PR #1049, opened 2020-11-16
Fix: userId should be set for hard crashes

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Fix: userId should be set for hard crashes

we got reports that hard crashes were not making to user count.

ran the App, hard crash, Opened the App, an event on Sentry has userId.

## :pencil: Checklist

- [X] All tests passing
- [ ] No breaking changes

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
@@ -143,7 +143,7 @@ final class DefaultAndroidEventProcessor implements EventProcessor {
 
     // userId should be set even if event is Cached as the userId is static and won't change anyway.
     if (event.getUser() == null) {
-      event.setUser(getUser());
+      event.setUser(getDefaultUser());
     }
 
     if (event.getContexts().getDevice() == null) {
@@ -809,7 +809,12 @@ final class DefaultAndroidEventProcessor implements EventProcessor {
     return null;
   }
 
-  public @NotNull User getUser() {
+  /**
+   * Sets the default user which contains only the userId.
+   *
+   * @return the User object
+   */
+  public @NotNull User getDefaultUser() {
     User user = new User();
     user.setId(getDeviceId());
 
--- a/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
@@ -141,6 +141,11 @@ final class DefaultAndroidEventProcessor implements EventProcessor {
           event.getEventId());
     }
 
+    // userId should be set even if event is Cached as the userId is static and won't change anyway.
+    if (event.getUser() == null) {
+      event.setUser(getUser());
+    }
+
     if (event.getContexts().getDevice() == null) {
       event.getContexts().setDevice(getDevice());
     }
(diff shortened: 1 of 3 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -141,6 +141,11 @@ public DefaultAndroidEventProcessor(
           event.getEventId());
     }
 
+    // userId should be set even if event is Cached as the userId is static and won't change anyway.
```
bruno-garcia, line 144: Is it static though? I mean it is if we use ANDROID_ID or the installation ID, but if it was a user_id defined by the client, it could be a different one after the restart.

bruno-garcia (changes requested):
I think it's the right solution, but the approach would need to change.

We should check if the current user is based on the static values such as ANDROID_ID or installation ID, because if this was in fact set by the user programatically, it would end up reporting the wrong user in here, right?

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
marandaneto: `ANDROID_ID` won't change, only if uninstall & install the App.
`installationID` is only for sessions.

if it was an user defined by the client, it won't overwrite it, not sure if I follow your comment.

marandaneto:
> We should check if the current user is based on the static values such as ANDROID_ID or installation ID, because if this was in fact set by the user programatically, it would end up reporting the wrong user in here, right?

not sure if I follow this user case, can you elaborate?

maciejwalkowiak (approved):
Looks good to me.

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
bruno-garcia: we're setting the whole user object here though after restart

bruno-garcia:
> not sure if I follow this user case, can you elaborate?

I wouldn't blindly copy the current scope's User into the event given that we don't know what that looks like (did the user set a customer user there programatically?) 

It's OK to set the ANDROID_ID or even the installation_id (yes today only used for sessions but that's for historical reasons, in hindsight we'd never have used ANDROID_ID)

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
bruno-garcia: I see now that getUser actually creates a brand new users and sets default values.

Could we rename this to `getDefaultUser` or something more clear?

bruno-garcia (approved):
The fix makes sense, I was confused because of the function name `getUser` which doesn't "just get a user" but actually creates a brand new user with the default Id, and hence has no risk of leaking a user-defined user object into it.

I propose renaming that private method to `getDefaultUser` to be explicit on intent

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
marandaneto: yep, done

PR #1058, opened 2020-11-21
Android Performance Experiment

bruno-garcia:
Goes without saying that'll never merge this.
But the idea is to discuss the API and what hooks we'll get

1. We're not applying device data here. For Java backend it makes sense to me because server info on each transaction might be just noise, but on Android I believe we really need it.

image

image

### DIFF
--- a/sentry-samples/sentry-samples-android/src/main/java/io/sentry/sample/MainActivity.java
+++ b/sentry-samples/sentry-samples-android/src/main/java/io/sentry/sample/MainActivity.java
@@ -2,8 +2,7 @@ package io.sentry.sample;
 
 import android.os.Bundle;
 import androidx.appcompat.app.AppCompatActivity;
-import io.sentry.Sentry;
-import io.sentry.UserFeedback;
+import io.sentry.*;
 import io.sentry.protocol.SentryId;
 import io.sentry.protocol.User;
 import io.sentry.sample.databinding.ActivityMainBinding;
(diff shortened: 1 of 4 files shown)


sentry-samples/sentry-samples-android/src/main/java/io/sentry/sample/MainActivity.java
```diff
@@ -74,18 +87,39 @@ protected void onCreate(Bundle savedInstanceState) {
...
+    innerSpan.finish();
 
+    innerSpan = activityTransaction.startChild();
+    innerSpan.setOperation("view");
+    innerSpan.setDescription("setContentView");
     setContentView(binding.getRoot());
+    innerSpan.setStatus(SpanStatus.OK);
+    innerSpan.finish();
+
+    activityTransaction.finish();
```
bruno-garcia, line 122: I thought in Python and other SDKs, they keep the reference to the Hub and flush itself on `finish()` ?

sentry-samples/sentry-samples-android/src/main/java/io/sentry/sample/MainActivity.java
```diff
@@ -74,18 +87,39 @@ protected void onCreate(Bundle savedInstanceState) {
...
+          }
         });
+    innerSpan.setStatus(SpanStatus.OK);
+    innerSpan.finish();
 
+    innerSpan = activityTransaction.startChild();
+    innerSpan.setOperation("view");
+    innerSpan.setDescription("setContentView");
     setContentView(binding.getRoot());
+    innerSpan.setStatus(SpanStatus.OK);
```
bruno-garcia, line 119: Would it make sense to have a wrapper that will do:

```java
void runTransaction(String name, String opertion, Callable code) {
var transaction = startTransaction();
transaction.operation = operation;
try {
code.Invoke();
transaction.setState(OK);
} catch {
transaction.setState(broken);
}
}


```

marandaneto: that's a good idea.
or annotation processors for annotated methods

sentry-samples/sentry-samples-android/src/main/java/io/sentry/sample/MainActivity.java
```diff
@@ -74,18 +87,39 @@ protected void onCreate(Bundle savedInstanceState) {
...
+    innerSpan.finish();
+
+    activityTransaction.finish();
+    Sentry.captureTransaction(activityTransaction, null);
```
marandaneto, line 123: we should not require a `, null`, so an overload will be handy.

sentry-samples/sentry-samples-android/src/main/java/io/sentry/sample/MainActivity.java
maciejwalkowiak: As you said, calling `finish` calls `Hub#captureTransaction`.

sentry-samples/sentry-samples-android/src/main/java/io/sentry/sample/MainActivity.java
maciejwalkowiak: I like the idea of having this method.

sentry-samples/sentry-samples-android/src/main/java/io/sentry/sample/MainActivity.java
maciejwalkowiak: I'll add it.

Issue #952, opened 2020-09-29
Figure out pushing a package to bintray via API


bruno-garcia:
Depends on #950 #951

* get the artifacts from zeus (#951) which craft does automatically.
* unzip the artifact to get the individual files
  * It'll be one artifact per package (like `io.sentry_sentry.zip`, `io.sentry_sentry-android.zip`)
* pgp sign them (expected the local machine to have the key)
* it's OK to expect the device has environment variables set with API keys
* Push the files to maven central via an API

Something like:

```sh
curl -v -F r=central -F hasPom=true -F e=jar -F file=@pom.xml -F file=@project-1.0.jar -u admin:admin123 http://localhost:8081/nexus/service/local/artifact/maven/content
```

The ultimate goal is to create a craft target to maven central: 
https://github.com/getsentry/craft/tree/master/src/targets

This target could be invoking a command (like `mvn`, or ./gradle task where task is somethign in this repo written in Kotlin) but ideally it would just be written there in TypeScript.

We can push to bintray only, which does the sync now that the repositories got sorted out. Pushing directly to jcenter and maven central will be more complex and now bintray is sorted to sync all Java and Android packages.

Some resources:
https://central.sonatype.org/pages/ossrh-guide.html
https://support.sonatype.com/hc/en-us/articles/213465818-How-can-I-programmatically-upload-an-artifact-into-Nexus-2-

bruno-garcia:
We really need this and #950 and #951 so we can get our docs and wizard pull the latest version automatically including the other benefits. Also craft will run as a service soon and SDK devs won't have keys to push to repositories either for compliance.

maciejwalkowiak:
Considering that my Bintray repo name is `github-packages-playground`, this is the CURL command that worked for me to upload a file to a new - `1.3-BUILD.SNAPSHOT` - version:

```curl -XPUT https://maciejwalkowiak:<api-key>@bintray.com/api/v1/maven/maciejwalkowiak/github-packages-playground/github-packages-playground/com/github/maciejwalkowiak/github-packages-playground/1.3-BUILD.SNAPSHOT/github-packages-playground-1.3-BUILD.SNAPSHOT.jar -o target/github-packages-playground-1.0-BUILD.SNAPSHOT.jar -vvv```

bruno-garcia:
🥇

Issue #951, opened 2020-09-29
Pull artifacts from CI/zeus to publish


bruno-garcia:
Dependent on #950, follow up #952

We want to use getsentry/craft to publish. That means splitting CI from publishing.
Once we get the artifacts (see #950) we would publish them to GH Actions (or zeus.ci).

With that, we can add the `.craft.yaml` to this repo so craft pulls the artifacts in order to make a release.

i.e:

```
github:
  owner: getsentry
  repo: sentry-java
targets:
  - name: maven
  - name: jcenter
  - name: github
  - name: registry
    type: sdk
    config:
      canonical: "maven:io.sentry:sentry"
 ....
```
For example in [sentry-cocoa](https://github.com/getsentry/sentry-cocoa/blob/master/.craft.yml), [sentry-dotnet](https://github.com/getsentry/sentry-dotnet/blob/main/.craft.yml), [sentry-react-native](https://github.com/getsentry/sentry-react-native/blob/master/.craft.yml).

marandaneto:
@maciejwalkowiak could be closed as well right?

maciejwalkowiak:
Yup. I wish moving cards on board also changes issue state.

bruno-garcia:
💯

PR #1062, opened 2020-11-24
name artifact so craft can find it

bruno-garcia:
#951 

With this we just need to get the info from #952 and add a craft target `bintray`


bruno-garcia:
Craft should be able to find the artifact now

Issue #984, opened 2020-10-15
Thread info is not present with 3.1.0 (Android-ANR)


ItsReddi:
_Platform:_
- [x]  Android

_IDE:_
- [x]  Android Studio

_Build system:_
- [x]  Gradle

_Sentry Android Gradle Plugin:_
- [x]  Yes
- [ ]  No

_Proguard/R8:_
- [x]  Enabled
- [ ]  Disabled

The version of the SDK:
**3.1.0**

---
I have the following issue:

After upgrading from `2.3.2` to `3.1.0` ANR reports do not include any thread informations.

**Steps to reproduce:**
- 1. use 2.3.2
- 2. example ANR
```
if (true) {
                    int a = 0;
                    while (true) {
                        a++;
                    }
                }
```
Result:
![grafik](https://user-images.githubusercontent.com/3147888/96109995-90840880-0edf-11eb-8c91-a7c9988dacd5.png)

- 3. switch to `3.1.0`
Result:
![grafik](https://user-images.githubusercontent.com/3147888/96110108-b4dfe500-0edf-11eb-88aa-efdd09335890.png)

marandaneto:
thanks for raising this, would you mind to try again but enabling `SentryOptions.attachThreads`? thanks

marandaneto:
this should have done the job: https://github.com/getsentry/sentry-java/blob/5ebd49cbfd17eb66175e9247db6d27e4df84a56b/sentry-android-core/src/main/java/io/sentry/android/core/ApplicationNotResponding.java#L24
I mean it does, but it doesn't look like having the whole stack traces or it's reporting the capturing thread but not the UI one.

ItsReddi:
`SentryOptions.attachThreads` does not seem to make a difference.
Still no threads attached on ANR.
but im unsure it is correct how it is enabled?

![grafik](https://user-images.githubusercontent.com/3147888/96220067-5d498400-0f88-11eb-9854-b57cc5fb132f.png)

marandaneto:
@ItsReddi this config is not available on the manifest config.

you need to: https://docs.sentry.io/platforms/android/usage/advanced-usage/#manual-instrumentation
and do `options.setAttachThreads(true)`

ItsReddi:
I can confirm it works with `setAttachThreads(true)`

marandaneto:
thanks, @ItsReddi we'll be looking into it, this flag should not have interfered in the ANR threads/stack traces if I am guessing it right.
[...]

PR #1064, opened 2020-11-24
Fix: Set current thread only if theres no exceptions

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Fix: Set current thread only if theres no exceptions

We'd like to set the default (current) thread if `attachStacktrace` is enabled and there are no exceptions only, exceptions have its own stack traces already.
one of the use cases for this is events captured by hybrid SDKs like React-native or Flutter, the current thread will be always the File observer, and it doesn't make much sense.

## :pencil: Checklist

- [X] All tests passing
- [ ] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 # vNext
 
 * Enhancement: Load `sentry.properties` from the application's current working directory (#1046)
-* ref: Refactor JSON deserialization (#1047)
+* Ref: Refactor JSON deserialization (#1047)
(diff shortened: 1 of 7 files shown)


marandaneto:
This also helps with #984

maciejwalkowiak (approved):
LGTM

bruno-garcia (approved):
Thanks!

Issue #981, opened 2020-10-14
SENTRY_TAGS environment variable doesn't work after upgrading from 1.7.30 to 3.0.0


bbednarek:
_Platform:_
- [ ]  Android -> If yes, which Device API (and compileSdkVersion/targetSdkVersion/Build tools) version?
- [X]  Java -> If yes, which Java (and sourceCompatibility/targetCompatibility) version? 
```
    targetCompatibility = sourceCompatibility = 14
```
- [ ]  Kotlin -> If yes, which Kotlin (and jvmTarget) version?
- [ ]  NDK -> If yes, which NDK/CMake version?
- [ ]  React-Native -> If yes, which version?
- [ ]  Timber -> If yes, which version?
- [ ]  Log4j -> If yes, which version?
- [X]  Logback -> If yes, which version? `1.2.3`
- [x]  Spring -> If yes, which version? `2.3.4.RELEASE`

_IDE:_
- [ ]  Android Studio -> If yes, which version?
- [X]  IntelliJ -> If yes, which version? `2020.2.2`
- [ ]  Other -> If yes, which one?

_Build system:_
- [ ]  Gradle -> If yes, which version?
- [ ]  Buck -> If yes, which version?
- [ ]  Bazel -> If yes, which version?
- [ ]  Maven -> If yes, which version?
- [ ]  Other -> If yes, which one?

_Android Gradle Plugin:_
- [ ]  Yes -> If yes, which version?
- [ ]  No

_Sentry Android Gradle Plugin:_
- [ ]  Yes -> If yes, which version?
- [ ]  No

_Proguard/R8:_
- [ ]  Enabled
- [ ]  Disabled

_Platform installed with:_
- [X] JCenter
- [ ] Bintray
- [ ] Maven Central
- [ ] Manually

The version of the SDK:
**3.0.0**

---
I have the following issue:

We used to set following environment variable
```
"SENTRY_TAGS": "service:%s" %(self.name),
```
and tags were available in sentry 
![Screenshot 2020-10-14 at 17 10 06](https://user-images.githubusercontent.com/895578/96008784-33854580-0e40-11eb-8c24-26ac9dffddc8.png)

and after upgrading to version `3.0.0` this variable does't work any more.
![Screenshot 2020-10-14 at 17 11 43](https://user-images.githubusercontent.com/895578/96008932-5a437c00-0e40-11eb-87f6-0836efb69fd7.png)

**Steps to reproduce:**
Upgrade 
```
[...]

PR #1066, opened 2020-11-24
Set tags on SentryOptions through external configuration

maciejwalkowiak:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

Add an option to set tags on SentryOptions through external configuration.

Fixes gh-981. Agreed with @bruno-garcia to add `tags` to `SentryOptions`.

Unit & integration tests.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/sentry/src/main/java/io/sentry/MainEventProcessor.java
+++ b/sentry/src/main/java/io/sentry/MainEventProcessor.java
@@ -5,6 +5,7 @@ import io.sentry.util.ApplyScopeUtils;
 import io.sentry.util.Objects;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
@@ -85,6 +86,12 @@ public final class MainEventProcessor implements EventProcessor {
     }
     if (event.getSdk() == null) {
       event.setSdk(options.getSdkVersion());
+    }
+
+    for (Map.Entry<String, String> tag : options.getTags().entrySet()) {
+      if (event.getTag(tag.getKey()) == null) {
+        event.setTag(tag.getKey(), tag.getValue());
+      }
     }
 
     if (event.getThreads() == null) {
--- a/sentry/src/main/java/io/sentry/MainEventProcessor.java
+++ b/sentry/src/main/java/io/sentry/MainEventProcessor.java
@@ -88,7 +88,7 @@ public final class MainEventProcessor implements EventProcessor {
       event.setSdk(options.getSdkVersion());
     }
 
-    for (Map.Entry<String, String> tag : options.getTags().entrySet()) {
+    for (final Map.Entry<String, String> tag : options.getTags().entrySet()) {
       if (event.getTag(tag.getKey()) == null) {
         event.setTag(tag.getKey(), tag.getValue());
       }
--- a/sentry/src/main/java/io/sentry/SentryOptions.java
+++ b/sentry/src/main/java/io/sentry/SentryOptions.java
@@ -11,7 +11,10 @@ import io.sentry.transport.NoOpTransport;
 import io.sentry.transport.NoOpTransportGate;
 import java.io.File;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.SSLSocketFactory;
@@ -240,6 +243,9 @@ public class SentryOptions {
    */
   private boolean enableExternalConfiguration;
 
+  /** Tags applied to every event and transaction */
+  private @NotNull Map<String, String> tags = new ConcurrentHashMap<>();
+
   /**
    * Creates {@link SentryOptions} from properties provided by a {@link PropertiesProvider}.
    *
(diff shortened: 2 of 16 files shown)


sentry/src/main/java/io/sentry/MainEventProcessor.java
```diff
@@ -87,6 +88,12 @@ private void processNonCachedEvent(SentryEvent event) {
       event.setSdk(options.getSdkVersion());
     }
 
+    for (Map.Entry<String, String> tag : options.getTags().entrySet()) {
```
marandaneto, line 91:
```suggestion
    for (final Map.Entry<String, String> tag : options.getTags().entrySet()) {
```

sentry/src/main/java/io/sentry/MainEventProcessor.java
```diff
@@ -87,6 +88,12 @@ private void processNonCachedEvent(SentryEvent event) {
...
+      if (event.getTag(tag.getKey()) == null) {
```
marandaneto, line 92: what's about `containsKey` instead?

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -240,6 +243,9 @@
    */
   private boolean enableExternalConfiguration;
 
+  /** Tags applied to every event and transaction */
+  private @NotNull Map<String, String> tags = new ConcurrentHashMap<>();
```
marandaneto, line 247: it doesn't need to be a thread-safe Map IMO

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -1118,6 +1125,34 @@ public void setEnableExternalConfiguration(boolean enableExternalConfiguration)
...
+    return tags;
+  }
+
+  /**
+   * Sets tags applied to all events and transactions.
+   *
+   * @param tags the tags map
+   */
+  public void setTags(final @NotNull Map<String, String> tags) {
+    this.tags = tags;
```
marandaneto, line 1143: one could set `null` or a non-thread-safe list, what's about only exposing the `setTag(key, value)`?

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -1203,6 +1238,10 @@ void merge(final @NotNull SentryOptions options) {
     if (options.getProxy() != null) {
       setProxy(options.getProxy());
     }
+    final Map<String, String> tags = new HashMap<>(options.getTags());
+    for (Map.Entry<String, String> tag : tags.entrySet()) {
```
marandaneto, line 1242:
```suggestion
    for (final Map.Entry<String, String> tag : tags.entrySet()) {
```

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -1203,6 +1238,10 @@ void merge(final @NotNull SentryOptions options) {
...
+      this.tags.put(tag.getKey(), tag.getValue());
```
marandaneto, line 1243: should we define precedence here? if both have the same `keys`, which one should win?

sentry/src/main/java/io/sentry/SentryOptions.java
marandaneto: we can make it `final` if we don't expose a setter, but only setTag(key, value)

sentry/src/main/java/io/sentry/config/AbstractPropertiesProvider.java
```diff
@@ -0,0 +1,58 @@
...
+    this.prefix = Objects.requireNonNull(prefix, "prefix is required");
+    this.properties = Objects.requireNonNull(properties, "properties are required");
+  }
+
+  protected AbstractPropertiesProvider(final @NotNull Properties properties) {
+    this("", properties);
+  }
+
+  @Override
+  public @Nullable String getProperty(@NotNull String property) {
```
marandaneto, line 29:
```suggestion
  public @Nullable String getProperty(final @NotNull String property) {
```

sentry/src/main/java/io/sentry/config/CompositePropertiesProvider.java
```diff
@@ -26,4 +28,15 @@ public CompositePropertiesProvider(@NotNull List<PropertiesProvider> providers)
     }
     return null;
   }
+
+  @Override
+  public @NotNull Map<String, String> getMap(@NotNull String property) {
```
marandaneto, line 33:
```suggestion
  public @NotNull Map<String, String> getMap(final @NotNull String property) {
```

sentry/src/main/java/io/sentry/SentryOptions.java
maciejwalkowiak: The one that is passed as a parameter. External configuration overwrites the configuration provided in the code. This precedence is verified in unit tests.

sentry/src/main/java/io/sentry/MainEventProcessor.java
maciejwalkowiak: This would be then:
```java
if (event.getTags() == null || !event.getTags().containsKey(tag.getKey())) {
    event.setTag(tag.getKey(), tag.getValue());
}
```
I think I prefer the original version.

marandaneto: true, `event.getTags()` can be null, my bad

marandaneto (approved):
LGTM

PR #1071, opened 2020-11-26
docs: Fix comment for cacheDirSize

philipphofmann:
The default for `cacheDirSize` is 30. The comment for the getter and setter stated 10 instead
of 30. This is fixed now.


maciejwalkowiak (approved):
Thanks!

PR #1082, opened 2020-12-01
feat: Add basic support for attachments

philipphofmann:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Add basic support for attachments.

Unit tests and with a sample project.

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes


sentry/src/main/java/io/sentry/Scope.java
```diff
@@ -55,6 +55,9 @@
   /** Scope's contexts */
   private @NotNull Contexts contexts = new Contexts();
 
+  /** Scope's attachments */
+  private @NotNull List<Attachment> attachments = new ArrayList<>();
```
bruno-garcia, line 59: `ArrayList` isnt' thread safe

philipphofmann: Good point thanks @bruno-garcia

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/MainActivity.java
```diff
@@ -35,6 +41,18 @@ protected void onCreate(Bundle savedInstanceState) {
...
+    binding.sendAttachment.setOnClickListener(
+        view -> {
+          Thread thread =
+              new Thread() {
+                public void run() {
+                  captureExceptionWithAttachments();
+                }
+              };
+          // Fire and forget
+          thread.start();
```
marandaneto, line 53: L: why do we use a Thread?

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/MainActivity.java
```diff
@@ -88,4 +106,30 @@ protected void onCreate(Bundle savedInstanceState) {
 
     setContentView(binding.getRoot());
   }
+
+  private void captureExceptionWithAttachments() {
+    Sentry.withScope(
```
marandaneto, line 111: L: do we need to use a `Sentry.withScope`? we try to avoid them on Mobile at least

sentry/src/main/java/io/sentry/Attachment.java
```diff
@@ -0,0 +1,109 @@
...
+import org.jetbrains.annotations.Nullable;
+
+/** You can use an attachment to store additional files alongside an event or transaction. */
+@ApiStatus.Experimental
+public final class Attachment {
+
+  private @Nullable byte[] bytes;
+  private @Nullable String path;
+  private final String filename;
+  private String contentType;
```
marandaneto, line 15: M: if they are supposed to be NotNull, let's add an annotation as well, it's easier to handle on Kotlin

sentry/src/main/java/io/sentry/Attachment.java
```diff
@@ -0,0 +1,109 @@
...
+  /**
+   * Initializes an Attachment with bytes.
+   *
+   * <p>The file located at the path is read lazily when the SDK captures an event or transaction
+   * not when the attachment is initialized.
+   *
+   * @param bytes The bytes of file.
+   * @param filename The name of the attachment to display in Sentry.
+   */
+  public Attachment(@NotNull byte[] bytes, @NotNull String filename) {
```
marandaneto, line 35:
```suggestion
  public Attachment(final @NotNull byte[] bytes, final @NotNull String filename) {
```

and any other missing param

sentry/src/main/java/io/sentry/Attachment.java
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  /**
+   * Gets the content type of the attachment. Default is "application/octet-stream".
+   *
+   * @return the content type.
+   */
+  public String getContentType() {
+    return contentType;
+  }
```
marandaneto, line 99: M: @NotNull

sentry/src/main/java/io/sentry/Attachment.java
```diff
@@ -0,0 +1,109 @@
...
+
+  /**
+   * Sets the content type of the attachment. Default is "application/octet-stream".
+   *
+   * @param contentType the content type of the attachment.
+   */
+  public void setContentType(String contentType) {
```
marandaneto, line 106: @NotNull or @Nullable

sentry/src/main/java/io/sentry/Scope.java
```diff
@@ -430,6 +435,28 @@ public void removeContexts(final @NotNull String key) {
     contexts.remove(key);
   }
 
+  /**
+   * Returns the Scopes's attachments
+   *
+   * @return the attachments
+   */
+  @ApiStatus.Experimental
```
marandaneto, line 443: L: fine keeping it, but it's a package-private method, so it's not supposed to be used anyway.

sentry/src/main/java/io/sentry/Scope.java
```diff
@@ -430,6 +435,28 @@ public void removeContexts(final @NotNull String key) {
...
+  }
+
+  /**
+   * Adds an attachment to the Scope's list of attachments. The SDK adds the attachment to every
+   * event and transaction sent to Sentry.
+   *
+   * @param attachment The attachment to add to the Scope's list of attachments.
+   */
+  @ApiStatus.Experimental
+  public void addAttachment(Attachment attachment) {
```
marandaneto, line 456:
```suggestion
  public void addAttachment(final @NotNull Attachment attachment) {
```

sentry/src/main/java/io/sentry/Scope.java
```diff
@@ -495,6 +522,7 @@ public void removeContexts(final @NotNull String key) {
     clone.extra = extraClone;
 
     clone.contexts = contexts.clone();
+    clone.attachments = new CopyOnWriteArrayList<>(attachments);
```
marandaneto, line 525: M: We might need to clone the Objects too, we do with breadcrumbs

sentry/src/main/java/io/sentry/SentryClient.java
```diff
@@ -124,13 +124,27 @@ public SentryClient(final @NotNull SentryOptions options, @Nullable Connection c
     return sentryId;
   }
 
-  private @Nullable SentryEnvelope buildEnvelope(final @Nullable SentryBaseEvent event)
+  private @NotNull List<Attachment> getAttachmentsFromScope(@Nullable Scope scope) {
+    List<Attachment> attachments;
+    if (scope != null) {
+      attachments = scope.getAttachments();
+    } else {
+      attachments = new ArrayList<>();
```
marandaneto, line 132: M: if the `buildEnvelope` accepts a nullable `List<Attachment>`, we'd not need to allocate this, right?

sentry/src/main/java/io/sentry/SentryClient.java
```diff
@@ -148,6 +162,14 @@ public SentryClient(final @NotNull SentryOptions options, @Nullable Connection c
       envelopeItems.add(sessionItem);
     }
 
+    if (attachments != null) {
+      for (Attachment attachment : attachments) {
```
marandaneto, line 166:
```suggestion
      for (final Attachment attachment : attachments) {
```

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
```diff
@@ -148,6 +149,50 @@ public static SentryEnvelopeItem fromUserFeedback(
...
+                } catch (IOException | SecurityException exception) {
+                  logger.log(
+                      SentryLevel.ERROR,
+                      exception,
+                      "Serializing attachment %s failed.",
+                      attachment.getFilename());
+                  outputStream = new ByteArrayOutputStream();
+                }
+
+                return outputStream.toByteArray();
```
marandaneto, line 180: M: a few things here concerns me, if one of the `write` throws, we still return a portion of it with `outputStream.toByteArray()`, not sure if this is a good idea, I am all in for either the whole thing or nothing.

`ByteArrayOutputStream` could be created inside of the `try-with-resources` so it's gonna be deallocated/closed automatically.

Might be possible to wrap the `FileInputStream` with `BufferedInputStream`, usually Buffered solutions would work better on Mobile

marandaneto:
M: Question, does the `deserializeEnvelope` work ootb for cached envelopes? (with attachments?)

Right now envelopes with attachments created by sentry-native (C/C++ layer) won't work, this would require changes on the `OutboxSender` class, are we aware of this limitation? the OutboxSender is only able to process/capture SentryEvents for now, so if we roll up this feature, we should document this limitation even if sentry-native supports it.

sentry/src/test/java/io/sentry/SentryEnvelopeItemTest.kt
```diff
@@ -20,7 +42,141 @@ class SentryEnvelopeItemTest {
...
+        verify(logger)
+            .log(eq(SentryLevel.ERROR), any<T>(),
+                eq("Serializing attachment %s failed."), eq(filename))
+    }
+
+    private inline fun <reified T : Any> T.injectForField(name: String, value: Any?) {
+        T::class.java.getDeclaredField(name)
+            .apply { isAccessible = true }
+            .set(this, value)
+    }
```
marandaneto, line 170: L: we could extract it to its own Util class and reuse it, I see this happening more often after making the usage of @NotNull and @Nullable
`AndroidOptionsInitializerTest` for example, but its another Module so we can't reuse it, just to keep in mind.

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/MainActivity.java
philipphofmann: Because writing a 2 MB file blocks the UI thread for a bit. Would you prefer to introduce a proper way of running tasks in the background, like `ExecutorService`, Coroutines, Rx, ...?

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/MainActivity.java
philipphofmann: I just want to send the attachment with the exception and not all events. If I add the attachment to the global scope it is going to be sent with every event. I'm also fine with doing that.

sentry/src/main/java/io/sentry/Attachment.java
philipphofmann: Do you know if there is a linter for this? So you don't have to make this comment.

sentry/src/main/java/io/sentry/Scope.java
philipphofmann: I'm going to remove it.

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/MainActivity.java
marandaneto: mm, I'd say adding a comment would be good enough, just to not bound that attachments feature 'require' a Thread or so, I mean, it's highly recommended, but not really needed

sentry/src/main/java/io/sentry/Scope.java
philipphofmann: Makes sense

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/MainActivity.java
marandaneto: maybe adding a comment, because that means none of the other Scope fields will be sent either, if people just copy and paste, other fields will be missing.

sentry/src/main/java/io/sentry/Attachment.java
marandaneto: no afaik, that's why I try to be the human linter about that, Kotlin fails compilation for that, but not on Java side.

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/MainActivity.java
philipphofmann: Ah, now I get your point. Didn't think about users might look this up and think we need a thread for this. I'm going to add a comment.

sentry/src/main/java/io/sentry/Attachment.java
philipphofmann: If you say Kotlin fails compilation for that. What do you exactly mean?

sentry/src/main/java/io/sentry/SentryClient.java
philipphofmann: I usually prefer passing an empty list around instead of `null`, because I don't like `null`. I could change it to `new ArrayList<>(0)` so we don't allocate too much memory, but returning `null` saves us even more memory. I'm going to change it.

sentry/src/main/java/io/sentry/SentryClient.java
philipphofmann: Again manual (Manoel 😁) linter. Would be nice to have an automatic linter for that.

sentry/src/main/java/io/sentry/Attachment.java
marandaneto: if a Java method is annotated with @NotNull or @Nullable, when Kotlin code is run against those annotations, they use the null safety feature accordingly, so you can't pass a `null` value to a `@NotNull` Java method, unless you use reflection.
you should know that, as you've written tests using reflection :D

sentry/src/test/java/io/sentry/SentryEnvelopeItemTest.kt
philipphofmann: I strongly agree. What about a project private test-util Gradle module? I used them heavily in the past and it made my life and the life of many other devs way easier.

sentry/src/main/java/io/sentry/SentryClient.java
marandaneto: either way is fine for me, but it's not fine to return an empty instance and still, the other method accepts null and do a NPE check, ideally, either we null-safety using `new ArrayList<>(0)` and `@NotNull` for `buildEnvelope` or we just return null and let `buildEnvelope` deal with the nullability pain.
the current way is doing an extra allocation + extra NPE check for nothing.

sentry/src/main/java/io/sentry/Attachment.java
philipphofmann: Now I get your point. I thought you meant that Kotlin would fail if I didn't add the annotation. That would be news for me.

sentry/src/main/java/io/sentry/SentryClient.java
marandaneto: Java is verbose, unless we write a custom lint, there's not one for this.
I like `val` in Kotlin, so that's why I use `final` in Java for everything, or do you use `var` for everything in Kotlin? :P

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
philipphofmann: Oh my! Was I out of my mind 🙈? Thanks for the important input. I'm going to change this.

Where do you have the info from, that buffered solutions work better on mobile?

sentry/src/test/java/io/sentry/SentryEnvelopeItemTest.kt
marandaneto: it already exists, its called `sentry-test-support`, it was midnight so totally forgot about it :P
this module was created afterwards, so a lot of things could be moved in there

sentry/src/test/java/io/sentry/SentryEnvelopeItemTest.kt
```diff
@@ -20,7 +41,138 @@ class SentryEnvelopeItemTest {
         }
     }
 
+    @Test
+    fun `fromAttachment with bytes`() {
```
marandaneto, line 45: H: I'd add a test specifically for non ascii file content like a binary or so, to be sure this won't break the parsing of the envelope either on serialization nor deserialization.
this file could exist in the `test/resources` folder, maybe just download a random exported `sqlite` database from the webz?

sentry/src/main/java/io/sentry/Attachment.java
```diff
@@ -0,0 +1,118 @@
...
+    this.bytes = bytes;
+    this.filename = filename;
+    this.contentType = DEFAULT_CONTENT_TYPE;
+  }
+
+  /**
+   * Initializes an Attachment with a path. The filename of the file located at the path is used.
+   *
+   * @param path The path of the file to upload as an attachment.
+   */
```
marandaneto, line 45: L: maybe include that we also require an `absolute` path? or rename the String param to make it obvious

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/MainActivity.java
philipphofmann: I removed the Thread and just create a small file.

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
```diff
@@ -148,6 +150,49 @@ public static SentryEnvelopeItem fromUserFeedback(
...
+            });
+
+    SentryEnvelopeItemHeader itemHeader =
+        new SentryEnvelopeItemHeader(
+            SentryItemType.Attachment,
+            () -> cachedItem.getBytes().length,
+            attachment.getContentType(),
+            attachment.getFilename());
+
+    return new SentryEnvelopeItem(itemHeader, cachedItem::getBytes);
```
marandaneto, line 193: H: `::` can cause issues on Android.
not all AGP versions do the method references translation properly

See: DefaultAndroidEventProcessor
there's a comment about it.

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
```diff
@@ -148,6 +150,49 @@ public static SentryEnvelopeItem fromUserFeedback(
     return new SentryEnvelopeItem(itemHeader, cachedItem::getBytes);
```
marandaneto, line 150: H: consider changing this too
https://github.com/getsentry/sentry-java/pull/1082/files#r535112346

marandaneto:
H: I ran a small test with a binary (sqlite database) file and got an error
See the PR https://github.com/getsentry/sentry-java/pull/1090

See the error: {"detail":"invalid event envelope","causes":["missing newline after header or payload"]}

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
philipphofmann: Thanks, didn't know that. Again a linter would be nice.

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
philipphofmann: Changed it here, thanks: https://github.com/getsentry/sentry-java/pull/1091

sentry/src/main/java/io/sentry/Attachment.java
philipphofmann: I renamed `path` to `pathname` and also updated the javadoc. This is more aligned with [`new File(String pathname)`](https://docs.oracle.com/javase/8/docs/api/java/io/File.html#File-java.lang.String-). It doesn't have to be an absolute path. Is it clearer now?

marandaneto: Files on Android are a bit messy, for example, a `pathname` from a taken picture simply doesn't work well if you don't resolve full address with the `MediaStore` intent classes (Uri etc), also it changes by OS versions.

An important note is that this should be run on a Java App as well which would be good enough already.

If people are familiar with the File APIs, it's good already, I was stressing only the docs.

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
```diff
@@ -151,6 +153,50 @@ public static SentryEnvelopeItem fromUserFeedback(
...
+  public static SentryEnvelopeItem fromAttachment(
+      @NotNull ILogger logger, final @NotNull Attachment attachment) {
+
+    final CachedItem cachedItem =
+        new CachedItem(
+            () -> {
+              if (attachment.getBytes() != null) {
+                return attachment.getBytes();
+              } else if (attachment.getPathname() != null) {
+                try (FileInputStream fileInputStream =
```
marandaneto, line 165: M: should we do a safe check for `isFile()` and `canRead()`? it's probably cheaper than `SecurityException`

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
marandaneto: It's not only on mobile but as we have limited resources, it's more important on mobile.
official docs already mention a bit about it https://docs.oracle.com/javase/tutorial/essential/io/buffers.html

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
```diff
@@ -151,6 +153,50 @@ public static SentryEnvelopeItem fromUserFeedback(
...
+                        new FileInputStream(attachment.getPathname());
+                    BufferedInputStream inputStream = new BufferedInputStream(fileInputStream);
+                    ByteArrayOutputStream outputStream = new ByteArrayOutputStream()) {
```
marandaneto, line 168: H: we could still wrap the outputStream to a buffered writer, see `fromSession` example

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
```diff
@@ -151,6 +153,50 @@ public static SentryEnvelopeItem fromUserFeedback(
...
+                  int offset = 0;
+                  while ((length = inputStream.read(bytes)) != -1) {
+                    outputStream.write(bytes, offset, length);
+                  }
+                  return outputStream.toByteArray();
+                } catch (IOException | SecurityException exception) {
+                  logger.log(
+                      SentryLevel.ERROR,
+                      exception,
+                      "Serializing attachment %s failed.",
```
bruno-garcia, line 181:
```suggestion
                      "Reading the attachment %s failed.",
```

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
```diff
@@ -151,6 +153,50 @@ public static SentryEnvelopeItem fromUserFeedback(
...
+                      attachment.getPathname());
+                }
+              }
+
+              return new byte[0];
```
bruno-garcia, line 186: This means we'll be sending an 0 byte attachments when we have errors reading?
Just want to make sure we've thought this through, as it will be confusing to customers given no additional data to why this was a 0 byte file.

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
```diff
@@ -151,6 +153,50 @@ public static SentryEnvelopeItem fromUserFeedback(
...
+                    ByteArrayOutputStream outputStream = new ByteArrayOutputStream()) {
+
+                  byte[] bytes = new byte[1024];
+                  int length;
+                  int offset = 0;
+                  while ((length = inputStream.read(bytes)) != -1) {
+                    outputStream.write(bytes, offset, length);
+                  }
+                  return outputStream.toByteArray();
+                } catch (IOException | SecurityException exception) {
```
bruno-garcia, line 177: Could this throw OOM here too? Given that we'll have to allocate a byte array for the whole size of this file?

sentry/src/main/java/io/sentry/GsonSerializer.java
```diff
@@ -151,34 +151,30 @@ public GsonSerializer(
...
    *
    * @param envelope the SentryEnvelope
-   * @param writer the Writer
+   * @param stream the Stream
    * @throws IOException an IOException
    */
   @Override
-  public void serialize(final @NotNull SentryEnvelope envelope, final @NotNull Writer writer)
-      throws Exception {
+  public void serialize(SentryEnvelope envelope, OutputStream stream) throws Exception {
```
marandaneto, line 158:
```suggestion
  public void serialize(final @NotNull SentryEnvelope envelope, final @NotNull OutputStream stream) throws Exception {
```

M: double-check if there are more missing.

marandaneto: H: not sure if it's possible to use, as the OutputStream comes from the connection, but `BufferedOutputStream` exists

marandaneto: H: I'd rename to `outputStream` as `inputStream` also exists.

sentry/src/main/java/io/sentry/GsonSerializer.java
```diff
@@ -151,34 +151,30 @@ public GsonSerializer(
...
-      try (final BufferedReader reader =
-          new BufferedReader(
-              new InputStreamReader(new ByteArrayInputStream(item.getData()), UTF_8))) {
-        final char[] buffer = new char[1024];
-        int charsRead;
-        while ((charsRead = reader.read(buffer, 0, buffer.length)) > 0) {
-          writer.write(buffer, 0, charsRead);
-        }
-      }
+      for (SentryEnvelopeItem item : envelope.getItems()) {
```
marandaneto, line 166:
```suggestion
      for (final SentryEnvelopeItem item : envelope.getItems()) {
```

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
philipphofmann: As discussed on Friday, it doesn't really make sense to use a buffered writer, which is a character-based output stream, as we are dealing with bytes here and not characters as we do in `fromSession`.

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/MainActivity.java
```diff
@@ -35,6 +41,31 @@ protected void onCreate(Bundle savedInstanceState) {
...
+          } catch (IOException e) {
+            Sentry.captureException(e);
+          }
+
+          Sentry.configureScope(
+              scope -> {
+                String json = "{ \"number\": 10 }";
+                Attachment attachment = new Attachment(json.getBytes(), "log.json");
+                scope.addAttachment(attachment);
+                scope.addAttachment(new Attachment(file.getPath()));
```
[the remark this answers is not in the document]
philipphofmann: We can also add attachments with bytes, so I would keep it as it is.

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
marandaneto: if you both agreed, it's fine, I was not part of the discussion

sentry/src/main/java/io/sentry/Attachment.java
```diff
@@ -0,0 +1,120 @@
...
+   * Sets the content type of the attachment. Default is "application/octet-stream".
+   *
+   * @param contentType the content type of the attachment.
+   */
+  public void setContentType(final @NotNull String contentType) {
+    this.contentType = contentType;
+  }
+
+  @Override
+  protected @NotNull Object clone() throws CloneNotSupportedException {
```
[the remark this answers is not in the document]
philipphofmann: I think that is the use case yes. Currently, the only thing that you can change on the attachment is the `contentType`. So I think it makes to ditch the cloning here. What do you think about that @marandaneto?

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
philipphofmann: I changed it. Please have a look.

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
philipphofmann: Yes, it could. If you try to send a 10GB file. Anyway relay has a default 50Mib limit for envelopes, see https://docs.sentry.io/product/relay/options/. So we could anyway discard files that are bigger than that. 
I would rather avoid catching `OutOfMemoryError` and fix the underlying issue. We allocate memory in 1024 byte blocks inside a loop, if we run out of memory in there the whole application is going to be in trouble. If we would allocate all the memory for the file in one big chunk and we would get an `OutOfMemoryError` we could be fine, but we don't do that.

sentry/src/main/java/io/sentry/Scope.java
```diff
@@ -430,6 +435,27 @@ public void removeContexts(final @NotNull String key) {
...
   }
 
+  /**
+   * Returns the Scopes's attachments
+   *
+   * @return the attachments
+   */
+  @NotNull
+  List<Attachment> getAttachments() {
+    return attachments;
```
bruno-garcia, line 445: Here we need to return a new instance of a list instead of our private one

bruno-garcia (approved):
Lets chips and iterate! 🚀

PR #1083, opened 2020-12-02
build: Add make format to Makefile

philipphofmann:
## :scroll: Description
Call `make format` to format the code with spotless.

We already have `make checkFormat`.

Calling it on the command line.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,3 +35,6 @@ stop:
 
 checkFormat:
 	./gradlew spotlessJavaCheck spotlessKotlinCheck
+
+format:
+	./gradlew spotlessApply


marandaneto:
let's also update the git commit hook to use this `make` call then

https://github.com/getsentry/sentry-java/blob/main/hooks/pre-commit#L6-L7

philipphofmann:
> let's also update the git commit hook to use this `make` call then
> 
> https://github.com/getsentry/sentry-java/blob/main/hooks/pre-commit#L6-L7

I'm not sure if it is a good idea to modify files in a pre-commit hook. If we say it's a good idea, I'm not sure how to achieve this. Just calling `git add .` won't do the job as we might add unwanted unstaged changes.

maciejwalkowiak (approved):

marandaneto (approved):

PR #1084, opened 2020-12-02
Make InAppIncludesResolver public.

### LINKED ISSUE #1080: No sane way of configuring SSLSocketFactory in Spring Boot integration
_Platform:_
- [x]  Java 11
- [x]  Spring Boot 2.3.0

_IDE:_
- [x]  IntelliJ 2020.2

_Build system:_
- [x]  Gradle 6.3

_Platform installed with:_
- [x] Maven Central

The version of the SDK:
**3.2.0**

---
I have the following issue:

Trying to configure custom `SSLSocketFactory` with Sentry's Spring Boot integration, but there does not seem to be a sane way.

1. Creating an own `HttpTransport` bean, which is seemingly the suggested way, is not trivial. I would have used `HttpTransportFactory`, but it is package-private. It's added value is that it parses DSN from `SentryOptions` to extract credentials necessary for `HttpTransport`. Trying to recreate this logic is also difficult, because `Dsn` class, which does the parsing, is also package-private.

2. Provide own `Sentry.OptionsConfiguration<SentryOptions>` bean to be able to manually configure `SentryOptions` has one significant problem: I am not able to set the default `inAppIncludes`. Those are provided by auto-configured `InAppIncludesResolver`, which is package-private. It can be only configured within the autoconfiguration class.

Is there a way that I am missing?

maciejwalkowiak:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

Make InAppIncludesResolver public.

Fixes #1080.

In case users want to overwrite `OptionsConfiguration` before this change they do not have an option to use `InAppIncludesResolver`. 

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * Enhancement: Add more convenient method to start a child span (#1073)
 * Enhancement: Autoconfigure traces callback in Spring Boot integration (#1074)
 * Enhancement: Resolve in-app-includes and in-app-excludes parameters from the external configuration
+* Enhancement: Make InAppIncludesResolver public (#1084)
 
 # 4.0.0-alpha.1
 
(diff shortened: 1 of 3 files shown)


marandaneto (approved):
LGTM

Issue #1081, opened 2020-12-01
Adding breadcrumbs taking too much on Android


samiazar:
Recently I've decided to add breadcrumbs for getting more details on crash reports.
But I notice each time by using the `addBreadcrumb()` method it take too much time (about 10 ms) on the UI thread.

Here is how I try to measure the time of this method.

```
long time = System.currentTimeMillis();
Breadcrumb breadcrumb = new Breadcrumb();
breadcrumb.setCategory("Event");
breadcrumb.setMessage(s);
breadcrumb.setLevel(SentryLevel.INFO);
mScope.addBreadcrumb(breadcrumb);
Timber.w("Adding breadcrumbs take %s millis", System.currentTimeMillis() - time);
```

And here is the result:
<img width="496" alt="Screen Shot 1399-09-11 at 18 02 42" src="https://user-images.githubusercontent.com/23658371/100753719-69d25080-33ff-11eb-97d4-760adfeed9de.png">
.

I think it's not expected. For logging, I almost dropping a frame on the UI.
Do you have a plan to optimize this?
I really need this feature but now its cost is heavy

bruno-garcia:
Do you have a `beforeBreadcrumb` callback defined in your `SentryOptions`?
Is scope synchronization enabled? 

https://github.com/getsentry/sentry-java/blob/3ed2ff105dd246cbf0242e802543c710bcb14a0b/sentry/src/main/java/io/sentry/SentryOptions.java#L1106-L1108

What version of the SDK. Is this on a emulator, device? What device?

samiazar:
Thank you for your quick response.

The sentry version was 3.0.0 and I didn't have a `beforeBreadcrumb` callback and scope synchronization was disabled.
But after more investigating finally, I figure it out that the problem is about using `getCurrentDateTimeOrNull` method from `DateUtils` class.
https://github.com/getsentry/sentry-java/blob/3ed2ff105dd246cbf0242e802543c710bcb14a0b/sentry/src/main/java/io/sentry/Breadcrumb.java#L64

This method takes too much time. I try to use the constructor of the Breadcrumb that get `Date` object but this constructor was package-private. 
After all, by updating the sentry version into 3.2.0 I see this constructor being public.
So I creating breadcrumb with a simple `Date` object. (just `new Date()`)
My issue is resolved but maybe you want to optimize the `getCurrentDateTimeOrNull` for the future.

All these benchmarking had been tested on the Samsung-j7 prime.

[...]

PR #1085, opened 2020-12-02
Optimize DateUtils performance.

maciejwalkowiak:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

Improves `DateUtils` performance. See benchmark results here: https://github.com/maciej-scratches/simple-date-format-benchmark.

While this improves the current state, we can consider also opting-out from using SDF completely and rely on `Calendar` class instead similar to: https://askldjd.wordpress.com/2013/03/04/simpledateformat-is-slow/

I noticed getting current timestamp in UTC takes significant amount of CPU time while running benchmarks for the performance feature. 

This class has been also pointed in https://github.com/getsentry/sentry-java/issues/1081

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/sentry/src/main/java/io/sentry/DateUtils.java
+++ b/sentry/src/main/java/io/sentry/DateUtils.java
@@ -17,18 +17,30 @@ public final class DateUtils {
   private static final String ISO_FORMAT = "yyyy-MM-dd'T'HH:mm:ss'Z'";
   private static final String ISO_FORMAT_WITH_MILLIS = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'";
   private static final ThreadLocal<SimpleDateFormat> SDF_ISO_FORMAT_WITH_MILLIS_UTC =
-      ThreadLocal.withInitial(
-          () -> {
-            final TimeZone tz = TimeZone.getTimeZone(UTC);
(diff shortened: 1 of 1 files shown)


marandaneto:
I'd love to use the new Java Date API from Java 8, but this would break a lot of Android devices :(

maciejwalkowiak:
Are we taking thi decision anytime soon? If not, i believe we should optimize this class either with ThreadLocal or Calendar class.

marandaneto:
> Are we taking thi decision anytime soon? If not, i believe we should optimize this class either with ThreadLocal or Calendar class.

nop, we can't break older OS versions, If the `ThreadLocal` version already speeds up, let's go for it.
Calendar is also broken if using UTC, it'll need workarounds too so not sure if it's gonna help anyway, let's go for ThreadLocal

marandaneto (approved):
LGTM

bruno-garcia:
could we introduce an abstraction and inject a different one for Android?
So we can have the non shitty version for Java then?

marandaneto:
> could we introduce an abstraction and inject a different one for Android?
> So we can have the non shitty version for Java then?

that sounds reasonable.

Issue #993, opened 2020-10-16
Support two instances of sentry client 


alexandresoler:
Hey guys,

I'm developing an Android SDK and I'm using the Sentry version 1.7.22
I'm getting my personal instance of Sentry with val mySentry = AndroidSentryClientFactory.sentryClient(sentryDSN)
Then, my integrator can use his Sentry in his code with the standard SentryClient with Sentry.init() and Sentry.capture() etc
Then, I can receive my events on my sentry server and he can receive his events on his server.

On the new 3.1.0 version, AndroidSentryClientFactory doesn't exist anymore.

Any solution?

Thanks

marandaneto:
you need to migrate and follow the unified API conventions:

https://docs.sentry.io/platforms/android/migration/

you could create your own `Hub` instance, for example, the `Hub` doesn't add the default integrations btw, be aware of that.

arunshankar87:
Thanks to @marandaneto, was able to achieve this. Just putting the steps if it can help others

1.7.27

Creating a client instance

```
SentryClient client = new AndroidSentryClientFactory(context).createSentryClient(new Dsn(dsn));
```

Capturing the event

```
EventBuilder eventBuilder = new EventBuilder()
            .withMessage(message)
            .withSentryInterface(new ExceptionInterface(e));
client.sendEvent(eventBuilder);
```

3.1.1 

Creating a hub instance

```
SentryOptions options = new SentryOptions();
options.setDsn(dsn);
options.setSerializer(new GsonSerializer(options.getLogger(), options.getEnvelopeReader()));
Hub hub = new Hub(options);
```

Capturing the event

```
SentryEvent event;
event = new SentryEvent(e);
Message sentryMessage = new Message();
sentryMessage.setFormatted(message);
event.setMessage(sentryMessage);
hub.captureEvent(event);
```

rrodrigues:
Thanks @arunshankar87 for your steps.
I was being able to do this because I was missing the line
> `options.setSerializer(new GsonSerializer(options.getLogger(), options.getEnvelopeReader()));` 

Also I think that could be useful to have `AndroidOptionsInitializer.init` public.

marandaneto:
@rrodrigues why would you need `AndroidOptionsInitializer.init` public?

Default integrations are supposed to be added only once, by the final App and not a library, otherwise, they would conflict, that's why it's not public.

rrodrigues:
Because if the final app does not use sentry there are integrations that I would like to include anyway.
[...]

PR #1088, opened 2020-12-02
Fix: Options create GsonSerializer by default

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Fix: Options create GsonSerializer by default

when creating a new `Hub` or `SentryClient`, they should be ready to be used and right now the Serializer is only created on the Sentry init static class, so one needs to set it manually.

One would need to set a Serializer to a fresh SentryOptions if they want to have a usable Hub or SentryClient.
Helps with #1080 #993

## :pencil: Checklist

- [X] All tests passing
- [ ] No breaking changes


sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -79,11 +79,12 @@
...
   private @NotNull SentryLevel diagnosticLevel = DEFAULT_DIAGNOSTIC_LEVEL;
 
-  /** Serializer interface to serialize/deserialize json events */
-  private @NotNull ISerializer serializer = NoOpSerializer.getInstance();
-
+  /** Envelope reader interface */
   private @NotNull IEnvelopeReader envelopeReader = new EnvelopeReader();
 
+  /** Serializer interface to serialize/deserialize json events */
+  private @NotNull ISerializer serializer = new GsonSerializer(logger, envelopeReader);
```
maciejwalkowiak, line 86: The only problem I see is that even if users provide their own serializer implementation, they are not able anymore to exclude Gson dependency. Which is maybe fine, i think we only need to make it explicit that Gson is a hard dependency from now on.

maciejwalkowiak (approved):

sentry/src/main/java/io/sentry/SentryOptions.java
marandaneto: that's true, ideally, we'd get rid of Gson altogether, and for that, we'd be refactoring this too, right now its suboptimal even as it is, as have a dependency and it won't work as well.

bruno-garcia (approved):

PR #1094, opened 2020-12-04
Fix: Sort breadcrumbs by Date if the event has breadcrumbs

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Fix: Sort breadcrumbs by Date if the event has breadcrumbs

If an event captured on a top SDK like Flutter and later sent by the Android SDK, the Scope breadcrumbs are merged along with the breadcrumbs from the event, but it's not respecting the timeline.
Ideally, this would be solved by the Scope sync, but if the Scope sync. can be opt-out, the same problem would happen.

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/sentry/src/main/java/io/sentry/SentryClient.java
+++ b/sentry/src/main/java/io/sentry/SentryClient.java
@@ -426,8 +426,6 @@ public final class SentryClient implements ISentryClient {
       sortedBreadcrumbs.addAll(breadcrumbs);
       Collections.sort(sortedBreadcrumbs, sortBreadcrumbsByDate);
     }
-
-    event.setBreadcrumbs(sortedBreadcrumbs);
   }
 
   private @Nullable SentryEvent executeBeforeSend(
--- a/sentry/src/main/java/io/sentry/SentryClient.java
+++ b/sentry/src/main/java/io/sentry/SentryClient.java
@@ -9,6 +9,8 @@ import io.sentry.util.ApplyScopeUtils;
 import io.sentry.util.Objects;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


sentry/src/main/java/io/sentry/SentryClient.java
```diff
@@ -413,6 +418,18 @@ public SentryId captureTransaction(
     return event;
   }
 
+  private void sortBreadcrumbsByDate(
+      final @NotNull SentryEvent event, final @NotNull Collection<Breadcrumb> breadcrumbs) {
+    final List<Breadcrumb> sortedBreadcrumbs = event.getBreadcrumbs();
```
maciejwalkowiak, line 423: Here we modify `event#breadcrumbs`. Shouldn't we do instead:

```
final List<Breadcrumb> sortedBreadcrumbs = new ArrayList<>(event.getBreadcrumbs());?
```

marandaneto: the outcome of this is a sorted breadcrumb in the event payload, so it's not needed to allocate a new ArrayList, I just got the reference to not call a getter a few times

maciejwalkowiak: Ok. Do they need to be set then again in line 430?

bruno-garcia (approved):

sentry/src/main/java/io/sentry/SentryClient.java
marandaneto: yep, good catch, it's a leftover of my previous implementation, gonna remove it, thx

sentry/src/main/java/io/sentry/SentryClient.java
```diff
@@ -413,6 +418,18 @@ public SentryId captureTransaction(
...
+
+    if (!breadcrumbs.isEmpty()) {
+      sortedBreadcrumbs.addAll(breadcrumbs);
+      Collections.sort(sortedBreadcrumbs, sortBreadcrumbsByDate);
+    }
+
+    event.setBreadcrumbs(sortedBreadcrumbs);
```
marandaneto, line 430:
```suggestion
```

maciejwalkowiak (approved):

Issue #1099, opened 2020-12-08
Failed to capture traces if `sentry-trace` header exists


ano1404:
_Platform:_
- [x]  Spring -> 2.2.1.RELEASE

_IDE:_
- [x]  IntelliJ -> IntelliJ IDEA 2020.2.4 (Community Edition)
Build #IC-202.8194.7, built on November 24, 2020

_Build system:_

- [x]  Maven -> 3.6.3

_Android Gradle Plugin:_
- [ ]  Yes -> If yes, which version?
- [x]  No

_Sentry Android Gradle Plugin:_
- [ ]  Yes -> If yes, which version?
- [x]  No

_Proguard/R8:_
- [ ]  Enabled
- [ ]  Disabled

_Platform installed with:_
- [ ] JCenter
- [ ] Bintray
- [x] Maven Central
- [ ] Manually

The version of the SDK:
**4.0.0-alpha.1**

---
I have the following issue:

So if the `sentry-trace` header is exists it will `startTransaction` in this line https://github.com/getsentry/sentry-java/blob/main/sentry-spring/src/main/java/io/sentry/spring/tracing/SentryTracingFilter.java#L88.
The transactionContext only set the parentSampled in https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/TransactionContext.java#L42, but does not set the `sampled` (set to null) https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/TransactionContext.java#L40

This will lead to always skipping the capture process in https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/Hub.java#L652-L658 because the `transaction.isSampled()` returns null

**Expected result:**
- Result it should capture the traces even with `sentry-trace` header exists

maciejwalkowiak:
Thanks @ano1404 for reporting. Fix is on a way.

bruno-garcia:
@ano1404 thanks a lot for trying it out and raising the issue ❤️

PR #1100, opened 2020-12-09
Fix inheriting sampling decision from parent.

maciejwalkowiak:
Fixes gh-1099

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Fix inheriting sampling decision from parent.

When transaction was created from transaction context without providing custom sampling context, `tracingSampler` was not used.

Unit test.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/sentry/src/test/java/io/sentry/HubTest.kt
+++ b/sentry/src/test/java/io/sentry/HubTest.kt
@@ -1006,6 +1006,17 @@ class HubTest {
     }
 
     @Test
+    fun `when startTransaction with parent sampled and no traces sampler provided, transaction inherits sampling decision`() {
+        val hub = generateHub()
+        val transactionContext = TransactionContext("name")
+        transactionContext.parentSampled = true
+        val transaction = hub.startTransaction(transactionContext)
+        assertNotNull(transaction)
+        assertNotNull(transaction.isSampled)
+        assertTrue(transaction.isSampled!!)
+    }
+
+    @Test
     fun `Hub should close the sentry executor processor on close call`() {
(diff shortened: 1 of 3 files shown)


sentry/src/test/java/io/sentry/HubTest.kt
```diff
@@ -1005,6 +1005,17 @@ class HubTest {
...
 
+    @Test
+    fun `when startTransaction with parent sampled and no traces sampler provided, transaction inherits sampling decision`() {
+        val hub = generateHub()
+        val transactionContext = TransactionContext("name")
+        transactionContext.parentSampled = true
+        val transaction = hub.startTransaction(transactionContext)
+        assertNotNull(transaction)
+        assertNotNull(transaction.isSampled)
+        assertTrue(transaction.isSampled!!)
```
marandaneto, line 1016: would not be more readable `assertFalse` ?

marandaneto, line 1016: do we need `!!` here? Kotlin should do it automatically for you as you did `assertNotNull(transaction.isSampled)` before

marandaneto (approved):

sentry/src/test/java/io/sentry/HubTest.kt
maciejwalkowiak: we wouldn't if its immutable. But unfortunately it is mutable.

bruno-garcia (approved):

PR #1104, opened 2020-12-10
WIP: Added non-blocking connection implementation.

maciejwalkowiak:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

Added non-blocking connection implementation based on [async-http-client](https://github.com/AsyncHttpClient/async-http-client).

**This is a draft** - please don't focus on details :-)

With current `AsyncConnection` and `HttpTransport`, there is a single thread sending envelopes. Practically it means that there can be up to 3 envelopes send per second. While this can be tweaked to reuse connections instead of closing them improving at the same time throughput to ~8 envelopes per second, it is still not enough for the performance feature.

One option is to increase the number of threads in `AsyncConnection` thread-pool, the other much more resource efficient is to use NIO based HTTP client.

This PR is an experiment on how this implementation could look like and how far we can get with the results.

Current state: with `async-http-client` I was able to send ~350 envelopes per second - not a single envelope was dropped - and was hitting rate limits from the Sentry server with no issues.

Tests are in progress.

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps

As @marandaneto mentioned, there are more popular HTTP clients than `async-http-client` so the next step is to look into them and see if there are any pros/cons.

### DIFF
--- a/sentry-samples/sentry-samples-console/src/main/java/io/sentry/samples/console/Main.java
+++ b/sentry-samples/sentry-samples-console/src/main/java/io/sentry/samples/console/Main.java
@@ -158,7 +158,7 @@ public class Main {
 
     // All events that have not been sent yet are being flushed on JVM exit. Events can be also
     // flushed manually:
-    // Sentry.close();
+    Sentry.close();
   }
 
   private static class SomeEventProcessor implements EventProcessor {
--- a/sentry/build.gradle.kts
+++ b/sentry/build.gradle.kts
@@ -19,6 +19,7 @@ tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach 
 dependencies {
     // Envelopes require JSON. Until a parse is done without GSON, we'll depend on it explicitly here
     implementation(Config.Libs.gson)
+    compileOnly("org.asynchttpclient:async-http-client:2.12.1")
 
     compileOnly(Config.CompileOnly.nopen)
     errorprone(Config.CompileOnly.nopenChecker)
--- a/sentry/src/main/java/io/sentry/ConnectionFactory.java
+++ b/sentry/src/main/java/io/sentry/ConnectionFactory.java
@@ -1,0 +1,26 @@
+package io.sentry;
+
+import io.sentry.transport.Connection;
+import org.jetbrains.annotations.NotNull;
+
+public final class ConnectionFactory {
+  private ConnectionFactory() {}
+
+  public static @NotNull Connection create(final @NotNull SentryOptions options) {
+    if (classExists("org.asynchttpclient.AsyncHttpClient")) {
+      return NIOConnectionFactory.create(options);
+    } else {
+      return AsyncConnectionFactory.create(options, options.getEnvelopeDiskCache());
+    }
+  }
+
+  private static boolean classExists(String clazz) {
+    try {
+      Class.forName(clazz, false, ClassLoader.getSystemClassLoader());
+      return true;
+    } catch (ClassNotFoundException e) {
+      e.printStackTrace();
+      return false;
+    }
+  }
+}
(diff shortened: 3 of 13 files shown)


sentry-samples/sentry-samples-console/src/main/java/io/sentry/samples/console/Main.java
```diff
@@ -158,7 +158,7 @@ public static void main(String[] args) throws InterruptedException {
 
     // All events that have not been sent yet are being flushed on JVM exit. Events can be also
     // flushed manually:
-    // Sentry.close();
+    Sentry.close();
```
maciejwalkowiak, line 161: Drawback of `async-http-client` - http client has to be explicitly closed, otherwise it will prevent JVM from shutting down.

sentry/src/main/java/io/sentry/ConnectionFactory.java
```diff
@@ -0,0 +1,26 @@
+package io.sentry;
+
+import io.sentry.transport.Connection;
+import org.jetbrains.annotations.NotNull;
+
+public final class ConnectionFactory {
+  private ConnectionFactory() {}
+
+  public static @NotNull Connection create(final @NotNull SentryOptions options) {
```
maciejwalkowiak, line 9: This is one way to determine which client should be used, the other would be to make it more explicit for users.

sentry/src/main/java/io/sentry/transport/AsyncConnection.java
```diff
@@ -14,8 +13,8 @@
 import io.sentry.util.Objects;
 import java.io.Closeable;
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.List;
+import java.time.Duration;
+import java.time.Instant;
```
marandaneto, line 17: this is not Android compatible and cannot live in this package

sentry/src/main/java/io/sentry/ConnectionFactory.java
```diff
@@ -0,0 +1,26 @@
...
+
+  private static boolean classExists(String clazz) {
+    try {
+      Class.forName(clazz, false, ClassLoader.getSystemClassLoader());
+      return true;
+    } catch (ClassNotFoundException e) {
+      e.printStackTrace();
+      return false;
+    }
+  }
```
marandaneto, line 25: reflection on Android is very bad and we already do some + IO, adding one more does not sound a good idea, this slows down the Sdk init which might drop frames on Mobile, wondering if it makes sense to split the whole "transport" dependent package to a different module

sentry/src/main/java/io/sentry/transport/AsyncConnection.java
maciejwalkowiak: Yes, it's a leftover. Anyway, don't spend too much time on details till it's a draft. The most important part is using the `Connection` as the construct that will be replacable with help of objects like `RateLimiter` and `HttpTransport` would become an implementation detail.

sentry/src/main/java/io/sentry/transport/AsyncConnection.java
```diff
@@ -33,19 +32,22 @@
...
+  private final @NotNull RateLimiter rateLimiter;
 
   public AsyncConnection(
       final ITransport transport,
       final ITransportGate transportGate,
       final IEnvelopeCache envelopeCache,
       final int maxQueueSize,
-      final SentryOptions options) {
+      final SentryOptions options,
+      final @NotNull RateLimiter rateLimiter) {
```
marandaneto, line 41: I loved that!

sentry/build.gradle.kts
```diff
@@ -19,6 +19,7 @@ tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach
 dependencies {
     // Envelopes require JSON. Until a parse is done without GSON, we'll depend on it explicitly here
     implementation(Config.Libs.gson)
+    compileOnly("org.asynchttpclient:async-http-client:2.12.1")
```
marandaneto, line 22: I believe the android build will complain because of `InvalidPackage` lints

maciejwalkowiak:
Superseded by https://github.com/getsentry/sentry-java/pull/1114 and https://github.com/getsentry/sentry-java/pull/1136

PR #1111, opened 2020-12-14
Ref: using Calendar to generate Dates

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [X] Refactoring

## :scroll: Description
Ref: using Calendar instead of Dates directly

this would avoid the Date/String parsing and improve the performance #1081

## :pencil: Checklist

- [X] All tests passing
- [ ] No breaking changes

### DIFF
--- a/sentry/src/main/java/io/sentry/DateUtils.java
+++ b/sentry/src/main/java/io/sentry/DateUtils.java
@@ -3,96 +3,81 @@ package io.sentry;
 import java.text.DateFormat;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
+import java.util.Calendar;
 import java.util.Date;
 import java.util.Locale;
 import java.util.TimeZone;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
-import org.jetbrains.annotations.Nullable;
 
 /** Utilities to deal with dates */
 @ApiStatus.Internal
 public final class DateUtils {
   private static final String UTC = "UTC";
+  // ISO 8601
   private static final String ISO_FORMAT = "yyyy-MM-dd'T'HH:mm:ss'Z'";
   private static final String ISO_FORMAT_WITH_MILLIS = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'";
-  private static final ThreadLocal<SimpleDateFormat> SDF_ISO_FORMAT_WITH_MILLIS_UTC =
+
+  private static @NotNull TimeZone getTimeZone() {
+    TimeZone timeZone = TimeZone.getTimeZone(UTC);
+    // if UTC is not available, let's get the current one and avoid NPE problems
+    if (timeZone == null) {
+      timeZone = TimeZone.getDefault();
+    }
+    return timeZone;
+  }
+
+  private static final @NotNull TimeZone UTC_TIMEZONE = getTimeZone();
+
+  private static final @NotNull ThreadLocal<SimpleDateFormat> SDF_ISO_FORMAT_WITH_MILLIS_UTC =
       new ThreadLocal<SimpleDateFormat>() {
         @Override
         protected SimpleDateFormat initialValue() {
-          final TimeZone tz = TimeZone.getTimeZone(UTC);
           final SimpleDateFormat simpleDateFormat =
               new SimpleDateFormat(ISO_FORMAT_WITH_MILLIS, Locale.ROOT);
-          simpleDateFormat.setTimeZone(tz);
+          simpleDateFormat.setTimeZone(UTC_TIMEZONE);
           return simpleDateFormat;
         }
       };
-  private static final ThreadLocal<SimpleDateFormat> SDF_ISO_FORMAT_WITH_MILLIS =
+
+  private static final @NotNull ThreadLocal<SimpleDateFormat> SDF_ISO_FORMAT_UTC =
       new ThreadLocal<SimpleDateFormat>() {
         @Override
         protected SimpleDateFormat initialValue() {
-          return new SimpleDateFormat(ISO_FORMAT_WITH_MILLIS, Locale.ROOT);
-        }
-      };
-  private static final ThreadLocal<SimpleDateFormat> SDF_ISO_FORMAT =
-      new ThreadLocal<SimpleDateFormat>() {
-        @Override
-        protected SimpleDateFormat initialValue() {
-          return new SimpleDateFormat(ISO_FORMAT, Locale.ROOT);
+          final SimpleDateFormat simpleDateFormat =
+                  new SimpleDateFormat(ISO_FORMAT, Locale.ROOT);
+          simpleDateFormat.setTimeZone(UTC_TIMEZONE);
+          return simpleDateFormat;
         }
       };
 
   private DateUtils() {}
 
   /**
-   * Get date formatted as expected by Sentry.
+   * Get the current date and time (UTC)
    *
-   * @param date the current date with local timezone
-   * @return the ISO formatted UTC date with millis precision.
-   */
-  public static @NotNull String getTimestampIsoFormat(final @NotNull Date date) {
-    return SDF_ISO_FORMAT_WITH_MILLIS_UTC.get().format(date);
-  }
-
-  /**
-   * Get the current date and time as ISO UTC
-   *
-   * @return the ISO UTC date and time
+   * @return the UTC date and time
    */
   @SuppressWarnings("JdkObsolete")
-  public static @NotNull Date getCurrentDateTime() throws IllegalArgumentException {
-    final String timestampIsoFormat = getTimestampIsoFormat(new Date());
-    return getDateTime(timestampIsoFormat);
+  public static @NotNull Date getCurrentDateTime() {
+    final Calendar calendar = Calendar.getInstance(UTC_TIMEZONE);
+    return calendar.getTime();
   }
 
   /**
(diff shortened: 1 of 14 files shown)


maciejwalkowiak (approved):
Great! Did you measure maybe how much more efficient it is?

marandaneto:
> Great! Did you measure maybe how much more efficient it is?

yes but on my device, even the old implementation is returning ~0 and ~1ms so it's hard to say without proper benchmarking.
I'd say it saves the current Date <-> String transformation so it should be already way better.

maciejwalkowiak:
I used JMH to test SDF vs thread local SDF https://github.com/maciej-scratches/simple-date-format-benchmark

bruno-garcia:
Ideally we'd measure it before making the change. Getting the habit of microbenchmarking when it matters is a good thing and the only way to confirm the changes we're making are worth it

philipphofmann:
> Ideally we'd measure it before making the change. Getting the habit of microbenchmarking when it matters is a good thing and the only way to confirm the changes we're making are worth it

If you want to even go further, adding a test that fails if the thing you are benchmarking is getting slower to avoid a regression.

marandaneto:
> > Ideally we'd measure it before making the change. Getting the habit of microbenchmarking when it matters is a good thing and the only way to confirm the changes we're making are worth it
> 
> If you want to even go further, adding a test that fails if the thing you are benchmarking is getting slower to avoid a regression.

I'm afraid this is not easily possible if we don't test on real devices (low-mid-high end).
this was slow in low-end devices (Samsung s7 initially), as I stated before, I did benchmarking on my device (which is high-end), but even the old implementation was returning either 0 or 1 ms, and Java is not good at microseconds.

so I'd correct @bruno-garcia phrase.
ideally, `we'd keep doing` benchmarking :)
the question here is if we wish to use proper benchmarking tooling + regression tests + running on real devices, I'd say it's not a 1h job, but I'm in to do it.
Executing benchmarking on the JVM using a powerful machine is useless, at least when it's a matter of micro-optimizations., the Thread local SFD already did a pretty good job.

sentry/src/main/java/io/sentry/DateUtils.java
```diff
@@ -114,32 +90,32 @@ private DateUtils() {}
...
-   * @param date the Date with local timezone
-   * @return the Date UTC timezone
+   * @param millis the UTC millis from the epoch
+   * @return the UTC Date
    */
-  public static @NotNull Date getDateTime(final @NotNull Date date)
-      throws IllegalArgumentException {
-    final String timestampIsoFormat = getTimestampIsoFormat(date);
-    return getDateTime(timestampIsoFormat);
+  public static @NotNull Date getDateTime(final long millis) {
```
philipphofmann, line 116: `m`: Why not just use `new Date(millis)`? No need to create an instance of `Calendar` I think. We could remove this method I think or am I missing something?

sentry/src/main/java/io/sentry/DateUtils.java
```diff
@@ -3,96 +3,72 @@
...
    *
-   * @return the ISO UTC date and time
+   * @return the UTC Date
    */
   @SuppressWarnings("JdkObsolete")
-  public static @NotNull Date getCurrentDateTime() throws IllegalArgumentException {
-    final String timestampIsoFormat = getTimestampIsoFormat(new Date());
-    return getDateTime(timestampIsoFormat);
+  public static @NotNull Date getCurrentDateTime() {
+    final Calendar calendar = Calendar.getInstance(UTC_TIMEZONE);
```
philipphofmann, line 54: `m`: Does it maybe make sense to move this to `private static final Calendar calendar = Calendar.getInstance(UTC_TIMEZONE);` so we save creating an instance on each call?

marandaneto:
> I used JMH to test SDF vs thread local SDF https://github.com/maciej-scratches/simple-date-format-benchmark

not sure if there's much difference, but you target Java 11 btw

sentry/src/main/java/io/sentry/DateUtils.java
```diff
@@ -3,96 +3,72 @@
...
   private static final String UTC = "UTC";
+  // ISO 8601
   private static final String ISO_FORMAT = "yyyy-MM-dd'T'HH:mm:ss'Z'";
   private static final String ISO_FORMAT_WITH_MILLIS = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'";
-  private static final ThreadLocal<SimpleDateFormat> SDF_ISO_FORMAT_WITH_MILLIS_UTC =
+
+  // if UTC is not found, it fallback to "GMT" which is UTC equivalent
+  private static final @NotNull TimeZone UTC_TIMEZONE = TimeZone.getTimeZone(UTC);
+
+  private static final @NotNull ThreadLocal<SimpleDateFormat> SDF_ISO_FORMAT_WITH_MILLIS_UTC =
```
philipphofmann, line 24: `l`: Just out of curiosity, why do we use `ThreadLocal` here?

sentry/src/main/java/io/sentry/DateUtils.java
marandaneto: no, Calendar is equivalent to `new Date()`, we need to create a new one to get the current date/time

sentry/src/main/java/io/sentry/DateUtils.java
marandaneto: `Date` is not TimeZone aware, we need a UTC Date

sentry/src/main/java/io/sentry/DateUtils.java
marandaneto: see the BM https://github.com/maciej-scratches/simple-date-format-benchmark
tldr, SimpleDateFormat is expensive and not thread-safe, so using a ThreadLocal speeds up A LOT

sentry/src/main/java/io/sentry/DateUtils.java
marandaneto: ps* Calendar actually creates a new Date(), but the UTC ctor is deprecated, and ask to use a Calendar
https://docs.oracle.com/javase/7/docs/api/java/util/Date.html#UTC(int,%20int,%20int,%20int,%20int,%20int)

sentry/src/main/java/io/sentry/DateUtils.java
marandaneto: the thing is, even that calendat.getTime() returns a new Date, makes sense that we'd cache it, but when you look at the Calendar class, it has some Global fields like `isTimeSet` etc being used to generate new Times, so I'm afraid that this will give us wrong values, as I don't know much about the Calendar class, I've used to work with Jodatime or other tools.
Do you know if its safe to Cache the Calendar and use only it?

marandaneto:
See calendarDateTime

Benchmark                             Mode  Cnt        Score       Error  Units
DateUtilsBenchmark.calendarDateTime  thrpt   20  7839604.986 ± 47744.236  ops/s
DateUtilsBenchmark.sdfConstructor    thrpt   20   346013.913 ±  6255.217  ops/s
DateUtilsBenchmark.sdfThreadLocal    thrpt   20   710661.007 ± 17248.333  ops/s

calendarDateTime but caching the `Calendar.getInstance(UTC_TIMEZONE)` to a static var.

Benchmark                             Mode  Cnt        Score       Error  Units
DateUtilsBenchmark.calendarDateTime  thrpt   20  7884103.710 ± 63048.777  ops/s
DateUtilsBenchmark.sdfConstructor    thrpt   20   356422.835 ±  7655.604  ops/s
DateUtilsBenchmark.sdfThreadLocal    thrpt   20   675566.794 ±  6334.113  ops/s

sentry/src/main/java/io/sentry/DateUtils.java
marandaneto: `Calendar` is not thread-safe so adding to a static field won't work.

philipphofmann (approved):
Benchmarking would be nice, but I have no real experience with doing that properly in Java. It's up to you @marandaneto if you want to set it up or not. Thanks for answering all my questions. DateTime is always complicated and you can easily get it wrong, so I'm always super careful when touching it.

bruno-garcia:
> Executing benchmarking on the JVM using a powerful machine is useless, at least when it's a matter of micro-optimizations., the Thread local SFD already did a pretty good job.

I wouldn't say it's useless.

Benchmarking is not just measuring one thing alone. It's measuring it against a baseline. If you're doing that on the JVM it's very likely the difference between benchmark and the change, if relevant, would show up on whatever other runtime or environment you put it.

I didn't even think of benchmarking this on devices to be honest. Could be nice to do but it would be a last test to make sure we're not hitting something odd, device/android runtime specific.

marandaneto:
> > Executing benchmarking on the JVM using a powerful machine is useless, at least when it's a matter of micro-optimizations., the Thread local SFD already did a pretty good job.
> 
> I wouldn't say it's useless.
> 
> Benchmarking is not just measuring one thing alone. It's measuring it against a baseline. If you're doing that on the JVM it's very likely the difference between benchmark and the change, if relevant, would show up on whatever other runtime or environment you put it.
> 
> I didn't even think of benchmarking this on devices to be honest. Could be nice to do but it would be a last test to make sure we're not hitting something odd, device/android runtime specific.

yeah, it's not useless of course, just a way to say, but what I mean is, the performance issues were most identified by a low-end device and we were not running benchmarking on a similar env, the slowness was not noticeable on the JVM using a powerful machine, after the thread-local refactoring.

PR #1118, opened 2020-12-16
Make ITransport top-level abstraction for sending events to Sentry.

maciejwalkowiak:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description

Flip the relation between `AsyncConnection` and `ITransport` and make `BlockingHttpTransport` use `HttpConnection`, making it possible to provide alternative transport implementations not bound to use thread pool.

In order to provide non blocking transport, transport itself has to be responsible for managing threads (not the AsyncConnection).

Unit & integration tests.

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes


maciejwalkowiak:
@marandaneto, as a next step - I believe we should create an interface `ITransportFactory` and put it on `SentryOptions`, instead of `ITransfer` itself. Currently, since transport is set on SentryOptions, but transport itself also likely depends on SentryOptions, it may be difficult for users to provide custom transport implementations.

sentry/src/main/java/io/sentry/transport/BlockingHttpTransport.java
```diff
@@ -11,57 +11,79 @@
...
+    }
+
+    final SentryEnvelope filteredEnvelope = rateLimiter.filter(envelope, hint);
+
+    if (filteredEnvelope == null) {
+      if (cached) {
+        envelopeCache.discard(envelope);
+      }
+    } else {
+      executor.submit(new EnvelopeSender(filteredEnvelope, hint, currentEnvelopeCache));
```
bruno-garcia, line 85: I think the name `BlockingHttpTransport` is misleading since it gives the impression the interface is going to block until for the event submission, which isnt' the case.

We're scheduling the event through the executor here.

Maybe we can use another name?

sentry/src/main/java/io/sentry/transport/StdoutTransport.java
```diff
@@ -15,14 +15,13 @@ public StdoutTransport(final @NotNull ISerializer serializer) {
...
-  public TransportResult send(final @NotNull SentryEnvelope envelope) throws IOException {
+  public void send(final @NotNull SentryEnvelope envelope, Object hint) throws IOException {
     Objects.requireNonNull(envelope, "SentryEnvelope is required");
 
     try {
       serializer.serialize(envelope, System.out);
-      return TransportResult.success();
     } catch (Exception e) {
-      return TransportResult.error();
+      // do nothing
```
marandaneto, line 24: not really against but why the change?

sentry/src/main/java/io/sentry/transport/BlockingHttpTransport.java
marandaneto: yeah I kinda agree, not sure about a suitable name though.

SingleThreadHttpTransport?

sentry/src/main/java/io/sentry/transport/HttpConnection.java
```diff
@@ -27,37 +24,21 @@
...
 
-/**
- * An implementation of the {@link ITransport} interface that sends the events to the Sentry server
- * over HTTP(S).
- */
 @Open
-@ApiStatus.NonExtendable // only not final because of testing
-@ApiStatus.Internal
-public class HttpTransport implements ITransport {
+class HttpConnection {
```
marandaneto, line 32: is the `HttpConnection` reusable for the NIO implementation as well? just trying to understand why we have the BlockingHttpTransport + HttpConnection, I do get their difference, but what would be the advantage of it right now?

marandaneto:
do we need a changelog? how do we communicate the breaking changes later on?

sentry/src/main/java/io/sentry/transport/BlockingHttpTransport.java
maciejwalkowiak: Can be also `AsyncHttpTransport`. It is async, but blocking.

sentry/src/main/java/io/sentry/transport/StdoutTransport.java
maciejwalkowiak: The interface has changed and it does not return anymore.

sentry/src/main/java/io/sentry/transport/HttpConnection.java
maciejwalkowiak: Just to split the responsibilities - and I agree that the split is not perfect. I tried to create decorators like `EnvelopeCachingHttpTransport` and `RateLimitingHttpTransport` but since `send` returns `void` it's also not possible. 

`HttpConnection` is package private, to give us flexibility of refactoring later without breaking users' code.

maciejwalkowiak:
@marandaneto @bruno-garcia I need this PR reviewed.

sentry/src/main/java/io/sentry/AsyncHttpTransportFactory.java
```diff
@@ -0,0 +1,36 @@
...
+import io.sentry.util.Objects;
+import java.net.MalformedURLException;
+import java.net.URL;
+import org.jetbrains.annotations.NotNull;
+
+/** Creates {@link AsyncHttpTransport}. */
+final class AsyncHttpTransportFactory implements TransportFactory {
+
+  @Override
+  public ITransport create(final @NotNull SentryOptions options) {
```
marandaneto, line 16: should we add `throws IllegalArgumentException` in the method signature?

marandaneto, line 16:
```suggestion
  public @NotNull ITransport create(final @NotNull SentryOptions options) {
```

sentry/src/main/java/io/sentry/transport/AsyncHttpTransport.java
```diff
@@ -139,7 +122,6 @@ public void close() throws IOException {
                 "Failed to shutdown the async connection async sender within 1 minute. Trying to force it now.");
         executor.shutdownNow();
       }
-      transport.close();
```
marandaneto, line 142: how does the connection get closed now?

sentry/src/main/java/io/sentry/transport/HttpConnection.java
marandaneto: sounds good to me

sentry/src/main/java/io/sentry/transport/HttpConnection.java
```diff
@@ -27,37 +24,21 @@
...
-  private final @Nullable SSLSocketFactory sslSocketFactory;
-  private final @Nullable HostnameVerifier hostnameVerifier;
-
-  private final @NotNull SentryOptions options;
-
-  private final @NotNull ILogger logger;
-
-  private final @NotNull RateLimiter rateLimiter;
+  private final SentryOptions options;
+  private final RateLimiter rateLimiter;
```
marandaneto, line 41:
```suggestion
  private final @NotNull SentryOptions options;
  private final @NotNull RateLimiter rateLimiter;
```

sentry/src/main/java/io/sentry/transport/HttpConnection.java
```diff
@@ -67,51 +48,31 @@
...
+  public HttpConnection(
       final @NotNull SentryOptions options,
       final @NotNull IConnectionConfigurator connectionConfigurator,
-      final int connectionTimeoutMillis,
-      final int readTimeoutMillis,
-      final @Nullable SSLSocketFactory sslSocketFactory,
-      final @Nullable HostnameVerifier hostnameVerifier,
-      final @NotNull URL sentryUrl) {
+      final @NotNull URL sentryUrl,
+      RateLimiter rateLimiter) {
```
marandaneto, line 58:
```suggestion
      final @NotNull RateLimiter rateLimiter) {
```

sentry/src/main/java/io/sentry/transport/HttpConnection.java
```diff
@@ -67,51 +48,31 @@
...
       final @NotNull SentryOptions options,
       final @NotNull IConnectionConfigurator connectionConfigurator,
-      final int connectionTimeoutMillis,
-      final int readTimeoutMillis,
-      final @Nullable SSLSocketFactory sslSocketFactory,
-      final @Nullable HostnameVerifier hostnameVerifier,
       final @NotNull URL sentryUrl,
       final @NotNull AuthenticatorWrapper authenticatorWrapper,
-      final @NotNull RateLimiter rateLimiter) {
+      RateLimiter rateLimiter) {
```
marandaneto, line 72:
```suggestion
      final @NotNull RateLimiter rateLimiter) {
```

sentry/src/main/java/io/sentry/transport/AsyncHttpTransport.java
marandaneto: well the method was not implemented anyway I guess

marandaneto (approved):
few missing jetbrains annotations but other than that LGTM.

I didn't find a plugin that fails the build if annotations are missing, but it'd be nice to write a custom linter for that, so I don't need to be the human linter here :D

I am only picky about that cus it's nicer when writing Kotlin code.

sentry/src/main/java/io/sentry/AsyncHttpTransportFactory.java
maciejwalkowiak: I think we should not, what's the reasonable way to handle this exception here anyways?

marandaneto: if we handle it down its fine, I was just curious cus the `TransportFactory` now is the way to go for custom transports right

sentry/src/main/java/io/sentry/TransportFactory.java
```diff
@@ -0,0 +1,9 @@
+package io.sentry;
+
+import io.sentry.transport.ITransport;
+import org.jetbrains.annotations.NotNull;
+
+/** Creates instances of {@link ITransport}. */
+public interface TransportFactory {
```
marandaneto, line 7: should we do `ITransportFactory`?

maciejwalkowiak: I have mixed feelings about putting `I` in the front of names for interfaces, but I will just follow your preference. Tell me :-)

marandaneto: no feelings or preferences here, but see `ITransport`, so ideally, we do a single way, either all of them or none of them.
chose which ones to rename :)

maciejwalkowiak: Let it be `ITransportFactory`.

sentry/src/main/java/io/sentry/AsyncHttpTransportFactory.java
maciejwalkowiak: We don't handle it, the exception is propagated to the very top and in the end Sentry init will fail.

This PR is a PR to another branch, so once whole package is there we can take a look at it again.

PR #1126, opened 2020-12-18
Do not return null from startTransaction.

maciejwalkowiak:
## :scroll: Description

Changes `startTransaction` method to never return `null`.

`null` could have been returned only if hub is closed or if stack contains zero items.

- when hub is closed, transaction will not be captured anyway since `captureTransaction` does the check for closed hub anyways
- it's not technically possible that `stack` contains zero items

Tests already cover it.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

Create abstraction over stack to make sure that stack can never be empty.

### DIFF
--- a/sentry/src/main/java/io/sentry/Hub.java
+++ b/sentry/src/main/java/io/sentry/Hub.java
@@ -639,7 +639,7 @@ public final class Hub implements IHub {
   @ApiStatus.Internal
   @Override
   public @NotNull SentryId captureTransaction(
-      final @NotNull SentryTransaction transaction, final @Nullable Object hint) {
+      final @NotNull ITransaction transaction, final @Nullable Object hint) {
     Objects.requireNonNull(transaction, "transaction is required");
 
     SentryId sentryId = SentryId.EMPTY_ID;
@@ -683,13 +683,13 @@ public final class Hub implements IHub {
   }
 
   @Override
-  public @NotNull SentryTransaction startTransaction(
+  public @NotNull ITransaction startTransaction(
       final @NotNull TransactionContext transactionContexts) {
     return this.startTransaction(transactionContexts, null);
   }
 
   @Override
-  public @NotNull SentryTransaction startTransaction(
+  public @NotNull ITransaction startTransaction(
       final @NotNull TransactionContext transactionContexts,
       final @Nullable CustomSamplingContext customSamplingContext) {
     Objects.requireNonNull(transactionContexts, "transactionContexts is required");
@@ -699,12 +699,22 @@ public final class Hub implements IHub {
     boolean samplingDecision = tracingSampler.sample(samplingContext);
     transactionContexts.setSampled(samplingDecision);
 
-    SentryTransaction transaction = new SentryTransaction(transactionContexts, this);
-    final StackItem item = stack.peek();
-    if (item != null) {
-      item.scope.setTransaction(transaction);
+    ITransaction transaction = null;
+    if (!isEnabled()) {
+      options
+          .getLogger()
+          .log(
+              SentryLevel.WARNING,
+              "Instance is disabled and this 'startTransaction' returns a no-op.");
+      transaction = new NoOpTransaction();
     } else {
-      options.getLogger().log(SentryLevel.FATAL, "Stack peek was null when startTransaction");
+      final StackItem item = stack.peek();
+      if (item != null) {
+        transaction = new SentryTransaction(transactionContexts, this);
+        item.scope.setTransaction(transaction);
+      } else {
+        options.getLogger().log(SentryLevel.FATAL, "Stack peek was null when startTransaction");
+      }
     }
     return transaction;
   }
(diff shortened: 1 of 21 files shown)


sentry/src/main/java/io/sentry/Hub.java
```diff
@@ -699,21 +699,12 @@ public void flush(long timeoutMillis) {
...
+      item.scope.setTransaction(transaction);
     } else {
-      final StackItem item = stack.peek();
-      if (item != null) {
-        transaction = new SentryTransaction(transactionContexts, this);
-        item.scope.setTransaction(transaction);
-      } else {
-        options.getLogger().log(SentryLevel.FATAL, "Stack peek was null when startTransaction");
-      }
+      options.getLogger().log(SentryLevel.FATAL, "Stack peek was null when startTransaction");
```
marandaneto, line 707: not sure if it makes sense, SDK should be enabled in order to do its operations, and right now we don't check it anymore, what's the purpose of it?

maciejwalkowiak: SDK can be disabled only after it is closed. When SDK is closed, check if it is closed will happen on transaction send in `captureTransaction` - so we still have these checks present.

Without this change, since `startTransaction` can return `null`, every time users call `startTransaction` they have to check if it's null and the code gets very verbose. While in Java you can simply ignore the fact that it can return null, imagine translating this code to Kotlin:

```java
SentryTransaction transaction = Sentry.startTransaction("processOrderBatch()");
transaction.setOperation("task");
try {
    procesOrderBatch();
} catch (Exception e) {
    transaction.setThrowable(e);
    transaction.setStatus(SpanStatus.INTERNAL_ERROR)
} finally {
    transaction.finish();
}
```

sentry/src/main/java/io/sentry/Hub.java
```diff
@@ -699,21 +699,12 @@ public void flush(long timeoutMillis) {
...
     transactionContexts.setSampled(samplingDecision);
 
-    SentryTransaction transaction = null;
-    if (!isEnabled()) {
-      options
-          .getLogger()
-          .log(
-              SentryLevel.WARNING,
-              "Instance is disabled and this 'startTransaction' call is a no-op.");
+    SentryTransaction transaction = new SentryTransaction(transactionContexts, this);
```
bruno-garcia, line 702: Can we return `NoOpTransaction` if SDK is disabled instead?

If the goal is to have no overhead, we can avoid the allocations.

sentry/src/main/java/io/sentry/Hub.java
marandaneto: it should return `NotNull`, that's for sure, and ideally, `NoOpTransaction` as just @bruno-garcia said.

true about being checked already in `captureTransaction`, agreed with the removal, I'd pass already a `NoOpTransaction` though.

maciejwalkowiak:
As requested - extracted `ITransaction` interface and created `NoOpTransaction` class.

maciejwalkowiak:
In this case I think we can ignore @codecov-io. I can artificially bump the coverage but I don't see the point.

sentry/src/main/java/io/sentry/Hub.java
```diff
@@ -699,12 +699,22 @@ public void flush(long timeoutMillis) {
...
+              "Instance is disabled and this 'startTransaction' returns a no-op.");
+      transaction = new NoOpTransaction();
     } else {
-      options.getLogger().log(SentryLevel.FATAL, "Stack peek was null when startTransaction");
+      final StackItem item = stack.peek();
+      if (item != null) {
+        transaction = new SentryTransaction(transactionContexts, this);
+        item.scope.setTransaction(transaction);
+      } else {
+        options.getLogger().log(SentryLevel.FATAL, "Stack peek was null when startTransaction");
```
marandaneto: should we do `transaction = new NoOpTransaction();` here too? otherwise it could return null

sentry/src/main/java/io/sentry/HubAdapter.java
```diff
@@ -154,23 +154,23 @@ public IHub clone() {
...
 
   @Override
-  public SentryId captureTransaction(SentryTransaction transaction, Object hint) {
+  public SentryId captureTransaction(ITransaction transaction, Object hint) {
     return Sentry.getCurrentHub().captureTransaction(transaction, hint);
   }
 
   @Override
-  public SentryTransaction startTransaction(TransactionContext transactionContexts) {
+  public ITransaction startTransaction(TransactionContext transactionContexts) {
```
marandaneto, line 162: the suggestion button is just gone oO
I'd add `@NotNull` to the return type

marandaneto (approved):
LGTM

sentry/src/main/java/io/sentry/Hub.java
maciejwalkowiak: I left it intentionally untouched as merging https://github.com/getsentry/sentry-java/pull/1127 removes this use case.

sentry/src/main/java/io/sentry/HubAdapter.java
maciejwalkowiak: 👍

Issue #1131, opened 2020-12-21
User.setOthers is marked @Nullable but crashes when passing in null


angusholder:
_Platform:_
- Android target SDK 29, compile SDK 29
- Java sourceCompatibility 1.8/targetCompatibility 1.8
- Kotlin 1.4.10

_IDE:_
- Android Studio 4.1

_Build system:_
- Gradle 6.5

_Android Gradle Plugin:_
- 4.1

The version of the SDK:
**io.sentry:sentry-android:3.2.0**

---
I have the following issue:

[`User.setOthers`](https://github.com/getsentry/sentry-java/blob/082b4b56575ae7c20bf423f085c38605b194e680/sentry/src/main/java/io/sentry/protocol/User.java#L128) is documented as accepting null, but crashes when passing null, as it passes this straight into `new ConcurrentHashMap(Map)`, which does not accept null.

marandaneto:
@angusholder thanks for raising this, we'll work on it.

marandaneto:
it'll be available in the next release, thanks for reporting.

PR #1133, opened 2020-12-22
Fix: Do not crash when passing null values to @Nullable methods, eg User and Scope

marandaneto:
## :scroll: Description
Fix: User.setOthers is marked @Nullable but crashes when passing in null

#1131

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * Feat: set isSideLoaded info tags
 * Enhancement: Read tracesSampleRate from AndroidManifest
 * Ref: Return NoOpTransaction instead of null (#1126)
+* Fix: User.setOthers is marked @Nullable but crashes when passing in null #1131
 
 # 4.0.0-alpha.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 * Feat: set isSideLoaded info tags
 * Enhancement: Read tracesSampleRate from AndroidManifest
 * Ref: Return NoOpTransaction instead of null (#1126)
-* Fix: User.setOthers is marked @Nullable but crashes when passing in null #1131
+* Fix: Do not crash when passing null values to @Nullable methods, eg User and Scope
 
 # 4.0.0-alpha.2
 
--- a/sentry/src/main/java/io/sentry/Scope.java
+++ b/sentry/src/main/java/io/sentry/Scope.java
@@ -83,7 +83,7 @@ public final class Scope implements Cloneable {
    *
    * @param level the SentryLevel
    */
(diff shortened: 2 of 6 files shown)


sentry/src/main/java/io/sentry/Scope.java
```diff
@@ -175,7 +175,10 @@ public void setUser(@Nullable User user) {
    *
    * @param fingerprint the fingerprint list
    */
-  public void setFingerprint(@NotNull List<String> fingerprint) {
+  public void setFingerprint(final @NotNull List<String> fingerprint) {
+    if (fingerprint == null) {
```
philipphofmann, line 179: `m`: What about throwing `IllegalArgumentException`?

sentry/src/test/java/io/sentry/protocol/UserTest.kt
```diff
@@ -79,7 +53,28 @@ class UserTest {
...
+
+    @Test
+    fun `setting null others do not crash`() {
+        val user = createUser()
+        user.others = null
+
+        assertNull(user.others)
+    }
+
+    private fun createUser(): User {
```
philipphofmann, line 68: Nice refactoring 💯

philipphofmann (commented):
The changes LGTM, but I think the title of the PR doesn't reflect all the changes in here. We make small fixes in both Scope and User. I think we can either change the title of the PR or we split this up into 2 PRs.

sentry/src/main/java/io/sentry/Scope.java
marandaneto: we cant do that
https://develop.sentry.dev/sdk/philosophy/#degrade-gracefully
this function is called by the users, not the SDK itself

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 * Feat: set isSideLoaded info tags
 * Enhancement: Read tracesSampleRate from AndroidManifest
 * Ref: Return NoOpTransaction instead of null (#1126)
+* Fix: User.setOthers is marked @Nullable but crashes when passing in null #1131
```
marandaneto, line 11:
```suggestion
* Fix: Do not crash when passing null values to @Nullable methods, eg User and Scope
```

marandaneto:
> The changes LGTM, but I think the title of the PR doesn't reflect all the changes in here. We make small fixes in both Scope and User. I think we can either change the title of the PR or we split this up into 2 PRs.

good catch, title and changelog are renamed

philipphofmann (approved):
LGTM

maciejwalkowiak (approved):

Issue #1123, opened 2020-12-17
Log errors while reading attachment


philipphofmann:
We currently put the error message when reading an attachment [into the attachment](https://github.com/getsentry/sentry-java/blob/753cf991d597ebc1766d15abb1b91125c178db66/sentry/src/main/java/io/sentry/SentryEnvelopeItem.java#L172-L186). This is not a good solution, because for example if the attachment is an image we just put the error message in the image, and the attachment is broken. Ideally, we would put them into [event errors](https://github.com/getsentry/develop/pull/228), but we need some changes in the SDK to achieve this. For the first release of attachments @bruno-garcia and I came to the conclusion we are just going to log the error message.

PR #1134, opened 2020-12-22
feat: Drop invalid attachments

philipphofmann:
## :scroll: Description
SentryEnvelopeItem.fromAttachment throws now a SentryEnvelopeException when it fails
to create an envelope item from an attachment. The GsonSerializer logs the exception
and drops the envelope item. So we avoid ingesting empty attachments.

This should have been a draft PR. It is not completely finished yet.

Fixes GH-1123

Unit tests and emulator

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/sentry/src/main/java/io/sentry/GsonSerializer.java
+++ b/sentry/src/main/java/io/sentry/GsonSerializer.java
@@ -169,13 +169,20 @@ public final class GsonSerializer implements ISerializer {
       writer.write("\n");
 
       for (final SentryEnvelopeItem item : envelope.getItems()) {
-        gson.toJson(item.getHeader(), SentryEnvelopeItemHeader.class, writer);
-        writer.write("\n");
-        writer.flush();
+        try {
+          final byte[] data = item.getData();
 
-        outputStream.write(item.getData());
+          gson.toJson(item.getHeader(), SentryEnvelopeItemHeader.class, writer);
+          writer.write("\n");
+          writer.flush();
 
-        writer.write("\n");
+          outputStream.write(data);
+
+          writer.write("\n");
+        } catch (Exception exception) {
+          logger.log(
+              SentryLevel.ERROR, "Failed to get data for envelope item. Dropping it.", exception);
+        }
       }
       writer.flush();
     }
--- a/sentry/src/main/java/io/sentry/GsonSerializer.java
+++ b/sentry/src/main/java/io/sentry/GsonSerializer.java
@@ -180,8 +180,7 @@ public final class GsonSerializer implements ISerializer {
 
           writer.write("\n");
         } catch (Exception exception) {
-          logger.log(
-              SentryLevel.ERROR, "Failed to get data for envelope item. Dropping it.", exception);
+          logger.log(SentryLevel.ERROR, "Failed to create envelope item. Dropping it.", exception);
         }
       }
       writer.flush();
--- a/sentry/src/main/java/io/sentry/GsonSerializer.java
+++ b/sentry/src/main/java/io/sentry/GsonSerializer.java
@@ -170,6 +170,7 @@ public final class GsonSerializer implements ISerializer {
 
       for (final SentryEnvelopeItem item : envelope.getItems()) {
         try {
+          // When this throws we don't write anything and continue with the next item.
           final byte[] data = item.getData();
 
           gson.toJson(item.getHeader(), SentryEnvelopeItemHeader.class, writer);
--- a/sentry/src/test/java/io/sentry/GsonSerializerTest.kt
+++ b/sentry/src/test/java/io/sentry/GsonSerializerTest.kt
@@ -4,8 +4,11 @@ import com.google.gson.JsonObject
 import com.google.gson.JsonParser
 import com.google.gson.JsonPrimitive
 import com.nhaarman.mockitokotlin2.any
+import com.nhaarman.mockitokotlin2.eq
 import com.nhaarman.mockitokotlin2.mock
+import com.nhaarman.mockitokotlin2.verify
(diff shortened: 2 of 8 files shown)


sentry/src/main/java/io/sentry/GsonSerializer.java
```diff
@@ -169,13 +169,19 @@ public void serialize(
       writer.write("\n");
 
       for (final SentryEnvelopeItem item : envelope.getItems()) {
-        gson.toJson(item.getHeader(), SentryEnvelopeItemHeader.class, writer);
-        writer.write("\n");
-        writer.flush();
+        try {
```
bruno-garcia, line 172: It's open up for the case of sending envelopes without items but if the stream in question is the network stream, it would be too late to do anything about it anyway unless we supported some trailing header to communicate the failure

bruno-garcia (approved):
LGTM!

sentry/src/main/java/io/sentry/GsonSerializer.java
marandaneto: I think its already done but if not yet, it'd be nice to check if the envelope items are not empty before calling this method

sentry/src/test/java/io/sentry/GsonSerializerTest.kt
```diff
@@ -31,7 +35,14 @@ import kotlin.test.assertTrue
...
 
-    private val serializer = GsonSerializer(mock(), EnvelopeReader())
+    private lateinit var logger: ILogger
+    private lateinit var serializer: GsonSerializer
+
+    @BeforeTest
+    fun before() {
+        logger = mock()
+        serializer = GsonSerializer(logger, EnvelopeReader())
+    }
```
marandaneto, line 45: M: what's about refactoring this and adding the Fixture?

sentry/src/test/java/io/sentry/SentryEnvelopeItemTest.kt
```diff
@@ -68,21 +67,21 @@ class SentryEnvelopeItemTest {
...
-        val item = SentryEnvelopeItem.fromAttachment(logger, attachment)
+        val item = SentryEnvelopeItem.fromAttachment(attachment)
 
-        assertAttachment(attachment,
-                "Reading the attachment ${attachment.pathname} failed, because the file located at " +
-                        "the path is not a file.", item)
+        assertFailsWith(SentryEnvelopeException::class, "Reading the attachment ${attachment.pathname} failed, because the file located at " +
+                "the path is not a file.") {
+            item.data
+        }
```
marandaneto, line 84: M: you can use the reified option

eg:
`assertFailsWith< SentryEnvelopeException > { val item = SentryEnvelopeItem.fromAttachment(attachment) }`

sentry/src/main/java/io/sentry/GsonSerializer.java
philipphofmann: @marandaneto what exactly do you mean by checking if the envelope items are not empty?

philipphofmann: Good point @bruno-garcia, but I think we already had the problem before. I think this change slightly improves it. 


**Given an envelope with two items and the first to send throws an exception.**
_Old_
Send header
Send 1st item -> Exception
Stop

We just sent an envelope without items.

_New_
Send header
Send 1st item -> Exception, Drop the item
Send 2nd 

We send the envelope but dropped the 1st item.


**Given an envelope with one item that throws.**
_Old and New_
Send header
Send 1st item -> Exception

We just sent an envelope without items.


For output streams that are all or nothing we just changed the behavior, because if there is one error in any of the items we previously discarded the whole envelope and now we keep the envelope. Therefore we keep an envelope without any items. I think we could solve this by throwing an exception when we serialize an envelope and we can't serialize a single item.

sentry/src/test/java/io/sentry/SentryEnvelopeItemTest.kt
philipphofmann: Good point thanks @marandaneto 👏🏼

sentry/src/main/java/io/sentry/GsonSerializer.java
marandaneto: > @marandaneto what exactly do you mean by checking if the envelope items are not empty?

I just checked up and we don't do it:

so here:
https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/transport/AsyncConnection.java#L156

we could do:

```
if (!envelope.getItems().isEmpty()) {
      executor.submit(new EnvelopeSender(envelope, hint, currentEnvelopeCache));
}
```

ps: `getItems` is an iterator so I don't think `isEmpty` would work, just a pseudocode

so we'd never try to send an envelope without items.

philipphofmann: I think we already check that in https://github.com/getsentry/sentry-java/blob/ef1eef782fb0eeb2e1a0e5e7fbbb00fe253be167/sentry/src/main/java/io/sentry/transport/AsyncConnection.java#L143, so I think it is not needed.

marandaneto: @philipphofmann this condition is only met if at least 1 item of the envelope is rate limited

marandaneto: it's not related to this PR though, and the backend accepts an empty envelope, so feel free to ignore it, I just wrote it because I thought that the backend would trigger an error, it doesn't, as said by Armin.

PR #1136, opened 2020-12-23
Add non blocking Apache HttpClient 5 based Transport.

maciejwalkowiak:
## :scroll: Description

Add non blocking Apache HttpClient 5 based Transport suitable for high traffic backend applications.

Current AsyncTransport can perform up to 3 requests per second which is too little for performance feature running in backend applications.

Unit tests plus sample app for running performance tests.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

Auto-configure new transport in Spring Boot auto-configuration when apache http client is on the classpath.

### DIFF
--- a/buildSrc/src/main/java/Config.kt
+++ b/buildSrc/src/main/java/Config.kt
@@ -57,6 +57,8 @@ object Config {
         val springAop = "org.springframework:spring-aop"
         val aspectj = "org.aspectj:aspectjweaver"
         val servletApi = "javax.servlet:javax.servlet-api"
+
+        val apacheHttpClient = "org.apache.httpcomponents.client5:httpclient5:5.0.3"
     }
 
     object AnnotationProcessors {
--- a/sentry/build.gradle.kts
+++ b/sentry/build.gradle.kts
@@ -19,6 +19,7 @@ tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach 
 dependencies {
     // Envelopes require JSON. Until a parse is done without GSON, we'll depend on it explicitly here
     implementation(Config.Libs.gson)
+    implementation(Config.Libs.apacheHttpClient)
 
     compileOnly(Config.CompileOnly.nopen)
     errorprone(Config.CompileOnly.nopenChecker)
--- a/sentry/src/main/java/io/sentry/ApacheHttpClientTransportFactory.java
+++ b/sentry/src/main/java/io/sentry/ApacheHttpClientTransportFactory.java
@@ -1,0 +1,55 @@
+package io.sentry;
+
+import io.sentry.transport.ApacheHttpClientTransport;
+import io.sentry.transport.ITransport;
+import io.sentry.transport.RateLimiter;
+import io.sentry.util.Objects;
+import org.apache.hc.client5.http.impl.DefaultConnectionKeepAliveStrategy;
+import org.apache.hc.client5.http.impl.async.CloseableHttpAsyncClient;
+import org.apache.hc.client5.http.impl.async.HttpAsyncClients;
+import org.apache.hc.client5.http.impl.nio.PoolingAsyncClientConnectionManager;
+import org.apache.hc.client5.http.impl.nio.PoolingAsyncClientConnectionManagerBuilder;
+import org.apache.hc.core5.http.impl.DefaultConnectionReuseStrategy;
+import org.apache.hc.core5.pool.PoolConcurrencyPolicy;
+import org.apache.hc.core5.util.TimeValue;
+import org.jetbrains.annotations.NotNull;
+
+/** Creates {@link ApacheHttpClientTransport}. */
+public final class ApacheHttpClientTransportFactory implements ITransportFactory {
+  private final @NotNull TimeValue connectionTimeToLive;
+
+  public ApacheHttpClientTransportFactory() {
+    this(TimeValue.ofMinutes(1));
+  }
+
+  public ApacheHttpClientTransportFactory(final @NotNull TimeValue connectionTimeToLive) {
+    this.connectionTimeToLive =
+        Objects.requireNonNull(connectionTimeToLive, "connectionTimeToLive is required");
+  }
+
+  @Override
+  public ITransport create(
+      final @NotNull SentryOptions options, final @NotNull RequestDetails requestDetails) {
+    Objects.requireNonNull(options, "options is required");
+    Objects.requireNonNull(requestDetails, "requestDetails is required");
+
+    final PoolingAsyncClientConnectionManager connectionManager =
+        PoolingAsyncClientConnectionManagerBuilder.create()
+            .setPoolConcurrencyPolicy(PoolConcurrencyPolicy.LAX)
+            .setConnectionTimeToLive(connectionTimeToLive)
+            .setMaxConnTotal(options.getMaxQueueSize())
+            .setMaxConnPerRoute(options.getMaxQueueSize())
+            .build();
+
+    final CloseableHttpAsyncClient httpclient =
+        HttpAsyncClients.custom()
+            .setKeepAliveStrategy(DefaultConnectionKeepAliveStrategy.INSTANCE)
+            .setConnectionManager(connectionManager)
+            .setConnectionReuseStrategy(DefaultConnectionReuseStrategy.INSTANCE)
+            .build();
+
+    final RateLimiter rateLimiter = new RateLimiter(options.getLogger());
+
+    return new ApacheHttpClientTransport(options, requestDetails, httpclient, rateLimiter);
+  }
+}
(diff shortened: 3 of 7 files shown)


maciejwalkowiak:
@marandaneto can you take a look what's the problem with Android sample here? There cannot be optional dependencies for the sentry module?

sentry/build.gradle.kts
```diff
@@ -19,6 +19,7 @@ tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach
 dependencies {
     // Envelopes require JSON. Until a parse is done without GSON, we'll depend on it explicitly here
     implementation(Config.Libs.gson)
+    implementation(Config.Libs.apacheHttpClient)
```
marandaneto, line 22: I think this is very problematic for Android, first because there's a dependency that is not used, and even worse, there are a bunch of transitive dependencies.
I'd even say that they are not Android compatible.

sentry/src/main/java/io/sentry/ApacheHttpClientTransportFactory.java
```diff
@@ -0,0 +1,55 @@
...
+import org.apache.hc.client5.http.impl.async.HttpAsyncClients;
+import org.apache.hc.client5.http.impl.nio.PoolingAsyncClientConnectionManager;
+import org.apache.hc.client5.http.impl.nio.PoolingAsyncClientConnectionManagerBuilder;
+import org.apache.hc.core5.http.impl.DefaultConnectionReuseStrategy;
+import org.apache.hc.core5.pool.PoolConcurrencyPolicy;
+import org.apache.hc.core5.util.TimeValue;
+import org.jetbrains.annotations.NotNull;
+
+/** Creates {@link ApacheHttpClientTransport}. */
+public final class ApacheHttpClientTransportFactory implements ITransportFactory {
```
marandaneto, line 18: I'd say that we should extract this to a separate module and the other integrations (spring, spring-boot) would pull this dependency instead of the `sentry` dependency directly.
like `sentry-nio`, or something like that, by default `sentry` has the standard BlockingHttpTransport.

marandaneto:
> @marandaneto can you take a look what's the problem with Android sample here? There cannot be optional dependencies for the sentry module?

see my comments, I bet they are not Android compatible.
I'm on vacation till the 4th so can't look into it in detail right now.

sentry/src/main/java/io/sentry/ApacheHttpClientTransportFactory.java
maciejwalkowiak: Sounds good. I'll move it to another module

maciejwalkowiak:
Thanks @marandaneto. Merry Christmas and see you on the 4th!

sentry/src/main/java/io/sentry/ApacheHttpClientTransportFactory.java
marandaneto: maybe @bruno-garcia has opinions here too

maciejwalkowiak: The module has been renamed.

maciejwalkowiak:
@marandaneto @bruno-garcia please find some time to review this PR.

bruno-garcia:
@maciejwalkowiak Manoel is back next week.

It LGTM but I have a note on the stream->byte array and the json type of the body, lets discuss that before merging

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@
 * Ref: `ITransport` implementations are now responsible for executing request in asynchronous or synchronous way (#1118)
 * Ref: Add option to set `TransportFactory` instead of `ITransport` on `SentryOptions` (#1124)
 * Ref: Simplify ITransport creation in ITransportFactory (#1135) 
+* Feat: Add non blocking Apache HttpClient 5 based Transport (#1136)
+* Enhancement: Autoconfigure Apache HttpClient 5 based Transport in Spring Boot integration (#1143)
```
marandaneto, line 17: this is part of the PR template next steps, but it's in the changelog as well, which one would be not up to date?

buildSrc/src/main/java/Config.kt
```diff
@@ -57,6 +57,8 @@ object Config {
         val springAop = "org.springframework:spring-aop"
         val aspectj = "org.aspectj:aspectjweaver"
         val servletApi = "javax.servlet:javax.servlet-api"
+
+        val apacheHttpClient = "org.apache.httpcomponents.client5:httpclient5:5.0.3"
```
marandaneto, line 61: 1st stable version of this lib v5 appeared in Feb, 2020, do you think that the majority of Apps have been upgraded already?

sentry-apache-http-client-5/src/main/java/io/sentry/transport/apache/ApacheHttpClientTransport.java
```diff
@@ -0,0 +1,147 @@
...
+          if (options.getLogger().isEnabled(DEBUG)) {
+            options.getLogger().log(DEBUG, "Currently running %d requests", currentlyRunning.get());
+          }
+
+          httpclient.execute(
+              request,
+              new FutureCallback<SimpleHttpResponse>() {
+                @Override
+                public void completed(SimpleHttpResponse response) {
+                  currentlyRunning.decrementAndGet();
```
marandaneto, line 95: should this be the last line to be called? theoretically there's still something happening.

sentry-apache-http-client-5/src/main/java/io/sentry/transport/apache/ApacheHttpClientTransport.java
```diff
@@ -0,0 +1,147 @@
...
+        try (final ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+            final GZIPOutputStream gzip = new GZIPOutputStream(outputStream)) {
+          options.getSerializer().serialize(filteredEnvelope, gzip);
+
+          final SimpleHttpRequest request =
+              SimpleHttpRequests.post(requestDetails.getUrl().toString());
+          request.setBody(outputStream.toByteArray(), ContentType.APPLICATION_JSON);
+          request.setHeader("Content-Encoding", "gzip");
+          request.setHeader("Content-Type", "application/x-sentry-envelope");
+          request.setHeader("Accept", "application/json");
```
marandaneto, line 80: Should this be part of RequestDetailsResolver or even extracting to a new class? it's a duplicated code right now.

bruno-garcia: If it's just 3 lines of code I wouldn't bother

marandaneto: if `RequestDetailsResolver` is used in both transports and they are meant to configure the connection with its headers, I don't see why duplicating it though.
https://github.com/getsentry/sentry-java/blob/gh-1097-apache-non-blocking-transport/sentry/src/main/java/io/sentry/transport/HttpConnection.java#L124-L126

sentry-apache-http-client-5/src/main/java/io/sentry/transport/apache/ApacheHttpClientTransport.java
```diff
@@ -0,0 +1,147 @@
...
+      if (filteredEnvelope != null) {
+        currentlyRunning.incrementAndGet();
+
+        try (final ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+            final GZIPOutputStream gzip = new GZIPOutputStream(outputStream)) {
+          options.getSerializer().serialize(filteredEnvelope, gzip);
+
+          final SimpleHttpRequest request =
+              SimpleHttpRequests.post(requestDetails.getUrl().toString());
+          request.setBody(outputStream.toByteArray(), ContentType.APPLICATION_JSON);
```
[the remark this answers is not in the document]
maciejwalkowiak: > Reading the stream into a bytearray sounds suboptimal.

As discussed it will be changed in the future. There is no straightforward way to write to output of the Apache client's request.

buildSrc/src/main/java/Config.kt
maciejwalkowiak: There is no package collision between version 5 and 4 so even if they did not, it means just an extra dependency. We still can add transport based on 4.x if there will be such request.

sentry-apache-http-client-5/src/main/java/io/sentry/transport/apache/ApacheHttpClientTransport.java
maciejwalkowiak: There is but it means that the connection is not occupied anymore. I don't see any risk here.

sentry-apache-http-client-5/src/main/java/io/sentry/transport/apache/ApacheHttpClientTransport.java
maciejwalkowiak: It's up to the transport to decide if it sends gzip or not, I also don't think that we should set fixed accept header. `RequestDetailsResolver` returns headers that users MUST send to Sentry.

maciejwalkowiak:
@bruno-garcia @marandaneto take another look please.

bruno-garcia (approved):

PR #1137, opened 2020-12-23
Ref: Bump AGP to 4.2.x and fix linter issues

marandaneto:
## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps
Fix all the `LambdaLast`, it's breaking changes though.
Fix all the `UnknownNullness` issues, quite a lot.

### DIFF
--- a/sentry-android-core/build.gradle.kts
+++ b/sentry-android-core/build.gradle.kts
@@ -56,6 +56,8 @@ android {
 
         // We run a full lint analysis as build part in CI, so skip vital checks for assemble tasks.
         isCheckReleaseBuilds = false
+
+        disable.add("UnknownNullness")
     }
 
     // needed because of Kotlin 1.4.x
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidLogger.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidLogger.java
@@ -9,7 +9,7 @@ final class AndroidLogger implements ILogger {
 
   private static final String tag = "Sentry";
 
-  @SuppressWarnings("AnnotateFormatMethod")
+  @SuppressWarnings({"AnnotateFormatMethod", "LogConditional"})
   @Override
   public void log(SentryLevel level, String message, Object... args) {
     Log.println(toLogcatLevel(level), tag, String.format(message, args));
--- a/sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserverIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserverIntegration.java
@@ -17,6 +17,7 @@ public abstract class EnvelopeFileObserverIntegration implements Integration, Cl
   private @Nullable EnvelopeFileObserver observer;
   private @Nullable ILogger logger;
 
+  @SuppressWarnings("SyntheticAccessor")
   public static @NotNull EnvelopeFileObserverIntegration getOutboxFileObserver() {
     return new OutboxEnvelopeFileObserverIntegration();
   }
@@ -67,6 +68,7 @@ public abstract class EnvelopeFileObserverIntegration implements Integration, Cl
   @TestOnly
   abstract @Nullable String getPath(final @NotNull SentryOptions options);
 
+  @SuppressWarnings("SyntheticAccessor")
   private static final class OutboxEnvelopeFileObserverIntegration
       extends EnvelopeFileObserverIntegration {
 
--- a/sentry-test-support/src/main/kotlin/io/sentry/test/assertions.kt
+++ b/sentry-test-support/src/main/kotlin/io/sentry/test/assertions.kt
@@ -10,7 +10,7 @@ import io.sentry.SentryOptions
 /**
  * Verifies is [SentryEnvelope] contains first event matching a predicate.
  */
-inline fun checkEvent(noinline predicate: (SentryEvent) -> Unit): SentryEnvelope {
+fun checkEvent(predicate: (SentryEvent) -> Unit): SentryEnvelope {
     val options = SentryOptions().apply {
         setSerializer(GsonSerializer(NoOpLogger.getInstance(), envelopeReader))
     }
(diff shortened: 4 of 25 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/AndroidLogger.java
```diff
@@ -9,7 +9,7 @@
 
   private static final String tag = "Sentry";
 
-  @SuppressWarnings("AnnotateFormatMethod")
+  @SuppressWarnings({"AnnotateFormatMethod", "LogConditional"})
```
marandaneto, line 12: we check if logger is enabled by ourselves, we don't need to check the `BuildConfig.DEBUG`
Sentry logger != Android logger

sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserverIntegration.java
```diff
@@ -17,6 +17,7 @@
   private @Nullable EnvelopeFileObserver observer;
   private @Nullable ILogger logger;
 
+  @SuppressWarnings("SyntheticAccessor")
```
marandaneto, line 20: there are a bunch of SyntheticAccessor errors for private static final classes, not sure if they make sense or are false positive, I don't see how could it be problematic

sentry-test-support/src/main/kotlin/io/sentry/test/assertions.kt
```diff
@@ -10,7 +10,7 @@ import io.sentry.SentryOptions
 /**
  * Verifies is [SentryEnvelope] contains first event matching a predicate.
  */
-inline fun checkEvent(noinline predicate: (SentryEvent) -> Unit): SentryEnvelope {
+fun checkEvent(predicate: (SentryEvent) -> Unit): SentryEnvelope {
```
marandaneto, line 13: no reason to be inlined, first because has no effect for this usecase, 2nd because its a testing class

sentry/src/main/java/io/sentry/CircularFifoQueue.java
```diff
@@ -335,6 +335,7 @@ private int decrement(int index) {
    *
    * @return an iterator over this queue's elements
    */
+  @SuppressWarnings("SyntheticAccessor")
```
marandaneto, line 338: I prefer to not touch this class that is borrowed from Apache package

sentry/src/main/java/io/sentry/Hub.java
```diff
@@ -69,7 +69,7 @@ public boolean isEnabled() {
   public @NotNull SentryId captureEvent(
       final @NotNull SentryEvent event, final @Nullable Object hint) {
     SentryId sentryId = SentryId.EMPTY_ID;
-    if (!isEnabled()) {
+    if (!isEnabled) {
```
marandaneto, line 72: access to the field is always faster than a getter, on Java.
on Dart, for example, field access and getters are the same thing.

sentry/src/main/java/io/sentry/Scope.java
```diff
@@ -165,6 +165,7 @@ public void setUser(final @Nullable User user) {
    *
    * @return the fingerprint list
    */
+  @SuppressWarnings("KotlinPropertyAccess")
```
marandaneto, line 168: they cant be public, users only have access to setters, but it cant be private as well, we need to access from a different package

sentry/src/main/java/io/sentry/Sentry.java
```diff
@@ -90,6 +90,7 @@ public static void init() {
    * @throws NoSuchMethodException the NoSuchMethodException
    * @throws InvocationTargetException the InvocationTargetException
    */
+  @SuppressWarnings("LambdaLast")
```
marandaneto, line 93: fixing all of those would be great, so Kotlin could use the last argument with a lambda if its an interface function, but would be a breaking change

sentry/src/main/java/io/sentry/protocol/App.java
```diff
@@ -34,7 +35,7 @@
   @SuppressWarnings("unused")
   private Map<String, Object> unknown;
 
-  public String getAppIdentifier() {
+  public @Nullable String getAppIdentifier() {
```
marandaneto, line 38: this would bring better interopatibility with Kotlin null safety, but yeah, maybe around 700 annotations to be added

sentry/src/main/java/io/sentry/transport/AsyncConnection.java
```diff
@@ -62,6 +63,7 @@ public AsyncConnection(
     this.executor = executorService;
   }
 
+  @SuppressWarnings("SyntheticAccessor")
```
marandaneto, line 66: inner classes have access to private fields but this has side effects, see https://stackoverflow.com/questions/4501324/synthetic-accessor-method-warning
so either we create a getter which cant be private, nor protected as the class is final, but I'd not like to make it public, to passing the instance is the right choice

sentry-android-core/build.gradle.kts
```diff
@@ -56,6 +56,8 @@ android {
 
         // We run a full lint analysis as build part in CI, so skip vital checks for assemble tasks.
         isCheckReleaseBuilds = false
+
+        disable.add("UnknownNullness")
```
marandaneto, line 60: @philipphofmann help me to fix those
so no manoelly linter :)

bruno-garcia: LOL

marandaneto: basic all public getters and setters would need either @NotNull or @Nullable, overwhelming hehe, I'd guess around 700

marandaneto: I just added this disable rule to make CI happier and focus on the other lint issues.

bruno-garcia: image

sentry/src/main/java/io/sentry/Sentry.java
marandaneto: this is called trailing lambda btw

sentry/src/main/java/io/sentry/Scope.java
marandaneto: maybe we should add `KotlinPropertyAccess` as excluding rule, otherwise, it's gonna be annoying

PR #1142, opened 2020-12-30
Read `uncaught.handler.enabled` property from the external configuration

### LINKED ISSUE #1112: Sentry Logback + default UncaughtExceptionHandler duplication
Platform:

    Java -> If yes, which Java (and sourceCompatibility/targetCompatibility) version? openjdk 11.0.8 2020-07-14
    Logback -> If yes, which version? 1.2.3

Build system:

    Other -> If yes, which one? sbt

Platform installed with:

    Maven Central

The version of the SDK:
3.2.0

---
I have the following issue:

This is somewhat specific for the logback integration

By default sentry changes the default jvm behaviour by making it no longer log stack traces, with a possibility that it can't send the error to sentry.

If we make sure we have our own handler configured before any logging starts up, and that handler writes the error to logback, we end up with duplicate entries in sentry. As the logback appender will log the error and the sentry exception handler will log it.

So on 1.x we disabled the handler using the sentry.properties file, however the `uncaught.handler.enabled=false` config option no longer exists in 3.x

Also we don't want compile time dependencies and configure everything using the env/properties/system properties. So calling setEnableUncaughtExceptionHandler is not an option.

**Steps to reproduce:**
- set up sentry logback integration
- in your main set up a UncaughtExceptionHandler that logs to logback (make sure no logging happens before this)
- now add `throw new RuntimeException("oops")`

**Actual result:**
The error is reported twice, once by the logback appender and once by the UncaughtExceptionHandler

**Expected result:**

- One log entry in the logs with the error and trace
- One error reported towards sentry

(previously reported on #1077 but not really related)

maciejwalkowiak:
## :scroll: Description

Read `uncaught.handler.enabled` property from the external configuration.

Fixes #1112 

Unit tests.

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 * Fix: Initialize Logback after context refreshes (#1129)
 * Ref: Return NoOpTransaction instead of null (#1126)
 * Fix: Do not crash when passing null values to @Nullable methods, eg User and Scope
-* Enhancement: Read `uncaught.handler.enabled` property from the external configuration 
+* feat: Read `uncaught.handler.enabled` property from the external configuration 
 
 # 4.0.0-alpha.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * Fix: Initialize Logback after context refreshes (#1129)
 * Ref: Return NoOpTransaction instead of null (#1126)
 * Fix: Do not crash when passing null values to @Nullable methods, eg User and Scope
+* Enhancement: Read `uncaught.handler.enabled` property from the external configuration 
 
 # 4.0.0-alpha.2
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 * Fix: Initialize Logback after context refreshes (#1129)
 * Ref: Return NoOpTransaction instead of null (#1126)
 * Fix: Do not crash when passing null values to @Nullable methods, eg User and Scope
+* Enhancement: Read `uncaught.handler.enabled` property from the external configuration
```
bruno-garcia, line 14: Lets just use `feat` like the other projects in Sentry:

```suggestion
* feat: Read `uncaught.handler.enabled` property from the external configuration 
```

bruno-garcia (approved):

PR #1144, opened 2020-12-30
Send user.ip_address = {{auto}} when sendDefaultPii is true

### LINKED ISSUE #1015: Send user.ip_address = {{ auto }} when sendDefaultPii is true
If no explicit IpAddress was set by the user code (or framework integration) the SDK should set `{{ auto }}` which tells the server to infer the IP Address to client connection as the `user.ip_address`.

This is true for all SDKs. Like: https://github.com/getsentry/sentry-dotnet/issues/561

maciejwalkowiak:
## :scroll: Description

Send user.ip_address = {{ auto }} when sendDefaultPii is true

Fixes #1015

Unit tests.

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # vNext
 
-* Feat: Drop invalid attachments #1134
+* Feat: Add maxAttachmentSize to SentryOptions (#1138)
+* Feat: Drop invalid attachments (#1134)
 * Ref: Make Attachment immutable (#1120)
 * Fix inheriting sampling decision from parent (#1100)
 * Fixes and Tests: Session serialization and deserialization
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 * Fix: Initialize Logback after context refreshes (#1129)
 * Ref: Return NoOpTransaction instead of null (#1126)
 * Fix: Do not crash when passing null values to @Nullable methods, eg User and Scope
-* Enhancement: Send user.ip_address = {{ auto }} when sendDefaultPii is true
+* Enhancement: Send user.ip_address = {{auto}} when sendDefaultPii is true
 * Fix: Resolving dashed properties from external configuration
 * Feat: Read `uncaught.handler.enabled` property from the external configuration 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@
 * Ref: Return NoOpTransaction instead of null (#1126)
 * Fix: Do not crash when passing null values to @Nullable methods, eg User and Scope
 * Enhancement: Send user.ip_address = {{ auto }} when sendDefaultPii is true
+* Fix: Resolving dashed properties from external configuration
+* Feat: Read `uncaught.handler.enabled` property from the external configuration 
 
 # 4.0.0-alpha.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Fix: Initialize Logback after context refreshes (#1129)
 * Ref: Return NoOpTransaction instead of null (#1126)
 * Fix: Do not crash when passing null values to @Nullable methods, eg User and Scope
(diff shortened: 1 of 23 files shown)


sentry/src/main/java/io/sentry/MainEventProcessor.java
```diff
@@ -123,5 +130,14 @@ private void processNonCachedEvent(final @NotNull SentryEvent event) {
         event.setThreads(sentryThreadFactory.getCurrentThread());
       }
     }
+    if (options.isSendDefaultPii()) {
+      if (event.getUser() == null) {
+        final User user = new User();
+        user.setIpAddress(DEFAULT_IP_ADDRESS);
+        event.setUser(user);
+      } else if (event.getUser().getIpAddress() == null) {
+        event.getUser().setIpAddress(DEFAULT_IP_ADDRESS);
```
bruno-garcia, line 139: You probably want to _undo_ this in the `spring` integration, since `{{auto}}` there will give the IP of the server.

Like check for default and set null instead.

maciejwalkowiak: In such case wouldn't it be better if we make it an opt in which is pre-configured in Android? Undoing it in Spring looks for me like a workaround.

bruno-garcia: @maciejwalkowiak That means desktop apps would have to opt-in too.
I think it's fair to say: It's enabled by default but you can opt-out. And we know better in some integrations to opt out already.

We need to document this behavior. Similar to how the default of another option is documented here:
https://github.com/getsentry/sentry-docs/pull/2815

marandaneto: if we check for `{{auto}}` and set it to null if `isSendDefaultPii` is disabled would be a breaking change btw which is fine as 4.0 is a major bump, but yeah we'd need to document.

Also, it could be written `{{ auto }}` with spaces, so we'd need to trim spaces off if we want to compare that properly as well, otherwise, it'd not be a match and IPs will be set by the server.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 * Fix: Initialize Logback after context refreshes (#1129)
 * Ref: Return NoOpTransaction instead of null (#1126)
 * Fix: Do not crash when passing null values to @Nullable methods, eg User and Scope
+* Enhancement: Send user.ip_address = {{ auto }} when sendDefaultPii is true
```
marandaneto, line 15:
```suggestion
* Enhancement: Send user.ip_address = {{auto}} when sendDefaultPii is true
```

marandaneto: or change the value of `DEFAULT_IP_ADDRESS`

sentry/src/main/java/io/sentry/MainEventProcessor.java
marandaneto: this is breaking the Android user, see:

https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java#L151-L154

now the `user` will never be null, we need to fix the `if` condition in the `DefaultAndroidEventProcessor` class

maciejwalkowiak: Undoing in Spring integration ✅
Updating Android code ✅

bruno-garcia (approved):

sentry/src/main/java/io/sentry/MainEventProcessor.java
marandaneto: this has been merged but there's a bug here.

`{{ auto }}` is valid
`{{auto}}` is also valid

previously `{{auto}}` was not valid and its been fixed on the backend, so people were setting `{{ auto }}` manually, those people, it won't work as the comparison is not trimming off the spaces.

maciejwalkowiak: Got it. Will be fixed in next PR.

PR #1146, opened 2020-12-30
Resolve servername from the localhost address.

### LINKED ISSUE #1109: Servername no longer defaults to hostname
_Platform:_
*  Java -> If yes, which Java (and sourceCompatibility/targetCompatibility) version? openjdk 11.0.8 2020-07-14
*  Logback -> If yes, which version? 1.2.3

_Build system:_
* Other -> If yes, which one? sbt

_Platform installed with:_
* Maven Central

The version of the SDK:
**3.2.0**

---
I have the following issue:

this is a continuation of #992

On sentry 1.x servername was automatically set by picking up the local hostname
On sentry 3.x that default seems to be removed

The legacy docs have this section:

> For many integrations the server name actually corresponds to the device hostname, even in situations where the machine is not actually a server. Most SDKs will attempt to auto-discover this value.

We try to configure sentry without binary dependencies (only properties file/env vars/system properties). Is there a property we can set to have sentry look up the hostname like it used to do?

maciejwalkowiak:
## :scroll: Description

Resolve servername from the localhost address.

Fixes #1109.

Large piece of code - resolving and caching the server name has been copied from 1.7.x and updated to new standards.

Unit tests.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


sentry/src/main/java/io/sentry/ServerNameResolvingEventProcessor.java
```diff
@@ -0,0 +1,134 @@
...
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * An {@link EventProcessor} that sets the local ip address to an {@link
+ * SentryEvent#getServerName()} property.
+ *
+ * <p>Note - this event processor is not used by SDK by default and must be configured manually on
+ * {@link SentryOptions} if there is intention to use it.
+ */
+public final class ServerNameResolvingEventProcessor implements EventProcessor {
```
bruno-garcia, line 21: We could add this to the list automatically if `sendDefaultPii` is set to true in Java. In the core SDK I mean. 

Though we should probably remote this on Android. /cc @marandaneto

maciejwalkowiak: Similar to https://github.com/getsentry/sentry-java/pull/1144#discussion_r550449131 - isn't it better to keep it as an opt-in rather than something our other integrations have to opt-out from?

bruno-garcia: Since we can just use the IP address, maybe opt-in for this one sounds better (since it's not so cheap to get it)

sentry/src/main/java/io/sentry/ServerNameResolvingEventProcessor.java
```diff
@@ -0,0 +1,134 @@
...
+    @Nullable private volatile String hostname;
+    /** Time at which the cache should expire. */
+    private volatile long expirationTimestamp;
+    /** Whether a cache update thread is currently running or not. */
+    private final @NotNull AtomicBoolean updateRunning = new AtomicBoolean(false);
+
+    private final @NotNull Callable<InetAddress> getLocalhost;
+
+    private HostnameCache(long cacheDuration) {
+      this(cacheDuration, InetAddress::getLocalHost);
```
marandaneto, line 73: method references on Android doesn't work on all the versions and linter also does not complain :(

sentry/src/main/java/io/sentry/ServerNameResolvingEventProcessor.java
```diff
@@ -0,0 +1,134 @@
...
+            } finally {
+              updateRunning.set(false);
+            }
+
+            return null;
+          };
+
+      try {
+        final FutureTask<Void> futureTask = new FutureTask<>(hostRetriever);
+        new Thread(futureTask).start();
```
marandaneto, line 120: should we use a `Executors.newSingleThreadExecutor()` and `submit` which returns a `Future` already? not a big fan of bare Threads

sentry/src/main/java/io/sentry/ServerNameResolvingEventProcessor.java
```diff
@@ -0,0 +1,134 @@
...
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+        handleCacheUpdateFailure();
+      } catch (ExecutionException | TimeoutException | RuntimeException e) {
+        handleCacheUpdateFailure();
+      }
+    }
+
+    private void handleCacheUpdateFailure() {
+      expirationTimestamp = System.currentTimeMillis() + TimeUnit.SECONDS.toMillis(1);
```
marandaneto, line 131: there's a class called `CurrentDateProvider` which returns a `System.currentTimeMillis()` so you could unit test if necessary.

sentry/src/main/java/io/sentry/ServerNameResolvingEventProcessor.java
```diff
@@ -0,0 +1,134 @@
...
+ * {@link SentryOptions} if there is intention to use it.
+ */
+public final class ServerNameResolvingEventProcessor implements EventProcessor {
+
+  /**
+   * Duration of the hostname caching.
+   *
+   * @see HostnameCache
+   */
+  private static final long HOSTNAME_CACHE_DURATION = TimeUnit.HOURS.toMillis(5);
```
marandaneto, line 28: is it a common thing to change the server name? 5h seems too little, looks like also something we could have on the options, but not worth it at the moment

sentry/src/main/java/io/sentry/ServerNameResolvingEventProcessor.java
marandaneto: on Android, it won't make sense, adding by default would require that the Android integration removes it,  we'd be allocating something that would not be used by the Android integration and this kinda slows down the App. init (if every option would be added by default and removed by Android) which has limited resources.

I'd prefer to add by default in the needed integrations (Spring, etc...), the downside is, when somebody depends on `sentry` only, this integration won't be there.

I guess this trade-off will become more common with the number of features that have been adding to the `sentry` package, wondering if a `flag` on options would make sense, which differs mobile from different environments and based on that, we allocate those objects or not.

sentry/src/main/java/io/sentry/ServerNameResolvingEventProcessor.java
marandaneto: https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java#L103-L104

sentry/src/main/java/io/sentry/ServerNameResolvingEventProcessor.java
maciejwalkowiak: I honestly don't know. This is how it was implemented in 1.x so I believe there was a reason for that.

sentry/src/main/java/io/sentry/ServerNameResolvingEventProcessor.java
maciejwalkowiak: If opting out would be removing integrations by name or class name I think it could be an option. But currently there's no simple way of opting out from event processor activated by default. 

We could have something like: `SentryOptions#removeEventProcessor(ServerNameResolvingEventProcessor.class)`

But as @marandaneto said, it's too late, objects have already been initialized. I remain in favour of activating things where we are sure they make sense and keeping `sentry` core as the lowest common denominator.

bruno-garcia: Folks at Sentry have strong opinions about default EventProcessors. The idea is that the SDK shouldn't ship any eventprocessor but integrations instead (that dig through the static API to find the client, then the options, to then look inside the integrations list of that options object, and only then do its work or not.

That said, I also prefer not to have such cryptic API like `SentryOptions#removeEventProcessor(ServerNameResolvingEventProcessor.class)`. I rather have proper Options for things when possible.

We can have an option such as `AttachServerName` and if this flag is on, the SDK adds the integration. On Android we can just `new SentryOptions` and set this flat to false so it won't mean any runtime cost.

sentry/src/main/java/io/sentry/ServerNameResolvingEventProcessor.java
```diff
@@ -0,0 +1,134 @@
...
+     *
+     * @return the hostname of the current machine.
+     */
+    private @Nullable String getHostname() {
+      if (expirationTimestamp < System.currentTimeMillis()
+          && updateRunning.compareAndSet(false, true)) {
+        updateCache();
+      }
+
+      return hostname;
```
bruno-garcia, line 101: First time we capture an error it just doesn't include a hostname? It scheduled a thread to run and returns `null`?

If so, should we call `updateCache()` to schedule the thread when the class is constructed?

sentry/src/main/java/io/sentry/ServerNameResolvingEventProcessor.java
bruno-garcia: If it fails to resolve, it'll kick off again after 1 second.
We catch any error here and set it to 1 second again. So basically this can mean we keep spawning a new thread every second. I think in case of error we should just stop running this and just return `null`.

WHat makes this fail? Are those errors recoverable? is this a reverse DNS query on the local interface IP addresses?

sentry/src/main/java/io/sentry/ServerNameResolvingEventProcessor.java
maciejwalkowiak: Agreed!

sentry/src/main/java/io/sentry/ServerNameResolvingEventProcessor.java
maciejwalkowiak: True!

maciejwalkowiak:
@bruno-garcia I can't think of scenario where resolving server name would fail, as `InetAddress#getHostFromNameService` in case of failure returns just an ip address.

Btw, this is how it was implemented in 1.x.

sentry/src/main/java/io/sentry/ServerNameResolvingEventProcessor.java
maciejwalkowiak: Failure is unlikely to happen, I can't think of the case when it would actually happen. Code has changed to use single daemon thread so that at least we don't spawn new threads all the time.

sentry/src/main/java/io/sentry/ServerNameResolvingEventProcessor.java
maciejwalkowiak: Understood. It's enabled by default now and new flag has been added to `SentryOptions`.

sentry/src/main/java/io/sentry/ServerNameResolvingEventProcessor.java
marandaneto: I see, makes sense, not sure if it's worth copying blindly tbh, it's a lot of code and complexity for a simple workaround that would be restarting/redeploying the App, of course, it's suboptimal, but a redeploy would happen soon or later right.

thoughts @bruno-garcia ?

maciejwalkowiak:
It's now disabled for android and turned on by default when just using Sentry dependency.

bruno-garcia (approved):

PR #1156, opened 2021-01-11
Automatically assign span context to captured events.

maciejwalkowiak:
## :scroll: Description

Automatically assign span context to captured events.

One step less for users using manual performance reporting.
Result of the discussion: https://github.com/getsentry/sentry-docs/pull/2818/files#r551077018

Unit & Integration tests.

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

### DIFF
--- a/sentry/src/main/java/io/sentry/Hub.java
+++ b/sentry/src/main/java/io/sentry/Hub.java
@@ -78,6 +78,7 @@ public final class Hub implements IHub {
       options.getLogger().log(SentryLevel.WARNING, "captureEvent called with null parameter.");
     } else {
       try {
+        assignTraceContext(event);
         final StackItem item = stack.peek();
         sentryId = item.getClient().captureEvent(event, item.getScope(), hint);
       } catch (Exception e) {
@@ -154,7 +155,9 @@ public final class Hub implements IHub {
     } else {
       try {
         final StackItem item = stack.peek();
-        sentryId = item.getClient().captureException(throwable, item.getScope(), hint);
+        final SentryEvent event = new SentryEvent(throwable);
+        assignTraceContext(event);
+        sentryId = item.getClient().captureEvent(event, item.getScope(), hint);
       } catch (Exception e) {
         options
             .getLogger()
@@ -164,6 +167,15 @@ public final class Hub implements IHub {
     }
     this.lastEventId = sentryId;
     return sentryId;
+  }
+
+  private void assignTraceContext(final @NotNull SentryEvent event) {
+    if (event.getThrowable() != null) {
+      final SpanContext spanContext = throwableToSpanContext.get(event.getThrowable());
+      if (spanContext != null && event.getContexts().getTrace() == null) {
+        event.getContexts().setTrace(spanContext);
+      }
+    }
   }
 
   @Override
--- a/sentry/src/main/java/io/sentry/Hub.java
+++ b/sentry/src/main/java/io/sentry/Hub.java
@@ -3,6 +3,7 @@ package io.sentry;
 import io.sentry.Stack.StackItem;
 import io.sentry.hints.SessionEndHint;
 import io.sentry.hints.SessionStartHint;
+import io.sentry.protocol.Contexts;
 import io.sentry.protocol.SentryId;
 import io.sentry.protocol.User;
 import io.sentry.util.Objects;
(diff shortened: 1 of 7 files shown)


sentry/src/main/java/io/sentry/Hub.java
```diff
@@ -166,6 +169,15 @@ public SentryId captureEnvelope(
     return sentryId;
   }
 
+  private void assignTraceContext(final @NotNull SentryEvent event) {
+    if (event.getThrowable() != null) {
+      final SpanContext spanContext = throwableToSpanContext.get(event.getThrowable());
+      if (spanContext != null && event.getContexts().getTrace() == null) {
```
bruno-garcia, line 175: If these have public setters and are nullable we should navigate through considering they might be.

sentry/src/main/java/io/sentry/ISentryClient.java
```diff
@@ -95,52 +95,6 @@ default SentryId captureMessage(String message, SentryLevel level) {
     return captureMessage(message, level, null);
   }
 
-  /**
```
bruno-garcia, line 98: Why are we getting rid of this API?

bruno-garcia:
Could you please attach a screenshot of what the issue event look like having that context attached to it?

sentry/src/main/java/io/sentry/ISentryClient.java
marandaneto: looks like wrong

maciejwalkowiak: When exception is captured there is a lookup in Hub to assign a trace context. If we allow SentryEvents to be constructed in the `SentryClient` they will never have trace context attached. 

I don't see the reason why should we keep these methods in `ISentryClient` but maybe I am missing something?

marandaneto: Java 1.7.x didn't have the Static class but only "SentryClient" and they were using it directly, I guess this could break a lot of user's code (if people migrated the easy way).
Also, people developing libraries would not use the Static class, either Hub or SentryClient, that would be another segment that would be broken.
I don't think removing them would be a good thing, but rather documenting the limitation.
the other possibility would be to `assignTraceContext` in the SentryClient, but comes with the probably of whom should hold the state, Hub or SentryClient? in theory, only the Hub

maciejwalkowiak:
image
image

I do not know why in event trace details are also shown on the righthand side in tags and in other they are not.

sentry/src/main/java/io/sentry/Hub.java
maciejwalkowiak: Fixed.

sentry/src/main/java/io/sentry/ISentryClient.java
maciejwalkowiak: Yes it's about the state - that has been already discussed. I'm ok to bring back these methods. Maybe I misunderstood @bruno-garcia when we discussed removing them as they are a part of interface that's not used directly by users anyway.

bruno-garcia: Oh I think there was a misunderstanding.
Instead of removing this, lets just document that for connecting errors to span, `Hub` must be involved since that's what adds state to Sentry. Accessing the Client doesn't add any Scope or other state so it shouldn't be a surprise

bruno-garcia: Also, can't we inspect `event.getThrowable()` when one calls `captureEvent`?

sentry/src/main/java/io/sentry/Hub.java
```diff
@@ -154,7 +156,9 @@ public SentryId captureEnvelope(
     } else {
       try {
         final StackItem item = stack.peek();
-        sentryId = item.getClient().captureException(throwable, item.getScope(), hint);
+        final SentryEvent event = new SentryEvent(throwable);
```
bruno-garcia, line 159: THis LGTM, we could just bring back the default methods on `ISentryClient` and merge

sentry/src/main/java/io/sentry/ISentryClient.java
maciejwalkowiak: Methods are back now.

bruno-garcia (approved):

Issue #1160, opened 2021-01-13
'SentryTransaction' throws if 'getSpanId()' is called and there's no 'contexts' or 'trace'


marandaneto:
We'd need to revisit every method call and annotation, as we can't annotate `getSpanId()` with `@NotNull` if `contexts` and `trace` are `@Nullable`.
The same issue applies to a few other methods.

Adding `throws` to the method signature is also ugly, we don't want the user's code handling exception on every method call, ideally, we'd return something similar to NoOp if possible.

cc @maciejwalkowiak

marandaneto:
fixed by #1173

PR #1164, opened 2021-01-14
Fix potential NPE by separating transaction contexts.

maciejwalkowiak:
## :scroll: Description

Fix potential NPE by separating transaction contexts from event contexts.

Even though event contexts look similar to transaction contexts, there are different nullability constraints - with transactions, its guaranteed that trace cannot be null. With this change, we fix potential NPE reported in #1160.

Unit tests.

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps

Perhaps change the `Contexts` class to also not inherit from the HashMap?

### DIFF
--- a/sentry/src/main/java/io/sentry/SentryEvent.java
+++ b/sentry/src/main/java/io/sentry/SentryEvent.java
@@ -97,6 +97,9 @@ public final class SentryEvent extends SentryBaseEvent implements IUnknownProper
   /** List of breadcrumbs recorded before this event. */
   private List<Breadcrumb> breadcrumbs;
 
+  /** Contexts describing the environment (e.g. device, os or browser). */
+  private final @NotNull Contexts contexts = new Contexts();
+
   /**
    * Arbitrary extra information set by the user.
    *
@@ -259,6 +262,10 @@ public final class SentryEvent extends SentryBaseEvent implements IUnknownProper
 
   public void addBreadcrumb(final @Nullable String message) {
     this.addBreadcrumb(new Breadcrumb(message));
+  }
+
+  public @NotNull Contexts getContexts() {
+    return contexts;
   }
 
   Map<String, Object> getExtras() {
--- a/sentry/src/main/java/io/sentry/TransactionContexts.java
+++ b/sentry/src/main/java/io/sentry/TransactionContexts.java
@@ -1,1 +1,93 @@
+package io.sentry;
 
+import io.sentry.protocol.App;
+import io.sentry.protocol.Browser;
+import io.sentry.protocol.Device;
+import io.sentry.protocol.Gpu;
+import io.sentry.protocol.OperatingSystem;
+import io.sentry.protocol.SentryRuntime;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public final class TransactionContexts {
+  private @Nullable App app;
+  private @Nullable Browser browser;
+  private @Nullable Device device;
+  private @Nullable OperatingSystem operatingSystem;
+  private @Nullable SentryRuntime runtime;
+  private @Nullable Gpu gpu;
+  private final @NotNull SpanContext trace;
+  private final @NotNull Map<String, Object> other = new ConcurrentHashMap<>();
+
+  public TransactionContexts(final @NotNull SpanContext trace) {
+    this.trace = trace;
+  }
+
+  public @Nullable App getApp() {
+    return app;
+  }
+
+  public void setApp(@Nullable App app) {
+    this.app = app;
+  }
+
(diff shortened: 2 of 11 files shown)


maciejwalkowiak:
@marandaneto there are few tests to add there, but before I put more time into it please take a look if you believe we should go this direction

sentry/src/main/java/io/sentry/SentryEvent.java
```diff
@@ -97,6 +97,9 @@
   /** List of breadcrumbs recorded before this event. */
   private List<Breadcrumb> breadcrumbs;
 
+  /** Contexts describing the environment (e.g. device, os or browser). */
+  private final @NotNull Contexts contexts = new Contexts();
```
marandaneto, line 101: in theory, this is a Nullable field.

maciejwalkowiak: This has been changed to not-nullable in previous PR.

sentry/src/main/java/io/sentry/adapters/TransactionContextsDeserializerAdapter.java
```diff
@@ -0,0 +1,122 @@
...
+  public TransactionContextsDeserializerAdapter(@NotNull final ILogger logger) {
+    this.logger = logger;
+  }
+
+  @Override
+  public TransactionContexts deserialize(
+      JsonElement json, Type typeOfT, JsonDeserializationContext context)
+      throws JsonParseException {
+    try {
+      if (json != null && !json.isJsonNull()) {
```
marandaneto, line 39: should we extract the content of this method to a Util class and reuse it in `ContextsDeserializerAdapter` and `ContextsSerializerAdapter`? they look pretty much the same.

sentry/src/main/java/io/sentry/adapters/TransactionContextsSerializerAdapter.java
```diff
@@ -0,0 +1,62 @@
...
+  private final @NotNull ILogger logger;
+
+  public TransactionContextsSerializerAdapter(@NotNull final ILogger logger) {
+    this.logger = logger;
+  }
+
+  @Override
+  public JsonElement serialize(
+      TransactionContexts src, Type typeOfSrc, JsonSerializationContext context) {
+    if (src == null) {
```
marandaneto, line 36: same as https://github.com/getsentry/sentry-java/pull/1164/files#r557351508

sentry/src/main/java/io/sentry/adapters/TransactionContextsSerializerAdapter.java
```diff
@@ -0,0 +1,62 @@
...
+      return null;
+    }
+
+    final JsonObject object = new JsonObject();
+    object.add(SpanContext.TYPE, context.serialize(src.getTrace(), Object.class));
```
marandaneto, line 41: `ContextsDeserializerAdapter` uses `SpanContext.TYPE` and apparently after this PR, it'd not be necessary anymore, right?

maciejwalkowiak: It will be, as SpanContext (aka "trace") is also a part of events.

sentry/src/main/java/io/sentry/adapters/TransactionContextsDeserializerAdapter.java
maciejwalkowiak: I'll see what i can do - they don't share the same class hierarchy, at least not at the moment.

maciejwalkowiak:
@marandaneto if we go this path, perhaps it makes sense to change also `Contexts` class to NOT inherit from hashmap.

sentry/src/main/java/io/sentry/TransactionContexts.java
```diff
@@ -1 +1,93 @@
...
+import io.sentry.protocol.Gpu;
+import io.sentry.protocol.OperatingSystem;
+import io.sentry.protocol.SentryRuntime;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public final class TransactionContexts {
```
marandaneto, line 15: while I agree that this would avoid all sort of NPE around`Contexts` and `trace`, looks like a junk of code duplication, wondering if just moving the `Contexts` inheritance away and adding an `other` field in there would not solve the problem for good (not adding a new class `TransactionContexts`).
Also, `Contexts` would need to become @NotNull by default on SentryBaseEvent.

maybe @bruno-garcia would like to chime in

maciejwalkowiak: Then we would eliminate in practical terms the possibility of NPE but we won't get away with warnings in the IDE.

maciejwalkowiak:
Not valid anymore.

PR #1165, opened 2021-01-14
Polish Performance API

maciejwalkowiak:
Polish Performance related API as a result of a call with @marandaneto @Tyrrrz.


sentry/src/main/java/io/sentry/Hub.java
```diff
@@ -616,16 +615,19 @@ public void flush(long timeoutMillis) {
   }
 
   @Override
-  public void setSpanContext(
-      final @NotNull Throwable throwable, final @NotNull SpanContext spanContext) {
+  public void setSpanContext(final @NotNull Throwable throwable, final @NotNull ISpan span) {
```
maciejwalkowiak, line 618: This method deserves a better name now. I am open for suggestions.

sentry/src/main/java/io/sentry/NoOpSpan.java
```diff
@@ -0,0 +1,50 @@
+package io.sentry;
+
+import io.sentry.protocol.SentryId;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public final class NoOpSpan implements ISpan {
```
marandaneto, line 7: every NoOp object today is a singleton with a `getInstance` method, what's about doing the same here and `NoOpTransaction`?

sentry/src/main/java/io/sentry/NoOpSpan.java
```diff
@@ -0,0 +1,50 @@
...
+  public void setThrowable(@Nullable Throwable throwable) {}
+
+  @Override
+  public @Nullable Throwable getThrowable() {
+    return null;
+  }
+
+  @Override
+  public @NotNull SpanContext getSpanContext() {
+    return new SpanContext();
```
marandaneto, line 45: better creating a `SpanContext` using `SentryId.EMPTY_ID` where possible, also `sampled=false` like you do with `SentryTraceHeader`

sentry/src/main/java/io/sentry/NoOpTransaction.java
```diff
@@ -107,7 +94,7 @@ public void setThrowable(@Nullable Throwable throwable) {}
 
   @Override
   public @NotNull SpanContext getSpanContext() {
-    return startChild();
+    return new SpanContext();
```
marandaneto, line 97: same with https://github.com/getsentry/sentry-java/pull/1165/files#r557497834

sentry/src/main/java/io/sentry/SentryTransaction.java
```diff
@@ -90,7 +90,7 @@ public void setName(final @NotNull String name) {
    */
   @Override
   public @NotNull Span startChild(
```
marandaneto, line 92: should the signatures be `ISpan` now?

sentry/src/main/java/io/sentry/Hub.java
marandaneto: naming is hard, off by 1 errors too :)

should it be `@ApiStatus.Internal`?

sentry/src/main/java/io/sentry/NoOpSpan.java
maciejwalkowiak: 👍

sentry/src/main/java/io/sentry/Hub.java
maciejwalkowiak: It is actually on the interface, I'll add it to implementation too. 👍

sentry/src/main/java/io/sentry/NoOpSpan.java
maciejwalkowiak: 👍

sentry/src/main/java/io/sentry/SentryTransaction.java
maciejwalkowiak: 👍

sentry/src/main/java/io/sentry/Hub.java
```diff
@@ -566,20 +565,20 @@ public void flush(long timeoutMillis) {
...
           .log(
               SentryLevel.WARNING,
               "Instance is disabled and this 'startTransaction' returns a no-op.");
-      transaction = new NoOpTransaction();
+      transaction = NoOpTransaction.getInstance();
     } else {
+      final SamplingContext samplingContext =
+          new SamplingContext(transactionContexts, customSamplingContext);
+      boolean samplingDecision = tracingSampler.sample(samplingContext);
+      transactionContexts.setSampled(samplingDecision);
```
bruno-garcia, line 580: M: Could we even take this decision in the ctor now and hide this setter? THat's until we decide on the right approach to have deferred sampling?

sentry/src/main/java/io/sentry/NoOpSpan.java
```diff
@@ -0,0 +1,59 @@
...
+    return instance;
+  }
+
+  @Override
+  public @NotNull ISpan startChild() {
+    return new NoOpSpan();
+  }
+
+  @Override
+  public @NotNull ISpan startChild(@Nullable String operation, @Nullable String description) {
```
bruno-garcia, line 23: should `operation` but NonNullable and set to `noop` instead?

bruno-garcia (approved):

sentry/src/main/java/io/sentry/Hub.java
maciejwalkowiak: What would be the benefit of that? Too many options in constructor means many constructors.

sentry/src/main/java/io/sentry/NoOpSpan.java
maciejwalkowiak: Has it been documented somewhere? "op" is not a required field (https://develop.sentry.dev/sdk/event-payloads/transaction/)

Tyrrrz: Apparently it's required:

image
https://discord.com/channels/621778831602221064/740915087220408350/800720296432500746

maciejwalkowiak: image
@marandaneto please clarify this.

Tyrrrz: cc @rhcarvalho

sentry/src/main/java/io/sentry/Hub.java
marandaneto: See https://github.com/getsentry/develop/issues/243

sentry/src/main/java/io/sentry/NoOpSpan.java
marandaneto: See https://github.com/getsentry/sentry-java/issues/1170

maciejwalkowiak: 👍

sentry/src/main/java/io/sentry/Hub.java
bruno-garcia: The problem is that changing sampling mid-flight means you send out some sentry-trace with one decision then others with another which will result in broken traces.

Tyrrrz: @bruno-garcia this is apparently allowed by design: https://github.com/getsentry/develop/issues/243#issuecomment-760971206

> What you're asking for is not possible with head-based sampling.
You can, however, overwrite the sampling decision for the current transaction at any point before it goes to Sentry (that is, you can drop the transaction), but you cannot expect that it will affect peer services, because at that point sampling decisions were already propagated via HTTP headers.

If I understand correctly, it's possible to sample out a transaction on creation and also at any point before it's sent to Sentry. In case of the latter, it will not affect peer services.

bruno-garcia: Once we start receiving reports of broken traces, it'll be a nightmare to debug why. We'll play a blame game where we'll say the user must be changing sampling, and users will say we'll no propagating properly, or dropping transactions on the Transport/Relay/else.

@rhcarvalho Am I overreacting here?

sentry/src/main/java/io/sentry/NoOpSpan.java
```diff
@@ -0,0 +1,63 @@
...
+
+  private NoOpSpan() {}
+
+  public static NoOpSpan getInstance() {
+    return instance;
+  }
+
+  @Override
+  public @NotNull ISpan startChild(final @NotNull String operation) {
+    return new NoOpSpan();
```
marandaneto, line 19: this one and other methods could return `instance` directly, right

sentry/src/main/java/io/sentry/SentryTransaction.java
```diff
@@ -120,11 +120,11 @@ public void setName(final @NotNull String name) {
...
    * @return a new transaction span
    */
-  @Override
-  public @NotNull Span startChild(
+  @NotNull
+  Span startChild(
       final @NotNull SpanId parentSpanId,
-      final @NotNull String operation,
-      final @NotNull String description) {
+      final @Nullable String operation,
```
marandaneto, line 126:
```suggestion
      final @NotNull String operation,
```

sentry/src/main/java/io/sentry/SentryTransaction.java
```diff
@@ -104,8 +104,8 @@ public void setName(final @NotNull String name) {
    * @param parentSpanId - parent span id
    * @return a new transaction span
    */
-  @Override
-  public @NotNull Span startChild(final @NotNull SpanId parentSpanId) {
+  @NotNull
+  Span startChild(final @NotNull SpanId parentSpanId) {
```
marandaneto, line 108: should all of them return `ISpan` instead?

marandaneto:
to not forget: https://github.com/getsentry/sentry-docs/issues/2912

sentry-spring/src/main/java/io/sentry/spring/tracing/SentrySpanClientHttpRequestInterceptor.java
```diff
@@ -41,10 +41,9 @@ public SentrySpanClientHttpRequestInterceptor(final @NotNull IHub hub) {
       return execution.execute(request, body);
     }
 
-    final ISpan span = activeSpan.startChild();
+    final ISpan span = activeSpan.startChild("http");
```
marandaneto, line 44: not really related to this PR but: https://github.com/getsentry/develop/issues/242
shouldn't it be `http.server` then?

sentry-spring/src/main/java/io/sentry/spring/tracing/SentrySpan.java
```diff
@@ -19,21 +19,21 @@
...
   String operation() default "";
 
   /**
    * Span description.
    *
    * @return description.
    */
-  @AliasFor("description")
+  @AliasFor("operation")
   String value() default "";
```
marandaneto, line 38: not sure if this is right, `AliasFor`= operation, javadocs is `description` and name of the param is `value`.
I'd double check this class and its methods, aliases and javadocs

sentry/src/main/java/io/sentry/Hub.java
rhcarvalho: > it's possible to sample out a transaction on creation and also at any point before it's sent to Sentry

@Tyrrrz 
While it is _technically possible_, I believe that's not how it was envisioned to be used.
Sampling should be done "behind the scenes" when starting the transaction. In Python, JS, etc, there's no stopping anyone from mutating an object later on, though.

The proper way to influence sampling is with an explicit traces sampler.

After some back-and-forth with the Go implementation, I let go of making everything "super protected" to stay inline with how the error reporting part of the SDK works, it is rather liberal.

I'll let ya'll think what design makes more sense for Java.

> Am I overreacting here?

@bruno
Maybe :P

sentry/src/main/java/io/sentry/Hub.java
```diff
@@ -566,20 +565,20 @@ public void flush(long timeoutMillis) {
...
       options
           .getLogger()
           .log(
               SentryLevel.WARNING,
               "Instance is disabled and this 'startTransaction' returns a no-op.");
-      transaction = new NoOpTransaction();
+      transaction = NoOpTransaction.getInstance();
     } else {
+      final SamplingContext samplingContext =
+          new SamplingContext(transactionContexts, customSamplingContext);
```
rhcarvalho, line 578: nit: transactionContexts -> transactionContext

sentry/src/main/java/io/sentry/Hub.java
```diff
@@ -566,20 +565,20 @@ public void flush(long timeoutMillis) {
...
+      boolean samplingDecision = tracingSampler.sample(samplingContext);
```
rhcarvalho, line 579: nit: tracingSampler -> tracesSampler (https://develop.sentry.dev/sdk/performance/#sdk-configuration)

sentry/src/main/java/io/sentry/NoOpSpan.java
bruno-garcia: Yeah the goal is to reduce the instrumentation cost so makes sense to return the singleton. Otherwise we'd be better off adding a new Span so we can have an `id` laying around at least.

sentry/src/main/java/io/sentry/SentryTransaction.java
maciejwalkowiak: Not this one but it can be private.

sentry-spring/src/main/java/io/sentry/spring/tracing/SentrySpan.java
maciejwalkowiak: Fixed.

maciejwalkowiak:
@bruno-garcia @marandaneto all polishing applied, please have another look.

sentry/src/main/java/io/sentry/SentryTransaction.java
```diff
@@ -104,8 +104,8 @@ public void setName(final @NotNull String name) {
    * @param parentSpanId - parent span id
    * @return a new transaction span
    */
-  @Override
-  public @NotNull Span startChild(final @NotNull SpanId parentSpanId) {
+  @NotNull
+  private Span startChild(final @NotNull SpanId parentSpanId) {
```
marandaneto, line 108:
```suggestion
  private ISpan startChild(final @NotNull SpanId parentSpanId) {
```

sentry/src/main/java/io/sentry/SentryTransaction.java
```diff
@@ -120,11 +120,11 @@ public void setName(final @NotNull String name) {
    * @param description - span description
    * @return a new transaction span
    */
-  @Override
-  public @NotNull Span startChild(
+  @NotNull
+  Span startChild(
```
marandaneto, line 124:
```suggestion
  ISpan startChild(
```

sentry/src/main/java/io/sentry/Span.java
```diff
@@ -44,12 +44,13 @@
   }
 
   @Override
-  public @NotNull Span startChild() {
-    return transaction.startChild(super.getSpanId());
+  public @NotNull Span startChild(final @NotNull String operation) {
```
marandaneto, line 47:
```suggestion
  public @NotNull ISpan startChild(final @NotNull String operation) {
```

sentry/src/main/java/io/sentry/Span.java
```diff
@@ -44,12 +44,13 @@
...
+    return this.startChild(operation, null);
   }
 
   @Override
-  public Span startChild(final @NotNull String operation, final @NotNull String description) {
+  public @NotNull Span startChild(
```
marandaneto, line 52:
```suggestion
  public @NotNull ISpan startChild(
```

sentry/src/main/java/io/sentry/Span.java
```diff
@@ -17,7 +17,7 @@
    * A transaction this span is attached to. Marked as transient to be ignored during JSON
    * serialization.
    */
-  private final transient @NotNull ITransaction transaction;
+  private final transient @NotNull SentryTransaction transaction;
```
marandaneto, line 20: why the change, just out of curiosity?

sentry-spring/src/main/java/io/sentry/spring/tracing/SentrySpanClientHttpRequestInterceptor.java
marandaneto: makes sense, should @Tyrrrz align outgoing requests?

sentry/src/main/java/io/sentry/Span.java
maciejwalkowiak: Because a method to start span that takes parent span is not anymore there on the `ITransaction` as we don't want to pollute public API with it. `SentryTransaction` has it package private.

sentry-spring/src/main/java/io/sentry/spring/tracing/SentrySpanClientHttpRequestInterceptor.java
rhcarvalho: FWIW Python uses `"http"` https://github.com/getsentry/sentry-python/blob/ca828c9486f4a0030a744849a4e8e10c5a1c7aeb/sentry_sdk/integrations/stdlib.py#L81

bruno-garcia: `http.client` and `http.server` make sense to me. Is this something we can change Python into alignment or we just deviate here? /cc @untitaker and @brustolin for visibility

Tyrrrz:
image

I thought this PR was already Polish 🇵🇱 :trollface:

bruno-garcia:
> image
> 
> I thought this PR was already Polish 🇵🇱 :trollface:

Never gets old :D 🇵🇱 🇲🇨 🇵🇱 🇲🇨 🇵🇱 🇲🇨 🇵🇱 🇲🇨

sentry-spring/src/main/java/io/sentry/spring/tracing/SentrySpanClientHttpRequestInterceptor.java
marandaneto: I'd also prefer `http.client` as `http` is just ambiguous.
I will create an issue for that so we can move ahead with this PR

marandaneto: https://github.com/getsentry/develop/issues/247

sentry/src/main/java/io/sentry/NoOpTransaction.java
```diff
@@ -10,31 +10,26 @@
...
 
   @Override
-  public @NotNull Span startChild(
-      final @NotNull SpanId parentSpanId,
-      final @NotNull String operation,
-      final @NotNull String description) {
-    return startChild();
+  public @NotNull ISpan startChild(
+      final @NotNull String operation, final @Nullable String description) {
+    return startChild(operation);
```
marandaneto, line 32: just a nit: you could return `NoOpSpan.getInstance();` directly so you spare a method call

marandaneto (approved):
LGTM

PR #1171, opened 2021-01-15
Support SENTRY_TRACES_SAMPLE_RATE conf. via env variables.

### LINKED ISSUE #1169: Support SENTRY_TRACES_SAMPLE_RATE conf. via env. variables
One would be able to set traces sample rate via env. variables

@maciejwalkowiak all we need is to read/set the env. var. here: https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/SentryOptions.java#L263-L293
is that right? `propertiesProvider` already does the job of reading from env. variables

maciejwalkowiak:
## :scroll: Description

Support SENTRY_TRACES_SAMPLE_RATE conf. via env variables.

Fixes #1169

Unit tests.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/sentry/src/main/java/io/sentry/config/PropertiesProvider.java
+++ b/sentry/src/main/java/io/sentry/config/PropertiesProvider.java
@@ -46,7 +46,7 @@ public interface PropertiesProvider {
    * @return property value or the default value if not found.
    */
   @NotNull
-  default String getProperty(@NotNull String property, @NotNull String defaultValue) {
+  default String getProperty(final @NotNull String property, final @NotNull String defaultValue) {
     final String result = getProperty(property);
     return result != null ? result : defaultValue;
   }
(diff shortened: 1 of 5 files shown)


sentry/src/main/java/io/sentry/config/PropertiesProvider.java
```diff
@@ -58,8 +58,20 @@ default String getProperty(@NotNull String property, @NotNull String defaultValu
...
+  /**
+   * Resolves a {@link Double} property given by it's name.
+   *
+   * @param property - the property name
+   * @return property value or the default value if not found.
+   */
+  @Nullable
+  default Double getDoubleProperty(final @NotNull String property) {
+    final String result = getProperty(property);
+    return result != null ? Double.valueOf(result) : null;
```
marandaneto, line 75: should we account if it's not a Double? like catch `NumberFormatException` and return null?

marandaneto (approved):

sentry/src/main/java/io/sentry/config/PropertiesProvider.java
bruno-garcia: Given this is done at startup and the value must be double, probably OK to blow up here

bruno-garcia (approved):

sentry/src/main/java/io/sentry/config/PropertiesProvider.java
marandaneto: that's o but kinda goes against our philosophy

https://develop.sentry.dev/sdk/philosophy/#degrade-gracefully

I'd prefer blowing up only if there's no DSN, any other reason would be a bug IMO

PR #1173, opened 2021-01-15
Set SpanContext on SentryTransaction to avoid potential NPE.

maciejwalkowiak:
## :scroll: Description

Set trace context on SentryTransaction and pass it to `contexts` on `transaction#finish`.

The alternative to https://github.com/getsentry/sentry-java/pull/1164

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


maciejwalkowiak:
@bruno-garcia @marandaneto I think this is a better than #1164 approach - the result the discussion we had. I think trace context on `finish` is cleaner than doing it during serialization.

bruno-garcia (approved):
Much nicer!

sentry/src/main/java/io/sentry/SentryTransaction.java
```diff
@@ -57,7 +59,7 @@ public SentryTransaction(
     this.transaction = Objects.requireNonNull(name, "name is required");
     this.startTimestamp = DateUtils.getCurrentDateTime();
     this.hub = Objects.requireNonNull(hub, "hub is required");
-    this.getContexts().setTrace(contexts);
+    this.context = contexts;
```
marandaneto, line 62: do the assignment with `Objects.requireNonNull(contexts, "contexts are required");`

sentry/src/main/java/io/sentry/SentryTransaction.java
```diff
@@ -28,7 +28,9 @@
    * A hub this transaction is attached to. Marked as transient to be ignored during JSON
    * serialization.
    */
-  private @NotNull final transient IHub hub;
+  private final transient @NotNull IHub hub;
+
+  private final transient @NotNull SpanContext context;
```
marandaneto, line 33:
```suggestion
  private final transient @NotNull SpanContext contexts;
```

sentry/src/main/java/io/sentry/SentryTransaction.java
```diff
@@ -239,4 +242,8 @@ Date getTimestamp() {
     }
     return null;
   }
+
+  SpanContext getContext() {
```
marandaneto, line 246:
```suggestion
  @NotNull SpanContext getContext() {
```

sentry/src/main/java/io/sentry/SentryTransaction.java
bruno-garcia: Wasn't the goal to move the name to `context`? Since `SpanContext`?

marandaneto: no goal, but I'd rather prefer the keep consistency, either the ctor param or the global field should rename, so they match.
it's a nitpick but why not naming the same?

sentry/src/main/java/io/sentry/SentryTransaction.java
```diff
@@ -156,6 +158,7 @@ public void finish() {
     if (this.throwable != null) {
       hub.setSpanContext(this.throwable, this.getSpanContext());
     }
+    this.getContexts().setTrace(this.context);
```
marandaneto, line 161: here `this.getContexts()` is guaranteed to not be null, right?

maciejwalkowiak: Yes.

sentry/src/main/java/io/sentry/SentryTransaction.java
maciejwalkowiak: renamed to "context"

marandaneto (approved):

PR #1183, opened 2021-01-18
Bring back support for setting transaction name without ongoing transaction

maciejwalkowiak:

Discussion in: https://github.com/getsentry/develop/issues/246

Even if performance event is not used, users still must be able to set the transaction name on the scope in order to apply it on the events.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/sentry/src/main/java/io/sentry/Scope.java
+++ b/sentry/src/main/java/io/sentry/Scope.java
@@ -109,9 +109,8 @@ public final class Scope implements Cloneable {
     final ITransaction tx = this.transaction;
     if (tx != null) {
       tx.setName(transaction);
-    } else {
-      this.transactionName = transaction;
     }
+    this.transactionName = transaction;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


sentry/src/main/java/io/sentry/Scope.java
```diff
@@ -106,6 +109,8 @@ public void setTransaction(final @NotNull String transaction) {
     final ITransaction tx = this.transaction;
     if (tx != null) {
       tx.setName(transaction);
```
Tyrrrz, line 111: I would probably do `this.transactionName = transaction;` in this case too, just in case. I know this probably cannot happen, but if `tx` becomes null at some point later, this makes sure the name doesn't revert to the previous one.

maciejwalkowiak: 👍

marandaneto (approved):
LGTM

PR #1186, opened 2021-01-19
Set "debug" through external properties.

maciejwalkowiak:
## :scroll: Description

Allows setting "debug" through external properties.

Useful when debugging Spring (non-boot) integration.

Unit tests.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/sentry/src/main/java/io/sentry/SentryOptions.java
+++ b/sentry/src/main/java/io/sentry/SentryOptions.java
@@ -371,8 +371,7 @@ public class SentryOptions {
    *
    * @return true if ON or false otherwise
    */
-  @Nullable
-  Boolean getDebug() {
+  private @Nullable Boolean getDebug() {
     return debug;
   }
 
--- a/sentry/src/main/java/io/sentry/SentryOptions.java
+++ b/sentry/src/main/java/io/sentry/SentryOptions.java
@@ -371,7 +371,8 @@ public class SentryOptions {
    *
    * @return true if ON or false otherwise
    */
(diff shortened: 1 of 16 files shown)


sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -353,18 +354,27 @@ public void setDsn(@Nullable String dsn) {
...
+  public void setDebug(final @Nullable Boolean debug) {
     this.debug = debug;
   }
 
+  /**
+   * Check if debug mode is ON, OFF or not set.
+   *
+   * @return true if ON or false otherwise
+   */
+  public @Nullable Boolean getDebug() {
```
marandaneto, line 374: can we make it either private or package-private? `isDebug()` is all we need.

maciejwalkowiak: Package private 👍

maciejwalkowiak: .. or actually private would be better.

bruno-garcia (commented):
Couldn't we get away with `boolean` and default to `false`? You only need to set the value if it's `true` anyway (since we always have it false)

maciejwalkowiak:
@bruno-garcia practically yes, but theoretically there can be a case when it is set to `true` in the code and then turned off in external configuration.

bruno-garcia (approved):

marandaneto (approved):

Issue #1185, opened 2021-01-19
Add addToTransactions to Attachment


philipphofmann:
The SentryClient currently adds attachments to transactions
https://github.com/getsentry/sentry-java/blob/2f8360a001ffe1dff08a374b5ec91cecd34aed12/sentry/src/main/java/io/sentry/SentryClient.java#L359-L360
This can add a huge overhead when putting large attachments to the scope. In order to avoid this, we can add `addToTransactions` to attachments and set the default to `false`. The Python SDK already has [this flag](https://github.com/getsentry/sentry-python/blob/2af3274de22ee00b5254cc6700cc26ddc06dbb66/sentry_sdk/attachments.py#L18).

bruno-garcia:
@philipphofmann we plan to release 4.0 next week, is this going to make into it?

philipphofmann:
@bruno-garcia I can take care of it.

bruno-garcia:
Thanks @philipphofmann !

PR #1191, opened 2021-01-21
feat: Add addToTransactions to Attachment

philipphofmann:
## :scroll: Description
Add extra parameter addToTransaction to attachment, which specifies if the SDK adds
it to transactions or not. The default is false.

Fixes GH-1185

Unit tests.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/sentry/src/main/java/io/sentry/Attachment.java
+++ b/sentry/src/main/java/io/sentry/Attachment.java
@@ -13,6 +13,7 @@ public final class Attachment {
   private @Nullable String pathname;
   private final @NotNull String filename;
   private final @NotNull String contentType;
+  private final boolean addToTransactions;
 
   /**
    * We could use Files.probeContentType(path) to determine the content type of the filename. This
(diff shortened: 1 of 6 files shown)


sentry/src/main/java/io/sentry/SentryClient.java
```diff
@@ -377,6 +377,21 @@ public void captureSession(final @NotNull Session session, final @Nullable Objec
     return sentryId;
   }
 
+  private List<Attachment> filterForTransaction(List<Attachment> attachments) {
```
marandaneto, line 380: M: for nonprimitive types, lets always add `final @NotNull` or `final @Nullable`
a linter is on its way if we work together on https://github.com/getsentry/sentry-java/pull/1137/files#r548151954 :)

sentry/src/main/java/io/sentry/Attachment.java
```diff
@@ -131,4 +177,14 @@ public Attachment(
...
     return contentType;
   }
+
+  /**
+   * Returns <code>true</code> if the SDK should add this attachment to every {@link ITransaction}
+   * and <code>false</code> if it shouldn't. Default is <code>false</code>.
+   *
+   * @return <code>true</code> if attachment should be added to every {@link ITransaction}.
+   */
+  public boolean isAddToTransactions() {
```
marandaneto, line 187: L: should this be package private? not sure if its useful as public API

marandaneto (approved):
2 minor comments but other than that LGTM

PR #1198, opened 2021-01-21
Copy options tags to transactions.

### LINKED ISSUE #1163: Transactions are missing configured tags
_Platform:_
- [ ]  Android -> If yes, which Device API (and compileSdkVersion/targetSdkVersion/Build tools) version?
- [x]  Java -> If yes, which Java (and sourceCompatibility/targetCompatibility) version? JDK 11
- [ ]  Kotlin -> If yes, which Kotlin (and jvmTarget) version?
- [ ]  NDK -> If yes, which NDK/CMake version?
- [ ]  React-Native -> If yes, which version?
- [ ]  Timber -> If yes, which version?
- [ ]  Log4j2 -> If yes, which version?
- [ ]  Logback -> If yes, which version?
- [ ]  Spring -> If yes, which version?

_IDE:_
- [ ]  Android Studio -> If yes, which version?
- [ ]  IntelliJ -> If yes, which version?
- [ ]  Other -> If yes, which one?

_Build system:_
- [ ]  Gradle -> If yes, which version?
- [ ]  Buck -> If yes, which version?
- [ ]  Bazel -> If yes, which version?
- [ ]  Maven -> If yes, which version?
- [ ]  Other -> If yes, which one?

_Android Gradle Plugin:_
- [ ]  Yes -> If yes, which version?
- [ ]  No

- [ ]  Yes -> If yes, which version?
- [ ]  No

_Proguard/R8:_
- [ ]  Enabled
- [ ]  Disabled

_Platform installed with:_
- [ ] JCenter
- [ ] Bintray
- [ ] Maven Central
- [ ] Manually

The version of the SDK:
**4.0.0-alpha.2**

---
I have the following issue:

I've noticed that all transactions are missing tags, including the environment and release. These are configured using `SentryOptions` that we pass in to `Sentry.init()`. These tags appear on all of our Issues.

It also seems like the `SentryBeforeSendCallback` we configure is not called, so it is not possible to work around the issue.

We are using the logback and Spring integrations but I am able to reproduce even with just manual calls to send a transaction.

Inspecting the code and setting a breakpoint with a debugger, I can see that `SentryClient.captureTransaction()` has the correctly configured scope, but the transaction has no tags.

https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/SentryClient.java#L343-L378 

**Steps to reproduce:**

Initialize Sentry:
```
SentryOptions options = SentryOptions.from(PropertiesProviderFactory.create());
options.setDsn("your-dsn-here");
options.setEnvironment("test");
options.setRelease("1234");
options.setTracesSampleRate(1.0);
options.setTag("tag", "bag");

Sentry.init(options);
```

Send a transaction:
```
final SentryTransaction transaction = Sentry.startTransaction("test transaction");
final ISpan span = Sentry.getSpan();
final Span innerSpan = span.startChild("test", "a test span");
innerSpan.finish();
transaction.finish();
```

**Actual result:**
- Transaction is visible in the UI but it has none of the expected tags (environment, release, "tag").

**Expected result:**
- Transaction should have the same tags we would expect to see on Issues as configured via SentryOptions.
- SentryBeforeSendCallback should be invoked, as SentryTransaction is another SentryEvent

maciejwalkowiak:
## :scroll: Description

Copy options tags to transactions.

Fixes #1163

## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next steps

Perhaps tags from the scope (and potentially other properties from the scope) should be applied on transactions too?

### DIFF
--- a/sentry/src/main/java/io/sentry/SentryClient.java
+++ b/sentry/src/main/java/io/sentry/SentryClient.java
@@ -400,6 +400,15 @@ public final class SentryClient implements ISentryClient {
     if (transaction.getEnvironment() == null) {
       transaction.setEnvironment(options.getEnvironment());
     }
+    if (transaction.getTags() == null) {
+      transaction.setTags(new HashMap<>(options.getTags()));
+    } else {
+      for (Map.Entry<String, String> item : options.getTags().entrySet()) {
+        if (!transaction.getTags().containsKey(item.getKey())) {
+          transaction.setTag(item.getKey(), item.getValue());
+        }
+      }
+    }
     return transaction;
   }
 
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -12,9 +12,8 @@
 
 ## :pencil: Checklist
 <!--- Put an `x` in the boxes that apply -->
-- [ ] I reviewed submitted code
-- [ ] I added tests to verify changes
-- [ ] All tests passing
+- [ ] I reviewed the submitted code
+- [ ] I added tests to verify the changes
 - [ ] No breaking changes
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,9 +12,11 @@
 * Enhancement: Polish Performance API (#1165)
 * Enhancement: Set "debug" through external properties (#1186)
 * Enhancement: Simplify Spring integration (#1188)
+* Enhancement: Init overload with dsn (#1195)
 * Enhancement: Enable Kotlin map-like access on CustomSamplingContext (#1192)
 * Enhancement: Auto register custom ITransportFactory in Spring integration (#1194)
 * Enhancement: Improve Kotlin property access in Performance API (#1193)
+* Enhancement: Add convenient method for accessing event's throwable (1202)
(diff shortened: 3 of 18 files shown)


bruno-garcia (approved):

sentry/src/main/java/io/sentry/SentryClient.java
```diff
@@ -400,6 +400,15 @@ public void captureSession(final @NotNull Session session, final @Nullable Objec
...
     }
+    if (transaction.getTags() == null) {
+      transaction.setTags(new HashMap<>(options.getTags()));
+    } else {
+      for (Map.Entry<String, String> item : options.getTags().entrySet()) {
+        if (!transaction.getTags().containsKey(item.getKey())) {
+          transaction.setTag(item.getKey(), item.getValue());
+        }
+      }
+    }
```
marandaneto, line 411: what's about the tags set on the scope? should we also copy them?

cc @bruno-garcia

bruno-garcia: oof, yeah we're missing that. All tags should be added

PR #1209, opened 2021-01-26
Log when capturing unfinished transaction

### LINKED ISSUE #1208: Check if transaction is finished during captureTransaction call
If the transaction is captured but not finished yet, what to do?
I'd say let's log it as INFO, finish and capture it.

Ideally `SentryClient.captureTransaction` does that as every call forwards to this method.

Context: https://github.com/getsentry/sentry-dotnet/pull/758#discussion_r564125517

maciejwalkowiak:
## :scroll: Description

Log warning when capturing unfinished transaction.

Fixes #1208

Unit tests.

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/sentry/src/main/java/io/sentry/ITransaction.java
+++ b/sentry/src/main/java/io/sentry/ITransaction.java
@@ -75,4 +75,11 @@ public interface ITransaction extends ISpan {
   @Nullable
   @ApiStatus.Internal
   String getTransaction();
+
+  /**
+   * Returns if transaction has finished.
+   *
+   * @return if transaction has finished.
+   */
+  boolean isFinished();
 }
--- a/sentry/src/main/java/io/sentry/ITransaction.java
+++ b/sentry/src/main/java/io/sentry/ITransaction.java
@@ -75,11 +75,4 @@ public interface ITransaction extends ISpan {
   @Nullable
   @ApiStatus.Internal
   String getTransaction();
-
-  /**
-   * Returns if transaction has finished.
-   *
-   * @return if transaction has finished.
-   */
-  boolean isFinished();
 }
--- a/sentry/src/main/java/io/sentry/SentryClient.java
+++ b/sentry/src/main/java/io/sentry/SentryClient.java
@@ -353,22 +353,29 @@ public final class SentryClient implements ISentryClient {
     SentryId sentryId = transaction.getEventId();
 
     if (transaction instanceof SentryTransaction) {
-      final SentryTransaction sentryTransaction =
-          processTransaction((SentryTransaction) transaction);
-      try {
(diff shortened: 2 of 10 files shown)


sentry/src/main/java/io/sentry/SentryClient.java
```diff
@@ -353,22 +353,29 @@ public void captureSession(final @NotNull Session session, final @Nullable Objec
...
-            buildEnvelope(sentryTransaction, filterForTransaction(getAttachmentsFromScope(scope)));
-        if (envelope != null) {
-          transport.send(envelope, hint);
-        } else {
+      final SentryTransaction sentryTransaction = (SentryTransaction) transaction;
+
+      if (!sentryTransaction.isFinished()) {
+        options.getLogger().log(SentryLevel.INFO, "Finishing unfinished transaction %s.", sentryId);
+        sentryTransaction.finish();
+      } else {
```
marandaneto, line 361: is that right? wouldn't it be better to `finish` AND capture the transaction? or am I missing something?

marandaneto: ahhhh `finish` calls `captureTransaction` already, guess that's the trick, right?

marandaneto:
@maciejwalkowiak changelog

bruno-garcia (changes requested):
I thought we had just agreed not to do this

bruno-garcia:
The rationale against this change is that it'll be hiding a potential programming error. A transaction is expected to be finished through `finish()` (of if captured explicitly, have its close timestamp be set explicitly via an API) so we can avoid odd bugs where transactions look normal but have been captured before they were properly finished.

marandaneto:
> I thought we had just agreed not to do this

> commented 4 hours ago

maciejwalkowiak:
PR changed to include just logging a warning. I think we don't need changelog here.

sentry/src/main/java/io/sentry/ITransaction.java
```diff
@@ -75,4 +75,11 @@
   @Nullable
   @ApiStatus.Internal
   String getTransaction();
+
+  /**
+   * Returns if transaction has finished.
+   *
+   * @return if transaction has finished.
+   */
+  boolean isFinished();
```
marandaneto, line 84: should this go to `ISpan` instead?

maciejwalkowiak: Thats true, fixed.

marandaneto (approved):

bruno-garcia:
> > I thought we had just agreed not to do this
> 
> commented 4 hours ago :) so even before the discussion on discord

Right, sorry about that

PR #1211, opened 2021-01-27
Support logging via JUL

### LINKED ISSUE #1037: Support for java.util.logging (jul)
_Platform:_
- [ ]  Android -> If yes, which Device API (and compileSdkVersion/targetSdkVersion/Build tools) version?
- [x ]  Java -> If yes, which Java 11
- [ ]  Kotlin -> If yes, which Kotlin (and jvmTarget) version?
- [ ]  NDK -> If yes, which NDK/CMake version?
- [ ]  React-Native -> If yes, which version?
- [ ]  Timber -> If yes, which version?
- [x ]  Log4j2 -> If yes, which version? 3.13
- [ ]  Logback -> If yes, which version?
- [ ]  Spring -> If yes, which version?

_IDE:_
- [ ]  Android Studio -> If yes, which version?
- [x ]  IntelliJ -> If yes, which version? 2020.2
- [ ]  Other -> If yes, which one?

_Build system:_
- [ ]  Gradle -> If yes, which version?
- [ ]  Buck -> If yes, which version?
- [ ]  Bazel -> If yes, which version?
- [x ]  Maven -> If yes, which version?
- [ ]  Other -> If yes, which one?

_Android Gradle Plugin:_
- [ ]  Yes -> If yes, which version?
- [ x]  No

- [ ]  Yes -> If yes, which version?
- [x ]  No

_Proguard/R8:_
- [ ]  Enabled
- [x ]  Disabled

_Platform installed with:_
- [ ] JCenter
- [ ] Bintray
- [x ] Maven Central
- [ ] Manually

The version of the SDK:
**3.x.x**

---
I have the following issue:
The legacyAPI contained a java.util.logging Logger Handler named io.sentry.jul.SentryHander. This class has been dropped for API 3.x.x Is there a specific reasoning for this? 
The lack of such a handler makes the integration of Sentry into wildfly much more difficult (if you want Sentry to be active during startup/deployment) because the wildfly container only allows custom handlers that extend java.util.logging.Handler ([ref](https://docs.wildfly.org/19/wildscribe/subsystem/logging/custom-handler/index.html))

**Steps to reproduce:**
Checkout the new java-sentry api

**Actual result:**
There is no jul handler.

**Expected result:**
A handler extending java.util.logger.Handler should exist

maciejwalkowiak:
## :scroll: Description

Support logging via JUL + sample project.

Fixes #1037

Unit & Integration tests.

## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next steps

Create a PR in Quarkus or let know Quarkus team about JUL logger in Sentry SDK, so that they do not need to maintain their custom one.


maciejwalkowiak:
Afterthought - for the backward compatibility we should also support MDC tags, which means we need a dependency for SLF4J, which means that JUL appender would have to go to separate module.

bruno-garcia:
> Afterthought - for the backward compatibility we should also support MDC tags, which means we need a dependency for SLF4J, which means that JUL appender would have to go to separate module.

Sorry but I don't follow. before it wasnt' a separate package, so why does it need to be now for backwards compat?

marandaneto:
> > Afterthought - for the backward compatibility we should also support MDC tags, which means we need a dependency for SLF4J, which means that JUL appender would have to go to separate module.
> 
> Sorry but I don't follow. before it wasnt' a separate package, so why does it need to be now for backwards compat?

adding a dependency of `SLF4J` I believe.

maciejwalkowiak:
As @marandaneto said, since 3.x we do not have slf4j dependency in the SDK core. I am not really sure how people use slf4j AND jul at the same time, but since it was in 1.x and it's also in handler added to Quarkus, I think we should follow the same path.

bruno-garcia:
makes sense

bruno-garcia (approved):

sentry-jul/src/main/java/io/sentry/jul/SentryHandler.java
```diff
@@ -0,0 +1,283 @@
...
+    // Do not log the event if the current thread is managed by sentry
+    if (!isLoggable(record)) {
+      return;
+    }
+    try {
+      if (record.getLevel().intValue() >= minimumEventLevel.intValue()) {
+        Sentry.captureEvent(createEvent(record));
+      }
+      if (record.getLevel().intValue() >= minimumBreadcrumbLevel.intValue()) {
+        Sentry.addBreadcrumb(createBreadcrumb(record));
```
marandaneto, line 71: can we use the `HubAdapter` as a package-private ctor so we can assert this via the mocked hub?

sentry-jul/src/main/java/io/sentry/jul/SentryHandler.java
```diff
@@ -0,0 +1,283 @@
...
+   * Creates {@link SentryEvent} from JUL's {@link LogRecord}.
+   *
+   * @param record the log record
+   * @return the sentry event
+   */
+  // for the Android compatibility we must use old Java Date class
+  @SuppressWarnings("JdkObsolete")
+  @NotNull
+  SentryEvent createEvent(final @NotNull LogRecord record) {
+    final SentryEvent event = new SentryEvent(new Date(record.getMillis()));
```
marandaneto, line 152: `record.getMillis()` is it UTC or local time?
otherwise we need to use `DateUtils.getDateTime(millis)`

sentry-jul/src/main/java/io/sentry/jul/SentryHandler.java
```diff
@@ -0,0 +1,283 @@
...
+  }
+
+  /**
+   * Returns formatted Event message when provided the message template and parameters.
+   *
+   * @param message Message template body.
+   * @param parameters Array of parameters for the message.
+   * @return Formatted message.
+   */
+  private String formatMessage(final @NotNull String message, final @Nullable Object[] parameters) {
```
marandaneto, line 206:
```suggestion
  private @NotNull String formatMessage(final @NotNull String message, final @Nullable Object[] parameters) {
```

sentry-jul/src/main/java/io/sentry/jul/SentryHandler.java
```diff
@@ -0,0 +1,283 @@
...
+      Sentry.close();
+    } catch (RuntimeException e) {
+      reportError(
+          "An exception occurred while closing the Sentry connection",
+          e,
+          ErrorManager.CLOSE_FAILURE);
+    }
+  }
+
+  private @NotNull SdkVersion createSdkVersion(@NotNull SentryOptions sentryOptions) {
```
marandaneto, line 231:
```suggestion
  private @NotNull SdkVersion createSdkVersion(final @NotNull SentryOptions sentryOptions) {
```

sentry-jul/src/main/java/io/sentry/jul/SentryHandler.java
marandaneto: oh just saw you've mocked the `transport` instead

sentry-jul/src/main/java/io/sentry/jul/SentryHandler.java
marandaneto: ok theres a test for it, please ignore

sentry-jul/src/main/java/io/sentry/jul/SentryHandler.java
```diff
@@ -0,0 +1,283 @@
...
+  @TestOnly
+  SentryHandler(final @NotNull SentryOptions options, final boolean configureFromLogManager) {
+    setFilter(new DropSentryFilter());
+    if (configureFromLogManager) {
+      retrieveProperties();
+    }
+    if (!Sentry.isEnabled()) {
+      options.setEnableExternalConfiguration(true);
+      options.setSdkVersion(createSdkVersion(options));
+      Sentry.init(options);
```
marandaneto, line 56: I don't think this is going to be used by Android at all, but it'd be possible, right now we are limiting it as the Handler inits Sentry itself, another approach would be to make it an Integration, as we do with `Timber`, so people initialize either the pure Java or Android SDK.
maybe @bruno-garcia have opinions?

marandaneto (approved):
a few comments but looks good and ready to go.
LGTM
The next step would be to create docs too, probably linking to the migration page.

sentry-jul/src/main/java/io/sentry/jul/SentryHandler.java
bruno-garcia: is `jul` used on Android at all?

The limitations with this approach is that on Android the SDK would need to be init'ed before the logging integration which should be the case (either the AndroidManifest or initing right at the start of the app) so I think it's OK to go ahead with this.

marandaneto: that's what I've tried to say, it's not common at all, but it'd be possible, JUL is part of the Android SDK, for people that care about APK size, its been always an option, it's a trade-off, I'd prefer one-liner integration and making it possible on the mobile world other than not possible at all, well, at least no device context, etc.

maciejwalkowiak: On mobile, Sentry initializes before or after handler? If Sentry gets initialized after handler is created, it will overwrite what was configured in the handler.

marandaneto: if `auto-init` mode is on, it's before

marandaneto: as the JUL supported has been asked by Java and not Android people, let's keep it as it is and if at some point Android people ask for that, we reconsider that.

marandaneto:
@maciejwalkowiak CI is not happy

PR #1213, opened 2021-01-29
Improve Kotlin compatibility for SdkVersion class

marandaneto:
## :scroll: Description
Improve Kotlin compatibility for SdkVersion class

https://github.com/getsentry/sentry-dart/pull/287
Kotlin integrations will need to take care of null fields that are not null if not annotated

## :pencil: Checklist

- [X] No breaking changes


sentry/src/main/java/io/sentry/protocol/SdkVersion.java
```diff
@@ -23,7 +25,7 @@
    * <p>Official Sentry SDKs use the entity `sentry`, as in `sentry.python` or
    * `sentry.javascript.react-native`. Please use a different entity for your own SDKs.
    */
-  private String name;
```
maciejwalkowiak, line 26: If we make these fields not nullable, they have to be either required in the constructor on they have to have default non-null values. Otherwise there is a mismatch when `new SdkVersion()` is executed.

marandaneto: good catch, we can't make it final though because of JSON deserialization

maciejwalkowiak (approved):

PR #1214, opened 2021-02-01
fix: Use a different method to get strings from JNI

Swatinem:
## :scroll: Description

This uses a different method to get string contents from JNI into native options, and avoids excessive stack allocation.

## :pencil: Checklist

- [ ] No breaking changes


marandaneto:
@Swatinem currently this causes: `JNI DETECTED ERROR IN APPLICATION: use of deleted local reference 0x71`

marandaneto:
@Swatinem can we close this?

Swatinem:
I think this would still be interesting to do, as I think the current way is a bit fragile.
More specifically, I think explicitly handling jni references, and cleaning up all the string code would be a good idea.

marandaneto (approved):

PR #1219, opened 2021-02-02
Update README to reflect new links and remove dup links

marandaneto:
## :scroll: Description
Update README to reflect new links and remove dup links

Added new links, fixed versions, and removed dup. links (which is already under Resources)

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,21 +26,17 @@ Sentry SDK for Java and Android
 
 
 
-# Java SDK 3.0 Docs
+# Useful links and docs
 
-The Java SDK documentation [can be found on docs.sentry.io](https://docs.sentry.io/platforms/java/).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,22 +26,18 @@ Sentry SDK for Java and Android
 
 
 
-# Java SDK 3.0 Docs
+# Useful links and docs
 
-The Java SDK documentation [can be found on docs.sentry.io](https://docs.sentry.io/platforms/java/).
```
philipphofmann, line 31: `m`: Shouldn't we keep a link to the SKD docs?

philipphofmann (approved):
LGTM

README.md
marandaneto: > removed dup. links (which is already under Resources)

PR #1220, opened 2021-02-02
Fix: set inApp=true for native events

marandaneto:
## :pencil: Checklist

- [ ] No breaking changes

### DIFF
--- a/sentry/src/main/java/io/sentry/MainEventProcessor.java
+++ b/sentry/src/main/java/io/sentry/MainEventProcessor.java
@@ -1,6 +1,8 @@
 package io.sentry;
 
 import io.sentry.protocol.SentryException;
+import io.sentry.protocol.SentryStackFrame;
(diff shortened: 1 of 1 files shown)


sentry/src/main/java/io/sentry/MainEventProcessor.java
```diff
@@ -85,6 +87,36 @@
               event.getEventId());
     }
 
+    // read packages name from options (without @version+build)
+    final String appPackage = "io.sentry.samples.android";
+    final String appJniPackage = appPackage.replace(".", "_");
+
+    // other way would be to pass the packages name as inAppIncludes to sentry-native, but they dont
+    // expose this option, yet.
```
marandaneto, line 95: ideally, we'd go for https://github.com/getsentry/sentry-native/issues/472 instead of this hacky solution.

marandaneto:
the way to go is https://github.com/getsentry/sentry-native/issues/472

PR #1222, opened 2021-02-02
Fix: Remove misleading error logs

### LINKED ISSUE #1221: Logging 'Stack peek was null when {method}' is misleading for some Hub calls.
#1127 introduced changes on how to deal with the Scope stack but logging is off in a few cases.

eg:
`setFingerprint`
`setUser`
`clearBreadcrumbs`

If the stack can't be null, should we just remove them?

maciejwalkowiak:
## :scroll: Description

Fix: Remove misleading error logs

Fixes #1221

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/sentry/src/main/java/io/sentry/Hub.java
+++ b/sentry/src/main/java/io/sentry/Hub.java
@@ -312,7 +312,6 @@ public final class Hub implements IHub {
           .log(SentryLevel.WARNING, "Instance is disabled and this 'setUser' call is a no-op.");
     } else {
       stack.peek().getScope().setUser(user);
-      options.getLogger().log(SentryLevel.FATAL, "Stack peek was null when setUser");
     }
   }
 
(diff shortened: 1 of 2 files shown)


sentry/src/main/java/io/sentry/Hub.java
```diff
@@ -482,13 +479,8 @@ public void bindClient(final @NotNull ISentryClient client) {
           .log(SentryLevel.WARNING, "Instance is disabled and this 'bindClient' call is a no-op.");
     } else {
       final StackItem item = stack.peek();
-      if (client != null) {
-        options.getLogger().log(SentryLevel.DEBUG, "New client bound to scope.");
-        item.setClient(client);
-      } else {
-        options.getLogger().log(SentryLevel.DEBUG, "NoOp client bound to scope.");
-        item.setClient(NoOpSentryClient.getInstance());
-      }
```
marandaneto, line 491: I guess we should keep this, `client` can be null, it's not really about the stack peek being null like the others

maciejwalkowiak: `client` is marked as `@NotNull` in `StackItem`, is this incorrect?

marandaneto: `client` is a method param.

that's why we set either the `client` instance or `NoOpSentryClient.getInstance()`
you're correct about `StackItem` having non-null `client`

maciejwalkowiak: @marandaneto thanks, you're right. Fixed.

marandaneto (approved):

PR #1224, opened 2021-02-03
Fix: Makefile upload symbols and use jfrog cli to sync over to maven central

marandaneto:
## :scroll: Description
Fix: Makefile upload symbols and use jfrog cli to sync over to maven central

we always need to remember uploading debug symbols.

maven central sync always fails via the gradle plugin, now we guarantee that upload is done to bintray/jcenter and we use the `jfrog cli` to sync over to maven central, which we could retry standalone if it fails.

you must export `$VERSION`.

## :pencil: Checklist

- [X] No breaking changes

## :crystal_ball: Next steps
Craft

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,9 +14,30 @@ compile:
 dryRelease:
 	./gradlew bintrayUpload -PbintrayUser=dryUser -PbintrayKey=dryKey
 
-# deploy the current build to bintray, jcenter and maven central
+# deploy the current build to bintray, jcenter and maven central, upload debug symbols to symbol collector
 doRelease:
-	./gradlew bintrayUpload -PbintrayUser="$(BINTRAY_USERNAME)" -PbintrayKey="$(BINTRAY_KEY)" -PmavenCentralUser="$(MAVEN_USER)" -PmavenCentralPassword="$(MAVEN_PASS)" -PmavenCentralSync=true -PdryRun=false --info
+	# bintray/jcenter upload
+	./gradlew bintrayUpload -PbintrayUser="$(BINTRAY_USERNAME)" -PbintrayKey="$(BINTRAY_KEY)" -PdryRun=false --info
+
+	# upload symbols
+	cd sentry-android-ndk
+	# be sure SymbolCollector.Console is in your $PATH
+	SymbolCollector.Console --upload directory --path ./ --batch-type android --bundle-id sentry-android-ndk-$VERSION --server-endpoint https://symbol-collector.services.sentry.io
+	cd ..
+
+	# maven central sync using jfrog cli
+	jfrog bt mcs --user=$BINTRAY_USERNAME --key=$BINTRAY_KEY --sonatype-username=$MAVEN_USER --sonatype-password=$MAVEN_PASS getsentry/sentry-java/io.sentry:sentry/$VERSION
+	jfrog bt mcs --user=$BINTRAY_USERNAME --key=$BINTRAY_KEY --sonatype-username=$MAVEN_USER --sonatype-password=$MAVEN_PASS getsentry/sentry-java/io.sentry:sentry-log4j2/$VERSION
+	jfrog bt mcs --user=$BINTRAY_USERNAME --key=$BINTRAY_KEY --sonatype-username=$MAVEN_USER --sonatype-password=$MAVEN_PASS getsentry/sentry-java/io.sentry:sentry-logback/$VERSION
+	jfrog bt mcs --user=$BINTRAY_USERNAME --key=$BINTRAY_KEY --sonatype-username=$MAVEN_USER --sonatype-password=$MAVEN_PASS getsentry/sentry-java/io.sentry:sentry-servlet/$VERSION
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -14,9 +14,30 @@ compile:
...
+# deploy the current build to bintray, jcenter and maven central, upload debug symbols to symbol collector
 doRelease:
-	./gradlew bintrayUpload -PbintrayUser="$(BINTRAY_USERNAME)" -PbintrayKey="$(BINTRAY_KEY)" -PmavenCentralUser="$(MAVEN_USER)" -PmavenCentralPassword="$(MAVEN_PASS)" -PmavenCentralSync=true -PdryRun=false --info
+	# bintray/jcenter upload
+	./gradlew bintrayUpload -PbintrayUser="$(BINTRAY_USERNAME)" -PbintrayKey="$(BINTRAY_KEY)" -PdryRun=false --info
+
+	# upload symbols
+	cd sentry-android-ndk
+	# be sure SymbolCollector.Console is in your $PATH - https://github.com/getsentry/symbol-collector/releases
+	SymbolCollector.Console --upload directory --path ./ --batch-type android --bundle-id sentry-android-ndk-$VERSION --server-endpoint https://symbol-collector.services.sentry.io
```
bruno-garcia, line 25: Maybe throw this as the last step so if we don't have this on the path it's 1 thing only to fix and rerun instead of all the other jfrog stuff after

marandaneto: I thought about that but tbh I'd prefer that we upload artifacts and symbols which is almost never-failing and only later the maven central pain.
we could even split it up to a different make command for example.
`make mc sync` or whatever, so we don't bound a task that is always successful with another one that is always broken

marandaneto: although my comment above is not really about your comment which is not having in the `path`, maybe the script could test if `SymbolCollector.Console` and `jfrog` are valid cms even before starting the whole thing

Makefile
```diff
@@ -14,9 +14,30 @@ compile:
...
+	jfrog bt mcs --user=$BINTRAY_USERNAME --key=$BINTRAY_KEY --sonatype-username=$MAVEN_USER --sonatype-password=$MAVEN_PASS getsentry/sentry-java/io.sentry:sentry-spring-boot-starter/$VERSION
+	jfrog bt mcs --user=$BINTRAY_USERNAME --key=$BINTRAY_KEY --sonatype-username=$MAVEN_USER --sonatype-password=$MAVEN_PASS getsentry/sentry-java/io.sentry:sentry-jul/$VERSION
+	jfrog bt mcs --user=$BINTRAY_USERNAME --key=$BINTRAY_KEY --sonatype-username=$MAVEN_USER --sonatype-password=$MAVEN_PASS getsentry/sentry-android/io.sentry:sentry-android/$VERSION
+	jfrog bt mcs --user=$BINTRAY_USERNAME --key=$BINTRAY_KEY --sonatype-username=$MAVEN_USER --sonatype-password=$MAVEN_PASS getsentry/sentry-android/io.sentry:sentry-android-core/$VERSION
+	jfrog bt mcs --user=$BINTRAY_USERNAME --key=$BINTRAY_KEY --sonatype-username=$MAVEN_USER --sonatype-password=$MAVEN_PASS getsentry/sentry-android/io.sentry:sentry-android-ndk/$VERSION
+	jfrog bt mcs --user=$BINTRAY_USERNAME --key=$BINTRAY_KEY --sonatype-username=$MAVEN_USER --sonatype-password=$MAVEN_PASS getsentry/sentry-android/io.sentry:sentry-android-timber/$VERSION
```
philipphofmann, line 40: `l`: Maybe something like this would make it a bit easier to maintain.

```suggestion
	make upload-to-jfrog url=getsentry/sentry-java/io.sentry:sentry/$(VERSION)
	make upload-to-jfrog url=getsentry/sentry-java/io.sentry:sentry-log4j2/$(VERSION)
	make upload-to-jfrog url=getsentry/sentry-java/io.sentry:sentry-logback/$(VERSION)
	make upload-to-jfrog url=getsentry/sentry-java/io.sentry:sentry-servlet/$(VERSION)
	make upload-to-jfrog url=getsentry/sentry-java/io.sentry:sentry-spring/$(VERSION)
	make upload-to-jfrog url=getsentry/sentry-java/io.sentry:sentry-apache-http-client-5/$(VERSION)
	make upload-to-jfrog url=getsentry/sentry-java/io.sentry:sentry-spring-boot-starter/$(VERSION)
	make upload-to-jfrog url=getsentry/sentry-java/io.sentry:sentry-jul/$(VERSION)
	make upload-to-jfrog url=getsentry/sentry-android/io.sentry:sentry-android/$(VERSION)
	make upload-to-jfrog url=getsentry/sentry-android/io.sentry:sentry-android-core/$(VERSION)
	make upload-to-jfrog url=getsentry/sentry-android/io.sentry:sentry-android-ndk/$(VERSION)
	make upload-to-jfrog url=getsentry/sentry-android/io.sentry:sentry-android-timber/$(VERSION)

upload-to-jfrog:
	jfrog bt mcs --user=$BINTRAY_USERNAME --key=$BINTRAY_KEY --sonatype-username=$MAVEN_USER --sonatype-password=$MAVEN_PASS $(url)
```

marandaneto: this PR is probably gonna be closed due to recent announcement, bintray and jcenter are gonna be retired.

philipphofmann: Ah, that's true.

marandaneto:
not useful because of #1228

PR #1231, opened 2021-02-04
Fix: Mark stacktrace as snapshot if captured at arbitrary moment

marandaneto:
## :scroll: Description
Fix: Mark stacktrace as snapshot if captured at arbitrary moment

ANR stack traces and the thread list attach a stack trace captured at an arbitrary moment and this could affect grouping, this flag will be used for better grouping, maybe using hierarchical grouping or something else on the server-side.

## :pencil: Checklist

- [X] No breaking changes

### DIFF
--- a/sentry-android-core/src/test/java/io/sentry/android/core/AnrIntegrationTest.kt
+++ b/sentry-android-core/src/test/java/io/sentry/android/core/AnrIntegrationTest.kt
@@ -1,9 +1,12 @@
 package io.sentry.android.core
 
+import android.content.Context
 import com.nhaarman.mockitokotlin2.any
+import com.nhaarman.mockitokotlin2.check
 import com.nhaarman.mockitokotlin2.mock
 import com.nhaarman.mockitokotlin2.verify
 import io.sentry.IHub
+import io.sentry.exception.ExceptionMechanismException
 import kotlin.test.BeforeTest
 import kotlin.test.Test
 import kotlin.test.assertNotNull
@@ -12,49 +15,80 @@ import kotlin.test.assertTrue
 
 class AnrIntegrationTest {
 
-    private val integration = AnrIntegration(mock())
+    private class Fixture {
+        val context = mock<Context>()
+        val hub = mock<IHub>()
+        val options = SentryAndroidOptions()
+
+        fun getSut(): AnrIntegration {
+            return AnrIntegration(context)
+        }
+    }
+
+    private val fixture = Fixture()
 
     @BeforeTest
     fun `before each test`() {
+        val sut = fixture.getSut()
         // watch dog is static and has shared state
-        integration.close()
+        sut.close()
     }
 
     @Test
     fun `When ANR is enabled, ANR watch dog should be started`() {
-        val options = SentryAndroidOptions()
-        val hub = mock<IHub>()
-        integration.register(hub, options)
-        assertNotNull(integration.anrWatchDog)
-        assertTrue((integration.anrWatchDog as ANRWatchDog).isAlive)
+        val sut = fixture.getSut()
+
+        sut.register(fixture.hub, fixture.options)
+
+        assertNotNull(sut.anrWatchDog)
+        assertTrue((sut.anrWatchDog as ANRWatchDog).isAlive)
     }
 
     @Test
     fun `When ANR is disabled, ANR should not be started`() {
-        val options = SentryAndroidOptions()
-        options.isAnrEnabled = false
-        val hub = mock<IHub>()
-        val integration = AnrIntegration(mock())
-        integration.register(hub, options)
-        assertNull(integration.anrWatchDog)
+        val sut = fixture.getSut()
+        fixture.options.isAnrEnabled = false
+
+        sut.register(fixture.hub, fixture.options)
+
+        assertNull(sut.anrWatchDog)
     }
 
     @Test
     fun `When ANR watch dog is triggered, it should capture exception`() {
-        val hub = mock<IHub>()
-        val integration = AnrIntegration(mock())
-        integration.reportANR(hub, mock(), mock())
-        verify(hub).captureException(any())
+        val sut = fixture.getSut()
+
+        sut.reportANR(fixture.hub, mock(), getApplicationNotResponding())
+
+        verify(fixture.hub).captureException(any())
     }
 
     @Test
     fun `When ANR integration is closed, watch dog should stop`() {
-        val options = SentryAndroidOptions()
-        val hub = mock<IHub>()
-        val integration = AnrIntegration(mock())
-        integration.register(hub, options)
-        assertNotNull(integration.anrWatchDog)
-        integration.close()
-        assertNull(integration.anrWatchDog)
+        val sut = fixture.getSut()
+
+        sut.register(fixture.hub, fixture.options)
+
+        assertNotNull(sut.anrWatchDog)
+
+        sut.close()
+
+        assertNull(sut.anrWatchDog)
+    }
+
+    @Test
+    fun `When ANR watch dog is triggered, snapshot flag should be true`() {
+        val sut = fixture.getSut()
+
+        sut.reportANR(fixture.hub, mock(), getApplicationNotResponding())
+
+        verify(fixture.hub).captureException(check {
+            val ex = it as ExceptionMechanismException
+            assertTrue(ex.isSnapshot)
+        })
+    }
+
+    private fun getApplicationNotResponding(): ApplicationNotResponding {
+        return ApplicationNotResponding("ApplicationNotResponding", Thread.currentThread())
     }
 }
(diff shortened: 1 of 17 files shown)


sentry/src/main/java/io/sentry/SentryExceptionFactory.java
```diff
@@ -86,7 +88,11 @@ public SentryExceptionFactory(final @NotNull SentryStackTraceFactory sentryStack
     final List<SentryStackFrame> frames =
         sentryStackTraceFactory.getStackFrames(throwable.getStackTrace());
     if (frames != null && !frames.isEmpty()) {
-      exception.setStacktrace(new SentryStackTrace(frames));
+      final SentryStackTrace sentryStackTrace = new SentryStackTrace(frames);
+      if (snapshot) {
+        sentryStackTrace.setSnapshot(true);
+      }
```
bruno-garcia, line 94: The preference is to not add the flag at all if it isn't snapshot?

Otherwise:
```suggestion
      sentryStackTrace.setSnapshot(snapshot);
```

marandaneto: yep, so no need for serialization if its null, we do the same for all the other booleans/nullable fields

marandaneto:
@jan-auer added you as a reviewer just to be aware that we are adding it to Android

bruno-garcia:
@jan-auer is it correct to assume that `snapshot` stacktraces will be used for grouping based on the top-most frames?

This is what we're aiming for ANR and if that's not what `snapshot` is for, I'm afraid it won't be helpful to add this at all.

jan-auer:
> is it correct to assume that `snapshot` stacktraces will be used for grouping based on the top-most frames?

@bruno-garcia Since `snapshot` stack traces are captured at an indeterministic time (based on external influence), it is likely that the top frame is random just like the rest of the stack trace. I'm afraid that grouping by the top frame wouldn't achieve better results in that case. If there are other indicators like the error type or a signal, we could group by that.

marandaneto:
> > is it correct to assume that `snapshot` stacktraces will be used for grouping based on the top-most frames?
> 
> @bruno-garcia Since `snapshot` stack traces are captured at an indeterministic time (based on external influence), it is likely that the top frame is random just like the rest of the stack trace. I'm afraid that grouping by the top frame wouldn't achieve better results in that case. If there are other indicators like the error type or a signal, we could group by that.

`mechanism.type=ANR` for ANR events.
@jan-auer when you say the `top` frame do you mean where it started or ended? I ask that because we revert the order of the frames before sending it right, so I am confused here.

sentry-android-core/src/test/java/io/sentry/android/core/AnrIntegrationTest.kt
```diff
@@ -12,49 +15,80 @@ import kotlin.test.assertTrue
 
 class AnrIntegrationTest {
 
-    private val integration = AnrIntegration(mock())
+    private class Fixture {
```
philipphofmann, line 18: Thanks for refactoring this 👍🏼

sentry-android-core/src/test/java/io/sentry/android/core/AnrIntegrationTest.kt
```diff
@@ -12,49 +15,80 @@ import kotlin.test.assertTrue
...
     fun `When ANR watch dog is triggered, it should capture exception`() {
-        val hub = mock<IHub>()
-        val integration = AnrIntegration(mock())
-        integration.reportANR(hub, mock(), mock())
-        verify(hub).captureException(any())
+        val sut = fixture.getSut()
+
+        sut.reportANR(fixture.hub, mock(), getApplicationNotResponding())
+
+        verify(fixture.hub).captureException(any())
```
philipphofmann, line 63: `l`: We could check for the actual exception and not just any.

sentry-android-core/src/test/java/io/sentry/android/core/AnrIntegrationTest.kt
```diff
@@ -12,49 +15,80 @@ import kotlin.test.assertTrue
...
+
+        assertNotNull(sut.anrWatchDog)
+
+        sut.close()
+
+        assertNull(sut.anrWatchDog)
+    }
+
+    @Test
+    fun `When ANR watch dog is triggered, snapshot flag should be true`() {
```
philipphofmann, line 80: Nice test method naming

sentry/src/main/java/io/sentry/SentryExceptionFactory.java
```diff
@@ -62,11 +62,13 @@ public SentryExceptionFactory(final @NotNull SentryStackTraceFactory sentryStack
    * @param exceptionMechanism The optional {@link Mechanism} of the {@code throwable}. Or null if
    *     none exist.
    * @param thread The optional {@link Thread} which the exception originated. Or null if not known.
+   * @param snapshot if the captured {@link java.lang.Thread}'s stacktrace is a snapshot.
```
philipphofmann, line 65: `l`: Maybe a short explanation of what a snapshot is would be lovely. A link to `SentryStackTrace.snapshot` would do the job.

sentry/src/main/java/io/sentry/protocol/SentryStackTrace.java
```diff
@@ -72,21 +79,29 @@ public SentryStackTrace(List<SentryStackFrame> frames) {
...
+  public @Nullable Map<String, String> getRegisters() {
     return registers;
   }
 
-  public void setRegisters(Map<String, String> registers) {
+  public void setRegisters(final @Nullable Map<String, String> registers) {
     this.registers = registers;
   }
+
+  public @Nullable Boolean getSnapshot() {
```
philipphofmann, line 100: `m`: Java naming convention.
```suggestion
  public @Nullable Boolean isSnapshot() {
```

sentry/src/main/java/io/sentry/protocol/SentryStackTrace.java
```diff
@@ -40,21 +41,27 @@
...
    *
    * <p>A map of register names and their values. The values should contain the actual register
    * values of the thread, thus mapping to the last frame in the list.
    */
-  private Map<String, String> registers;
+  private @Nullable Map<String, String> registers;
+
+  /**
+   * This flag indicates that this stack trace is captured at an arbitrary moment in time and this
+   * would affect the quality of grouping, Sentry will special case if this is set to true.
```
philipphofmann, line 55: `m`: Can we maybe explain more in detail how this would affect grouping and how Sentry is going to special case this?

sentry/src/test/java/io/sentry/SentryExceptionFactoryTest.kt
```diff
@@ -70,11 +71,22 @@ class SentryExceptionFactoryTest {
...
         val error = Exception("Exception")
 
-        val throwable = ExceptionMechanismException(mechanism, error, null)
+        val throwable = ExceptionMechanismException(mechanism, error, Thread.currentThread())
 
         val sentryExceptions = fixture.getSut().getSentryExceptions(throwable)
         assertEquals("anr", sentryExceptions[0].mechanism.type)
-        assertEquals(false, sentryExceptions[0].mechanism.isHandled)
+        assertFalse(sentryExceptions[0].mechanism.isHandled)
+        assertNull(sentryExceptions[0].stacktrace?.snapshot)
```
philipphofmann, line 79: `l`: It would be great to test this also with multiple exceptions and not just one.

philipphofmann (approved):
LGTM, just made a few minor comments.

sentry/src/main/java/io/sentry/protocol/SentryStackTrace.java
marandaneto: the convention `is` is for the `boolean` primitive type, but here the object is nullable.

sentry/src/main/java/io/sentry/protocol/SentryStackTrace.java
marandaneto: it's not defined yet, backend implementation is not even started.

sentry-android-core/src/test/java/io/sentry/android/core/AnrIntegrationTest.kt
marandaneto: `ANRWatchDogTest` already does it, which is the class that actually creates the exception

sentry/src/main/java/io/sentry/SentryExceptionFactory.java
marandaneto: SentryExceptionFactory is an internal class, private API.
`SentryStackTrace.snapshot` already has a user friendly javadocs, a link makes sense

jan-auer:
> @jan-auer when you say the `top` frame do you mean where it started or ended? I ask that because we revert the order of the frames before sending it right, so I am confused here.

@marandaneto Good point, this is ambiguous terminology. With "top" I always mean the inner-most frame.

sentry/src/test/java/io/sentry/SentryExceptionFactoryTest.kt
marandaneto: this class already has tests for an exception with multiple causes (aka inner exception)

PR #1235, opened 2021-02-08
Add "data" property to Span

### LINKED ISSUE #1233: Span.data is missing
https://develop.sentry.dev/sdk/event-payloads/span/

We don't have a field to set arbitrary data associated with the Span, wondering if we decided to drop it or was an oversight.

Context: https://github.com/getsentry/sentry-cocoa/pull/932#discussion_r570927177

maciejwalkowiak:
## :scroll: Description

Add "data" property to Span.

Fixes #1233

What I discovered in meantime is that `transactions` do not support `data` attribute. It's been [suggested](https://discord.com/channels/621778831602221064/777210445442187274/808378671161278534) to use `extra` field on transactions to comply with `ISpan` interface.

Unit tests.

## :pencil: Checklist

- [ ] No breaking changes

### DIFF
--- a/sentry/src/main/java/io/sentry/SentryBaseEvent.java
+++ b/sentry/src/main/java/io/sentry/SentryBaseEvent.java
@@ -7,6 +7,7 @@ import io.sentry.protocol.SdkVersion;
 import io.sentry.protocol.SentryId;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
@@ -58,6 +59,13 @@ public abstract class SentryBaseEvent {
    * <p>```json { "environment": "production" } ```
    */
   private String environment;
+
+  /**
+   * Arbitrary extra information set by the user.
+   *
+   * <p>```json { "extra": { "my_key": 1, "some_other_value": "foo bar" } }```
+   */
+  private final Map<String, Object> extra = new ConcurrentHashMap<>();
 
   /** The captured Throwable */
   protected transient @Nullable Throwable throwable;
@@ -173,5 +181,28 @@ public abstract class SentryBaseEvent {
 
   public void setEnvironment(String environment) {
     this.environment = environment;
+  }
+
+  Map<String, Object> getExtras() {
+    return extra;
+  }
+
+  public void setExtras(Map<String, Object> extra) {
+    final Map<String, Object> copy = new HashMap<>(extra);
+    for (Map.Entry<String, Object> entry : copy.entrySet()) {
+      this.extra.put(entry.getKey(), entry.getValue());
+    }
+  }
+
+  public void setExtra(String key, Object value) {
+    extra.put(key, value);
+  }
+
+  public void removeExtra(@NotNull String key) {
+    extra.remove(key);
+  }
+
+  public @Nullable Object getExtra(final @NotNull String key) {
+    return extra.get(key);
   }
 }
--- a/sentry/src/main/java/io/sentry/SentryBaseEvent.java
+++ b/sentry/src/main/java/io/sentry/SentryBaseEvent.java
@@ -183,22 +183,29 @@ public abstract class SentryBaseEvent {
     this.environment = environment;
   }
 
+  @NotNull
   Map<String, Object> getExtras() {
     return extra;
   }
 
-  public void setExtras(Map<String, Object> extra) {
+  /**
+   * Copies extras from the map given by parameter to extras on the event. If key already exists,
+   * its value gets overwritten.
+   *
+   * @param extra - the extras
+   */
+  public void setExtras(final @NotNull Map<String, Object> extra) {
     final Map<String, Object> copy = new HashMap<>(extra);
     for (Map.Entry<String, Object> entry : copy.entrySet()) {
       this.extra.put(entry.getKey(), entry.getValue());
     }
   }
 
-  public void setExtra(String key, Object value) {
+  public void setExtra(final @NotNull String key, final @NotNull Object value) {
     extra.put(key, value);
   }
 
-  public void removeExtra(@NotNull String key) {
+  public void removeExtra(final @NotNull String key) {
     extra.remove(key);
   }
 
--- a/sentry/src/main/java/io/sentry/SentryTransaction.java
+++ b/sentry/src/main/java/io/sentry/SentryTransaction.java
@@ -24,6 +24,7 @@ public final class SentryTransaction extends SentryBaseEvent implements ITransac
 
   /** A list of spans within this transaction. Can be empty. */
   private final @NotNull List<Span> spans = new CopyOnWriteArrayList<>();
+
   /**
    * A hub this transaction is attached to. Marked as transient to be ignored during JSON
    * serialization.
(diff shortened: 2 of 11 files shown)


sentry/src/main/java/io/sentry/SentryBaseEvent.java
```diff
@@ -174,4 +182,27 @@ public String getEnvironment() {
...
+  public void setExtras(Map<String, Object> extra) {
+    final Map<String, Object> copy = new HashMap<>(extra);
+    for (Map.Entry<String, Object> entry : copy.entrySet()) {
+      this.extra.put(entry.getKey(), entry.getValue());
+    }
+  }
+
+  public void setExtra(String key, Object value) {
+    extra.put(key, value);
+  }
```
marandaneto, line 199: annotations

sentry/src/main/java/io/sentry/SentryBaseEvent.java
```diff
@@ -174,4 +182,27 @@ public String getEnvironment() {
...
+
+  Map<String, Object> getExtras() {
+    return extra;
+  }
+
+  public void setExtras(Map<String, Object> extra) {
+    final Map<String, Object> copy = new HashMap<>(extra);
+    for (Map.Entry<String, Object> entry : copy.entrySet()) {
+      this.extra.put(entry.getKey(), entry.getValue());
+    }
```
marandaneto, line 194: if we overwrite its existent keys, let's add java docs about this usecase

marandaneto:
missing changelog

sentry/src/test/java/io/sentry/GsonSerializerTest.kt
```diff
@@ -486,6 +494,8 @@ class GsonSerializerTest {
         assertNotNull(transaction.timestamp)
         assertNotNull(transaction.contexts)
         assertNotNull(transaction.contexts.trace)
+        assertEquals("data-value", transaction.getData("string-data-key"))
```
marandaneto, line 497: add a test setting a `Span#setData` too, right now only `Transaction` is being tested

sentry/src/test/java/io/sentry/GsonSerializerTest.kt
```diff
@@ -503,6 +503,56 @@ class GsonSerializerTest {
...
+        val parentSpanId = SpanId()
+        val span = Span(sentryId, parentSpanId, SentryTransaction("tx"), NoOpHub.getInstance())
+        span.setData("data-key", "data-value")
+        span.finish()
+
+        val stringWriter = StringWriter()
+        fixture.serializer.serialize(span, stringWriter)
+
+        val toString = stringWriter.toString()
+        println(toString)
```
marandaneto, line 518: use the function `serializeToString`

sentry/src/test/java/io/sentry/GsonSerializerTest.kt
```diff
@@ -503,6 +503,56 @@ class GsonSerializerTest {
...
+                "data-key":"data-value"
+              },
+              "trace_id":"bb185b9ad3ae489eb3645c1d6657b2e0",
+              "span_id":"f67e9eafb33f444e",
+              "parent_span_id":"15c304828711435c",
+              "tags":{}
+            }
+        """.trimIndent()
+
+        val span = fixture.serializer.deserialize(StringReader(json), Span::class.java)
```
marandaneto, line 546: let's make `serializeToString` `reified` so we could val str = serializeToString<T>(instance)?

maciejwalkowiak: It can be just `Any`

sentry/src/main/java/io/sentry/SentryTransaction.java
```diff
@@ -204,6 +205,16 @@ public void setDescription(@Nullable String description) {
     return this.context;
   }
 
+  @Override
+  public void setData(@NotNull String key, @NotNull Object value) {
```
bruno-garcia, line 209: Is it worth adding setData if event has setExtra already?
The other SDKs where transactions is just an event (without the separation we did here) will have just `setExtra` in the docs and samples?

marandaneto: https://github.com/getsentry/sentry-cocoa/pull/932#discussion_r573211307

bruno-garcia: Made a comment there. I think adding stuff to transaction such as that field isn't the way to go since this isn't what the protocol has. It'll be confusing to have both Extra and Data.

We have to remember that in reality Transaction is an Event, we just turned things around here to simplify (not sure it did at this point) things.

maciejwalkowiak: It had to be added here because `ITransaction` extends `ISpan`. Docs also say that each transaction is a span, that's why we have such design. Now I believe it would be just better if we make this statement true again and add `data` to transactions on the Sentry backend.

bruno-garcia: Maybe that's the problem. Having transation inherit from Span.

`data` will definitely not be added on the backend.

Sentry and its SDKs: Transction is event with type=transaction.

We decided to create a class for it and split some fields. Maybe we just don't inherit from Span? 
Composition only in this case.

maciejwalkowiak: We can think about it but it will probably change the design quite a bit. RIght now, when you get the current span to create a child span, you don't need to worry if it's a transaction or if it's a span - you just get an ISpan that lets you create children.

bruno-garcia: We should discuss this because we'll need to get to a design where spans can be created alone.
Transaction is a container to capture a set of spans (and is a Sentry event) where a Span is not, and one day could be ingested individually (rather in batches) but outside of a Transaction.

Friday we can discuss this if not earlier. I'll see if @mitsuhiko can join us.

marandaneto: this is part of our agenda for tomorrow

marandaneto: See https://github.com/getsentry/sentry-java/issues/1233#issuecomment-779235496

maciejwalkowiak:
Since there is a plan to hide `ITransaction` completely from the public API (https://github.com/getsentry/sentry-java/issues/1251), this PR is not relevant anymore.

marandaneto:
> Since there is a plan to hide `ITransaction` completely from the public API (#1251), this PR is not relevant anymore.

this PR was about the `Span#data` though, I still see it relevant, we'd need to revert the `Transaction#data` and `Transaction#extra` though, are you planning to open a new PR then?

maciejwalkowiak:
Since `ITransaction` extends `ISpan`, adding `data` to spans means also adding it to transactions. Do you have ideas on how to add `data` to spans and not to transactions without doing (#1251) first?

marandaneto:
> Since `ITransaction` extends `ISpan`, adding `data` to spans means also adding it to transactions. Do you have ideas on how to add `data` to spans and not to transactions without doing (#1251) first?

yeah makes sense, composition over inheritance would help here but I agree to do the other one first would be easier

PR #1247, opened 2021-02-12
Optimize DuplicateEventDetectionEventProcessor performance.

maciejwalkowiak:
## :scroll: Description

Optimize potential `DuplicateEventDetectionEventProcessor` performance issue by changing how objects are stored from `WeakHashMap` to `ConcurrentLinkedQueue`.

Using WeakHashMap to store objects opens up a possibility that the map will grow to thousands of entries and causes performance loss when searching for stored exception. This may happen either if outside of the SDK code references to exceptions are kept or if GC does not run for long time.

Changing the implementation to `ConcurrentLinkedQueue` gives the event processor control over how many exceptions are stored in the memory.

Unit tests.

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Enchancement: Support @SentrySpan and @SentryTransaction on classes and interfaces. (#1243)
 * Enchancement: Do not serialize empty collections and maps (#1245)
 * Ref: Simplify RestTemplate instrumentation (#1246)
+* Ref: Optimize DuplicateEventDetectionEventProcessor performance.
 
 # 4.1.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 * Enchancement: Support @SentrySpan and @SentryTransaction on classes and interfaces. (#1243)
 * Enchancement: Do not serialize empty collections and maps (#1245)
 * Ref: Simplify RestTemplate instrumentation (#1246)
-* Ref: Optimize DuplicateEventDetectionEventProcessor performance.
+* Ref: Optimize DuplicateEventDetectionEventProcessor performance (#1247).
 
 # 4.1.0
 
--- a/sentry/src/main/java/io/sentry/DuplicateEventDetectionEventProcessor.java
+++ b/sentry/src/main/java/io/sentry/DuplicateEventDetectionEventProcessor.java
@@ -37,6 +37,8 @@ public final class DuplicateEventDetectionEventProcessor implements EventProcess
           capturedObjects.put(throwable, null);
         }
       }
+    } else {
+      options.getLogger().log(SentryLevel.DEBUG, "Event deduplication is disabled.");
     }
     return event;
   }
--- a/sentry/src/main/java/io/sentry/DuplicateEventDetectionEventProcessor.java
+++ b/sentry/src/main/java/io/sentry/DuplicateEventDetectionEventProcessor.java
@@ -27,21 +27,23 @@ public final class DuplicateEventDetectionEventProcessor implements EventProcess
 
   @Override
   public SentryEvent process(final @NotNull SentryEvent event, final @Nullable Object hint) {
-    final Throwable throwable = event.getOriginThrowable();
-    if (throwable != null) {
-      if (capturedObjects.contains(throwable)
-          || containsAnyKey(capturedObjects, allCauses(throwable))) {
-        options
-            .getLogger()
-            .log(
-                SentryLevel.DEBUG,
-                "Duplicate Exception detected. Event %s will be discarded.",
-                event.getEventId());
-        return null;
-      } else {
-        capturedObjects.add(throwable);
-        if (capturedObjects.size() > bufferSize) {
-          capturedObjects.poll();
+    if (options.isEnableDeduplication()) {
+      final Throwable throwable = event.getOriginThrowable();
+      if (throwable != null) {
+        if (capturedObjects.contains(throwable)
+            || containsAnyKey(capturedObjects, allCauses(throwable))) {
+          options
+              .getLogger()
+              .log(
+                  SentryLevel.DEBUG,
+                  "Duplicate Exception detected. Event %s will be discarded.",
+                  event.getEventId());
+          return null;
+        } else {
+          capturedObjects.add(throwable);
+          if (capturedObjects.size() > bufferSize) {
+            capturedObjects.poll();
+          }
         }
       }
     }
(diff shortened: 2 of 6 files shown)


sentry/src/main/java/io/sentry/DuplicateEventDetectionEventProcessor.java
```diff
@@ -2,26 +2,34 @@
...
     this.options = Objects.requireNonNull(options, "options are required");
+    this.bufferSize = bufferSize;
   }
 
   @Override
   public SentryEvent process(final @NotNull SentryEvent event, final @Nullable Object hint) {
     final Throwable throwable = event.getOriginThrowable();
     if (throwable != null) {
-      if (capturedObjects.containsKey(throwable)
+      if (capturedObjects.contains(throwable)
```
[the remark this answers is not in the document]
maciejwalkowiak: Not sure if I understand. `allCauses` is executed on the exception that is being captured now. We are not looking into causes of all previously captured exceptions.

> And with this refactor, we'll build up a list of 100 and always keep them around, even long after the exception was collected. This means for a long running process we'll always lookup on the full map. Still an improvement in case a service is throwing a ton of exceptions and for some reason it doesn't run GC because it has TB of RAM but does sound suboptimal to other scenarios.

That's true. Do you have a suggestion that would work in all scenarios? We can also store exceptions along with the time they were captured, and once `DuplicateEventProcessor` while iterating for checking for equality, check also if it's not older than X, if so - remove.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 * Enchancement: Support @SentrySpan and @SentryTransaction on classes and interfaces. (#1243)
 * Enchancement: Do not serialize empty collections and maps (#1245)
 * Ref: Simplify RestTemplate instrumentation (#1246)
+* Ref: Optimize DuplicateEventDetectionEventProcessor performance (#1247).
```
marandaneto, line 9: should also go under the `breaking change` section

maciejwalkowiak: It's changing the algorithm that sits under the hood. I am not sure why this would be qualified as a breaking change.

marandaneto: its a behavior change, there's a `bufferSize` now and there is the possibility to report the same exception twice, right?

maciejwalkowiak: Unlikely, only if applications sends hundreds/thousands errors per second. Buffer size is hidden from the user. I can add it to breaking change if you like but I don't believe we should bump a version because of this.

sentry/src/main/java/io/sentry/DuplicateEventDetectionEventProcessor.java
```diff
@@ -2,45 +2,63 @@
...
-        options
-            .getLogger()
-            .log(
-                SentryLevel.DEBUG,
-                "Duplicate Exception detected. Event %s will be discarded.",
-                event.getEventId());
-        return null;
-      } else {
-        capturedObjects.put(throwable, null);
+    if (options.isEnableDeduplication()) {
```
marandaneto, line 30: maybe add an `else` block and log it if it's disabled, we do this for all other integrations

marandaneto (approved):

sentry/src/main/java/io/sentry/DuplicateEventDetectionEventProcessor.java
```diff
@@ -2,45 +2,63 @@
...
-import java.util.WeakHashMap;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import org.jetbrains.annotations.TestOnly;
 
 /** Deduplicates events containing throwable that has been already processed. */
 public final class DuplicateEventDetectionEventProcessor implements EventProcessor {
-  private final WeakHashMap<Throwable, Object> capturedObjects = new WeakHashMap<>();
+  private final ConcurrentLinkedQueue<Throwable> capturedObjects = new ConcurrentLinkedQueue<>();
```
bruno-garcia, line 14: I'm not confortable with this to be honest. Can we discuss this before going ahead?

bruno-garcia (changes requested):

sentry/src/main/java/io/sentry/DuplicateEventDetectionEventProcessor.java
maciejwalkowiak: Sure sure, I was waiting for your blessing anyway :-) We must change the `WeakHashMap` though (https://kdowbecki.github.io/WeakHashMap-is-not-thread-safe/ - this likely contributed to our issues). The option could be to synchronize `WeakHashMap` but this comes with drawbacks too.

maciejwalkowiak: There's no thread safe `WeakHashMap` in Java standard library, but perhaps we can include this dependency? https://github.com/raphw/weak-lock-free

maciejwalkowiak: Or: considering that Deduplication is neccessary only with Spring + logging framework integration, perhaps we can move this event processor to Spring module and use `ConcurrentReferenceHashMap` from Spring Framework. This would solve the problem assuming that only thread safety is an issue.

bruno-garcia: We discussed this on a call and decided to go with @maciejwalkowiak's suggestion of using `Collections.synchronizedMap`.

maciejwalkowiak:
`WeakHashMap` is not thread safe and can [end up in endless loop](https://www.adam-bien.com/roller/abien/entry/endless_loops_in_unsychronized_weakhashmap). We've discussed with @bruno-garcia to synchronize map. This should make the performance issue go away.

maciejwalkowiak:
@marandaneto take a look please at the recent changes. What do you think about this implementation?

sentry/src/test/java/io/sentry/DuplicateEventDetectionEventProcessorTest.kt
```diff
@@ -9,10 +9,22 @@ import kotlin.test.assertNull
...
 class DuplicateEventDetectionEventProcessorTest {
 
-    val processor = DuplicateEventDetectionEventProcessor(SentryOptions())
+    class Fixture {
+        fun getSut(enableDeduplication: Boolean? = null): DuplicateEventDetectionEventProcessor {
+            val options = SentryOptions().apply {
+                if (enableDeduplication != null) {
+                    this.setEnableDeduplication(enableDeduplication)
+                }
+            }
```
marandaneto, line 18: you could init `SentryOptions` outside of `getSut` but still inside of `Fixture` and call options setters directly on each test, I think that's the way we've been doing so far.
`getSut` only takes params if the Sut (eg DuplicateEventDetectionEventProcessor) itself requires those params

marandaneto:
> @marandaneto take a look please at the recent changes. What do you think about this implementation?

looks good to me, left a comment about the testing, its a nit though, feel free to ignore.
update the PR template please, its outdated now.

PR #1248, opened 2021-02-12
SentryExceptionResolver should not send handled errors by default

maciejwalkowiak:
## :scroll: Description

Change the default SentryExceptionResolver behavior to not send errors handled with Spring exception handlers.

I believe current default behavior of `SentryExceptionResolver` is surprising for users - even if they define an exception handler for an exception - to handle it gracefully - it's still being sent to Sentry as an error. This PR changes the default behavior (that can still be overwritten through setting `exception-resolver-order`) to send only unhandled exceptions.

Integration tests

## :pencil: Checklist

- [ ] No breaking changes

## :crystal_ball: Next steps

While it does not break the compilation process, it is a breaking change from the behavior point of view.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,10 @@
 * Enchancement: Support @SentrySpan and @SentryTransaction on classes and interfaces. (#1243)
 * Enchancement: Do not serialize empty collections and maps (#1245)
 * Ref: Simplify RestTemplate instrumentation (#1246)
+
+Breaking Changes:
 * Enchancement: SentryExceptionResolver should not send handled errors by default (#1248).
+
 
 # 4.1.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Enchancement: Support @SentrySpan and @SentryTransaction on classes and interfaces. (#1243)
 * Enchancement: Do not serialize empty collections and maps (#1245)
 * Ref: Simplify RestTemplate instrumentation (#1246)
+* Enchancement: SentryExceptionResolver should not send handled errors by default
 
 # 4.1.0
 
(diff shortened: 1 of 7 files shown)


bruno-garcia (approved):
It's a considerable change in behavior, would be nice to bump a minor here and add a big note on the top of the release notes (the GitHub one)

sentry-spring-boot-starter/src/test/kotlin/io/sentry/spring/boot/it/SentrySpringIntegrationTest.kt
```diff
@@ -137,6 +142,17 @@ class SentrySpringIntegrationTest {
...
+
+    @Test
+    fun `does not send events for handled exceptions`() {
+        val restTemplate = TestRestTemplate().withBasicAuth("user", "password")
+
+        restTemplate.getForEntity("http://localhost:$port/throws-handled", String::class.java)
+
+        await.during(Duration.ofSeconds(2)).untilAsserted {
+            verifyZeroInteractions(transport)
+        }
```
marandaneto, line 154: wouldn't be better to `verify` the `Hub` instead which is a sync call and you wouldn't need to wait up to 2s?

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 * Enchancement: Support @SentrySpan and @SentryTransaction on classes and interfaces. (#1243)
 * Enchancement: Do not serialize empty collections and maps (#1245)
 * Ref: Simplify RestTemplate instrumentation (#1246)
+* Enchancement: SentryExceptionResolver should not send handled errors by default (#1248).
```
marandaneto, line 9: what's about introducing a breaking change section in the changelog when we do such behavior changes?

sentry-spring-boot-starter/src/test/kotlin/io/sentry/spring/boot/it/SentrySpringIntegrationTest.kt
maciejwalkowiak: Good idea!

CHANGELOG.md
maciejwalkowiak: Added a separate section.

marandaneto (approved):

PR #1250, opened 2021-02-15
Fix: Make the ANR Atomic flags immutable

marandaneto:
## :scroll: Description
Fix: Make the ANR Atomic flags immutable

https://stackoverflow.com/questions/66095728/sentry-android-anr-could-there-be-a-false-alarm/66097964?noredirect=1#comment116884716_66097964

could not reproduce the issue not being final, but the behavior is still the same.

## :pencil: Checklist

- [X] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 * Fix: Disable Gson HTML escaping
 * Enchancement: Support @SentrySpan and @SentryTransaction on classes and interfaces. (#1243)
 * Ref: Simplify RestTemplate instrumentation (#1246)
-* Fix: Make the ANR AtomicLong final 
+* Fix: Make the ANR Atomic flags immutable
 
 # 4.1.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Fix: Disable Gson HTML escaping
 * Enchancement: Support @SentrySpan and @SentryTransaction on classes and interfaces. (#1243)
 * Ref: Simplify RestTemplate instrumentation (#1246)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Fix: Disable Gson HTML escaping
 * Enchancement: Support @SentrySpan and @SentryTransaction on classes and interfaces. (#1243)
 * Ref: Simplify RestTemplate instrumentation (#1246)
+* Fix: Make the ANR AtomicLong final
```
marandaneto, line 8:
```suggestion
* Fix: Make the ANR Atomic flags immutable
```

philipphofmann (approved):
LGTM, but these properties are now not immutable. They are final. So maybe we want to change this in the title and the changelog?

marandaneto:
> LGTM, but these properties are now not immutable. They are final. So maybe we want to change this in the title and the changelog?

@philipphofmann not sure if I follow, they are immutable

philipphofmann:
Immutable means that you can't change the value of the object. Final means you can't change the reference. Both `tick` and `reported` are now final, but you can still change their value by using their setters and other methods.

marandaneto:
> Immutable means that you can't change the value of the object. Final means you can't change the reference. Both `tick` and `reported` are now final, but you can still change their value by using their setters and other methods.

ok so to clarify what I want to achieve is the immutable state of assigning new references to the ANR Atomic flags, that's why I meant `ANR Atomic flags` in the title/PR change and not the whole class or its internal fields (which are mutable but atomic).
what would you suggest as a title/changelog?

bruno-garcia (approved):
good catch!

PR #1256, opened 2021-02-16
Test Android Breakpad Minidumps

Swatinem:
## :scroll: Description

Just for testing android breakpad

## :pencil: Checklist

- [ ] No breaking changes

### DIFF
--- a/sentry/src/main/java/io/sentry/OutboxSender.java
+++ b/sentry/src/main/java/io/sentry/OutboxSender.java
@@ -1,25 +1,25 @@
 package io.sentry;
 
-import static io.sentry.SentryLevel.ERROR;
+//import static io.sentry.SentryLevel.ERROR;
 import static io.sentry.cache.EnvelopeCache.PREFIX_CURRENT_SESSION_FILE;
 
 import io.sentry.hints.Flushable;
-import io.sentry.hints.Resettable;
+//import io.sentry.hints.Resettable;
 import io.sentry.hints.Retryable;
-import io.sentry.hints.SubmissionResult;
+//import io.sentry.hints.SubmissionResult;
 import io.sentry.util.CollectionUtils;
 import io.sentry.util.LogUtils;
 import io.sentry.util.Objects;
 import java.io.BufferedInputStream;
-import java.io.BufferedReader;
-import java.io.ByteArrayInputStream;
+//import java.io.BufferedReader;
+//import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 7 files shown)


sentry/src/main/java/io/sentry/OutboxSender.java
```diff
@@ -107,89 +108,97 @@ private void processEnvelope(final @NotNull SentryEnvelope envelope, final @Null
...
+//        ((Resettable) hint).reset();
+//      }
+//    }
+
+    // event and minidump need to be within the same envelope, so right now we are not enriching
+    // the event as a hacky solution, but event is gonna ne symbolicated using the minidump
+    hub.captureEnvelope(envelope, hint);
+    if (!waitFlush(hint)) {
+      return;
     }
```
marandaneto, line 200: we'd need to special case this if we go ahead with mini dumps, as we need to enrich the event but still sending all the left items within the same envelope

sentry/src/main/java/io/sentry/SentryEnvelopeItemHeader.java
```diff
@@ -14,6 +14,7 @@
   private final @NotNull SentryItemType type;
   private final int length;
   @Nullable private final Callable<Integer> getLength;
+  private final @Nullable String attachmentType;
```
marandaneto, line 17: as attachmentType had a default value on backend and Android didn't support any other than the default one, we didn't add it, now we need to as values come from 3rd party SDKs eg sentry-native

marandaneto:
@Swatinem can we close this?

Swatinem:
@marandaneto lets keep this open for now, I would love to be able to play around more with breakpad on Android.

Half of the experiment is adding support for passing through the special minidump attachment, I guess that would be interesting to do independently in a separate PR?

marandaneto:
> @marandaneto lets keep this open for now, I would love to be able to play around more with breakpad on Android.
> 
> Half of the experiment is adding support for passing through the special minidump attachment, I guess that would be interesting to do independently in a separate PR?

attachmentType is now preserved but event and minidump would still be sent separated which is still a problem as it's not going to symbolicated.
I will open a separated issue for that

romtsn:
@Swatinem can we close this now? looks stale for a while

Swatinem:
I don’t think it hurts keeping the branch around. And it might also be a good idea to properly implement this. I know of one customer who wanted to have proper minidumps on Android, with all the different threads in it.

romtsn:
> I don’t think it hurts keeping the branch around. And it might also be a good idea to properly implement this. I know of one customer who wanted to have proper minidumps on Android, with all the different threads in it.

let's keep the branch around but close the PR since it's just noise in the ever-growing list of PRs, we can always reopen later

PR #1260, opened 2021-02-17
Make SentryTransaction#finish thread-safe.

### LINKED ISSUE #1253: Transacrion/Span#finish should be thread-safe
The method `finish` should be thread-safe as it could be called from multiple threads.

cc @Tyrrrz @brustolin

maciejwalkowiak:
## :scroll: Description

Make SentryTransaction#finish thread-safe.

Fixes #1253

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -8,7 +8,6 @@ statusProvider:
 artifactProvider:
   name: github
 targets:
-  - name: bintray
   - name: github
   - name: registry
     type: sdk
@@ -34,6 +33,14 @@ targets:
     type: sdk
     config:
       canonical: "maven:io.sentry:sentry-log4j2"
+  - name: registry
+    type: sdk
+    config:
+      canonical: "maven:io.sentry:sentry-jul"
+  - name: registry
+    type: sdk
+    config:
+      canonical: "maven:io.sentry:sentry-apache-http-client-5"
   - name: registry
     type: sdk
     config:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # vNext
 
+* Fix: Remove experimental annotation for Attachment #1257
 * Fix: Mark stacktrace as snapshot if captured at arbitrary moment #1231
 * Enchancement: Improve EventProcessor nullability annotations (#1229).
 * Bump: sentry-native to 0.4.7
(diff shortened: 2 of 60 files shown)


maciejwalkowiak:
Is this @marandaneto what you meant?

marandaneto:
> Is this @marandaneto what you meant?

I think that's a start, but not the whole thing, if the finish is called from multiple threads it'd be still racey, if we consider that a usage API error, that's ok, if we want to avoid these problems, then we'd need to do more.
let's see what @bruno-garcia says about https://github.com/getsentry/sentry-java/issues/1253#issuecomment-780406961
cus I am a bit lost

bruno-garcia:
"make this thread-safe" gives me the impression we're designing the API to be called concurrently which doesn't make sense for `finish` which should be called once. It would then make more sense to flip the flag atomically so we don't capture the same transaction twice then. The name of this PR could then be:

`ref: use same throwable reference` to avoid confusion on what do we mean by it's making it thread-safe.
We're mutating the scope context without any further synchronization so by looking at this, finish is still "not thread-safe"

marandaneto:
> "make this thread-safe" gives me the impression we're designing the API to be called concurrently which doesn't make sense for `finish` which should be called once. It would then make more sense to flip the flag atomically so we don't capture the same transaction twice then. The name of this PR could then be:
> 
> `ref: use same throwable reference` to avoid confusion on what do we mean by it's making it thread-safe.
> We're mutating the scope context without any further synchronization so by looking at this, finish is still "not thread-safe"

the naming `ref: use same throwable reference` would not be what I meant btw, this is the change that @maciejwalkowiak has changed to fix one of the racey possibilities, we didn't discuss only this field, but the whole method.

See this https://github.com/getsentry/sentry-java/issues/1253#issuecomment-780406961

eg call transaction.finish thread X (transaction is set on the scope, single hub is enabled)
thread Y and thread Z starts new child, finishes child, sets tags, sets throwable, mutate things
transaction.X finishes the execution of the finish method, but while so, the transaction/span state has been mutated by thread Y and Z, this is what we have discussed on the call.

bruno-garcia:
> > "make this thread-safe" gives me the impression we're designing the API to be called concurrently which doesn't make sense for `finish` which should be called once. It would then make more sense to flip the flag atomically so we don't capture the same transaction twice then. The name of this PR could then be:
> > `ref: use same throwable reference` to avoid confusion on what do we mean by it's making it thread-safe.
> > We're mutating the scope context without any further synchronization so by looking at this, finish is still "not thread-safe"
> 
> the naming `ref: use same throwable reference` would not be what I meant btw, this is the change that @maciejwalkowiak has changed to fix one of the racey possibilities, we didn't discuss only this field, but the whole method.
> 
> See this [#1253 (comment)](https://github.com/getsentry/sentry-java/issues/1253#issuecomment-780406961)
> 
> eg call transaction.finish thread X (transaction is set on the scope, single hub is enabled)
> thread Y and thread Z starts new child, finishes child, sets tags, sets throwable, mutate things
> transaction.X finishes the execution of the finish method, but while so, the transaction/span state has been mutated by thread Y and Z, this is what we have discussed on the call.

TO make this thread-safe we'd need to synchronize access to all public methods here and make sure nothing gets mutated while we're running `finish`. This PR doesn't make "finish thread-safe".

maciejwalkowiak:
Considering we are changing the `SentryTransaction` implementation anyway, I am closing this one.

PR #1270, opened 2021-02-18
Add Request to the Scope.

maciejwalkowiak:
## :scroll: Description

Add Request to the Scope.

We need to have a simple way to pass scope to another thread - including the request. Sample scenario:

HTTP request coming to a server application, this application executes a task on the thread from the pool, Sentry event is sent from this thread. There should be an easy way to pass the HTTP request context to this event.

Unit tests.

## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next steps

Refactor Spring and Servlet integrations to set the request on the scope instead of attaching event processor.

### DIFF
--- a/sentry/src/main/java/io/sentry/Scope.java
+++ b/sentry/src/main/java/io/sentry/Scope.java
@@ -309,6 +309,7 @@ public final class Scope implements Cloneable {
     level = null;
     transaction = null;
     user = null;
+    request = null;
     fingerprint.clear();
     breadcrumbs.clear();
     tags.clear();
@@ -510,6 +511,9 @@ public final class Scope implements Cloneable {
 
     final User userRef = user;
     clone.user = userRef != null ? userRef.clone() : null;
+
+    final Request requestRef = request;
+    clone.request = requestRef != null ? requestRef.clone() : null;
 
     clone.fingerprint = new ArrayList<>(fingerprint);
     clone.eventProcessors = new CopyOnWriteArrayList<>(eventProcessors);
--- a/sentry/src/main/java/io/sentry/Scope.java
+++ b/sentry/src/main/java/io/sentry/Scope.java
@@ -1,6 +1,7 @@
 package io.sentry;
 
 import io.sentry.protocol.Contexts;
+import io.sentry.protocol.Request;
 import io.sentry.protocol.User;
 import io.sentry.util.Objects;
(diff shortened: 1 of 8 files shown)


sentry/src/main/java/io/sentry/Scope.java
```diff
@@ -29,6 +30,9 @@
   /** Scope's user */
   private @Nullable User user;
 
+  /** Scope's request */
+  private @Nullable Request request;
```
marandaneto, line 34: missing `clone` and `clear` implementation

sentry/src/main/java/io/sentry/protocol/Request.java
```diff
@@ -159,9 +163,38 @@ public void setOthers(Map<String, String> other) {
...
+   * Clones an User aka deep copy
+   *
+   * @return the cloned User
+   * @throws CloneNotSupportedException if the User is not cloneable
+   */
+  @Override
+  public @NotNull Request clone() throws CloneNotSupportedException {
+    final Request clone = (Request) super.clone();
+
+    clone.headers = CollectionUtils.shallowCopy(headers);
```
marandaneto, line 193: I think `clone.data` could be a List or Map and it might require a shallow copy, I guess we've done this in another clone methods.

marandaneto (approved):
LGTM

sentry/src/main/java/io/sentry/protocol/Request.java
maciejwalkowiak: you mean to check if with `instanceof`? Since `data` can be anything for now it's just a reference copy.

bruno-garcia:
Regarding the description of the PR (motivation): Passing the scope data to another thread is achieved via clink ng the whole hub and passing to another thread, not really just a single scope. The hub has an actual scope stack in it. That's what the unfied design specifies at least.

bruno-garcia (approved):

sentry/src/main/java/io/sentry/protocol/Request.java
marandaneto: yes, but if it's a list/map and items are added/removed, it would mutate both, right, that's what we could avoid.
we did something similar here https://github.com/getsentry/sentry-dart/blob/main/dart/lib/src/protocol/request.dart#L25-L38

PR #1271, opened 2021-02-18
Simplify configuring Logback integration when environment variable with the DSN is not set.

maciejwalkowiak:
## :scroll: Description

Simplify configuring Logback integration when environment variable with the DSN is not set.

When an environment property that is referenced in logback.xml is not set, Logback sets its value to PROPERTY_NAME_IS_UNDEFINED (not a `null` like it could have been expected). This complicates setting up SentryLogback appender configuration where the integration should be disabled in environments that do not have this property.

Before this change, users had to do:

```xml
<options>
  <dsn>${PROPERTY:- }</dsn>
</options>
```

With this change:

```xml
<options>
  <dsn>${PROPERTY}</dsn>
</options>
```

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
+++ b/sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
@@ -48,7 +48,7 @@ public final class SentryAppender extends UnsynchronizedAppenderBase<ILoggingEve
       } else {
         options
             .getLogger()
-            .log(SentryLevel.DEBUG, "DSN is null. SentryLogbackAppender is not being initialized");
+            .log(SentryLevel.WARNING, "DSN is null. SentryAppender is not being initialized");
       }
     }
     super.start();
--- a/sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
+++ b/sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
@@ -35,14 +35,20 @@ public final class SentryAppender extends UnsynchronizedAppenderBase<ILoggingEve
   @Override
   public void start() {
     if (!Sentry.isEnabled()) {
-      options.setEnableExternalConfiguration(true);
-      options.setSentryClientName(BuildConfig.SENTRY_LOGBACK_SDK_NAME);
-      options.setSdkVersion(createSdkVersion(options));
-      Optional.ofNullable(transportFactory).ifPresent(options::setTransportFactory);
-      try {
-        Sentry.init(options);
-      } catch (IllegalArgumentException e) {
-        addWarn("Failed to init Sentry during appender initialization: " + e.getMessage());
+      if (options.getDsn() != null && !options.getDsn().endsWith("_IS_UNDEFINED")) {
+        options.setEnableExternalConfiguration(true);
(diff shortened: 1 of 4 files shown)


sentry-samples/sentry-samples-logback/src/main/resources/logback.xml
```diff
@@ -8,8 +8,9 @@
 
   <appender name="sentry" class="io.sentry.logback.SentryAppender">
     <options>
+      <debug>true</debug>
       <!-- NOTE: Replace the test DSN below with YOUR OWN DSN to see the events from this app in your Sentry project/dashboard -->
-      <dsn>https://502f25099c204a2fbf4cb16edc5975d1@o447951.ingest.sentry.io/5428563</dsn>
+      <dsn>${SENTRY_XXX}</dsn>
```
marandaneto, line 13: should this be reverted?

sentry-logback/src/test/kotlin/io/sentry/logback/SentryAppenderTest.kt
```diff
@@ -305,4 +306,11 @@ class SentryAppenderTest {
             }, anyOrNull())
         }
     }
+
+    @Test
+    fun `does not initialize Sentry when DSN is null`() {
```
marandaneto, line 311: on Android or standard `Sentry.init`, if no DSN is passed (null), we throw an exception, because "" empty DSN means disabled SDK, are we allowing to do differently on logback?

maciejwalkowiak: I fixed the test name, as we are not really handling null here. Logback has this weird behavior, that when the env is not set it gets a value "ENV_NAME_IS_UNDEFINED" and the only way to handle it properly is to write it like `${PROPERTY:- }`.

We can either clarify it in the docs or make it easier for users. This has been roughly discussed with @bruno-garcia - would be nice if he also has a quick look

bruno-garcia (approved):

sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
```diff
@@ -35,14 +35,20 @@
...
+        Optional.ofNullable(transportFactory).ifPresent(options::setTransportFactory);
+        try {
+          Sentry.init(options);
+        } catch (IllegalArgumentException e) {
+          addWarn("Failed to init Sentry during appender initialization: " + e.getMessage());
+        }
+      } else {
+        options
+            .getLogger()
+            .log(SentryLevel.DEBUG, "DSN is null. SentryLogbackAppender is not being initialized");
```
marandaneto, line 51:
```suggestion
            .log(SentryLevel.WARNING, "DSN is null. SentryAppender is not being initialized");
```

marandaneto: I'd log it at least a `WARNING`, class is called `SentryAppender` and not `SentryLogbackAppender`

marandaneto (approved):
left a suggestion, otherwise LGTM

PR #1273, opened 2021-02-18
Fix SentryTransaction#getStatus NPE

### LINKED ISSUE #1268: SentryTransaction.getStatus throws NPE
SentryTransaction.getStatus throws NPE if the transaction is not finished yet.

https://github.com/getsentry/sentry-java/blob/b016da47681036b854fc62d3d141751d8c249a2a/sentry/src/main/java/io/sentry/SentryTransaction.java#L238-L242

it should probably be `context.getStatus()` instead as `this.getContexts().setTrace(this.context);` only happens on `finish`

maciejwalkowiak:
## :scroll: Description

Fix SentryTransaction#getStatus NPE

Fixes #1268.

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/sentry/src/main/java/io/sentry/SentryTransaction.java
+++ b/sentry/src/main/java/io/sentry/SentryTransaction.java
@@ -238,7 +238,8 @@ public final class SentryTransaction extends SentryBaseEvent implements ITransac
   @Override
   @Nullable
   public SpanStatus getStatus() {
-    return this.getContexts().getTrace().getStatus();
+    final SpanContext trace = this.getContexts().getTrace();
+    return trace != null ? trace.getStatus() : null;
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


sentry/src/main/java/io/sentry/SentryTransaction.java
```diff
@@ -238,7 +238,8 @@ Date getTimestamp() {
   @Override
   @Nullable
   public SpanStatus getStatus() {
-    return this.getContexts().getTrace().getStatus();
+    final SpanContext trace = this.getContexts().getTrace();
+    return trace != null ? trace.getStatus() : null;
```
marandaneto, line 242: why not `context.getStatus()`?
I could set a status and check for it, why do I need to wait the transaction to be finished?

maciejwalkowiak: Thanks Manoel, good point.

marandaneto (approved):

PR #1277, opened 2021-02-19
Feat: Auto activity instrumentation

marandaneto:
## :scroll: Description
Feat: Auto activity instrumentation

We'd like to start and finish a transaction automatically during Activity's lifecycle.
It should have a escape hatch so users can finish the transaction manually if they know the best when to finish it (after an API call or IO, etc...)

running on devices

## :pencil: Checklist

- [X] No breaking changes

## :crystal_ball: Next steps
Docs


sentry/src/main/java/io/sentry/SentryTransaction.java
```diff
@@ -238,7 +238,7 @@ Date getTimestamp() {
   @Override
   @Nullable
   public SpanStatus getStatus() {
-    return this.getContexts().getTrace().getStatus();
+    return context.getStatus();
```
marandaneto, line 241: revert after #1268

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -0,0 +1,181 @@
...
+
+  @Override
+  public synchronized void onActivityCreated(
+      @NonNull Activity activity, @Nullable Bundle savedInstanceState) {
+    addBreadcrumb(activity, "created");
+
+    // if activity has global fields being init. and
+    // they are slow, this won't count the whole fields/ctor initialization time, but only
+    // when onCreate is actually called.
+    startTracing(activity);
```
philipphofmann, line 134: `m`: Why not call this in `onActivityPreCreated`?

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -0,0 +1,181 @@
...
+  private void startTracing(final @NonNull Activity activity) {
+    if (performanceEnabled && !isRunningTransaction(activity)) {
+      final ITransaction transaction =
+          hub.startTransaction(getActivityName(activity), "navigation");
+
+      // lets bind to the scope so other integrations can pick it up
+      hub.configureScope(
+          scope -> {
+            // we we'd not like to overwrite existent transactions bound to the Scope manually.
+            if (scope.getTransaction() == null) {
```
philipphofmann, line 99: `h`: I think we will run into trouble with putting transactions on the scope. I think we can't use that to give the users the possibility to finish the transaction manually in the activity. We need another way to store and expose the transaction for the current activity.

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/MainActivity.java
```diff
@@ -146,6 +148,37 @@ protected void onCreate(Bundle savedInstanceState) {
...
+  }
+
+  private void finishTransaction(SpanStatus status) {
+    Sentry.configureScope(
+        scope -> {
+          final ITransaction transaction = scope.getTransaction();
+          if (transaction != null) {
+            transaction.finish(status);
+          }
+        });
```
philipphofmann, line 182: `h`: I really don't like this API. We must find a better way IMO.

philipphofmann (commented):
It's nice to see this going forward. I only addressed important issues. We already talked about some of them in the past, but still, I added comments to not forget about them.

Another idea: We could add spans to measure start with `onActivityPreStarted` and `onActivityPostStarted` and measure resume with `onActivityPreResumed` and `onActivityPostResumed`. We could do this for basically everything. Also, for pause, stop, and destroy.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -0,0 +1,181 @@
...
+  private void stopTracing(final @NonNull Activity activity) {
+    if (performanceEnabled && options.isEnableAutoActivityLifecycleTracingFinish()) {
+      final ITransaction transaction = activities.get(activity);
+      if (transaction != null) {
+        SpanStatus status = transaction.getStatus();
+        // status might be set by other integrations, let's not overwrite it
+        if (status == null) {
+          status = SpanStatus.OK;
+        }
+        transaction.finish(status);
```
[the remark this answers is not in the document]
marandaneto: > It's nice to see this going forward. I only addressed important issues. We already talked about some of them in the past, but still, I added comments to not forget about them.
> 
> Another idea: We could add spans to measure start with `onActivityPreStarted` and `onActivityPostStarted` and measure resume with `onActivityPreResumed` and `onActivityPostResumed`. We could do this for basically everything. Also, for pause, stop, and destroy.

all of these possibilities are written down in our milestones and won't be covered by this PR, first, we need to gather more feedback, see if our code is fast enough to run on each method etc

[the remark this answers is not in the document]
marandaneto: > `h`: Shouldn't we only finish, if the transaction is not finished?

if `isEnableAutoActivityLifecycleTracingFinish` is enabled, we are the only one finishing it, so it wont be never be finished before.

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/MainActivity.java
philipphofmann: What about having something like `Sentry.finishTransaction(this)`? We do the magic internally, by somehow calling the integration and then finish the transaction. Of course, we need to make sure to not create memory leaks.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -0,0 +1,181 @@
...
+      }
+    }
+  }
+
+  @Override
+  public synchronized void onActivityCreated(
+      @NonNull Activity activity, @Nullable Bundle savedInstanceState) {
+    addBreadcrumb(activity, "created");
```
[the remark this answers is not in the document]
marandaneto: > `m`: Why not keep the `ActivityBreadcrumbsIntegration` just for the breadcrumbs and have this integration here just for the transactions? Now the logic is mixed in this class.

why would we have the overhead of one more callback to be called if we can do it here? it's trade-off and we don't want to overload somebody App's

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
marandaneto: this is called already `before` the App's onCreate, so its not necessary.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
marandaneto: > `h`: I think we will run into trouble with putting transactions on the scope. I think we can't use that to give the users the possibility to finish the transaction manually in the activity. We need another way to store and expose the transaction for the current activity.

what type of troubles? can you elaborate?

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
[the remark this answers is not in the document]
philipphofmann: We put the transaction on the Scope, so some other code could finish it before we do. Or am I missing something?

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/MainActivity.java
marandaneto: 100% agreed and already discussed, see: https://github.com/getsentry/sentry-java/blob/feat/auto-activity-tracing/sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/MainActivity.java#L178

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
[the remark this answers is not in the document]
marandaneto: then it's misusing the API, they should disable the `enableAutoActivityLifecycleTracingFinish` and do themselves, they will probably be finishing spans, not the transaction (root span)

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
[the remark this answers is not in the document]
philipphofmann: I don't think the overhead is big. We only add an extra object to the heap. The callbacks are a simple for loop. see
```Java
private void dispatchActivityPreCreated(@android.annotation.Nullable Bundle savedInstanceState) {
  getApplication().dispatchActivityPreCreated(this, savedInstanceState);
  Object[] callbacks = collectActivityLifecycleCallbacks();
  if (callbacks != null) {
    for (int i = 0; i < callbacks.length; i++) {
      ((Application.ActivityLifecycleCallbacks) callbacks[i]).onActivityPreCreated(this,
              savedInstanceState);
    }
  }
}
```
I don't have a strong opinion on this one. Just seems strange to me to mix two features in one class while the overhead wouldn't be big.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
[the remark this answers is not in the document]
philipphofmann: But I think this API is very easy to misuse then. Why not let the user have `enableAutoActivityLifecycleTracingFinish ` and still give him the possibility to finish the transaction manually? We just would need to check here if the transaction is not finished yet.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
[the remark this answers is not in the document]
marandaneto: I'd prefer to have a single callback for the sake of optimization.
this loop happens for each method of the lifecycle btw, always on the UI thread.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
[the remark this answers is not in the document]
marandaneto: that's a fair point, a check won't kill, will do.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
philipphofmann: I don't think so. When taking a look at the source code of Activity, there is a lot of stuff happening before `onActivityCreated` is called, like restoring fragments, stuff with `savedInstanceState`, etc. This could also be slow and I think it would be good to include that.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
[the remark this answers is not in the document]
philipphofmann: I think it doesn't matter. I leave this up to you. Again no strong opinion.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
marandaneto: the user's code is only executed after this method, not before, except the global fields as stated before.

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/MainActivity.java
marandaneto: @philipphofmann what would be `this`? the `Activity` Context? the way how to access the transaction is thru the Scope, the only thing that comes to my mind is via the Static API like JS/Python does, Sentry.getSpan() but not accessing an integration that has a WeakRef ma, this would also make racey as it could be called from different threads during Screen's lifecycle.
I didn't spend time thinking about this yet, the new API should also be compatible with #1286 which makes a bit harder as parts are moving as we speak

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -0,0 +1,181 @@
...
+        .log(
+            SentryLevel.DEBUG,
+            "ActivityBreadcrumbsIntegration enabled: %s",
+            this.options.isEnableActivityLifecycleBreadcrumbs());
+
+    performanceEnabled = isPerformanceEnabled(this.options);
+
+    if (this.options.isEnableActivityLifecycleBreadcrumbs() || performanceEnabled) {
+      application.registerActivityLifecycleCallbacks(this);
+      options.getLogger().log(SentryLevel.DEBUG, "ActivityBreadcrumbsIntegration installed.");
```
bruno-garcia, line 58:
```suggestion
    if (this.options.isEnableActivityLifecycleBreadcrumbs() || performanceEnabled) {
      application.registerActivityLifecycleCallbacks(this);
      options.getLogger().log(SentryLevel.DEBUG, "ActivityLifecycleIntegration installed.");
```

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -0,0 +1,181 @@
...
+  private boolean isPerformanceEnabled(final @NotNull SentryAndroidOptions options) {
+    return ((options.getTracesSampleRate() != null || options.getTracesSampler() != null)
+        && options.isEnableAutoActivityLifecycleTracing());
+  }
+
+  @Override
+  public void close() throws IOException {
+    application.unregisterActivityLifecycleCallbacks(this);
+
+    options.getLogger().log(SentryLevel.DEBUG, "ActivityBreadcrumbsIntegration removed.");
```
bruno-garcia, line 71:
```suggestion
    options.getLogger().log(SentryLevel.DEBUG, "ActivityLifecycleIntegration removed.");
```

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
bruno-garcia: If we don't put the transaction on the scope, user code or our own integrations can't find it and no spans will be added to it.

Reading and writing isn't an atomic operation here and scope is accessed concurrently.
One alternative is a single method that can do the condition and reference swap atomically like `scope.setIfNull(transaction)` but the fact we got to this situation is a sign we probably need to discuss this further. In other SDKs/general Sentry unified API a scope is part of a hub which is usually accessed by a single thread (thread local storage). It must be thread-safe as it can be accessed concurrently particularly when it's being cloned to make room for a new instance bound to a new thread, but it's generally not being mutated by multiple threads at the same time. This isn't true for the mobile SDKs since we wanted to have this global state for breadcrumbs and tags etc.

For transactions this seems not to fit very well so we might need to discuss what to do next here.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -0,0 +1,181 @@
...
+
+  private final @NotNull Application application;
+  private @NotNull IHub hub;
+  private @NotNull SentryAndroidOptions options;
+
+  private boolean performanceEnabled = false;
+
+  // WeakHashMap isn't thread safe but ActivityLifecycleCallbacks is only called from the
+  // main-thread
+  private final @NotNull WeakHashMap<Activity, ITransaction> activities = new WeakHashMap<>();
```
bruno-garcia, line 32: Could have a more descriptive name:

```suggestion
  private final @NotNull WeakHashMap<Activity, ITransaction> activitiesWithOngoingTransactions = new WeakHashMap<>();
```

^ suggestions is maybe too descriptive but the goal is to describe what this is about

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  @Override
+  public synchronized void onActivitySaveInstanceState(
+      @NonNull Activity activity, @NonNull Bundle outState) {
+    addBreadcrumb(activity, "saveInstanceState");
+  }
+
+  @Override
+  public synchronized void onActivityDestroyed(@NonNull Activity activity) {
```
bruno-garcia, line 178: Could this be called instead of `onActivityPostResumed` leaving an open transaction hanging?

Such edge case makes me realize we probably want to implement 1k (or other value) max Span and start dropping it because if we end up with a transaction that isnt' finished, we could add spans to it indefinitely. Another edge case is a transaction that's finished but still bound to the scope and the SDK allows adding spans to finished transactions. We need to disallow that and log out a warning.

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/NetworkInterceptor.kt
```diff
@@ -0,0 +1,21 @@
...
+import okhttp3.Response
+
+class NetworkInterceptor : Interceptor {
+    override fun intercept(chain: Interceptor.Chain): Response {
+        val request = chain.request()
+
+        // read transaction from the bound scope
+        val span = Sentry.getSpan()?.startChild("http.client", request.url().toString())
+
+        val response = chain.proceed(request)
```
bruno-garcia, line 15: I believe we need a try catch here in case proceed actually throws.

bruno-garcia (commented):
The activity stuff looks good but using the API here surfaced a few edge cases we need to discuss further like potentially adding TLS for performance alone for mobile.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
marandaneto: that's a good point, we do it atomically for `startSession` for example, maybe this would be a workaround for making it work now.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
marandaneto: not sure if I got it, onActivityPostResumed means the App went foreground.
`onActivityDestroyed` means App went background.

if you meant `onActivityStopped` instead of `onActivityPostResumed`, the answer is no, `onActivityStopped` always run before `onActivityDestroyed` hence a transaction will never be hanging unless the user opt-out `enableAutoActivityLifecycleTracingFinish` and didn't finish the transaction manually and never closed the screen either, we could be sure that the transaction is ended at `onActivityDestroyed` even if `enableAutoActivityLifecycleTracingFinish` was disabled.

marandaneto: I will end the transaction regardely if `enableAutoActivityLifecycleTracingFinish` is disabled, on `onActivityDestroyed`

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/NetworkInterceptor.kt
marandaneto: no, Retrofit implements its own Error handler, it bails to the `onFailure` callback, and this is part of our sample, not SDK

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -0,0 +1,190 @@
...
+
+  private boolean isRunningTransaction(final @NonNull Activity activity) {
+    return activitiesWithOngoingTransactions.containsKey(activity);
+  }
+
+  private void stopTracing(
+      final @NonNull Activity activity, final boolean enableAutoActivityLifecycleTracingFinish) {
+    if (performanceEnabled && enableAutoActivityLifecycleTracingFinish) {
+      final ITransaction transaction = activitiesWithOngoingTransactions.get(activity);
+      if (transaction != null && !transaction.isFinished()) {
```
marandaneto, line 117: ideally `finish` does it atomically, so we don't fish twice, use AtomicBoolean and flip it

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -0,0 +1,190 @@
...
+  private void startTracing(final @NonNull Activity activity) {
+    if (performanceEnabled && !isRunningTransaction(activity)) {
+      final ITransaction transaction =
+          hub.startTransaction(getActivityName(activity), "navigation");
+
+      // lets bind to the scope so other integrations can pick it up
+      hub.configureScope(
+          scope -> {
+            // we we'd not like to overwrite existent transactions bound to the Scope manually.
+            if (scope.getTransaction() == null) {
```
marandaneto, line 100: this should go to the docs, its a special case.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
bruno-garcia: Makes sense, thanks for clarifying.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
marandaneto: added `onActivityPreCreated` for thr sake of correctness

sentry/src/main/java/io/sentry/Scope.java
```diff
@@ -140,7 +144,13 @@ public ISpan getSpan() {
    * @param transaction the transaction
    */
   public void setTransaction(final @NotNull ITransaction transaction) {
-    this.transaction = Objects.requireNonNull(transaction, "transaction is required");
+    synchronized (transactionLock) {
+      this.transaction = Objects.requireNonNull(transaction, "transaction is required");
+    }
+
+    // TODO: double-check this
+    // setTransaction(String) replaces the transaction.name so I guess this makes sense too
```
marandaneto, line 152: @bruno-garcia @maciejwalkowiak opinions?

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -0,0 +1,198 @@
...
+        .log(
+            SentryLevel.DEBUG,
+            "ActivityLifecycleIntegration enabled: %s",
+            this.options.isEnableActivityLifecycleBreadcrumbs());
+
+    performanceEnabled = isPerformanceEnabled(this.options);
+
+    if (this.options.isEnableActivityLifecycleBreadcrumbs() || performanceEnabled) {
+      application.registerActivityLifecycleCallbacks(this);
+      options.getLogger().log(SentryLevel.DEBUG, "ActivityLifecycleIntegration installed.");
```
philipphofmann, line 59: `l`
```suggestion
      this.options.getLogger().log(SentryLevel.DEBUG, "ActivityLifecycleIntegration installed.");
```

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -0,0 +1,198 @@
...
+    if (performanceEnabled && !isRunningTransaction(activity)) {
+      final ITransaction transaction =
+          hub.startTransaction(getActivityName(activity), "navigation");
+
+      // lets bind to the scope so other integrations can pick it up
+      hub.configureScope(
+          scope -> {
+            scope.withTransaction(
+                scopeTransaction -> {
+                  // we we'd not like to overwrite existent transactions bound to the Scope
```
philipphofmann, line 101: `l`
```suggestion
                  // we'd not like to overwrite existent transactions bound to the Scope
```

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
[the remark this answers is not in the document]
philipphofmann: > that's a fair point, a check won't kill, will do.

I think this point is not addressed yet.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -0,0 +1,198 @@
...
+  @Override
+  public synchronized void onActivityPaused(final @NonNull Activity activity) {
+    addBreadcrumb(activity, "paused");
+  }
+
+  @Override
+  public synchronized void onActivityPostPaused(final @NonNull Activity activity) {
+    // in case people opt-out enableAutoActivityLifecycleTracingFinish and forgot to finish it,
+    // we do it automatically before moving to a new Activity.
+    stopTracing(activity, true);
```
philipphofmann, line 173: `h`: If someone gets a call after 2 seconds being on the activity, this is going to produce wrong statistics, I guess. We don't know here if someone navigates to a different activity or puts the app in the background. Maybe we need some kind of strategy as we have it already for sessions. Store it somewhere, but only end it if the user doesn't come back.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -0,0 +1,198 @@
...
+    stopTracing(activity, options.isEnableAutoActivityLifecycleTracingFinish());
+  }
+
+  @Override
+  public synchronized void onActivityPaused(final @NonNull Activity activity) {
+    addBreadcrumb(activity, "paused");
+  }
+
+  @Override
+  public synchronized void onActivityPostPaused(final @NonNull Activity activity) {
```
philipphofmann, line 170: `h`: I think this should be moved to onActivityStopped. OnPause is also called when you display a dialog. It just states the activity has no focus anymore, but it could be still visible.

philipphofmann (commented):
One of the biggest problems for me right now is that with the current approach, I think we end the transaction when the user puts the app into the background. I also added a comment for this.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
marandaneto: we can't, when you start a new activity, the new activity's onCreate is called before the current's activity onStop, it was like this before, onActivityPostPaused is the solution

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
marandaneto: we are doing performance, not analytics, so it's fair to finish it if not yet.
I was not finishing before and @bruno-garcia brought some uses cases that would make sense to do it (like non finished transactions, etc)

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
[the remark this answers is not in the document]
marandaneto: it is, looks at the implementation of the method finish()

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
marandaneto: I'm fine either way, both use cases has pros/cons, its a best effort

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
philipphofmann: Oh, I didn't think about that. But is it fine to finish the transaction when you display a dialog?

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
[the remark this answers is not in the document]
philipphofmann: Ah sorry, didn't see that you changed it there.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
philipphofmann: But what's the big deal calling this after the `onCreate` of the other activity is called? 

Another important point is for multi-window activities. The activity could be paused, but still visible, see https://developer.android.com/guide/topics/ui/multi-window#lifecycle. I know this is an edge case, but finishing the transaction on `onPause` feels a bit weird to me.

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/MainActivity.java
marandaneto: this is now obsolete by `Sentry.getSpan().finish(...)`

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
marandaneto: @philipphofmann we have a single Hub and Scope, which takes only one transaction per time.
If I start a transaction before finishing the previous one, its not gonna set the new transaction to the Scope.
If we always overwrite the current transaction set to the Scope, its also a problem, as they wont be able to finish themselves as the transaction is not accessible anymore.

multi-window is indeed an issue, but I'd prefer to tackle this problem at a later problem, with many other problems that would come, probably.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -0,0 +1,198 @@
...
+      // lets bind to the scope so other integrations can pick it up
+      hub.configureScope(
+          scope -> {
+            scope.withTransaction(
+                scopeTransaction -> {
+                  // we we'd not like to overwrite existent transactions bound to the Scope
+                  // manually.
+                  if (scopeTransaction == null) {
+                    scope.setTransaction(transaction);
+                  }
```
bruno-garcia, line 105: worth a log.debug("Activity transaction won't be bound since {scope.transaction.name} is alread ythere?

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
```diff
@@ -33,6 +37,48 @@
...
+  /**
+   * Enables the Auto instrumentation for Activity lifecycle tracing, but specifically when to
+   * finish the transaction, read {@link SentryAndroidOptions#enableAutoActivityLifecycleTracing}.
+   *
+   * <p>If you require a specific lifecycle to finish a transaction or even after the Activity is
+   * fully rendered but still waiting for an IO operation, you could call {@link ISpan#finish()}
+   * yourself on {@link Sentry#getSpan()}, be sure that you've finished all of your manually created
+   * Spans.
+   */
+  private boolean enableAutoActivityLifecycleTracingFinish = false;
```
bruno-garcia, line 80: This means the user must call `finish` manually?
Shouldn't this be `true` by default?

sentry-samples/sentry-samples-android/src/main/AndroidManifest.xml
```diff
@@ -72,6 +74,15 @@
...
         <meta-data android:name="io.sentry.session-tracking.timeout-interval-millis" android:value="10000" />
 
+        <!--    how to enable the performance API by setting a sample-rate-->
+        <meta-data android:name="io.sentry.traces.sample-rate" android:value="1.0" />
+
+        <!--    how to disable the Activity auto instrumentation for tracing-->
+        <!--    <meta-data android:name="io.sentry.traces.activity.enable" android:value="false" />-->
+
+        <!--    how to disable the Activity auto instrumentation for tracing (finishing automatically)-->
+        <!--    <meta-data android:name="io.sentry.traces.activity.finish.enable" android:value="false" />-->
```
bruno-garcia, line 84: For clarity here I suggest a change in the name of the options:

```suggestion
        <!--    how to disable the Activity auto instrumentation automatically finished on onClose -->
        <!--    <meta-data android:name="io.sentry.traces.activity.auto-finish.enable" android:value="false" />-->
```

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/MainActivity.java
```diff
@@ -146,6 +147,13 @@ protected void onCreate(Bundle savedInstanceState) {
           }
         });
 
+    binding.openSecondActivity.setOnClickListener(
+        view -> {
+          // finishing so its completely destroyed
+          finish();
```
bruno-garcia, line 153: Is this needed? As a developer reading this code, should I be using this pattern on my app?

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/NetworkInterceptor.kt
```diff
@@ -0,0 +1,31 @@
...
+import io.sentry.Sentry
+import io.sentry.SpanStatus
+import java.io.IOException
+import okhttp3.Interceptor
+import okhttp3.Response
+
+class NetworkInterceptor : Interceptor {
+    override fun intercept(chain: Interceptor.Chain): Response {
+
+        val request = chain.request()
```
bruno-garcia, line 12: Does this line have side effect or is it just a getter?
If this could potentially be slow we could consider getting a timestamp before calling it and setting the transaction timestamp to that explicitly.

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/SecondActivity.kt
```diff
@@ -0,0 +1,112 @@
...
+        service.listRepos(binding.editRepo.text.toString()).enqueue(object : Callback<List<Repo>> {
+            override fun onFailure(call: Call<List<Repo>>?, t: Throwable) {
+                span.finish(SpanStatus.INTERNAL_ERROR)
+                Sentry.captureException(t)
+
+                showText(true, "error: ${t.message}")
+
+                // be sure to finish all your spans before this
+                val transaction = Sentry.getSpan()
+                transaction?.finish(SpanStatus.INTERNAL_ERROR)
```
bruno-garcia, line 94: Is this the place to finish the root span?
In case of the activity auto transaction wouldn't we be closing that prematurely?

There's time running from when we handle the HTTP failure and the user interaction like showing an error message so I believe we shouldn't close the transaction here.

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/SecondActivity.kt
```diff
@@ -0,0 +1,112 @@
...
+
+                span.finish(SpanStatus.OK)
+
+                showText(text = "items: ${repos.size}")
+
+                // I opt out enableAutoActivityLifecycleTracingFinish so I when best when to end my transaction
+
+                // be sure to finish all your spans before this
+                val transaction = Sentry.getSpan()
+                transaction?.finish(SpanStatus.OK)
```
bruno-garcia, line 108: Same as above

sentry/src/main/java/io/sentry/Scope.java
bruno-garcia: When we phase out `Transaction` being on the scope this would introduce a breaking change since creating a root span with `startTransaction` would now require also changing the scope's `transaction` property with the given string. No strong opinions here but this seems like a can of warms.

bruno-garcia (commented):
Left some comments but generally lgtm, can see a preview cooking already

sentry/src/main/java/io/sentry/Scope.java
marandaneto: I will keep as it is then

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
marandaneto: good point

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
marandaneto: yes, it should, I was testing and commit/push wrongly, thx

sentry-samples/sentry-samples-android/src/main/AndroidManifest.xml
marandaneto: yeah, makes sense, renamed the options too to `enableActivityLifecycleTracingAutoFinish`

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/NetworkInterceptor.kt
marandaneto: its just a getter

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/SecondActivity.kt
marandaneto: not sure if I follow your comment.

`updateRepos` is called during Activity.onResume which means the Activity is responsive and rendered.
if the API call returned a result or error, that's the end of my transaction, this is a sample app, I don't have any other screen to go nor a dialog to be shown.

the error message is updated via `showText` just before finishing the transaction.

@bruno-garcia can you elaborate?

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
philipphofmann: What about moving this to `onDestroy` and before starting a new transaction on `onCreate` we check if there is still an active transaction of another activity and just finish it?

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
marandaneto: consider iterating over the activity weakref list and finish pending transactions instead of finishing during onPause before creating the new Activity, this would give a bit more time to the user to finish the transaction than just finishing on onPause, this would solve the problem if the user gets a call.

marandaneto: this method call should stay on onStop or onDestroy

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
marandaneto: discussed along with https://github.com/getsentry/sentry-java/pull/1277#discussion_r585417497

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/MainActivity.java
marandaneto: yes, I wanted a cold start

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
marandaneto: onDestroy because if you get a call from somebody, App goes to onStop as well

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/SecondActivity.kt
```diff
@@ -0,0 +1,112 @@
...
+        binding.text.text = if (visible) text else ""
+        binding.text.visibility = if (visible) View.VISIBLE else View.GONE
+
+        binding.indeterminateBar.visibility = if (visible) View.GONE else View.VISIBLE
+    }
+
+    override fun onResume() {
+        super.onResume()
+
+        val span = Sentry.getSpan()?.startChild("onResume", javaClass.simpleName)
```
philipphofmann, line 67: `m`: Isn't this something we should do, and you shouldn't have to do manually?

philipphofmann (commented):
This gets better and better. Do you plan to add tests for the preview? This would be the main thing missing for approval.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -88,8 +89,19 @@ private void addBreadcrumb(final @NonNull Activity activity, final @NotNull Stri
     return activity.getClass().getSimpleName();
   }
 
+  private void stopPreviousTransactions() {
+    for (final Map.Entry<Activity, ITransaction> entry :
+        activitiesWithOngoingTransactions.entrySet()) {
+      final ITransaction transaction = entry.getValue();
+      finishTransaction(transaction);
+    }
```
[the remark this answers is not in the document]
marandaneto: `values` creates a new Collection and `clear` iterates the whole `Map` again, I'd not go this route

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/SecondActivity.kt
marandaneto: we already discussed this in a call, ideally, yes, not now, this is a sample activity

marandaneto:
> This gets better and better. Do you plan to add tests for the preview? This would be the main thing missing for approval.

I already have a few tests locally and will make sure it's fully tested as discussed in the call, this is just early feedback for the behavioral feature.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -0,0 +1,217 @@
...
+
+      // lets bind to the scope so other integrations can pick it up
+      hub.configureScope(
+          scope -> {
+            scope.withTransaction(
+                scopeTransaction -> {
+                  // we'd not like to overwrite existent transactions bound to the Scope
+                  // manually.
+                  if (scopeTransaction == null) {
+                    scope.setTransaction(transaction);
```
bruno-garcia, line 116: Proposing @maciejwalkowiak uses this API and just realized someone could call `setTransaction` outside of this `withTransaction` hence bypassing the synchronization block.

Can we get rid of `get/setTransaction`?

marandaneto: @bruno-garcia what's the problem? they are all guarded by a lock -> `transactionLock`

bruno-garcia: I missed that. And the lock being reentrant this should work fine!

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/SecondActivity.kt
bruno-garcia: > the error message is updated via showText just before finishing the transaction.

Would this be the place to finish the transaction then?

I believe it would be at the very end of the transaction as in: Activity loaded, we queried the backend, got an error from it, checked the status code, prepared some error message and showed the dialog to the user: Finish.

But this is a sample and not a reason to block this PR so we can discuss this offline.

bruno-garcia (approved):

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -0,0 +1,231 @@
...
+      final ITransaction transaction = activitiesWithOngoingTransactions.get(activity);
+      finishTransaction(transaction);
+    }
+  }
+
+  private void finishTransaction(final @Nullable ITransaction transaction) {
+    if (transaction != null) {
+      SpanStatus status = transaction.getStatus();
+      // status might be set by other integrations, let's not overwrite it
+      if (status == null) {
```
maciejwalkowiak, line 155: If status can be set in other threads perhaps status could be also an atomic reference and `compareAndSet` could be used.

maciejwalkowiak (approved):

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
marandaneto: that's a good point

PR #1278, opened 2021-02-19
Fix: SentryTransaction#finish should not clear another transaction from the scope

### LINKED ISSUE #1269: SentryTransaction.finish() clears transaction from the Scope even if it was not bound to the Scope at all
SentryTransaction.finish() clears transaction from the Scope even if it was not bound to the Scope at all, that means, we'd be clearing a transaction bound to the Scope that might not be the same transaction at all.

https://github.com/getsentry/sentry-java/blob/1431c5e09734014084506c8a9aad98a8bdf959dc/sentry/src/main/java/io/sentry/Hub.java#L558-L560

I think before clearing it up, we'd need to check if it's actually the very same running transaction, this might be a problem only on Mobile though because of the single Hub thingie

maciejwalkowiak:
## :scroll: Description

SentryTransaction#finish should not clear another transaction from the scope

Fixes #1269

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/sentry/src/main/java/io/sentry/Hub.java
+++ b/sentry/src/main/java/io/sentry/Hub.java
@@ -556,7 +556,10 @@ public final class Hub implements IHub {
                   e);
         } finally {
           if (item != null) {
-            item.getScope().clearTransaction();
+            final Scope scope = item.getScope();
+            if (scope.getTransaction() == transaction) {
+              scope.clearTransaction();
+            }
           }
         }
       }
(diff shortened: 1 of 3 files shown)


marandaneto (approved):
LGTM
changelog though

sentry/src/main/java/io/sentry/Hub.java
```diff
@@ -556,7 +556,10 @@ public void flush(long timeoutMillis) {
                   e);
         } finally {
           if (item != null) {
-            item.getScope().clearTransaction();
+            final Scope scope = item.getScope();
+            if (scope.getTransaction() == transaction) {
+              scope.clearTransaction();
```
bruno-garcia, line 561: On Android where the scope is shared this would race since in between another transaction could have been set.

Another approach would be:

`scope.clear(transaction)`
And internal conditional atomic reference swap (i.e: https://docs.microsoft.com/en-us/dotnet/api/system.threading.interlocked.compareexchange?view=net-5.0)

PR #1286, opened 2021-02-23
Remove SentryTransaction from public API

maciejwalkowiak:
## :pencil: Checklist

- [ ] No breaking changes


maciejwalkowiak:
Spans are collected with SentryTracer class. The only interface user is dealing with is `ISpan`.

@marandaneto @bruno-garcia please take a look for early feedback.

maciejwalkowiak:
In addition to that, in such setup I would like to rename `Span` to `SentrySpan` and `ISpan` to `Span`.

sentry/src/main/java/io/sentry/Hub.java
```diff
@@ -516,8 +516,8 @@ public void flush(long timeoutMillis) {
   @ApiStatus.Internal
   @Override
   public @NotNull SentryId captureTransaction(
-      final @NotNull ITransaction transaction, final @Nullable Object hint) {
-    Objects.requireNonNull(transaction, "transaction is required");
+      final @NotNull SentryTracer tracer, final @Nullable Object hint) {
```
bruno-garcia, line 519: I was thinking we'd keep `captureTransaction(Transaction ..)`
`SentryTracer.finish` would in fact do:

```
var tran = new SentryTransaction() { ..tags = self.tags }
hub.captureTransaction(tran);
```

sentry/src/main/java/io/sentry/ISpan.java
```diff
@@ -5,6 +5,22 @@
...
 /** Represents performance monitoring Span. */
 public interface ISpan {
+  String NAME_TAG = "sentry-name";
+
+  /**
+   * Sets the span name.
+   *
+   * @param name - span name
+   */
+  void setName(@Nullable String name);
```
bruno-garcia, line 15: I propose we don't have these method on Span since this is a Transaction thing only.

The idea of a Transaction to get a name is either through `startTransaction(name)` in which case whatever object instance it creates knows that it's in fact a Transaction (i.e when `finish` is called on that object, a wire-protocol SentryTransaction is created and values copied into it, including the name).

For the edge case of setting a name after the fact (or changing the name) we could have a special-cased tag called: `sentry-transaction-name` and if that's set we'll use that as the name instead of whatever was given on `startTransaction` as a way to allow overriding the name after the fact.

bruno-garcia (commented):
half way through pass

marandaneto:
@maciejwalkowiak please revert `sentry-android-ndk/sentry-native ` changes.

sentry/src/main/java/io/sentry/ISpan.java
marandaneto: is there an use case to make the name mutable? I'd make it immutable and only taking on startTransaction

sentry/src/main/java/io/sentry/Hub.java
marandaneto: yeah me too

marandaneto: is there a reason to take `SentryTracer` instead of the `Transaction` itself? the reason is, the reason I ask is, we'd not need ot change SentryClient

sentry/src/main/java/io/sentry/ISpan.java
```diff
@@ -118,10 +134,29 @@
...
+  @Nullable
+  Boolean isSampled();
+
+  /**
+   * Returns the last active span within the span hierarchy
+   *
+   * @return the last active span of {@code null} if not found
+   */
+  @Nullable
+  ISpan getLatestActiveSpan();
```
marandaneto, line 161: why would we need this in the ISpan interface?

sentry/src/main/java/io/sentry/Scope.java
```diff
@@ -122,25 +122,22 @@ public void setTransaction(final @NotNull String transaction) {
    * @return current active Span or Transaction or null if transaction has not been set.
    */
   @Nullable
-  public ISpan getSpan() {
-    final ITransaction tx = transaction;
+  public ISpan getLatestActiveSpan() {
```
marandaneto, line 125: is `Scope. getSpan()` totally removed? should we keep the same name at least?

sentry/src/main/java/io/sentry/Scope.java
```diff
@@ -300,14 +297,14 @@ public void clearTransaction() {
    * @return the transaction
    */
   @Nullable
-  public ITransaction getTransaction() {
-    return this.transaction;
+  public ISpan getSpan() {
```
marandaneto, line 300: so never about about https://github.com/getsentry/sentry-java/pull/1286/files#r583536154

the thing is, what previously was returning the latest active Span is returning the root span, same method but different behavior, I guess this would cause a lot of problems, people might fix the compilation error due to return type, but it'll mess up their span hierarchy.

settings.gradle.kts
```diff
@@ -1,24 +1,25 @@
...
+//    "sentry-spring-boot-starter",
+//    "sentry-android-timber",
+//    "sentry-samples:sentry-samples-android",
+    "sentry-samples:sentry-samples-console")
+//    "sentry-samples:sentry-samples-jul",
+//    "sentry-samples:sentry-samples-log4j2",
+//    "sentry-samples:sentry-samples-logback",
+//    "sentry-samples:sentry-samples-servlet",
+//    "sentry-samples:sentry-samples-spring",
+//    "sentry-samples:sentry-samples-spring-boot")
```
marandaneto, line 25: oh boy, I get you :D

sentry/src/main/java/io/sentry/Hub.java
maciejwalkowiak: We need to clear the transaction from the scope on capture, since on the scope we will have instance of a Span (likely SentryTracer), not a transaction we need to pass the original captured transaction.

sentry/src/main/java/io/sentry/ISpan.java
maciejwalkowiak: Either this or we use `instance of` in Scope:

```
public ISpan getLatestActiveSpan() {
  final ISpan tx = span;
  if (tx instanceof SentryTracer) {
    return ((SentryTracer) tx).getLatestActiveSpan();
  } else {
    return tx;
  }
}
```

sentry/src/main/java/io/sentry/ISpan.java
maciejwalkowiak: In web frameworks, route may not be known when the transaction is started as the components that resolve the route from HTTP request have not run yet.

sentry/src/main/java/io/sentry/ISpan.java
maciejwalkowiak: When I look at it now, as much as I don't like using `instanceof` i think in this case it will be a lesser evil.

maciejwalkowiak:
Replaced by #1304

PR #1292, opened 2021-02-25
Fix: Craft artifact provider

marandaneto:
Craft publish requires no artifact provider if we want only to publish to release-registry, command:

eg 4.2.0

> craft publish 4.2.0 --rev 4.2.0 --target registry

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -1,4 +1,4 @@
-minVersion: "0.13.2"
+minVersion: "0.18.0"
 github:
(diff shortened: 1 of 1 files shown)


tonyo (approved):
Note that `craft` will also assume certain things for the changelog management (https://github.com/getsentry/craft#changelog-policies) when running `craft prepare`, so it's worth trying some dry runs to make sure that things are aligned.

PR #1304, opened 2021-03-04
Remove SentryTransaction from public API

maciejwalkowiak:
## :scroll: Description

Remove SentryTransaction from public API & use separate classes for user facing & protocol operations in the Performance feature.

Unit tests.

## :pencil: Checklist

- [x] No breaking changes


sentry-spring/src/main/java/io/sentry/spring/tracing/SentryTracingFilter.java
```diff
@@ -73,6 +73,11 @@ protected void doFilterInternal(
     if (hub.isEnabled()) {
       final String sentryTraceHeader = httpRequest.getHeader(SentryTraceHeader.SENTRY_TRACE_HEADER);
 
+      hub.configureScope(
```
maciejwalkowiak, line 76: `ITransaction#setRequest` is deprecated now as these kind of properties should be set on the scope and then applied to transaction in `SentryClient`.

sentry/src/main/java/io/sentry/SentryClient.java
```diff
@@ -424,6 +431,16 @@ public void captureSession(final @NotNull Session session, final @Nullable Objec
     return transaction;
   }
 
+  private @NotNull SentryTransaction applyScope(
```
maciejwalkowiak, line 434: Here very likely we should populate transactions with more data, i just don't want to put everything into single PR.

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -0,0 +1,251 @@
...
+  @Override
+  public @NotNull String getName() {
+    return this.root.getTag(ISpan.NAME_TAG);
+  }
+
+  @Override
+  @Deprecated
+  @ApiStatus.ScheduledForRemoval
+  public void setRequest(@Nullable Request request) {
+    hub.configureScope(scope -> scope.setRequest(request));
```
maciejwalkowiak, line 196: This is not great - ideally we would remove these methods already as it may be surprising to set request on a transaction and realise that it's set on the scope.

maciejwalkowiak:
Few tests are still todo and likely some polishing in tests, other than that, please take another look @marandaneto @bruno-garcia

sentry/src/main/java/io/sentry/ISpan.java
```diff
@@ -32,15 +33,22 @@
...
 
-  /** Sets span timestamp marking this span as finished. */
-  void finish();
+  /**
+   * Sets span timestamp marking this span as finished.
+   *
+   * @return true if transaction has been successfully finished or false if span has been already
+   *     finished
+   */
+  boolean finish();
```
marandaneto, line 42: wondering why we need to return a boolean here

sentry/src/main/java/io/sentry/ITransaction.java
```diff
@@ -72,7 +80,15 @@
...
+   * Returns the transaction name.
+   *
+   * @deprecated use {@link #getName()}
+   * @return transaction name
+   */
   @Nullable
   @ApiStatus.Internal
+  @Deprecated
+  @ApiStatus.ScheduledForRemoval
   String getTransaction();
```
marandaneto, line 93: if we already had `@ApiStatus.Internal`, we could simply remove it

sentry/src/main/java/io/sentry/SentryClient.java
marandaneto: true, we have an issue for that already

sentry/src/main/java/io/sentry/SentryTracer.java
marandaneto: I wonder if there are 2 transactions running at the same time and both call `setRequest`, they are gonna overwrite each other (at least on Android)

sentry/src/main/java/io/sentry/ISpan.java
maciejwalkowiak: Look into `SentryTracer#finish` - we need to know if finishing the root span ended up in finishing or not.

bruno-garcia: Also a bit surprised about returning this. 
It's also a breaking change.

sentry/src/main/java/io/sentry/ITransaction.java
```diff
@@ -69,10 +51,38 @@
...
+   *
+   * @return the request or {@code null} if not set
+   * @deprecated use {@link Scope#getRequest()}
+   */
   @Nullable
-  @ApiStatus.Internal
-  String getTransaction();
+  @Deprecated
+  @ApiStatus.ScheduledForRemoval
+  Request getRequest();
```
bruno-garcia, line 76: Do we need to add this here if it's already marked as deprecated?

sentry/src/main/java/io/sentry/NoOpSpan.java
```diff
@@ -31,10 +31,14 @@ public static NoOpSpan getInstance() {
   }
 
   @Override
-  public void finish() {}
+  public boolean finish() {
+    return true;
```
bruno-garcia, line 35: Shouldn't this be `false` then?
Or ideally `void`

sentry/src/main/java/io/sentry/NoOpTransaction.java
```diff
@@ -38,14 +39,20 @@ public void setName(@NotNull String name) {}
...
   @Override
+  @Deprecated
+  @ApiStatus.ScheduledForRemoval
   public @Nullable Request getRequest() {
     return null;
   }
 
   @Override
+  @Deprecated
+  @ApiStatus.ScheduledForRemoval
```
bruno-garcia, line 55: Each usage of this annotation should follow a javadoc on the replacement API

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -0,0 +1,245 @@
...
+  private final @NotNull SentryId eventId = new SentryId();
+  private final @NotNull Span root;
+  private final @NotNull List<Span> children = new CopyOnWriteArrayList<>();
+  private final @NotNull IHub hub;
+
+  public SentryTracer(final @NotNull TransactionContext context, final @NotNull IHub hub) {
+    Objects.requireNonNull(context, "context is required");
+    Objects.requireNonNull(hub, "hub is required");
+    this.root = new Span(context, this, hub);
+    this.root.setTag(ISpan.NAME_TAG, context.getName());
```
bruno-garcia, line 28: `setTag` was added to Span just so that we can do this, right?
Couldn't we just have a string field for `name` and an API to set TransationName on `ITransaction` instead?

sentry/src/main/java/io/sentry/SentryTracer.java
bruno-garcia: We could just have a `request` field for now and when you do `var transation = new Transation` you set the fields into that. Same for anything else we need to call into the scope like this.

The deprecation note is there so later we can drop the code on next major.
Would this work?

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -0,0 +1,245 @@
...
+    final AtomicReference<Request> contexts = new AtomicReference<>();
+    hub.configureScope(scope -> contexts.set(scope.getRequest()));
+    return contexts.get();
+  }
+
+  @Override
+  @Deprecated
+  @ApiStatus.ScheduledForRemoval
+  public @NotNull Contexts getContexts() {
+    final AtomicReference<Contexts> contexts = new AtomicReference<>();
```
bruno-garcia, line 213: This can have surprising side effects. `Contexts` is thread-safe so we could have a instance here and set it to the SentryTransaction when we capture it.

IMHO changing this instance having side effects on the scope isn't the way to go.

sentry/src/main/java/io/sentry/SentryTracer.java
maciejwalkowiak: Yep, done that.

sentry/src/main/java/io/sentry/ITransaction.java
maciejwalkowiak: "ScheduledForRemoval" is a stronger variant of `Deprecated` and should be used together with `Deprecated`.

sentry/src/main/java/io/sentry/ISpan.java
maciejwalkowiak: Now moved back to `void`. Technically it's possible that `SentryTracer#finish` will be invoked by two threads, only one will finish the root span and both will capture the same transaction. As far as I can see transactions are de-duplicated on the Sentry backend side, and it's not something that is likely to happen, so I guess it can stay as it is right now.

bruno-garcia (approved):

bruno-garcia:
@marandaneto good to merge?

sentry/src/main/java/io/sentry/SentryTracer.java
maciejwalkowiak: This tag is a leftover from the refactoring to drop `ITransaction` completely.

Tags were on span before, but I agree that we can create separate field `name` on SentryTracer.

Done.

sentry-spring/src/test/kotlin/io/sentry/spring/tracing/SentryTransactionAdviceTest.kt
```diff
@@ -80,9 +79,10 @@ class SentryTransactionAdviceTest {
     fun `creates transaction around method in class annotated with @SentryTransaction`() {
         classAnnotatedSampleService.hello()
         verify(hub).captureTransaction(check {
+            println(it)
```
marandaneto, line 82: we can remove this

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -0,0 +1,241 @@
...
+  public void finish(@Nullable SpanStatus status) {
+    if (!root.isFinished()) {
+      root.finish(status);
+      hub.configureScope(
+          scope -> {
+            scope.withTransaction(
+                transaction -> {
+                  if (transaction == this) {
+                    scope.clearTransaction();
+                  }
```
marandaneto, line 110: I guess we are introducing the same bug here, how do we know we are clearing the right transaction? and if its another one? we used to do an equal check, we need to find a way here.

sentry/src/test/java/io/sentry/SentryTracerTest.kt
```diff
@@ -0,0 +1,283 @@
...
+
+class SentryTracerTest {
+
+    private class Fixture {
+        val hub: Hub
+
+        init {
+            val options = SentryOptions()
+            options.dsn = "https://key@sentry.io/proj"
+            hub = spy(Hub(options))
```
marandaneto, line 25: just out of curiosity, what would be the benefit of using `spy` instead of `mock` as usual? are we asserting the `hub` itself? if so, should we not test that thru the HubTest class?

sentry/src/main/java/io/sentry/ISpan.java
bruno-garcia: But there's no need to return to the user whether it was captured or not due double calling it though so void works fine

sentry/src/main/java/io/sentry/SentryTracer.java
maciejwalkowiak: Since we are inside `withTransaction`, does the lock prevent from setting any transaction during this operation?

sentry/src/test/java/io/sentry/SentryTracerTest.kt
maciejwalkowiak: With using mock we can't really test that transaction is cleared from the scope. If we use mock for `finish` method, we won't really test much.

marandaneto (approved):
LGTM
huge win, thanks!

PR #1312, opened 2021-03-09
Fix: Use logger set on SentryOptions in GsonSerializer

### LINKED ISSUE #1308: Event/Transaction payloads are not logged by GsonSerializer
The version of the SDK: **4.2.0**

Initializing Sentry with debug=true does not cause `GsonSerializer` to log event payloads.

maciejwalkowiak:
## :scroll: Description

Use logger set on SentryOptions in GsonSerializer.

Fixes #1308

Unit tests.

## :pencil: Checklist

- [ ] No breaking changes

The breaking change is that `GsonSerializer` does not take a `ILogger` as a constructor parameter but `SentryOptions` instead.

### DIFF
--- a/sentry/src/main/java/io/sentry/GsonSerializer.java
+++ b/sentry/src/main/java/io/sentry/GsonSerializer.java
@@ -48,8 +48,8 @@ public final class GsonSerializer implements ISerializer {
   @SuppressWarnings("CharsetObjectCanBeUsed")
   private static final Charset UTF_8 = Charset.forName("UTF-8");
 
-  /** the ILogger interface */
-  private final @NotNull ILogger logger;
+  /** the SentryOptions */
+  private final @NotNull SentryOptions options;
 
   /** the Gson instance */
   private final @NotNull Gson gson;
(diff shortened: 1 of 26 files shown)


sentry/src/main/java/io/sentry/GsonSerializer.java
```diff
@@ -60,12 +60,12 @@
...
    * AndroidSerializer ctor
    *
-   * @param logger the ILogger interface
+   * @param options the SentryOptions object
    * @param envelopeReader the IEnvelopeReader interface
    */
   public GsonSerializer(
-      final @NotNull ILogger logger, final @NotNull IEnvelopeReader envelopeReader) {
-    this.logger = Objects.requireNonNull(logger, "The ILogger object is required.");
+      final @NotNull SentryOptions options, final @NotNull IEnvelopeReader envelopeReader) {
```
marandaneto, line 67: `envelopeReader` is already a property from `SentryOptions`, we could remove it then

marandaneto:
looks like that every property that could be mutated during SDK init should take Options instead of the reference and this would be a lot of changes across the SDKs as at the beginning we're trying to be more specific.
happy to merge as it is, but we need to keep this in mind from now on.

sentry/src/main/java/io/sentry/GsonSerializer.java
maciejwalkowiak: 👍

marandaneto (approved):
LGTM

Issue #1316, opened 2021-03-11
RateLimiter logs the the number of dropped item for each dropped item in the SentryEnvelope


bobinx:
https://github.com/getsentry/sentry-java/blob/bb19d62bda9b161e366dd792d1dd1122c28f87cd/sentry/src/main/java/io/sentry/transport/RateLimiter.java#L64-L76

Just saw that the count of dropped items in envelope is logged every time an item is added to the 'dropItems' list. 
Is this the expected behavior? Can the logging be moved outside the loop to reduce the amount of logs? 
I am new to sentry 🙂 Please excuse if I am missing something here.

marandaneto:
@bobinx thanks for raising this, you're right, would you like contributing with a PR? :)

bobinx:
@marandaneto I have created a PR with the changes.
Please have a look

marandaneto:
thanks :)

PR #1318, opened 2021-03-12
Log dropped items size in RateLimiter only after all items are added to list

bobinx:
## :scroll: Description

Currently, The debug logs for dropped item size is printed each time a dropped item is added to the list from SentryEnvelop in RateLimiter class.
Updated this to log the dropped item size only after we have the complete list of dropped items to avoid unnecessary logs.

https://github.com/getsentry/sentry-java/issues/1316

## :pencil: Checklist

- [x] No breaking changes


marandaneto (approved):
LGTM

thanks @bobinx

PR #1325, opened 2021-03-14
ref: contact through sdk email

bruno-garcia:
new DL for sdks


bruno-garcia:
using a DL that gets routed to sdk folks

PR #1328, opened 2021-03-15
Allow 0.0 to be set on tracesSampleRate.

### LINKED ISSUE #1326: Allow `tracesSampleRate` = 0.0
We should allow `tracesSampleRate` to be set to 0.0, 0.0 means no transaction will be created by this SDK unless an incoming trace has been sampled.

cc @Tyrrrz this is relevant for you too, please have a look and create an issue if necessary.

maciejwalkowiak:
## :scroll: Description

Allow 0.0 to be set on tracesSampleRate.

Fixes #1326.

## :pencil: Checklist

- [x] No breaking changes


maciejwalkowiak:
The question is what's the point of having `null` on `tracesSampleRate`? Shouldn't we default it to `0.0`? Or this is actually a bug, and having `null` on `tracesSampleRate` should make all `startTransaction` calls return no-op?

marandaneto:
> The question is what's the point of having `null` on `tracesSampleRate`? Shouldn't we default it to `0.0`? Or this is actually a bug, and having `null` on `tracesSampleRate` should make all `startTransaction` calls return no-op?

@maciejwalkowiak that was the point.

@bruno-garcia to confirm but if I understood correctly means:

`tracesSampleRate` and `tracesSampler` == null -> Tracing is fully disabled, we could do NoOp IMO as discussed in the meeting.

`tracesSampleRate` == 0.0, means no transaction will be created by this SDK unless an incoming trace has been sampled

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -671,7 +671,12 @@ public void setProxy(@Nullable Proxy proxy) {
    * @param sampleRate the sample rate
    */
   public void setSampleRate(Double sampleRate) {
-    this.validateRate(sampleRate);
+    if (sampleRate != null && (sampleRate > 1.0 || sampleRate <= 0.0)) {
+      throw new IllegalArgumentException(
+          "The value "
+              + sampleRate
+              + " is not valid. Use null to disable or values between 0.01 and 1.0.");
```
marandaneto, line 678:
```suggestion
              + " is not valid. Use null to disable or values > 0.0 and <= 1.0.");
```

the error message is misleading, 0.001 would be valid already but the log says 0.01

marandaneto:
missing @tests and changelog, let's wait @bruno-garcia to confirm though

marandaneto:
@bruno-garcia thats all we need right?
@maciejwalkowiak could u work on that, so I can cut a release with it :)

bruno-garcia (approved):
Other than the note from @marandaneto LGTM.
As Manoel already wrote on this PR: `null` means *no transactions should be emitted*.

But 0.0 means no chance of random sampling creating transactions, but if an incoming request is sampled, the SDK will honor that decision.

bruno-garcia:
Looks like we're missing docs too: https://docs.sentry.io/platforms/java/configuration/filtering/#using-sampling-to-filter-transaction-events

marandaneto (approved):

maciejwalkowiak:
@bruno-garcia docs updated: https://github.com/getsentry/sentry-docs/pull/3277

Disabling performance feature when tracesSampleRate is null will come in subsequent PR.

PR #1329, opened 2021-03-16
GraalVM Native Image Compatibility.

### LINKED ISSUE #1323: Error reporting does not work if used with native image (graalvm) due to reflection usage in the lib
_Platform:_
- [x] Kotlin -> 1.4.20, graalvm 11
- [x] Logback -> 1.2.3 
- [x] Quarkus.io -> 1.12.1.Final

_Build system:_
- [x]  Gradle -> 6.3

The version of the SDK:
**4.1.0**

---
I have the following issue:

All of our services use sentry.io to report errors. This works nicely for our legacy spring services that do not use graalvm to build native container images. Our latest services use quarkus.io and graalvm though and the reporting does not work for them. Instead of reporting a meaningful error to sentry, the error lacks any content - no stack trace, error message, whatsoever: 

image

If I build any of these quarkus.io services with a traditional image (openjdk-11), the reporting works as expected. In native images it does not work though. 

I also created an error on the quarkus.io github: https://github.com/quarkusio/quarkus/issues/15623

This issue can be temporarily worked around by providing the sentry classes that need reflection in a reflection-config.json file, which is read by graalvm. IMO this should be fixed in the sentry library though. 

**Steps to reproduce:**
- Provoke an error that will be reported to sentry 
- Ensure that this application runs natively (graalvm)

**Actual result:**
- Unlabelled error with no content (message, stacktrace)

**Expected result:**
- Regular error with content

### LINKED ISSUE #1324: Spring Native Compatibility
Hi there,

[Spring Native](https://spring.io/blog/2021/03/11/announcing-spring-native-beta) was recently released.

With `implementation 'org.springframework.experimental:spring-native:0.9.0'` and `compile 'io.sentry:sentry-spring-boot-starter:4.3.0'` present, building a native image per the spring native docs gives the below error:

`Caused by: java.lang.IllegalStateException: ERROR: in 'io.sentry.spring.tracing.SentryAdviceConfiguration' these methods are directly invoking methods marked @Bean: [sentryTransactionAdvisor, sentrySpanAdvisor] - due to the enforced proxyBeanMethods=false for components in a native-image, please consider refactoring to use instance injection. If you are confident this is not going to affect your application, you may turn this check off using -Dspring.native.verify=false.
`

In order to be compatible with Spring Native, it would be good if `io.sentry.spring.tracing.SentryAdviceConfiguration` was refactored per the guidance from Spring Native above - avoid directly using the Bean, and instead inject an instance. Naturally, it would be better to avoid disabling verification on builds.

I don't know if there are going to be any other issues to be compatible but it seems like a goo place to start. What do you think?

maciejwalkowiak:
## :scroll: Description

GraalVM Native Image Compatibility.

- adds all necessary configuration files for running Sentry SDK in GraalVM native image
- fixes Spring configuration files to be compatible with Spring Native 0.10

Fixes #1323.

Partially fixes #1324 - error reporting works with performance there is one gotcha - it does not work with `@SentrySpan` and `@SentryTransaction` annotations. Since Spring Native is evolving project potentially it may handle this use case in the future without our custom code. I'll sync with Spring team about it but this can be anyway solved in separate PR.

Building native image with Spring Native and Quarkus and checked if no exceptions are thrown in runtime and if errors sent to Sentry have complete information.

## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next steps

On the side in meantime work on Spring Native integration to handle properly `@SentrySpan` and `@SentryTransaction` annotations.


sentry/src/main/resources/META-INF/native-image/io.sentry/sentry/reflect-config.json
```diff
@@ -0,0 +1,146 @@
+[
```
marandaneto, line 1: missing the envelope classes that rely on serialization too eg SentryEnvelopeItem ?

maciejwalkowiak: Hm I'll integration works without these classes. Just in case I'll add them in case I missed some use case.

marandaneto (approved):
approving blindly as I know nothing about GraalVM compatibility :D

maciejwalkowiak:
To be fair I am also not an expert in this subject. I think it is a step forward and definitely it makes it work with GraalVM but I am not sure if it covers it all. I hope we will get feedback once its there.

PR #1330, opened 2021-03-16
Feat: Add OkHttp client application interceptor

marandaneto:
## :scroll: Description
Create spans and add breadcrumbs
A sort of guideline: https://github.com/getsentry/develop/issues/254

Unit tests, sample project.

## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next steps

Resolve TODOs in the code.


sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpInterceptor.kt
```diff
@@ -0,0 +1,61 @@
...
+        } catch (e: IOException) {
+            span?.throwable = e
+            throw e
+        } finally {
+            span?.finish(SpanStatus.fromHttpStatusCode(code, SpanStatus.INTERNAL_ERROR))
+        }
+
+        val breadcrumb = Breadcrumb.http(request.url.toString(), request.method, code)
+        request.body?.contentLength().ifHasValidLength {
+            breadcrumb.setData("requestBodySize", it)
```
maciejwalkowiak, line 42: I am not sure how really valuable this information is, both for requests and responses.

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpInterceptor.kt
```diff
@@ -0,0 +1,61 @@
...
+        }
+        response.body?.contentLength().ifHasValidLength {
+            breadcrumb.setData("responseBodySize", it)
+        }
+        hub.addBreadcrumb(breadcrumb)
+
+        // TODO: collect ideas
+        // https://github.com/square/okhttp/blob/master/okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt
+        // TODO: add package to options? how do we get options? does it even make sense?
+        // sdkVersion?.addPackage("maven:io.sentry:sentry-android-timber", BuildConfig.VERSION_NAME)
```
maciejwalkowiak, line 52: we can do it through `Hub#getOptions` but I somehow doubt it makes sense to add it for each and every instrumentation we do. I am not the one with access to reporting tools though.

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/SecondActivity.kt
```diff
@@ -18,7 +19,10 @@ class SecondActivity : AppCompatActivity() {
 
     private lateinit var repos: List<Repo>
 
-    private val client = OkHttpClient.Builder().addInterceptor(NetworkInterceptor()).build()
+    // TODO: its an application interceptor
+    // whats about network interceptor, does it make sense?
```
maciejwalkowiak, line 23: I think higher level application interceptor is a good start and likely enough.

maciejwalkowiak:
@marandaneto I left the TODOs you created as they are more like open questions. Other than that, interceptor is working, breadcrumbs are added, everything is tested.

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpInterceptor.kt
marandaneto: for debugging (when there are events, we know each packages were being used).
or usage of our packages to take decisions like dropping a version or stuff like this.

true, forgot about `Hub#getOptions`, let's add then :)

buildSrc/src/main/java/Config.kt
```diff
@@ -1,7 +1,7 @@
 import java.math.BigDecimal
 
 object Config {
-    val kotlinVersion = "1.4.10"
+    val kotlinVersion = "1.4.30"
```
marandaneto, line 4:
```suggestion
    val kotlinVersion = "1.4.10"
```

lets rollback to avoid conflicts, i am likely downgrading to v1.3.x

buildSrc/src/main/java/Config.kt
```diff
@@ -34,6 +35,9 @@ object Config {
     object Libs {
         val appCompat = "androidx.appcompat:appcompat:1.2.0"
         val timber = "com.jakewharton.timber:timber:4.7.1"
+        val okhttpBom = "com.squareup.okhttp3:okhttp-bom:4.9.0"
+        val okhttp = "com.squareup.okhttp3:okhttp"
+        val okhttpInterceptor = "com.squareup.okhttp3:logging-interceptor"
```
marandaneto, line 40: not being used, okhttp ships the interface already

sentry-android-okhttp/build.gradle.kts
```diff
@@ -0,0 +1,96 @@
...
+    explicitApi()
+}
+
+dependencies {
+    api(project(":sentry"))
+
+    implementation(Config.Libs.okhttpBom)
+    implementation(Config.Libs.okhttp)
+// not needed
+//    implementation(Config.Libs.okhttpInterceptor)
```
marandaneto, line 76: remove it

marandaneto, line 76:
```suggestion
```

buildSrc/src/main/java/Config.kt
marandaneto, line 40:
```suggestion
```

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpInterceptor.kt
marandaneto: I agree, @bruno-garcia suggested it and I also found in the okhttp interceptors
https://github.com/square/okhttp/blob/master/okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt
so kinda has an use case

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/SecondActivity.kt
marandaneto, line 23:
```suggestion
```

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpInterceptor.kt
```diff
@@ -0,0 +1,61 @@
...
+            response = chain.proceed(request)
+            code = response.code
+        } catch (e: IOException) {
+            span?.throwable = e
+            throw e
+        } finally {
+            span?.finish(SpanStatus.fromHttpStatusCode(code, SpanStatus.INTERNAL_ERROR))
+        }
+
+        val breadcrumb = Breadcrumb.http(request.url.toString(), request.method, code)
```
marandaneto, line 40: wondering if it makes sense to call `Breadcrumb.http` without `code` if its -1, otherwise its adding -1 which does not make sense, or Breadcrumb.http inside of it checks if its -1, or changing it to Integer and check if its null

maciejwalkowiak: 👍 +1

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpInterceptor.kt
maciejwalkowiak: Concerns:
- we would need to add it to all events and transactions
- since `packages` on `sdkVersions` is a list (not set) creating multiple interceptors would add multiple same packages

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpInterceptor.kt
```diff
@@ -0,0 +1,58 @@
...
+        if (code != -1) {
+            val breadcrumb = Breadcrumb.http(request.url.toString(), request.method, code)
+            request.body?.contentLength().ifHasValidLength {
+                breadcrumb.setData("requestBodySize", it)
+            }
+            response.body?.contentLength().ifHasValidLength {
+                breadcrumb.setData("responseBodySize", it)
+            }
+            hub.addBreadcrumb(breadcrumb)
+        }
```
marandaneto, line 49: should this be inside of the `finally` block? otherwise, if it throws, we won't be able to record breadcrumbs or am I missing smt?

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpInterceptor.kt
marandaneto: yeah, lets keep without it for now then and only do it when the OkHttp integration is able to captureEvent then, which would be then an integration for now it just works as an interceptor.

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpInterceptor.kt
maciejwalkowiak: I changed it to always add breadcrumb even if call fails with network error.

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpInterceptor.kt
maciejwalkowiak: fixed

marandaneto:
@maciejwalkowiak run `apiDump` and we are good

marandaneto:
I cannot approve my own PR but you have my amen

PR #1332, opened 2021-03-18
Fix: set "java" platform to transactions

maciejwalkowiak:
## :scroll: Description

Fix: set "java" platform to transactions

Each event/transaction that SDK sends should have platform set.

Unit tests.

## :pencil: Checklist

- [x] No breaking changes


maciejwalkowiak:
The alternative implementation that I think I would be in favour of is that `SentryBaseEvent#platform` is set to "java" by default when object is created. This way we could mark it as `@NotNull` and just let user overwrite it with non-null values if needed.

If it's not something that users should be able to overwrite we could then remove the setter completely. What do you think @bruno-garcia @marandaneto

sentry/src/main/java/io/sentry/SentryBaseEvent.java
```diff
@@ -176,4 +190,12 @@ public String getEnvironment() {
...
     this.environment = environment;
   }
+
+  public String getPlatform() {
+    return platform;
+  }
+
+  public void setPlatform(String platform) {
+    this.platform = platform;
+  }
```
marandaneto, line 200:
```suggestion
  public @Nullable String getPlatform() {
    return platform;
  }

  public void setPlatform(final @Nullable String platform) {
    this.platform = platform;
  }
```

marandaneto:
> The alternative implementation that I think I would be in favour of is that `SentryBaseEvent#platform` is set to "java" by default when object is created. This way we could mark it as `@NotNull` and just let user overwrite it with non-null values if needed.
> 
> If it's not something that users should be able to overwrite we could then remove the setter completely. What do you think @bruno-garcia @marandaneto

better not, we also capture events from native platforms (Android NDK), it's important to know if there was a value, or not. 
I think Gson would overwrite anyway but seems wrong.

maciejwalkowiak:
But do we set different value for NDK? The javadocs says:

```
Acceptable values are: `as3`, `c`, `cfml`, `cocoa`, `csharp`, `elixir`, `haskell`, `go`,
   * `groovy`, `java`, `javascript`, `native`, `node`, `objc`, `other`, `perl`, `php`, `python`,
   * `ruby`
```

marandaneto:
> But do we set different value for NDK? The javadocs says:
> 
> ```
> Acceptable values are: `as3`, `c`, `cfml`, `cocoa`, `csharp`, `elixir`, `haskell`, `go`,
>    * `groovy`, `java`, `javascript`, `native`, `node`, `objc`, `other`, `perl`, `php`, `python`,
>    * `ruby`
> ```

NDK events are `native`, Dart events are `other`, RN events are `javascript`

maciejwalkowiak:
Ok, then PR is ready to review as it is.

marandaneto (approved):
LGTM

bruno-garcia:
> But do we set different value for NDK? The javadocs says:
> 
> ```
> Acceptable values are: `as3`, `c`, `cfml`, `cocoa`, `csharp`, `elixir`, `haskell`, `go`,
>    * `groovy`, `java`, `javascript`, `native`, `node`, `objc`, `other`, `perl`, `php`, `python`,
>    * `ruby`
> ```

NDK will use `native`

PR #1337, opened 2021-03-19
Allow disabling tracing through SentryOptions.

maciejwalkowiak:
## :scroll: Description

Allow disabling tracing through SentryOptions. When both tracesSampleRate = null and tracesSampler = null tracing is disabled.

Since this behavior is in place, `sentry.enable-tracing` property in Spring Boot integration is not needed anymore and it gets deprecated.

According to specs, users can disable tracing by setting both tracesSampleRate to null and tracesSampler to null.

Unit tests.

## :pencil: Checklist

- [ ] No breaking changes

Breaking change: for users who have set `null` on both `tracesSampleRate` and `tracesSampler` - transactions with explicit sampling decision and transactions with inherited sampling decision will not be send to Sentry anymore. To bring back the old behaviour users would have to set `tracesSampleRate` to `0.0`.


sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -1268,6 +1268,16 @@ public void setEnableDeduplication(final @Nullable Boolean enableDeduplication)
...
 
+  /**
+   * Returns if tracing should be enabled. If tracing is disabled, starting transactions returns
+   * {@link NoOpTransaction}.
+   *
+   * @return if tracing is enabled.
+   */
+  public boolean isTracingEnabled() {
+    return getTracesSampleRate() != null || getTracesSampler() != null;
+  }
```
marandaneto, line 1279: this method could be used here too now: https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java#L67-L70

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryAutoConfiguration.java
```diff
@@ -103,6 +103,9 @@
 
       options.setSentryClientName(BuildConfig.SENTRY_SPRING_BOOT_SDK_NAME);
       options.setSdkVersion(createSdkVersion(options));
+      if (options.getTracesSampleRate() == null) {
+        options.setTracesSampleRate(0.0);
+      }
```
marandaneto, line 108: why do we need that here? isn't users choice to set it up?

maciejwalkowiak: Spring has already a property "enable-tracing" which triggers auto configuration. If we don't do this change, users would have to set both `enable-tracing=true` and `traces-sample-rate=0.0` to enable integration.

marandaneto: but that's what is described here right https://docs.sentry.io/platforms/java/guides/spring-boot/performance/#enable-tracing 
if it's a matter of not doing multiple things, I'd say that removing `enable-tracing` would be the way to go instead.

maciejwalkowiak: We can do it but it will be breaking. Probably better to do it in another PR and leave this one as it is?

marandaneto: wfm

maciejwalkowiak: As decided on Friday, I'll deprecate `enable-tracing` and make `traces-sample-rate` turn on integration too.

sentry/src/main/java/io/sentry/SentryOptions.java
maciejwalkowiak: 👍

maciejwalkowiak:
@marandaneto please have another look

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryAutoConfiguration.java
marandaneto: @maciejwalkowiak we need docs for this

maciejwalkowiak: Sure, I'll prepare a PR

marandaneto: after deprecating `enable-tracing`, we should remove this right now, right?

maciejwalkowiak: I don't think so. This keeps the old behavior - users only need to set `enable-tracing=true` to get tracing working. If we remove it, it is a breaking change.

marandaneto (approved):
LGTM

marandaneto:
@maciejwalkowiak a conflict to be fixed and GG

PR #1352, opened 2021-03-24
Add option to ignore exceptions by type.

### LINKED ISSUE #1306: Ignore Exception by type
The SDK needs a first class API to filter out exceptions by their type.

The way to filter out today is through `beforeSend` such as:

```java
Sentry.init(options -> {
  options.setBeforeSend((event, _hint_) -> {
    return event.getThrowable() instanceof SQLException ? null : event;
  });
});
```

This is a feature in the [.NET SDK documented here for reference](https://docs.sentry.io/platforms/dotnet/ignoring-exceptions/).

i.e:
`SentrySdk.Init(o => o.AddExceptionFilterForType<OperationCanceledException>());`

In hindsight maybe not the best name `AddExceptionFilterForType`, perhaps `FilterException<TException>` would be a better candidate although keeping the name consistent is also a nice to have.

This issue was originally raised here: https://github.com/getsentry/sentry-java/issues/575

maciejwalkowiak:
## :scroll: Description

Add option to `SentryOptions` to ignore exceptions by type. Works also for Spring Boot configuration and external configuration for other integrations.

Fixes #1306

Unit & integration tests.

## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next steps

Update reference docs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+* Feat: Add option to ignore exceptions by type. #1352
+
 # 4.4.0-alpha.1
 
 * Bump: sentry-native to 0.4.8
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * Feat: Set SDK version on Transactions (#1307)
 * Fix: Use logger set on SentryOptions in GsonSerializer (#1308)
 * Fix: Use the bindToScope correctly
+* Feat: GraalVM Native Image Compatibility (#1329)
 * Fix: Allow 0.0 to be set on tracesSampleRate (#1328)
 * Fix: set "java" platform to transactions #1332
 * Feat: Add OkHttp client application interceptor (#1330)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Fix: Allow 0.0 to be set on tracesSampleRate (#1328)
 * Fix: set "java" platform to transactions #1332
 * Feat: Add OkHttp client application interceptor (#1330)
+* Fix: Allow disabling tracing through SentryOptions (#1337)
 
 # 4.3.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-* Feat: Add option to ignore exceptions by type. #1352
+* Feat: Add option to ignore exceptions by type (#1352)
 
 # 4.4.0-alpha.1
 
--- a/sentry/src/main/java/io/sentry/SentryClient.java
+++ b/sentry/src/main/java/io/sentry/SentryClient.java
@@ -60,19 +60,6 @@ public final class SentryClient implements ISentryClient {
 
     options.getLogger().log(SentryLevel.DEBUG, "Capturing event: %s", event.getEventId());
 
-    SentryId sentryId = SentryId.EMPTY_ID;
-
-    if (event.getThrowable() != null
-        && options.containsIgnoredExceptionForType(event.getThrowable())) {
-      options
-          .getLogger()
-          .log(
-              SentryLevel.DEBUG,
-              "Event was dropped as the exception %s is ignored",
-              event.getThrowable().getClass());
-      return sentryId;
-    }
-
     if (ApplyScopeUtils.shouldApplyScopeData(hint)) {
       // Event has already passed through here before it was cached
       // Going through again could be reading data that is no longer relevant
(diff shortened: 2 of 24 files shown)


sentry/src/main/java/io/sentry/SentryClient.java
```diff
@@ -60,6 +60,19 @@ public boolean isEnabled() {
 
     options.getLogger().log(SentryLevel.DEBUG, "Capturing event: %s", event.getEventId());
 
+    SentryId sentryId = SentryId.EMPTY_ID;
+
+    if (event.getThrowable() != null
```
marandaneto, line 65: this should be probably just before the `executeBeforeSend` callback, because we don't want to lose the chance to mark our session as crashed or something, it'd be the same concept of the `beforeSend` actualy

sentry/src/test/java/io/sentry/SentryOptionsTest.kt
```diff
@@ -469,13 +483,42 @@ class SentryOptionsTest {
...
             temporaryFolder.delete()
         }
     }
 
+    @Test
+    fun `creates options with ignored exception types using external properties`() {
+        // create a sentry.properties file in temporary folder
+        val temporaryFolder = TemporaryFolder()
+        temporaryFolder.create()
+        val file = temporaryFolder.newFile("sentry.properties")
```
marandaneto, line 498: can we extract this to an aux method for all the tests? its being duplicated

sentry/src/main/java/io/sentry/SentryClient.java
maciejwalkowiak: I guess it should be aligned with other SDKs. I think ignoring means also doing as little as possible - in this case skipping all event processing - and not not marking session as crashed.

marandaneto: @maciejwalkowiak it's a topic we've touched before, if you dropped an event in your beforeSend, you decided to do it, still, we prefer to give proper stats about your sessions.
the ignored exceptions are pretty much the same concept and I don't why it'd be different.
@bruno-garcia

maciejwalkowiak: 👍

maciejwalkowiak:
"Danger" does not recognize that changelog has been already fixed.

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
 
+* Feat: Add option to ignore exceptions by type. #1352
```
bruno-garcia, line 3: Danger expects `-`?

```suggestion
- Feat: Add option to ignore exceptions by type. #1352
```

bruno-garcia (approved):
👍

CHANGELOG.md
marandaneto, line 3:
```suggestion
* Feat: Add option to ignore exceptions by type. (#1352)
```

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -1278,6 +1306,60 @@ public boolean isTracingEnabled() {
...
+
+  /**
+   * Adds exception type to the list of ignored exceptions.
+   *
+   * @param exceptionType - the exception type
+   */
+  public void addIgnoredExceptionForType(final @NotNull Class<? extends Throwable> exceptionType) {
+    if (this.ignoredExceptionsForType == null) {
+      this.ignoredExceptionsForType = new HashSet<>();
+    }
```
marandaneto, line 1347: what's about creating it directly and using a thread-safe `Set` as we do with all the other list/array/map? see eventProcessors and integrations, we already got issues that people were iterating the list and removing from a loop which obviously would throw (if not thread-safe type).

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -1278,6 +1306,60 @@ public boolean isTracingEnabled() {
...
+  }
+
+  /**
+   * Sets the list of exception types that once captured will not be sent to Sentry as {@link
+   * SentryEvent}.
+   *
+   * @param ignoredExceptionsForType list of ignored exception types
+   */
+  @SuppressWarnings("unchecked")
+  public void setIgnoredExceptionsForType(final @NotNull Set<Class<?>> ignoredExceptionsForType) {
```
marandaneto, line 1327: we don't offer a set for any other array/list, but they should call `addIgnoredExceptionForType` instead so if at some point we come with some default types, they don't get overwritten

marandaneto:
> "Danger" does not recognize that changelog has been already fixed.

you have to use the #id between ()

sentry/src/main/java/io/sentry/SentryOptions.java
maciejwalkowiak: This is an edge case how Spring Boot binds properties. If the property is a collection and is initialized in a field or constructor, it calls `collection#add()` - in this case it would mean that it's technically possible to set types that do not extend throwable into this list.

maciejwalkowiak: I'll take another look if there is a better way to overcome it

marandaneto (approved):
LGTM

marandaneto:
@maciejwalkowiak please add to the docs -> https://docs.sentry.io/platforms/java/configuration/options/

PR #1363, opened 2021-03-30
Remove signal handlers side by side limitation from README

marandaneto:
After running a few tests, everything worked out as expected

_#skip-changelog_


bruno-garcia (approved):
Raises more questions to why one system had more crashes than the other and without proper repros, it'll be fun to investigate this.

bruno-garcia:
It usually defaults the commit message like the PR name, but it didn't here (and I hit merge quickly). Not sure why this happens but it isn't the first time.

Issue #1366, opened 2021-03-30
NPE while applying `Map` with `nullable` value to `User#setOthers`


wzieba:
_Platform:_
- [x]  Android -> If yes, which Device API (and compileSdkVersion/targetSdkVersion/Build tools) version?
Device `API`: 26
`compileSdkVersion`: 30
`targetSdkVersion`: 30
`buildTools`: applied by Android Gradle Plugin

_IDE:_
- [x]  Android Studio -> If yes, which version?
`4.2 beta 3` 

_Build system:_
- [x]  Gradle -> If yes, which version?
`6.7.1`

_Android Gradle Plugin:_
- [x]  Yes -> If yes, which version?
`4.2.0-beta03`

_Sentry Android Gradle Plugin:_
- [x]  No

_Proguard/R8:_
- [x]  Disabled

_Platform installed with:_
- [x] Maven Central

The version of the SDK:
**4.3.0**

---
I have the following issue:

Setting `map` with `Nullable` value results with `NullPointerException`. The reason is that SDK uses `ConcurentHashMap` which does not accept nullable values ([ConcurentHashMap source](http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/concurrent/ConcurrentHashMap.java#l1011))

**Steps to reproduce:**
Run
```
User().apply {
    others = mapOf("test" to null)
}
```

**Actual result:**
- `ConcurrentHashMap.putVal` throws `NullPointerException`

**Expected result:**
- `User#setOthers` should not accept map with nullable values

---

I know you work on better null-safety (https://github.com/getsentry/sentry-java/issues/1357#issuecomment-809265747) but would you consider accepting a small PR with just a signature update, like this: https://github.com/wzieba/sentry-java/commit/84eb804a8d4967e3cef9fb8e45bbbf3344897edb?

marandaneto:
@wzieba thanks for reporting.

yes, adding such annotations is actually one of the things we'd do, adding to the key makes sense too IMO.
btw TIL you can add such annotations inside of the diamond operator.

even better if we search for all the usage of `ConcurrentHashMap` within this project and when such method exists, add the annotations.

PR #1368, opened 2021-03-30
Accept only non null value maps

wzieba:
Resolves: #1366 

## :scroll: Description

This PR adds `org.jetbrains.annotations.NotNull` annotation for a places, where SDK could receive a `Map` from a client, which has a `nullable` type for `value` generic parameter.

I've also added this annotation to internal API methods and some fields to which `ConcurrentHashMap` is assigned.

`ConcurrentHashMap` can't have a `null` for any of it's entries value, [otherwise `NullPointerException` is thrown](http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/concurrent/ConcurrentHashMap.java#l1011). By adding this `NotNull` annotation we:
- raise awareness of this issue for Java clients
- have a build-time check for Kotlin clients:
image

Unfortunately, I don't know how to test it, I'm open to any suggestions.

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/sentry/src/main/java/io/sentry/config/EnvironmentVariablePropertiesProvider.java
+++ b/sentry/src/main/java/io/sentry/config/EnvironmentVariablePropertiesProvider.java
@@ -33,12 +33,14 @@ final class EnvironmentVariablePropertiesProvider implements PropertiesProvider 
   public @NotNull Map<String, String> getMap(final @NotNull String property) {
     final String prefix = propertyToEnvironmentVariableName(property) + "_";
 
-    final Map<String, String> result = new ConcurrentHashMap<>();
+    final Map<String, @NotNull String> result = new ConcurrentHashMap<>();
     for (Map.Entry<String, String> entry : System.getenv().entrySet()) {
       final String key = entry.getKey();
       if (key.startsWith(prefix)) {
         final String value = StringUtils.removeSurrounding(entry.getValue(), "\"");
-        result.put(key.substring(prefix.length()).toLowerCase(Locale.ROOT), value);
+        if (value != null) {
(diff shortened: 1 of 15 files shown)


sentry/src/main/java/io/sentry/config/EnvironmentVariablePropertiesProvider.java
```diff
@@ -33,12 +33,14 @@
...
     final String prefix = propertyToEnvironmentVariableName(property) + "_";
 
-    final Map<String, String> result = new ConcurrentHashMap<>();
+    final Map<String, @NotNull String> result = new ConcurrentHashMap<>();
     for (Map.Entry<String, String> entry : System.getenv().entrySet()) {
       final String key = entry.getKey();
       if (key.startsWith(prefix)) {
         final String value = StringUtils.removeSurrounding(entry.getValue(), "\"");
-        result.put(key.substring(prefix.length()).toLowerCase(Locale.ROOT), value);
+        if (value != null) {
```
wzieba, line 41: `StringUtils#removeSurrounding` can return `null` value, therefore this check.

marandaneto: makes sense

marandaneto (approved):
thanks for the contribution @wzieba :)

marandaneto:
@wzieba 

> Unfortunately, I don't know how to test it, I'm open to any suggestions.

as we write tests in Kotlin and compile fails if passing a null value to a @NotNull param, the only way is thru reflection, it's not worth IMO, we've done at the beginning but if users are aware that they can't pass null, it's already good enough :) thanks once more.

PR #1372, opened 2021-04-01
Make NoOpHub public

egalpin:
## :scroll: Description
Makes `NoOpHub` public.

This enables easier resource management in an Apache Beam use case that I have. In Beam, I must manage an IHub manually rather than using static methods in `Sentry`. Allowing end-users to create NoOpHub instances will provide the kind of flexibility needed for this kind of use case.

If there is currently a method by which I can create an instance of NoOpHub directly, please point it out and close this PR. Thanks!

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/sentry/src/main/java/io/sentry/NoOpHub.java
+++ b/sentry/src/main/java/io/sentry/NoOpHub.java
@@ -7,7 +7,7 @@ import java.util.List;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-final class NoOpHub implements IHub {
+public final class NoOpHub implements IHub {
 
   private static final NoOpHub instance = new NoOpHub();
 


maciejwalkowiak:
I think we can make it public. @marandaneto @bruno-garcia?

sentry/src/main/java/io/sentry/NoOpHub.java
```diff
@@ -7,7 +7,7 @@
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-final class NoOpHub implements IHub {
+public final class NoOpHub implements IHub {
```
marandaneto, line 10: you need to annotate this class with `@Open` too

marandaneto:
> I think we can make it public. @marandaneto @bruno-garcia?

I think it's fine but I quite didn't understand why it's necessary.

sentry/src/main/java/io/sentry/NoOpHub.java
maciejwalkowiak: It stays final, why would it need to be `@Open`?

maciejwalkowiak (commented):
@egalpin please run `./gradlew apiDump`. It will create a file listing changes in the public API. Then add the change and push. Once it's there the build should pass.

sentry/src/main/java/io/sentry/NoOpHub.java
marandaneto: ups my bad, CI failed and I misread the issue, they need to run `./gradlew apiDump` because there's a new public class in our public API

egalpin:
I just ran `./gradlew apiDump` but ran into a failure:

```
A problem occurred configuring project ':sentry-android-ndk'.
> SDK location not found. Define location with an ANDROID_SDK_ROOT environment variable or by setting the sdk.dir path in your project's local properties file at '/me/code/sentry-java/local.properties'.
```

Do I need to have the Android SDK installed? I'd prefer not to haha. Is there an automated build system I could leverage? Also feel free to push directly to this PR as needed if that's a path to a resolution 🙂 Thanks!

marandaneto:
in favor of https://github.com/getsentry/sentry-java/pull/1379

PR #1375, opened 2021-04-06
Upgrade Apache HttpComponents Core to 5.0.3.

maciejwalkowiak:
## :scroll: Description

Upgrade Apache HttpComponents Core to 5.0.3.

HttpComponents Core 5.0.3 contains a bugfix for a potential infinite loop and CPU exhaustion.

[According to library authors, HttpComponents Core 5.0.3 is compatible with HttpComponents Client 5.0.3](https://lists.apache.org/thread.html/rdb608c0d93474e6e19a56fe48075240104f0f83121f6cd3ca2b7650d%40%3Chttpclient-users.hc.apache.org%3E).

Covered by existing tests. Could not reproduce the infinite loop.

## :pencil: Checklist

- [x] No breaking changes


maciejwalkowiak:
cc @bruno-garcia

sentry-apache-http-client-5/build.gradle.kts
```diff
@@ -21,6 +21,8 @@ tasks.withType<KotlinCompile>().configureEach {
 dependencies {
     api(project(":sentry"))
     api(Config.Libs.apacheHttpClient)
+    api(Config.Libs.apacheHttpCore)
+    api(Config.Libs.apacheHttpCoreH2)
```
marandaneto, line 25: do people need to access its classes? otherwise only `implementation` is fine

maciejwalkowiak: They don't need to but they may want to. We bring `ApacheTransportFactory` but they may want to create their custom factory with custom client configuration. I am not 100% sure if at some level of configuration they would not need to use lower level classes.

buildSrc/src/main/java/Config.kt
```diff
@@ -74,6 +74,8 @@ object Config {
         val servletApi = "javax.servlet:javax.servlet-api"
 
         val apacheHttpClient = "org.apache.httpcomponents.client5:httpclient5:5.0.3"
+        val apacheHttpCore = "org.apache.httpcomponents.core5:httpcore5:5.0.3"
+        val apacheHttpCoreH2 = "org.apache.httpcomponents.core5:httpcore5-h2:5.0.3"
```
marandaneto, line 78: we could create an internal `val apacheHttpVersion = "5.0.3"` and reuse it

marandaneto (approved):

buildSrc/src/main/java/Config.kt
maciejwalkowiak: The thing is that these versions are not really aligned. The same number is just an coincidence.

maciejwalkowiak:
Converting to draft until we decide that this is the way to go. The alternative is to upgrade httpcomponents-core directly to 5.1

marandaneto:
> Converting to draft until we decide that this is the way to go. The alternative is to upgrade httpcomponents-core directly to 5.1

if 5.0.3 fixes the bug, even though they said 5.1 is also compatible, not sure how extensible this has been tested as its a new release, it'd be safer just bumping the patch.

maciejwalkowiak:
5.0.3 fixes a bug, but i am not sure if this is the one we were facing. Just to double check I've filed an issue: https://issues.apache.org/jira/browse/HTTPCLIENT-2146

PR #1381, opened 2021-04-07
Device context for transactions

marandaneto:
## :pencil: Checklist

- [ ] No breaking changes

### DIFF
--- a/sentry/src/main/java/io/sentry/EventProcessor.java
+++ b/sentry/src/main/java/io/sentry/EventProcessor.java
@@ -3,7 +3,12 @@ package io.sentry;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+// other approach is to use EventProcessor<T extends SentryBaseEvent>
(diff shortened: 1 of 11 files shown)


sentry/src/main/java/io/sentry/EventProcessor.java
```diff
@@ -3,7 +3,12 @@
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+// other approach is to use EventProcessor<T extends SentryBaseEvent>
+// and having a processor which is the common fields and specific processors for Event and Transactions
+// not sure if our event processor interface should be deal with this
+// the downside using only the base class, if people want to process a field which is not part of SentryBaseEvent
+// they need to cast it manually
```
marandaneto: @bruno-garcia can we discuss this

marandaneto: after the discussion, we could try to use default methods for interfaces, return it so it's not a breaking change.

marandaneto:
favor of https://github.com/getsentry/sentry-java/pull/1430

PR #1389, opened 2021-04-09
Fix: `SentryTransactionAdvice` should operate on the new scope.

maciejwalkowiak:
## :scroll: Description

Fix: `SentryTransactionAdvice` should operate on the new scope.

`@SentryTransaction` is meant to be used for integrations that we do not support out of the box like messaging, scheduled jobs. These integrations work on thread pools, meaning if users modify the scope, without having push & pop scope changes would last also for further invocations happening on the same thread.

Unit tests.

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * Enhancement: Make NoOpHub public (#1379)
 * Fix: Do not bind transactions to scope by default. (#1376)
 * Fix: fix Hub thread safety (#1388)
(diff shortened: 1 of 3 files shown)


marandaneto:
@maciejwalkowiak changelog and mind updating the issue's description about why do we need that? I am fine with the changes but I don't know why its required to be in a new scope

maciejwalkowiak:
@marandaneto updated.

marandaneto (approved):

PR #1390, opened 2021-04-12
Fix: set user on the transaction.

### LINKED ISSUE #1335: User is not set on transactions

maciejwalkowiak:
## :scroll: Description

Set user on transaction.

Instead of setting user with event processor, user is set on the scope in Servlet filter - `SentryUserFilter`. This filter must run after Spring security (or any other filter verifying authentication details) so that `SentryUserProvider` implementations can access the currently logged in user data.

Sentry request flow

In Spring Boot integration, `SentryUserFilter` is set to run:

- by default runs as the last filter (lowest precedence)
- user can explicitly define the order

In Spring integration, filter bean must be created manually and a corresponding `DelegatingFilterProxy` servlet filter must be registered.

Fixes #1335 

In Spring integration, `User` used to be set on the events using `EventProcessors`. Since `Transactions` are not a subject to event processing, `user` has to be set on the `scope`, and then in `SentryClient`, scope data gets propagated to captured transactions.

## :pencil: Checklist

- [ ] No breaking changes

## Breaking Changes

In Spring integration, its not enough to register custom `SentryUserProvider` anymore. To propagate user information to events and transactions, in addition to `SentryUserProvider`, a `SentryUserFilter` must be registered.


sentry-samples/sentry-samples-spring/src/main/java/io/sentry/samples/spring/SentryConfig.java
```diff
@@ -32,4 +37,15 @@ TracesSamplerCallback tracesSamplerCallback() {
...
+  @Bean
+  SentryUserFilter sentryUserFilter(
+      final IHub hub, final List<SentryUserProvider> sentryUserProviders) {
+    return new SentryUserFilter(hub, sentryUserProviders);
+  }
+
+  @Bean
+  SpringSecuritySentryUserProvider springSecuritySentryUserProvider() {
+    return new SpringSecuritySentryUserProvider();
+  }
```
marandaneto, line 50: @NotNull or @Nullable annotations

sentry-samples/sentry-samples-spring/src/main/java/io/sentry/samples/spring/WebConfig.java
```diff
@@ -24,7 +24,7 @@
    * @return RestTemplate
    */
   @Bean
-  RestTemplate restTemplate(IHub hub) {
+  RestTemplate restTemplate(final IHub hub) {
```
marandaneto, line 27: always add @NotNull or @Nullable annotations too when touching code, so little by little, we'd have a better Kotlin interop. with Kotlin null-safety

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryAutoConfiguration.java
```diff
@@ -125,6 +125,53 @@
     @Open
     static class SentryWebMvcConfiguration {
 
+      @Bean
+      @ConditionalOnClass(SecurityContextHolder.class)
+      @Order(1)
+      public SpringSecuritySentryUserProvider springSecuritySentryUserProvider(
```
marandaneto, line 131: also @NotNull or @Nullable annotations for return types

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryAutoConfiguration.java
```diff
@@ -125,6 +125,53 @@
...
+       *
+       * @param hub the Sentry hub
+       * @param sentryProperties the Sentry properties
+       * @param sentryUserProvider the user provider
+       * @param securityProperties the Spring Security properties
+       * @return {@link SentryUserFilter} registration bean
+       */
+      @Bean
+      @ConditionalOnBean(SentryUserProvider.class)
+      FilterRegistrationBean<SentryUserFilter> sentryUserFilter(
```
marandaneto, line 150: @NotNull or @Nullable annotations return type

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryAutoConfiguration.java
```diff
@@ -125,6 +125,53 @@
...
+          final @NotNull SentryProperties sentryProperties,
+          final @NotNull Optional<SecurityProperties> securityProperties) {
+        return Optional.ofNullable(sentryProperties.getUserFilterOrder())
+            .orElseGet(
+                () ->
+                    securityProperties
+                        .map(
+                            props ->
+                                props.getFilter().getOrder()
+                                    + 10) // filter should run after security filters
```
marandaneto, line 171: mm, is it a magic number? how can we improve that?

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryProperties.java
```diff
@@ -26,6 +26,13 @@
   /** Report all or only uncaught web exceptions. */
   private int exceptionResolverOrder = 1;
 
+  /**
+   * Defines the {@link io.sentry.spring.SentryUserFilter} order. The default value is {@link
+   * org.springframework.core.Ordered.LOWEST_PRECEDENCE}, if Spring Security is auto-configured, its
+   * guaranteed to run after Spring Security filter chain.
+   */
+  private Integer userFilterOrder;
```
marandaneto, line 34: @NotNull or @Nullable annotations

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryProperties.java
```diff
@@ -67,6 +74,14 @@ public void setExceptionResolverOrder(int exceptionResolverOrder) {
     this.exceptionResolverOrder = exceptionResolverOrder;
   }
 
+  public Integer getUserFilterOrder() {
+    return userFilterOrder;
+  }
+
+  public void setUserFilterOrder(Integer userFilterOrder) {
+    this.userFilterOrder = userFilterOrder;
+  }
```
marandaneto, line 83: @NotNull or @Nullable annotations

sentry-spring/src/test/kotlin/io/sentry/spring/SentryUserFilterTest.kt
```diff
@@ -0,0 +1,156 @@
...
+
+    @Test
+    fun `sets provided user data on the scope`() {
+        val filter = fixture.getSut(userProviders = listOf(SentryUserProvider {
+            val user = User()
+            user.username = "john.doe"
+            user.id = "user-id"
+            user.ipAddress = "192.168.0.1"
+            user.email = "john.doe@example.com"
+            user.others = mapOf("key" to "value")
```
marandaneto, line 47: use `apply` always when possible, so we avoid all the `user.`

marandaneto:
most comments are nitpicking about @NotNull or @nullable annotations.
so feel free to do it all in one single fix.
it'd be nice to use Kotlin `apply` whenever possible.

as `No breaking changes` isn't marked, I'm wondering what would be required to enable such behavior, do we need docs? do we need a migration page? do we need a major/minor bump?

sentry-samples/sentry-samples-spring/src/main/java/io/sentry/samples/spring/WebConfig.java
maciejwalkowiak: These are samples. As I understand, we did not add these annotation to samples as they are not used by Kotlin code plus adding them makes the samples less approachable to users.

marandaneto:
I think this change conflicts with https://github.com/getsentry/sentry-java/pull/1381 which is just a Draft but kinda makes it event processors to be run so we could enrich transactions with more context.
@bruno-garcia we need to discuss that, if we really want event processors running for transactions too, this PR would not require this big change.

sentry-samples/sentry-samples-spring/src/main/java/io/sentry/samples/spring/SentryConfig.java
maciejwalkowiak: Same as https://github.com/getsentry/sentry-java/pull/1390#discussion_r612311867

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryAutoConfiguration.java
maciejwalkowiak: If you insist we can extract it to a constant but not sure how to name it. Filter order is defined with magic numbers in Spring Security (https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/SecurityProperties.java). Here we make sure that the order is higher, but we leave place for extra filters to be added.

We can also leave out this option, use only `LOWEST_PRECEDENCE` and give users a way to overwrite complete filter definition.

maciejwalkowiak:
Even with event processors, I think it's a better design (the way it should have been done from start), to set a `User` on the scope, rather than setting `User` with event processor.

marandaneto:
> Even with event processors, I think it's a better design (the way it should have been done from start), to set a `User` on the scope, rather than setting `User` with event processor.

sounds good, but my 1st question still applies:

I'm wondering what would be required to enable such behavior, do we need docs? do we need a migration page? do we need a major/minor bump? why is it a breaking change?

sentry/src/main/java/io/sentry/SentryClient.java
```diff
@@ -452,6 +452,7 @@ public void captureSession(final @NotNull Session session, final @Nullable Objec
     if (scope != null) {
       if (transaction.getRequest() == null) {
         transaction.setRequest(scope.getRequest());
+        transaction.setUser(scope.getUser());
```
marandaneto, line 455: this should have its own `if` block and not within `transaction.getRequest() == null`

maciejwalkowiak:
We need docs and migration page. The breaking change is that a filter now must be added in Spring integration.

sentry-samples/sentry-samples-spring/src/main/java/io/sentry/samples/spring/WebConfig.java
marandaneto: yep makes sense, my bad

marandaneto:
> We need docs and migration page. The breaking change is that a filter now must be added in Spring integration to set user data on events.

tbh I'm not really sure that this sounds like a good idea, user is such a basic requirement and should not require any additional steps other than the basic stuff, if we go this way, this should be even in the wizard so people won't miss it.

maciejwalkowiak:
Great majority of people use Spring Boot where things just work. When you use Spring, you must configure things manually. There is some space to configure things automatically like we do with `@EnableSentry`, but it has its limits.

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryAutoConfiguration.java
maciejwalkowiak: Removed this option and left LOWEST_PRECEDENCE since this will cover 99% of cases anyways. If users need to customize the order, they can set a value to `sentry.user-filter-order`

sentry-spring-boot-starter/src/test/kotlin/io/sentry/spring/boot/it/SentrySpringIntegrationTest.kt
```diff
@@ -147,6 +148,19 @@ class SentrySpringIntegrationTest {
...
 
+    @Test
+    fun `sets user on transaction`() {
+        val restTemplate = TestRestTemplate().withBasicAuth("user", "password")
+
+        restTemplate.getForEntity("http://localhost:$port/performance", String::class.java)
+
+        await.untilAsserted {
+            verify(transport).send(checkTransaction { transaction ->
+                assertThat(transaction.user).isNotNull()
```
bruno-garcia, line 159: `L:` Ideally we test it's the right user.

bruno-garcia (approved):

bruno-garcia:
> > We need docs and migration page. The breaking change is that a filter now must be added in Spring integration to set user data on events.
> 
> tbh I'm not really sure that this sounds like a good idea, user is such a basic requirement and should not require any additional steps other than the basic stuff (installing the SDK itself), if we go this way, this should be even in the wizard so people won't miss it.
> unless adding such filters is just the Spring way to go and people know about it.

This is indeed not great. We'll need to bump to 5.0.0 to land it then

marandaneto:
@maciejwalkowiak please update the docs migration page and required spring setup docs for adding such filter
https://github.com/getsentry/sentry-docs/issues/3405

sentry/src/main/java/io/sentry/SentryBaseEvent.java
```diff
@@ -198,4 +201,12 @@ public void setEnvironment(String environment) {
...
     this.platform = platform;
   }
+
+  public User getUser() {
+    return user;
+  }
+
+  public void setUser(User user) {
+    this.user = user;
+  }
```
marandaneto, line 211: lets add annotations @notnull @nullable

maciejwalkowiak:
Converting back to draft, since we need first to marge event processing on transactions.

marandaneto:
after discussion, we've decided that we wanted to find a mid-way rather than doing a breaking change, the important thing is, set the user to the scope (via event processor or not, it does not matter), and read the user from the scope when capturing the event, so clients don't need to add a filter.

maciejwalkowiak:
This fix will be reimplemented with event processor once https://github.com/getsentry/sentry-java/pull/1381 is finished & merged.

marandaneto:
> This fix will be reimplemented with event processor once #1381 is finished & merged.

lets please fix this before https://github.com/getsentry/sentry-java/pull/1381
this is a bigger change and will require way more things to be done.

PR #1397, opened 2021-04-13
Fix: use connection and read timeouts in ApacheHttpClient based transport

maciejwalkowiak:
## :scroll: Description

Fix: use connection and read timeouts in ApacheHttpClient based transport

When `ApacheHttpClientTransport` was used, timeouts were not respected.

Console sample and slowing down connections with ToxiProxy connected to Sentry relay running locally.

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/sentry-apache-http-client-5/src/test/kotlin/io/sentry/transport/apache/ApacheHttpClientTransportFactoryTest.kt
+++ b/sentry-apache-http-client-5/src/test/kotlin/io/sentry/transport/apache/ApacheHttpClientTransportFactoryTest.kt
@@ -11,10 +11,8 @@ import org.apache.hc.client5.http.impl.async.InternalHttpAsyncClient
 class ApacheHttpClientTransportFactoryTest {
 
     class Fixture {
-        fun getSut(options: SentryOptions = SentryOptions()): ApacheHttpClientTransport {
-            val factory = ApacheHttpClientTransportFactory()
-            return factory.create(options, mock()) as ApacheHttpClientTransport
-        }
+        fun getSut(options: SentryOptions = SentryOptions()) =
+            ApacheHttpClientTransportFactory().create(options, mock()) as ApacheHttpClientTransport
     }
 
     private val fixture = Fixture()
--- a/sentry-apache-http-client-5/src/test/kotlin/io/sentry/transport/apache/ApacheHttpClientTransportFactoryTest.kt
+++ b/sentry-apache-http-client-5/src/test/kotlin/io/sentry/transport/apache/ApacheHttpClientTransportFactoryTest.kt
@@ -10,21 +10,26 @@ import org.apache.hc.client5.http.impl.async.InternalHttpAsyncClient
 
 class ApacheHttpClientTransportFactoryTest {
 
+    class Fixture {
+        fun getSut(options: SentryOptions = SentryOptions()): ApacheHttpClientTransport {
+            val factory = ApacheHttpClientTransportFactory()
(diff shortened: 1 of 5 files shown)


sentry-apache-http-client-5/src/test/kotlin/io/sentry/transport/apache/ApacheHttpClientTransportFactoryTest.kt
```diff
@@ -11,13 +13,16 @@ class ApacheHttpClientTransportFactoryTest {
...
-        }
-        val requestDetails = mock<RequestDetails>()
-
-        val transport = factory.create(options, requestDetails)
+        val transport = factory.create(SentryOptions(), mock())
         assertNotNull(transport)
     }
+
+    @Test
+    fun `options timeouts are used when creating ApacheHttpClientTransport`() {
```
maciejwalkowiak, line 21: This test is very questionable. There's nothing we can assert here on the client itself. To write a proper test we would need to set up a mock http server with ability to set connect and response timeouts. Not sure if worth the effort.

marandaneto: agreed, can we actually read these values from `CloseableHttpAsyncClient` and see if they are the same as from options? it'd be more useful I guess

maciejwalkowiak: Only with reflection. There are pros & cons but after all, with a hope that httpclient won't change it internal structure much, i think it's the better way. Code updated.

sentry-apache-http-client-5/src/test/kotlin/io/sentry/transport/apache/ApacheHttpClientTransportFactoryTest.kt
```diff
@@ -1,23 +1,45 @@
...
 import kotlin.test.assertNotNull
+import org.apache.hc.client5.http.config.RequestConfig
+import org.apache.hc.client5.http.impl.async.InternalHttpAsyncClient
 
 class ApacheHttpClientTransportFactoryTest {
 
     @Test
     fun `creates ApacheHttpClientTransport`() {
+        val factory = ApacheHttpClientTransportFactory()
+        val transport = factory.create(SentryOptions(), mock())
```
marandaneto, line 16: this repeats in both tests, whats about making the Fixture and reusing it

sentry-apache-http-client-5/src/test/kotlin/io/sentry/transport/apache/ApacheHttpClientTransportFactoryTest.kt
```diff
@@ -1,23 +1,48 @@
...
+    private fun InternalHttpAsyncClient.getRequestConfig(): RequestConfig = this.getProperty("defaultConfig")
+
+    private inline fun <reified T> Any.getProperty(name: String): T =
+        try {
+            this::class.java.getDeclaredField(name)
+        } catch (_: NoSuchFieldException) {
+            this::class.java.superclass.getDeclaredField(name)
+        }.apply {
+            this.isAccessible = true
+        }.get(this) as T
```
marandaneto, line 47: this could go to the sentry-test-support module, reflection.kt, we already have some util functions there

marandaneto (approved):
left a comment but other than that, LGTM

marandaneto (approved):
LGTM

Issue #1395, opened 2021-04-13
Device's memory for Android NDK hard crashes is added on restart


marandaneto:
Device's memory for Android NDK crashes is added on the restart, so these values are not accurate.

Ideally, `sentry-native` would add this to the event before writing the event to the disk, so on restart, the Android SDK enriches only with static data.

marandaneto:
on the Android bits, we can read it from https://developer.android.com/reference/android/app/ActivityManager.MemoryInfo but if reading it directly on the NDK side, we'd need to figure this out and if it's actually safe to read such values when the App. is crashing because of OOM.

bruno-garcia:
Can't make JNI calls on the signal handler, so we should just make sure we don't add this on events read off of the disk after restart

marandaneto:
@bruno-garcia that's the 1st step, it'd be cool if the signal handler would be able to read free/used memory on the signal handler, so on the restart, we'd just merge the device info with the static data.

PR #1399, opened 2021-04-14
Fix: do not set free memory and is low memory fields when it's a NDK hard crash

marandaneto:
#1395

## :pencil: Checklist

- [ ] No breaking changes


marandaneto:
@bruno-garcia let's discuss this, there are more device data that is nonstatic and should not be applied if it's a hard crash.
I'm afraid that not setting them would make the device context too poor, lets decide case by case?

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -314,12 +316,15 @@ private void setArchitectures(final @NotNull Device device) {
     device.setModelId(Build.ID);
     setArchitectures(device);
 
+    // likely to be applied only if applyScopeData
```
marandaneto, line 319: the chances that the user opens the app again right away is high, so this won't change much

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -314,12 +316,15 @@ private void setArchitectures(final @NotNull Device device) {
...
     Intent batteryIntent = getBatteryIntent();
     if (batteryIntent != null) {
       device.setBatteryLevel(getBatteryLevel(batteryIntent));
       device.setCharging(isCharging(batteryIntent));
       device.setBatteryTemperature(getBatteryTemperature(batteryIntent));
     }
+
+    // likely to be applied only if applyScopeData
```
marandaneto, line 327: the chances that the user opens the app again right away is high, so this won't change much

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -347,12 +352,15 @@ private void setArchitectures(final @NotNull Device device) {
...
-      device.setLowMemory(memInfo.lowMemory);
+      if (applyScopeData) {
+        device.setFreeMemory(memInfo.availMem);
+        device.setLowMemory(memInfo.lowMemory);
+      }
       // there are runtime.totalMemory() and runtime.freeMemory(), but I kept the same for
       // compatibility
     }
 
+    // likely to be applied only if applyScopeData
```
marandaneto, line 363: the chances that the user opens the app again right away is high, so this won't change much

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -362,6 +370,7 @@ private void setArchitectures(final @NotNull Device device) {
       device.setFreeStorage(getUnusedInternalStorage(internalStorageStat));
     }
 
+    // likely to be applied only if applyScopeData
```
marandaneto, line 373: the chances that the user opens the app again right away is high, so this won't change much

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -376,7 +385,9 @@ private void setArchitectures(final @NotNull Device device) {
       device.setScreenDpi(displayMetrics.densityDpi);
     }
 
+    // likely to be applied only if applyScopeData
```
marandaneto, line 388: the chances that the user opens the app again right away is high, so this won't change much

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -385,6 +396,8 @@ private void setArchitectures(final @NotNull Device device) {
     if (device.getLanguage() == null) {
       device.setLanguage(Locale.getDefault().toString()); // eg en_US
     }
+
+    // likely to be applied only if applyScopeData
```
marandaneto, line 400: the chances that the user opens the app again right away is high, so this won't change much

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
bruno-garcia: Even though it's more likely that an app that crashes is reopened immediately by the user, I believe we can agree this is also not the case sometimes. And this sometimes multipled by the number of events we'll get through it'll be millions of false data.

That said, we could 'meet in the middle' and verify the timestamp of the event being within "n seconds" of the current time. And if so, apply any of such data.

Similar approach could be used with the performance product (transactions) for data that is costly to retrieve, so that we cache the value for that amount of seconds and hence, accept that the value could be slightly off from when the event actually happened.
How does that sound?

marandaneto: I still think the right approach would be doing that thru sentry-native
https://github.com/getsentry/sentry-native/issues/520
hence I'm going to remove free memory and is low memory on this PR, but let's collaborate on the other fields/approach in another PR.

bruno-garcia (approved):
Lgtm

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -302,7 +304,7 @@ private void setArchitectures(final @NotNull Device device) {
 
   // we can get some inspiration here
   // https://github.com/flutter/plugins/blob/master/packages/device_info/android/src/main/java/io/flutter/plugins/deviceinfo/DeviceInfoPlugin.java
-  private @NotNull Device getDevice() {
+  private @NotNull Device getDevice(final boolean applyScopeData) {
```
philipphofmann, line 307: `l`: Maybe we can find a better name for `applyScopeData`. We use it to control whether we add free and low memory. The name could reflect that.

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -347,8 +350,10 @@ private void setArchitectures(final @NotNull Device device) {
     if (memInfo != null) {
       // in bytes
       device.setMemorySize(getMemorySize(memInfo));
-      device.setFreeMemory(memInfo.availMem);
-      device.setLowMemory(memInfo.lowMemory);
+      if (applyScopeData) {
```
philipphofmann, line 353: `m`: I think it would be nice to add a comment here or somewhere else why we do this. Ideally, we could extract the bottom line from https://github.com/getsentry/sentry-java/issues/1395, so we know in the future why we did this.

sentry-android-core/src/test/java/io/sentry/android/core/DefaultAndroidEventProcessorTest.kt
```diff
@@ -290,4 +290,26 @@ class DefaultAndroidEventProcessorTest {
...
+
+    @Test
+    fun `When hint is not Cached, memory data should be applied`() {
+        val sut = fixture.getSut(context)
+
+        var event = SentryEvent()
+        event = sut.process(event, null)
+
+        assertNotNull(event.contexts.device!!.freeMemory)
+        assertNotNull(event.contexts.device!!.isLowMemory)
```
philipphofmann, line 313: `l`: Do we know the expected values for `freeMemory` and `isLowMemory`? If so, we could assert against them. This would make the test a bit better. Otherwise, just ignore me.

philipphofmann (approved):
LGTM

sentry-android-core/src/test/java/io/sentry/android/core/DefaultAndroidEventProcessorTest.kt
marandaneto: no, it can vary depending on robolectric tests

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
marandaneto: the naming cant be related to free memory, because the same flag would be used to every nonstatic data that cant be applied after the restart, so it's pretty much how we call on every integration that has such check, `applyScopeData`

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
marandaneto: makes sense, will do

PR #1406, opened 2021-04-14
Fix: set correct transaction status for unhandled exceptions in `SentryTracingFilter`.

maciejwalkowiak:
## :scroll: Description

Fix: set correct transaction status for unhandled exceptions in `SentryTracingFilter`.

For exceptions that have not been handled with `@ControllerAdvice` or `@ExceptionHandler`, the `response#status` in `SentryTracingFilter` is equal to `200`, which results in setting `SpanStatus.OK` on transaction.

With this change, we handle also the case when the exception is not handled and set correct status code.

Unit tests.

## :pencil: Checklist

- [x] No breaking changes


sentry-spring/src/main/java/io/sentry/spring/tracing/SentryTracingFilter.java
```diff
@@ -67,16 +67,22 @@ protected void doFilterInternal(
...
-        // we should not report it to Sentry
+        // and we should not report it to Sentry
         if (transactionName != null) {
           transaction.setName(transactionName);
           transaction.setOperation(TRANSACTION_OP);
-          transaction.setStatus(SpanStatus.fromHttpStatusCode(httpResponse.getStatus()));
+          // if exception has been thrown, transaction status is already set to INTERNAL_ERROR, and
+          // httpResponse.getStatus() returns 200.
+          if (transaction.getStatus() == null) {
+            transaction.setStatus(SpanStatus.fromHttpStatusCode(httpResponse.getStatus()));
```
marandaneto, line 84: pass a sensible default? okhttp uses `SpanStatus.INTERNAL_ERROR`

sentry-spring/src/main/java/io/sentry/spring/tracing/SentryTracingFilter.java
```diff
@@ -67,16 +67,22 @@ protected void doFilterInternal(
       final ITransaction transaction = startTransaction(httpRequest, sentryTraceHeader);
       try {
         filterChain.doFilter(httpRequest, httpResponse);
+      } catch (Exception e) {
+        // exceptions that are not handled by Spring
+        transaction.setStatus(SpanStatus.INTERNAL_ERROR);
```
marandaneto, line 72: should we `throw e;`?

maciejwalkowiak: There is no need to setThrowable as the exception is captured while transaction is active - SentryClient sets the span context on the event.

sentry-spring/src/main/java/io/sentry/spring/tracing/SentryTracingFilter.java
maciejwalkowiak: I am not sure if this is right. All 3xx, 422 and more responses would get INTERNAL_ERROR

marandaneto: better than no status at all?

marandaneto: https://github.com/getsentry/develop/issues/307

marandaneto: could be also `UNKNOWN`, but I'd rather change it everywhere to `UNKNOWN`

marandaneto: now its `UNKNOWN` if not a match, but other integrations do `INTERNAL_ERROR`, so I'd either keep this one as `INTERNAL_ERROR` or change all the others to use `UNKNOWN` as a default, I'd not like to differ here within the same SDK.

marandaneto: we recently had a discussion about that (during a call), cc @bruno-garcia 
we might not set any status actually because it'll fall back on the server side, so then we should remove any default I believe, wdyt @bruno-garcia

bruno-garcia: agree no status is better than possibly misleading status.

maciejwalkowiak: Reverted the change to not set status.

marandaneto (approved):
LGTM

PR #1409, opened 2021-04-15
Fix: set scope on transaction

maciejwalkowiak:
## :scroll: Description

Fix: set scope on transaction

Until now, only request has been set from the scope on transaction.

Unit tests

## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next steps

I believe applying scope deserves a refactoring. We can add a method on `SentryBaseEvent#applyScope(scope)` which will get overwritten in `SentryTransaction` and `SentryEvent`.

### DIFF
--- a/sentry/src/main/java/io/sentry/SentryClient.java
+++ b/sentry/src/main/java/io/sentry/SentryClient.java
@@ -1,6 +1,7 @@
 package io.sentry;
 
 import io.sentry.hints.DiskFlushNotification;
+import io.sentry.protocol.Contexts;
 import io.sentry.protocol.SentryId;
 import io.sentry.protocol.SentrySpan;
 import io.sentry.protocol.SentryTransaction;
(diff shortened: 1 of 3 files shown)


maciejwalkowiak:
cc @bruno-garcia

sentry/src/main/java/io/sentry/SentryClient.java
```diff
@@ -521,6 +490,37 @@ public void captureSession(final @NotNull Session session, final @Nullable Objec
     return event;
   }
 
+  private <T extends SentryBaseEvent> T applyScope(
+      final @NotNull T sentryBaseEvent, final @Nullable Scope scope) {
```
marandaneto, line 494: there's already a NPE check to call this method, hence `Scope` can be non null

maciejwalkowiak: For the transaction we use this method directly, that's why it's kept as nullable.

marandaneto (approved):

maciejwalkowiak:
Converting back to draft, since we need first to marge event processing on transactions.

maciejwalkowiak:
> Converting back to draft, since we need first to marge event processing on transactions.

Wrong PR.

PR #1424, opened 2021-04-19
Fix: Apply user from the scope to transaction.

maciejwalkowiak:
## :scroll: Description

Apply user from the scope to transaction.

#1335 Note: this does not solve the issue of setting user on transaction in Spring/Spring Boot integrations.

Unit tests.

## :pencil: Checklist

- [ ] No breaking changes

Adding nullability annotations to `user` makes it a breaking change for Kotlin users.

## :crystal_ball: Next steps

Fix setting user on transaction for Spring & Spring Boot.

### DIFF
--- a/sentry-spring-boot-starter/src/test/kotlin/io/sentry/spring/boot/it/SentrySpringIntegrationTest.kt
+++ b/sentry-spring-boot-starter/src/test/kotlin/io/sentry/spring/boot/it/SentrySpringIntegrationTest.kt
@@ -84,8 +84,8 @@ class SentrySpringIntegrationTest {
                 assertThat(event.request).isNotNull()
                 assertThat(event.request!!.url).isEqualTo("http://localhost:$port/hello")
                 assertThat(event.user).isNotNull()
-                assertThat(event.user.username).isEqualTo("user")
-                assertThat(event.user.ipAddress).isEqualTo("169.128.0.1")
+                assertThat(event.user!!.username).isEqualTo("user")
+                assertThat(event.user!!.ipAddress).isEqualTo("169.128.0.1")
             }, anyOrNull())
         }
     }
@@ -101,7 +101,8 @@ class SentrySpringIntegrationTest {
 
         await.untilAsserted {
             verify(transport).send(checkEvent { event ->
-                assertThat(event.user.ipAddress).isEqualTo("169.128.0.1")
+                assertThat(event.user).isNotNull()
+                assertThat(event.user!!.ipAddress).isEqualTo("169.128.0.1")
             }, anyOrNull())
         }
     }
--- a/sentry-spring/src/test/kotlin/io/sentry/spring/SentryUserProviderEventProcessorTest.kt
+++ b/sentry-spring/src/test/kotlin/io/sentry/spring/SentryUserProviderEventProcessorTest.kt
@@ -37,12 +37,13 @@ class SentryUserProviderEventProcessorTest {
         val result = processor.process(event, null)
 
         assertNotNull(result)
-        assertNotNull(result.user)
-        assertEquals("john.doe", result.user.username)
-        assertEquals("user-id", result.user.id)
-        assertEquals("192.168.0.1", result.user.ipAddress)
-        assertEquals("john.doe@example.com", result.user.email)
-        assertEquals(mapOf("key" to "value"), result.user.others)
+        assertNotNull(result.user) {
+            assertEquals("john.doe", it.username)
+            assertEquals("user-id", it.id)
+            assertEquals("192.168.0.1", it.ipAddress)
+            assertEquals("john.doe@example.com", it.email)
+            assertEquals(mapOf("key" to "value"), it.others)
+        }
     }
 
     @Test
@@ -62,12 +63,13 @@ class SentryUserProviderEventProcessorTest {
         val result = processor.process(event, null)
 
         assertNotNull(result)
-        assertNotNull(result.user)
-        assertEquals("john.doe", result.user.username)
-        assertEquals("user-id", result.user.id)
-        assertEquals("192.168.0.1", result.user.ipAddress)
-        assertEquals("john.doe@example.com", result.user.email)
-        assertEquals(mapOf("key" to "value"), result.user.others)
+        assertNotNull(result.user) {
+            assertEquals("john.doe", it.username)
+            assertEquals("user-id", it.id)
+            assertEquals("192.168.0.1", it.ipAddress)
+            assertEquals("john.doe@example.com", it.email)
+            assertEquals(mapOf("key" to "value"), it.others)
+        }
     }
 
     @Test
@@ -78,22 +80,24 @@ class SentryUserProviderEventProcessorTest {
         }
 
         val event = SentryEvent()
-        event.user = User()
-        event.user.username = "jane.smith"
-        event.user.id = "jane-smith"
-        event.user.ipAddress = "192.168.0.3"
-        event.user.email = "jane.smith@example.com"
-        event.user.others = mapOf("key" to "value")
+        event.user = User().apply {
+            username = "jane.smith"
+            id = "jane-smith"
+            ipAddress = "192.168.0.3"
+            email = "jane.smith@example.com"
+            others = mapOf("key" to "value")
+        }
 
         val result = processor.process(event, null)
 
         assertNotNull(result)
-        assertNotNull(result.user)
-        assertEquals("jane.smith", result.user.username)
-        assertEquals("jane-smith", result.user.id)
-        assertEquals("192.168.0.3", result.user.ipAddress)
-        assertEquals("jane.smith@example.com", result.user.email)
-        assertEquals(mapOf("key" to "value"), result.user.others)
+        assertNotNull(result.user) {
+            assertEquals("jane.smith", it.username)
+            assertEquals("jane-smith", it.id)
+            assertEquals("192.168.0.3", it.ipAddress)
+            assertEquals("jane.smith@example.com", it.email)
+            assertEquals(mapOf("key" to "value"), it.others)
+        }
     }
 
     @Test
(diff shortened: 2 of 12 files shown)


sentry/src/main/java/io/sentry/SentryBaseEvent.java
```diff
@@ -74,6 +75,9 @@
    */
   private @Nullable String platform;
 
+  /** Information about the user who triggered this event. */
+  private User user;
```
marandaneto, line 79:
```suggestion
  private @Nullable User user;
```

sentry-spring/src/test/kotlin/io/sentry/spring/SentryUserProviderEventProcessorTest.kt
```diff
@@ -37,12 +37,13 @@ class SentryUserProviderEventProcessorTest {
...
-        assertEquals("192.168.0.1", result.user.ipAddress)
-        assertEquals("john.doe@example.com", result.user.email)
-        assertEquals(mapOf("key" to "value"), result.user.others)
+        assertNotNull(result.user) {
+            assertEquals("john.doe", it.username)
+            assertEquals("user-id", it.id)
+            assertEquals("192.168.0.1", it.ipAddress)
+            assertEquals("john.doe@example.com", it.email)
+            assertEquals(mapOf("key" to "value"), it.others)
+        }
```
marandaneto, line 46: this block is duplicated across tests, maybe we extract it to a method called asserUser or something like that?

sentry/src/main/java/io/sentry/SentryClient.java
```diff
@@ -496,6 +496,9 @@ public void captureSession(final @NotNull Session session, final @Nullable Objec
       if (sentryBaseEvent.getRequest() == null) {
         sentryBaseEvent.setRequest(scope.getRequest());
       }
+      if (sentryBaseEvent.getUser() == null) {
+        sentryBaseEvent.setUser(scope.getUser());
+      }
```
marandaneto, line 501: what's about the combination with this check and `MainEventProcessor`? do they conflict with each other when `isSendDefaultPii` is enabled? I believe one of them will never be executed if it sets to a non-null value

marandaneto:
@maciejwalkowiak this PR cannot be released alone though, if it reads from Scope but integrations don't set the user request to the Scope, Scope.user will be always null

sentry/src/main/java/io/sentry/SentryClient.java
maciejwalkowiak: If user is not set on the event, `SentryClient` sets it from the scope. Then in `MainEventProcessor` if user is null it sets the user with a default ip address, if its not null but the ip address is null it sets the default ip address. What conflict do you mean?

maciejwalkowiak:
@marandaneto for anyone who is not using integrations but does manual instrumentation this PR does provide a value. As I understand no matter how we solve the integrations, scope should be applied to transactions anyways.

sentry/src/main/java/io/sentry/SentryClient.java
marandaneto: > @marandaneto for anyone who is not using integrations but does manual instrumentation this PR does provide a value. As I understand no matter how we solve the integrations, scope should be applied to transactions anyways.

I agree, but ideally this PR or another one should fix for our integrations too, see https://github.com/getsentry/sentry-java/issues/1335#issuecomment-814891298

marandaneto:
> @marandaneto for anyone who is not using integrations but does manual instrumentation this PR does provide a value. As I understand no matter how we solve the integrations, scope should be applied to transactions anyways.

I agree, but ideally this PR or another one should fix for our integrations too, see https://github.com/getsentry/sentry-java/issues/1335#issuecomment-814891298

sentry/src/main/java/io/sentry/SentryClient.java
marandaneto: indeed, apply scope happens before event processors, that's why, the other way around would be problematic I believe.

marandaneto: a problem that I see is, MainEventProcessor takes into consideration the `isSendDefaultPii` and sets IP address with `DEFAULT_IP_ADDRESS`, but that won't happen for transactions, this can be fixed later IMO when we run event processors for transactions too.

marandaneto:
what we could do is: creating the Spring filter that sets the user to the Scope and add it to https://docs.sentry.io/platforms/java/guides/spring/performance/#enable-tracing
so the error tracking works as it is, but the Spring (without spring boot) requires this additional step to fix the issue.

wdyt @maciejwalkowiak @bruno-garcia ?

maciejwalkowiak:
@marandaneto yes I agree. Please note, that adding extra filter for Spring integration is not as big problem as it may look. Using Spring without Spring Boot, means that users must configure a lot of things manually, and while it is possible to configure some things that happen inside Spring context (like `@EnableSentry` does), it's idiomatic to add filters manually. 

To simplify user experience we could have just a single filter that does everything, but then the performance measurement would not start as early as it is right now.

sentry-spring/src/test/kotlin/io/sentry/spring/SentryUserProviderEventProcessorTest.kt
maciejwalkowiak: I think in case of this test class duplicated code is a good trade off. Having fixed string values in `assertUser` method although resolves duplication, doesn't feel right to couple this method with test methods. Feel free to refactor it if you like.

sentry/src/main/java/io/sentry/SentryClient.java
bruno-garcia: We could start setting the default value even if a user is set, but the IpAddress is null.
I don't believe the "setting user semantics" dictates it's a all or nothing thing, or does it?

Can't we merge values instead? 

If I do `captureEvent(Event{ user = User(name))` while having `sendDefaultPii=true` I'd still like the the `{{auto}}` IP Address to be added to the user. I just set the `name` because that's all I have.

bruno-garcia (approved):
👍 
Though as Manoel pointed out, we'll miss the default IP address at times. Left a note on that

maciejwalkowiak:
@bruno-garcia the question is should we lift setting the IP address from `MainEventProcessor` to `SentryClient` or wait till event processors support transactions?

marandaneto:
> @bruno-garcia the question is should we lift setting the IP address from `MainEventProcessor` to `SentryClient` or wait till event processors support transactions?

let's not block this PR because of that, here is the issue to follow up later when we execute event processors https://github.com/getsentry/sentry-java/issues/1427

sentry-spring-boot-starter/src/test/kotlin/io/sentry/spring/boot/it/SentrySpringIntegrationTest.kt
```diff
@@ -84,8 +84,8 @@ class SentrySpringIntegrationTest {
                 assertThat(event.request).isNotNull()
                 assertThat(event.request!!.url).isEqualTo("http://localhost:$port/hello")
                 assertThat(event.user).isNotNull()
-                assertThat(event.user.username).isEqualTo("user")
-                assertThat(event.user.ipAddress).isEqualTo("169.128.0.1")
+                assertThat(event.user!!.username).isEqualTo("user")
+                assertThat(event.user!!.ipAddress).isEqualTo("169.128.0.1")
```
marandaneto, line 88: maybe use `assertNotNull` with the lambda so we avoid `!!` for all user asserts on this PR?

sentry-spring/src/test/kotlin/io/sentry/spring/SentryUserProviderEventProcessorTest.kt
marandaneto: tbh I'm not a big fan of that, it's the same strings and same asserts across every test, every time you'd need to change either a value or adding/removing a field, you've to change it everywhere.
I know @philipphofmann has strong opinions about that too, it's not scalable when the number of tests increases.

philipphofmann: I agree 100% with @marandaneto. Treat your test code as production code. A few duplications are fine, but in this case, I would extract them to a method.

marandaneto: I will merge as it is because I am blocked by this PR, but lets please discuss and refactor this as soon as we touch this class again

bruno-garcia: I'm not willing to die on this hill so bare with me:

I believe a test should be as self-sufficient as possible. When a test calls out to a private method `getUser` then `assertUser` it isn't clear what `user` is, and how `assert` works, unless I peek into those methods. It also means that I can be setting the user now thinking of a particular test, but be negatively affecting another, without knowing, since multiple tests "change" by means of modifying these common functions.

It has the advantage, obviously, of having a single place to "fix" things, or even to add a new property to be set and asserted. It  But that comes with the cost of potentially changing a test that wasn't supposed to be changed.

Perhaps what we're missing here is a library that validates two objects are "equivalent". It solves the problem of adding a new property to the field and also expecting that to be mapped through which is a common mistake when mapping objects by hand.

Both these principles/approaches are used in the .NET SDK. The assertion library (introduced by @Tyrrrz, who from what I noticed also follows the idea of self-sufficient tests) is called `FluentAssertions `, is there something similar for Java?

All of that said, this class in particular does have a lot of duplication for basic object mapping, to the point that this diff was mainly changes to asserting `user`. I'd be very tempted to at least create a `assertUserEquals(User a, User b)` to get rid of this duplication.

marandaneto: Kotlin generates `equals` and `hashCode` for data classes, but Java does not, we'd need to implement and bloat the SDK with that (which is gonna be used only in tests), so we could do an equality check.

another approach is having a test per field, instead of the whole user every time.

I do agree that's a trade-off, but I often spend more time scaling the tests other than "understanding" them.

PR #1426, opened 2021-04-20
Fix: set status and associate events with transactions

maciejwalkowiak:
## :scroll: Description

Set status when exception is thrown during transaction execution, and associate exceptions with transactions.

## :bulb: Motivation and Context

#1351 

## :green_heart: How did you test it?

Unit & Integration tests

## :pencil: Checklist

- [x] I reviewed the submitted code
- [x] I added tests to verify the changes
- [ ] I updated the docs if needed
- [x] No breaking changes

PR #1429, opened 2021-04-21
Fix setting transaction name on events sent after transaction is over.

maciejwalkowiak:
## :scroll: Description

Fix setting transaction name on events sent after transaction is over.

Events that got triggered after span/transaction is over, but are connected with a span, must have also `transaction` field set, so that they are searchable in the UI by transaction name.

Also, without `transaction` field set, the red error box is not shown on the transaction details page (like below)
image

This topic has been discussed on Discord: https://discord.com/channels/621778831602221064/777210445442187274/834151506620973058

Unit tests.

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/sentry/src/main/java/io/sentry/HubAdapter.java
+++ b/sentry/src/main/java/io/sentry/HubAdapter.java
@@ -3,7 +3,6 @@ package io.sentry;
 import io.sentry.protocol.SentryId;
 import io.sentry.protocol.SentryTransaction;
 import io.sentry.protocol.User;
-import io.sentry.util.Pair;
 import java.util.List;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
@@ -179,8 +178,11 @@ public final class HubAdapter implements IHub {
   }
 
   @Override
-  public void setSpanContext(final @NotNull Throwable t, final @NotNull Pair<ISpan, String> sc) {
-    Sentry.getCurrentHub().setSpanContext(t, sc);
+  public void setSpanContext(
+      final @NotNull Throwable t,
+      final @NotNull ISpan span,
+      final @NotNull String transactionName) {
+    Sentry.getCurrentHub().setSpanContext(t, span, transactionName);
   }
 
   @Override
--- a/sentry/src/main/java/io/sentry/HubAdapter.java
+++ b/sentry/src/main/java/io/sentry/HubAdapter.java
@@ -179,10 +179,10 @@ public final class HubAdapter implements IHub {
 
   @Override
   public void setSpanContext(
-      final @NotNull Throwable t,
+      final @NotNull Throwable throwable,
       final @NotNull ISpan span,
       final @NotNull String transactionName) {
-    Sentry.getCurrentHub().setSpanContext(t, span, transactionName);
+    Sentry.getCurrentHub().setSpanContext(throwable, span, transactionName);
   }
 
   @Override
--- a/sentry/src/main/java/io/sentry/HubAdapter.java
+++ b/sentry/src/main/java/io/sentry/HubAdapter.java
@@ -3,6 +3,7 @@ package io.sentry;
 import io.sentry.protocol.SentryId;
 import io.sentry.protocol.SentryTransaction;
 import io.sentry.protocol.User;
+import io.sentry.util.Pair;
 import java.util.List;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
@@ -178,7 +179,7 @@ public final class HubAdapter implements IHub {
   }
 
   @Override
-  public void setSpanContext(final @NotNull Throwable t, final @NotNull ISpan sc) {
+  public void setSpanContext(final @NotNull Throwable t, final @NotNull Pair<ISpan, String> sc) {
     Sentry.getCurrentHub().setSpanContext(t, sc);
   }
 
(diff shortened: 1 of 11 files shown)


sentry/src/main/java/io/sentry/IHub.java
```diff
@@ -402,14 +403,14 @@ ITransaction startTransaction(
...
+   * Associates {@link ISpan} and the transaction name with the {@link Throwable}. Used to determine
+   * in which trace the exception has been thrown in framework integrations.
    *
    * @param throwable the throwable
-   * @param span the span context
+   * @param span a pair of span context and transaction name
    */
   @ApiStatus.Internal
-  void setSpanContext(@NotNull Throwable throwable, @NotNull ISpan span);
+  void setSpanContext(@NotNull Throwable throwable, @NotNull Pair<ISpan, String> span);
```
maciejwalkowiak, line 413: Perhaps we can come up with a dedicated class, instead of `Pair`. I am open for suggestions as I could not figure out a class name that would add value.

maciejwalkowiak:
Changelog is not needed, as this PR is going to be merged to https://github.com/getsentry/sentry-java/pull/1426 that already has a changelog entry.

sentry/src/main/java/io/sentry/util/Pair.java
```diff
@@ -0,0 +1,39 @@
...
+    if (first != null ? !first.equals(pair.first) : pair.first != null) return false;
+    return second != null ? second.equals(pair.second) : pair.second == null;
+  }
+
+  @Override
+  public int hashCode() {
+    int result = first != null ? first.hashCode() : 0;
+    result = 31 * result + (second != null ? second.hashCode() : 0);
+    return result;
+  }
```
marandaneto, line 38: do we need that if it's just a container?

sentry/src/main/java/io/sentry/IHub.java
marandaneto: Pair or Tuple are fine

sentry/src/main/java/io/sentry/util/Pair.java
```diff
@@ -0,0 +1,39 @@
+package io.sentry.util;
+
+import org.jetbrains.annotations.Nullable;
+
+public final class Pair<A, B> {
```
marandaneto, line 5: can we make it package-private or mark it as internal

sentry/src/test/java/io/sentry/HubTest.kt
```diff
@@ -314,7 +314,7 @@ class HubTest {
         val exception = RuntimeException()
         val span = mock<Span>()
         whenever(span.spanContext).thenReturn(SpanContext("op"))
-        sut.setSpanContext(exception, span)
+        sut.setSpanContext(exception, io.sentry.util.Pair(span, "tx-name"))
```
marandaneto, line 317: import the class instead full package

maciejwalkowiak: Kotlin Pair is already used in this class.

sentry/src/main/java/io/sentry/util/Pair.java
maciejwalkowiak: Simplifies equality check in tests

sentry/src/main/java/io/sentry/util/Pair.java
maciejwalkowiak: Package private only if we move it out to root package, so i guess making internal is the way to go.

sentry/src/main/java/io/sentry/util/Pair.java
marandaneto: true but better asserting field per field in tests, we don't want to bloat the SDK with stuff used only in tests

sentry/src/test/java/io/sentry/HubTest.kt
marandaneto: we can still import and use an alias

`import kotlin.test.assertEquals as hi`

sentry/src/main/java/io/sentry/IHub.java
marandaneto, line 413: I know this method is `@ApiStatus.Internal` but if we don't want to leak the `Pair` (due to package visibility) outside of the package, we could get 2 params and handle the Pair within the Hub class

maciejwalkowiak:
@marandaneto all fixes applied, please re-review

sentry/src/main/java/io/sentry/HubAdapter.java
```diff
@@ -178,8 +178,11 @@ public IHub clone() {
   }
 
   @Override
-  public void setSpanContext(final @NotNull Throwable t, final @NotNull ISpan sc) {
-    Sentry.getCurrentHub().setSpanContext(t, sc);
+  public void setSpanContext(
+      final @NotNull Throwable t,
```
marandaneto, line 182:
```suggestion
      final @NotNull Throwable throwable,
```

marandaneto:
@maciejwalkowiak changelog

maciejwalkowiak:
@marandaneto as mentioned above:

> Changelog is not needed, as this PR is going to be merged to #1426 that already has a changelog entry.

marandaneto:
> @marandaneto as mentioned above:
> 
> > Changelog is not needed, as this PR is going to be merged to #1426 that already has a changelog entry.

ups my bad, then use `_#skip-changelog_`, make CI happier

marandaneto (approved):

PR #1430, opened 2021-04-21
Fix: Run event processors and enrich transactions with contexts

marandaneto:
## :scroll: Description
Fix: Run event processors and enrich transactions with contexts

Event processors were not run for transactions, so contexts were missing.
Some contexts are heavy for Android, hence there's a flag to not do IO nor IPC if it's a transaction.
Spring User is also run, hence transactions for the Spring integration will contain the user info.
Some fields from Event were moved to BaseEvent so transactions contain more data like breadcrumbs, extra, etc...

running and unit tests

## :pencil: Checklist

- [ ] No breaking changes

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
@@ -22,6 +22,7 @@ import android.util.DisplayMetrics;
 import io.sentry.DateUtils;
 import io.sentry.EventProcessor;
 import io.sentry.ILogger;
+import io.sentry.SentryBaseEvent;
 import io.sentry.SentryEvent;
 import io.sentry.SentryLevel;
 import io.sentry.android.core.util.ConnectivityChecker;
@@ -34,6 +35,7 @@ import io.sentry.protocol.DebugMeta;
 import io.sentry.protocol.Device;
 import io.sentry.protocol.OperatingSystem;
 import io.sentry.protocol.SentryThread;
+import io.sentry.protocol.SentryTransaction;
 import io.sentry.protocol.User;
 import io.sentry.util.ApplyScopeUtils;
 import io.sentry.util.Objects;
@@ -139,15 +141,38 @@ final class DefaultAndroidEventProcessor implements EventProcessor {
   @Override
   public @NotNull SentryEvent process(
       final @NotNull SentryEvent event, final @Nullable Object hint) {
-    if (ApplyScopeUtils.shouldApplyScopeData(hint)) {
+    if (shouldApplyScopeData(event, hint)) {
       processNonCachedEvent(event);
+      mergeDebugImages(event);
+      setThreads(event);
+    }
+
+    setCommons(event, true);
+
+    return event;
+  }
+
+  private void setCommons(final @NotNull SentryBaseEvent event, final boolean doIO) {
+    mergeUser(event);
+    setDevice(event, doIO);
+    mergeOS(event);
+    setSideLoadedInfo(event);
+  }
+
+  private boolean shouldApplyScopeData(
+      final @NotNull SentryBaseEvent event, final @Nullable Object hint) {
+    if (ApplyScopeUtils.shouldApplyScopeData(hint)) {
+      return true;
     } else {
       logger.log(
           SentryLevel.DEBUG,
           "Event was cached so not applying data relevant to the current app execution/version: %s",
           event.getEventId());
+      return false;
     }
+  }
 
+  private void mergeUser(final @NotNull SentryBaseEvent event) {
     // userId should be set even if event is Cached as the userId is static and won't change anyway.
     final User user = event.getUser();
     if (user == null) {
@@ -155,19 +180,15 @@ final class DefaultAndroidEventProcessor implements EventProcessor {
     } else if (user.getId() == null) {
       user.setId(getDeviceId());
     }
-
-    if (event.getContexts().getDevice() == null) {
-      event.getContexts().setDevice(getDevice());
-    }
-
-    mergeOS(event);
-
-    setSideLoadedInfo(event);
-
-    return event;
   }
 
-  private void mergeOS(final @NotNull SentryEvent event) {
+  private void setDevice(final @NotNull SentryBaseEvent event, final boolean doIO) {
+    if (event.getContexts().getDevice() == null) {
+      event.getContexts().setDevice(getDevice(doIO));
+    }
+  }
+
+  private void mergeOS(final @NotNull SentryBaseEvent event) {
     final OperatingSystem currentOS = event.getContexts().getOperatingSystem();
     final OperatingSystem androidOS = getOperatingSystem();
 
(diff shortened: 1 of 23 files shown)


sentry-spring/src/main/java/io/sentry/spring/SentryUserProviderEventProcessor.java
```diff
@@ -13,6 +13,7 @@
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+// TODO: check if I need to support process(SentryTransaction) here
```
marandaneto, line 16: @maciejwalkowiak

sentry/src/main/java/io/sentry/SentryEvent.java
```diff
@@ -80,9 +64,12 @@
    * <p>```json { "fingerprint": ["myrpc", "POST", "/foo.bar"] }
    */
   private List<String> fingerprint;
+
+  // TODO: should breadcrumbs be part of a transaction?
```
marandaneto, line 68: @bruno-garcia

sentry/src/main/java/io/sentry/SentryEvent.java
```diff
@@ -80,9 +64,12 @@
...
   /** List of breadcrumbs recorded before this event. */
   private List<Breadcrumb> breadcrumbs;
 
+  // TODO: should extras be part of a transaction?
```
marandaneto, line 72: @bruno-garcia

marandaneto:
@bruno-garcia can we discuss this?

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -302,7 +331,7 @@ private void setArchitectures(final @NotNull Device device) {
 
   // we can get some inspiration here
   // https://github.com/flutter/plugins/blob/master/packages/device_info/android/src/main/java/io/flutter/plugins/deviceinfo/DeviceInfoPlugin.java
-  private @NotNull Device getDevice() {
+  private @NotNull Device getDevice(final boolean doIO) {
```
marandaneto, line 334: figure out better naming for the var

sentry-spring/src/main/java/io/sentry/spring/SentryUserProviderEventProcessor.java
marandaneto: does it make sense to run this event processor for transactions too?

sentry/src/main/java/io/sentry/SentryEvent.java
marandaneto: does make sense

sentry/src/main/java/io/sentry/SentryEvent.java
marandaneto: does make sense

sentry-spring/src/main/java/io/sentry/spring/SentryUserProviderEventProcessor.java
maciejwalkowiak: If it would run for transactions we would not need to add user filter - so I guess yes

marandaneto:
@bruno-garcia @maciejwalkowiak please review

sentry/src/main/java/io/sentry/EventProcessor.java
```diff
@@ -1,9 +1,15 @@
...
 
+import io.sentry.protocol.SentryTransaction;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 public interface EventProcessor {
   @Nullable
   SentryEvent process(@NotNull SentryEvent event, @Nullable Object hint);
+
+  @Nullable
```
maciejwalkowiak, line 11: I think this interface can be improvemed. Either we can also make `process` for `SentryEvent` return `event` by default, leaving user the choice which methods should be implemented, or make `EventProcessor` a parameterized type `T extends SentryBaseEvent` - this API would probably be the cleanest.

marandaneto: `T extends SentryBaseEvent` was the first option, but it was a breaking change.
making `SentryEvent` also default would make sense though if a processor is only about transactions, makes sense

marandaneto: made `process(SentryEvent event, @Nullable Object hint)` also a default method

bruno-garcia (approved):

marandaneto:
@maciejwalkowiak could you please check that I didn't break anything in the backend integrations for error monitoring and performance? thanks
Also test that it fixes https://github.com/getsentry/sentry-java/issues/1335

sentry-spring/src/test/kotlin/io/sentry/spring/SentryUserProviderEventProcessorTest.kt
```diff
@@ -200,4 +206,13 @@ class SentryUserProviderEventProcessorTest {
             assertNull(it.ipAddress)
         }
     }
+
+    @Test
+    fun `Transaction sets the user`() {
```
maciejwalkowiak, line 211: Either the name of this test is off or the test itself is not finished, as it does not seem that it is testing if user is set on the transaction. Also maybe `user is set on transaction` rather than `transaction sets the user`?

sentry/src/main/java/io/sentry/EventProcessor.java
```diff
@@ -1,9 +1,17 @@
 package io.sentry;
 
+import io.sentry.protocol.SentryTransaction;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 public interface EventProcessor {
   @Nullable
-  SentryEvent process(@NotNull SentryEvent event, @Nullable Object hint);
+  default SentryEvent process(@NotNull SentryEvent event, @Nullable Object hint) {
```
maciejwalkowiak, line 9: As much as I understand and agree with the implementation, I think this interface is less obvious now since it can process both events and transactions and both have default implementation - Javadoc explaining the purpose and what does it do would be nice.

sentry/src/test/java/io/sentry/MainEventProcessorTest.kt
```diff
@@ -342,6 +341,36 @@ class MainEventProcessorTest {
         assertEquals("optionsHost", event.serverName)
     }
 
+    @Test
+    fun `Transaction sets server name`() {
```
maciejwalkowiak, line 345: I don't follow the naming. `server name is set on transaction` and so on I think better reflects whats happening here.

maciejwalkowiak:
As mentioned on Discord, unfortunately event processing for transactions doesn't save us from having a user filter. When transaction is finished, user is cleared from the thread local already. 

I'll now just double check if nothing is off with the integrations.

maciejwalkowiak (approved):
Other than these small things mentioned above, LGTM

sentry-spring/src/test/kotlin/io/sentry/spring/SentryUserProviderEventProcessorTest.kt
marandaneto: good catch

sentry/src/main/java/io/sentry/EventProcessor.java
marandaneto: makes sense

sentry/src/test/java/io/sentry/MainEventProcessorTest.kt
marandaneto: indeed, naming could be improved

PR #1439, opened 2021-04-23
Add nullability annotations to `sentry` module.

maciejwalkowiak:
## :scroll: Description

Adds `@Nullable` & `@NotNull` annotations to classes and interfaces in `sentry` module.

Better Kotlin interoperatibility and cleaner API.

With `NullAway` Errorprone plugin.

I did not unfortunately find a way to ensure that API is 100% documented. Errorprone triggers violation if the null safety is violated in the code, but if the code is not used, it won't trigger an error.

## :pencil: Checklist

- [ ] No breaking changes

Breaking changes affect mainly Kotlin users.

## :crystal_ball: Next steps

Add nullability annotations to other modules - as PRs to `nullability` branch.


bruno-garcia:
Excited about this change!

sentry-android-core/src/main/java/io/sentry/android/core/AndroidLogger.java
```diff
@@ -11,7 +11,10 @@
 
   @SuppressWarnings("AnnotateFormatMethod")
   @Override
-  public void log(SentryLevel level, String message, Object... args) {
+  public void log(
+      final @Nullable SentryLevel level,
```
maciejwalkowiak, line 15: Perhaps these parameters should not be `nullable`?

maciejwalkowiak:
@marandaneto have a look please. Since this is a breaking change anyway, perhaps we can re-evaluate if some parameters should/should not be nullable

sentry-android-core/src/main/java/io/sentry/android/core/AndroidLogger.java
bruno-garcia: Agreed, can't think of a use case were we can't provide a log here and we really need one to make sense of things.
Otherwise we'll need to add some default and odd stuff happens

sentry/src/main/java/io/sentry/ILogger.java
```diff
@@ -31,7 +31,11 @@
    * @param message The message.
    * @param args the formatting arguments
    */
-  void log(SentryLevel level, Throwable throwable, String message, Object... args);
+  void log(
+      @Nullable SentryLevel level,
+      @Nullable Throwable throwable,
+      @Nullable String message,
```
bruno-garcia, line 37: IMO `level` and `message` should be required.
One argument is that if I have a throwable I might now need to pass a message but perhaps run a search on the code base right now and see if we have any of such cases.

sentry/src/main/java/io/sentry/IScopeObserver.java
```diff
@@ -1,18 +1,20 @@
...
 /** Observer for the sync. of Scopes across SDKs */
 public interface IScopeObserver {
-  void setUser(User user);
+  void setUser(@Nullable User user);
 
-  void addBreadcrumb(Breadcrumb crumb);
+  void addBreadcrumb(@NotNull Breadcrumb crumb);
 
-  void setTag(String key, String value);
+  void setTag(@NotNull String key, @Nullable String value);
```
bruno-garcia, line 13: I believe Sentry is unhappy with a `null` value and we have an API to `removeTag` so here we're probably better off with `NotNull`

sentry/src/main/java/io/sentry/Scope.java
```diff
@@ -112,10 +112,14 @@ public void setLevel(final @Nullable SentryLevel level) {
...
    */
-  public void setTransaction(final @NotNull String transaction) {
+  public void setTransaction(final @Nullable String transaction) {
     final ITransaction tx = this.transaction;
     if (tx != null) {
-      tx.setName(transaction);
+      if (transaction != null) {
+        tx.setName(transaction);
+      } else {
+        this.setTransaction((ITransaction) null);
```
bruno-garcia, line 121: I think this could be unwanted side effect.

If I have a Sentry error monitoring instrumentation that sets transaction to null, I add Sentry performance monitoring and I'd be kicking out a transaction bound to the scope without knowing.

Sounds like there are not good solutions because of this odd state where we bind an object adn a string to represent transaction due to backward compatibility but at least we should log a warning here, but potentially not doing this. Just setting the scope transaction to null and not dropping the transaction object

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
```diff
@@ -48,6 +48,7 @@
   }
 
   // TODO: Should be a Stream
+  @SuppressWarnings("NullAway")
```
bruno-garcia, line 51: What is it warning about exactly? And why do we ignore?
```suggestion
  @SuppressWarnings("NullAway") // Suppress because ...
```

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -266,7 +266,7 @@
    * deduplication prevents from receiving the same exception multiple times when there is more than
    * one framework active that captures errors, for example Logback and Spring Boot.
    */
-  private Boolean enableDeduplication = true;
+  private @Nullable Boolean enableDeduplication = true;
```
bruno-garcia, line 269: Since we have a default, can't we get away with this being `NotNull`?

sentry/src/main/java/io/sentry/Session.java
```diff
@@ -73,7 +73,7 @@ public Session(
       final @Nullable String ipAddress,
       final @Nullable String userAgent,
       final @Nullable String environment,
-      final @NotNull String release) {
+      final @Nullable String release) {
```
bruno-garcia, line 76: A session without a release is invalid. Why does need to be Nullable?

sentry/src/main/java/io/sentry/SessionAdapter.java
```diff
@@ -73,7 +73,9 @@ public void write(JsonWriter writer, Session value) throws IOException {
     boolean hasInitAttrs = false;
     hasInitAttrs = initAttrs(writer, hasInitAttrs);
 
-    writer.name("release").value(value.getRelease());
+    if (value.getRelease() != null) {
```
bruno-garcia, line 76: This would result in a broken session update (if it's null)

bruno-garcia (commented):
Made it half the way through.. need to continue later

sentry/src/main/java/io/sentry/IScopeObserver.java
maciejwalkowiak: Good spot. This was my mistake.

sentry/src/main/java/io/sentry/Scope.java
maciejwalkowiak: Maybe `transaction` name parameter can be set to `@NotNull`? We still give option to explicitly call `Scope#clearTransaction`.

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
maciejwalkowiak: `dataFactory` is a `Callable` which returns theoretically a nullable result. Our implementations always provide non-null values. We can handle potential `null`, but how to handle it? Log and return an empty byte array from this method?

sentry/src/main/java/io/sentry/SentryOptions.java
maciejwalkowiak: This is unfortunately coming from binding external configuration. I think it was a mistake to bind external config directly to `SentryOptions` - there should be an object dedicated to it, which can be applied to `SentryOptions`. 

I am happy to prepare a PR for it in case you believe it makes sense.

sentry/src/main/java/io/sentry/Session.java
maciejwalkowiak: `SentryOptions#release` is nullable. If the release cannot be null on sessions, we need to handle it somehow then. Perhaps @marandaneto has ideas?

sentry/src/main/java/io/sentry/Scope.java
bruno-garcia: Sounds like that's the best solution. @marandaneto ?

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
bruno-garcia: If I read it right we currently return an empty byte array.

The note was mainly to suggest adding a comment:

```suggestion
  // dataFactory is a Callable which returns theoretically a nullable result. Our implementations always provide non-null values. 
  // In the even that it did return null, we compensate by returning an empty byte array (and logging an error)
  @SuppressWarnings("NullAway")
```

sentry/src/main/java/io/sentry/SentryOptions.java
bruno-garcia: if we change it to `boolean` is it possible to bind? assuming in that case it would either set as-is (false) or set the value if it existed.

We can discuss changing this but for the sake of this PR then I guess its fine as Nullable if no other way

sentry/src/main/java/io/sentry/cache/CacheStrategy.java
```diff
@@ -218,7 +219,10 @@ private void moveInitFlagIfNecessary(
     return null;
   }
 
-  private @Nullable Session getFirstSession(final @NotNull SentryEnvelope envelope) {
+  private @Nullable Session getFirstSession(final @Nullable SentryEnvelope envelope) {
```
bruno-garcia, line 222: As a private method can't we make this NotNull and check at the caller?

sentry/src/main/java/io/sentry/exception/InvalidDsnException.java
```diff
@@ -1,29 +1,32 @@
...
 
+import org.jetbrains.annotations.Nullable;
+
 public final class InvalidDsnException extends RuntimeException {
   private static final long serialVersionUID = 412945154259913013L;
-  private final String dsn;
+  private final @Nullable String dsn;
 
-  public InvalidDsnException(String dsn) {
+  public InvalidDsnException(final @Nullable String dsn) {
```
bruno-garcia, line 9: Do we currently have a use case for this to be Nullable?
Ideally it would be non nullable.

Or best yet since this is a major bump: Get rid of this exception and save up on some code. Simplifies things with `new RuntimeException("Invalid DSN: " + dsn)`

@marandaneto wdyt?

it's not like someone is expecting InvalidDsnException to handle it somehow, are we?

sentry/src/main/java/io/sentry/transport/RateLimiter.java
```diff
@@ -203,7 +203,10 @@ public void updateRetryAfterLimits(
               for (final String catItem : categories) {
                 DataCategory dataCategory = DataCategory.Unknown;
                 try {
-                  dataCategory = DataCategory.valueOf(StringUtils.capitalize(catItem));
+                  final String catItemCapitalized = StringUtils.capitalize(catItem);
+                  if (catItemCapitalized != null) {
+                    dataCategory = DataCategory.valueOf(catItemCapitalized);
+                  }
```
bruno-garcia, line 209:
```suggestion
                  }
                  else {
                    logger.log(ERROR, "Couldn't capitalize: %s", catItem);
                  }
```

Also, looks like like 211 should be ERROR, no?

bruno-garcia (commented):
Made it to the end 🥳 

I did skim through the tests, and the protocol is pretty much all `Nullable` so also skimmed throuhg.

Other than the points I've raised it LGTM

sentry-android-core/src/main/java/io/sentry/android/core/AndroidLogger.java
```diff
@@ -11,18 +11,28 @@
...
+      final @Nullable String message,
```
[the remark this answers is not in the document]
marandaneto: agreed, the only problem is if we log the result of something instead of hardcoded value, eg, the response from the API, we'd need to check NPE before calling the method `log`, its a trade-off, I guess making it nullable makes it dev. friendlier, but doing NPE check within the method has its performance impacts too.

sentry/src/main/java/io/sentry/Breadcrumb.java
```diff
@@ -39,7 +39,7 @@
    *
    * @param timestamp the timestamp
    */
-  public Breadcrumb(final @Nullable Date timestamp) {
+  public Breadcrumb(final @NotNull Date timestamp) {
```
marandaneto, line 42: its recommended but not mandatory https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/
wondering if there's an use case where it makes sense not having a date

sentry/src/main/java/io/sentry/EnvelopeReader.java
```diff
@@ -87,6 +87,10 @@
             deserializeEnvelopeItemHeader(
                 envelopeBytes, itemHeaderStartOffset, lineBreakIndex - itemHeaderStartOffset);
 
+        if (itemHeader == null) {
+          throw new IllegalArgumentException(
+              "Item header at index '" + items.size() + "' is null.");
+        }
         if (itemHeader.getLength() <= 0) {
```
marandaneto, line 94: I guess it'd be fine to just do a `||` op here

sentry/src/main/java/io/sentry/Hub.java
```diff
@@ -177,18 +181,18 @@ private void assignTraceContext(final @NotNull SentryEvent event) {
     if (event.getThrowable() != null) {
       final Pair<ISpan, String> pair = throwableToSpan.get(event.getThrowable());
       if (pair != null) {
-        if (event.getContexts().getTrace() == null) {
+        if (event.getContexts().getTrace() == null && pair.getFirst() != null) {
           event.getContexts().setTrace(pair.getFirst().getSpanContext());
         }
-        if (event.getTransaction() == null) {
+        if (event.getTransaction() == null && pair.getSecond() != null) {
           event.setTransaction(pair.getSecond());
```
marandaneto, line 188: lets cache `pair.getFirst()` and `pair.getSecond()` to a var instead of calling twice, we avoid multiple lookups

sentry/src/main/java/io/sentry/Hub.java
```diff
@@ -260,7 +264,10 @@ public void close() {
             ((Closeable) integration).close();
           }
         }
-        options.getExecutorService().close(options.getShutdownTimeout());
+        final ISentryExecutorService executorService = options.getExecutorService();
+        if (executorService != null) {
```
marandaneto, line 268: `executorService` isnt nullable

sentry/src/main/java/io/sentry/Hub.java
```diff
@@ -653,7 +660,7 @@ public void setSpanContext(
   SpanContext getSpanContext(final @NotNull Throwable throwable) {
     Objects.requireNonNull(throwable, "throwable is required");
     final Pair<ISpan, String> span = this.throwableToSpan.get(throwable);
-    if (span != null) {
+    if (span != null && span.getFirst() != null) {
       return span.getFirst().getSpanContext();
```
marandaneto, line 664: multiple lookups

sentry/src/main/java/io/sentry/MainEventProcessor.java
```diff
@@ -191,11 +191,10 @@ private void setThreads(final @NotNull SentryEvent event) {
       // crashed properly
       List<Long> mechanismThreadIds = null;
 
-      final boolean hasExceptions =
-          event.getExceptions() != null && !event.getExceptions().isEmpty();
+      final List<SentryException> eventExceptions = event.getExceptions();
```
marandaneto, line 194: wondering why we've changed the `hasExceptions` instead of checking NPE and emptiness 2 times

sentry/src/main/java/io/sentry/NoOpSentryExecutorService.java
```diff
@@ -0,0 +1,23 @@
...
+import java.util.concurrent.Future;
+import java.util.concurrent.FutureTask;
+import org.jetbrains.annotations.NotNull;
+
+final class NoOpSentryExecutorService implements ISentryExecutorService {
+  private static final NoOpSentryExecutorService instance = new NoOpSentryExecutorService();
+
+  private NoOpSentryExecutorService() {}
+
+  public static ISentryExecutorService getInstance() {
```
marandaneto, line 12:
```suggestion
  public static @NotNull ISentryExecutorService getInstance() {
```

sentry/src/main/java/io/sentry/Scope.java
marandaneto: having `Scope#clearTransaction` makes sense, wondering if it'd exist such an use case that you want to clean the String but not the transaction itself, I don't think so, so not nullable makes sense for me

sentry/src/main/java/io/sentry/SendFireAndForgetEnvelopeSender.java
```diff
@@ -26,7 +26,7 @@ public SendFireAndForgetEnvelopeSender(
     Objects.requireNonNull(options, "SentryOptions is required");
 
     final String dirPath = sendFireAndForgetDirPath.getDirPath();
-    if (!hasValidPath(dirPath, options.getLogger())) {
+    if (dirPath == null || !hasValidPath(dirPath, options.getLogger())) {
```
marandaneto, line 29: `hasValidPath` already checks for NPE

sentry/src/main/java/io/sentry/SendFireAndForgetOutboxSender.java
```diff
@@ -26,7 +26,7 @@ public SendFireAndForgetOutboxSender(
     Objects.requireNonNull(options, "SentryOptions is required");
 
     final String dirPath = sendFireAndForgetDirPath.getDirPath();
-    if (!hasValidPath(dirPath, options.getLogger())) {
+    if (dirPath == null || !hasValidPath(dirPath, options.getLogger())) {
```
marandaneto, line 29: same as above

sentry/src/main/java/io/sentry/Sentry.java
```diff
@@ -210,14 +210,16 @@ private static boolean initConfigurations(final @NotNull SentryOptions options)
     // eg release, distinctId, sentryClientName
 
     // this should be after setting serializers
-    if (options.getCacheDirPath() != null && !options.getCacheDirPath().isEmpty()) {
+    if (options.getOutboxPath() != null
+        && options.getCacheDirPath() != null
+        && !options.getCacheDirPath().isEmpty()) {
```
marandaneto, line 215: we should break into 2 `ifs` here, the Java SDK may work without the `outboxPath`

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
```diff
@@ -247,11 +248,15 @@ public CachedItem(final @Nullable Callable<byte[]> dataFactory) {
...
-      if (bytes == null) {
+    public @NotNull byte[] getBytes() throws Exception {
+      if (bytes == null && dataFactory != null) {
         bytes = dataFactory.call();
       }
-      return bytes;
+      return orEmptyArray(bytes);
+    }
+
+    private static @NotNull byte[] orEmptyArray(final @Nullable byte[] bytes) {
```
marandaneto, line 258: it does not need to be static

sentry/src/main/java/io/sentry/Session.java
marandaneto: there's already checks in the code to not create a session if release isn't available, where is it breaking it notnull?

sentry/src/main/java/io/sentry/SessionAdapter.java
marandaneto: side effect of https://github.com/getsentry/sentry-java/pull/1439/files#r625319745
should not be nullable

sentry/src/main/java/io/sentry/cache/CacheStrategy.java
```diff
@@ -120,13 +120,13 @@ private void moveInitFlagIfNecessary(
       final @NotNull File currentFile, final @NotNull File[] notDeletedFiles) {
     final SentryEnvelope currentEnvelope = readEnvelope(currentFile);
 
-    if (!isValidEnvelope(currentEnvelope)) {
+    if (currentEnvelope != null && !isValidEnvelope(currentEnvelope)) {
```
marandaneto, line 123: `isValidEnvelope` already does NPE check

sentry/src/main/java/io/sentry/cache/CacheStrategy.java
marandaneto: getFirstSession is only called if `isValidEnvelope` validates it, so it should be not nullable already

sentry/src/main/java/io/sentry/cache/EnvelopeCache.java
```diff
@@ -54,8 +55,22 @@
 
   private final @NotNull Map<SentryEnvelope, String> fileNameMap = new WeakHashMap<>();
 
-  public EnvelopeCache(final @NotNull SentryOptions options) {
-    super(options, options.getCacheDirPath(), options.getCacheDirSize());
+  public static IEnvelopeCache create(final @NotNull SentryOptions options) {
```
marandaneto, line 58:
```suggestion
  public static @NotNull IEnvelopeCache create(final @NotNull SentryOptions options) {
```

sentry/src/main/java/io/sentry/exception/InvalidDsnException.java
marandaneto: I'd go for `IllegalArgumentException` instead though

sentry/src/main/java/io/sentry/transport/RateLimiter.java
marandaneto: makes sense

marandaneto:
- missing changelog.
- there are a few methods missing a return nullability annotation.
- there are a few changes that actually may introduce new behaviors, they are not covered by unit tests
- major junk of work is done, well done

sentry/src/main/java/io/sentry/SentryOptions.java
maciejwalkowiak: If we change it to `boolean` we will not know if it was set externally to true, or if its a default value. Leaving a nullable then.

sentry/src/main/java/io/sentry/MainEventProcessor.java
maciejwalkowiak: `NullAway` is not able to recognize this check. If we want to keep using it, we would either have to do it as in PR, or disable warning for this method.

sentry/src/main/java/io/sentry/SendFireAndForgetEnvelopeSender.java
maciejwalkowiak: Same reason as here https://github.com/getsentry/sentry-java/pull/1439#discussion_r628094104

sentry/src/main/java/io/sentry/SendFireAndForgetOutboxSender.java
maciejwalkowiak: Same reason as here https://github.com/getsentry/sentry-java/pull/1439#discussion_r628094104

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
maciejwalkowiak: Since it does not access the state, isn't it a good candidate for a static method?

sentry/src/main/java/io/sentry/Session.java
maciejwalkowiak: `Hub#startSession` -> `Scope#startSession` - startSession uses `SentryOptions#release` which is a nullable field. Technically its possible that null sneaks in into a session. We can either disable the null check for this method or make `Scope#startSession` return `null` or throw if the `release` is `null`.

sentry/src/main/java/io/sentry/cache/CacheStrategy.java
maciejwalkowiak: Fixed

sentry/src/main/java/io/sentry/exception/InvalidDsnException.java
maciejwalkowiak: `InvalidDsnException` removed.

sentry/src/main/java/io/sentry/Breadcrumb.java
maciejwalkowiak: Then we just need a decision. Right now when passing `null` as timestamp, `getTimestamp` throws NPE.

sentry/src/main/java/io/sentry/Sentry.java
maciejwalkowiak: 👍

maciejwalkowiak:
Few comments left to resolve and take decision. The rest is fixed.

sentry/src/main/java/io/sentry/Session.java
bruno-garcia: Can we return null in that case?

sentry/src/main/java/io/sentry/Breadcrumb.java
bruno-garcia: Agree with `NotNull` here.

Sentry tries to be very lenient on what it receives but we can require this value from the SDK point of view.
One could always use the current time when instantiating. This is anyway more an internal API (should be IMHO) since users usually don't care about date unless writing some integration that's reading old logs (with dates) and creating crumbs.

Ideally we'd have overloads even such as: `Sentry.addBreadcrumb("message")` so there's no need to instantiate a `Breadcrumb` to begin with.

sentry/src/main/java/io/sentry/Hub.java
bruno-garcia: What lookup happens with these methods? Isn't this just a method call on an object? That under the hood return a reference? If that's the case, and there's no concurrency, it could be OK to just call the getter twice.

sentry/src/main/java/io/sentry/cache/CacheStrategy.java
bruno-garcia: I believe the static analyzer can't follow the flow of null check through the private method calls so the check is needed to make the compiler happy. aka: tool limitation

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
bruno-garcia: if access no state, doesn't override a base method, best to be static

sentry/src/main/java/io/sentry/Session.java
maciejwalkowiak: Changed to return `null`.

maciejwalkowiak:
`Session#release` changed back to `@NotNull`.

bruno-garcia:
Merging this to move things! Looks like nothing major pending, but lets work on follow up PRs if something comes up

PR #1443, opened 2021-04-27
Fix: set user on transaction

### LINKED ISSUE #1335: User is not set on transactions

maciejwalkowiak:
## :scroll: Description

Set user on transaction.

User is set on the scope in Servlet filter - `SentryUserFilter`. This filter must run after Spring security (or any other filter verifying authentication details) so that `SentryUserProvider` implementations can access the currently logged in user data.

Sentry request flow

In Spring Boot integration, `SentryUserFilter` is set to run:

- by default runs as the last filter (lowest precedence)
- user can explicitly define the order

In Spring integration, filter bean must be created manually and a corresponding `DelegatingFilterProxy` servlet filter must be registered.

Fixes #1335 

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 * Fix: Pass maxBreadcrumbs config. to sentry-native (#1425)
 * Fix: Run event processors and enrich transactions with contexts (#1430)
 * Bump: sentry-native to 0.4.9 (#1431)
-* Fix: Set user on transaction (#1443)
+* Fix: Set user on transaction in Spring & Spring Boot integrations (#1443)
 
 # 4.4.0-alpha.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Fix: Pass maxBreadcrumbs config. to sentry-native (#1425)
 * Fix: Run event processors and enrich transactions with contexts (#1430)
 * Bump: sentry-native to 0.4.9 (#1431)
+* Fix: Set user on transaction (#1443)
 
 # 4.4.0-alpha.2
 
--- a/sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryAutoConfiguration.java
+++ b/sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryAutoConfiguration.java
@@ -12,9 +12,11 @@ import io.sentry.protocol.SdkVersion;
 import io.sentry.spring.SentryExceptionResolver;
 import io.sentry.spring.SentryRequestResolver;
 import io.sentry.spring.SentrySpringRequestListener;
+import io.sentry.spring.SentryUserFilter;
 import io.sentry.spring.SentryUserProvider;
 import io.sentry.spring.SentryUserProviderEventProcessor;
 import io.sentry.spring.SentryWebConfiguration;
+import io.sentry.spring.SpringSecuritySentryUserProvider;
 import io.sentry.spring.tracing.SentryAdviceConfiguration;
 import io.sentry.spring.tracing.SentrySpanPointcutConfiguration;
 import io.sentry.spring.tracing.SentryTracingFilter;
@@ -22,6 +24,8 @@ import io.sentry.spring.tracing.SentryTransactionPointcutConfiguration;
 import io.sentry.transport.ITransportGate;
 import io.sentry.transport.apache.ApacheHttpClientTransportFactory;
 import java.util.List;
+import java.util.Optional;
+import javax.servlet.http.HttpServletRequest;
 import org.aspectj.lang.ProceedingJoinPoint;
 import org.jetbrains.annotations.NotNull;
 import org.springframework.beans.factory.ObjectProvider;
@@ -42,6 +46,7 @@ import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
 import org.springframework.core.Ordered;
 import org.springframework.core.annotation.Order;
+import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.web.client.RestTemplate;
 
 @Configuration(proxyBeanMethods = false)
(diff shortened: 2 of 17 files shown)


sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryAutoConfiguration.java
```diff
@@ -125,6 +130,43 @@
     @Open
     static class SentryWebMvcConfiguration {
 
+      @Bean
```
bruno-garcia, line 133: Missing Javadocs on this one?

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 * Fix: Pass maxBreadcrumbs config. to sentry-native (#1425)
 * Fix: Run event processors and enrich transactions with contexts (#1430)
 * Bump: sentry-native to 0.4.9 (#1431)
+* Fix: Set user on transaction (#1443)
```
bruno-garcia, line 15: This change affects only Spring, right?

```suggestion
* Fix: Set user on transaction on Spring (#1443)
```

sentry-spring/api/sentry-spring.api
```diff
@@ -54,6 +54,12 @@ public class io/sentry/spring/SentrySpringServletContainerInitializer : javax/se
 	public fun onStartup (Ljava/util/Set;Ljavax/servlet/ServletContext;)V
 }
 
+public class io/sentry/spring/SentryUserFilter : javax/servlet/Filter {
```
bruno-garcia, line 57: Looks like we need to add these to the docs and a note on a migration plan for the next version (should be 5.0 I believe)?

sentry-spring/src/main/java/io/sentry/spring/SentryUserFilter.java
```diff
@@ -0,0 +1,80 @@
...
+      if (IpAddressUtils.isDefault(user.getIpAddress())) {
+        // unset {{auto}} as it would set the server's ip address as a user ip address
+        user.setIpAddress(null);
+      }
+    }
+    hub.configureScope(scope -> scope.setUser(user));
+    chain.doFilter(request, response);
+  }
+
+  private void apply(final @NotNull User existingUser, final @Nullable User user) {
```
bruno-garcia, line 59: Reading this method alone wasn't clear what `user` was in relation to `existingUser`.

What about something more specific like:

```suggestion
  private void apply(final @NotNull User existingUser, final @Nullable User userFromProvider) {
```

Or Maybe `existingUser` could be named `scopeUser`

sentry-spring-boot-starter/src/test/kotlin/io/sentry/spring/boot/SentryAutoConfigurationTest.kt
```diff
@@ -319,11 +320,12 @@ class SentryAutoConfigurationTest {
         contextRunner.withPropertyValues("sentry.dsn=http://key@localhost/proj", "sentry.send-default-pii=true")
             .withConfiguration(UserConfigurations.of(SentryUserProviderConfiguration::class.java))
             .run {
-                val options = it.getBean(SentryProperties::class.java)
-                val userProviderEventProcessors = options.eventProcessors.filterIsInstance<SentryUserProviderEventProcessor>()
-                assertEquals(2, userProviderEventProcessors.size)
-                assertTrue(userProviderEventProcessors[0].sentryUserProvider is HttpServletRequestSentryUserProvider)
-                assertTrue(userProviderEventProcessors[1].sentryUserProvider is CustomSentryUserProvider)
+                val options = it.getBean("sentryUserFilter", FilterRegistrationBean::class.java).filter as SentryUserFilter
+                val userProviderEventProcessors = options.sentryUserProviders
```
marandaneto, line 324: this is duplicated across tests, we could extract to `internal getUserProviderEventProcessors(it)`

sentry-spring/src/main/java/io/sentry/spring/SentryUserFilter.java
```diff
@@ -0,0 +1,80 @@
...
+    if (user != null) {
+      Optional.ofNullable(user.getEmail()).ifPresent(existingUser::setEmail);
+      Optional.ofNullable(user.getId()).ifPresent(existingUser::setId);
+      Optional.ofNullable(user.getIpAddress()).ifPresent(existingUser::setIpAddress);
+      Optional.ofNullable(user.getUsername()).ifPresent(existingUser::setUsername);
+      if (user.getOthers() != null && !user.getOthers().isEmpty()) {
+        if (existingUser.getOthers() == null) {
+          existingUser.setOthers(new ConcurrentHashMap<>());
+        }
+        for (Map.Entry<String, String> entry : user.getOthers().entrySet()) {
```
marandaneto, line 69:
```suggestion
        for (final Map.Entry<String, String> entry : user.getOthers().entrySet()) {
```

sentry-spring/src/main/java/io/sentry/spring/SentryUserFilter.java
```diff
@@ -0,0 +1,80 @@
...
+          existingUser.getOthers().put(entry.getKey(), entry.getValue());
```
marandaneto, line 70: should we check if the `key` already exists or it has precedence?

sentry-spring/src/main/java/io/sentry/spring/SentryUserProviderEventProcessor.java
```diff
@@ -15,6 +15,10 @@
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+/**
+ * Deprecated: instead of using event processor for setting user on events and transactions, user
+ * should be set on the scope using {@link SentryUserFilter}.
+ */
```
marandaneto, line 21: I don't follow this comment, this filter should be only to address the missing user from transactions and not events, events should not be affected otherwise its a breaking change experience even for events or am I missing something?

Also, we should annotate with `@Deprecated`

sentry-spring/src/test/kotlin/io/sentry/spring/SentryUserFilterTest.kt
```diff
@@ -0,0 +1,142 @@
...
+import org.springframework.mock.web.MockHttpServletRequest
+import org.springframework.mock.web.MockHttpServletResponse
+
+class SentryUserFilterTest {
+    class Fixture {
+        val hub = mock<IHub>()
+        val request = MockHttpServletRequest()
+        val response = MockHttpServletResponse()
+        val chain = mock<FilterChain>()
+        val options = SentryOptions()
```
marandaneto, line 25: hub and options are not used outside of the method `getSut`, so they can be hidden in there, not really exposed here

marandaneto:
@maciejwalkowiak we need docs for this

sentry-spring/src/main/java/io/sentry/spring/SentryUserFilter.java
maciejwalkowiak: Changed to `userFromProvider`.

sentry-spring/src/main/java/io/sentry/spring/SentryUserFilter.java
maciejwalkowiak: It has a precedence.

sentry-spring/src/main/java/io/sentry/spring/SentryUserProviderEventProcessor.java
maciejwalkowiak: This means that using this filter alone solves the problem of setting user on events and transactions. We keep it to not break codebases that do not add the filter. In long term: users should use filter and not this event processor.

sentry-spring/src/test/kotlin/io/sentry/spring/SentryUserFilterTest.kt
maciejwalkowiak: Refactored the filter a bit, to use `hub` instead of configuring scope.

sentry-spring/src/main/java/io/sentry/spring/SentryUserProviderEventProcessor.java
marandaneto: ah ok, `SentryInitBeanPostProcessor` still adds it, so `SentryUserProviderEventProcessor` will get executed for the Spring integration/events not causing a breaking change.

marandaneto (approved):
LGTM

PR #1463, opened 2021-05-05
Fix: associate event with transaction when thrown exception is not a direct cause

### LINKED ISSUE #1418: Event is not associated with transaction when thrown exception is not a direct cause
In `Hub` where span context is associated to an event there should be check if any cause of the event exception is not associated with a span.

The issue becomes a problem when integrating for example with Spring AMQP, where every exception thrown from the listener is wrapped with `ListenerExecutionFailedException`.

maciejwalkowiak:
## :scroll: Description

Fix: associate event with transaction when thrown exception is not a direct cause

Fixes #1418

Unit tests.

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/sentry/src/main/java/io/sentry/util/ExceptionUtils.java
+++ b/sentry/src/main/java/io/sentry/util/ExceptionUtils.java
@@ -1,0 +1,23 @@
+package io.sentry.util;
+
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+@ApiStatus.Internal
+public final class ExceptionUtils {
+
+  /**
+   * Returns exception root cause or the exception itself if there are no causes
+   *
+   * @param throwable - the throwable
+   * @return the root cause
+   */
+  public static @NotNull Throwable findRootCause(final @NotNull Throwable throwable) {
+    Objects.requireNonNull(throwable, "throwable cannot be null");
+    Throwable rootCause = throwable;
+    while (rootCause.getCause() != null && rootCause.getCause() != rootCause) {
+      rootCause = rootCause.getCause();
+    }
+    return rootCause;
+  }
+}
--- a/sentry/src/test/java/io/sentry/HubTest.kt
+++ b/sentry/src/test/java/io/sentry/HubTest.kt
@@ -341,6 +341,23 @@ class HubTest {
     }
 
     @Test
+    fun `when captureEvent is called and event has exception which non-root cause has been previously attached with span context, sets span context to the event`() {
+        val (sut, mockClient) = getEnabledHub()
+        val rootCause = RuntimeException()
+        val exceptionAssignedToSpan = RuntimeException(rootCause)
+        val span = mock<Span>()
+        whenever(span.spanContext).thenReturn(SpanContext("op"))
+        sut.setSpanContext(exceptionAssignedToSpan, span, "tx-name")
+
+        val event = SentryEvent(RuntimeException(exceptionAssignedToSpan))
+
+        val hint = { }
+        sut.captureEvent(event, hint)
+        assertEquals(span.spanContext, event.contexts.trace)
+        verify(mockClient).captureEvent(eq(event), any(), eq(hint))
+    }
+
+    @Test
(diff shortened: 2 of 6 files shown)


sentry/src/main/java/io/sentry/util/ExceptionUtils.java
```diff
@@ -0,0 +1,23 @@
...
+   * @return the root cause
+   */
+  public static @NotNull Throwable findRootCause(final @NotNull Throwable throwable) {
+    Objects.requireNonNull(throwable, "throwable cannot be null");
+    Throwable rootCause = throwable;
+    while (rootCause.getCause() != null && rootCause.getCause() != rootCause) {
+      rootCause = rootCause.getCause();
+    }
+    return rootCause;
+  }
```
marandaneto, line 22: while this would fix the problem, it'd also be a breaking change and introduce new issues.

See the use case:

`NotFoundException(IOException)`
`InternalError(IOException)`

both would be linked as `IOException` but either the App. wrapped into a known type or 3rd party libs did it, not sure if it's a good idea to do it here, but rather doing the check before setting the Span-Exception to the Hub WeakRef Map.

@bruno-garcia wdyt?

maciejwalkowiak: What do you mean by "linked as"? Sentry events stay the same, its only about associating them with spans. I don't see how this is a breaking change

marandaneto: there's an association between Span and Throwable right.
we're changing the behavior now, from the Throwable itself to its cause, given the example above with a dummy `NotFoundException` and `InternalError`.

Maybe a better solution would be either, getting the cause in such integrations (eg Spring AMQP and integrations that always wrap up in a generic Throwable) and passing to `throwableToSpan` the correct Throwable, the cause.

2nd, maybe a bit more expensive, which is setting to `throwableToSpan` every `Throwable` and its cause.

eg:
The example: NotFoundException(IOException)
`throwableToSpan` will contain 2 entries, with the `NotFoundException` and `IOException`

marandaneto: lets see what @bruno-garcia says

maciejwalkowiak: I am still not getting it. Can you show an example test that would be passing now, but fails after this change?

marandaneto: tbh I am not sure how to explain differently, maybe we jump on a call? DM me

bruno-garcia: Good idea to discuss this on a call. Lets sync up if y'all didn't yet

sentry/src/test/java/io/sentry/HubTest.kt
```diff
@@ -324,6 +324,22 @@ class HubTest {
         verify(mockClient).captureEvent(eq(event), any(), eq(hint))
     }
 
+    @Test
+    fun `when captureEvent is called and event has exception which root cause has been previously attached with span context, sets span context to the event`() {
```
bruno-garcia, line 328: We're missing a test where we called `setThrowable` passing a throwable that itself wasn't the root one

sentry/src/main/java/io/sentry/util/ExceptionUtils.java
bruno-garcia: Manoel and I talked btw on a call. We realized that we need to consider the root also when adding to the weak ref, but the type of the exception doesn't play a role so I think we can mark this one as resolved (I added a note for the other point)

sentry/src/test/java/io/sentry/HubTest.kt
maciejwalkowiak: Got it.

maciejwalkowiak:
I thought it through over the weekend and realised the missing scenario. Please take a look @marandaneto

marandaneto (approved):
LGTM

PR #1466, opened 2021-05-07
Feat: App start up and slow/frozen frames metrics

marandaneto:
## :scroll: Description
Just a POC

## :pencil: Checklist

- [ ] No breaking changes


sentry-android-core/build.gradle.kts
```diff
@@ -79,6 +79,8 @@ dependencies {
     // lifecycle processor, session tracking
     implementation(Config.Libs.lifecycleProcess)
     implementation(Config.Libs.lifecycleCommonJava8)
+    // core is a transitive dependency of appcompat btw
+    implementation(Config.Libs.androidxCore)
```
marandaneto, line 83: androidx core is a transitive dependency of the appcompat package which contains the `AppCompatActivity` so I believe almost every app already has this dependency anyway

marandaneto: either we keep it as it is and deal with dependency conflicts or we create a new package and tell people to add it, wdyt?

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -196,11 +225,29 @@ public synchronized void onActivityCreated(
...
     addBreadcrumb(activity, "started");
   }
 
+  //  private boolean isHardwareAccelerated(Activity activity) {
+  //    // we can't observe frame rates for a non hardware accelerated view
+  //    return activity.getWindow() != null
+  //        && ((activity.getWindow().getAttributes().flags
+  //                & WindowManager.LayoutParams.FLAG_HARDWARE_ACCELERATED)
+  //            != 0);
+  //  }
```
marandaneto, line 241: I read that reading slow and frozen frames are only possible if `FLAG_HARDWARE_ACCELERATED` is enabled, but it worked with it disabled during tests, so I'm not sure if I should keep this check or not.
`FLAG_HARDWARE_ACCELERATED` is disabled by default by the way

sentry-android-core/src/main/java/io/sentry/android/core/PerformanceAndroidEventProcessor.java
```diff
@@ -0,0 +1,35 @@
...
+      @NotNull SentryTransaction transaction, @Nullable Object hint) {
+    if (!AppStartState.getInstance().isSentStartMetric()) {
+      Long appStartUpInterval = AppStartState.getInstance().getAppStartInterval();
+      if (appStartUpInterval != null) {
+        MeasurementValue value = new MeasurementValue((float) appStartUpInterval);
+
+        String appStartKey =
+            AppStartState.getInstance().getColdStart() ? "app_start_cold" : "app_start_warm";
+
+        transaction.getMeasurements().put(appStartKey, value);
```
marandaneto, line 23: app start metric is sent only once

sentry-android-core/src/main/java/io/sentry/android/core/PerformanceAndroidEventProcessor.java
```diff
@@ -0,0 +1,35 @@
...
+        AppStartState.getInstance().setSentStartUp();
+      }
+    }
+    final Map<String, @NotNull MeasurementValue> framesMetrics =
+        ActivityFramesState.getInstance().getMetrics();
+    if (framesMetrics != null) {
+      transaction.getMeasurements().putAll(framesMetrics);
```
marandaneto, line 30: slow/frozen frames are always sent, to every transaction, and represents the overall amount of the app lifecycle, and not from the start-end time of the transaction, we can't track only during the start/end of the transaction unless we implement the whole counting ourselves and this changes based on OS version and so on

sentry-android-core/src/main/java/io/sentry/android/core/SentryPerformanceProvider.java
```diff
@@ -0,0 +1,73 @@
...
+import android.database.Cursor;
+import android.net.Uri;
+import android.os.SystemClock;
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import io.sentry.DateUtils;
+import org.jetbrains.annotations.ApiStatus;
+
+@ApiStatus.Internal
+public final class SentryPerformanceProvider extends ContentProvider {
```
marandaneto, line 16: Performance as a generic name instead of AppStartProvider because we may reuse this in the future for more things like starting the app start transaction etc

marandaneto:
@philipphofmann and @bruno-garcia can you give some early feedback? I have left a few comments and TODO with my findings as well

sentry-android-core/build.gradle.kts
philipphofmann: The last release for the Android Support Library was in September 2018 with [28.0.0](https://developer.android.com/topic/libraries/support-library/revisions#28-0-0). It doesn't receive any bug fixes anymore. So you have to migrate to AndroidX. I think it is fine to pull in that dependency.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesState.java
```diff
@@ -0,0 +1,82 @@
...
+  void addActivity(final @NotNull Activity activity) {
+    frameMetricsAggregator.add(activity);
+  }
+
+  void removeActivity(final @NotNull Activity activity) {
+    frameMetricsAggregator.remove(activity);
+  }
+
+  @Nullable
+  Map<String, @NotNull MeasurementValue> getMetrics() {
```
philipphofmann, line 34: `l`: What about returning a data class with total, slow and frozen frames here instead? The `PerformanceAndroidEventProcessor` could then convert it to the correct format.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
philipphofmann: Did you try it in a release build?

sentry-android-core/src/main/java/io/sentry/android/core/PerformanceAndroidEventProcessor.java
philipphofmann: Whenever we start a transaction, we could check the current count of the frames in `ActivityFramesState`. When you finish a transaction, you again call `ActivityFramesState`, and then you can calculate the different types of frames during the time of the transaction.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -70,6 +85,15 @@ public void register(final @NotNull IHub hub, final @NotNull SentryOptions optio
     if (this.options.isEnableActivityLifecycleBreadcrumbs() || performanceEnabled) {
       application.registerActivityLifecycleCallbacks(this);
       this.options.getLogger().log(SentryLevel.DEBUG, "ActivityLifecycleIntegration installed.");
+
+      // this is called after the Activity is created, so we know if the App is a warm or cold
+      // start.
```
philipphofmann, line 90: `l`: Can you maybe elaborate a bit on why this is called after the Activity is created? Why don't we use `onActivityCreated` instead?

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -196,11 +227,29 @@ public synchronized void onActivityCreated(
 
   @Override
   public synchronized void onActivityStarted(final @NonNull Activity activity) {
+    if (performanceEnabled) {
```
philipphofmann, line 230: `l`: Why don't we do this in `onActivityCreated`?

sentry-android-core/src/main/java/io/sentry/android/core/AppStartState.java
```diff
@@ -0,0 +1,68 @@
...
+  @Nullable
+  Date getAppStartTime() {
+    return appStartTime;
+  }
+
+  synchronized void setAppStartTime() {
+    if (appStartSet) {
+      return;
+    }
+    final long millis = SystemClock.uptimeMillis();
```
philipphofmann, line 63: `m`: Why do we use `SystemClock.uptimeMillis()` and not the current date? Can we add the reason somewhere in a comment and also document this in the setter of appStartEnd.

philipphofmann (commented):
Good job for a draft PR. Although it's a draft PR, I added many comments. I guess you would anyway fix most of them before making this PR ready for review. Still, I added them to not forget about them.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
marandaneto: yes, maybe thats the case for some OS versions only, I could not find anything though

sentry-android-core/src/main/java/io/sentry/android/core/PerformanceAndroidEventProcessor.java
marandaneto: that would not scale well if I have multiple transactions at the same time, the same issue as counting on iOS the frames on iOS, we can do it next maybe?

sentry-android-core/src/main/AndroidManifest.xml
```diff
@@ -8,5 +8,11 @@
       android:name=".SentryInitProvider"
       android:authorities="${applicationId}.SentryInitProvider"
       android:exported="false"/>
+
+    <provider
+        android:name=".SentryPerformanceProvider"
+        android:authorities="${applicationId}.SentryPerformanceProvider"
+        android:initOrder="200"
```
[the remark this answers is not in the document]
marandaneto: the higher the more important, so it'd make sense to put Long.MAX_VALUE :D

sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesState.java
```diff
@@ -0,0 +1,82 @@
...
+        for (int i = 0; i < totalIndexArray.size(); i++) {
+          int frameTime = totalIndexArray.keyAt(i);
+          int numFrames = totalIndexArray.valueAt(i);
+          totalFrames += numFrames;
+          // hard coded values, its also in the official android docs and frame metrics API
+          if (frameTime > 700) {
+            // frozen frames, threshold is 700ms
+            frozenFrames += numFrames;
+          }
+          if (frameTime > 16) {
```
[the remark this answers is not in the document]
marandaneto: I don't think it changes on Android, it's written in their docs and source code in the same way

sentry-android-core/src/main/java/io/sentry/android/core/AppStartState.java
```diff
@@ -0,0 +1,68 @@
...
+final class AppStartState {
+
+  private static final @NotNull AppStartState instance = new AppStartState();
+
+  private @Nullable Long appStart;
+  private @Nullable Long appStartEnd;
+  private @Nullable Boolean coldStart;
+  private boolean sentStartMetric = false;
+  private @Nullable Date appStartTime;
+  private volatile boolean appStartSet = false;
```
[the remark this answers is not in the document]
marandaneto: yep this can be improved now, good catch, before `appStartTime` was nullable so such check would not work, will fix

sentry-android-core/src/main/java/io/sentry/android/core/AppStartState.java
```diff
@@ -0,0 +1,68 @@
...
+    this.coldStart = coldStart;
+  }
+
+  boolean isSentStartMetric() {
+    return sentStartMetric;
+  }
+
+  void setSentStartUp() {
+    this.sentStartMetric = true;
+  }
```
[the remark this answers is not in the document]
marandaneto: yep, good catch

sentry-android-core/src/main/java/io/sentry/android/core/PerformanceAndroidEventProcessor.java
```diff
@@ -0,0 +1,38 @@
...
+      Long appStartUpInterval = AppStartState.getInstance().getAppStartInterval();
+      if (appStartUpInterval != null) {
+        MeasurementValue value = new MeasurementValue((float) appStartUpInterval);
+
+        String appStartKey =
+            AppStartState.getInstance().getColdStart() ? "app_start_cold" : "app_start_warm";
+
+        transaction.getMeasurements().put(appStartKey, value);
+        AppStartState.getInstance().setSentStartUp();
+      }
```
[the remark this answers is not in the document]
marandaneto: mm thats a good point, it'll make contention but I don't think it'd be a big problem, maybe I move the `sentStartMetric` out from the `AppStartState` class to `PerformanceAndroidEventProcessor` so I can resolve this issue easily with a volatile field

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
 
+Feat: App start up and slow/frozen frames metrics (#1466)
```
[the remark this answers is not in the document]
marandaneto: both share `ActivityLifecycleIntegration` and `PerformanceAndroidEventProcessor` so it just sounded more complicated

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
```diff
@@ -51,10 +51,14 @@ public static void init(
    * @param logger your custom logger that implements ILogger
    * @param configuration Sentry.OptionsConfiguration configuration handler
    */
-  public static void init(
+  public static synchronized void init(
```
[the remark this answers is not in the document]
marandaneto: it should be always like this, Sentry#init is also `synchronized`, in case somebody inits the SDK twice in different threads

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
```diff
@@ -51,10 +51,14 @@ public static void init(
...
       @NotNull final Context context,
       @NotNull ILogger logger,
       @NotNull Sentry.OptionsConfiguration<SentryAndroidOptions> configuration) {
+    // if SentryPerformanceProvider was disabled or removed, we set the App Start when
+    // the SDK is called.
+    AppStartState.getInstance().setAppStartTime();
```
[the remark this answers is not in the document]
marandaneto: its still useful, if we dont know when it started, we assume when the SDK is inited, yes, it should be documented, but not setting at all is even less useful as you don't have metrics at all.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesState.java
marandaneto: I thought about it, but we just need to add these measurements to the current transactions.measurements, nothing more, its not used outside of this scope, it just sounded over-engineering, like, creating a new object and moving back to a Map anyway

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
marandaneto: I will add a better comment, but it's hard to explain and requires understanding deeply how the whole process start work, its a few pages that cant be easily resumed in one liner

sentry-android-core/src/main/java/io/sentry/android/core/AppStartState.java
marandaneto: to measure intervals, you have to use `SystemClock.uptimeMillis()`, the current time might be changed by the user or by the phone provider, timezone etc

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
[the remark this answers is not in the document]
marandaneto: for context, it was like this already, `event.contexts.app.appStartTime` was set when the SDK was inited. as well (on the creation of DefaultAndroidEventProcessor)

sentry-android-core/src/main/java/io/sentry/android/core/PerformanceAndroidEventProcessor.java
jennmueng: Running into a similar issue on RN as well. Considered @philipphofmann's suggestion at first but on JS users can change the start/end timestamps which makes it not work and way more difficult to implement metrics at the start/end of the transaction.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesState.java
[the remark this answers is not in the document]
philipphofmann: Interesting. How does it work then if you have a phone with 120 fps?

sentry-android-core/src/main/java/io/sentry/android/core/PerformanceAndroidEventProcessor.java
philipphofmann: Why do we run into scaling problems? Hmm, changing it afterward might cause problems. Having it different on platforms is also not a good idea. The advantage of always sending the total number of frames during the lifetime of an application gives you more insights on how many slow frames and frozen frames you have, but the downside is that you have no clue where they are coming from. Being able to correlate them with transactions gives you way more insight. Maybe it would even make sense to send both lifetime frames and frames during the transaction.

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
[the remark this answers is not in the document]
philipphofmann: It would be nice to open up another PR for this. We can then quickly merge it and don't have to wait for this PR to fix the master.

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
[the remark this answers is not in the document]
philipphofmann: Okay, then good docs are the key here to avoid confusion.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
philipphofmann: Can we maybe link to resources so you can find out on your own even when you have to read some articles?

sentry/src/main/java/io/sentry/protocol/MeasurementValue.java
```diff
@@ -0,0 +1,13 @@
+package io.sentry.protocol;
+
+import org.jetbrains.annotations.ApiStatus;
+
+@ApiStatus.Internal
+public final class MeasurementValue {
```
bruno-garcia, line 6: I wonder if we're better off without this class and using the `float` directly to avoid allocating these objects

bruno-garcia:
Is this blocked? Is there anything I can do to help move this PR (particularly the App Startup bits)

marandaneto:
> Is this blocked? Is there anything I can do to help move this PR (particularly the App Startup bits)

nop, this PR was the PoC, the PR for App startup is https://github.com/getsentry/sentry-java/pull/1487
Next week I'll be focused only on that to make the discussed changes and land it.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesState.java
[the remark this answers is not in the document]
marandaneto: afaik Android displays are limited to 60 fps

sentry/src/main/java/io/sentry/protocol/MeasurementValue.java
marandaneto: could be easily removed, its not a public API, but the idea was to be extendable like the API is, its an object with a single float property

sentry-android-core/src/main/AndroidManifest.xml
```diff
@@ -8,5 +8,11 @@
       android:name=".SentryInitProvider"
       android:authorities="${applicationId}.SentryInitProvider"
       android:exported="false"/>
+
+    <provider
+        android:name=".SentryPerformanceProvider"
+        android:authorities="${applicationId}.SentryInitProvider"
+        android:initOrder="200"
```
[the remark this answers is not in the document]
marandaneto: see https://github.com/getsentry/sentry-java/pull/1466/files#r635893582

PR #1468, opened 2021-05-10
Fix handling immutable collections on SentryEvent and protocol objects

### LINKED ISSUE #1357: Applying immutable map in `SentryEvent#setExtras` causes `UnsupportedOperationException`
_Platform:_
- [x]  Android -> If yes, which Device API (and compileSdkVersion/targetSdkVersion/Build tools) version?
Device `API`: 26
`compileSdkVersion`: 30
`targetSdkVersion`: 30
`buildTools`: applied by Android Gradle Plugin

_IDE:_
- [x]  Android Studio -> If yes, which version?
`4.2 beta 3` 

_Build system:_
- [x]  Gradle -> If yes, which version?
`6.7.1`

_Android Gradle Plugin:_
- [x]  Yes -> If yes, which version?
`4.2.0-beta03`

- [x]  No

_Proguard/R8:_
- [x]  Disabled

_Platform installed with:_
- [x] Maven Central

The version of the SDK:
**4.3.0**

---
I have the following issue:

When applying extra data by `Sentry#setExtra` and then applying extra data for a single event by `SentryEvent#setExtras` with `immutable map` (from `Kotlin`), I'm experiencing `UnsupportedOperationException` as SDK tries to edit the immutable map that is passed.

It crashes on the runtime as Java's method signatures will accept `immutable` Kotlin's map.

**Steps to reproduce:**
I've prepared [a reproduction repository](https://github.com/wzieba/SentryImmutableMapCrashRepro).

To check it, please go to [`SentryProxyTest`](https://github.com/wzieba/SentryImmutableMapCrashRepro/blob/main/app/src/test/java/io/github/sentrymutablemaprepro/SentryProxyTest.kt). Both of the use cases are tested there.

You can also check the results of those tests on CI.
[The test with mutable map passes](https://github.com/wzieba/SentryImmutableMapCrashRepro/runs/2208270028?check_suite_focus=true#step:5:132) while test with immutable map [throws an exception](https://github.com/wzieba/SentryImmutableMapCrashRepro/runs/2208270028?check_suite_focus=true#step:5:199)

**Actual result:**
When applying extra with `MutableMap` everything works fine, the exception is not thrown. When applying extra with `Map` (which is immutable), `UnsupportedOperationException` is thrown.

**Expected result:**
Sentry SDK should accept both the `mutable` and `immutable` types. For `immutable` types, if editing is required, it should map a `Map` to `mutable` type.

---

maciejwalkowiak:
## :scroll: Description

Fix handling immutable collections on SentryEvent and protocol objects. With this change, when immutable collection is set on any of `SentryEvent` field, it is still possible to add more entries to this collection on the event.

Fixes #1357

Unit tests.

## :pencil: Checklist

- [x] No breaking changes


sentry/src/main/java/io/sentry/SentryBaseEvent.java
```diff
@@ -185,8 +185,15 @@ public void setThrowable(final @Nullable Throwable throwable) {
     return tags;
   }
 
-  public void setTags(Map<String, String> tags) {
-    this.tags = tags;
+  public void addTags(Map<String, String> tags) {
```
maciejwalkowiak, line 188: I believe its better to explicit here that we are not actually "setting" tags but adding tags from one collection to existing tags. 

This also means that this change is breaking on the compilation level. I am not sure if this is what we want to do or we want to keep it backward compatible @marandaneto @bruno-garcia

marandaneto: that does against our unified API, I do agree its a convenience but the public API already exposes `setField(x, y)` and `setFields(fields)` so I'd stick with the unified API.
One of them only tests the value to a specific key and the other replace everything.

For Kotlin, when we work on the Kotlin extensions, we could have extension methods so the syntax would be Kotlin friendlier eg `keys[x] = y`

marandaneto: indeed, I totally agree with the Java convention point of view, actually, I thought those methods were in the Scope but actually, they are in the SentryBaseEvent so these additional methods actually make sense to me, unless @bruno-garcia is against them.

maciejwalkowiak: Decided to drop `addXXX` method and in `setFields(fields)` follow Java convention to replace the collection just that instead of assigning the value given by parameter, new mutable collection is created containing all elements from the parameter.

marandaneto:
worth doing the same with `Scope#setFingerprint`, `SentryValues`, `Device#archs`, `Mehanism#meta` and `data`, `Message#params`, `Request#headers`, I mean, every List/Array/Map that are public and mutable and are exposed i nthe public API to swap the reference with an immutable type

sentry/src/main/java/io/sentry/SentryBaseEvent.java
```diff
@@ -186,7 +186,7 @@ public void setThrowable(final @Nullable Throwable throwable) {
   }
 
   public void setTags(@Nullable Map<String, String> tags) {
-    this.tags = tags;
+    this.tags = tags != null ? new HashMap<>(tags) : null;
```
marandaneto, line 189: maybe worth adding a method to `CollectionUtils` that does all of this

sentry/src/main/java/io/sentry/SentryBaseEvent.java
```diff
@@ -262,7 +262,7 @@ public void setUser(final @Nullable User user) {
   }
 
   public void setBreadcrumbs(final @Nullable List<Breadcrumb> breadcrumbs) {
-    this.breadcrumbs = breadcrumbs;
+    this.breadcrumbs = breadcrumbs != null ? new ArrayList<>(breadcrumbs) : null;
```
marandaneto, line 265: same as above for ArrayList

sentry/src/main/java/io/sentry/protocol/Request.java
```diff
@@ -144,23 +145,23 @@ public void setCookies(final @Nullable String cookies) {
   }
 
   public void setHeaders(final @Nullable Map<String, String> headers) {
-    this.headers = headers;
+    this.headers = headers != null ? new ConcurrentHashMap<>(headers) : null;
```
marandaneto, line 148: we could reuse `CollectionUtils#shallowCopy`

sentry/src/main/java/io/sentry/SentryBaseEvent.java
marandaneto: nice, thanks

sentry/src/main/java/io/sentry/SentryBaseEvent.java
maciejwalkowiak: `CollectionUtils.newHashMap(@Nullable Map<K,V> map)` and rename `shallowCopy` to `CollectionUtils.newConcurrentHashMap(@Nullable Map<K,V> map)`?

marandaneto: sounds good, maybe we add in the javadocs that this is pretty much a shallow copy as well, as its being used in a lot of cloning methods

maciejwalkowiak:
`Device#archs` is an array, so i think there is nothing to be done there. The rest is done.

marandaneto (approved):
if CI is happy, I am happy :)
Thanks @maciejwalkowiak

PR #1472, opened 2021-05-12
Clean up sample and add with Scope sample

marandaneto:
## :scroll: Description
_#skip-changelog_

## :pencil: Checklist

- [ ] No breaking changes

### DIFF
--- a/sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/MainActivity.java
+++ b/sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/MainActivity.java
@@ -71,8 +71,6 @@ public class MainActivity extends AppCompatActivity {
                 String json = "{ \"number\": 10 }";
                 Attachment attachment = new Attachment(json.getBytes(), "log.json");
                 scope.addAttachment(attachment);
-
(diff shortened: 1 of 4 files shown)


sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/MainActivity.java
```diff
@@ -92,41 +71,42 @@ protected void onCreate(Bundle savedInstanceState) {
...
+          // Use the client's IP address or enable 'sendDefaultPii' that does it automatically.
           user.setIpAddress("{{auto}}");
           Sentry.setUser(user);
+          // pass null to Sentry#setUser to unset the user
+          Sentry.captureMessage("Some message with user.");
+        });
+
+    binding.localScope.setOnClickListener(
+        view -> {
+          Sentry.withScope(
```
bruno-garcia, line 102: Since Hub is global, access to it from another thread would race to the data changed in this scope so I would rather avoid adding this example. We should consider adding the overloads to capture like we have on Cocoa and Flutter and bring up to the SDK evolution that we want to drop this methods.

PR #1478, opened 2021-05-17
Feat: OkHttp callback for Customising the Span

### LINKED ISSUE #1365: OkHttp callback for Customising the Span
Lets add a callback to the OkHttp integration that allows the user to customize the `Span`, there's no way to do it right now.

https://github.com/getsentry/sentry-java/blob/main/sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpInterceptor.kt#L11

Related to https://github.com/getsentry/sentry-java/issues/1362

maciejwalkowiak:
## :scroll: Description

OkHttp callback for Customising the Span

Fixes #1365

Unit tests.

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpInterceptor.kt
+++ b/sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpInterceptor.kt
@@ -3,13 +3,15 @@ package io.sentry.android.okhttp
 import io.sentry.Breadcrumb
 import io.sentry.HubAdapter
 import io.sentry.IHub
+import io.sentry.ISpan
 import io.sentry.SpanStatus
 import java.io.IOException
 import okhttp3.Interceptor
 import okhttp3.Response
 
 class SentryOkHttpInterceptor(
-    private val hub: IHub = HubAdapter.getInstance()
+    private val hub: IHub = HubAdapter.getInstance(),
+    private val spanCustomizer: (span: ISpan) -> Unit = { }
 ) : Interceptor {
 
     override fun intercept(chain: Interceptor.Chain): Response {
@@ -39,6 +41,7 @@ class SentryOkHttpInterceptor(
             }
             throw e
         } finally {
+            span?.apply(spanCustomizer)
             span?.finish()
             val breadcrumb = Breadcrumb.http(request.url.toString(), request.method, code)
             request.body?.contentLength().ifHasValidLength {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 * Fix: Enrich Transactions with Context Data (#1469)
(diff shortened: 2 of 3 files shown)


sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpInterceptor.kt
```diff
@@ -3,13 +3,15 @@ package io.sentry.android.okhttp
...
+import io.sentry.ISpan
 import io.sentry.SpanStatus
 import java.io.IOException
 import okhttp3.Interceptor
 import okhttp3.Response
 
 class SentryOkHttpInterceptor(
-    private val hub: IHub = HubAdapter.getInstance()
+    private val hub: IHub = HubAdapter.getInstance(),
+    private val spanCustomizer: (span: ISpan) -> Unit = { }
```
marandaneto, line 14: what's about using a name that sort of matches our beforeSend and beforeBreadcrumb callbacks?
eg `beforeSpan`? cc @bruno-garcia 

I believe this could also be an interface already available in the sentry package, it may be reused in other integrations, I'd not add to options though, wdyt?

marandaneto: 2nd thing is that using `Unit` isn't the best definition if using it on Java, but as OkHttp is already written in Kotlin, I guess it's just fine.

maciejwalkowiak: Agreed with naming. I think `Unit` can stay.

marandaneto:
as @bruno-garcia came up with this callback after talking to a few people, let's get his review before merging it, looks good to me.

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpInterceptor.kt
```diff
@@ -39,6 +41,7 @@ class SentryOkHttpInterceptor(
             }
             throw e
         } finally {
+            span?.apply(beforeSpan)
```
bruno-garcia, line 44: Is there a way we can drop the span? Similar to how we do `beforeSend` and `beforeBreadcrumb`?

We would also need to pass some context to the callback, like the request (and resposne?) objects.

Some context: We're trying to add a way for a user to do:

```
if (request.url == "/health" || response.status == 200) {
return null;
```

maciejwalkowiak: Good point about dropping the span. url and status can be obtained from span but in less convenient way than with plain request and response. I'll change the implementation.

marandaneto (approved):

PR #1479, opened 2021-05-18
Add coroutines support

### LINKED ISSUE #973: Kotlin Coroutine Support
When we switch threads via coroutine, all information in hub (e.g. extra and breadcrumbs) are lost because it's thread local.

```kotlin
class MyClass {
    suspend fun example() {
        // Record a breadcrumb in the scope.
        Sentry.addBreadcrumb("On First Thread")
        Sentry.captureMessage("This message captured with breadcrumb 'On First Thread'")
        
        withContext(Dispatchers.IO) {
            Sentry.captureMessage("This message captured without breadcrumb")
            Sentry.addBreadcrumb("On Second Thread")
            Sentry.captureMessage("This message captured with breadcrumb 'On Second Thread'")
        }
    }
}
```

I think sentry-java should provide something to keep thread local information when switching thread or the implementation of IHub that delegates everything to CorotuineContext.

maciejwalkowiak:
## :scroll: Description

Adds the ability to use Sentry scope with Kotlin coroutines through a coroutines context element `SentryContext`.

Fixes #973.

Unit tests.

## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next steps

It would be nice if someone with coroutines experience review it.


sentry/src/test/java/io/sentry/SentryContext.kt
```diff
@@ -0,0 +1,23 @@
+package io.sentry
+
+import kotlinx.coroutines.ThreadContextElement
+import kotlin.coroutines.AbstractCoroutineContextElement
+import kotlin.coroutines.CoroutineContext
+
+class SentryContext : ThreadContextElement<IHub>, AbstractCoroutineContextElement(Key) {
```
maciejwalkowiak, line 7: This is done in a similar way to how slf4j support is implemented in https://github.com/Kotlin/kotlinx.coroutines/blob/master/integration/kotlinx-coroutines-slf4j/src/MDCContext.kt

sentry/src/test/java/io/sentry/SentryContextTest.kt
```diff
@@ -0,0 +1,134 @@
...
+    fun testContextCanBePassedBetweenCoroutines() = runBlocking {
+        Sentry.setTag("myKey", "myValue")
+        // Scoped launch with MDCContext element
+        launch(SentryContext()) {
+            assertEquals("myValue", getTag("myKey"))
+        }.join()
+    }
+
+    @Test
+    fun testContextInheritanceMDC() = runBlocking {
```
maciejwalkowiak, line 57: This test is copied from sfl4j implementation just for temporary comparison purposes. It passes.

sentry/src/test/java/io/sentry/SentryContextTest.kt
```diff
@@ -0,0 +1,134 @@
...
+            // Scoped launch with inherited MDContext element
+            launch(Dispatchers.Default) {
+                assertEquals("myValue", MDC.get("myKey"))
+            }.join()
+        }
+        assertEquals("myValue", MDC.get("myKey"))
+    }
+
+    @Test
+    fun testContextInheritance() = runBlocking {
```
maciejwalkowiak, line 70: This test is similar to `testContextInheritanceMDC` but uses `SentryContext`. It fails. @marandaneto perhaps you have an idea why?

marandaneto: no idea just by looking at it, just wondering if we should actually clone the Hub or the Scope.
`Hub#clone` has never been used internally, we'd need to guarantee that its getting the right `stack`, something worth looking at

sentry/src/main/java/io/sentry/Sentry.java
```diff
@@ -33,7 +33,8 @@ private Sentry() {}
    *
    * @return the hub
    */
-  static @NotNull IHub getCurrentHub() {
+  @ApiStatus.Internal // exposed for the coroutines integration in SentryContext
+  public static @NotNull IHub getCurrentHub() {
```
marandaneto, line 37: what's about exposing it in the `IHub` and using `HubAdapter` instead of `Sentry` static class? same for `setCurrentHub`

sentry/src/main/java/io/sentry/Scope.java
```diff
@@ -337,9 +338,10 @@ public void clear() {
    *
    * @return the tags map
    */
-  @NotNull
-  Map<String, String> getTags() {
-    return tags;
+  @ApiStatus.Internal
+  @SuppressWarnings("NullAway") // tags are never null
+  public @NotNull Map<String, String> getTags() {
```
marandaneto, line 343:
```suggestion
  public @NotNull Map<@NotNull String, @NotNull String> getTags() {
```

does it work without `@SuppressWarnings("NullAway")`?

sentry/src/main/java/io/sentry/Sentry.java
marandaneto: Nvm, it should be under the sentry static class anyway

maciejwalkowiak: It can be added to `HubAdapter` and call package-private methods on `Sentry` class, but it cannot be added to `IHub` as there is no concept of the "current hub" on the hub itself. 

If the goal would be to hide the "current hub" concept from the public `Sentry` class API, we can add it to `HubAdapter` - am I getting it right?

sentry/src/main/java/io/sentry/Scope.java
maciejwalkowiak: `Collections.newConcurrentHashMap` returns a nullable object, this is why this one is marked with `SuppressWarnings`, as the `tags` are never null.

marandaneto: makes sense, thanks

sentry/src/main/java/io/sentry/Sentry.java
marandaneto: yes, the goal would be to not expose as a public method, even with `@ApiStatus.Internal`, if it can be package-private it'd be better IMO.
wdyt @bruno-garcia ?

sentry-kotlin-extensions/build.gradle.kts
```diff
@@ -0,0 +1,66 @@
...
+    targetCompatibility = JavaVersion.VERSION_1_8
+}
+
+tasks.withType<KotlinCompile>().configureEach {
+    kotlinOptions.jvmTarget = JavaVersion.VERSION_1_8.toString()
+}
+
+dependencies {
+    api(project(":sentry"))
+    implementation(Config.Libs.coroutinesCore)
```
marandaneto, line 24: could this be `compileOnly`? ideally `sentry-kotlin-extensions` would offer method extensions as well in the future, making it easier/friendlier working with Kotlin and the Unified API, but right now it depends on coroutines and people might not like it, another option would be to rename it, to be something more coroutines specific, @bruno-garcia opinions?

buildSrc/src/main/java/Config.kt
```diff
@@ -80,6 +80,8 @@ object Config {
         private val retrofit2Group = "com.squareup.retrofit2"
         val retrofit2 = "$retrofit2Group:retrofit:$retrofit2Version"
         val retrofit2Gson = "$retrofit2Group:converter-gson:$retrofit2Version"
+
+        val coroutinesCore = "org.jetbrains.kotlinx:kotlinx-coroutines-core:1.5.0"
```
marandaneto, line 84:
```suggestion
        val coroutinesCore = "org.jetbrains.kotlinx:kotlinx-coroutines-core:1.4.3"
```

1.5.0 depends on Kotlin 1.5.x, 1.4.3 depends on Kotlin 1.4.x, our projects right now are tested against Kotlin 1.4.30, so I'd stick to 1.4.3, wdyt?

sentry/src/main/java/io/sentry/Sentry.java
marandaneto: `currentHub` lives in the Sentry static class, moving to `HubAdapter` would be strange though, maybe we just need to be sure that its well tested and resetting the static values correctly in tests

maciejwalkowiak: `currentHub` can stay in Sentry class, we just make these methods package private in Sentry and expose them with `HubAdapter` - since users don't use HubAdapter public API won't be polluted

sentry-kotlin-extensions/build.gradle.kts
bruno-garcia: Are there valid use cases where folks can't have coroutines in their project? Other than perhaps small edge cases?
Personally I rather we have a single package with all kotlin related bits like kotlin extensions and the coroutine support and make it a "all or nothing" thing. Breaking into 2 packages will add to the complexity in terms of maintenance for us, publshing more packages to central, documenting extra packages.

Maybe I'll rephrase the question: If I"m creating a Kotlin project today, is there a reason I want it not to have?

sentry/src/main/java/io/sentry/Sentry.java
bruno-garcia: Makes sense to have this here wit @Internal and to be honest it's probably not going to be internal forever since this use case is real and users might want to propagate the Sentry state by themselves. This is done in Python, Rust and other SDKs where the user is expected to set the hub at the right places where applicable.

bruno-garcia (approved):
From my perspective LGTM!

We could get feedback from folks that have been interested in this before shipping though.
And validate that question whether it's OK to have 1 package with ktx + the coroutines, or if it's going to be a problem for too many ppl

sentry-kotlin-extensions/build.gradle.kts
marandaneto: the problem is on Android, it adds 1.9 MB as a dependency, and coroutines got traction also on Android apps but still not adopted by everyone, and dependency's size is one of the problems.
I see though Android apps wanting to use Sentry extensions, we can live with that, but it's a trade-off that I likely would not have as an Android dev

sentry/src/main/java/io/sentry/Sentry.java
marandaneto: sounds good, thanks

sentry-kotlin-extensions/build.gradle.kts
```diff
@@ -0,0 +1,66 @@
+import net.ltgt.gradle.errorprone.errorprone
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
+plugins {
+    `java-library`
+    kotlin("jvm")
+    jacoco
+    id(Config.QualityPlugins.errorProne)
+    id(Config.QualityPlugins.gradleVersions)
+    id(Config.QualityPlugins.detektPlugin)
```
marandaneto, line 10: you need to apply detekt plugin

https://github.com/getsentry/sentry-java/blob/main/sentry-android-timber/build.gradle.kts#L93-L96

sentry-kotlin-extensions/build.gradle.kts
```diff
@@ -0,0 +1,66 @@
...
+    toolVersion = Config.QualityPlugins.Jacoco.version
+}
+
+tasks.jacocoTestReport {
+    reports {
+        xml.isEnabled = true
+        html.isEnabled = false
+    }
+}
+
```
marandaneto, line 55: enable the explicitApi feature https://github.com/getsentry/sentry-java/blob/main/sentry-android-timber/build.gradle.kts#L65-L67

sentry-kotlin-extensions/build.gradle.kts
bruno-garcia: Would it be fair to say that given extension method (ktx) is not really a requirement, it's bundled in the `kotlin-extensions` package which includes coroutines. For users who have not adopted coroutines, they are not blocked, they continue to have the support we have today, that is: no extension methods?
Not necessarily the best for those users but best balance between maintenance burden to us and future thinking (Assuming that new projects with kotlin will use coroutines)?

marandaneto: lets publish as it is, once we're done with method extensions for Sentry, we evaluate this again, most Android people are starting with coroutines or willing to migrate it whenever possible (eg new features), so I agree that a balance here plays a bigger role, thanks

marandaneto:
@maciejwalkowiak I've created the docs issue as well https://github.com/getsentry/sentry-docs/issues/3628

marandaneto (approved):
nice, thanks @maciejwalkowiak 
LGTM

iamkdblue:
I am trying to add support for sentry coroutine in the Spring Boot project given
[Here](https://docs.sentry.io/platforms/java/enriching-events/scopes/#installation)

i am using  implementation `"io.sentry:sentry-kotlin-extensions:5.0.0"`


but its throwing error 
```
Execution failed for task ':controller:compileKotlin'.
> Could not resolve all files for configuration ':controller:compileClasspath'.
   > Could not find io.sentry:sentry-kotlin-extensions:5.0.0.
     Searched in the following locations:
       - https://repo.maven.apache.org/maven2/io/sentry/sentry-kotlin-extensions/5.0.0/sentry-kotlin-extensions-5.0.0.pom
       - https://jcenter.bintray.com/io/sentry/sentry-kotlin-extensions/5.0.0/sentry-kotlin-extensions-5.0.0.pom
     Required by:
         project :controller

Possible solution:
 - Declare repository providing the artifact, see the documentation at https://docs.
```gradle.org/current/userguide/declaring_repositories.html

marandaneto:
@iamkdblue 

> Support for Kotlin Coroutines is available as a Beta. Features available in Beta are still in-progress and may have bugs. We recognize the irony.

please use this version for now https://github.com/getsentry/sentry-java/releases/tag/5.0.0-beta.7, 5.0.0 isn't out yet

iamkdblue:
@marandaneto 

cant find `SentryContext() ` and launch in spring boot 
and i am using withContext(Dispatchers.IO) not launch, mostly launch(fire and forget) uses in android!

marandaneto:
@iamkdblue please raise a new issue with your setup and what you're trying to do, it'd be easier to track it down and fix it if there are issues, thanks

markallanson:
Tested this out in 5.0.1 but it doesn't seem to work. Using this with the grpc-kotlin support and providing a `SentryContrext` to the CoroutineScope it creates, but when I am inside coroutines the Hub seems to resolve to the `NoOpHub` hence sentry errors are not reported anywhere.

Using `sentry-spring-boot-starter` to initialize Sentry, and outside the coroutines the Hub is resolved correctly.

marandaneto:
@markallanson please raise a new issue with more detailed context, happy to look into it, thanks

markallanson:
I got to the bottom of this, the sentry-spring-boot-starter wasn't initialising sentry before the `SentryContext` was created causing it to use the `NoOpHub`.

marandaneto:
@markallanson so was it a bug in your end or what, does it mean you're creating the `SentryContext` before even initing Sentry?

markallanson:
@marandaneto I need to try and get the `sentry-spring-boot-starter` to initialize sentry before my services try and use the context.

PR #1480, opened 2021-05-18
Expose SentryOptions

danieltdt:
## :scroll: Description
Almost all methods of IHub are exposed on Sentry class, except by `getOptions`. This PR exposes that method.

If you have set `enableExternalConfiguration` when initializing with `Sentry.init()`, there is no public way to access `inAppInclude` list... Exposing the SentryOptions for the current context, all options resolved from system/environment/properties may be accessible by `Sentry.getOptions`. 

By checking if a custom option is accessible when calling the newly introduced method.

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/sentry/api/sentry.api
+++ b/sentry/api/sentry.api
@@ -559,6 +559,7 @@ public final class io/sentry/Sentry {
 	public static fun endSession ()V
 	public static fun flush (J)V
 	public static fun getLastEventId ()Lio/sentry/protocol/SentryId;
+	public static fun getOptions ()Lio/sentry/SentryOptions;
 	public static fun getSpan ()Lio/sentry/ISpan;
 	public static fun init ()V
 	public static fun init (Lio/sentry/OptionsContainer;Lio/sentry/Sentry$OptionsConfiguration;)V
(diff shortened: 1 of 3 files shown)


marandaneto:
@danieltdt thanks, but not sure if this is what we'd like to do, `Hub#getOptions` is actually an internal method, we probably have forgotten to add `@ApiStatus.Internal`.
You can access the options if needed by having your own `Integration` or initing the SDK manually and accessing the options via the `OptionsConfiguration` callback.
The idea is that once the Options is configured, it's immutable, why would you need to access it?

bruno-garcia:
Is there another way to get the list of inAppInclude? Curious, what do you need this list for?

The SentryOptions is also known as "the client options" even though it drives a lot of the SDK's behavior during initialization it ultimately belongs to the client. We prefer not to have a getter for this object through the public Sentry API which is the type most interfaced by users. It would ultimately mean Sentry.getCurrentHub().getClient().getOptions() and considering we have `bindClient` depending on which client yuo have bound to the scope it would return a different instance.

danieltdt:
I wanted to have access to external configurations before starting the client (so I could add more options and create my custom client).
But I don't use Sentry anymore, I'll close this PR :)

PR #1487, opened 2021-05-25
Feat: Measure app start time

marandaneto:
## :scroll: Description
Feat: App start metric

extract from https://github.com/getsentry/sentry-java/pull/1466

## :pencil: Checklist

- [X] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 * Feat: OkHttp callback for Customising the Span (#1478)
 * Feat: Add breadcrumb in Spring RestTemplate integration (#1481)
 * Fix: Cloning Stack (#1483)
-* Feat: App start up metrics (#)
+* Feat: App start up metrics (#1487)
 
 # 5.0.0-beta.4
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 * Feat: OkHttp callback for Customising the Span (#1478)
 * Feat: Add breadcrumb in Spring RestTemplate integration (#1481)
 * Fix: Cloning Stack (#1483)
+* Feat: App start up metrics (#)
 
 # 5.0.0-beta.4
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-* Feat: App start up metrics (#1487)
+* Feat: Measure app start time (#1487)
 
 ## 5.0.1
 
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
@@ -29,7 +29,8 @@ public final class ActivityLifecycleIntegration
     implements Integration, Closeable, Application.ActivityLifecycleCallbacks {
 
   private static final String UI_LOAD_OP = "ui.load";
-  static final String APP_START_OP = "app.start";
+  static final String APP_START_WARM = "app.start.warm";
+  static final String APP_START_COLD = "app.start.cold";
 
   private final @NotNull Application application;
   private @Nullable IHub hub;
@@ -139,7 +140,7 @@ public final class ActivityLifecycleIntegration
         transaction = hub.startTransaction(activityName, UI_LOAD_OP, appStartTime);
         // start specific span for app start
 
-        appStartSpan = transaction.startChild(APP_START_OP, getColdStartDesc(), appStartTime);
+        appStartSpan = transaction.startChild(getAppStartOp(), getAppStartDesc(), appStartTime);
       }
 
       // lets bind to the scope so other integrations can pick it up
@@ -306,11 +307,19 @@ public final class ActivityLifecycleIntegration
     }
   }
 
-  private @NotNull String getColdStartDesc() {
+  private @NotNull String getAppStartDesc() {
     if (AppStartState.getInstance().isColdStart()) {
       return "Cold Start";
     } else {
       return "Warm Start";
+    }
+  }
+
+  private @NotNull String getAppStartOp() {
+    if (AppStartState.getInstance().isColdStart()) {
+      return APP_START_COLD;
+    } else {
+      return APP_START_WARM;
     }
   }
 }
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
@@ -28,6 +28,8 @@ import org.jetbrains.annotations.VisibleForTesting;
 public final class ActivityLifecycleIntegration
     implements Integration, Closeable, Application.ActivityLifecycleCallbacks {
 
+  private static final String NAV_OP = "navigation";
+
   private final @NotNull Application application;
   private @Nullable IHub hub;
   private @Nullable SentryAndroidOptions options;
@@ -125,12 +127,17 @@ public final class ActivityLifecycleIntegration
       // we can only bind to the scope if there's no running transaction
       ITransaction transaction;
       final String activityName = getActivityName(activity);
-      final String op = "navigation";
+
       final Date appStartTime = AppStartState.getInstance().getAppStartTime();
+
+      // in case appStartTime isn't available, we don't create a span for it.
       if (firstActivityCreated || appStartTime == null) {
-        transaction = hub.startTransaction(activityName, op);
+        transaction = hub.startTransaction(activityName, NAV_OP);
       } else {
-        transaction = hub.startTransaction(activityName, op, appStartTime);
+        // start transaction with app start timestamp
+        transaction = hub.startTransaction(activityName, NAV_OP, appStartTime);
+        // start specific span for app start
+        // TODO: add description cold/warm
         appStartSpan = transaction.startChild("app.start", appStartTime);
       }
 
@@ -226,8 +233,10 @@ public final class ActivityLifecycleIntegration
     if (!firstActivityResumed && performanceEnabled) {
       // sets App start as finished when the very first activity calls onResume
       AppStartState.getInstance().setAppStartEnd();
+
+      // finishes app start span
       if (appStartSpan != null) {
-          appStartSpan.finish();
+        appStartSpan.finish();
       }
       firstActivityResumed = true;
     }
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
@@ -28,7 +28,7 @@ import org.jetbrains.annotations.VisibleForTesting;
 public final class ActivityLifecycleIntegration
     implements Integration, Closeable, Application.ActivityLifecycleCallbacks {
 
-  private static final String NAV_OP = "navigation";
+  private static final String UI_LOAD_OP = "ui.load";
   static final String APP_START_OP = "app.start";
 
   private final @NotNull Application application;
@@ -133,10 +133,10 @@ public final class ActivityLifecycleIntegration
 
       // in case appStartTime isn't available, we don't create a span for it.
       if (firstActivityCreated || appStartTime == null) {
-        transaction = hub.startTransaction(activityName, NAV_OP);
+        transaction = hub.startTransaction(activityName, UI_LOAD_OP);
       } else {
         // start transaction with app start timestamp
-        transaction = hub.startTransaction(activityName, NAV_OP, appStartTime);
+        transaction = hub.startTransaction(activityName, UI_LOAD_OP, appStartTime);
         // start specific span for app start
         // TODO: add description cold/warm or different operation? how do we break down
         // per app start cold/warm?
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
@@ -137,7 +137,8 @@ public final class ActivityLifecycleIntegration
         // start transaction with app start timestamp
         transaction = hub.startTransaction(activityName, NAV_OP, appStartTime);
         // start specific span for app start
-        // TODO: add description cold/warm
+        // TODO: add description cold/warm or different operation? how do we break down
+        // per app start cold/warm?
         appStartSpan = transaction.startChild("app.start", appStartTime);
       }
 
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
@@ -29,6 +29,7 @@ public final class ActivityLifecycleIntegration
     implements Integration, Closeable, Application.ActivityLifecycleCallbacks {
 
   private static final String NAV_OP = "navigation";
+  static final String APP_START_OP = "app.start";
 
   private final @NotNull Application application;
   private @Nullable IHub hub;
@@ -139,7 +140,7 @@ public final class ActivityLifecycleIntegration
         // start specific span for app start
         // TODO: add description cold/warm or different operation? how do we break down
         // per app start cold/warm?
-        appStartSpan = transaction.startChild("app.start", appStartTime);
+        appStartSpan = transaction.startChild(APP_START_OP, appStartTime);
       }
 
       // lets bind to the scope so other integrations can pick it up
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
@@ -8,6 +8,7 @@ import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import io.sentry.Breadcrumb;
 import io.sentry.IHub;
+import io.sentry.ISpan;
 import io.sentry.ITransaction;
 import io.sentry.Integration;
 import io.sentry.Scope;
@@ -17,6 +18,7 @@ import io.sentry.SpanStatus;
 import io.sentry.util.Objects;
 import java.io.Closeable;
 import java.io.IOException;
+import java.util.Date;
 import java.util.Map;
 import java.util.WeakHashMap;
 import org.jetbrains.annotations.NotNull;
@@ -36,6 +38,8 @@ public final class ActivityLifecycleIntegration
 
   private boolean firstActivityCreated = false;
   private boolean firstActivityResumed = false;
+
+  private @Nullable ISpan appStartSpan;
 
   // WeakHashMap isn't thread safe but ActivityLifecycleCallbacks is only called from the
   // main-thread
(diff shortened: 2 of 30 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 * Feat: OkHttp callback for Customising the Span (#1478)
 * Feat: Add breadcrumb in Spring RestTemplate integration (#1481)
 * Fix: Cloning Stack (#1483)
+* Feat: App start up metrics (#)
```
marandaneto, line 6:
```suggestion
* Feat: App start up metrics (#1487)
```

sentry-android-core/src/main/java/io/sentry/android/core/AppStartState.java
```diff
@@ -0,0 +1,70 @@
...
+/** AppStartState holds the state of the App Start metric and appStartTime */
+final class AppStartState {
+
+  private static @NotNull AppStartState instance = new AppStartState();
+
+  /** appStart in milliseconds */
+  private @Nullable Long appStart;
+
+  /** appStartEnd in milliseconds */
+  private @Nullable Long appStartEnd;
```
philipphofmann, line 18: `l`: When adding `millis` to the variable name we can ditch the comment.
```suggestion
  private @Nullable Long appStartMillis;
  private @Nullable Long appStartEndMillis;
```

sentry-android-core/src/main/java/io/sentry/android/core/SentryPerformanceProvider.java
```diff
@@ -0,0 +1,86 @@
...
+ * as ContentProvider is the only reliable hook for libraries that works across all the supported
+ * SDK versions. When minSDK is >= 24, we could use Process.getStartUptimeMillis()
+ */
+@ApiStatus.Internal
+public final class SentryPerformanceProvider extends ContentProvider {
+
+  // static to rely on Class load
+  private static final @NotNull Date appStartTime = DateUtils.getCurrentDateTime();
+  // SystemClock.uptimeMillis() isn't affected by phone provider or clock changes.
+  private static final long appStart = SystemClock.uptimeMillis();
```
philipphofmann, line 28: `l`
```suggestion
  private static final long appStartMillis = SystemClock.uptimeMillis();
```

sentry-android-core/src/main/java/io/sentry/android/core/AppStartState.java
```diff
@@ -0,0 +1,70 @@
...
+
+  /** The type of App start coldStart=true -> Cold start, coldStart=false -> Warm start */
+  private boolean coldStart;
+
+  /** appStart as a Date used in the App's Context */
+  private @Nullable Date appStartTime;
```
philipphofmann, line 24: `m`: Can't we just ditch this variable and use `appStart` instead to create the Date here? This would be less complex and we would use less memory.

sentry-android-core/src/test/java/io/sentry/android/core/ActivityLifecycleIntegrationTest.kt
```diff
@@ -420,4 +421,60 @@ class ActivityLifecycleIntegrationTest {
...
+        sut.register(fixture.hub, fixture.options)
+
+        // set by SentryPerformanceProvider so forcing it here
+        AppStartState.getInstance().setAppStartTime(0, Date())
+
+        val activity = mock<Activity>()
+        sut.onActivityCreated(activity, null)
+        sut.onActivityResumed(activity)
+
+        // SystemClock.uptimeMillis() always returns 0, can't assert real values
```
philipphofmann, line 477: `m`: We could use a wrapper around `SystemClock.uptimeMillis()` to make it testable or do, for example, `AppStartState.getInstance().setAppStartTime(-1.5, Date())` to make it testable.

sentry-android-core/src/test/java/io/sentry/android/core/AppStartStateTest.kt
```diff
@@ -0,0 +1,44 @@
...
+import java.util.Date
+import kotlin.test.BeforeTest
+import kotlin.test.Test
+import kotlin.test.assertNull
+import kotlin.test.assertSame
+
+class AppStartStateTest {
+
+    @BeforeTest
+    fun `reset instance`() {
```
philipphofmann, line 12: `l`: Maybe also do a reset in `@AfterTest` to leave the static `AppStartState` in a clean state.

sentry-android-core/src/test/java/io/sentry/android/core/AppStartStateTest.kt
```diff
@@ -0,0 +1,44 @@
...
+        AppStartState.getInstance().resetInstance()
+    }
+
+    @Test
+    fun `appStartInterval returns null if end time is not set`() {
+        val sut = AppStartState.getInstance()
+
+        sut.setAppStartTime(0, Date())
```
philipphofmann, line 20: `l`: Using always the same date makes the test deterministic.
```suggestion
        sut.setAppStartTime(0, Date(0))
```

sentry-android-core/src/test/java/io/sentry/android/core/AppStartStateTest.kt
```diff
@@ -0,0 +1,44 @@
+package io.sentry.android.core
+
+import java.util.Date
+import kotlin.test.BeforeTest
+import kotlin.test.Test
+import kotlin.test.assertNull
+import kotlin.test.assertSame
+
+class AppStartStateTest {
```
philipphofmann, line 9: `m`: I think we are still missing a test to validate the calculation of `getAppStartInterval`.

sentry-android-core/src/main/java/io/sentry/android/core/PerformanceAndroidEventProcessor.java
```diff
@@ -0,0 +1,43 @@
+package io.sentry.android.core;
+
+import io.sentry.EventProcessor;
+import io.sentry.protocol.MeasurementValue;
+import io.sentry.protocol.SentryTransaction;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/** Event Processor responsable for adding Android metrics to transactions */
```
philipphofmann, line 9:
```suggestion
/** Event Processor responsible for adding Android metrics to transactions */
```

sentry-android-core/src/test/java/io/sentry/android/core/PerformanceAndroidEventProcessorTest.kt
```diff
@@ -0,0 +1,100 @@
...
+
+import com.nhaarman.mockitokotlin2.mock
+import io.sentry.SentryTracer
+import io.sentry.protocol.SentryTransaction
+import java.util.Date
+import kotlin.test.BeforeTest
+import kotlin.test.Test
+import kotlin.test.assertTrue
+
+class PerformanceAndroidEventProcessorTest {
```
philipphofmann, line 11: `m`: Nice tests. It would be nice to have some async tests to make sure the measurement is only sent once and to validate if the measurement value is correct. As pointed out in another test we could use `-1` or something for the startTime to be able to validate the interval.

sentry-android-core/src/test/java/io/sentry/android/core/SentryPerformanceProviderTest.kt
```diff
@@ -0,0 +1,37 @@
...
+        val mockContext = ContextUtilsTest.createMockContext()
+        providerInfo.authority = AUTHORITY
+
+        val provider = SentryPerformanceProvider()
+        provider.attachInfo(mockContext, providerInfo)
+
+        // done by ActivityLifecycleIntegration so forcing it here
+        AppStartState.getInstance().setAppStartEnd()
+
+        assertNotNull(AppStartState.getInstance().appStartInterval)
```
philipphofmann, line 31: `l`: We could assert for an expected interval and not just assert if it's not null.

philipphofmann (commented):
The first pass. Looks already pretty good. 👍

sentry/src/main/java/io/sentry/protocol/SentryTransaction.java
```diff
@@ -85,4 +88,8 @@ public boolean isSampled() {
     final SpanContext trace = this.getContexts().getTrace();
     return trace != null && Boolean.TRUE.equals(trace.getSampled());
   }
+
+  public @NotNull Map<String, @NotNull MeasurementValue> getMeasurements() {
```
bruno-garcia, line 92: At these times I'm happy we have a separate class for transaction and event

sentry-android-core/src/main/java/io/sentry/android/core/AppStartState.java
marandaneto: `appStartTime` is used for `contexts.app.appStartTime`, which was previously created in `DefaultAndroidEventProcessor` but it was not accurate as now, so I'd keep it as it is

sentry-android-core/src/main/java/io/sentry/android/core/AppStartState.java
```diff
@@ -0,0 +1,70 @@
...
+  void setColdStart(final boolean coldStart) {
+    this.coldStart = coldStart;
+  }
+
+  @Nullable
+  Date getAppStartTime() {
+    return appStartTime;
+  }
+
+  synchronized void setAppStartTime(final long appStart, final @NotNull Date appStartTime) {
```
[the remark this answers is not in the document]
marandaneto: yes, everything else happens only in the main thread, so no concurrency

sentry-android-core/src/test/java/io/sentry/android/core/ActivityLifecycleIntegrationTest.kt
marandaneto: yep, just wanted to make it simple now, similar provider is `CurrentDateProvider`

sentry-android-core/src/test/java/io/sentry/android/core/AppStartStateTest.kt
marandaneto: good point

sentry-android-core/src/test/java/io/sentry/android/core/AppStartStateTest.kt
marandaneto: it'd not harm but it'd not have any effect either, as it cleans the state before each test, keeping it as it is.

sentry-android-core/src/test/java/io/sentry/android/core/PerformanceAndroidEventProcessorTest.kt
marandaneto: I added a lock to `PerformanceAndroidEventProcessor` and there are tests to be sent only once, so the async tests would not make much difference now, with the volatile field yes.
there's a new test that checks the value of `getAppStartInterval`

marandaneto: `setAppStartEnd` reads from `SystemClock.uptimeMillis()` so it cant be negative

sentry-android-core/src/test/java/io/sentry/android/core/SentryPerformanceProviderTest.kt
marandaneto: it'd not be possible to mock `appStartMillis` and `appStartTime` anyway because we want to get the time during class load which is before we can inject something via ctor, if we do via ctor, we are losing some millis, that's why just testing if its at least set

philipphofmann, line 31: `l`: Maybe something like this would make the test a bit more valuable.
```suggestion
        val providerUptimeMillis = SystemClock.uptimeMillis()
        val provider = SentryPerformanceProvider()
        provider.attachInfo(mockContext, providerInfo)

        val expectedAppStartInterval = 10
        val appStartEndMillis = providerUptimeMillis + expectedAppStartInterval

        // done by ActivityLifecycleIntegration so forcing it here
        AppStartState.getInstance().setAppStartEnd(appStartEndMillis)

        assertTrue(expectedAppStartInterval <= AppStartState.getInstance().appStartInterval ?: 0)
```

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Feat: App start up metrics (#1487)
```
philipphofmann, line 5: `m`: It's measurements, not metrics and we also add now spans.
```suggestion
* Feat: Measure app start time (#1487)
```

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -116,8 +125,21 @@ private void startTracing(final @NonNull Activity activity) {
...
+      final Date appStartTime = AppStartState.getInstance().getAppStartTime();
+
+      // in case appStartTime isn't available, we don't create a span for it.
+      if (firstActivityCreated || appStartTime == null) {
+        transaction = hub.startTransaction(activityName, NAV_OP);
+      } else {
+        // start transaction with app start timestamp
+        transaction = hub.startTransaction(activityName, NAV_OP, appStartTime);
+        // start specific span for app start
+        // TODO: add description cold/warm
```
philipphofmann, line 140: `m`: TODO. Just as a reminder.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -116,8 +125,21 @@ private void startTracing(final @NonNull Activity activity) {
...
+        appStartSpan = transaction.startChild("app.start", appStartTime);
```
philipphofmann, line 141: `h`: We still need to align on Cocoa and Android with operation name and span operations for this.

sentry/src/main/java/io/sentry/ISpan.java
```diff
@@ -14,6 +16,10 @@
   @NotNull
   ISpan startChild(@NotNull String operation);
 
+  @ApiStatus.Internal
+  @NotNull
+  ISpan startChild(@NotNull String operation, @NotNull Date timestamp);
```
philipphofmann, line 21: `l`: Maybe to algin with the other methods?
```suggestion
  ISpan startChild(@NotNull String operation, @NotNull Date startTimestamp);
```

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -24,9 +24,16 @@
   private @NotNull String name;
 
   public SentryTracer(final @NotNull TransactionContext context, final @NotNull IHub hub) {
+    this(context, hub, null);
+  }
+
+  SentryTracer(
```
philipphofmann, line 30: `m`: Should we add tests for adding startTimestamp?

philipphofmann (commented):
I think we are still missing tests validating the span creation of the `ActivityLifecycleIntegration`. 

Nice seeing this going forward 🚀.

sentry-android-core/src/main/java/io/sentry/android/core/AppStartState.java
```diff
@@ -0,0 +1,73 @@
+package io.sentry.android.core;
+
+import android.os.SystemClock;
+import java.util.Date;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.jetbrains.annotations.TestOnly;
+
+/** AppStartState holds the state of the App Start metric and appStartTime */
+final class AppStartState {
```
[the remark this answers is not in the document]
marandaneto: only `setAppStartTime` may be called out of the main thread, it's already `synchronized`, `PerformanceAndroidEventProcessor` is also `synchronized`, but I made all the setters in `AppStartState` as well, good point

sentry-android-core/src/test/java/io/sentry/android/core/ActivityLifecycleIntegrationTest.kt
marandaneto: its not possible due to `SentryPerformanceProvider` relies on class load time, which is before the ctor of the class, so I cant really mock `SystemClock.uptimeMillis()` for that use case.

sentry-android-core/src/test/java/io/sentry/android/core/SentryPerformanceProviderTest.kt
marandaneto: its flaky, `SentryPerformanceProvider` is already loaded by the Android FW already, hence `SentryPerformanceProvider.appStartMillis` already has a value, so subtracting a value may work but it'd make it probably flaky, its a tricky use case.

sentry/src/main/java/io/sentry/ISpan.java
marandaneto: a child (Span) has a timestamp, a transaction has a startTimestamp, not the same thing

sentry-android-core/src/main/java/io/sentry/android/core/PerformanceAndroidEventProcessor.java
```diff
@@ -0,0 +1,46 @@
...
+    tracingEnabled = options.isTracingEnabled();
+  }
+
+  @Override
+  public @NotNull SentryTransaction process(
+      @NotNull SentryTransaction transaction, @Nullable Object hint) {
+    // the app start metric is sent only once when the 1st transaction happens
+    // after the app start is collected.
+    synchronized (measurementLock) {
+      if (!sentStartMeasurement && tracingEnabled) {
```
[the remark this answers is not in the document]
marandaneto: yep, I forgot to commit that, I cant add directly to the transaction though, because when the span starts, I still don't know when the app start finishes.

sentry-android-core/src/test/java/io/sentry/android/core/SentryPerformanceProviderTest.kt
philipphofmann: We could fix that by resetting the value with a method annotated with `@TestOnly`. No strong opinion here. You decide if you think it's worth the effort.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
philipphofmann: `app.start` it is 😁

philipphofmann (commented):
Almost done 👍

sentry-android-core/src/main/java/io/sentry/android/core/AppStartState.java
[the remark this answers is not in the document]
marandaneto: you can't change the value of a getter, it's a value not a reference type object

sentry-android-core/src/test/java/io/sentry/android/core/SentryPerformanceProviderTest.kt
marandaneto: good point, done

sentry-android-core/src/main/java/io/sentry/android/core/AppStartState.java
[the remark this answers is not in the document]
marandaneto: `getAppStartInterval` isn't a simple getter, its a method, its also `synchronized`

philipphofmann (approved):
Thanks for adapting my suggested changes for the tests. 👍 
LGTM 🚀

PR #1491, opened 2021-05-26
Feat: Add secondary constructor to SentryOkHttpInterceptor

maciejwalkowiak:
## :scroll: Description

Add secondary constructor to SentryOkHttpInterceptor.

Without secondary constructor it is inconvenient to pass `beforeSpan` in constructor without using named parameter or from Java.

## :pencil: Checklist

- [ ] No breaking changes

### DIFF
--- a/sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpInterceptor.kt
+++ b/sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpInterceptor.kt
@@ -15,6 +15,8 @@ class SentryOkHttpInterceptor(
     private val beforeSpan: BeforeSpanCallback? = null
 ) : Interceptor {
 
+    constructor(beforeSpan: BeforeSpanCallback) : this(HubAdapter.getInstance(), beforeSpan)
+
     override fun intercept(chain: Interceptor.Chain): Response {
         var request = chain.request()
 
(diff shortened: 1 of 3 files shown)


sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpInterceptor.kt
```diff
@@ -15,6 +15,8 @@ class SentryOkHttpInterceptor(
     private val beforeSpan: BeforeSpanCallback? = null
 ) : Interceptor {
 
+    constructor(beforeSpan: BeforeSpanCallback) : this(HubAdapter.getInstance(), beforeSpan)
```
marandaneto, line 18: do you need to pass `HubAdapter.getInstance()` again as its the default value in the main ctor?

marandaneto (approved):

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpInterceptor.kt
maciejwalkowiak: Unless there is a trick that I don't know this is not possible
image

marandaneto: mm yeah guess that's it then, coincides with the very same ctor.

PR #1496, opened 2021-05-28
Replace `clone()` with copy constructor 

### LINKED ISSUE #1482: Replace `clone()` with copy constructor
For creating copy of objects we use `clone` method and `Cloneable` interfaces, which means we need to deal with the `CloneNotSupportedExceptions` and keep in mind the "magic" of the clone method. More about this topic in https://www.artima.com/articles/josh-bloch-on-design#part13 (section "Copy Constructor versus Cloning").

We could use instead copy constructors to simplify our code.

maciejwalkowiak:
## :scroll: Description

Replace `clone()` with copy constructor 

Fixes #1482

Unit tests.

## :pencil: Checklist

- [x] No breaking changes


sentry/src/main/java/io/sentry/Breadcrumb.java
```diff
@@ -43,6 +43,21 @@ public Breadcrumb(final @NotNull Date timestamp) {
...
+    this.message = breadcrumb.message;
+    this.type = breadcrumb.type;
+    this.category = breadcrumb.category;
+    final Map<String, Object> dataClone = CollectionUtils.newConcurrentHashMap(breadcrumb.data);
+    if (dataClone != null) {
+      this.data = dataClone;
+    }
+    this.unknown = CollectionUtils.newConcurrentHashMap(breadcrumb.unknown);
+    final SentryLevel levelRef = breadcrumb.level;
+    this.level =
```
maciejwalkowiak, line 57: I don't think this is actually needed for enums, is it @marandaneto?

marandaneto: enum is a sort of singleton, I don't remember if I had an issue or just went ahead and "serialized" again, if you remove this and there are tests asserting that you don't mutate the copied value over, and tests are green, we can remove this

sentry/src/main/java/io/sentry/protocol/Device.java
```diff
@@ -102,6 +102,46 @@
...
+    this.id = device.id;
+    this.language = device.language;
+    this.connectionType = device.connectionType;
+    this.batteryTemperature = device.batteryTemperature;
+    this.batteryLevel = device.batteryLevel;
+    final String[] archsRef = device.archs;
+    this.archs = archsRef != null ? archsRef.clone() : null;
+
+    final TimeZone timezoneRef = device.timezone;
+    this.timezone = timezoneRef != null ? (TimeZone) timezoneRef.clone() : null;
```
marandaneto, line 140: should we address them?

maciejwalkowiak: I think JDK classes on which we use `clone` can stay, as here its not even clear how to copy timezone.

marandaneto: yeah its fine, timezone may throw `InternalError` but that means its a JVM bug so not worth recovering anyway
the String array also, String is immutable anyway, lets keep it

marandaneto (approved):

bruno-garcia (approved):
Nice!

PR #1500, opened 2021-05-31
Fix: Graceful Shutdown flushes event instead of Closing SDK

### LINKED ISSUE #1458: Supporting Graceful Shutdown
Sentry registers automatically during initialization a JVM Shutdown Hook to close the hub after shutdown has been **initiated**. Unfortunately the sentry context is closed to early and following events are discarded.

Spring introduced Graceful Shutdown Mechanism with 2.3.x. It would be great if the sentry context shutdown could be done as last as possible. The benefit would be that sentry events during the graceful shutdown phase could be send to the sentry server.

A quick solution could be a simple new flag `registerShutdownHook` in `SentryOptions` to disable the current registration of the shutdown hook via `ShutdownHookIntegration`. 
Afterwards it would be possible to implement a custom shutdown hook via Spring `SmartLifecycle`.

marandaneto:
## :scroll: Description
Fix: Graceful Shutdown flushes event instead of Closing SDK

Fix #1458

## :pencil: Checklist

- [ ] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-* Fix: Graceful Shutdown flushes event instead of Closing SDK
+* Fix: Graceful Shutdown flushes event instead of Closing SDK (#1500)
 
 # 5.0.0-beta.6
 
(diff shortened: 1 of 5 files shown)


bruno-garcia (approved):
@Tyrrrz we need the same thing in .NET. Basically `Dispose` should only flush, particularly from Unity and ASP.NET Core and other integrations application shutdown hook where we don't guarantee we're the last thing to shut down.

PR #1504, opened 2021-06-01
Fix: Add attachmentType to envelope ser/deser.

marandaneto:
## :scroll: Description
Fix: Add attachmentType to envelope ser/deser.

Extract from https://github.com/getsentry/sentry-java/pull/1256/files

## :pencil: Checklist

- [X] No breaking changes

### DIFF
--- a/sentry/src/main/java/io/sentry/Attachment.java
+++ b/sentry/src/main/java/io/sentry/Attachment.java
@@ -147,7 +147,8 @@ public final class Attachment {
   }
 
   /**
-   * Initializes an Attachment with a path, a filename, a content type, and addToTransactions.
+   * Initializes an Attachment with a path, a filename, a content type, addToTransactions, and
+   * attachmentType.
    *
    * <p>The file located at the pathname is read lazily when the SDK captures an event or
    * transaction not when the attachment is initialized. The pathname string is converted into an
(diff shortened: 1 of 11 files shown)


sentry/src/main/java/io/sentry/Attachment.java
```diff
@@ -140,6 +146,33 @@ public Attachment(
     this.addToTransactions = addToTransactions;
   }
 
+  /**
+   * Initializes an Attachment with a path, a filename, a content type, and addToTransactions.
```
philipphofmann, line 150: `l`:
```suggestion
   * Initializes an Attachment with a path, a filename, a content type, addToTransactions, and attachmentType.
```

sentry/src/main/java/io/sentry/SentryEnvelopeItemHeader.java
```diff
@@ -42,23 +43,36 @@ public int getLength() {
       final @NotNull SentryItemType type,
       int length,
       final @Nullable String contentType,
-      final @Nullable String fileName) {
+      final @Nullable String fileName,
+      final @Nullable String attachmentType) {
```
philipphofmann, line 47: `m`: I see that we often pass `null` for `attachmentType`. Does it maybe make sense to define an overload so we don't need to pass `null`?

philipphofmann (approved):
Just two minor comments. TIL. Didn't know the abbreviation ser/deser.

sentry/src/main/java/io/sentry/SentryEnvelopeItemHeader.java
marandaneto: makes sense

Swatinem (approved):

PR #1516, opened 2021-06-07
ref(changelog): Make versions heading level 2

iker-barriocanal:
Craft's [Changelog Policies](https://github.com/getsentry/craft#changelog-policies) show valid examples of changelogs. Versions should be in heading level 2 (instead of the current 1). This will allow automating the release process using Craft.

_#skip-changelog_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,5 @@
 # Changelog
 
-## Unreleased
-
 ## 5.0.0
 
 This release brings many improvements but also new features:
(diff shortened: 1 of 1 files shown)


marandaneto:
@iker-barriocanal you can use `_#skip-changelog_` in the PR description to make Danger CI step happy

marandaneto (approved):

marandaneto:
you probably need to merge main into this branch, there's a 5.0.1 changelog entry

PR #1518, opened 2021-06-07
feat(release): Delegate version bumping to Craft

iker-barriocanal:
At the moment, `buildVersionCode` and `versionName` have to be updated manually before doing a release. This PR adds a version bumping script, run by Craft on [`prepare`](https://github.com/getsentry/craft#pre-release-command).

- `buildVersionCode` will be incremented by 1.
- `versionName` will be the new version to prepare by Craft. For example, `craft prepare 6.0.0` will set `versionName=6.0.0`, and `craft prepare 6.1.0-alpha.1` will set `versionName=6.1.0-alpha.1`.

Note [Changelog Policies](6.1.0-alpha.1). Current policy is `simple`, so there must be an entry in the changelog with a matching version name for the new release passed to Craft; if not, the command will fail.

#skip-changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,6 @@
 
 ## Unreleased
 
-* Add version bumping script for Craft (#1518)
-
 ## 5.0.1
 
 * Fix: Sources and Javadoc artifacts were mixed up (#1515)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Add version bumping script for Craft.
+
 ## 5.0.1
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Add version bumping script for Craft (#1518)
```
marandaneto, line 5: maybe we use skip-changelog for such changes? the changelog is meant for users facing changes.

marandaneto (approved):

marandaneto:
@iker-barriocanal looks like we could change https://github.com/getsentry/craft#changelog-policies to `auto` then, as we currently maintain the `Unreleased` label

Issue #1374, opened 2021-04-05
Provide default fragment breadcrumb implementation


VladyslavKasprov:
Hi. We use a lot of fragments in our Android application, but we don't get any breadcrumbs for them. Is it possible to add default implementation for collecting fragment breadcrumbs?

P.S. From what I know, we'll have to manually set it in each activity using FragmentManager#registeeFragmenrLifecycleCallbacks.

marandaneto:
@VladyslavKasprov thanks for raising this issue, yes, we're aware of that, the main reason is, there's no global hook that we can access and add our lifecycle callback automatically, you'd need to do it manually for each `FragmentManager`.

2nd reason is that we'd need to depend on `androidx.fragment`, so ideally we'd have a new package so we don't overload our android-core dependency as not everyone uses fragments.

VladyslavKasprov:
Separate dependency and manual hooking would be fine) Thank you) Is there a sample implementation that we can use in the meantime?

marandaneto:
@VladyslavKasprov yes, take a look at the Activity lifecycle for inspiration https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java

if you wish to contribute with a PR, happy to guide :) ideally creating a new Gradle module etc

wzieba:
👋 @marandaneto I'm working currently on a `FragmentLifecycleIntegration` POC but I have a problem with how it should be registered.

My concern is that `SentryAndroid#init` initializes `Sentry SDK` with passed options, [including all default `Integration` objects](https://github.com/getsentry/sentry-java/blob/75a87428fee926621ac397a6ffdeff2499caea8e/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java#L112). It's okay as default `Integration`s rely on `Application` lifecycle so `Sentry SDK` can be initialized when an app is starting.

The challenge I see here is how to add `Integration` during the app runtime. I'd like to add `FragmentLifecycleIntegration` which requires `FragmentManager` later than SDK initialization because when app is starting, I don't have yet access to `FragmentManager`.

My question is: is there any way to add `Integration` after `Sentry SDK` is initialized? If not, can you see a different way to make `FragmentLifecycleIntegration` work?

marandaneto:
@wzieba hey, thanks for your feedback.

`Integrations` and `EventProcessor` can't be added at runtime if not using the passed options.

[...]

PR #1522, opened 2021-06-10
feat: add fragment lifecycle breadcrumb logger

wzieba:
## :scroll: Description

This PR adds a new module `sentry-android-fragment` which exposes a new `Integration` for automated Android's `Fragment`s lifecycles breadcrumb logging.

### Sample
image

Closes: #1374 

From suggested solutions by @marandaneto , I've liked 
> Maybe installing a new ActivityFragmentLifecycleIntegration that relies on androidx.fragment.app and then calling activity.getSupportFragmentManager(), to add and remove the FragmentLifecycleIntegration callback. 
> this would likely require a new package as the SDK does not depend on the fragments package, like sentry-android-fragment module.

the most, so I've followed it in this implementation.

It's also highly inspired by [`ActivityLifecycleIntegration`](https://github.com/getsentry/sentry-java/blob/112c46d04edfcdf1507f60bac8a401854c834457/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java).

Aside from unit tests, I've tested it manually using a sample project. To do this, uncomment manual initialization in `MyApplication`, add a test DSN, run app and click on `Open sample fragment` button. 

From there you can send a message from inner fragment to check if breadcrumbs logging works for nested fragments too.

## :pencil: Checklist

- [x] No breaking changes


sentry-android-fragment/src/main/java/io/sentry/android/fragment/FragmentLifecycleIntegration.kt
```diff
@@ -0,0 +1,66 @@
...
+
+    override fun onActivityStopped(activity: Activity) {
+        // no-op
+    }
+
+    override fun onActivitySaveInstanceState(activity: Activity, outState: Bundle) {
+        // no-op
+    }
+
+    override fun onActivityDestroyed(activity: Activity) {
```
wzieba, line 63: I don't unregister `SentryFragmentLifecycleCallback` as stated in docs:

> All registered callbacks will be automatically unregistered when this FragmentManager is destroyed.

[Source](https://developer.android.com/reference/androidx/fragment/app/FragmentManager?hl=id#registerFragmentLifecycleCallbacks(androidx.fragment.app.FragmentManager.FragmentLifecycleCallbacks,%20boolean))

We probably want a callbacks to be registered for the whole lifecycle of the `FragmentManager`.

sentry-android-fragment/src/main/java/io/sentry/android/fragment/FragmentLifecycleIntegration.kt
```diff
@@ -0,0 +1,66 @@
...
+    override fun close() {
+        application.unregisterActivityLifecycleCallbacks(this)
+        logger.log(DEBUG, "FragmentLifecycleIntegration removed.")
+    }
+
+    override fun onActivityCreated(activity: Activity, savedInstanceState: Bundle?) {
+        (activity as? FragmentActivity)
+            ?.supportFragmentManager
+            ?.registerFragmentLifecycleCallbacks(
+                SentryFragmentLifecycleCallbacks(hub),
```
wzieba, line 38: This creates a new instance of `SentryFragmentLifecycleCallbacks`. Another way would be to create `SentryFragmentLifecycleCallbacks` and reuse them for all activities. I'm not sure which one is preferred, WDYT?

sentry-android-fragment/src/main/java/io/sentry/android/fragment/FragmentLifecycleIntegration.kt
```diff
@@ -0,0 +1,66 @@
...
+    private val application: Application,
+    private val logger: ILogger
+) : ActivityLifecycleCallbacks, Integration, Closeable {
+
+    private lateinit var hub: IHub
+
+    override fun register(hub: IHub, options: SentryOptions) {
+        this.hub = hub
+
+        application.registerActivityLifecycleCallbacks(this)
```
wzieba, line 25: Differently from `ActivityLifecycleIntegration`, I don't add a flag in `SentryAndroidOptions` and [I don't check if this integration is enabled.](https://github.com/getsentry/sentry-java/blob/112c46d04edfcdf1507f60bac8a401854c834457/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java#L70)

The reason for that is that it's an opt-in feature and the user has to add a new module and integration manually. I think there's no point to hide it behind a flag in this case.

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/SampleInnerFragment.kt
```diff
@@ -0,0 +1,28 @@
...
+
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import androidx.fragment.app.Fragment
+import io.sentry.Sentry
+import io.sentry.samples.android.databinding.FragmentSampleInnerBinding
+
+class SampleInnerFragment : Fragment() {
```
wzieba, line 11: This inner fragment is made to show that registered `FragmentLifecycleCallbacks` is indeed recursive.

buildSrc/src/main/java/Config.kt
```diff
@@ -82,6 +83,8 @@ object Config {
         val retrofit2Gson = "$retrofit2Group:converter-gson:$retrofit2Version"
 
         val coroutinesCore = "org.jetbrains.kotlinx:kotlinx-coroutines-core:1.4.3"
+
+        val fragment = "androidx.fragment:fragment-ktx:1.3.4"
```
marandaneto, line 87: the `-ktx` is written in Kotlin so it'd be fine if the whole module is written in Kotlin, but the base of this library is written in Java and could be used in plain Java as well https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:fragment/fragment/src/main/java/androidx/fragment/app/

to be honest, I also prefer to be written in Kotlin, but we've always to think about a final user that has a Java app and don't want to bring the Kotlin stblib in as a transitive dependency.

@bruno-garcia and @philipphofmann opinions? I'd go with Kotlin IMO

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
```diff
@@ -0,0 +1,91 @@
...
+import android.view.View
+import androidx.fragment.app.Fragment
+import androidx.fragment.app.FragmentManager
+import androidx.fragment.app.FragmentManager.FragmentLifecycleCallbacks
+import io.sentry.Breadcrumb
+import io.sentry.IHub
+import io.sentry.SentryLevel.INFO
+
+@Suppress("TooManyFunctions")
+internal class SentryFragmentLifecycleCallbacks(private val hub: IHub) : FragmentLifecycleCallbacks() {
```
wzieba, line 14: I've intentionally missed 

- onFragmentPreAttached
- onFragmentPreCreated
- onFragmentActivityCreated

and it's a completely subjective decision but I believe they might create too much noise. WDYT? Is it okay or should I add some configuration for the clients about which callbacks do they want to add breadcrumbs for?

sentry-android-fragment/build.gradle.kts
```diff
@@ -0,0 +1,73 @@
...
+        versionName = project.version.toString()
+        versionCode = project.properties[Config.Sentry.buildVersionCodeProp].toString().toInt()
+
+        // for AGP 4.1
+        buildConfigField("String", "VERSION_NAME", "\"$versionName\"")
+    }
+
+    buildTypes {
+        getByName("debug")
+        getByName("release")
```
marandaneto, line 28: we probably need to have its own `proguard-rules.pro`, we don't want this class to be renamed, see eg the `sentry-android-timber` module

sentry-android-fragment/build.gradle.kts
```diff
@@ -0,0 +1,73 @@
...
+    }
+
```
marandaneto, line 30: missing the `testOptions` block here, I recall having troubles if not setting them

sentry-android-fragment/src/main/java/io/sentry/android/fragment/FragmentLifecycleIntegration.kt
```diff
@@ -0,0 +1,66 @@
...
+import io.sentry.IHub
+import io.sentry.ILogger
+import io.sentry.Integration
+import io.sentry.SentryLevel.DEBUG
+import io.sentry.SentryOptions
+import java.io.Closeable
+
+class FragmentLifecycleIntegration(
+    private val application: Application,
+    private val logger: ILogger
```
marandaneto, line 17: the logger can be read from `options.logger`, so a `lateinit` var would help too

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/MyApplication.java
```diff
@@ -16,13 +16,19 @@ public void onCreate() {
...
+    //    this,
+    //    options -> {
+    //      options.setBeforeSend(
+    //          (event, hint) -> {
+    //            event.setTag("sample-key", "before-send");
+    //            return event;
+    //          });
+    //      options.setAnrTimeoutIntervalMillis(2000);
+    //
+    // options.setDsn("https://d3964e5d568641c484c536600fb433a1@o552180.ingest.sentry.io/5677396");
```
wzieba, line 28: Whoops, it was meant to be the same DSN as in [AndroidManifest.xml](https://github.com/getsentry/sentry-java/blob/main/sentry-samples/sentry-samples-android/src/main/AndroidManifest.xml#L37). Will change.

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
marandaneto, line 14:
```suggestion
internal class SentryFragmentLifecycleCallbacks(private val hub: IHub =  HubAdapter.getInstance()) : FragmentLifecycleCallbacks() {
```

so in case people wanna use the `SentryFragmentLifecycleCallbacks` directly, its possible, the `HubAdapter` is just a wrapper around the Sentry static class to make things testable

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
```diff
@@ -0,0 +1,91 @@
...
+    override fun onFragmentDetached(fragmentManager: FragmentManager, fragment: Fragment) {
+        addBreadcrumb(fragment, "detached")
+    }
+
+    private fun addBreadcrumb(fragment: Fragment, state: String) {
+        val breadcrumb = Breadcrumb().apply {
+            type = "navigation"
+            setData("state", state)
+            setData("screen", getFragmentName(fragment))
+            category = "ui.lifecycle"
```
marandaneto, line 82: should we be a bit more specific here to differentiate between Activity and fragment? like `fragment.lifecycle` or `ui.fragment.lifecycle`?

sentry-android-fragment/src/test/java/io/sentry/android/fragment/FragmentLifecycleIntegrationTest.kt
```diff
@@ -0,0 +1,70 @@
...
+        val sut = fixture.getSut()
+        val fragmentManager = mock<FragmentManager>()
+        val fragmentActivity = mock<FragmentActivity> {
+            on { supportFragmentManager } doReturn fragmentManager
+        }
+
+        sut.register(mock(), SentryOptions())
+        sut.onActivityCreated(fragmentActivity, savedInstanceState = null)
+
+        verify(fragmentManager, times(1)).registerFragmentLifecycleCallbacks(check { fragmentCallbacks ->
```
marandaneto, line 57: `times(1)` is already the default value of the predicate, so you can omit it

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/MyApplication.java
marandaneto, line 28: even if initing manually, it still reads the values from manifest, so if you have your `dsn` set in there, you don't actually need it here, it does not hurt being here either

sentry-android-fragment/src/main/java/io/sentry/android/fragment/FragmentLifecycleIntegration.kt
marandaneto: good point, maybe we can add it as a comment and linking to the docs, just for reference

sentry-android-fragment/src/main/java/io/sentry/android/fragment/FragmentLifecycleIntegration.kt
marandaneto: I agree, its not needed

sentry-android-fragment/src/main/java/io/sentry/android/fragment/FragmentLifecycleIntegration.kt
marandaneto: I don't know the cost at runtime using one or the other, there's no static field so either as a singleton or a new instance will behave exactly the same, I'd keep creating a new instance though, because it receives a `hub` as a param, so it could be used standalone with a custom `hub`

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
marandaneto: we also dont add the `Pre` breadcrumbs for the activity lifecycle, also because those methods are only available in the newer SDK versions.
`onFragmentActivityCreated` is probably only noise since the activity lifecycle will have its own, I'd keep as it is

marandaneto:
@wzieba nicely done, thanks a lot, left a few comments only the Java or Kotlin would be a question to be answered, all the rest are easy fixes.
This integration will likely grow in features, eg creating spans (performance monitoring) automatically for every fragment

wzieba:
> nicely done, thanks a lot, left a few comments only the Java or Kotlin would be a question to be answered, all the rest are easy fixes.

@marandaneto thank you a lot for the review!

I'm now fixing formatting issues - I couldn't reproduce them locally as I was using Java 11.0 on my machine. After changing JDK to 1.8 (the same as on CI) I was able to reproduce&fix them.

>  This integration will likely grow in features, eg creating spans (performance monitoring) automatically for every fragment

Sure, that's a good idea to add spans for this integration too.

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/MyApplication.java
marandaneto: we can actually disable the auto-mode for this sample and init. the SDK manually, so the samples for fragments would work as well, the other option would be creating a new sample that hooks up the fragment integration.

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
wzieba: Sure, good idea! I like ui.fragment.lifecycle for the category name

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/MyApplication.java
wzieba: gotcha, thanks! I think I'll remove it to make things simpler.

sentry-android-fragment/src/main/java/io/sentry/android/fragment/FragmentLifecycleIntegration.kt
wzieba: Sure, good idea. Comment added in a1344ffe99dec542bf3ea10088191a00ae48c42f

sentry-android-fragment/build.gradle.kts
wzieba: Sure, added in 1b538250268295dda65d75de73d254e5e4381946

sentry-android-fragment/build.gradle.kts
wzieba: Gotcha, I've removed it initially as those seemed not needed at a first glance, thanks for noticing. Added in 380b732a4181b3ca4efa041a665ca40fbc999e04

sentry-android-fragment/src/main/java/io/sentry/android/fragment/FragmentLifecycleIntegration.kt
wzieba: Right! Fixed in 09f1c6305cba5507e15f04eec13f7df46ab7ac3d

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
wzieba: Sure,

> in case people wanna use the SentryFragmentLifecycleCallbacks directly

so I also removed `internal` modifier. Done in 099ad04ccaa3fcfe85b20179f505995dad44293c

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
wzieba: Done in b673c00d95f97381da577259a043d6e6eaf7b142

sentry-android-fragment/src/test/java/io/sentry/android/fragment/FragmentLifecycleIntegrationTest.kt
wzieba: Good point, done in 910f3ed72c8c37f72002fdaf84e4cf83d72b8547

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/MyApplication.java
wzieba: Removed in 8090b9e7297431492d416dc72944967cb98a7c58

wzieba: I think it's a good call with disabling auto-mode here. I've enabled manual initialization in 513a3ec5a91f04bde6b5365a1f51af111d4220f0 - is this ok or you'd prefer to move setting `dsn` from `AndroidManifest` to `SentryAndroidOptions`?

bruno-garcia (commented):
wow this is beautiful stuff. Thanks @wzieba

sentry-android-fragment/build.gradle.kts
marandaneto: it was more of a problem when running `@RunWith(AndroidJUnit4::class)`, so lets keep it for the sake of safety

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
```diff
@@ -0,0 +1,94 @@
...
+import androidx.fragment.app.FragmentManager
+import androidx.fragment.app.FragmentManager.FragmentLifecycleCallbacks
+import io.sentry.Breadcrumb
+import io.sentry.HubAdapter
+import io.sentry.IHub
+import io.sentry.SentryLevel.INFO
+
+@Suppress("TooManyFunctions")
+class SentryFragmentLifecycleCallbacks(
+    private val hub: IHub = HubAdapter.getInstance()
```
marandaneto, line 16: let's also add a ctor overload for better Java compatibility, so the `hub` does not need to be passed at all, similar to https://github.com/getsentry/sentry-java/pull/1491/files

marandaneto (approved):
@wzieba good stuff, thanks a lot, also pretty nice the comments linking the commit hash, so easy to review, nicely done.
left 2 small comments, other than that LGTM.
I
l'll cut a release early next week :)

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
wzieba: Sure, but I think we already have that: there are two constructors for `SentryFragmentLifecycleCallbacks` in Java, and one of them is parameterless. You can check this on [generated API file](https://github.com/getsentry/sentry-java/pull/1522/files#diff-015ad5258f9a132c88f9e5da60e6137c5f55f6b9d031d6776d6742e15b8363e4R23) or here:

image

Looking how it'd look like in Java: invoking a parameterless constructor from 134 will invoke one from 124 and it'll use the default value: `HubAdapter.getInstance();`

image

--- 

I was surprised it works like this but further tests proved that if all arguments in the constructor have a default value, Kotlin will create a parameterless constructor too. I've found it also in docs [here](https://kotlinlang.org/docs/classes.html#secondary-constructors), starting with `On the JVM, if all of`

Is it okay to leave it as it is?

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/MyApplication.java
```diff
@@ -15,14 +15,17 @@ public void onCreate() {
...
-    // });
+    SentryAndroid.init(
+        this,
+        options -> {
+          options.setBeforeSend(
+              (event, hint) -> {
+                event.setTag("sample-key", "before-send");
+                return event;
+              });
+          options.setAnrTimeoutIntervalMillis(2000);
```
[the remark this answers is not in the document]
wzieba: Sure, commented in 9910d77de94f87fdd581ba4228c0a5c656ab0375

wzieba:
Thank you @marandaneto , it was a pleasure working on this. I've left [one comment](https://github.com/getsentry/sentry-java/pull/1522#discussion_r649790745) about parameterless constructor - please take a look and let me know if it needs changes.

>  l'll cut a release early next week :)

Great, thanks!

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
marandaneto: indeed, good point, TIL, I just tested and it works, thanks.

Side note (not related to this PR): not all the cases are covered though, just opened this issue: https://github.com/getsentry/sentry-java/issues/1526

sentry-android-fragment/src/main/java/io/sentry/android/fragment/FragmentLifecycleIntegration.kt
```diff
@@ -0,0 +1,74 @@
...
+        this.hub = hub
+        this.logger = options.logger
+
+        application.registerActivityLifecycleCallbacks(this)
+        logger.log(DEBUG, "FragmentLifecycleIntegration installed.")
+    }
+
+    override fun close() {
+        application.unregisterActivityLifecycleCallbacks(this)
+        logger.log(DEBUG, "FragmentLifecycleIntegration removed.")
```
marandaneto, line 33: if you don't check if `logger` is `isInitialized`, it'd crash the app when `close` is called, this would be a bug in the SDK, but always better to prevent it, so a simple check would save us some headache.

wzieba: Right, good point. Added check in 87275f641a00eee1dcdec96f1045c538f94f0442

Issue #1523, opened 2021-06-10
Support transaction finishing automatically with 'idle timeout'


bruno-garcia:
The goal is to have a mode for finishing transactions automatically once no new spans are added to it for N milliseconds, or a general timeout. 

We need to define the spec in the develop.sentry.dev docs and align with other implementations of it. This is an approach used by the JS on the browser (and React Native).

Note On Android: we introduced an opt-out for the auto activity transaction from closing on rendering (it should only add/end span instead of the whole transaction). In that case might not use the idle timeout transaction when that flag is opt-in and rely on the user programatic `finish` call instead.

philipphofmann:
Do we also need this on Cocoa? We currently have there a different approach with waiting for all spans to finish.

bruno-garcia:
@philipphofmann I was just talking about this with @maciejwalkowiak and this other approach came up. Possibly we could have two modes? I'm unsure yet how the 'finish transaction when all spans end' mode will work. For example:

What happens when:
* No span is created from the transaction?
* A span never finishes (i.e: it became a worker thread that doesn't end)

Sounds like we need to discuss this further because likely we can just align on a single approach on all mobile SDKs.

AbhiPrasad:
Figured I'd give some context here for those interested:

JS Browser implementation: https://github.com/getsentry/sentry-javascript/blob/master/packages/tracing/src/idletransaction.ts

An `IdleTransaction` keeps track child spans through the concept of activities. Activities are typed as `Set<span_id>`, where an activity is added when a child span is started, and removed when a child span is finished. When an activity is removed, and the `activities` set becomes empty, the `idleTransaction` finishes itself.

The `IdleTransaction` exposes functionality to register `beforeFinish` callbacks https://github.com/getsentry/sentry-javascript/blob/master/packages/tracing/src/idletransaction.ts#L159, which runs right before the transaction is sent to Sentry. This has to be done because the `IdleTransaction` finishes at an unknown time, so these callbacks have to be set up ahead of time. The browser tracing integration uses this to add measurements and additional spans based on heuristics (like browser specific resource spans). 

In addition, the `IdleTransaction` also trims its end timestamp to match the time of it's last child span end timestamp (as its end timestamp is kind of arbitrary, doesn't have that much value).

Let's talk about some possible gotchas:

1. No child spans are created:

[...]

PR #1528, opened 2021-06-11
Feat: Perf. for fragments

marandaneto:
## :scroll: Description
Feat: Perf. for fragments

Able to start spans out of fragments

## :pencil: Checklist

- [ ] No breaking changes


sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/ThirdActivityFragment.kt
```diff
@@ -0,0 +1,41 @@
...
+        // this is called before fragments are fully rendered, as they have their own lifecycle
+        // activity is only responsible of rendering the containers for fragments (FragmentContainerView)
+        // so I cant finish my transaction here manually.
+        // we'd need idle transactions here, I will simulate with a delayed coroutines
+
+        GlobalScope.launch {
+            delay(1500L)
+            val span = Sentry.getSpan()
+            span?.finish(SpanStatus.OK)
+        }
```
marandaneto, line 39: @bruno-garcia

bruno-garcia: We'll chat about this on Monday and discuss the idle transaction vs the approach taken on iOS.
From there we should have next steps

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
```diff
@@ -91,4 +107,43 @@ class SentryFragmentLifecycleCallbacks(
...
+        if (!isPerformanceEnabled || isRunningSpan(fragment)) {
+            return
+        }
+
+        val currentSpan = Sentry.getSpan()
+
+        val span: ISpan
+        val fragmentName = getFragmentName(fragment)
+
+        // or ui.load too?
```
bruno-garcia, line 124: I think it's nice to be specific here to leverage the 'ops breakdown'
```suggestion
```

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
```diff
@@ -91,4 +107,43 @@ class SentryFragmentLifecycleCallbacks(
...
+        val op = "fragment.load"
+
+        // should be a span of the activity transaction or its own transaction?
+        span = currentSpan?.startChild(op, fragmentName)
+            ?: Sentry.startTransaction(fragmentName, op)
```
bruno-garcia, line 129:
```suggestion
            ?: hub.startTransaction(fragmentName, op)
```

marandaneto: good catch

sentry-samples/sentry-samples-android/src/main/AndroidManifest.xml
```diff
@@ -79,7 +81,7 @@
         <!--    <meta-data android:name="io.sentry.traces.activity.enable" android:value="false" />-->
 
         <!--    how to disable the Activity auto instrumentation automatically finished on onActivityPostPaused-->
-        <meta-data android:name="io.sentry.traces.activity.auto-finish.enable" android:value="false" />
+<!--        <meta-data android:name="io.sentry.traces.activity.auto-finish.enable" android:value="false" />-->
```
marandaneto, line 84: the `waitForChildren` pretty much deprecates this auto-finish

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -134,10 +135,14 @@ private void startTracing(final @NonNull Activity activity) {
...
-        transaction = hub.startTransaction(activityName, UI_LOAD_OP);
+        transaction =
+            hub.startTransaction(
+                new TransactionContext(activityName, UI_LOAD_OP), null, false, null, true);
       } else {
         // start transaction with app start timestamp
-        transaction = hub.startTransaction(activityName, UI_LOAD_OP, appStartTime);
+        transaction =
+            hub.startTransaction(
+                new TransactionContext(activityName, UI_LOAD_OP), null, false, appStartTime, true);
```
marandaneto, line 145: @maciejwalkowiak I am branched from your branch `gh-1523`, can we make an easy overload for such option? that allows name, op and waitForChildren, with or without `startTimestamp`

maciejwalkowiak: Sorry for late reply. This overload is already merged to master https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/IHub.java#L387

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
```diff
@@ -38,6 +55,8 @@ class SentryFragmentLifecycleCallbacks(
         savedInstanceState: Bundle?
     ) {
         addBreadcrumb(fragment, "created")
+
+        startTracing(fragment)
```
bruno-garcia, line 59: Seems like we have a single instance of this class for the whole app regardless of Sentry.init or Hub.
We could have a publish subscriber apporach and have this class only 'raise events'. And a FragmentIntegration that upon `register` we subscribe to events from this class.

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
```diff
@@ -91,4 +117,45 @@ class SentryFragmentLifecycleCallbacks(
...
+
+        val fragmentName = getFragmentName(fragment)
+        val span = transaction?.startChild(FRAGMENT_LOAD_OP, fragmentName)
+
+        span?.let {
+            fragmentsWithOngoingTransactions[fragment] = it
+        }
+    }
+
+    private fun stopTracing(fragment: Fragment) {
```
bruno-garcia, line 142: Any chance this isn't called ever? I wonder if it's possible to leak fragments in `fragmentsWithOngoingTransactions`

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
```diff
@@ -55,6 +74,8 @@ class SentryFragmentLifecycleCallbacks(
 
     override fun onFragmentResumed(fragmentManager: FragmentManager, fragment: Fragment) {
         addBreadcrumb(fragment, "resumed")
+
+        stopTracing(fragment)
```
bruno-garcia, line 78: These always run in the UI thread. rigth? So we're good with `mutable map` and no synchronization here?

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
```diff
@@ -9,13 +9,30 @@ import androidx.fragment.app.FragmentManager.FragmentLifecycleCallbacks
...
 import io.sentry.HubAdapter
 import io.sentry.IHub
+import io.sentry.ISpan
 import io.sentry.SentryLevel.INFO
+import io.sentry.SpanStatus
 
 @Suppress("TooManyFunctions")
 class SentryFragmentLifecycleCallbacks(
-    private val hub: IHub = HubAdapter.getInstance()
+    private val hub: IHub = HubAdapter.getInstance(),
```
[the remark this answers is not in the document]
marandaneto: `HubAdapter` is just a wrapper around the `Sentry` static class, to make it testable, it's going to use the active `hub` in the Sentry class, or am I missing something?

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
```diff
@@ -9,13 +9,30 @@ import androidx.fragment.app.FragmentManager.FragmentLifecycleCallbacks
...
+    constructor(
+        enableFragmentLifecycleBreadcrumbs: Boolean = true,
+        enableAutoFragmentLifecycleTracing: Boolean = false
+    ) : this(
+        hub = HubAdapter.getInstance(),
+        enableFragmentLifecycleBreadcrumbs = enableFragmentLifecycleBreadcrumbs,
+        enableAutoFragmentLifecycleTracing = enableAutoFragmentLifecycleTracing
+    )
+
+    private val isPerformanceEnabled = hub.options.isTracingEnabled && enableAutoFragmentLifecycleTracing
```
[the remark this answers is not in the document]
philipphofmann: Good catch @bruno-garcia. I think this would to the trick
```suggestion
    private val isPerformanceEnabled get() = hub.options.isTracingEnabled && enableAutoFragmentLifecycleTracing
```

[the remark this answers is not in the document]
marandaneto: its a new instance per activity, as the callback is bound to the fragment manager, and each activity has its own fragmentmanager, I guess its the same as https://github.com/getsentry/sentry-java/pull/1528#discussion_r658519308
so in the case that a new activity is setup and `isTracingEnabled` is disabled, nothing is gonna happen.

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
```diff
@@ -9,13 +9,30 @@ import androidx.fragment.app.FragmentManager.FragmentLifecycleCallbacks
...
+
+    private val fragmentsWithOngoingTransactions = mutableMapOf<Fragment, ISpan>()
```
philipphofmann, line 34: `h`: Should we use a `WeakReference` for the fragment here?

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
marandaneto: It's an instance per activity, and it does the same approach as `ActivityLifecycleIntegration`, we could go with pub/sub yes, but I am not sure that it brings any benefit compared with the current approach, mind elaborating?

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
marandaneto: we follow the fragment's lifecycle, same with the Activity lifecycle, if the methods don't get called, it's a bug in the androidx library.

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
marandaneto: yes, same with `ActivityLifecycleIntegration`, no need for synchronization or thread-safe map.

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
```diff
@@ -91,4 +117,45 @@ class SentryFragmentLifecycleCallbacks(
...
+        val fragmentName = getFragmentName(fragment)
+        val span = transaction?.startChild(FRAGMENT_LOAD_OP, fragmentName)
+
+        span?.let {
+            fragmentsWithOngoingTransactions[fragment] = it
+        }
+    }
+
+    private fun stopTracing(fragment: Fragment) {
+        if (!isPerformanceEnabled || !isRunningSpan(fragment)) {
```
philipphofmann, line 143: `h`: If `isPerformanceEnabled` is evaluated on each call it could be that it is disabled here and never finish the span. If `isPerformanceEnabled` is set to `false` and we have a running span we could just discard the span.

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
```diff
@@ -91,4 +117,45 @@ class SentryFragmentLifecycleCallbacks(
...
+            if (status == null) {
+                status = SpanStatus.OK
+            }
+            it.finish(status)
+            fragmentsWithOngoingTransactions.remove(fragment)
+        }
+    }
+
+    companion object {
+        const val FRAGMENT_LOAD_OP = "fragment.load"
```
philipphofmann, line 159: `h`: Shouldn't this be?
```suggestion
        const val FRAGMENT_LOAD_OP = "ui.load"
```

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
```diff
@@ -71,13 +92,18 @@ class SentryFragmentLifecycleCallbacks(
 
     override fun onFragmentDestroyed(fragmentManager: FragmentManager, fragment: Fragment) {
         addBreadcrumb(fragment, "destroyed")
+
+        stopTracing(fragment)
```
philipphofmann, line 96: `h`: On Cocoa, we finish on `onResume`. The duration of the fragment transaction would vary greatly.

philipphofmann (commented):
On Cocoa, we only track spans until `onResume` is called because otherwise, the duration of a transaction would vary vastly. We could also create spans for `onCreate`, `onStart` and `onResume`. Then you get more insight into how long each of the lifecycle methods took.

marandaneto:
> On Cocoa, we only track spans until `onResume` is called because otherwise, the duration of a transaction would vary vastly. We could also create spans for `onCreate`, `onStart` and `onResume`. Then you get more insight into how long each of the lifecycle methods took.

the idea is that first, we measure the impact of adding spans for every single lifecycle method and later expand upon feedback, fragment has like 12 lifecycle methods, it's too noisy

sentry-android-core/src/test/java/io/sentry/android/core/ActivityLifecycleIntegrationTest.kt
```diff
@@ -41,7 +42,7 @@ class ActivityLifecycleIntegrationTest {
         fun getSut(apiVersion: Int = 29): ActivityLifecycleIntegration {
             whenever(hub.startTransaction(any<String>(), any())).thenReturn(transaction)
             whenever(hub.options).thenReturn(options)
-            whenever(hub.startTransaction(any(), any(), any<Date>())).thenReturn(transaction)
+            whenever(hub.startTransaction(any(), any(), anyOrNull<Date>(), any())).thenReturn(transaction)
```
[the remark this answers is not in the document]
marandaneto: I do the check the `verify` test already, no need to do in the `whenever`

sentry-android-fragment/src/main/java/io/sentry/android/fragment/FragmentLifecycleIntegration.kt
```diff
@@ -6,40 +6,48 @@ import android.app.Application.ActivityLifecycleCallbacks
...
 import io.sentry.Integration
 import io.sentry.SentryLevel.DEBUG
 import io.sentry.SentryOptions
 import java.io.Closeable
 
-class FragmentLifecycleIntegration(private val application: Application) :
+class FragmentLifecycleIntegration(
+    private val application: Application,
+    private val enableFragmentLifecycleBreadcrumbs: Boolean,
+    private val enableAutoFragmentLifecycleTracing: Boolean
```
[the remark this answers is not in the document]
marandaneto: I would keep as it is because the activity one is called `enableAutoActivityLifecycleTracing`

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
```diff
@@ -9,13 +9,30 @@ import androidx.fragment.app.FragmentManager.FragmentLifecycleCallbacks
 import io.sentry.Breadcrumb
 import io.sentry.HubAdapter
 import io.sentry.IHub
+import io.sentry.ISpan
 import io.sentry.SentryLevel.INFO
+import io.sentry.SpanStatus
 
 @Suppress("TooManyFunctions")
 class SentryFragmentLifecycleCallbacks(
```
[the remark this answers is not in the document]
marandaneto: if we depend on a library that uses Kotlin already, no reason to not do so :D
same for okhttp and timber integrations

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
marandaneto: good catch, we use `WeakHashMap` for activities, I will use the same then

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
marandaneto: you cannot mutate the `isPerformanceEnabled` after the SDK is inited, so I don't see the point of taking care of this corner case, the same happens on `ActivityLifecycleIntegration`

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
marandaneto: the transaction itself is `ui.load`, but this is the `op` of the span only, so it'd be nice to differ otherwise it'd be hard to understand what is the activity and fragment spans

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
marandaneto: we also finish on `onFragmentResumed`, this is more like a safe guard to not leak anything when the fragment gets destroyed

sentry-samples/sentry-samples-android/api/sentry-samples-android.api
```diff
@@ -11,6 +11,11 @@ public abstract interface class io/sentry/samples/android/GitHubService {
 	public abstract fun listRepos (Ljava/lang/String;)Lretrofit2/Call;
 }
 
+public final class io/sentry/samples/android/GithubAPI {
```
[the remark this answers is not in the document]
marandaneto: yep, good idea https://github.com/getsentry/sentry-java/issues/1564

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
philipphofmann: Couldn't you do some crazy stuff like this?
```Java
public class MyApplication extends Application {

  private SentryOptions myOptions;

  @Override
  public void onCreate() {
    super.onCreate();
    
    SentryAndroid.init(
        this,
        options -> {
          myOptions = options;
        });

    // some time passes
    myOptions.setTracesSampleRate(null);
  }
}
```

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
philipphofmann: But I think the prefix matters to the Sentry UI. Maybe `ui.fragment.load`? I would clarify this with the visibility team.

philipphofmann:
> the idea is that first, we measure the impact of adding spans for every single lifecycle method and later expand upon feedback, fragment has like 12 lifecycle methods, it's too noisy

Didn't know that. We currently don't have much "noise" on Android, so adding some makes sense to me as performance is about getting more insight.

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
marandaneto: you could, yes, this would be misusing the API hence not really a case that we should take care, if we take this option in consideration, there are N cases that would break the whole SDK, or using reflection and setting non null fields to null etc

marandaneto: also `init` should be called only once, its part of the spec.

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
marandaneto: `ui.fragment.load` sounds good to me.

marandaneto:
> > the idea is that first, we measure the impact of adding spans for every single lifecycle method and later expand upon feedback, fragment has like 12 lifecycle methods, it's too noisy
> 
> Didn't know that. We currently don't have much "noise" on Android, so adding some makes sense to me as performance is about getting more insight. But I'm fine with the let's see how it goes approach.

we've discussed this for Activity instrumentation, I'm just taking the same approach.
the difference is also that on Activitys and Fragments lifecycle, most of the work, if not only, is in the onCreate method, nothing else.

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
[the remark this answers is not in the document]
bruno-garcia: The problem is that when you do:

```
IHub hub = HubAdapter.getInstance();
Sentry.init();

hub.capture() -> This won't work because. 

But if you do `HubAdapter.getInstance().capture` that would work.

So the idea is to always do `HubAdapter.getInstance()` (through a local `Callable<IHub>`)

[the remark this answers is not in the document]
bruno-garcia: Fair, the Hub instance is always passing the call through to the static class so changes to the Hub instnace bound to that won't take effect here.

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
bruno-garcia: Clarified on a call, makes sense

bruno-garcia (approved):
:shipit:

PR #999, opened 2020-10-18
Webflux support

dbuos:
A working integration with spring-webflux. Associates the Hub with the reactive pipeline and captures unhandled errors in webflux applications.
related to: https://github.com/getsentry/sentry-java/issues/640

Issue #640, opened 2018-11-16
sentry integration with spring webflux


eddumelendez:
new spring boot starter module was added in #636. This is relying in spring boot `1.5.x` release. My question is how sentry should be configured for apps who do not rely in servlet implementations I mean reactive apps.

jnfeinstein:
Has WebFlux support been added?  I'm seeing a number of thread-bound resources and servlet-specific behaviors in the code.

weekens:
Hey guys! In WebFlux we do it like this:

```kotlin
    @Bean
    fun sentryExceptionHandler(): ErrorWebExceptionHandler {
        // Sentry DSN is taken from SENTRY_DSN environment variable.
        // If it is not defined, a client becomes no-op and does not
        // consume resources.
        val sentryClient = SentryClientFactory.sentryClient()
        sentryClient.release = appVersion

        return ErrorWebExceptionHandler { _, ex ->
            // Message is sent asynchronously using inner thread pool.
            sentryClient.sendException(ex)

            Mono.error(ex)
        }
    }
```

jnfeinstein:
@weekens I don't think you're capturing the HTTP properties.

weekens:
@jnfeinstein Yeah, you're right. The first parameter (`_`) in lambda is actually an HTTP exchange object, all HTTP properties can be taken from there. I just didn't figure out, where to put them in `sentryClient`.

pulse00:
@bruno-garcia thanks for merging #838 . 

@bretthoerner I'll be working on springwebflux integration next because we neDocumentEnvelopeFiltered this in our projects. If it's possible to have generic-enough support for webflux / project reactor, i'll open another PR with it - would you be interested? 

If yes - does your team have any ideas so far with adding support? I've a couple of approaches in my local fork, but i'm not quite happy so far.

marandaneto:
hey @pulse00 thanks for your contributions :)

I believe we don't have any work going on for webflux/project reactor.
not sure if @bruno-garcia has something in mind, otherwise PRs are welcome.

would you like to point us to these approaches so we can have a look at it? like a short description or even linking to the classes in your fork would be enough. thanks!

marandaneto:
> @jnfeinstein Yeah, you're right. The first parameter (`_`) in lambda is actually an HTTP exchange object, all HTTP properties can be taken from there. I just didn't figure out, where to put them in `sentryClient`.

[...]

Issue #1525, opened 2021-06-11
SentryExceptionResolver mechanism should have a `type`


marandaneto:
```
    final Mechanism mechanism = new Mechanism();
    mechanism.setHandled(false);
    final Throwable throwable =
        new ExceptionMechanismException(mechanism, ex, Thread.currentThread());
```

a tag is created automatically from the mechanism type and it helps to filter events by mechanism type.
let's add a type to `SentryExceptionResolver` mechanism. 

https://develop.sentry.dev/sdk/event-payloads/exception/#exception-mechanism

eg `HandlerExceptionResolver`

marandaneto:
cc @maciejwalkowiak

Issue #1524, opened 2021-06-11
Support SentryCoroutineExceptionHandler﻿


marandaneto:
As we have a `sentry-kotlin-extensions` module right now and `SentryContext` for Coroutines, it'd be a nice addition offering a `SentryCoroutineExceptionHandler﻿`, https://kotlinlang.org/docs/exception-handling.html#coroutineexceptionhandler
so people can use it and when it throws, it captures an event automatically, how does that sound?
Similar to our `UncaughtExceptionHandlerIntegration`

marandaneto:
```
val handler = SentryCoroutineExceptionHandler﻿ { _, exception -> 
    // ideally we create an event with a mechanism
    Sentry.captureException(exception)
}
val job = GlobalScope.launch(handler) {
    throw AssertionError()
}
```

marandaneto:
examples can be found on https://github.com/Kotlin/kotlinx.coroutines/tree/master/kotlinx-coroutines-core/jvm/test/guide
example-exceptions-{number}-kt files

marandaneto:
works like a try/catch, if no `CoroutineExceptionHandler﻿` is given, `UncaughtExceptionHandlerIntegration` gets called and crashes the App, but if a `CoroutineExceptionHandler﻿` is given, it gets caught and swallowed, so its a bit less boilerplate than a try/catch but works the same way

marandaneto:
one thing to note is:

> [CoroutineExceptionHandler] can be invoked from an arbitrary thread

so we cannot use the current thread when creating a `ExceptionMechanismException`

marandaneto:
```
public object SentryCoroutineExceptionHandler : AbstractCoroutineContextElement(CoroutineExceptionHandler), CoroutineExceptionHandler {
    override fun handleException(context: CoroutineContext, exception: Throwable) {
        val mechanism = Mechanism().apply {
            // null and true is the same
            // isHandled = true
            type = "CoroutineExceptionHandler"
        }
        // cannot use the Thread.currentThread()
        // [CoroutineExceptionHandler] can be invoked from an arbitrary thread
        val error = ExceptionMechanismException(mechanism, exception, Thread.currentThread())
        val event = SentryEvent(error)
        Sentry.captureEvent(event)
    }
}
```

```
            GlobalScope.launch(SentryCoroutineExceptionHandler) {
                throw AssertionError()
            }
```

[...]

PR #1529, opened 2021-06-11
Experimental Spring Webflux integration

maciejwalkowiak:
## :scroll: Description

**Experimental** Integration for Spring Webflux with Project Reactor.

Support reactive Spring stack. #640

Integration tests.

## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next steps

- add reactive Spring Security integration for obtaining current user (depending on effort needed, can be added in a separate PR)
- update reference docs


CHANGELOG.md
```diff
@@ -3,7 +3,8 @@
 ## Unreleased
 
 * Feat: Measure app start time (#1487)
-* Feat: Automatic breadcrumbs logging for fragment lifecycle (#1522) 
+* Feat: Automatic breadcrumbs logging for fragment lifecycle (#1522)
```
marandaneto, line 6: merge issue?

sentry-spring/src/main/java/io/sentry/spring/webflux/SentryWebExceptionHandler.java
```diff
@@ -0,0 +1,46 @@
...
+    this.hub = Objects.requireNonNull(hub, "hub is required");
+    this.transactionNameProvider = new TransactionNameProvider();
+  }
+
+  @Override
+  public @NotNull Mono<Void> handle(
+      final @NotNull ServerWebExchange serverWebExchange, final @NotNull Throwable ex) {
+    if (!(ex instanceof ResponseStatusException)) {
+      final Mechanism mechanism = new Mechanism();
+      mechanism.setHandled(false);
```
marandaneto, line 36: does this event crash the application? handled=false is only if it does, otherwise it's handled=true or null (same effect)

sentry-spring/src/main/java/io/sentry/spring/webflux/SentryWebExceptionHandler.java
```diff
@@ -0,0 +1,46 @@
...
+  public SentryWebExceptionHandler(final @NotNull IHub hub) {
+    this.hub = Objects.requireNonNull(hub, "hub is required");
+    this.transactionNameProvider = new TransactionNameProvider();
+  }
+
+  @Override
+  public @NotNull Mono<Void> handle(
+      final @NotNull ServerWebExchange serverWebExchange, final @NotNull Throwable ex) {
+    if (!(ex instanceof ResponseStatusException)) {
+      final Mechanism mechanism = new Mechanism();
```
marandaneto, line 35: `mechanism.type` is mandatory, similar to https://github.com/getsentry/sentry-java/issues/1525

sentry-spring/src/main/java/io/sentry/spring/webflux/SentryWebFilter.java
```diff
@@ -0,0 +1,44 @@
...
+import reactor.core.publisher.Mono;
+
+/** Manages {@link io.sentry.Scope} in Webflux request processing. */
+@ApiStatus.Experimental
+public final class SentryWebFilter implements WebFilter {
+  private final @NotNull IHub hub;
+  private final @NotNull SentryRequestResolver sentryRequestResolver;
+
+  public SentryWebFilter(final @NotNull IHub hub) {
+    this.hub = hub;
```
marandaneto, line 20: do we need `requireNonNull`?

sentry-spring/src/main/java/io/sentry/spring/webflux/SentryWebExceptionHandler.java
maciejwalkowiak: It does not crash in a sense that application shuts down and stops processing more requests. It is not handled in a way that user did not provide any code to handle the exception.

marandaneto: ok, `SentryExceptionResolver` does the same so its ok, questioned due to https://github.com/getsentry/sentry-dart/issues/456 its somehow similar.

sentry-spring/src/main/java/io/sentry/spring/webflux/SentryWebExceptionHandler.java
```diff
@@ -0,0 +1,47 @@
...
+      mechanism.setType("SentryWebExceptionHandler");
+      mechanism.setHandled(false);
+      final Throwable throwable =
+          new ExceptionMechanismException(mechanism, ex, Thread.currentThread());
```
marandaneto, line 39: more like a question, is `SentryWebExceptionHandler` always executed in the calling thread? otherwise marking the current thread will be wrong due to the usage of `Thread.currentThread()`.

Asking because of https://github.com/getsentry/sentry-java/issues/1524#issuecomment-859370797
in coroutines, the handler can get called in different threads

sentry-spring/src/main/java/io/sentry/spring/webflux/TransactionNameProvider.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Resolves transaction name using {@link ServerWebExchange#getRequest()} ()} and templated route
+ * that handled the request. To return correct transaction name, it must be used after request is
+ * processed by {@link
+ * org.springframework.web.reactive.result.method.RequestMappingInfoHandlerMapping} where {@link
+ * HandlerMapping#BEST_MATCHING_PATTERN_ATTRIBUTE} is set.
+ */
+@ApiStatus.Internal
+public final class TransactionNameProvider {
```
bruno-garcia, line 18: Couldn't this just be a static thing?
Code using it doesn't seem to be mocking it

sentry-spring/src/main/java/io/sentry/spring/webflux/SentryWebExceptionHandler.java
maciejwalkowiak: I don't think there is a guarantee that its executed on the same thread. `thread` is a required property - what should be set there other than `currentThread` in this case?

sentry-spring/src/main/java/io/sentry/spring/webflux/TransactionNameProvider.java
maciejwalkowiak: Yep, can be static. Changed.

sentry-spring/src/main/java/io/sentry/spring/webflux/SentryWebExceptionHandler.java
marandaneto: the side effect is that the `exception.setThreadId` will contain the id of the current thread, which could display the wrong stack trace in the sentry issues page if `SentryOptions#attachThreads` is enabled (all threads reported).
I'm wondering if we should make `ExceptionMechanismException#getThread` nullable, and add a new ctor to `ExceptionMechanismException` and just don't set the `threadId` if not available.

@bruno-garcia ?

marandaneto:
add new package entry to https://github.com/getsentry/sentry-java/blob/main/.craft.yml
new packages also require sentry-release-registry eg https://github.com/getsentry/sentry-release-registry/pull/52

bruno-garcia:
> add new package entry to https://github.com/getsentry/sentry-java/blob/main/.craft.yml
> new packages also require sentry-release-registry eg [getsentry/sentry-release-registry#52](https://github.com/getsentry/sentry-release-registry/pull/52)

Instructions for the release registry are on the repo README (it's just a `make sync-all-links` after you add 1 JSON file and the links are created for you

maciejwalkowiak:
@marandaneto only the sample is a new package and as far as I can see we don't have other samples configured in `craft`.

marandaneto:
> @marandaneto only the sample is a new package and as far as I can see we don't have other samples configured in `craft`.

oh, true, I thought it was a new package but indeed, its the sample, all good, thanks!

sentry-spring/src/main/java/io/sentry/spring/webflux/SentryWebExceptionHandler.java
marandaneto: ping @bruno-garcia

marandaneto:
I believe this closes https://github.com/getsentry/sentry-java/pull/999

CodeQL is unhappy

CHANGELOG.md
```diff
@@ -5,14 +5,16 @@
 * Fix: Handling missing Spring Security on classpath on Java 8 (#1552)
 * Feat: Support transaction waiting for children to finish. (#1535) 
 * Feat: Capture logged marker in log4j2 and logback appenders (#1551)
+* Feat: Spring Webflux integration (#1529) 
 * Fix: Add data field to SentrySpan (#1555)
 * Fix: Clock drift issue when calling DateUtils#getDateTimeWithMillisPrecision (#1557)
 * Feat: Set mechanism type in SentryExceptionResolver (#1556)
+* Feat: Spring Webflux integration (#1529)
```
marandaneto, line 12:
```suggestion
```

dup

sentry-spring/src/main/java/io/sentry/spring/webflux/SentryWebExceptionHandler.java
bruno-garcia: If we include Thread stack traces + Exception stack trace, the UI will display the Exception stack trace on the threads list (and ignore the Thread one) by matching its ID. So if we don't send an ID this could be broken.

Since this is asynchronous, the "thread" is less relevant here in terms of "This thread has done all the work up to this point". But it's OK to be "this thread did the work at the time of capture" which is what this is represented here.

So I think keeping as `Thread.currentThread()` is fine here and arguably a better approach than making it nullable given how Sentry works.

marandaneto: yep I am just afraid that the thread id may be wrong, and we'd be hiding the stack trace of that thread since it's going to be showing the exception stack trace, I don't know if it's a big problem though.

marandaneto (approved):
LGTM

marandaneto:
@maciejwalkowiak please open up the docs for it

marandaneto:
> @maciejwalkowiak please open up the docs for it

https://github.com/getsentry/sentry-docs/issues/3816

pkgonan:
Hi. when is the release date of this PR?

maciejwalkowiak:
@pkgonan we don't have a release schedule, but we release very often: https://github.com/getsentry/sentry-java/releases

I recommend watching this repo for releases and trying out next beta.

pkgonan:
@maciejwalkowiak Thank you.
Is this feature support to spring boot webflux with kotlin coroutine ? (Mean : suspend fun and Flow operator)

sentry-spring/src/main/java/io/sentry/spring/webflux/SentryWebExceptionHandler.java
```diff
@@ -0,0 +1,45 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.springframework.core.annotation.Order;
+import org.springframework.web.server.ResponseStatusException;
+import org.springframework.web.server.ServerWebExchange;
+import org.springframework.web.server.WebExceptionHandler;
+import reactor.core.publisher.Mono;
+
+/** Handles unhandled exceptions in Spring WebFlux integration. */
+@Order(
+    -2) // the DefaultErrorWebExceptionHandler provided by Spring Boot for error handling is ordered
```
sergey-volkov-lm, line 19: Maybe use existing property for order tuning? `sentry.exception-resolver-order`. But, for some reason, it has default value 1 for MVC handler.

PR #1531, opened 2021-06-14
kotlin-extensions to readme

bruno-garcia:
Do we have a min API for this? Not on the gradle file

#skip-changelog


marandaneto (approved):

marandaneto:
> Do we have a min API for this? Not on the gradle file

no, it should work on every version as we target `jvmTarget=8`, Kotlin will make it compatible with lower versions.

PR #1535, opened 2021-06-16
Support transaction waiting for children to finish.

### LINKED ISSUE #1523: Support transaction finishing automatically with 'idle timeout'

maciejwalkowiak:
## :scroll: Description

Adds an option to create transaction that finishes only when it is itself finished and all its children are finished.

Fixes #1523

Unit tests.

## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next steps

Reference docs.


maciejwalkowiak:
@marandaneto as we do not want to make it a part of public API, but rather use it for Android integration, what API method on the `IHub` would you expect?

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -150,8 +188,9 @@ public void finish() {
 
   @Override
   public void finish(@Nullable SpanStatus status) {
-    if (!root.isFinished()) {
-      root.finish(status);
+    this.finishStatus = status;
+    if (!root.isFinished() && (!waitForChildren || hasAllChildrenFinished())) {
```
marandaneto, line 192: `root.isFinished()` is an atomic operation, so I'd do the other 2 checks with an inner `if`

marandaneto:
@maciejwalkowiak we probably still need a `timeout` feature, for spans that never finish or such other corner cases, iOS right now does not do it either, @philipphofmann @brustolin do you recall if we've decided to add on iOS too?

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -150,8 +188,9 @@ public void finish() {
 
   @Override
   public void finish(@Nullable SpanStatus status) {
-    if (!root.isFinished()) {
-      root.finish(status);
+    this.finishStatus = status;
```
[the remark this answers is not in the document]
maciejwalkowiak: we must save the intentional finish status so that when all children are finished, we finish the transaction with proper status.

marandaneto:
> @marandaneto as we do not want to make it a part of public API, but rather use it for Android integration, what API method on the `IHub` would you expect?

to be honest, I think this would be an issue for Java Desktop Apps too, don't you? we can start exposing it to Hub with `@ApiStatus.Internal`, I'm not sure how we could do it otherwise, even if using `SentryAndroid`, looks like we still need something available in the Hub

maciejwalkowiak:
Sure for desktop too, but you I thought you have exact use case in mind. Then I'll just add another variant of `startTransaction`.

philipphofmann:
> @maciejwalkowiak we probably still need a `timeout` feature, for spans that never finish or such other corner cases, iOS right now does not do it either, @philipphofmann @brustolin do you recall if we've decided to add on iOS too?

We didn't add it yet on iOS. I created an issue to discuss this today at grooming https://github.com/getsentry/sentry-cocoa/issues/1173.

sentry/src/main/java/io/sentry/SentryTracer.java
marandaneto: yep, you're right, no difference, my bad

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -97,7 +124,18 @@ private ISpan createChild(
...
+            root.getTraceId(),
+            parentSpanId,
+            this,
+            operation,
+            this.hub,
+            timestamp,
+            __ -> {
+              if (waitForChildren && hasAllChildrenFinished()) {
+                finish(finishStatus);
+              }
```
marandaneto, line 137: should you just call `finish(x)` directly? the `finish(x)` method does pretty much the same 2 checks

marandaneto:
I have the impression that there's a bug.

```
val tr = startTranasction(x)
val span tr.startChild(x)
span.finish()
// at this point, tr has a single child and its finished, tr is gonna be finished automatically, it should not.
// `tr` should only auto finish once all children are finished once somebody called `tr.finish(x)`
```

can you confirm my use case? thanks

sentry/src/test/java/io/sentry/SentryTracerTest.kt
```diff
@@ -327,4 +331,33 @@ class SentryTracerTest {
...
+        transaction.finish()
+        verify(fixture.hub, never()).captureTransaction(any())
+    }
+
+    @Test
+    fun `when waiting for children, finishing transaction calls hub if all children are finished`() {
+        val transaction = fixture.getSut(waitForChildren = true)
+        val child = transaction.startChild("op")
+        child.finish()
+        transaction.finish()
```
marandaneto, line 348: if you remove this line, the test is still green I guess, but `transaction.finish()` should be called too

marandaneto:
@brustolin and @philipphofmann please review since this is done on iOS too

sentry/src/main/java/io/sentry/SentryTracer.java
brustolin, line 192: I'm missing a finish intention flag here.
Let's say the user did not call finish, but finishes all children, this condition will evaluate to true and finish the root span.

brustolin: IOS solved this by adding a ```isWaitingForChildren``` variable, and a private method ```canBeFinished```.
The finish method only sets isWaitingForChildren to true, saves the status and call canBeFinished that does the rest. 
The span listener should call ```canBeFinished```, otherwise it would set the isWaitingForChildren to true.

sentry/src/main/java/io/sentry/SentryTracer.java
maciejwalkowiak: Refactored. Take another look please.

sentry/src/test/java/io/sentry/SentryTracerTest.kt
maciejwalkowiak: Thanks, same has been caught by @brustolin. Test added and code fixed.

sentry/src/main/java/io/sentry/SentryTracer.java
maciejwalkowiak: I think its fixed now: https://github.com/getsentry/sentry-java/pull/1535/files#diff-8bb325ca6fc9666c24826415b357d8e76a34cb455edc5709dd962bdcd34c6f16R136

brustolin: Nice!
You just fixed while I was commenting 😂

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -97,7 +124,19 @@ private ISpan createChild(
...
+            parentSpanId,
+            this,
+            operation,
+            this.hub,
+            timestamp,
+            __ -> {
+              final FinishStatus finishStatus = this.finishStatus;
+              if (finishStatus.isFinishing) {
+                finish(finishStatus.spanStatus);
+              }
```
brustolin, line 138: L: I would just suggest extracting this to a function, let's say in the future we have another condition to finish, or other places this verification should be done.
It would help with maintenance.

Maybe SentryTracer itself could implement SpanListener.

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -289,4 +341,22 @@ public void setRequest(final @Nullable Request request) {
   Span getRoot() {
     return root;
   }
+
+  private static class FinishStatus {
```
marandaneto, line 345:
```suggestion
  private static final class FinishStatus {
```

philipphofmann (commented):
I currently only have one comment to add.

sentry/src/main/java/io/sentry/SentryTracer.java
marandaneto: its fixed, thanks!

sentry/src/main/java/io/sentry/SpanListener.java
```diff
@@ -0,0 +1,12 @@
...
+
+import org.jetbrains.annotations.NotNull;
+
+interface SpanListener {
+  /**
+   * Called when observed span finishes.
+   *
+   * @param span the span that has finished.
+   */
+  void onSpanFinished(final @NotNull Span span);
```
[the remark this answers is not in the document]
marandaneto: `Span` ctor takes a `SpanListener`

sentry/src/main/java/io/sentry/SentryTracer.java
marandaneto: not every span has the `waitForChildren` feat, so `Span` getting the listener as ctor makes sense to me, but its an implementation detail anyway

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -289,4 +341,22 @@ public void setRequest(final @Nullable Request request) {
   Span getRoot() {
     return root;
   }
+
+  private static final class FinishStatus {
+    private final boolean isFinishing;
+    private final @Nullable SpanStatus spanStatus;
+
+    static @NotNull FinishStatus notFinished() {
+      return new FinishStatus(false, null);
```
bruno-garcia, line 350: Looking at the usage of this method we'll need a instance of this for each tracer. On Spring we'll have on per request so could declare this as a field and just returned that single instance?

sentry/src/main/java/io/sentry/SpanListener.java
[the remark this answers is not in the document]
bruno-garcia: `span => { }`

sentry/src/main/java/io/sentry/SentryTracer.java
maciejwalkowiak: Done! Thanks good point

maciejwalkowiak:
@marandaneto @bruno-garcia is there anything more do you think we should change here?

CHANGELOG.md
```diff
@@ -5,7 +5,8 @@
 ## 5.1.0-beta.1
 
 * Feat: Measure app start time (#1487)
-* Feat: Automatic breadcrumbs logging for fragment lifecycle (#1522) 
+* Feat: Automatic breadcrumbs logging for fragment lifecycle (#1522)
+* Feat: Support transaction waiting for children to finish. (#1535)
```
marandaneto, line 9: goes to `Unreleased`

marandaneto:
> @marandaneto @bruno-garcia is there anything more do you think we should change here?

probably https://github.com/getsentry/sentry-java/pull/1535#issuecomment-862137902 and https://github.com/getsentry/sentry-java/pull/1528#discussion_r652642233

I'll discuss both this afternoon with the team, maybe the later could be already done

bruno-garcia:
> @maciejwalkowiak we probably still need a `timeout` feature, for spans that never finish or such other corner cases, iOS right now does not do it either, @philipphofmann @brustolin do you recall if we've decided to add on iOS too?

We could merge as-is and add the timeout later? It would unblock Android.

> @maciejwalkowiak I am branched from your branch gh-1523, can we make an easy overload for such option? that allows name, op and waitForChildren, with or without startTimestamp

Could the overload be added to the other branch instead and we just merge this?

CHANGELOG.md
bruno-garcia: fixed

bruno-garcia (approved):

CHANGELOG.md
```diff
@@ -2,12 +2,13 @@
 
 ## Unreleased
 
+* Feat: Automatic breadcrumbs logging for fragment lifecycle (#1522)
```
marandaneto, line 5: this is 5.1.0-beta.1

bruno-garcia: It was marked on ur review so I moved down. 

image


It's quick tp fix these things btw ^

I'll fix it now

CHANGELOG.md
```diff
@@ -2,12 +2,13 @@
...
+* Feat: Support transaction waiting for children to finish. (#1535) 
 * Feat: Capture logged marker in log4j2 and logback appenders (#1551)
 
 ## 5.1.0-beta.1
 
 * Feat: Measure app start time (#1487)
-* Feat: Automatic breadcrumbs logging for fragment lifecycle (#1522) 
+* Feat: Automatic breadcrumbs logging for fragment lifecycle (#1522)
 
 ## 5.0.1
```
bruno-garcia, line 13: @marandaneto this right?

marandaneto (approved):

PR #1545, opened 2021-06-18
feat(craft): Update config to support automatic releases

iker-barriocanal:
Updates Craft config. With this new configuration, Craft should be able to release Java & Android SDKs.

Note: draft until Craft is properly finished, see https://github.com/getsentry/craft/pull/258.


.craft.yml
```diff
@@ -20,4 +18,13 @@ targets:
       maven:io.sentry:sentry-android-timber:
       maven:io.sentry:sentry-android-okhttp:
       maven:io.sentry:sentry-kotlin-extensions:
-      maven:io:sentry:sentry-android-fragment:
+  - name: maven
```
bruno-garcia, line 21:
```suggestion
  - name: maven
```

.craft.yml
```diff
@@ -21,3 +19,13 @@ targets:
       maven:io.sentry:sentry-android-okhttp:
       maven:io.sentry:sentry-kotlin-extensions:
       maven:io:sentry:sentry-android-fragment:
+  - name: maven
+    gradleCliPath: ./gradlew
+    mavenCliPath: scripts/mvnw
```
marandaneto, line 24: is this accurate after moving the /mvn folder?

.craft.yml
```diff
@@ -1,7 +1,5 @@
 minVersion: 0.23.1
```
marandaneto, line 1: this probably has to changed after released right

.craft.yml
iker-barriocanal: Yes, `.mvn/` should be in the same directory as `.craft.yml`.

.craft.yml
iker-barriocanal: Yes, but the feature hasn't been released yet on Craft. Once it's released, I'll get this branch ready (including updating the minimum Craft version).

PR #1588, opened 2021-07-07
fix(release): Manage git state after post-release

iker-barriocanal:
From [Craft docs](https://github.com/getsentry/craft#post-release-script-conventions):

> The script is responsible for any and all `git` state management as `craft` will simply exit after running this script as the final step. This means the script is responsible for committing and pushing any changes that it may have made.

Git state isn't managed in the post-release script, meaning changes won't be pushed; this PR manages it.

_#skip-changelog_


marandaneto (approved):

marandaneto:
@iker-barriocanal there's a conflict, besides that, LGTM

PR #1592, opened 2021-07-12
feat(craft): Add `symbol-collector` target

iker-barriocanal:
This PR adds the `symbol-collector` target to support https://github.com/getsentry/craft/pull/266. It also changes the target order, so that:

- `symbol-collector` runs before `maven`, since the former is reentrant and the latter isn't.
- `github` and `registry` targets go after `symbol-collector` and `maven`, since the first two should only be run once the actual release has been made.

_#skip-changelog_

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -1,6 +1,21 @@
 minVersion: 0.25.0
 changelogPolicy: auto
 targets:
+  - name: symbol-collector
+    includeNames: /libsentry(-android)?\.so/
+    batchType: android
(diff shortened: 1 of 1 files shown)


bruno-garcia (approved):
Nice!

marandaneto:
@iker-barriocanal requires https://github.com/getsentry/craft to be released, and updating https://github.com/getsentry/sentry-java/blob/main/.craft.yml#L1 if necessary, right now it requires building craft manually

PR #1595, opened 2021-07-13
Add request body extraction for Spring MVC integration

### LINKED ISSUE #1334: Support HTTP Request Body extraction
This feature [exists in other SDKs and is configured like this](https://docs.sentry.io/platforms/dotnet/guides/aspnetcore/configuration/options/#max-request-body-size):

An option named `maxRequestBodySize.`

The values are:

* never: request bodies are never sent (default)
* small: only small request bodies will be captured where the cutoff for small depends on the SDK (typically 4KB)
* medium: medium and small requests will be captured (typically 10KB)
* always: the SDK will always capture the request body for as long as Sentry can make sense of it

If opting-in means the request stream need to be buffered, that should be transparent. Docs can mention this caveat but it's fair to assume that if you're looking to get the request body included in error
s, you're not running a server that's about to tip over and you can afford having the payload buffered.

- [ ] [Document the feature spec in the develop docs](Add to the spec as "Expected Features": https://develop.sentry.dev/sdk/features/)

The code can depend on a simple [HTTP abstraction to get the data.](https://github.com/getsentry/sentry-dotnet/blob/2337261333d5743996d6d96ca3aa289d34a9d3ec/src/Sentry/Extensibility/RequestBodyExtractionDispatcher.cs). That was needed in .NET to support both ASP.NET and ASP.NET Core.
Also: [1](https://github.com/getsentry/sentry-dotnet/blob/7166cc23f2f384b4e8dddf1fc3450dc7edf88a12/src/Sentry/Extensibility/BaseRequestPayloadExtractor.cs#L11), [2](https://github.com/getsentry/sentry-dotnet/search?q=IHttpRequest)

This feature was requested in #331 particularly for `application/json`.

maciejwalkowiak:
## :scroll: Description

Add request body extraction for Spring MVC integration

Fixes #1334 

Unit & integration tests, sample projects updated.

## :pencil: Checklist

- [x] No breaking changes


sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryAutoConfiguration.java
```diff
@@ -198,9 +195,16 @@
           final @NotNull IHub hub) {
         FilterRegistrationBean<SentryTracingFilter> filter =
             new FilterRegistrationBean<>(new SentryTracingFilter(hub));
-        filter.setOrder(Ordered.HIGHEST_PRECEDENCE);
+        filter.setOrder(Ordered.HIGHEST_PRECEDENCE + 1); // must run after SentrySpringFilter
```
bruno-garcia, line 198: nit: Could be worth then declaring these as field and name them so the fact one depends (must follow) on the other can never go out of sync:

```suggestion
        private static int SENTRY_SPRING_FILTER_PRESEDENCE = Ordered.HIGHEST_PRECEDENCE;
        // must run after SentrySpringFilter
        private static int SENTRY_TRACING_FILTER_PRESEDENCE + SENTRY_SPRING_FILTER_PRESEDENCE + 1;
        filter.setOrder(SENTRY_TRACING_FILTER_PRESEDENCE);
```

sentry-spring/src/main/java/io/sentry/spring/RequestPayloadExtractor.java
```diff
@@ -0,0 +1,30 @@
...
+
+final class RequestPayloadExtractor {
+
+  @Nullable
+  String extract(final @NotNull HttpServletRequest request, final @NotNull SentryOptions options) {
+    // request body can be read only once from the stream
+    // original request can be replaced with CachedBodyHttpServletRequest in SentrySpringFilter
+    if (request instanceof CachedBodyHttpServletRequest) {
+      try {
+        final byte[] body = StreamUtils.copyToByteArray(request.getInputStream());
```
bruno-garcia, line 20: To confirm: we're gettinga brand new input stream here so not moving the offset of the request stream, right?

In ASP.NET classic, IIRC, I had to keep track of the offset within the stream (which was cached, like here) so I could read from beginning to end and then set the offset back where it was.

sentry-spring/src/main/java/io/sentry/spring/SentrySpringFilter.java
```diff
@@ -0,0 +1,100 @@
...
+  protected void doFilterInternal(
+      final @NotNull HttpServletRequest servletRequest,
+      final @NotNull HttpServletResponse response,
+      final @NotNull FilterChain filterChain)
+      throws ServletException, IOException {
+    if (hub.isEnabled()) {
+      final HttpServletRequest request = resolveHttpServletRequest(servletRequest);
+      try {
+        hub.pushScope();
+        hub.addBreadcrumb(Breadcrumb.http(request.getRequestURI(), request.getMethod()));
```
bruno-garcia, line 50: Might need to confirm if this looks like it's an incoming request. I believe this "format" of crumb might have been designed to represent an outgoing HTTP request. 
If it won't get confusing with crumbs representing requests done from this process out to the another for example through rest-template, then we're good.

sentry-spring/src/main/java/io/sentry/spring/SentrySpringFilter.java
```diff
@@ -0,0 +1,100 @@
...
+        hub.configureScope(
+            scope -> {
+              scope.setRequest(requestResolver.resolveSentryRequest(request));
```
bruno-garcia, line 53: Here we're eagerly reading the request bytes into the scope which won't really be used unless we have an event.

Another alternative here is to add an event processor that holds the request (we have on the line below?) and we then only add the request information if needed and spare all the allocations.

Buffering the request is required anyway, since we need to be able to rewind the stream in case of an error.

sentry-spring/src/main/java/io/sentry/spring/SentrySpringFilter.java
```diff
@@ -0,0 +1,100 @@
...
+  @Override
+  protected void doFilterInternal(
+      final @NotNull HttpServletRequest servletRequest,
+      final @NotNull HttpServletResponse response,
+      final @NotNull FilterChain filterChain)
+      throws ServletException, IOException {
+    if (hub.isEnabled()) {
+      final HttpServletRequest request = resolveHttpServletRequest(servletRequest);
+      try {
+        hub.pushScope();
```
bruno-garcia, line 49: We're popping inside of the `finally` block so technically this `pushScope` should be before try:


```suggestion
      hub.pushScope();
      try {
```

sentry-spring/src/main/java/io/sentry/spring/SentrySpringFilter.java
bruno-garcia: One issue I see is that we're passing `request` here regardless of the outcome of `resolveHttpServletRequest`. Meaning, it could have returned the normal rquest without buffering it (so regardless of the result of `qualifiesForCaching`.

I think we need to flow the result of `qualifiesForCaching` to guard calling `setRequest` to add this altogether.

bruno-garcia (commented):
Looks good. I had a couple comments we could discuss on a call if best.
It's rather late and I'm tired so I might be missing something

sentry-spring/src/main/java/io/sentry/spring/RequestPayloadExtractor.java
maciejwalkowiak: `getInputStream` returns a new input stream so there is no need to track offset.

sentry-spring/src/main/java/io/sentry/spring/SentrySpringFilter.java
maciejwalkowiak: That's how it's been since 3.0 release. Where can I verify if this is the right format?

sentry-spring/src/main/java/io/sentry/spring/SentrySpringFilter.java
maciejwalkowiak: 👍

sentry-spring/src/main/java/io/sentry/spring/SentrySpringFilter.java
maciejwalkowiak: Updated. Now the request body is not set on the scope at all, and added in event processor.

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryAutoConfiguration.java
maciejwalkowiak: 👍

marandaneto:
deferring the review to @bruno-garcia as there's already the 1st pass.
The sentry module is barely untouched so all good for Android.
It'd be nice to bring this into the OkHttp integration too, another PR of course.

maciejwalkowiak:
@marandaneto how this would work for OkHttp or other http clients? On the server side, request body is added to events that are triggered within the scope of an HTTP request. I am not sure how would this translate to HTTP clients, as for the clients we just create a breadcrumb and a span.

sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryAutoConfiguration.java
```diff
@@ -198,9 +198,16 @@
           final @NotNull IHub hub) {
         FilterRegistrationBean<SentryTracingFilter> filter =
             new FilterRegistrationBean<>(new SentryTracingFilter(hub));
-        filter.setOrder(Ordered.HIGHEST_PRECEDENCE);
+        filter.setOrder(SENTRY_SPRING_FILTER_PRECEDENCE + 1); // must run after SentrySpringFilter
```
bruno-garcia, line 201: Much simpler than what I had in mind 👍

sentry-spring/src/main/java/io/sentry/spring/SentrySpringFilter.java
bruno-garcia: If we have an incoming request into a spring process that makes an outgoing request. Would we be able to differentiate the crumbs? Probably at least through the URL.
Doesn't block this PR

bruno-garcia (approved):
NIce! thanks @maciejwalkowiak

marandaneto:
> @marandaneto how this would work for OkHttp or other http clients? On the server side, request body is added to events that are triggered within the scope of an HTTP request. I am not sure how would this translate to HTTP clients, as for the clients we just create a breadcrumb and a span.

oh yeah indeed, the OkHttp integration still doesn't raise events if the request failed somehow, so yeah it does not make sense right now.
See the Dart integration https://github.com/getsentry/sentry-dart/blob/main/dart/lib/src/http_client/failed_request_client.dart#L187-L207
it captures failed requests if enabled, and also respects the `MaxRequestBodySize` if enabled too, the idea would be to bring this feature into the Java SDK too, but we're waiting for this feature to mature since we've just released it and gathering feedback about the usage.

Issue #1607, opened 2021-07-19
Why Min Sdk 21 is requrired for sentry-android-fragment and sentry-android-okhttp?


Alireza-Farahani:
Both mentioned artifacts have configured api 21 as minimum. `sentry-android-okhttp` depends on OkHttp 4.9 which itself requires minsdk 21, but the functionality doesn't require api 21. `sentry-android-fragment` doesn't require api 21 at all. I was able to use both artifacts in my app, leveraging `tools:overrideLibrary` in `AndroidManifest.xml` file.

Is there a specific reason behind choosing 21 as min sdk? Should I expect any runtime problems?

Anyway, I think related docs in "integrations" segment should explicitly mention about the min sdk requirement.

marandaneto:
@Alireza-Farahani thanks for raising this.

for OkHttp: https://github.com/square/okhttp#requirements

> OkHttp works on Android 5.0+ (API level 21+) and Java 8+.

Might be the case that it works for older versions, but as its not tested by us, we prefer to stick to the official OkHttp docs.

I'll check the fragments compatibility.

indeed docs could better state that, although we've it written down here in the README file https://github.com/getsentry/sentry-java#sentry-sdk-for-java-and-android

marandaneto:
apparently androidx.fragment is min API 14, See https://github.com/getsentry/sentry-java/pull/1608
Let's see if CI gets happy as there are code checks via AGP and code compatibility

PR #1608, opened 2021-07-19
Fix: set min sdk version of sentry-android-fragment to API 14

marandaneto:
## :scroll: Description
https://mvnrepository.com/artifact/androidx.fragment/fragment and https://androidx.tech/artifacts/fragment/fragment/ state min API 14

Fix https://github.com/getsentry/sentry-java/issues/1607

CI

## :pencil: Checklist

- [X] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Fix: set min sdk version of sentry-android-fragment to API 14
+
 ## 5.1.0-beta.5
 
 * Make SentryAppender non-final for Log4j2 and Logback (#1603) 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-* Fix: set min sdk version of sentry-android-fragment to API 14
+* Fix: set min sdk version of sentry-android-fragment to API 14 (#1608)
 
 ## 5.1.0-beta.5
 
--- a/buildSrc/src/main/java/Config.kt
+++ b/buildSrc/src/main/java/Config.kt
@@ -33,7 +33,6 @@ object Config {
 
         val minSdkVersion = 14
         val minSdkVersionOkHttp = 21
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Fix: set min sdk version of sentry-android-fragment to API 14
```
marandaneto, line 5:
```suggestion
* Fix: set min sdk version of sentry-android-fragment to API 14 (#1608)
```

buildSrc/src/main/java/Config.kt
```diff
@@ -33,7 +33,6 @@ object Config {
 
         val minSdkVersion = 14
         val minSdkVersionOkHttp = 21
-        val minSdkVersionFragment = 21
```
marandaneto, line 36: @wzieba do you remember why you've set 21 instead of 14?

buildSrc/src/main/java/Config.kt
```diff
@@ -86,7 +85,7 @@ object Config {
 
         val coroutinesCore = "org.jetbrains.kotlinx:kotlinx-coroutines-core:1.4.3"
 
-        val fragment = "androidx.fragment:fragment-ktx:1.3.4"
+        val fragment = "androidx.fragment:fragment-ktx:1.3.5"
```
marandaneto, line 88: theres a critical bugfix from 1.3.4

buildSrc/src/main/java/Config.kt
wzieba: I don't remember and I think it was rather not a deliberative decision - I think I've just copied the value from `minSdkVersionOkHttp`. Sorry!

marandaneto: not a problem at all, we're in a beta state so its the right moment to get it fixed :)
thanks again for the contribution @wzieba

maciejwalkowiak (approved):
LGTM

PR #1609, opened 2021-07-19
Feat: Slow/Frozen frames metrics

marandaneto:
## :scroll: Description
Feat: Slow/Frozen frames metrics

Closes https://github.com/getsentry/sentry-java/pull/1466

## :pencil: Checklist

- [X] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased
 
 * Fix: set min sdk version of sentry-android-fragment to API 14 (#1608)
-* Feat: Slow/Frozen frames metrics (#)
+* Feat: Slow/Frozen frames metrics (#1609)
 
 ## 5.1.0-beta.5
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 * Fix: set min sdk version of sentry-android-fragment to API 14 (#1608)
+* Feat: Slow/Frozen frames metrics (#)
 
 ## 5.1.0-beta.5
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,11 @@
 
 ## Unreleased
 
+* Feat: Slow/Frozen frames metrics (#1609)
+
 ## 5.1.0-beta.6
 
 * Fix: set min sdk version of sentry-android-fragment to API 14 (#1608)
-* Feat: Slow/Frozen frames metrics (#1609)
 * Fix: Ser/Deser of the UserFeedback from cached envelope (#1611)
 * Feat: Add request body extraction for Spring MVC integration (#1595)
 
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesState.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesState.java
@@ -81,6 +81,10 @@ public final class ActivityFramesState {
     return activityMeasurements.get(sentryId);
   }
 
+  void removeMetrics(final @NotNull SentryId sentryId) {
+    activityMeasurements.remove(sentryId);
+  }
+
   void close() {
     frameMetricsAggregator.stop();
   }
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesState.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesState.java
@@ -1,0 +1,87 @@
+package io.sentry.android.core;
+
+import android.app.Activity;
+import android.util.SparseIntArray;
+import androidx.core.app.FrameMetricsAggregator;
+import io.sentry.protocol.MeasurementValue;
+import io.sentry.protocol.SentryId;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.WeakHashMap;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+@ApiStatus.Internal
+public final class ActivityFramesState {
+
+  private static final @NotNull ActivityFramesState instance = new ActivityFramesState();
+
+  private final @NotNull FrameMetricsAggregator frameMetricsAggregator =
+      new FrameMetricsAggregator();
+
+  private ActivityFramesState() {}
+
+  static @NotNull ActivityFramesState getInstance() {
+    return instance;
+  }
+
+  private final @NotNull Map<SentryId, Map<String, @NotNull MeasurementValue>>
+      activityMeasurements = Collections.synchronizedMap(new WeakHashMap<>());
+
+  void addActivity(final @NotNull Activity activity) {
+    frameMetricsAggregator.add(activity);
+  }
+
+  void setMetrics(final @NotNull Activity activity, final @NotNull SentryId sentryId) {
+    int totalFrames = 0;
+    int slowFrames = 0;
+    int frozenFrames = 0;
+
+    final SparseIntArray[] framesRates = frameMetricsAggregator.remove(activity);
+
+    if (framesRates != null) {
+      final SparseIntArray totalIndexArray = framesRates[FrameMetricsAggregator.TOTAL_INDEX];
+      if (totalIndexArray != null) {
+        for (int i = 0; i < totalIndexArray.size(); i++) {
+          int frameTime = totalIndexArray.keyAt(i);
+          int numFrames = totalIndexArray.valueAt(i);
+          totalFrames += numFrames;
+          // hard coded values, its also in the official android docs and frame metrics API
+          if (frameTime > 700) {
+            // frozen frames, threshold is 700ms
+            frozenFrames += numFrames;
+          }
+          if (frameTime > 16) {
+            // slow frames, above 16ms, 60 frames/second
+            slowFrames += numFrames;
+          }
+        }
+      }
+    }
+
+    if (totalFrames == 0 && slowFrames == 0 && frozenFrames == 0) {
+      return;
+    }
+
+    final MeasurementValue tfValues = new MeasurementValue(totalFrames);
+    final MeasurementValue sfValues = new MeasurementValue(slowFrames);
+    final MeasurementValue ffValues = new MeasurementValue(frozenFrames);
+    final Map<String, @NotNull MeasurementValue> measurements = new HashMap<>();
+    measurements.put("frames_total", tfValues);
+    measurements.put("frames_slow", sfValues);
+    measurements.put("frames_frozen", ffValues);
+
+    activityMeasurements.put(sentryId, measurements);
+  }
+
+  @Nullable
+  Map<String, @NotNull MeasurementValue> getMetrics(final @NotNull SentryId sentryId) {
+    return activityMeasurements.get(sentryId);
+  }
+
+  void close() {
+    frameMetricsAggregator.stop();
+  }
+}
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesTracker.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesTracker.java
@@ -50,8 +50,7 @@ final class ActivityFramesTracker {
           if (frameTime > 700) {
             // frozen frames, threshold is 700ms
             frozenFrames += numFrames;
-          }
-          if (frameTime > 16 && frameTime <= 700) {
+          } else if (frameTime > 16) {
             // slow frames, above 16ms, 60 frames/second
             slowFrames += numFrames;
           }
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesTracker.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesTracker.java
@@ -74,12 +74,11 @@ final class ActivityFramesTracker {
   }
 
   @Nullable
-  Map<String, @NotNull MeasurementValue> getMetrics(final @NotNull SentryId sentryId) {
-    return activityMeasurements.get(sentryId);
-  }
-
-  void removeMetrics(final @NotNull SentryId sentryId) {
+  Map<String, @NotNull MeasurementValue> takeMetrics(final @NotNull SentryId sentryId) {
+    final Map<String, @NotNull MeasurementValue> stringMeasurementValueMap =
+        activityMeasurements.get(sentryId);
     activityMeasurements.remove(sentryId);
+    return stringMeasurementValueMap;
   }
 
   void close() {
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesTracker.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesTracker.java
@@ -5,30 +5,29 @@ import android.util.SparseIntArray;
 import androidx.core.app.FrameMetricsAggregator;
 import io.sentry.protocol.MeasurementValue;
 import io.sentry.protocol.SentryId;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.WeakHashMap;
+import java.util.concurrent.ConcurrentHashMap;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 @ApiStatus.Internal
-public final class ActivityFramesState {
+public final class ActivityFramesTracker {
 
-  private static final @NotNull ActivityFramesState instance = new ActivityFramesState();
+  private static final @NotNull ActivityFramesTracker instance = new ActivityFramesTracker();
 
   private final @NotNull FrameMetricsAggregator frameMetricsAggregator =
       new FrameMetricsAggregator();
 
-  private ActivityFramesState() {}
+  private ActivityFramesTracker() {}
 
-  static @NotNull ActivityFramesState getInstance() {
+  static @NotNull ActivityFramesTracker getInstance() {
     return instance;
   }
 
   private final @NotNull Map<SentryId, Map<String, @NotNull MeasurementValue>>
-      activityMeasurements = Collections.synchronizedMap(new WeakHashMap<>());
+      activityMeasurements = new ConcurrentHashMap<>();
 
   void addActivity(final @NotNull Activity activity) {
     frameMetricsAggregator.add(activity);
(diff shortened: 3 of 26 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Unreleased
 
 * Fix: set min sdk version of sentry-android-fragment to API 14 (#1608)
+* Feat: Slow/Frozen frames metrics (#)
```
marandaneto, line 6:
```suggestion
* Feat: Slow/Frozen frames metrics (#1609)
```

sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesState.java
```diff
@@ -0,0 +1,91 @@
...
+      new FrameMetricsAggregator();
+
+  private ActivityFramesState() {}
+
+  static @NotNull ActivityFramesState getInstance() {
+    return instance;
+  }
+
+  private final @NotNull Map<SentryId, Map<String, @NotNull MeasurementValue>>
+      activityMeasurements = Collections.synchronizedMap(new WeakHashMap<>());
```
philipphofmann, line 31: `m`: Why do we need a `WeakHashMap` here? Shouldn't a normal HashMap also be fine?

sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesState.java
```diff
@@ -0,0 +1,91 @@
...
+  Map<String, @NotNull MeasurementValue> getMetrics(final @NotNull SentryId sentryId) {
+    return activityMeasurements.get(sentryId);
+  }
+
+  void removeMetrics(final @NotNull SentryId sentryId) {
+    activityMeasurements.remove(sentryId);
+  }
+
+  void close() {
+    frameMetricsAggregator.stop();
```
philipphofmann, line 89: `m`: Should we maybe also clear here or do we want to keep the measurements?
```suggestion
    frameMetricsAggregator.stop();
    activityMeasurements.clear();
```

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -95,6 +95,8 @@ public void close() throws IOException {
     if (options != null) {
       options.getLogger().log(SentryLevel.DEBUG, "ActivityLifecycleIntegration removed.");
     }
+
+    ActivityFramesState.getInstance().close();
```
philipphofmann, line 99: `l`: Maybe state seems a bit weird to me. Maybe tracker fits better here?
```suggestion
    ActivityFramesTracker.getInstance().close();
```

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -185,6 +209,12 @@ private void stopTracing(final @NonNull Activity activity, final boolean shouldF
 
   private void finishTransaction(final @Nullable ITransaction transaction) {
     if (transaction != null) {
+      // if io.sentry.traces.activity.auto-finish.enable is disabled, transaction may be already
+      // finished when this method is called.
+      if (transaction.isFinished()) {
```
philipphofmann, line 214: `l`: It's good that we check that here, but I hope it isn't a problem when you call finish on a finished transaction. Maybe we don't need that check here.

sentry-android-core/src/main/java/io/sentry/android/core/AppStartState.java
```diff
@@ -25,7 +25,7 @@
 
   private AppStartState() {}
 
-  static @NotNull AppStartState getInstance() {
+  public static @NotNull AppStartState getInstance() {
```
philipphofmann, line 28: `m`: Is this a breaking change?

sentry/src/main/java/io/sentry/TransactionListener.java
```diff
@@ -0,0 +1,13 @@
+package io.sentry;
+
+import org.jetbrains.annotations.NotNull;
+
+public interface TransactionListener {
```
philipphofmann, line 5: `m`: This is new to me. Is this something we want to be public? Did we sync this with other SDKs?

philipphofmann (commented):
Looks already good. I added a few comments.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesState.java
```diff
@@ -0,0 +1,91 @@
...
+        for (int i = 0; i < totalIndexArray.size(); i++) {
+          int frameTime = totalIndexArray.keyAt(i);
+          int numFrames = totalIndexArray.valueAt(i);
+          totalFrames += numFrames;
+          // hard coded values, its also in the official android docs and frame metrics API
+          if (frameTime > 700) {
+            // frozen frames, threshold is 700ms
+            frozenFrames += numFrames;
+          }
+          if (frameTime > 16) {
```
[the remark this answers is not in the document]
marandaneto: 16 is hardcoded in the androidx libs and android docs, I'd not change to `1 / 60 = 16,666667` if that's not what's documented

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
marandaneto: this is needed since `stopPreviousTransactions` calls this method, the comment explains why its needed

sentry-android-core/src/main/java/io/sentry/android/core/AppStartState.java
marandaneto: needed for react native, making it public isn't really a breaking change

sentry/src/main/java/io/sentry/TransactionListener.java
marandaneto: this is needed since we have a Java/Android SDK and transactions live only in the Java SDK but the Android SDK needs some tweaks, it was discussed yesterday in the daily standup

sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesState.java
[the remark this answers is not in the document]
philipphofmann: Can we add this info as a comment then, please? It might confuse others as well.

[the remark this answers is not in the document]
marandaneto: it is a comment already

> hard coded values, its also in the official android docs and frame metrics API

check the code again

sentry/src/main/java/io/sentry/TransactionListener.java
philipphofmann: Ah, sorry, I missed that. As we have `BeforeSendCallback` and `BeforeBreadcrumbCallback`, should we maybe name this `TransactionFinishedCallback` and the method below `execute`? @bruno-garcia, what do you think about the naming here?

philipphofmann (commented):
Just one open comment about the naming. Is it hard to add tests for `ActivityFramesTracker` or how is it tested?

sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesState.java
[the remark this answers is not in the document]
philipphofmann: Sorry

sentry/src/main/java/io/sentry/TransactionListener.java
marandaneto: actually its not the same concept as `BeforeSendCallback` and `BeforeBreadcrumbCallback` which is called for every event/breadcrumb.
so I'd not like to add the `Callback` suffix as it could be misunderstood.
the `TransactionListener` is just a callback that is called only when using the `waitForChildren` feature (currently only on Android), so I can trigger some Android code before the transaction gets captured.

philipphofmann: > the TransactionListener is just a callback that is called

😁. No strong opinion here. For me, `TransactionListener` is a bit confusing.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -153,6 +171,12 @@ private void startTracing(final @NonNull Activity activity) {
     }
   }
 
+  private void setMetricsForActivity(
+      final @NotNull Activity activity, final @NotNull ITransaction transaction) {
+    // set metrics and remove the slow/frozen detection since transaction is finished.
```
bruno-garcia, line 176: `L`: Looks like this method doesn't add much. Maybe just inline it?

> and remove the slow/frozen detection since transaction is finished.

Doesn't look like removing it is happening here. Is this expected to happen by calling `setMetrics`? If so the comment is a bit disconnected (relies on a side effect of a method that isn't really named in a way that would indicate such side effect.

sentry-android-core/src/main/java/io/sentry/android/core/PerformanceAndroidEventProcessor.java
```diff
@@ -41,6 +48,16 @@
       }
     }
 
+    final SentryId eventId = transaction.getEventId();
+    if (eventId != null) {
+      final Map<String, @NotNull MeasurementValue> framesMetrics =
+          ActivityFramesTracker.getInstance().getMetrics(eventId);
```
bruno-garcia, line 54: Could we replace the `getMetrics` and `removeMetrics` with a single `takeMetrics` (or a better name) that would do both?

sentry/src/main/java/io/sentry/TransactionListener.java
bruno-garcia: By looking at the usage:
image

It seems to me it's a callback so something like `TransactionFinishedCallback` would make sense.

bruno-garcia (commented):
Just a couple notes (given it's a draft)

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
marandaneto: It's because when you read the metrics from `FrameMetricsAggregator`, it also removes them from the internal map.
I can inline, yes.

sentry-android-core/src/main/java/io/sentry/android/core/PerformanceAndroidEventProcessor.java
marandaneto: yes, thats possible

sentry/src/main/java/io/sentry/TransactionListener.java
marandaneto: I can rename, but the `Span` class takes a `SpanListener` which does pretty much the same thing, so ideally we'd align both of them introducing a breaking change, well its an internal class so not a big deal

https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/SpanListener.java

so do I rename both?

philipphofmann: If it's not a breaking change, then I would yes.

marandaneto: done thanks

sentry/src/main/java/io/sentry/TransactionFinishedCallback.java
```diff
@@ -0,0 +1,13 @@
...
+import org.jetbrains.annotations.NotNull;
+
+public interface TransactionFinishedCallback {
+
+  /**
+   * Called when observed transaction finishes
+   *
+   * @param transaction the transaction
+   */
+  void onTransactionFinished(@NotNull ITransaction transaction);
```
philipphofmann, line 12: `l`: Maybe we can rename this to `execute` as the name of the interface already contains this information. Same applies for `SpanFinishedCallback`.
```suggestion
  void execute(@NotNull ITransaction transaction);
```

philipphofmann (commented):
One tiny comment, and is it possible to add tests for `ActivityFramesTracker`?

sentry-android-core/src/main/java/io/sentry/android/core/AppStartState.java
```diff
@@ -44,14 +44,14 @@ void setAppStartEnd(final long appStartEndMillis) {
   }
 
   @Nullable
-  synchronized Long getAppStartInterval() {
+  public synchronized Long getAppStartInterval() {
```
jennmueng, line 47: Can we instead have public getter methods to separately get the `appStartMillis` and `appStartEndMillis` for RN? We would need the separate timestamps to compute RN's app start with own JS's launch time.

sentry/src/main/java/io/sentry/TransactionFinishedCallback.java
marandaneto: since its an interface with a single method, the method name will never be used (cus u can do lambda or function interfaces), but it makes sense so I will rename it

sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesTracker.java
```diff
@@ -0,0 +1,89 @@
...
+        for (int i = 0; i < totalIndexArray.size(); i++) {
+          int frameTime = totalIndexArray.keyAt(i);
+          int numFrames = totalIndexArray.valueAt(i);
+          totalFrames += numFrames;
+          // hard coded values, its also in the official android docs and frame metrics API
+          if (frameTime > 700) {
+            // frozen frames, threshold is 700ms
+            frozenFrames += numFrames;
+          }
+          if (frameTime > 16 && frameTime <= 700) {
```
bruno-garcia, line 54:
```suggestion
          } else if (frameTime > 16) {
```

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -134,10 +142,26 @@ private void startTracing(final @NonNull Activity activity) {
...
-        transaction = hub.startTransaction(activityName, UI_LOAD_OP, (Date) null, true);
+        transaction =
+            hub.startTransaction(
+                activityName,
+                UI_LOAD_OP,
+                (Date) null,
+                true,
+                (finishingTransaction) -> {
+                  activityFramesTracker.setMetrics(activity, finishingTransaction.getEventId());
+                });
```
bruno-garcia, line 153: I'm probably missing something but couldn't we remove the `if` (and the `else` branch) and just pass `appStartTime` regardless if it's null or not? 

```suggestion
        transaction =
            hub.startTransaction(
                activityName,
                UI_LOAD_OP,
                appStartTime,
                true,
                (finishingTransaction) -> {
                  activityFramesTracker.setMetrics(activity, finishingTransaction.getEventId());
                });
```

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -202,6 +232,9 @@ public synchronized void onActivityPreCreated(
     if (isAllActivityCallbacksAvailable) {
       setColdStart(savedInstanceState);
 
+      // start collecting frame metrics for transaction
+      activityFramesTracker.addActivity(activity);
```
bruno-garcia, line 236: Should we start tracking frames before we `setColdStart` or the order here is by design?

sentry-android-core/src/main/java/io/sentry/android/core/PerformanceAndroidEventProcessor.java
```diff
@@ -41,6 +57,20 @@
       }
     }
 
+    final SentryId eventId = transaction.getEventId();
+    final SpanContext spanContext = transaction.getContexts().getTrace();
+
+    // only add slow/frozen frames to transactions created by ActivityLifecycleIntegration
```
bruno-garcia, line 63: We're technically adding to transactions named `UI_LOAD_OP` which could set by the user.

```suggestion
    // only add slow/frozen frames to transactions created by ActivityLifecycleIntegration
    // which have the operation UI_LOAD_OP. If a user-defined (or hybrid SDK) transaction 
    // users it, we'll also add the metrics if available
```

Worth clarifying it?

sentry-android-core/src/test/java/io/sentry/android/core/PerformanceAndroidEventProcessorTest.kt
```diff
@@ -107,6 +111,31 @@ class PerformanceAndroidEventProcessorTest {
...
+    @Test
+    fun `do not add slow and frozen frames if not auto transaction`() {
+        val sut = fixture.getSut()
+        var tr = getTransaction("task")
+
+        tr = sut.process(tr, null)
+
+        assertTrue(tr.measurements.isEmpty())
+    }
+
```
bruno-garcia, line 123: I wonder if it compiles ;)

```suggestion
    @Test
    fun `do not add slow and frozen frames if not auto transaction`() {
        val sut = fixture.getSut()
        var tr = getTransaction("task")

        tr = sut.process(tr, null)

        assertTrue(tr.measurements.isEmpty())
    }

    @Test
    fun `do not add slow and frozen frames if tracing is disabled`() {
        val sut = fixture.getSut(0)
        var tr = getTransaction("task")

        tr = sut.process(tr, null)

        assertTrue(tr.measurements.isEmpty())
    }

```

bruno-garcia (approved):
Other than some small suggestions LGTM! 🚀

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
marandaneto: moved this block 1st thing when the activity methods are called.
the downside is, if tracing is disabled, we will be doing something that is not gonna be used, but if we do that just after all the checks, we might be losing some frames count, lets live with this trade off then and we start collecting the data as early as possible.

sentry-android-core/src/test/java/io/sentry/android/core/PerformanceAndroidEventProcessorTest.kt
marandaneto: almost, sut gets a Double, so only 0 wont work :D
added the test, although the same use case was already tested by `do not add app start metric if performance is disabled` since the `if` is above both blocks, added anyway as it makes sense.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
marandaneto: there are 2 conditions, `firstActivityCreated || appStartTime == null`, `appStartTime` might not be null and `firstActivityCreated` true so it'd give a wrong transaction start time.

marandaneto:
> Other than some small suggestions LGTM! 🚀

applied them and added one more test too, thanks

PR #1641, opened 2021-08-11
Build against different JVM versions

maciejwalkowiak:
_#skip-changelog_

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,13 +4,14 @@ on:
     branches:
       - main
       - release/**
+      - jvm-matrix
   pull_request:
     branches:
       - '*'
 
 jobs:
   build:
-    name: Build Job ${{ matrix.os }}
+    name: Build Job ${{ matrix.os }} - ${{ matrix.jvm }}
     runs-on: ${{ matrix.os }}
     strategy:
       # we want that the matrix keeps running, default is to cancel them if it fails.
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -4,31 +4,40 @@ on:
...
-          distribution: 'adopt'
-          java-version: '8'
+          candidate: java
+          version: ${{ matrix.jvm }}
+
+      - name: "Set up JVM: ${{ matrix.jvm }}"
+        uses: actions/setup-java@v1
+        with:
+          java-version: 8
+          jdkFile: ${{ steps.sdkman.outputs.file }}
```
marandaneto, line 40: instead of using the jdkFile, we can also use the java-version, see https://github.com/getsentry/sentry-android-gradle-plugin/blob/main/.github/workflows/test-publish-dry-run.yaml#L66-L70

.github/workflows/build.yml
```diff
@@ -4,31 +4,40 @@ on:
...
-      - name: set up JDK 1.8
-        uses: actions/setup-java@v2
+      - name: "Download JVM: ${{ matrix.jvm }}"
+        uses: sdkman/sdkman-action@master
+        id: sdkman
         with:
-          distribution: 'adopt'
-          java-version: '8'
+          candidate: java
+          version: ${{ matrix.jvm }}
```
marandaneto, line 34: GH Action already has all the versions, we don't need to install and download sdkman, `actions/setup-java` is enough

maciejwalkowiak: With SDKMan we can be more specific with versions and distributions. On the other hand, we don't really need it, so agreed - lets stick with what's offered in `setup-java`.

marandaneto:
@maciejwalkowiak we have to add an `if` condition for a few steps, like `Archive packages` should be only run for Java 8, `Upload coverage to Codecov` also only once

maciejwalkowiak:
@marandaneto looks like adding `matrix.java` to `if` statement did the job.

marandaneto (approved):

PR #1644, opened 2021-08-12
Type-safe project accessors in Gradle config.

### LINKED ISSUE #1452: Type-safe project accessors Gradle 7
https://docs.gradle.org/7.0/release-notes.html#upgrade-instructions

```
dependencies {
  // type-safe alternative to project(":client")
  implementation projects.client
}
```

maciejwalkowiak:
## :scroll: Description

Type-safe project accessors in Gradle config.

Fixes #1452.

Since this is an incubating feature, I am not convinced we should use it. I does not seem to hurt but also does not seem to bring much value.

## :pencil: Checklist

- [x] No breaking changes

_#skip-changelog_

### DIFF
--- a/sentry-android-core/build.gradle.kts
+++ b/sentry-android-core/build.gradle.kts
@@ -74,7 +74,7 @@ tasks.withType<JavaCompile>().configureEach {
 }
 
 dependencies {
-    api(project(":sentry"))
+    api(projects.sentry)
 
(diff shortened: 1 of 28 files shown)


marandaneto:
> I does not seem to hurt but also does not seem to bring much value.

it does when you rename a project, but that's the only difference I guess

marandaneto (approved):

PR #1648, opened 2021-08-13
fix: maven artifacts location

romtsn:
#skip-changelog

## :scroll: Description

No idea if it affects something, but it probably does?

## :pencil: Checklist

- [ ] No breaking changes


marandaneto:
actually, it does, good catch

See:
https://github.com/getsentry/sentry-release-registry/blob/master/packages/maven/io.sentry/sentry-android-fragment/5.1.0-beta.1.json

marandaneto (approved):

PR #1650, opened 2021-08-17
Bump AGP to 7.0.2

### LINKED ISSUE #1646: Bump AGP to 7.0.0
Android libs don't have `versionName` and `versionCode` anymore.

it requires Java 11, so GH Action has to be fixed too.

if `versionCode` is totally removed, we can remove the craft bump script too (also `buildVersionCode` from gradle.properties)

`jvmTarget` should still be Java 1.8 though because of compatibility.

marandaneto:
## :scroll: Description
Bump AGP to 7.0.2

Closes #1646

CI/running

## :pencil: Checklist

- [X] No breaking changes


sentry.properties
```diff
@@ -1,4 +1,3 @@
-# how a sentry.properties looks like
-defaults.project=android
-defaults.org=man
-auth.token=
+defaults.project=sentry-android
+defaults.org=sentry-sdks
+auth.token=5ac5c14f29074714a034379bb0d16dc0b33ec3be83e041e6bae979c62307ce8f
```
marandaneto, line 3:
```suggestion
auth.token=
```

marandaneto:
@maciejwalkowiak can you confirm that compiling the Java projects with Java 11 but target/source being Java 8 won't cause any harm? since the new Android Gradle Plugin requires Java 11, either we split up in 2 Gradle projects or we use Java 11 targeting Java 8
Maybe the Matrix instead of Java 8 and 11 now could be Java 11 and 16 or 17

marandaneto:
spotless hits the issue https://github.com/diffplug/spotless/issues/834 if using Java 16

.github/workflows/build.yml
```diff
@@ -19,7 +19,7 @@ jobs:
         # TODO: windows-latest
         os: [ubuntu-latest, macos-latest]
         # Zulu Community distribution of OpenJDK
-        java: [ '8', '11' ]
+        java: [ '11' ]
```
marandaneto, line 22: didn't remove the matrix because we want to add Java 16 later

maciejwalkowiak:
"target" means the latest Java version runtime we want to support. I don't see any way it can cause any harm.

romtsn (approved):

PR #1656, opened 2021-08-18
Merge external map properties

### LINKED ISSUE #874: Unable to set tags at different levels
If sentry is configured  for instance in sentry.properties with tags=commitHash:a5e39b08 and has env property, SENTRY_TAGS="namespaceK8s:test", then tags map is `{ namespaceK8s: test } `while I would have expect tags to be merged in `{ namespaceK8s: test,  commitHash: a5e39b08}`.

I could prepare a pull request to fix it, but I would to know if it may be accepted before.

maciejwalkowiak:
## :scroll: Description

Merge external map properties

Fixes #874

Now tags can be set in multiple places: some in properties file, some in env variables. The resulting options will contain tags from all places.

Unit tests

## :pencil: Checklist

- [ ] No breaking changes

Not breaking in terms of compilation, but resulting tags can differ if they were set in different places (unlikely though).

## :crystal_ball: Next steps

The order of loading properties can (and probably should) be changed. Providers order by priority I think should be:

- system property
- properties files
- environment variables

### DIFF
--- a/sentry/src/main/java/io/sentry/config/CompositePropertiesProvider.java
+++ b/sentry/src/main/java/io/sentry/config/CompositePropertiesProvider.java
@@ -31,12 +31,10 @@ final class CompositePropertiesProvider implements PropertiesProvider {
 
   @Override
   public @NotNull Map<String, String> getMap(final @NotNull String property) {
+    final Map<String, String> result = new ConcurrentHashMap<>();
     for (final PropertiesProvider provider : providers) {
-      final Map<String, String> result = provider.getMap(property);
-      if (!result.isEmpty()) {
-        return result;
-      }
+      result.putAll(provider.getMap(property));
     }
-    return new ConcurrentHashMap<>();
+    return result;
   }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Feat: Merge external map properties ()
+
 ## 5.1.0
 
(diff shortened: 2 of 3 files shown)


sentry/src/main/java/io/sentry/config/CompositePropertiesProvider.java
```diff
@@ -31,12 +31,10 @@ public CompositePropertiesProvider(@NotNull List<PropertiesProvider> providers)
 
   @Override
   public @NotNull Map<String, String> getMap(final @NotNull String property) {
+    final Map<String, String> result = new ConcurrentHashMap<>();
     for (final PropertiesProvider provider : providers) {
```
marandaneto, line 35: // can we add a comment about the precedence over providers? like:

// system property
// env vars.
// simple properties
// ...

just so that its easier to debug when tags are either missing or overwritten

marandaneto (approved):
left a suggestion but LGTM

sentry/src/main/java/io/sentry/config/CompositePropertiesProvider.java
maciejwalkowiak: The order of providers is defined in `PropertiesProviderFactory` where it's documented. `CompositePropertiesProvider` technically is unaware of the specific order.

marandaneto: indeed, sounds good then, thanks

PR #1670, opened 2021-08-20
fix: Properly bootstrap ndk

romtsn:
#skip-changelog

## :scroll: Description

The previous way of bootstrapping the native submodule didn't really work for me (the `sentry-native` folder was always empty after a gradle sync, hence `cmake` couldn't succeed)

* `rm -rf sentry-java`
* `git clone git@github.com:getsentry/sentry-java.git`
* `idea sentry-java`
* Do a gradle sync -> shouldn't fail

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -32,3 +32,12 @@ include(
     "sentry-samples:sentry-samples-spring",
     "sentry-samples:sentry-samples-spring-boot",
     "sentry-samples:sentry-samples-spring-boot-webflux")
+
+gradle.beforeProject {
+    if (project.name == "sentry-android-ndk" || project.name == "sentry-samples-android") {
+        exec {
+            logger.log(LogLevel.LIFECYCLE, "Initializing git submodules")
+            commandLine("git", "submodule", "update", "--init", "--recursive")
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


settings.gradle.kts
```diff
@@ -32,3 +32,12 @@ include(
     "sentry-samples:sentry-samples-spring",
     "sentry-samples:sentry-samples-spring-boot",
     "sentry-samples:sentry-samples-spring-boot-webflux")
+
+gradle.beforeProject {
```
marandaneto, line 36: before we were executing `outputs.dir("${project.projectDir}/$sentryNativeSrc")` so the task was cached and not executed every time, does it mean that now every project sync would try to init all the sub modules?

romtsn: this line wasn't really doing anything, because the submodules paths are dictated by the `.gitmodules` file, and there's no "output" when executing `git submodule ...` per se.

When the modules are already initialized, doing a project sync wouldn't trigger another init of submodules (the git command would just be ignored), unless the submodule was updated to the latest commit - but that's what we want then, to have this automatically bootstrapped/updated in case there are updates.

marandaneto: sounds good, lets merge it

marandaneto (approved):

PR #1680, opened 2021-08-25
Fix: Do not drop Contexts key if Collection, Array or Char

marandaneto:
## :scroll: Description
Fix: Do not drop Contexts key if Collection, Array or Char

```
Sentry.configureScope(scope -> {
  scope.setContexts("inventory", new String[] {"big potion", "small potion"});
});
```

would be dropped because it'd fall to `setContexts(String, Object)` but reflection would make an array instead of an object and array as a value, hence adding convenient method for the missing Java Data types

## :pencil: Checklist

- [X] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-* Fix: Do not drop Contexts key if Collection, Array or Char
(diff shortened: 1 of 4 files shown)


maciejwalkowiak (approved):
Looks good. I would add tests.

marandaneto:
> Looks good. I would add tests.

done

PR #1683, opened 2021-08-25
Feat: Add `tracestate` HTTP header support.

maciejwalkowiak:
## :scroll: Description

Add experimental `tracestate` HTTP header support defined in https://develop.sentry.dev/sdk/performance/trace-context/#client-options.

Trace context is added to each envelope containing transaction or an event.

In addition to that, `SentryOkHttpInterceptor` attaches a new `tracestate` header to all outgoing requests.

By default this feature is off and is controller by a `traceSampling` flag on `SentryOptions`. All new API classes or methods are marked as `@Experimental`.

Unit & Integration tests. Sending events to `sentry-java` project on in `sentry-sdks` and verifying if they appear.

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/sentry/src/main/java/io/sentry/Hub.java
+++ b/sentry/src/main/java/io/sentry/Hub.java
@@ -675,6 +675,18 @@ public final class Hub implements IHub {
   }
 
   @Override
+  public @Nullable TraceStateHeader traceStateHeader() {
+    final Scope scope = stack.peek().getScope();
+    final TraceState traceState = TraceState.create(scope, options);
+    if (traceState != null) {
+      return TraceStateHeader.fromTraceState(
+          traceState, options.getSerializer(), options.getLogger());
+    } else {
+      return null;
+    }
+  }
+
+  @Override
   public @Nullable ISpan getSpan() {
     ISpan span = null;
     if (!isEnabled()) {
--- a/sentry/src/main/java/io/sentry/Hub.java
+++ b/sentry/src/main/java/io/sentry/Hub.java
@@ -531,7 +531,9 @@ public final class Hub implements IHub {
   @ApiStatus.Internal
   @Override
   public @NotNull SentryId captureTransaction(
-      final @NotNull SentryTransaction transaction, final @Nullable Object hint) {
+      final @NotNull SentryTransaction transaction,
+      final @Nullable TraceState traceState,
+      final @Nullable Object hint) {
     Objects.requireNonNull(transaction, "transaction is required");
 
     SentryId sentryId = SentryId.EMPTY_ID;
@@ -561,7 +563,8 @@ public final class Hub implements IHub {
         StackItem item = null;
         try {
           item = stack.peek();
-          sentryId = item.getClient().captureTransaction(transaction, item.getScope(), hint);
+          sentryId =
+              item.getClient().captureTransaction(transaction, traceState, item.getScope(), hint);
         } catch (Exception e) {
           options
               .getLogger()
@@ -672,27 +675,6 @@ public final class Hub implements IHub {
       }
     }
     return traceHeader;
-  }
-
-  @Override
-  public @Nullable TraceStateHeader traceStateHeader() {
-    TraceStateHeader header = null;
-    if (!isEnabled()) {
-      options
-          .getLogger()
-          .log(
-              SentryLevel.WARNING,
-              "Instance is disabled and this 'traceStateHeader' call is a no-op.");
-    } else {
-      final Scope scope = stack.peek().getScope();
-      final TraceState traceState = TraceState.create(scope, options);
-      if (traceState != null) {
-        header =
-            TraceStateHeader.fromTraceState(
-                traceState, options.getSerializer(), options.getLogger());
-      }
-    }
-    return header;
   }
 
   @Override
--- a/sentry/src/main/java/io/sentry/Hub.java
+++ b/sentry/src/main/java/io/sentry/Hub.java
@@ -676,14 +676,23 @@ public final class Hub implements IHub {
 
   @Override
   public @Nullable TraceStateHeader traceStateHeader() {
-    final Scope scope = stack.peek().getScope();
-    final TraceState traceState = TraceState.create(scope, options);
-    if (traceState != null) {
-      return TraceStateHeader.fromTraceState(
-          traceState, options.getSerializer(), options.getLogger());
+    TraceStateHeader header = null;
+    if (!isEnabled()) {
+      options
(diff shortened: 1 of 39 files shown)


sentry/src/main/java/io/sentry/Hub.java
```diff
@@ -674,6 +674,27 @@ public void flush(long timeoutMillis) {
     return traceHeader;
   }
 
+  @Override
+  public @Nullable TraceStateHeader traceStateHeader() {
```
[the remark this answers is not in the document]
maciejwalkowiak: We don't need it on `Sentry`, but we do need it on `IHub` so that integrations - like OkHttp one, could set the header. At least thats one way to do it and I don't know yet about another way.

sentry/src/main/java/io/sentry/SentryClient.java
```diff
@@ -437,7 +434,11 @@ public void captureSession(final @NotNull Session session, final @Nullable Objec
 
     try {
       final SentryEnvelope envelope =
-          buildEnvelope(transaction, filterForTransaction(getAttachmentsFromScope(scope)));
+          buildEnvelope(
+              transaction,
+              filterForTransaction(getAttachmentsFromScope(scope)),
+              null,
+              scope != null ? TraceState.create(transaction, scope, options) : null);
```
brustolin, line 441: The only information extracted from scope is the user, which is an optional information.
So, why create the TraceState only if there is a scope?

bruno-garcia (approved):
Couple notes but lets merge and make a release of this please.
Lets make a beta first please

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -216,7 +220,35 @@ public void finish(@Nullable SpanStatus status) {
...
+        this.traceState = new TraceState(this, userAtomicReference.get(), hub.getOptions());
+      }
+      return this.traceState;
+    } else {
+      return null;
+    }
+  }
+
+  @Override
+  public @Nullable TraceStateHeader toTraceStateHeader() {
```
bruno-garcia, line 245: If `TraceState` has a reference to `options`, could we have `toTraceStateHeader` be part of `TraceState` instead and avoid this method?

maciejwalkowiak: `TraceState` does not have a reference to `options`. We can put it to a transient field, just somehow does not seem right to me as `TraceState` was meant to be just a data container.

sentry/src/main/java/io/sentry/ISpan.java
```diff
@@ -39,6 +39,26 @@ ISpan startChild(
   @NotNull
   SentryTraceHeader toSentryTrace();
 
+  /**
+   * Returns the trace state information. @see <a
+   * href="https://develop.sentry.dev/sdk/trace-context/">Trace Context</a>.
+   *
+   * @return a trace state or {@code null} if {@link SentryOptions#isTraceSampling()} is disabled.
```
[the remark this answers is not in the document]
maciejwalkowiak: It's a method on a span, so if i span exists, there is a transaction, right?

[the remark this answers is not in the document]
bruno-garcia: Right. I guess it could be null also if any required value for a trace state doesn't exist?

sentry/src/main/java/io/sentry/SentryTracer.java
bruno-garcia: That's a fair point. I'm just trying to find any way not to add these two public methods here but I'm about to lose hope (the PR is approved anyway so)

bruno-garcia:
@marandaneto we need to ship soon, would you like to give this a pass before we merge?

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -280,6 +280,9 @@
    */
   private @NotNull RequestSize maxRequestBodySize = RequestSize.NONE;
 
+  /** Controls if the `tracestate` header is attached to envelopes and HTTP client integrations. */
+  private boolean traceSampling;
```
marandaneto, line 284: is it disabled by default?
if this is intended for mobile too, let's add the ability to turn it on via Android Manifest
See: ManifestMetadataReader

sentry/src/main/java/io/sentry/SentryTracer.java
marandaneto: `SentryTracer` is `@ApiStatus.Internal` already, so it'd be fine to be there.

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -216,7 +220,35 @@ public void finish(@Nullable SpanStatus status) {
...
-      hub.captureTransaction(transaction);
+      hub.captureTransaction(transaction, this.traceState());
+    }
+  }
+
+  @Override
+  public @Nullable TraceState traceState() {
+    if (hub.getOptions().isTraceSampling()) {
+      if (traceState == null) {
+        final AtomicReference<User> userAtomicReference = new AtomicReference<>();
```
marandaneto, line 231: what's about adding a function to the Scope called `withUser` like we do with `withSession` instead of using `AtomicReference`? that executes a callback and returns the user atomically, it could be package private or `@ApiStatus.Internal`. not a big fun of  `AtomicReference`

maciejwalkowiak: I agree, also didn't like it, but in this PR we need to avoid introducing changes to the API, so i think `AtomicReference` as a temporary solution is a right trade off.

sentry/src/main/java/io/sentry/SentryOptions.java
maciejwalkowiak: 👍

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -216,7 +220,35 @@ public void finish(@Nullable SpanStatus status) {
...
+        hub.configureScope(
+            scope -> {
+              userAtomicReference.set(scope.getUser());
+            });
+        this.traceState = new TraceState(this, userAtomicReference.get(), hub.getOptions());
+      }
+      return this.traceState;
+    } else {
+      return null;
+    }
```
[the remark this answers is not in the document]
maciejwalkowiak: 👍

marandaneto (approved):

PR #1694, opened 2021-09-01
Update proguard keep rule for enums

### LINKED ISSUE #1684: Proguard rule for enums in sentry-android-core is too broad
The version of the SDK:
**5.1.2**

Proguard rule for enums is embedded into application proguard config and it's affecting all application enums, effectively preventing obfuscation:

https://github.com/getsentry/sentry-java/blob/eb30c909381c1df9d211f5c3bba6fa26663982a1/sentry-android-core/proguard-rules.pro#L15

Could we reduce it down to only library classes?
Something like this:
```
-keepclassmembers enum io.sentry.** { *; }
```

romankivalin:
## :scroll: Description
Updated proguard rule to target library enums only.

This is a consumer proguard file, which means it is included into consumer application proguard config, and if we target `*`, it will apply to each class of consumer application too. This unreasonably prevents obfuscation of consumer application enums.

Fixes #1684

Didn't test.

## :pencil: Checklist

- [ ] No breaking changes

I think it's a breaking change because consumer application could accidentally rely on this rule, and removing it may break obfucated code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 Breaking changes:
 
-* Updated proguard keep rule for enums, which affects consumer application code (#1684)
+* Updated proguard keep rule for enums, which affects consumer application code (#1694)
 
 ## 5.1.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+Breaking changes:
+
+* Updated proguard keep rule for enums, which affects consumer application code (#1684)
+
 ## 5.1.2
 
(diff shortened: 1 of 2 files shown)


marandaneto:
thanks for doing this @romankivalin

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+Breaking changes:
+
+* Updated proguard keep rule for enums, which affects consumer application code (#1684)
```
marandaneto, line 7:
```suggestion
* Updated proguard keep rule for enums, which affects consumer application code (#1694)
```

marandaneto:
@dmytroKarataiev the change is part of our changelog, how would you suggest then?

PR #1698, opened 2021-09-03
Feat: Add option to filter which origins receive tracing headers

### LINKED ISSUE #1585: Add support for tracing origins
Currently the [SentryOkHttpInterceptor](https://github.com/getsentry/sentry-java/blob/aca78d9ca024c75225e59c36e35622055d2f6cbf/sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpInterceptor.kt#L33-L35) adds the [sentry-trace HTTP header](https://develop.sentry.dev/sdk/performance/#header-sentry-trace) to every HTTP request. Some HTTP requests are targeting backends that don't have support for the sentry-trace HTTP header. Therefore JavaScript has the concept of [tracingOrigins](https://docs.sentry.io/platforms/javascript/performance/instrumentation/automatic-instrumentation/#tracingorigins), which is a list of URLs to which the integration should append the sentry-trace HTTP header.

### LINKED ISSUE #1697: Add allowList for the `tracestate` header
related to https://github.com/getsentry/sentry-java/pull/1683
reason is https://github.com/getsentry/develop/issues/425

cc @bruno-garcia 

it should be a List in the SentryOptions with an experimental annotation

maciejwalkowiak:
## :scroll: Description

Add `tracingOrigins` to `SentryOptions` and respect it in OkHttp, WebClient and RestTemplate integrations.

Fixes #1697
Fixes #1585

Integration tests.

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/sentry/src/main/java/io/sentry/SentryOptions.java
+++ b/sentry/src/main/java/io/sentry/SentryOptions.java
@@ -284,6 +284,11 @@ public class SentryOptions {
   private boolean traceSampling;
 
   /**
+   * Contains a list of origins to which `sentry-trace` header should be sent in HTTP integrations.
+   */
+  private @NotNull List<String> tracingOrigins = new ArrayList<>();
+
+  /**
    * Creates {@link SentryOptions} from properties provided by a {@link PropertiesProvider}.
    *
    * @param propertiesProvider the properties provider
@@ -1455,6 +1460,24 @@ public class SentryOptions {
   @ApiStatus.Experimental
   public void setTraceSampling(boolean traceSampling) {
     this.traceSampling = traceSampling;
+  }
+
+  /**
+   * Returns a list of origins to which `sentry-trace` header should be sent in HTTP integrations.
+   *
+   * @return the list of origins
+   */
+  public @NotNull List<String> getTracingOrigins() {
+    return tracingOrigins;
+  }
+
+  /**
+   * Sets a list of origins to which `sentry-trace` header should be sent in HTTP integrations.
+   *
+   * @param tracingOrigins - the tracing origins
+   */
+  public void setTracingOrigins(final @NotNull List<String> tracingOrigins) {
+    this.tracingOrigins = tracingOrigins;
   }
 
   /** The BeforeSend callback */
--- a/sentry/src/main/java/io/sentry/SentryOptions.java
+++ b/sentry/src/main/java/io/sentry/SentryOptions.java
@@ -333,6 +333,9 @@ public class SentryOptions {
     for (final String inAppExclude : propertiesProvider.getList("in-app-excludes")) {
       options.addInAppExclude(inAppExclude);
     }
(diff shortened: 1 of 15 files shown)


sentry/src/main/java/io/sentry/TracingOrigins.java
```diff
@@ -0,0 +1,37 @@
...
+
+  public TracingOrigins(final @NotNull List<String> origins) {
+    this.origins = origins;
+  }
+
+  public TracingOrigins(final @NotNull String... origins) {
+    this(Arrays.asList(origins));
+  }
+
+  public boolean contain(final @NotNull String url) {
```
marandaneto, line 22: what's about making it a static method that takes the list and the URL?
otherwise we have to create an instance of TracingOrigins for every check (eg every HTTP call)

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -1457,6 +1462,24 @@ public void setTraceSampling(boolean traceSampling) {
...
+  }
+
+  /**
+   * Sets a list of origins to which `sentry-trace` header should be sent in HTTP integrations.
+   *
+   * @param tracingOrigins - the tracing origins
+   */
+  public void setTracingOrigins(final @NotNull List<String> tracingOrigins) {
+    this.tracingOrigins = tracingOrigins;
+  }
```
marandaneto, line 1481: usually, we offer the methods `addX` eg `addEventProcessor` but never the instance of the List itself

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -283,6 +283,11 @@
   /** Controls if the `tracestate` header is attached to envelopes and HTTP client integrations. */
   private boolean traceSampling;
 
+  /**
+   * Contains a list of origins to which `sentry-trace` header should be sent in HTTP integrations.
+   */
+  private @NotNull List<String> tracingOrigins = new ArrayList<>();
```
marandaneto, line 289:
```suggestion
  private @NotNull List<String> tracingOrigins = new CopyOnWriteArrayList<>();
```

the reason is that people may iterate it and remove items on the go and if not thread-safe, the SDK crashes, its a client error, but it happened a few times.

maciejwalkowiak:
I realised we also need add option to set this property from the external configuration.

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -328,6 +333,9 @@
     for (final String inAppExclude : propertiesProvider.getList("in-app-excludes")) {
       options.addInAppExclude(inAppExclude);
     }
+    for (final String tracingOrigin : propertiesProvider.getList("tracing-origins")) {
+      options.addTracingOrigin(tracingOrigin);
+    }
```
marandaneto, line 338: we need this ability in the Android Manifest too

marandaneto (approved):
LGTM

PR #1699, opened 2021-09-04
Feat: Include unfinished spans in transaction.

### LINKED ISSUE #1690: Include unfinished spans in transactions
When a transaction is finished, the SDK tries to send the data to Sentry.
If an open Span (one that was not finished) exist in that transaction, the SDK removes that span before sending the data to Sentry.

That was done because [Sentry will discard the transaction if unfinished spans are included](https://github.com/getsentry/relay/blob/569678659bac656d6c4a6edcbdf3381677b5eaa2/relay-general/src/store/transactions.rs#L110-L130). More context on this thread: https://github.com/getsentry/develop/issues/274

An alternative solution used by the JavaScript SDK is to complete those spans with the timestamp of the finishing transaction and setting the status to `timeline_exceeded`. We should align with this approach as it shows something was going on even though it wasn't completed during the transaction.

On:

```[tasklist]
### Tasks
- [ ] https://github.com/getsentry/sentry-cocoa/issues/1303
- [ ] https://github.com/getsentry/sentry-dotnet/issues/1182
- [ ] https://github.com/getsentry/relay/issues/1244
- [ ] https://github.com/getsentry/sentry-dotnet/issues/1182
```

maciejwalkowiak:
## :scroll: Description

Include unfinished spans in transaction

Fixes #1690.

Unit tests.

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/sentry/src/main/java/io/sentry/SentryTracer.java
+++ b/sentry/src/main/java/io/sentry/SentryTracer.java
@@ -209,20 +209,20 @@ public final class SentryTracer implements ITransaction {
       root.finish(finishStatus.spanStatus);
 
       // finish unfinished children
-      for (Span child : children) {
+      Date finishTimestamp = root.getTimestamp();
+      if (finishTimestamp == null) {
+        hub.getOptions()
+            .getLogger()
+            .log(
+                SentryLevel.WARNING,
+                "Root span - op: %s, description: %s - has no timestamp set, when finishing unfinished spans.",
+                root.getOperation(),
+                root.getDescription());
+        finishTimestamp = DateUtils.getCurrentDateTime();
+      }
+      for (final Span child : children) {
         if (!child.isFinished()) {
-          final Date rootTimestamp = root.getTimestamp();
-          if (rootTimestamp != null) {
-            child.finish(SpanStatus.DEADLINE_EXCEEDED, rootTimestamp);
-          } else {
-            hub.getOptions()
-                .getLogger()
-                .log(
-                    SentryLevel.WARNING,
-                    "Root span - op: %s, description: %s - has no timestamp set, when finishing unfinished spans.",
-                    root.getOperation(),
-                    root.getDescription());
-          }
+          child.finish(SpanStatus.DEADLINE_EXCEEDED, finishTimestamp);
         }
       }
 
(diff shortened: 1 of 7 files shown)


sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -207,6 +207,25 @@ public void finish(@Nullable SpanStatus status) {
     this.finishStatus = FinishStatus.finishing(status);
     if (!root.isFinished() && (!waitForChildren || hasAllChildrenFinished())) {
       root.finish(finishStatus.spanStatus);
+
+      // finish unfinished children
+      for (Span child : children) {
```
marandaneto, line 212:
```suggestion
      for (final Span child : children) {
```

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -207,6 +207,25 @@ public void finish(@Nullable SpanStatus status) {
...
+          if (rootTimestamp != null) {
+            child.finish(SpanStatus.DEADLINE_EXCEEDED, rootTimestamp);
+          } else {
+            hub.getOptions()
+                .getLogger()
+                .log(
+                    SentryLevel.WARNING,
+                    "Root span - op: %s, description: %s - has no timestamp set, when finishing unfinished spans.",
+                    root.getOperation(),
+                    root.getDescription());
```
marandaneto, line 224: logging is fine but we still have to `finish` the `child` span with the current time if not `rootTimestamp`, otherwise the whole transaction would be dropped

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -207,6 +207,25 @@ public void finish(@Nullable SpanStatus status) {
     this.finishStatus = FinishStatus.finishing(status);
     if (!root.isFinished() && (!waitForChildren || hasAllChildrenFinished())) {
       root.finish(finishStatus.spanStatus);
+
+      // finish unfinished children
+      Date finishTimestamp = root.getTimestamp();
```
marandaneto, line 212:
```suggestion
      final Date finishTimestamp = root.getTimestamp();
```

marandaneto (approved):
LGTM

sentry/src/main/java/io/sentry/SentryTracer.java
maciejwalkowiak: It can't be final as it may be reassigned in line 221.

PR #1700, opened 2021-09-06
Fix: Move tags from transaction.contexts.trace.tags to transaction.tags

### LINKED ISSUE #1693: `transaction.contexts.trace.tags` should go under `transaction.tags`
right now the transaction tags go under `transaction.contexts.trace.tags` but it should go under `transaction.tags`

`transaction.contexts.trace` should only have a few fields which are documented here https://develop.sentry.dev/sdk/event-payloads/transaction/#examples

maciejwalkowiak:
## :scroll: Description

Move tags from `transaction.contexts.trace.tags` to `transaction.tags`.

Fixes #1693

The implementation could be cleaner - like introducing a dedicated `TraceContext` object, but this would mean we introduce a breaking change - something to keep in mind for 6.0.

Unit tests.

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/sentry/src/test/java/io/sentry/SentryTracerTest.kt
+++ b/sentry/src/test/java/io/sentry/SentryTracerTest.kt
@@ -328,8 +328,8 @@ class SentryTracerTest {
         transaction.finish(SpanStatus.UNKNOWN_ERROR)
 
         // call only once
-        verify(fixture.hub, times(1)).setSpanContext(ex, transaction.root, "name")
-        verify(fixture.hub, times(1)).captureTransaction(check {
+        verify(fixture.hub).setSpanContext(ex, transaction.root, "name")
+        verify(fixture.hub).captureTransaction(check {
             assertNotNull(it.contexts.trace) {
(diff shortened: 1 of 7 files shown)


sentry/src/test/java/io/sentry/SentryTracerTest.kt
```diff
@@ -298,9 +328,11 @@ class SentryTracerTest {
         transaction.finish(SpanStatus.UNKNOWN_ERROR)
 
         // call only once
-        verify(fixture.hub).setSpanContext(ex, transaction.root, "name")
-        verify(fixture.hub).captureTransaction(check {
-            assertEquals(transaction.root.spanContext, it.contexts.trace)
+        verify(fixture.hub, times(1)).setSpanContext(ex, transaction.root, "name")
+        verify(fixture.hub, times(1)).captureTransaction(check {
```
marandaneto, line 332: iirc `times(1)` is the default value, we can remove it

marandaneto (approved):
left a comment but rather than that LGTM, thanks @maciejwalkowiak

PR #1702, opened 2021-09-07
Feat: Add static helpers for creating breadcrumbs.

### LINKED ISSUE #361: Add better Breadcrumb helpers
Like so: https://github.com/joshdholtz/Sentry-Android/blob/master/sentry-android/src/main/java/com/joshdholtz/sentry/Sentry.java#L645

Also could use better Context helpers, such as one to set the User.

maciejwalkowiak:
## :scroll: Description

Add static helper methods for creating breadcrumbs, that render in special way in the UI.

Breadcrumbs created using these methods look following:

image

The question is if there is a value in having separate UI and User breadcrumb - as far as I can see they render the same way.

Fixes #361

Unit tests.

## :pencil: Checklist

- [x] No breaking changes


marandaneto (approved):
LGTM

PR #1703, opened 2021-09-07
Feat: Attach Java vendor and version to events and transactions.

maciejwalkowiak:
## :scroll: Description

Attach Java vendor and version to events and transactions.

Came up once during discussion with @bruno-garcia, that this functionality should be in the core of library and not just as a sample event processor in a Spring Boot sample.

Unit tests.

## :pencil: Checklist

- [x] No breaking changes


sentry/src/main/java/io/sentry/SentryRuntimeEventProcessor.java
```diff
@@ -0,0 +1,48 @@
...
+    return process(event);
+  }
+
+  @Override
+  public @Nullable SentryTransaction process(
+      final @NotNull SentryTransaction transaction, final @Nullable Object hint) {
+    return process(transaction);
+  }
+
+  private <T extends SentryBaseEvent> @Nullable T process(final @Nullable T event) {
```
marandaneto, line 35:
```suggestion
  private <T extends SentryBaseEvent> @Nullable T process(final @NotNull T event) {
```

the `event` cant be null, so it is `@NotNull`

sentry/src/main/java/io/sentry/SentryRuntimeEventProcessor.java
```diff
@@ -0,0 +1,48 @@
...
+    if (event != null) {
+      if (event.getContexts().getRuntime() == null) {
+        event.getContexts().setRuntime(new SentryRuntime());
+      }
+      final SentryRuntime runtime = event.getContexts().getRuntime();
+      if (runtime != null && runtime.getName() == null && runtime.getVersion() == null) {
```
marandaneto, line 41: should we make 2 ifs? in case there's the name but no version and the other way around?

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -1560,6 +1560,7 @@ private SentryOptions(final boolean empty) {
 
       eventProcessors.add(new MainEventProcessor(this));
       eventProcessors.add(new DuplicateEventDetectionEventProcessor(this));
+      eventProcessors.add(new SentryRuntimeEventProcessor());
```
marandaneto, line 1563: this is going to be executed by Android for every event and transaction although its never going to do anything since those properties don't exist, how can we prevent that? Android has its own Runtime, which is the ART or Dalvik

sentry/src/main/java/io/sentry/SentryRuntimeEventProcessor.java
maciejwalkowiak: I think we should not. If someone set the runtime name or version, likely he had already an idea on what should runtime contain.

marandaneto: yeah makes sense, lets keep as it is then

sentry/src/main/java/io/sentry/SentryRuntimeEventProcessor.java
maciejwalkowiak: 👍

sentry/src/main/java/io/sentry/SentryOptions.java
maciejwalkowiak: Hmm didn't know its not there in Android. Maybe then we add a method to this processor like `isConfigured()` (or better name), that checks if these properties are set and add to event processors only then?

marandaneto: to be honest I'd not add the event processor at all based on a "mode/flag" if possible, executing the event processor is already "wasting resources", maybe `SentryAndroidOptions` passes via the super ctor that its a mobile mode and don't add the event processor? we can reuse this flag for some other cases that run into the same issue, default value of this flag is obviously "true"

maciejwalkowiak: I meant that based on the value returned from `isConfigured` event processor would or wouldn't be added to options. So only during the initialization the check would have been done.

bruno-garcia: It could be an integration that when `register` is called it adds this processor to the list in the `options` it receives.

On the Android init code we just remove the integration from the list?

marandaneto: could be too but sounds like overengineering having an integration per event processor for every use case like this, a boolean would do the job if its something meant for Android or not

marandaneto: @maciejwalkowiak up to you either adding a new integration that gets added by `sentry` by default and removed by `sentry-android-core`, or just adding a boolean that decides if we add the event processor or not, the later sounds better IMO

maciejwalkowiak: I've changed it to add event processor only when running on JVM.

sentry/src/main/java/io/sentry/util/Platform.java
```diff
@@ -0,0 +1,19 @@
+package io.sentry.util;
+
+import org.jetbrains.annotations.ApiStatus;
+
+@ApiStatus.Internal
+public final class Platform {
+  // All system properties on Android:
+  // https://developer.android.com/reference/java/lang/System#getProperties()
+  private static final boolean isAndroid =
+      "The Android Project".equals(System.getProperty("java.vendor"));
```
marandaneto, line 10: it could throw `SecurityException` or `IllegalArgumentException`, if so, SDK will crash the App, should we handle such a use case?

marandaneto (approved):
left a comment, other than that, LGTM

sentry/src/main/java/io/sentry/util/Platform.java
maciejwalkowiak: Never seen it happening in wild, but better safe than sorry. Exceptions are now caught.

marandaneto (approved):

PR #1705, opened 2021-09-09
Feat: Performance support for Android Apollo

### LINKED ISSUE #1331: Support for performance tracking with Apollo on Android
We use [Apollo android ](https://github.com/apollographql/apollo-android) extensively through our app. 

This uses OkHttp under the hood which means I assume each request will be traced in the performance dashboard once this is complete -> https://github.com/getsentry/sentry-java/pull/1330

However with GraphQL each request posts to the same URL. The body of the post contains the query which tells the backend what to do eg a query or mutation... 

This I believe would make it difficult to understand what each network request is doing... 

In the meantime we can do something similar to what the OkHttp interceptor is doing but it would be great if this came out of the box for us.

maciejwalkowiak:
## :scroll: Description

Performance support for Android Apollo.

Apollo executes callback on another thread (or coroutine), which makes it impossible to set breadcrumbs, as the context is not propagated back to parent.

Since all spans are finished now automatically, `BeforeSpanCallback` does not allow dropping spans (this will likely change when we fix this for Feign and OkHttp).

Generated classes for test purposes, are generated using Apollo plugin for the API: https://apollo-fullstack-tutorial.herokuapp.com/graphql

Fixes #1331

Integration test.

## :pencil: Checklist

- [x] No breaking changes


buildSrc/src/main/java/Config.kt
```diff
@@ -111,6 +115,7 @@ object Config {
         val mockitoInline = "org.mockito:mockito-inline:3.10.0"
         val awaitility = "org.awaitility:awaitility-kotlin:4.1.0"
         val mockWebserver = "com.squareup.okhttp3:mockwebserver:4.9.0"
+        val mockWebserver3 = "com.squareup.okhttp3:mockwebserver:3.14.9"
```
bruno-garcia, line 118: Curious why do we need two versions? Android vs Java?

sentry-apollo/src/main/java/io/sentry/apollo/SentryApolloInterceptor.kt
```diff
@@ -0,0 +1,105 @@
...
+
+    override fun interceptAsync(request: InterceptorRequest, chain: ApolloInterceptorChain, dispatcher: Executor, callBack: CallBack) {
+        val activeSpan = hub.span
+        if (activeSpan == null) {
+            chain.proceedAsync(request, dispatcher, callBack)
+        } else {
+            val span = startChild(request, activeSpan)
+            val sentryTraceHeader = span.toSentryTrace()
+
+            // we have no access to URI, no way to verify tracing origins
```
bruno-garcia, line 35: Should we address this before we ship it?
Sounds like we should bring tracing options (allowlist of outgoing URLs to add the header info) going forward

sentry-apollo/src/main/java/io/sentry/apollo/SentryApolloInterceptor.kt
```diff
@@ -0,0 +1,105 @@
...
+            val headers = request.requestHeaders.toBuilder().addHeader(sentryTraceHeader.name, sentryTraceHeader.value).build()
+            val requestWithHeader = request.toBuilder().requestHeaders(headers).build()
+            span.setTag("operationId", requestWithHeader.operation.operationId())
+            span.setTag("variables", requestWithHeader.operation.variables().valueMap().toString())
+
+            chain.proceedAsync(requestWithHeader, dispatcher, object : CallBack {
+                override fun onResponse(response: InterceptorResponse) {
+                    span.status = response.httpResponse.map { SpanStatus.fromHttpStatusCode(it.code(), SpanStatus.INTERNAL_ERROR) }
+                        .or(SpanStatus.UNKNOWN)
```
bruno-garcia, line 44: Sounds like we use `INTERNAL_ERROR` as the default/fallback on the map so it won't even get to `or` here? Is this to avoid nulls?

I think we should also use `UNKNOWN` on the `fromHttpStatusCode`'s default arg:

```suggestion
                    span.status = response.httpResponse.map { SpanStatus.fromHttpStatusCode(it.code(), SpanStatus.UNKNOWN) }
                        .or(SpanStatus.UNKNOWN)
```

sentry-apollo/src/main/java/io/sentry/apollo/SentryApolloInterceptor.kt
```diff
@@ -0,0 +1,105 @@
...
+            else -> request.operation.javaClass.simpleName
+        }
+        val description = "$operationType $operation"
+        return activeSpan.startChild(operation, description)
+    }
+
+    private fun finish(span: ISpan, request: InterceptorRequest, response: InterceptorResponse? = null) {
+        var newSpan: ISpan = span
+        if (beforeSpan != null) {
+            newSpan = beforeSpan.execute(span, request, response)
```
bruno-garcia, line 87: there's no chance of `execute` throwing then? Since there's `onFailure` ?

sentry-apollo/src/main/java/io/sentry/apollo/SentryApolloInterceptor.kt
marandaneto: `INTERNAL_ERROR` if it throws, or do not set anything if `it.code()` does not match, discussed here https://github.com/getsentry/develop/issues/307#issuecomment-822240763

sentry-apollo/src/main/java/io/sentry/apollo/SentryApolloInterceptor.kt
marandaneto: looks like a limitation of apollo

buildSrc/src/main/java/Config.kt
marandaneto: apollo probably depends on an older version

sentry-apollo/src/main/java/io/sentry/apollo/SentryApolloInterceptor.kt
marandaneto: it can throw if the clients callback code throws, if this is an issue we should handle, we have to fix it on the other integrations too (okhttp, feign and this one), good catch

sentry-apollo/src/main/java/io/sentry/apollo/SentryApolloInterceptor.kt
marandaneto: you can do with the response object, `response.httpResponse.get().request().url()` but its too late

sentry-apollo/src/main/java/io/sentry/apollo/SentryApolloInterceptor.kt
```diff
@@ -0,0 +1,105 @@
...
+            chain.proceedAsync(request, dispatcher, callBack)
+        } else {
+            val span = startChild(request, activeSpan)
+            val sentryTraceHeader = span.toSentryTrace()
+
+            // we have no access to URI, no way to verify tracing origins
+            val headers = request.requestHeaders.toBuilder().addHeader(sentryTraceHeader.name, sentryTraceHeader.value).build()
+            val requestWithHeader = request.toBuilder().requestHeaders(headers).build()
+            span.setTag("operationId", requestWithHeader.operation.operationId())
+            span.setTag("variables", requestWithHeader.operation.variables().valueMap().toString())
```
marandaneto, line 39: a map as a tag isn't a good idea, whats about adding it as `data`? but then we are blocked by https://github.com/getsentry/sentry-java/issues/1233

marandaneto: the reason is that this tag can be very long, so it's not useful to be searchable and it'll be truncated by the UI anyway

bruno-garcia: Agree. We have to be careful with what tags we're adding out of the box as this results in potential costs on the server when high cardinality tags are used

Please default to `Contexts` or even `Extra` and if it should be a tag, lets discuss internally to consider a protocol change before adding as a tag

marandaneto: span does not have access to `contexts` or `extras` but `data`, so we have to work on https://github.com/getsentry/sentry-java/issues/1233 before going forward.

@maciejwalkowiak how `variables` are important? is that the map that tells the user which query has been used or what? if its something we can add later, lets do it later, if this is the most important part to identify where the span comes from, then we have to work on #1233

bruno-garcia: `data` is the way to go

maciejwalkowiak: @marandaneto variables add more context to the request, but its not absolutely neccessary. Removed tags, lets add variables to `data` later once #1233 is there.

marandaneto (approved):
LGTM

marandaneto:
docs issue https://github.com/getsentry/sentry-docs/issues/4151

PR #1713, opened 2021-09-14
Fix: drop spans in BeforeSpanCallback.

### LINKED ISSUE #1708: BeforeSpanCallback isnt dropping spans
Bug introduced by https://github.com/getsentry/sentry-java/pull/1699 since the span isn't finished, but the integration does it automatically

maciejwalkowiak:
## :scroll: Description

Fix: drop spans in BeforeSpanCallback. #1713

Fixes #1708

## :pencil: Checklist

- [x] No breaking changes

_#skip-changelog_

### DIFF
--- a/sentry-openfeign/src/main/java/io/sentry/openfeign/SentryFeignClient.java
+++ b/sentry-openfeign/src/main/java/io/sentry/openfeign/SentryFeignClient.java
@@ -60,9 +60,16 @@ public final class SentryFeignClient implements Client {
         throw e;
       } finally {
         if (beforeSpan != null) {
-          span = beforeSpan.execute(span, request, response);
-        }
-        if (span != null) {
+          final ISpan result = beforeSpan.execute(span, request, response);
+
+          if (result == null) {
+            // span is dropped
+            span.getSpanContext().setSampled(false);
+          } else {
+            span = result;
+            span.finish();
+          }
+        } else {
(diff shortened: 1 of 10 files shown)


sentry-openfeign/src/main/java/io/sentry/openfeign/SentryFeignClient.java
```diff
@@ -60,9 +60,16 @@ public Response execute(final @NotNull Request request, final @NotNull Request.O
...
+          final ISpan result = beforeSpan.execute(span, request, response);
+
+          if (result == null) {
+            // span is dropped
+            span.getSpanContext().setSampled(false);
+          } else {
+            span = result;
+            span.finish();
+          }
+        } else {
```
marandaneto, line 72: the logic here is a bit different than the one in the OkHttp, which one is correct?

marandaneto: if I got it right, we dont finish the span in case its dropped

maciejwalkowiak: Thanks, good catch.

marandaneto (approved):

marandaneto:
this one should have a changelog since beta.1 has been published, adding it

Issue #1711, opened 2021-09-11
Android SDK crashes on Unity player


bruno-garcia:
While trying to add the Android SDK to the Unity SDK through https://github.com/getsentry/sentry-unity/pull/307:

Building from `main` branch, with the SDK added but no DSN given, the app fails to launch with:

```
09-11 17:38:03.873  6108  6108 D Sentry  : io.sentry.auto-init read: true
09-11 17:38:03.873  6108  6108 I Sentry  : Retrieving auto-init from AndroidManifest.xml
09-11 17:38:03.890  1199  6134 E ResolverController: No valid NAT64 prefix (100, <unspecified>/0)
09-11 17:38:03.899  6108  6108 D AndroidRuntime: Shutting down VM
09-11 17:38:03.900  6108  6108 E AndroidRuntime: FATAL EXCEPTION: main
09-11 17:38:03.900  6108  6108 E AndroidRuntime: Process: com.DefaultCompany.unityofbugs, PID: 6108
09-11 17:38:03.900  6108  6108 E AndroidRuntime: java.lang.NoClassDefFoundError: Failed resolution of: Landroidx/core/app/FrameMetricsAggregator;
09-11 17:38:03.900  6108  6108 E AndroidRuntime: 	at io.sentry.android.core.ActivityFramesTracker.<init>(ActivityFramesTracker.java:20)
09-11 17:38:03.900  6108  6108 E AndroidRuntime: 	at io.sentry.android.core.AndroidOptionsInitializer.init(AndroidOptionsInitializer.java:103)
09-11 17:38:03.900  6108  6108 E AndroidRuntime: 	at io.sentry.android.core.AndroidOptionsInitializer.init(AndroidOptionsInitializer.java:69)
09-11 17:38:03.900  6108  6108 E AndroidRuntime: 	at io.sentry.android.core.AndroidOptionsInitializer.init(AndroidOptionsInitializer.java:53)
09-11 17:38:03.900  6108  6108 E AndroidRuntime: 	at io.sentry.android.core.SentryAndroid.lambda$init$1(SentryAndroid.java:74)
09-11 17:38:03.900  6108  6108 E AndroidRuntime: 	at io.sentry.android.core.-$$Lambda$SentryAndroid$g99QaqNwcEFxja-3CMUNWZ36_g4.configure(Unknown Source:8)
09-11 17:38:03.900  6108  6108 E AndroidRuntime: 	at io.sentry.Sentry.init(Sentry.java:116)
[...]

PR #1718, opened 2021-09-15
Check at runtime if AndroidX.Core is available

marandaneto:
## :scroll: Description
Check at runtime if AndroidX.Core is available

Closes https://github.com/getsentry/sentry-java/issues/1711

running without androidxcore/unit test

## :pencil: Checklist

- [X] No breaking changes


marandaneto:
docs PR https://github.com/getsentry/sentry-docs/pull/4150

sentry-android-core/src/main/java/io/sentry/android/core/LoadClass.java
```diff
@@ -2,8 +2,16 @@
 
 import org.jetbrains.annotations.NotNull;
 
-final class LoadClass implements ILoadClass {
-  @Override
+/** An Adapter for making Class.forName testable */
+final class LoadClass {
```
maciejwalkowiak, line 6: Following our other "util" classes, shouldn't it contain a static method?

maciejwalkowiak (approved):
LGTM

sentry-android-core/src/main/java/io/sentry/android/core/LoadClass.java
romtsn: would need to use powermock to test it then, right?

maciejwalkowiak: Mockito supports mocking static methods too: https://www.baeldung.com/mockito-mock-static-methods

But, perhaps its overengineering - having this method as non-static does not hurt.

marandaneto: yeah, it'd be harder to mock the static methods, but it could, the class already exists so I will keep it like this

romtsn (approved):

bruno-garcia (approved):
This is great. I'm sure you were resistant to do the runtime check with reflection but I appreciate as it'll help the unity use case and others.

PR #1728, opened 2021-09-16
Feat: Allow setting proguard via Options and/or external resources

### LINKED ISSUE #1236: Allow setting `proguard` via Options and/or external resources
A Java App (non Android-probably Desktop) may be obfuscated using the Proguard tooling and right now there's no easy way to set up the `proguard` id if not using the `beforeSend` callback.
Ideally, this would be possible via SentryOptions or external resources.

See on Android: 
https://github.com/getsentry/sentry-java/blob/f6bca67bde00b6b036c41a15c5a4c2f150b34080/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java#L239-L244

on Android the Sentry Gradle plugin uploads the mapping file via sentry-cli and injects its `id` in the assets folder of the App which is packaged in the final APK.
At runtime, the SDK reads this file and sets the `id` to the debug images.

maciejwalkowiak:
## :scroll: Description

Allow setting proguard via Options and/or external resources. 

UUIDs can be set on options either programmatically or using system properties, environment variables or `sentry.properties` file - same way as any other external property.

Fixes #1236

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
@@ -10,7 +10,6 @@ import android.content.IntentFilter;
 import android.content.pm.ApplicationInfo;
 import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
-import android.content.res.AssetManager;
 import android.os.BatteryManager;
 import android.os.Build;
 import android.os.Environment;
@@ -30,8 +29,6 @@ import io.sentry.android.core.util.DeviceOrientations;
 import io.sentry.android.core.util.MainThreadChecker;
 import io.sentry.android.core.util.RootChecker;
 import io.sentry.protocol.App;
-import io.sentry.protocol.DebugImage;
-import io.sentry.protocol.DebugMeta;
 import io.sentry.protocol.Device;
 import io.sentry.protocol.OperatingSystem;
 import io.sentry.protocol.SentryThread;
@@ -39,21 +36,15 @@ import io.sentry.protocol.SentryTransaction;
 import io.sentry.protocol.User;
 import io.sentry.util.ApplyScopeUtils;
 import io.sentry.util.Objects;
-import java.io.BufferedInputStream;
 import java.io.BufferedReader;
 import java.io.File;
-import java.io.FileNotFoundException;
 import java.io.FileReader;
 import java.io.IOException;
-import java.io.InputStream;
-import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Locale;
 import java.util.Map;
-import java.util.Properties;
 import java.util.TimeZone;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
@@ -106,10 +97,6 @@ final class DefaultAndroidEventProcessor implements EventProcessor {
 
   private @NotNull Map<String, Object> loadContextData() {
     Map<String, Object> map = new HashMap<>();
-    String[] proguardUUIDs = getProguardUUIDs();
-    if (proguardUUIDs != null) {
-      map.put(PROGUARD_UUID, proguardUUIDs);
-    }
 
     map.put(ROOTED, rootChecker.isDeviceRooted());
 
@@ -138,7 +125,6 @@ final class DefaultAndroidEventProcessor implements EventProcessor {
       // enriched on restart, so non static data might be wrong, eg lowMemory or availMem will
       // be different if the App. crashes because of OOM.
       processNonCachedEvent(event);
-      mergeDebugImages(event);
       setThreads(event);
     }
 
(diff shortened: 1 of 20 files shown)


sentry/src/main/java/io/sentry/MainEventProcessor.java
```diff
@@ -72,6 +75,21 @@
     return event;
   }
 
+  private void setDebugMeta(final @NotNull SentryEvent event) {
+    if (event.getDebugMeta() == null && !options.getProguardUuids().isEmpty()) {
+      final DebugMeta debugMeta = new DebugMeta();
+      final List<DebugImage> images = new ArrayList<>();
+      for (final String uuid : options.getProguardUuids()) {
+        final DebugImage debugImage = new DebugImage();
+        debugImage.setType("proguard");
```
marandaneto, line 84: lets create a constant for the String, so we reuse it here https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java#L289

sentry/src/main/java/io/sentry/MainEventProcessor.java
```diff
@@ -72,6 +75,21 @@
     return event;
   }
 
+  private void setDebugMeta(final @NotNull SentryEvent event) {
+    if (event.getDebugMeta() == null && !options.getProguardUuids().isEmpty()) {
```
marandaneto, line 79: let's remove this check `event.getDebugMeta()` and instantiate if its null when `getProguardUuids` isn't empty, the reason is, an Android event may have native images, so it's not gonna be null but still has to be appended

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -336,6 +339,9 @@
     for (final String tracingOrigin : propertiesProvider.getList("tracing-origins")) {
       options.addTracingOrigin(tracingOrigin);
     }
+    for (final String proguardUuids : propertiesProvider.getList("proguard-uuids")) {
+      options.addProguardUuid(proguardUuids);
+    }
```
marandaneto, line 344: let's also add support for Android via `ManifestMetadataReader`

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -336,6 +339,9 @@
     for (final String tracingOrigin : propertiesProvider.getList("tracing-origins")) {
       options.addTracingOrigin(tracingOrigin);
     }
+    for (final String proguardUuids : propertiesProvider.getList("proguard-uuids")) {
```
marandaneto, line 342: I'm not sure that an App can have more than 1 proguard-uuid, afaik the final App always output a single mapping file, so it'd not make sense to be a list I guess

maciejwalkowiak: All examples, including Android one from you has UUIDs represented as an array in #1236. I've never used Proguard.

marandaneto: @maciejwalkowiak it's because the `DebugMeta#images` is a list, since it's also used for debug images from native code, but proguard mapping file per see is a single one afaik.

We always add a single proguard id to the images list though.

sentry/src/main/java/io/sentry/SentryOptions.java
maciejwalkowiak: Done

sentry/src/main/java/io/sentry/MainEventProcessor.java
maciejwalkowiak: 👍

sentry/src/main/java/io/sentry/MainEventProcessor.java
maciejwalkowiak: 👍

sentry/src/main/java/io/sentry/SentryOptions.java
maciejwalkowiak: Changed from `proguard-uuids` (plural) to `proguard-uuid` (singular). There is just one string now.

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -286,7 +286,7 @@ private void mergeDebugImages(final @NotNull SentryEvent event) {
 
     for (String item : proguardUUIDs) {
       DebugImage debugImage = new DebugImage();
-      debugImage.setType("proguard");
+      debugImage.setType(DebugImage.PROGUARD);
```
marandaneto, line 289: since `MainEventProcessor` now applies the uuid to the event reading from the options, what's about setting the proguardUUID here to the options and let the MainEventProcessor create the DebugImage?
we only do it if the options isn't set though

marandaneto: the other thing to consider is, event processor runs for every event, reading the uuid is a 1-time thing, so maybe it'd also make sense to move this logic from here to somewhere else eg `AndroidOptionsInitializer`, the downside is that we do IO in the main thread, maybe we keep this refactor for another PR, we just need to tackle the comment above

sentry/src/main/java/io/sentry/MainEventProcessor.java
```diff
@@ -72,6 +75,27 @@
...
+    if (options.getProguardUuid() != null) {
+      DebugMeta debugMeta = event.getDebugMeta();
+
+      if (debugMeta == null) {
+        debugMeta = new DebugMeta();
+        debugMeta.setImages(new ArrayList<>());
+      } else if (debugMeta.getImages() == null) {
+        debugMeta.setImages(new ArrayList<>());
+      }
+      List<DebugImage> images = debugMeta.getImages();
```
marandaneto, line 88: could be its own `if` so we don't duplicate the `debugMeta.setImages(new ArrayList<>());` part.

eg
```
if (debugMeta == null) {
	debugMeta = new DebugMeta();
}
if (debugMeta.getImages() == null) {
	debugMeta.setImages(new ArrayList<>());
}
```

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
maciejwalkowiak: Here it reads a list of uuids (from all the resources), so it cannot be set on the options, or we change it here to single UUID.

marandaneto: we can get just a single id, there's even a comment about that.

https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java#L911-L912

maciejwalkowiak:
@marandaneto take a look please if reading order makes sense. Currently first the Proguard UUID is read from the manifest, if not found from `sentry-debug-meta.properties`.

sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
```diff
@@ -106,10 +97,6 @@ public DefaultAndroidEventProcessor(
 
   private @NotNull Map<String, Object> loadContextData() {
     Map<String, Object> map = new HashMap<>();
-    String[] proguardUUIDs = getProguardUUIDs();
-    if (proguardUUIDs != null) {
-      map.put(PROGUARD_UUID, proguardUUIDs);
```
marandaneto, line 111: var `PROGUARD_UUID` can be removed from this class now

maciejwalkowiak: Thanks, missed that.

marandaneto (approved):

PR #1737, opened 2021-09-23
Use `float` instead of `Date` for protocol types for higher precision.

### LINKED ISSUE #1689: Use `float` instead of `Date` for protocol types for higher precision
* If possible add an overload to with `float` but keep the `Date` one
* add backward compatibility when deserializing payloads

The goal is to have a higher precision in the performance product in particular. So types such as Span and Transaction should have `float` but it's less of a concern to something like Breadcrumb

maciejwalkowiak:
## :scroll: Description

Use `float` instead of `Date` for protocol types for higher precision.

To retrieve more precise timestamps, we are using `System#nanoTime` as it's the only method available both on Android and JVM. 

When Span is created with given start timestamp (for example with Android app start time), we are not able to measure high precision timestamp, we fall back then to regular timestamp with milliseconds precision.

How it looks in UI:
image

Fixes #1689

Unit & integration tests.

## :pencil: Checklist

- [x] No breaking changes


sentry/src/main/java/io/sentry/Span.java
```diff
@@ -72,15 +88,21 @@ public Span(
...
+      this.startNanos = System.nanoTime();
+    }
   }
 
   public @NotNull Date getStartTimestamp() {
     return startTimestamp;
   }
 
-  public @Nullable Date getTimestamp() {
+  public @Nullable Double getTimestamp() {
```
bruno-garcia, line 105: Is there a way to make this change without a breaking change?
Like adding another `timestampHigherPrecision()` and leave this one as `Date timestamp()` and have it return a `Date` from the float for example?

We'd need to bump a major to release this change

bruno-garcia (commented):
Looks so much better in the UI 💯

sentry/src/main/java/io/sentry/Span.java
maciejwalkowiak: Definitely there is a way, the question is if it's worth it considering this class is marked as internal. `getTimestamp` is never used by the user code as it's not exposed with `ISpan`.

marandaneto:
btw: `System#nanoTime` 

>  This clock is guaranteed
>  *     to be monotonic, and is suitable for interval timing when the
>  *     interval does not span device sleep

that means when the device is in sleep mode, the system does not increase the timer, is it possible that it'd cause off-by-1 sort of problems?

sentry/src/main/java/io/sentry/adapters/DoubleDeserializerAdapter.java
```diff
@@ -0,0 +1,50 @@
...
+import io.sentry.SentryLevel;
+import io.sentry.SentryOptions;
+import io.sentry.util.Objects;
+import java.lang.reflect.Type;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+@ApiStatus.Internal
+public final class DoubleDeserializerAdapter implements JsonDeserializer<Double> {
```
marandaneto, line 17: does that mean that this class and `DoubleSerializerAdapter` will be triggered for every Double in the protocol classes and if its not related to the Date/Double changes?

sentry/src/main/java/io/sentry/adapters/DoubleSerializerAdapter.java
```diff
@@ -0,0 +1,39 @@
...
+
+  @Override
+  public @Nullable JsonElement serialize(
+      final @Nullable Double src,
+      final @NotNull Type typeOfSrc,
+      final @NotNull JsonSerializationContext context) {
+    try {
+      return src == null
+          ? null
+          : new JsonPrimitive(BigDecimal.valueOf(src).setScale(6, RoundingMode.DOWN));
```
marandaneto, line 33: if we try to serialize a Double that is added by setExtra or setContexts (not related to the Date/Double changes), is it gonna try to scale down?

sentry/src/main/java/io/sentry/adapters/DoubleDeserializerAdapter.java
maciejwalkowiak: Yes, and I think there's no easy way to set up an adapter only for fields related to one type - we should I think have a separate serializer for `SentryTransaction` and `SentrySpan`. Perhaps its better to wait for the non-gson serializer to be merged till we merge this one?

sentry/src/main/java/io/sentry/adapters/DoubleSerializerAdapter.java
maciejwalkowiak: Good spot. Look comment [above](https://github.com/getsentry/sentry-java/pull/1737#discussion_r716774333)

maciejwalkowiak:
> that means when the device is in sleep mode, the system does not increase the timer, is it possible that it'd cause off-by-1 sort of problems?

Are our transactions meant to span over the system sleep? If so, we should at least check if the `endTime` - `startTime` isn't a negative number.

sentry/src/main/java/io/sentry/adapters/DoubleDeserializerAdapter.java
marandaneto: yeah I'd say so, so we can target the ser/deser correctly

marandaneto:
> > that means when the device is in sleep mode, the system does not increase the timer, is it possible that it'd cause off-by-1 sort of problems?
> 
> Are our transactions meant to span over the system sleep? If so, we should at least check if the `endTime` - `startTime` isn't a negative number.

good question and I don't know the answer right now, have to check if it makes sense or not

marandaneto:
should we target the v6 branch then?

maciejwalkowiak:
Yes but this PR is anyway on hold until we have JSON serialization merged.

bruno-garcia:
Gson PR merged into branch `6.x.x`. Lets get this one merged in there too? We'll release an alpha soon.

maciejwalkowiak:
6.x.x is now up to date with main, this PR is now pointing to 6.x.x. Ready for the final review.

marandaneto:
@romtsn would you like to have a look?

romtsn:
@marandaneto checking 👍

romtsn (approved):
LGTM

marandaneto (approved):

PR #1759, opened 2021-10-11
change put() to putIfAbsent()

XinyuLiu5566:
- Problem

You are using a ConcurrentHashMap, but your usage of get() and put() may not be thread-safe at lines: 246, 249, and 250. Two threads can perform this same check at the same time and one thread can overwrite the value written by the other thread.

- Fix

Consider replacing put() with putIfAbsent() to help prevent accidental overwriting. putIfAbsent() puts the value only if the ConcurrentHashMap does not contain the key and therefore avoids overwriting the value written there by the other thread's putIfAbsent().

- More info

putIfAbsent() returns null if the value did not exist and returns the value in the map if one already exists.

### DIFF
--- a/sentry/src/main/java/io/sentry/transport/RateLimiter.java
+++ b/sentry/src/main/java/io/sentry/transport/RateLimiter.java
@@ -247,7 +247,7 @@ public final class RateLimiter {
 
     // only overwrite its previous date if the limit is even longer
     if (oldDate == null || date.after(oldDate)) {
-      sentryRetryAfterLimit.put(dataCategory, date);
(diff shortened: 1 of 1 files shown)


marandaneto:
`putIfAbsent` isn't thread-safe either and also because Android does not support `putIfAbsent`, this code is shared with the Android SDK (only API >= 24 https://android-review.googlesource.com/c/platform/frameworks/support/+/772608).

`putIfAbsent` would not solve the described problem, hence I will close the issue.

Worth noting that by default the SDK executes 1 thread per time, so this would not happen anyway.

PR #1777, opened 2021-10-22
Feat: Add `graphql-java` instrumentation.

### LINKED ISSUE #1755: GraphQL integration for server side
Similar to our Apollo GraphQL integration but for server-side, using the libs:

com.graphql-java:graphql-java and com.graphql-java:graphql-java-spring-boot-starter-webmvc

https://www.graphql-java.com/tutorials/getting-started-with-spring-boot/

maciejwalkowiak:
## :scroll: Description

Add `graphql-java` instrumentation.

- exception handler - `SentryDataFetcherExceptionHandler` reports exception to Sentry and invokes a delegate

Sample usage with Netflix DGS:

```java
@Bean
SentryDataFetcherExceptionHandler sentryDataFetcherExceptionHandler() {
  // delegate to default Netflix DGS exception handler
  return new SentryDataFetcherExceptionHandler(new DefaultDataFetcherExceptionHandler());
}
```

- tracing - `SentryInstrumentation` creates spans around data fetchers. Since there can be more than a single query executed within a single HTTP request, the transaction name is `POST /graphql` and resolving each query/field is reported as separate span.

It is advised to turn on `sentry.max-request-body-size` reporting, to get a complete GraphQL query in request parameters.

image

Fixes #1755

Both [Netflix DGS](https://netflix.github.io/dgs) and [Spring GraphQL](https://spring.io/projects/spring-graphql) are based on [graphql-java](https://github.com/graphql-java/graphql-java). Changes introduced in this PR will work for both solutions as well as using `graphql-java` directly.

Unit tests.

## :pencil: Checklist

- [ ] No breaking changes

## :crystal_ball: Next steps

- [x] Add error handler
- [x] Create sample project

Optionally, once [Spring GraphQL](https://spring.io/projects/spring-graphql) project reaches GA, add corresponding sample.


sentry-samples/sentry-samples-netflix-dgs/src/main/java/io/sentry/samples/netflix/dgs/ShowsDatafetcher.java
```diff
@@ -0,0 +1,23 @@
...
+import java.util.Random;
+
+@DgsComponent
+public class ShowsDatafetcher {
+
+  @DgsQuery(field = DgsConstants.QUERY.Shows)
+  public List<Show> shows() throws InterruptedException {
+    Thread.sleep(new Random().nextInt(500));
+
+    //        return Arrays.asList(
```
bruno-garcia, line 17: deadcode to be removed?

maciejwalkowiak: I added another data fetcher that ends with success. Thanks for pointing it out.

maciejwalkowiak:
Few improvements to instrumentation:

- do not finish parent span
- handle exceptions thrown by data fetchers
- set parent span status if any of child spans fails

So now, if any of the child spans fails - meaning that there are errors present in GraphQL response, we set the parent span status - which in most cases will be a transaction - to `INTERNAL_ERROR`.

sentry-graphql-java/src/main/java/io/sentry/graphql/java/SentryDataFetcherExceptionHandler.java
```diff
@@ -0,0 +1,36 @@
+package io.sentry.graphql.java;
```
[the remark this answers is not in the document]
marandaneto: still called `sentry-graphql-java`, should we rename it to `sentry-graphql`? we only add the `java` suffix to Android modules

sentry-graphql-java/src/main/java/io/sentry/graphql/SentryInstrumentation.java
```diff
@@ -0,0 +1,130 @@
...
+
+  @Override
+  public @NotNull InstrumentationState createState() {
+    return new TracingState();
+  }
+
+  @Override
+  public @NotNull InstrumentationContext<ExecutionResult> beginExecution(
+      final @NotNull InstrumentationExecutionParameters parameters) {
+    TracingState tracingState = parameters.getInstrumentationState();
```
marandaneto, line 43:
```suggestion
    final TracingState tracingState = parameters.getInstrumentationState();
```

sentry-graphql-java/src/main/java/io/sentry/graphql/SentryInstrumentation.java
```diff
@@ -0,0 +1,130 @@
...
+        return dataFetcher.get(environment);
+      }
+    };
+  }
+
+  @Override
+  public @NotNull CompletableFuture<ExecutionResult> instrumentExecutionResult(
+      final @NotNull ExecutionResult executionResult,
+      final @NotNull InstrumentationExecutionParameters parameters) {
+    TracingState tracingState = parameters.getInstrumentationState();
```
marandaneto, line 95:
```suggestion
    final TracingState tracingState = parameters.getInstrumentationState();
```

sentry-graphql-java/src/main/java/io/sentry/graphql/SentryDataFetcherExceptionHandler.java
```diff
@@ -0,0 +1,36 @@
...
+
+  public SentryDataFetcherExceptionHandler(final @NotNull DataFetcherExceptionHandler delegate) {
+    this(HubAdapter.getInstance(), delegate);
+  }
+
+  @Override
+  @SuppressWarnings("deprecation")
+  public DataFetcherExceptionHandlerResult onException(
+      final @NotNull DataFetcherExceptionHandlerParameters handlerParameters) {
+    hub.captureException(handlerParameters.getException());
```
marandaneto, line 33: should we capture every exception? should we offer maybe an option to turn it off then?
I remember not doing it on SentryOkHttpInterceptor because the user might want to do it themselves based on their own logic.

cc @bruno-garcia

sentry-graphql-java/src/main/java/io/sentry/graphql/SentryInstrumentation.java
```diff
@@ -0,0 +1,130 @@
...
+    final ISpan transaction = tracingState.getTransaction();
+
+    final List<GraphQLError> errors = executionResult.getErrors();
+    if (transaction != null && errors != null && !errors.isEmpty()) {
+      transaction.setStatus(SpanStatus.INTERNAL_ERROR);
```
marandaneto, line 100: that is different than other integrations but might make sense, what do you think @bruno-garcia ?
just wondering if a parent has 10 children, 9 of them were successful and one of them failed, is the parent failed or not?

sentry-graphql-java/src/main/java/io/sentry/graphql/SentryDataFetcherExceptionHandler.java
maciejwalkowiak: Perhaps you mean `BeforeSpanCallback` in OkHttp interceptor? We can add it here too.

marandaneto: thats about capturing an exception and not really the span callback, but well remembered, the before span callback could also be a follow up tho

marandaneto:
missing the craft configuration, release-registry and docs later on :)

the module name is still with `-java`

sentry-graphql-java/src/main/java/io/sentry/graphql/java/SentryDataFetcherExceptionHandler.java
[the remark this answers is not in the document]
maciejwalkowiak: the library we integrate with is called `graphql-java` - i do think it should stay as the module name.

sentry-graphql-java/src/main/java/io/sentry/graphql/SentryDataFetcherExceptionHandler.java
bruno-garcia: Here's where `hints` could be helpful.
As we would capture the exception with the `handlersParameters` as a hint. So if users want to filter some out, they could access that data in the beforeSend or eventProcessor and do they customization/dropping there

maciejwalkowiak: `hint` added.

sentry-graphql-java/src/main/java/io/sentry/graphql/SentryInstrumentation.java
```diff
@@ -0,0 +1,130 @@
...
+  @Override
+  public @NotNull InstrumentationState createState() {
+    return new TracingState();
+  }
+
+  @Override
+  public @NotNull InstrumentationContext<ExecutionResult> beginExecution(
+      final @NotNull InstrumentationExecutionParameters parameters) {
+    final TracingState tracingState = parameters.getInstrumentationState();
+    tracingState.setTransaction(hub.getSpan());
```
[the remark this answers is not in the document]
maciejwalkowiak: there's nothing really interesting happening here to start a span i believe.

sentry-graphql-java/src/main/java/io/sentry/graphql/SentryInstrumentation.java
maciejwalkowiak: As far as I found out, this method is executed at the end when the execution is over. There can be more than one query sent with single GraphQL request - some of them can fail and be returned in "errors" part of the response. I don't think marking such request that has partially failed as a success reflects the reality.

Perhaps there is already a GraphQL instrumentation in other SDKs and we can just mimic its behavior?

marandaneto: the way how I see in all the other SDKs is, if you are able to detect the error of the span that caused the issue, you can indeed mark it as `INTERNAL_ERROR` as we do on the `instrumentDataFetcher` method, but it's parent that comes from `hub.getSpan()` could be even a manually created span by the user that is controlled by them, its up to them to mark as `errored` or `successful, not us, since we `rethrow` the exception, they should do it manually, and if its an automatically created by some other integrations, the exception is gonna bubble up and be marked as `INTERNAL_ERROR` anyway, so I'd avoid messing up with the parent span status.

marandaneto:
let's discuss https://github.com/getsentry/sentry-java/pull/1777/files#r742653695 today and approve/merge, nothing but this comment blocks me to approve, merge and release, thanks for doing this @maciejwalkowiak 

@romtsn wanna have a look?

sentry-graphql-java/src/main/java/io/sentry/graphql/SentryInstrumentation.java
```diff
@@ -0,0 +1,167 @@
...
+          final Object result = dataFetcher.get(environment);
+          if (result instanceof CompletableFuture) {
+            ((CompletableFuture<?>) result)
+                .whenComplete(
+                    (r, ex) -> {
+                      if (ex != null) {
+                        span.setThrowable(ex);
+                        span.setStatus(SpanStatus.INTERNAL_ERROR);
+                      }
+                      finish(span, environment, r);
```
romtsn, line 81: Can we run into some concurrency issues since it's an async callback here?

romtsn (approved):

sentry-graphql-java/src/main/java/io/sentry/graphql/SentryInstrumentation.java
```diff
@@ -0,0 +1,167 @@
...
+    if (beforeSpan != null) {
+      final ISpan newSpan = beforeSpan.execute(span, environment, result);
+      if (newSpan == null) {
+        // span is dropped
+        span.getSpanContext().setSampled(false);
+      } else {
+        newSpan.finish();
+      }
+    } else {
+      span.finish();
```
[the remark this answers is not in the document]
maciejwalkowiak: Thats something we agreed on, that by default in successful scenario we don't pass status. It's the same in all integrations.

sentry-graphql-java/src/main/java/io/sentry/graphql/SentryInstrumentation.java
maciejwalkowiak: No, it's made thread safe.

sentry-graphql-java/src/main/java/io/sentry/graphql/SentryInstrumentation.java
maciejwalkowiak: Changing status on parent span removed.

sentry-graphql-java/src/main/java/io/sentry/graphql/java/SentryDataFetcherExceptionHandler.java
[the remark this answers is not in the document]
marandaneto: @bruno-garcia are u fine with it? I recall you didn't want to add such sufixes

sentry-graphql-java/src/main/java/io/sentry/graphql/SentryInstrumentation.java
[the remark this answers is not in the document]
marandaneto: so the difference between this one and the other integrations is that the other integrations always try to get the status from the HTTP response code, but here we always leave it without status.
can we do the same here in case the request is successful?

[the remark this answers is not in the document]
marandaneto: it is, https://github.com/getsentry/develop/issues/307
there was the discussion but none of us worked on the docs yet, PR welcomed :P

marandaneto (approved):
LGTM

sentry-graphql-java/src/main/java/io/sentry/graphql/java/SentryDataFetcherExceptionHandler.java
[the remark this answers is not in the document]
bruno-garcia: I don't think it must match in all cases the package name. The proposal from manoel sounds better to me unless folks have strong options not to take that route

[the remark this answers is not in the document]
marandaneto: with or without fine with me, I'd rather keep it without -java if I'd have to vote.
@romtsn ?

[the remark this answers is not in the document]
romtsn: same maybe call it `sentry-graphql-server`? :D fine with everything

sentry-graphql-java/src/main/java/io/sentry/graphql/SentryInstrumentation.java
```diff
@@ -0,0 +1,153 @@
...
+      final @NotNull InstrumentationFieldFetchParameters parameters) {
+    // We only care about user code
+    if (parameters.isTrivialDataFetcher()) {
+      return dataFetcher;
+    }
+
+    return environment -> {
+      final TracingState tracingState = parameters.getInstrumentationState();
+      final ISpan transaction = tracingState.getTransaction();
+      if (transaction != null) {
```
koenpunt, line 67: I'm trying to use the `SentryInstrumentation` in our spring-graphql application (using Spring Boot), but the transaction here is always `null`. Is it necessary to manually start a transaction? Or how can I get this to work?

marandaneto: see the Docs PR https://github.com/getsentry/sentry-docs/pull/4385

koenpunt: Thanks, I didn't see that PR yet. But unfortunately it doesn't solve the issue; performance tracing is set up, but apparently a transaction isn't started before graphql execution. 

For now I think I've mitigated it by adding a `WebInterceptor` (a spring-graphql concept), that starts a transaction;

```kt
class SentryWebInterceptor : WebInterceptor {
    override fun intercept(webInput: WebInput, chain: WebInterceptorChain): Mono<WebOutput> {
        val span = Sentry.getSpan() ?: Sentry.startTransaction("GraphQL", webInput.operationName ?: "none", true)

        return chain.next(webInput).doFinally {
            span.finish()
        }
    }
}
```

According to the PR description "the transaction name is `POST /graphql`", but this doesn't seem to be defined explicitly, so could it be that the assumption is made that the the http/network layer already starts the transaction, but that somehow with spring-graphql this doesn't happen?

marandaneto: the graphql integration only creates spans if there's an active transaction, so ideally you'd use it along with an integration that has the ability to create a transaction, eg https://docs.sentry.io/platforms/java/guides/spring-boot/performance/instrumentation/automatic-instrumentation/
or just create a transaction yourself

koenpunt: Just found that out too, but unfortunately there's no auto instrumentation available for applications using webflux, so I guess I have to implement what's in [`SentryTracingFilter`](https://github.com/getsentry/sentry-java/blob/main/sentry-spring/src/main/java/io/sentry/spring/tracing/SentryTracingFilter.java), but then reactive.

marandaneto: good catch, https://github.com/getsentry/sentry-java/issues/1807
no priorities for now, but if you feel adding support via PR, happy to guide :)

PR #1791, opened 2021-11-02
Release 6.0.0

bruno-garcia:
Draft PR so we can keep up with 6.0.0 and `main` diff.

The idea is to merge this branch into `main` once we GA.

_#skip-changelog_

- [ ] Document the interface for custom serialization for: https://docs.sentry.io/platforms/java/enriching-events/context/

PR #1793, opened 2021-11-04
Update gson to 2.8.9 for CWE-502

CodemonkeyOSS:
## :scroll: Description
CWE-502 is a CVSS 7.7 scored vulnerability in the gson library that was fixed in gson 2.8.9.

Some users of the Sentry library are subject to security gates that will fail on the presence of this vuln. Instead of just making one-off changes, I am contributing back ❤️ 

Tested it on a local project and confirmed that Sentry event submission did not suffer any new failures.

## :pencil: Checklist

- [ x] I reviewed the submitted code
- [ x] I updated the docs if needed
- [ x] No breaking changes


marandaneto:
since we've dropped the Gson dependency on 6.0, I'd just wait a bit longer, doing this now would break older clients
See https://github.com/getsentry/sentry-java/pull/1791

Issue #1798, opened 2021-11-08
Sentry Performance crash: Caused by java.lang.NullPointerException: Attempt to invoke virtual method 'int java.util.ArrayList.size()' on a null object reference


dmytroKarataiev:
_Platform:_
- [x]  Android -> If yes, which Device API (and compileSdkVersion/targetSdkVersion/Build tools) version?

compileSdkVersion 30
targetSdkVersion 30
Device API: Galaxy S8, Android 9

_IDE:_
- [x]  Android Studio -> If yes, which version?
4.3

_Build system:_
- [x]  Gradle -> If yes, which version?
7.0.1

_Android Gradle Plugin:_
- [x]  Yes -> If yes, which version?
4.2.2

_Sentry Android Gradle Plugin:_
- [x]  Yes -> If yes, which version?
2.1.5

_Proguard/R8:_
- [x]  Enabled
- [x]  Disabled

_Platform installed with:_
- [x] Maven Central

The version of the SDK:
**5.3.0**

---
I have the following issue:

```
Caused by java.lang.NullPointerException: Attempt to invoke virtual method 'int java.util.ArrayList.size()' on a null object reference
       at android.view.View.findFrameMetricsObserver(View.java:7279)
       at android.view.View.removeFrameMetricsListener(View.java:7250)
       at android.view.Window.removeOnFrameMetricsAvailableListener(Window.java:886)
       at androidx.core.app.FrameMetricsAggregator$FrameMetricsApi24Impl.remove(FrameMetricsAggregator.java:433)
       at androidx.core.app.FrameMetricsAggregator.remove(FrameMetricsAggregator.java:243)
       at io.sentry.android.core.ActivityFramesTracker.setMetrics(ActivityFramesTracker.java:69)
       at io.sentry.android.core.ActivityLifecycleIntegration.lambda$startTracing$0$ActivityLifecycleIntegration(ActivityLifecycleIntegration.java:152)
       at io.sentry.android.core.-$$Lambda$ActivityLifecycleIntegration$OZU_BVfUTV1Z-vCLPnw148HJdoA.execute(:4)
       at io.sentry.SentryTracer.finish(SentryTracer.java:241)
       at io.sentry.android.core.ActivityLifecycleIntegration.finishTransaction(ActivityLifecycleIntegration.java:223)
       at io.sentry.android.core.ActivityLifecycleIntegration.stopPreviousTransactions(ActivityLifecycleIntegration.java:128)
[...]

PR #1799, opened 2021-11-10
Fix: ActivityFramesTracker does not throw if Activity has no observers

marandaneto:
## :scroll: Description
Fix: ActivityFramesTracker does not throw if Activity has no observers

Closes https://github.com/getsentry/sentry-java/issues/1798

## :pencil: Checklist

- [X] No breaking changes

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesTracker.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesTracker.java
@@ -67,9 +67,11 @@ final class ActivityFramesTracker {
     SparseIntArray[] framesRates = null;
     try {
       framesRates = frameMetricsAggregator.remove(activity);
-    } catch (IllegalArgumentException ignored) {
-      // it throws when attempt to remove OnFrameMetricsAvailableListener that was never added.
-      // there's no contains method
+    } catch (Exception ignored) {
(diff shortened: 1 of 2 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesTracker.java
```diff
@@ -67,9 +67,11 @@ synchronized void setMetrics(final @NotNull Activity activity, final @NotNull Se
     SparseIntArray[] framesRates = null;
     try {
       framesRates = frameMetricsAggregator.remove(activity);
-    } catch (IllegalArgumentException ignored) {
-      // it throws when attempt to remove OnFrameMetricsAvailableListener that was never added.
-      // there's no contains method
+    } catch (Exception ignored) {
```
romtsn, line 70: hm, should we rather do a multi-catch? cause now it will swallow all other exceptions as well?

romtsn (approved):

sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesTracker.java
marandaneto: that's the goal, to avoid problems with the further version compatibility, since it's the 2nd, I'd rather prefer to debug once slow/frozen frames are missing rather than crashing the final app.

PR #1800, opened 2021-11-11
Fix: Do not crash when event processors throw a lower level Throwable class

marandaneto:
## :scroll: Description
Fix: Do not crash when event processors throw a lower level Throwable class

https://github.com/getsentry/sentry-unity/issues/398

running it on newer Unity versions

## :pencil: Checklist

- [X] No breaking changes


bitsandfoxes (approved):

romtsn (approved):
LGTM, but I have no idea what I'm doing anyway :D

sentry/src/main/java/io/sentry/SentryClient.java
```diff
@@ -235,7 +235,7 @@ private SentryTransaction processTransaction(
     for (final EventProcessor processor : eventProcessors) {
       try {
         transaction = processor.process(transaction, hint);
-      } catch (Exception e) {
```
bruno-garcia, line 238: Do we have other places in the codebase we should make this change?

sentry-android-core/src/main/java/io/sentry/android/core/PerformanceAndroidEventProcessor.java
```diff
@@ -32,6 +33,22 @@
         Objects.requireNonNull(activityFramesTracker, "ActivityFramesTracker is required");
   }
 
+  /**
+   * Returns the event itself
+   *
+   * @param event the SentryEvent the SentryEvent
+   * @param hint the Hint the Hint
```
lucas-zimerman, line 40: confusing 🤪

sentry/src/main/java/io/sentry/SentryClient.java
marandaneto: maybe yes...

sentry-android-core/src/main/java/io/sentry/android/core/PerformanceAndroidEventProcessor.java
marandaneto: indeed, it was auto-generated, didn't realize the confusion :(

Issue #1806, opened 2021-11-15
Replace SDKs `catch (Exception e)` to `catch (Throwable e)`


marandaneto:
SDK should not crash the finals App, but rather swallow and log it.
this would avoid issues such as https://github.com/getsentry/sentry-java/pull/1800

Context: https://github.com/getsentry/sentry-java/pull/1800#discussion_r747575295

philipphofmann:
Make sure to log out when we swallow a Throwable. We still need to find out how to log out.

bruno-garcia:
The PR #1812 partially addresses this. See PR description for details.
Goal: Changes all Java catch `Exception` to `Throwable` so we avoid crashes on Unity that should have been suppressed.

For this issue I suggest we add a linter. We we avoid adding `catch Exception` again in the future. The linter should warn that throwable should be used to suppress all errors instead.

PR #1812, opened 2021-11-17
catch throwable to suppress errors in Java

bruno-garcia:
The goal is to change code locations we suppress error to avoid the SDK bubbling errors to customers. 

Firs try: I changed every match of `catch (Exception` to `catch (Throwable`. Not matching Kotlin code though, the main focus was to help avoid the case we had when debugging a crash on Unity with the Android SDK so mainly `sentry-java` and `sentry-android-core`.

Adding additional logging isn't in the scope of this PR. When it was simple I just added a debug log, but if no logger available, I added a TODO and we can have a follow up PR.

@maciejwalkowiak does this make sense on the Spring and other backend packages or should I revert?

Partially addresses #1806 (not doing the Kotlin code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Feat: Refactor OkHttp and Apollo to Kotlin functional interfaces (#1797)
 * Feat: Add secondary constructor to SentryInstrumentation (#1804)
+* Ref: catch Throwable instead of Exception to suppress internal SDK errors (#1812)
 
 ## 5.4.0
 
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesTracker.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesTracker.java
@@ -67,7 +67,7 @@ final class ActivityFramesTracker {
     SparseIntArray[] framesRates = null;
     try {
       framesRates = frameMetricsAggregator.remove(activity);
-    } catch (Exception ignored) {
+    } catch (Throwable ignored) {
       // throws IllegalArgumentException when attempting to remove OnFrameMetricsAvailableListener
       // that was never added.
       // there's no contains method.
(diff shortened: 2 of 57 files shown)


maciejwalkowiak:
I don't see how this could hurt anything server side related. Looks good to me.

marandaneto:
Missing specific catch blocks that don't use `Exception` but rather something specific eg `RuntimeException`, `InterruptedException`, `ParseException`, and many more.

Missing changes on Kotlin projects.

the PR description states that  logging isn't the focus but resolves the issue https://github.com/getsentry/sentry-java/issues/1806
The issue describes that logging isn't part of this PR so either we don't close #1806 or we open a new issue, we don't want to silently swallow a Throwable and not log out, it'd be even harder than crashing for debugging eventual issues.

Tests are not adjusted to using a `Throwable` error instead of `Exception`

bruno-garcia:
> Missing specific catch blocks that don't use `Exception` but rather something specific eg `RuntimeException`, `InterruptedException`, `ParseException`, and many more.

I wouldn't go that far. When we picked a specific exception type, the intent was to handle only that one. That implies already that any other type was supposed to bubble up so for that reason I wouldn't change those. 

> Missing changes on Kotlin projects.
> the PR description states that logging isn't the focus but resolves the issue #1806 The PR description describes that logging isn't part of this PR so either we don't close (resolve) #1806 or we open a new issue, we don't want to silently swallow a Throwable and not log out, it'd be even harder than crashing for debugging eventual issues.

Fair point. I'll remove the `Resolves` from the PR. We can keep that for adding any missing logging and a linter so we don't have `catch (Exception)` anymore in the code base.

> Tests are not adjusted to using a `Throwable` error instead of `Exception`, so we cannot guarantee that those changes don't have a side effect during code execution (at least when testing)

I'm not sure what we can test in this case. The intent of the code I changed was to suppress any unhandled error. It was done on the "wrong" exception type. I wrote many of those and as a C# dev I just wrote what I knew and missed the point `Throwable`s could go through my catch block so it's just a correction of those.

bruno-garcia:
Discussed on a call with @marandaneto and agreed to merge this not as resolving the inital issue but partially addressing it.

Issue #1787, opened 2021-10-29
Odd activity auto tracing duration


LouisFn:
_Platform:_
- [X]  Android -> minSdk: 24, compileSdk: 30, targetSdk: 30, buildTools: 31.0.0
- [X]  Kotlin -> 1.5.31, jvmTarget: VERSION_11
- [X]  Timber -> 5.0.1

_Build system:_
- [X]  Gradle -> 7.2

_Android Gradle Plugin: 7.0.2

The version of the SDK:
**5.2.0**

---
I have the following issue:

We get odd result with activity auto tracing. We receive many transactions with a very long duration on prod. 

<img width="1204" alt="Screenshot 2021-10-29 at 10 59 40" src="https://user-images.githubusercontent.com/54979514/139409679-ff0c67ec-d63a-451e-a260-81e5120dfb0c.png">

marandaneto:
@LouisYubo I can't really reproduce it, not sure how to proceed, any ideas on how to reproduce otherwise I'd have to close it.

marandaneto:
@LouisFn does your app have notifications? is it possible that users open your app via notification?

LouisFn:
Completely, I can reproduce it by opening the app from notification

marandaneto:
@LouisFn ok thats a good start, can you tell me the exact steps to reproduce it? or provide a minimal reproducible example? thanks

LouisFn:
![Screenshot 2021-11-16 at 14 12 16](https://user-images.githubusercontent.com/82385769/141992227-715b1689-4449-4eea-a7a2-b942a5d05610.png)

@marandaneto It seems to be because you are waiting for the child traces to finish before stoping the transaction (waitForChildren) and the child trace from FragmentLifecycleIntegration never ends.

It's due to NavDeepLinkBuilder, when the deep link is triggered, the destination and all of its parents will be on the back stack. But for its parents, only onCreate is called but not onResume. So the HomeFragment trace never finished.

You can find a reproducible example here:
https://github.com/LouisFn/SentryFrozenFrame 

BTW, I think it would be usefull to add an option to disable "waitForChildren"

marandaneto:
will have a look, thanks

marandaneto:
I can confirm that `HomeFragment` isn't calling `onFragmentResumed`

marandaneto:
@LouisFn in your use case, maybe there's a bug in your example as well, you set:

`app:startDestination="@id/navigation_fragment_home"`

[...]

PR #1813, opened 2021-11-18
Fix: Do not start fragment span if not added to the Activity

marandaneto:
## :scroll: Description
Fix: Do not start fragment span if not added to the Activity

Closes https://github.com/getsentry/sentry-java/issues/1787

running https://github.com/LouisFn/SentryFrozenFrame and repo's sample with added/visible fragments

## :pencil: Checklist

- [X] No breaking changes

### DIFF
--- a/sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
+++ b/sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
@@ -57,7 +57,11 @@ class SentryFragmentLifecycleCallbacks(
     ) {
         addBreadcrumb(fragment, "created")
 
-        startTracing(fragment)
+        // we only start the tracing for the fragment if the fragment has been added to its activity
+        // and not only to the backstack
+        if (fragment.isAdded) {
+            startTracing(fragment)
+        }
     }
 
     override fun onFragmentViewCreated(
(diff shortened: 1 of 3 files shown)


sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
```diff
@@ -57,7 +57,11 @@ class SentryFragmentLifecycleCallbacks(
     ) {
         addBreadcrumb(fragment, "created")
 
-        startTracing(fragment)
+        // we only start the tracing for the fragment if the fragment has been added to its activity
+        // and not only to the backstack
+        if (fragment.isAdded) {
```
romtsn, line 62: just curious if we should just call `startTracing` inside `onAttach`?

romtsn (approved):

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
marandaneto: `onFragmentAttached` is called even before `onFragmentCreated`

marandaneto: and also called by non added fragments

PR #1815, opened 2021-11-18
Bump to Kotlin 1.5

marandaneto:
## :scroll: Description
should this go to Branch 6.0.0 due to Kotlin compatibility? (Only for Timber tho)

## :pencil: Checklist

- [ ] No breaking changes


sentry-graphql/build.gradle.kts
```diff
@@ -16,7 +16,7 @@ configure<JavaPluginExtension> {
 
 tasks.withType<KotlinCompile>().configureEach {
     kotlinOptions.jvmTarget = JavaVersion.VERSION_1_8.toString()
-    kotlinOptions.languageVersion = Config.springKotlinCompatibleLanguageVersion
```
marandaneto, line 19: @maciejwalkowiak cant use 1.3 due to the functional interfaces

sentry-android-timber/build.gradle.kts
```diff
@@ -34,6 +33,7 @@ android {
 
     kotlinOptions {
         jvmTarget = JavaVersion.VERSION_1_8.toString()
+        kotlinOptions.languageVersion = Config.kotlinCompatibleLanguageVersion
```
marandaneto, line 36: thats a breaking change from 1.2 to 1.4, Kotlin 1.2 isn't supported on Kotlin 1.5 anymore

buildSrc/src/main/java/Config.kt
```diff
@@ -1,12 +1,13 @@
 import java.math.BigDecimal
 
 object Config {
-    val kotlinVersion = "1.4.32"
+    val kotlinVersion = "1.5.31"
```
romtsn, line 4: Hm, can't we even bump it to 1.6 because of this?
image

sentry-android-timber/build.gradle.kts
romtsn: Btw, why do we stick to the older version of TImber? Java compat.? They say in 5.0.0 that it should be binary compatible with 4.x

sentry-graphql/build.gradle.kts
maciejwalkowiak: Spring now supports Kotlin 1.5

marandaneto: but then I guess we'd need to bump Spring versions I guess?

sentry-android-timber/build.gradle.kts
marandaneto: that's about Kotlin compatibility and not Timber compatibility, we could technically upgrade Timber to v5 if that's the case, it'd still be a breaking change for users relying on Kotlin 1.2 compatibility.

buildSrc/src/main/java/Config.kt
marandaneto: because it's just been released, we cant use the 1.6 features anyway till adoption goes higher, so it's safer to be in a more compatible/adopted version.

romtsn (approved):

sentry-graphql/build.gradle.kts
marandaneto: since this is a separate issue, I will merge, but we could bump the `springKotlinCompatibleLanguageVersion` libs to 1.4 if its already supported (under the v6 branch)

PR #1821, opened 2021-11-29
Fix: Only report App start measurement for full launch on Android

marandaneto:
## :pencil: Checklist

- [ ] No breaking changes

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
@@ -1,9 +1,14 @@
 package io.sentry.android.core;
 
+import static android.app.ActivityManager.RunningAppProcessInfo.IMPORTANCE_FOREGROUND;
+
 import android.app.Activity;
+import android.app.ActivityManager;
 import android.app.Application;
+import android.content.Context;
 import android.os.Build;
 import android.os.Bundle;
+import android.os.Process;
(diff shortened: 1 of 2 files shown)


marandaneto:
docs https://github.com/getsentry/sentry-java/pull/1821

romtsn (approved):

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -64,6 +71,10 @@ public ActivityLifecycleIntegration(
     if (buildInfoProvider.getSdkInfoVersion() >= Build.VERSION_CODES.Q) {
       isAllActivityCallbacksAvailable = true;
     }
+
+    // we only track app start for processes that will show an Activity (full launch).
+    // Here we check the process importance which will tell us that.
+    foregroundImportance = isForegroundImportance(this.application);
```
romtsn, line 77: hm, just thinking if this will be called in case of the warm app start? This integration is added in `Application.onCreate`, which won't be called in case of warm start, right?

marandaneto: it's in the ctor of the integration, so it'll be called too.

romtsn: But the integration ctor itself - will it be called in case of the warm start, if the process wasn't killed?

PR #1826, opened 2021-12-03
feat: Add runtime classes for FileIOStream instrumentation

romtsn:
## :scroll: Description

* Added `SentryFileInputStream` implementation for `FileInputStream` instrumentation
* Added `SentryFileOutputStream` implementation for `FileOutputStream` instrumentation

This is a supporting PR with runtime classes for File I/O instrumentation see https://github.com/getsentry/sentry-android-gradle-plugin/pull/239

## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next steps
* Will follow up with another PR for `FileReader/Writer`
* Will update docs to show how to manually use it

### DIFF
--- a/sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
+++ b/sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
@@ -10,11 +10,12 @@ import java.io.IOException;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-class FileIOSpanManager {
+final class FileIOSpanManager {
 
   private final @Nullable ISpan currentSpan;
   private final @Nullable File file;
 
+  private @Nullable Throwable throwable = null;
   private @NotNull SpanStatus spanStatus = SpanStatus.OK;
   private long byteCount;
 
@@ -48,6 +49,7 @@ class FileIOSpanManager {
       return result;
     } catch (IOException exception) {
       spanStatus = SpanStatus.INTERNAL_ERROR;
+      throwable = exception;
       throw exception;
     }
   }
@@ -57,6 +59,7 @@ class FileIOSpanManager {
       delegate.close();
     } catch (IOException exception) {
       spanStatus = SpanStatus.INTERNAL_ERROR;
+      throwable = exception;
       throw exception;
     }
     finishSpan();
@@ -77,6 +80,7 @@ class FileIOSpanManager {
         currentSpan.setDescription(byteCountToString);
       }
       currentSpan.setData("file.size", byteCount);
+      currentSpan.setThrowable(throwable);
       currentSpan.finish(spanStatus);
     }
   }
--- a/sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
+++ b/sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
@@ -18,7 +18,7 @@ final class FileIOSpanManager {
 
   private final @Nullable ISpan currentSpan;
   private final @Nullable File file;
-  private final @NotNull IHub hub;
+  private final boolean isSendDefaultPii;
 
   private @NotNull SpanStatus spanStatus = SpanStatus.OK;
   private long byteCount;
@@ -29,10 +29,10 @@ final class FileIOSpanManager {
   }
 
   FileIOSpanManager(
-      final @Nullable ISpan currentSpan, final @Nullable File file, final @NotNull IHub hub) {
+      final @Nullable ISpan currentSpan, final @Nullable File file, final boolean isSendDefaultPii) {
     this.currentSpan = currentSpan;
     this.file = file;
-    this.hub = hub;
+    this.isSendDefaultPii = isSendDefaultPii;
   }
 
   /**
@@ -87,14 +87,13 @@ final class FileIOSpanManager {
       if (file != null) {
         final String description = file.getName() + " " + "(" + byteCountToString + ")";
         currentSpan.setDescription(description);
-        if (Platform.isAndroid() || hub.getOptions().isSendDefaultPii()) {
+        if (Platform.isAndroid() || isSendDefaultPii) {
           currentSpan.setData("file.path", file.getAbsolutePath());
         }
       } else {
         currentSpan.setDescription(byteCountToString);
       }
       currentSpan.setData("file.size", byteCount);
-      currentSpan.setThrowable(throwable);
       currentSpan.finish(spanStatus);
     }
   }
--- a/sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
+++ b/sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
@@ -1,0 +1,92 @@
+package io.sentry.instrumentation.file;
+
+import io.sentry.ISpan;
+import io.sentry.Sentry;
+import io.sentry.SpanStatus;
+import io.sentry.util.StringUtils;
+import java.io.Closeable;
+import java.io.File;
+import java.io.IOException;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+class FileIOSpanManager {
+
+  private final @Nullable ISpan currentSpan;
+  private final @Nullable File file;
+
+  private @NotNull SpanStatus spanStatus = SpanStatus.OK;
+  private long byteCount;
+
+  static @Nullable ISpan startSpan(final @NotNull String op) {
+    final ISpan parent = Sentry.getSpan();
+    return parent != null ? parent.startChild(op) : null;
+  }
+
+  FileIOSpanManager(
+    final @Nullable ISpan currentSpan,
+    final @Nullable File file
+  ) {
+    this.currentSpan = currentSpan;
+    this.file = file;
+  }
+
+  <T> T performIO(final @NotNull FileIOCallable<T> operation) throws IOException {
+    try {
+      final T result = operation.call();
+      if (result instanceof Integer) {
+        final int res = (int) result;
+        if (res != -1) {
+          byteCount += res;
+        }
+      } else if (result instanceof Long) {
+        final long res = (long) result;
+        if (res != -1) {
+          byteCount += res;
+        }
+      }
+      return result;
+    } catch (Exception exception) {
+      spanStatus = SpanStatus.INTERNAL_ERROR;
+      throw exception;
+    }
+  }
+
+  void finish(final @NotNull Closeable delegate) throws IOException {
+    try {
+      delegate.close();
+    } catch (IOException exception) {
+      spanStatus = SpanStatus.INTERNAL_ERROR;
+      throw exception;
+    }
+    finishSpan();
+  }
+
+  private void finishSpan() {
+    if (currentSpan != null) {
+      if (file != null) {
+        String description = file.getName()
+          + " "
+          + "("
+          + StringUtils.byteCountToString(byteCount)
+          + ")";
+        currentSpan.setDescription(description);
+        currentSpan.setData("filepath", file.getAbsolutePath());
+      }
+      currentSpan.finish(spanStatus);
+    }
+  }
+
+  /**
+   * A task that returns a result and may throw an IOException.
+   * Implementors define a single method with no arguments called
+   * {@code call}.
+   *
+   * Derived from {@link java.util.concurrent.Callable}
+   */
+  @FunctionalInterface
+  interface FileIOCallable<V> {
+
+    V call() throws IOException;
+  }
+}
--- a/sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
+++ b/sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
@@ -20,7 +20,6 @@ final class FileIOSpanManager {
   private final @Nullable File file;
   private final @NotNull IHub hub;
 
-  private @Nullable Throwable throwable = null;
   private @NotNull SpanStatus spanStatus = SpanStatus.OK;
   private long byteCount;
 
@@ -61,7 +60,9 @@ final class FileIOSpanManager {
       return result;
     } catch (IOException exception) {
       spanStatus = SpanStatus.INTERNAL_ERROR;
-      throwable = exception;
+      if (currentSpan != null) {
+        currentSpan.setThrowable(exception);
+      }
       throw exception;
     }
   }
@@ -71,7 +72,9 @@ final class FileIOSpanManager {
       delegate.close();
     } catch (IOException exception) {
       spanStatus = SpanStatus.INTERNAL_ERROR;
-      throwable = exception;
+      if (currentSpan != null) {
+        currentSpan.setThrowable(exception);
+      }
       throw exception;
     } finally {
       finishSpan();
--- a/sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
+++ b/sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
@@ -46,7 +46,7 @@ class FileIOSpanManager {
         }
       }
       return result;
-    } catch (Exception exception) {
+    } catch (IOException exception) {
       spanStatus = SpanStatus.INTERNAL_ERROR;
       throw exception;
     }
--- a/sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
+++ b/sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
@@ -3,7 +3,6 @@ package io.sentry.instrumentation.file;
 import io.sentry.IHub;
 import io.sentry.ISpan;
 import io.sentry.SpanStatus;
-import io.sentry.util.Objects;
 import io.sentry.util.Pair;
 import io.sentry.util.Platform;
 import io.sentry.util.StringUtils;
@@ -31,10 +30,7 @@ final class FileIOSpanManager {
   }
 
   FileIOSpanManager(
-    final @Nullable ISpan currentSpan,
-    final @Nullable File file,
-    final @NotNull IHub hub
-  ) {
+      final @Nullable ISpan currentSpan, final @Nullable File file, final @NotNull IHub hub) {
     this.currentSpan = currentSpan;
     this.file = file;
     this.hub = hub;
@@ -43,12 +39,12 @@ final class FileIOSpanManager {
   /**
    * Performs file IO, counts the read/written bytes and handles exceptions in case of occurence
    *
-   * @param operation An IO operation to execute (e.g. {@link FileInputStream#read()} or {@link FileOutputStream#write(int)}
-   * The operation is of a type {@link Pair}, where the first element is the result of the IO operation,
-   * and the second element is the number of bytes read/written/skipped/etc.
+   * @param operation An IO operation to execute (e.g. {@link FileInputStream#read()} or {@link
+   *     FileOutputStream#write(int)} The operation is of a type {@link Pair}, where the first
+   *     element is the result of the IO operation, and the second element is the number of bytes
+   *     read/written/skipped/etc.
    */
-  <T> T performIO(final @NotNull FileIOCallable<T> operation)
-    throws IOException {
+  <T> T performIO(final @NotNull FileIOCallable<T> operation) throws IOException {
     try {
       final T result = operation.call();
       if (result instanceof Integer) {
@@ -86,11 +82,7 @@ final class FileIOSpanManager {
     if (currentSpan != null) {
       final String byteCountToString = StringUtils.byteCountToString(byteCount);
       if (file != null) {
-        final String description = file.getName()
-          + " "
-          + "("
-          + byteCountToString
-          + ")";
+        final String description = file.getName() + " " + "(" + byteCountToString + ")";
         currentSpan.setDescription(description);
         if (Platform.isAndroid() || hub.getOptions().isSendDefaultPii()) {
           currentSpan.setData("file.path", file.getAbsolutePath());
@@ -105,11 +97,10 @@ final class FileIOSpanManager {
   }
 
   /**
-   * A task that returns a result and may throw an IOException.
-   * Implementors define a single method with no arguments called
-   * {@code call}.
+   * A task that returns a result and may throw an IOException. Implementors define a single method
+   * with no arguments called {@code call}.
    *
-   * Derived from {@link java.util.concurrent.Callable}
+   * <p>Derived from {@link java.util.concurrent.Callable}
    */
   @FunctionalInterface
   interface FileIOCallable<V> {
--- a/sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
+++ b/sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
@@ -3,7 +3,6 @@ package io.sentry.instrumentation.file;
 import io.sentry.IHub;
 import io.sentry.ISpan;
 import io.sentry.SpanStatus;
-import io.sentry.util.Pair;
 import io.sentry.util.Platform;
 import io.sentry.util.StringUtils;
 import java.io.Closeable;
@@ -29,7 +28,9 @@ final class FileIOSpanManager {
   }
 
   FileIOSpanManager(
-      final @Nullable ISpan currentSpan, final @Nullable File file, final boolean isSendDefaultPii) {
+      final @Nullable ISpan currentSpan,
+      final @Nullable File file,
+      final boolean isSendDefaultPii) {
     this.currentSpan = currentSpan;
     this.file = file;
     this.isSendDefaultPii = isSendDefaultPii;
--- a/sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
+++ b/sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
@@ -47,25 +47,22 @@ final class FileIOSpanManager {
    * The operation is of a type {@link Pair}, where the first element is the result of the IO operation,
    * and the second element is the number of bytes read/written/skipped/etc.
    */
-  <T> T performIO(final @NotNull FileIOCallable<Pair<T, T>> operation)
+  <T> T performIO(final @NotNull FileIOCallable<T> operation)
     throws IOException {
     try {
-      final Pair<T, T> result = operation.call();
-      final T res = result.getFirst();
-      if (res instanceof Integer) {
-        final int resUnboxed = (int) result.getFirst();
-        final int count = (int) result.getSecond();
+      final T result = operation.call();
+      if (result instanceof Integer) {
+        final int resUnboxed = (int) result;
         if (resUnboxed != -1) {
-          byteCount += count;
+          byteCount += resUnboxed;
         }
-      } else if (res instanceof Long) {
-        final long resUnboxed = (long) result.getFirst();
-        final long count = (long) result.getSecond();
+      } else if (result instanceof Long) {
+        final long resUnboxed = (long) result;
         if (resUnboxed != -1L) {
-          byteCount += count;
+          byteCount += resUnboxed;
         }
       }
-      return Objects.requireNonNull(res, "Result of File IO is required");
+      return result;
     } catch (IOException exception) {
       spanStatus = SpanStatus.INTERNAL_ERROR;
       throwable = exception;
--- a/sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
+++ b/sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
@@ -39,9 +39,8 @@ final class FileIOSpanManager {
    * Performs file IO, counts the read/written bytes and handles exceptions in case of occurence
    *
    * @param operation An IO operation to execute (e.g. {@link FileInputStream#read()} or {@link
-   *     FileOutputStream#write(int)} The operation is of a type {@link Pair}, where the first
-   *     element is the result of the IO operation, and the second element is the number of bytes
-   *     read/written/skipped/etc.
+   *     FileOutputStream#write(int)} The operation is of a type {@link Integer} or {@link Long},
+   *     where the output is the result of the IO operation (byte count read/written)
    */
   <T> T performIO(final @NotNull FileIOCallable<T> operation) throws IOException {
     try {
--- a/sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
+++ b/sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
@@ -105,8 +105,8 @@ final class FileIOSpanManager {
    * <p>Derived from {@link java.util.concurrent.Callable}
    */
   @FunctionalInterface
-  interface FileIOCallable<V> {
+  interface FileIOCallable<T> {
 
-    V call() throws IOException;
+    T call() throws IOException;
   }
 }
--- a/sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
+++ b/sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
@@ -1,11 +1,16 @@
 package io.sentry.instrumentation.file;
 
+import io.sentry.IHub;
 import io.sentry.ISpan;
-import io.sentry.Sentry;
 import io.sentry.SpanStatus;
+import io.sentry.util.Objects;
+import io.sentry.util.Pair;
+import io.sentry.util.Platform;
 import io.sentry.util.StringUtils;
 import java.io.Closeable;
 import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
 import java.io.IOException;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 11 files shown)


sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
```diff
@@ -0,0 +1,197 @@
...
+ */
+@SuppressWarnings("unused") // used via bytecode manipulation (SAGP)
+@Open
+public class SentryFileInputStream extends FileInputStream {
+
+  private final @Nullable ISpan currentSpan;
+  private final @Nullable File file;
+  private final @NotNull FileInputStream delegate;
+
+  private @NotNull SpanStatus spanStatus = SpanStatus.OK;
```
romtsn, line 31: Not sure if we should handle concurrency here - generally, it's discouraged to use IOStreams across multiple threads

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
```diff
@@ -0,0 +1,197 @@
...
+  private static FileInputStreamInitData init(
+    @NotNull FileDescriptor fd,
+    @Nullable FileInputStream delegate
+  ) {
+    final ISpan span = startSpan();
+    if (delegate == null) {
+      delegate = new FileInputStream(fd);
+    }
+    // TODO: it's only possible to get filename from FileDescriptor via reflection AND when it's
+    // TODO: running on Android, should we do that?
```
marandaneto, line 85:
```suggestion
    // TODO: it's only possible to get filename from FileDescriptor via reflection AND when it's
    // running on Android, should we do that?
```

marandaneto: FileDescriptor is always used with FIOS right, and FIOS have file name, so we don't need to I guess

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
```diff
@@ -0,0 +1,197 @@
...
+      }
+      return result;
+    } catch (IOException exception) {
+      spanStatus = SpanStatus.INTERNAL_ERROR;
+      throw exception;
+    }
+  }
+
+  @Override
+  public int read(byte @NotNull [] b) throws IOException {
```
marandaneto, line 110:
```suggestion
  public int read(@NotNull byte [] b) throws IOException {
```

nit only, annotations always before

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
```diff
@@ -0,0 +1,197 @@
...
+    if (delegate == null) {
+      delegate = new FileInputStream(fd);
+    }
+    // TODO: it's only possible to get filename from FileDescriptor via reflection AND when it's
+    // TODO: running on Android, should we do that?
+    return new FileInputStreamInitData(null, span, delegate);
+  }
+
+  private static @Nullable
+  ISpan startSpan() {
```
marandaneto, line 90: nit only: I'd rather do only `@Nullable` on this line

sentry/src/main/java/io/sentry/instrumentation/file/FileInputStreamInitData.java
```diff
@@ -0,0 +1,21 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+final class FileInputStreamInitData {
+
+  @Nullable File file;
+  @Nullable ISpan span;
+  @NotNull FileInputStream delegate;
+
+  public FileInputStreamInitData(@Nullable File file, @Nullable ISpan span,
```
marandaneto, line 15: always mark fields as `final` to make them immutable

eg `var` and `val` on Kotlin

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
```diff
@@ -0,0 +1,197 @@
...
+  private void finishSpan() {
+    if (currentSpan != null) {
+      if (file != null) {
+        String description = file.getName()
+          + " "
+          + "("
+          + StringUtils.byteCountToString(byteCount)
+          + ")";
+        currentSpan.setDescription(description);
+        currentSpan.setData("filepath", file.getAbsolutePath());
```
marandaneto, line 169: wondering if it makes sense to be relative path because of PII

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
```diff
@@ -0,0 +1,197 @@
...
+    finishSpan();
+  }
+
+  private void finishSpan() {
+    if (currentSpan != null) {
+      if (file != null) {
+        String description = file.getName()
+          + " "
+          + "("
+          + StringUtils.byteCountToString(byteCount)
```
marandaneto, line 166: whats about byte count being a `data` too?

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileOutputStream.java
```diff
@@ -0,0 +1,196 @@
...
+public class SentryFileOutputStream extends FileOutputStream {
+
+  private final @Nullable ISpan currentSpan;
+  private final @Nullable File file;
+  private final @NotNull FileOutputStream delegate;
+
+  private @NotNull SpanStatus spanStatus = SpanStatus.OK;
+  private long byteCount;
+
+  public SentryFileOutputStream(String name) throws FileNotFoundException {
```
marandaneto, line 34: missing all the @NotNull @Nullable annotations

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileOutputStream.java
```diff
@@ -0,0 +1,196 @@
...
+      byteCount += len;
+    } catch (IOException exception) {
+      spanStatus = SpanStatus.INTERNAL_ERROR;
+      throw exception;
+    }
+  }
+
+  @Override public void close() throws IOException {
+    try {
+      delegate.close();
```
marandaneto, line 135: this class has plenty of very similar methods compared to `SentryFileInputStream`, can we do some DRY?

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
romtsn: you mean in addition to the description or instead?

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
marandaneto: good question, lets discuss with @bruno-garcia too since he always has opinions about that

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
marandaneto: instead, not sure if its a good idea, food for thought

romtsn: https://github.com/getsentry/develop/issues/463 we agreed to have it in the description initially, but now I see that there's a point to have under `data`, so prob. makes sense to move it there

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
romtsn: Good point, I think for Android it's not concerning, since it's always under `data/user/0/etc.`, but for pure JVM/backend projects can be dangerous. @bruno-garcia opinions?

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileOutputStream.java
romtsn: yeah, thought of this already, will try to make smthng up

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
romtsn: hm, not sure I got you, FIOS have file name when it was instantiated with a filename or a File instance, but when the FileDescriptor ctor is used, then there's no access to the underlying File instance

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
romtsn: I think it has to be where it is now, otherwise it has no effect
image

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
marandaneto: @romtsn tbh I don't recall using `FileDescriptor` so maybe what I've said indeed does not make sense, but looking into the Java docs and examples, eg https://www.javatpoint.com/java-filedescriptor-class
`FileDescriptor` is never used in the same way as standard File or FIOS, but always with it.

See `fd = fos.getFD(); `

in that case, we have access to file name because we'll be instrumenting the `fos` anyway

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileOutputStream.java
romtsn: Extracted the logic into a `FileIOSpanManager` - not a big fan of having a `manager` in the name, but couldn't come up with anything better

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
marandaneto: could be both actually, the description has a user-friendly size, in KB, MB or...
in the data, we have the real byte count

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
romtsn: Right, but the problem is that this can be returned from the Android SDK as a return type, and we do not instrument FIOS there. Anyway, I think it's fine for now, we will still instrument it, just w/o the filename/path

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
romtsn: Sounds good, will change to that 👍

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
marandaneto: then we are fine.

sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
```diff
@@ -0,0 +1,97 @@
...
+      if (file != null) {
+        description = file.getName()
+          + " "
+          + "("
+          + StringUtils.byteCountToString(byteCount)
+          + ")";
+        currentSpan.setDescription(description);
+        currentSpan.setData("file.path", file.getAbsolutePath());
+      } else {
+        description = StringUtils.byteCountToString(byteCount);
```
marandaneto, line 77:
```suggestion
      String description;
      final String byteCountToString = StringUtils.byteCountToString(byteCount);
      if (file != null) {
        description = file.getName()
          + " "
          + "("
          + byteCountToString
          + ")";
        currentSpan.setDescription(description);
        currentSpan.setData("file.path", file.getAbsolutePath());
      } else {
        description = byteCountToString;
```

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
bruno-garcia: `FileStreamInput` can't/isn't designed be used concurrently (not without synchronization) so I think we're safe here.

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
bruno-garcia: I think it's valuable to have the FQP. The problem is knowing where in the path we should cut off not to throw away valuable info. There are some heuristics like `/Users/Xyz` and `C:\Users\Xyz` and IIRC that was done on the server.

There's PII filtering in ingest but I'm not sure if it'll apply to Spans. @untitaker  might be able to help us here.

untitaker: Hi, I replied in Slack because I am not sure of the answer, but this is what I am sure of:

> 
> If you can't find a reasonable cutoff in platform- and domain-specific code, the server will do no better. The [advanced data scrubbing rules](https://docs.sentry.io/product/data-management-settings/scrubbing/advanced-datascrubbing/) you can configure at the moment are, and yes they will also work on span data:
> 
> 1. set a custom regex to strip away matches
> 2. strip the entire path and leave the basename
> 
> This has to be explicitly set up by the user. However, we communicate in all SDK docs that our SDKs do not send PII by default. So you cannot intentionally send PII and just expect the server to strip it away. That's not the guarantee we make, and even if the server does manage to strip away all the PII it will be too late for some customers. Safe defaults first.

That doesn't necessarily resolve your discussion because it's unclear to me at the moment as to whether we consider absolute paths as "PII".

marandaneto: I'd rather go with file name then, better safe than sorry.
@romtsn maybe confirm with Jan/Armin, they might know if the file path is PII or not

romtsn: looks like we can easily distinguish between desktop vs server using https://docs.oracle.com/javase/8/docs/api/java/awt/GraphicsEnvironment.html#isHeadless--, so maybe we just skip absolute paths for desktops

marandaneto: this wont compile for Android, awt isn't part of the Android SDK iirc, unless you add a layer of abstraction.
check `io.sentry.util.Platform#isAndroid` tho, it'd be fine to add for Mobile I believe at least.

sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
```diff
@@ -0,0 +1,100 @@
...
+
+  private final @Nullable ISpan currentSpan;
+  private final @Nullable File file;
+
+  private @Nullable Throwable throwable = null;
+  private @NotNull SpanStatus spanStatus = SpanStatus.OK;
+  private long byteCount;
+
+  static @Nullable ISpan startSpan(final @NotNull String op) {
+    final ISpan parent = Sentry.getSpan();
```
marandaneto, line 23: to make it testable, make a ctor with IHub and without, that defaults to `HubAdapter.getInstance()`, see eg SentryOkHttpInterceptor

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
romtsn: good concern. There was also mentioned `sendDefaultPii` in the slack thread, so perhaps, we could send it for server and desktop, if this flag is `true`, and always for mobile

sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
```diff
@@ -0,0 +1,110 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+final class FileIOSpanManager {
+
+  private final @Nullable ISpan currentSpan;
+  private final @Nullable File file;
+  private final @NotNull IHub hub;
+
+  private @Nullable Throwable throwable = null;
```
marandaneto, line 23: if you have the instance of `currentSpan`, should we not set the `throwable` directly?

sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
```diff
@@ -0,0 +1,110 @@
...
+    final ISpan parent = hub.getSpan();
+    return parent != null ? parent.startChild(op) : null;
+  }
+
+  FileIOSpanManager(
+      final @Nullable ISpan currentSpan, final @Nullable File file, final @NotNull IHub hub) {
+    this.currentSpan = currentSpan;
+    this.file = file;
+    this.hub = hub;
+  }
```
marandaneto, line 37:
```suggestion
  FileIOSpanManager(
      final @Nullable ISpan currentSpan, final @Nullable File file, final @NotNull IHub hub) {
    this.currentSpan = currentSpan;
    this.file = file;
    this.hub = hub;
  }

  FileIOSpanManager(
          final @Nullable ISpan currentSpan, final @Nullable File file) {
    this(currentSpan, file, HubAdapter.getInstance());
  }
```

the ctor that accepts the hub is only used for testing

sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
```diff
@@ -0,0 +1,110 @@
...
+
+  private final @Nullable ISpan currentSpan;
+  private final @Nullable File file;
+  private final @NotNull IHub hub;
+
+  private @Nullable Throwable throwable = null;
+  private @NotNull SpanStatus spanStatus = SpanStatus.OK;
+  private long byteCount;
+
+  static @Nullable ISpan startSpan(final @NotNull IHub hub, final @NotNull String op) {
```
marandaneto, line 27: I'm a bit lost here, we have a static method within an instance class, but also a ctor that takes a `hub`, mind elaborating the use case?

sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
romtsn: True, missed that 👍

sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  /**
+   * A task that returns a result and may throw an IOException. Implementors define a single method
+   * with no arguments called {@code call}.
+   *
+   * <p>Derived from {@link java.util.concurrent.Callable}
+   */
+  @FunctionalInterface
+  interface FileIOCallable<V> {
```
marandaneto, line 106: we usually use `T` for generics

sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
```diff
@@ -0,0 +1,110 @@
...
+      final @Nullable ISpan currentSpan, final @Nullable File file, final @NotNull IHub hub) {
+    this.currentSpan = currentSpan;
+    this.file = file;
+    this.hub = hub;
+  }
+
+  /**
+   * Performs file IO, counts the read/written bytes and handles exceptions in case of occurence
+   *
+   * @param operation An IO operation to execute (e.g. {@link FileInputStream#read()} or {@link
```
romtsn, line 42: Wrong doc, should change it

sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
marandaneto: unless this class is never gonna be used by clients, so then it makes sense to always use the very same hub

sentry/src/main/java/io/sentry/instrumentation/file/FileInputStreamInitData.java
```diff
@@ -0,0 +1,27 @@
...
+  final @Nullable File file;
+  final @Nullable ISpan span;
+  final @NotNull FileInputStream delegate;
+  final @NotNull IHub hub;
+
+  public FileInputStreamInitData(
+      final @Nullable File file,
+      final @Nullable ISpan span,
+      final @NotNull FileInputStream delegate,
+      final @NotNull IHub hub) {
```
marandaneto, line 21: same as https://github.com/getsentry/sentry-java/pull/1826/files#r766443142

sentry/src/main/java/io/sentry/instrumentation/file/FileInputStreamInitData.java
```diff
@@ -0,0 +1,27 @@
...
+import org.jetbrains.annotations.Nullable;
+
+final class FileInputStreamInitData {
+
+  final @Nullable File file;
+  final @Nullable ISpan span;
+  final @NotNull FileInputStream delegate;
+  final @NotNull IHub hub;
+
+  public FileInputStreamInitData(
```
marandaneto, line 17:
```suggestion
  FileInputStreamInitData(
```

if the class is package private, I guess we don't need to make it public here

sentry/src/main/java/io/sentry/instrumentation/file/FileInputStreamInitData.java
```diff
@@ -0,0 +1,27 @@
...
+import java.io.FileInputStream;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+final class FileInputStreamInitData {
+
+  final @Nullable File file;
+  final @Nullable ISpan span;
+  final @NotNull FileInputStream delegate;
+  final @NotNull IHub hub;
```
marandaneto, line 15: can we make them private?

sentry/src/main/java/io/sentry/instrumentation/file/FileOutputStreamInitData.java
```diff
@@ -0,0 +1,30 @@
+package io.sentry.instrumentation.file;
+
+import io.sentry.IHub;
+import io.sentry.ISpan;
+import java.io.File;
+import java.io.FileOutputStream;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+final class FileOutputStreamInitData {
```
marandaneto, line 10: same comments from FileInputStreamInitData

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
```diff
@@ -0,0 +1,144 @@
...
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * An implementation of {@link java.io.FileInputStream} that creates a {@link io.sentry.ISpan} for
+ * reading operation with filename and byte count set as description
+ *
+ * <p>Note, that span is started when this InputStream is instantiated via constructor and finishes
+ * when the {@link java.io.FileInputStream#close()} is called.
+ */
+@Open
```
marandaneto, line 23: if we are not extending, should we make it `final`?

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
```diff
@@ -0,0 +1,144 @@
...
+
+  public SentryFileInputStream(final @Nullable File file) throws FileNotFoundException {
+    this(file, HubAdapter.getInstance());
+  }
+
+  public SentryFileInputStream(final @NotNull FileDescriptor fdObj) {
+    this(fdObj, HubAdapter.getInstance());
+  }
+
+  public SentryFileInputStream(final @Nullable File file, final @NotNull IHub hub)
```
marandaneto, line 41: maybe every ctor that have a hub should be package private

sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
romtsn: yeah, if you take a look at how it's used, we need this static method, because we start the span *before* the FIOS class is instantiated to also measure how much time the ctor of FIOS will take. the hub is propagated here because we need to test this.

As for the ctor, the only reason to have it there is because we are using it to retrieve the options to check if `sendDefaultPii` is enabled in `finishSpan` (which we also want to test)

sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
romtsn: See https://github.com/getsentry/sentry-java/pull/1826#discussion_r766445062

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
```diff
@@ -0,0 +1,144 @@
...
+
+  @Override
+  public int read() throws IOException {
+    // this is the only case, when the read() operation returns the byte value, and not the count
+    // hence we need this special handling
+    AtomicInteger result = new AtomicInteger(0);
+    spanManager.performIO(
+        () -> {
+          result.set(delegate.read());
+          return result.get() != -1 ? 1 : 0;
```
marandaneto, line 93:
```suggestion
	        final read = delegate.read();
          result.set(read);
          return read != -1 ? 1 : 0;
```

its probably cheaper since AtomicInteger needs to be atomic

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
romtsn: I thought it's good to leave it open, because the users might want to extend it, for instance, if they have their own impls of FIOS, but also want to have them instrumented

sentry/src/main/java/io/sentry/instrumentation/file/FileInputStreamInitData.java
romtsn: mm, and use getters instead? we can, but if they are final, is there a difference?

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
```diff
@@ -0,0 +1,144 @@
...
+        final @NotNull FileInputStream delegate, final @NotNull FileDescriptor descriptor) {
+      return new SentryFileInputStream(
+          init(descriptor, delegate, HubAdapter.getInstance()), descriptor);
+    }
+
+    public static FileInputStream create(
+        final @NotNull FileInputStream delegate, final @Nullable File file, final @NotNull IHub hub)
+        throws FileNotFoundException {
+      return new SentryFileInputStream(init(file, delegate, hub));
+    }
```
marandaneto, line 142: if this is for testing only, we can make package-private, or test the ctor directly

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileOutputStream.java
```diff
@@ -0,0 +1,155 @@
...
+
+/**
+ * An implementation of {@link java.io.FileOutputStream} that creates a {@link io.sentry.ISpan} for
+ * writing operation with filename and byte count set as description
+ *
+ * <p>Note, that span is started when this OutputStream is instantiated via constructor and finishes
+ * when the {@link java.io.FileOutputStream#close()} is called.
+ */
+@Open
+public class SentryFileOutputStream extends FileOutputStream {
```
marandaneto, line 23: same comments from SentryFileInputStream

sentry/src/main/java/io/sentry/util/StringUtils.java
```diff
@@ -51,4 +54,22 @@ private StringUtils() {}
...
     }
   }
+
+  /**
+   * Converts the given number of bytes to a human-readable string.
+   *
+   * @param bytes the number of bytes
+   * @return a string representing the human-readable byte count (e.g. 1kB, 20 MB, etc.)
+   */
+  public static @NotNull String byteCountToString(long bytes) {
```
marandaneto, line 64:
```suggestion
  public static @NotNull String byteCountToString(final long bytes) {
```

sentry/src/main/java/io/sentry/util/StringUtils.java
```diff
@@ -51,4 +54,22 @@ private StringUtils() {}
...
+    if (-1000 < bytes && bytes < 1000) {
+      return bytes + " B";
+    }
+    CharacterIterator ci = new StringCharacterIterator("kMGTPE");
+    while (bytes <= -999_950 || bytes >= 999_950) {
+      bytes /= 1000;
+      ci.next();
+    }
+    return String.format(Locale.US, "%.1f %cB", bytes / 1000.0, ci.current());
```
marandaneto, line 73: can it be `Locale.ROOT`? cant guarantee that `Locale.US` is available

sentry/src/test/java/io/sentry/instrumentation/file/SentryFileInputStreamTest.kt
```diff
@@ -0,0 +1,201 @@
...
+        fun getSut(
+            tmpFile: File? = null,
+            activeTransaction: Boolean = true,
+            delegate: FileInputStream? = null,
+            fileDescriptor: FileDescriptor? = null,
+            sendDefaultPii: Boolean = false
+        ): SentryFileInputStream {
+            tmpFile?.writeText("Text")
+            whenever(hub.options).thenReturn(
+                SentryOptions().apply {
```
marandaneto, line 35: cheaper having a SentryOptions as an attribute of Fixture, and flipping only the flag

sentry/src/test/java/io/sentry/instrumentation/file/SentryFileInputStreamTest.kt
```diff
@@ -0,0 +1,201 @@
...
+            )
+            if (activeTransaction) {
+                whenever(hub.span).thenReturn(sentryTracer)
+            }
+            return if (delegate == null) {
+                if (fileDescriptor == null) {
+                    SentryFileInputStream(tmpFile, hub)
+                } else {
+                    SentryFileInputStream(fileDescriptor, hub)
+                }
```
marandaneto, line 47: maybe split up in 2 getSut? logic inside of the getSut makes it harder to read

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
romtsn: If you mean Factory, nope, it's for the bytecode instrumentation, since we do wrapping there.

romtsn: Or if you mean this specific method, yep, it's only for testing

sentry/src/test/java/io/sentry/instrumentation/file/SentryFileInputStreamTest.kt
```diff
@@ -0,0 +1,201 @@
...
+            }
+        }
+    }
+
+    @get:Rule
+    val tmpDir = TemporaryFolder()
+
+    private val fixture = Fixture()
+
+    private val tmpFile: File get() = tmpDir.newFile("test.txt")
```
marandaneto, line 63: I'd add an after test to delete this file, see DirectoryProcessorTest#shutdown

sentry/src/main/java/io/sentry/util/StringUtils.java
romtsn: Can't do that since we modify the bytes in line 70

sentry/src/test/java/io/sentry/instrumentation/file/SentryFileInputStreamTest.kt
```diff
@@ -0,0 +1,201 @@
...
+
+    @Test
+    fun `when no active transaction does not capture a span`() {
```
marandaneto, line 66: I like the test in this class, names are concise and well tested

sentry/src/main/java/io/sentry/util/StringUtils.java
romtsn: image
or are you aware of some edge case?

sentry/src/test/java/io/sentry/instrumentation/file/SentryFileOutputStreamTest.kt
```diff
@@ -0,0 +1,106 @@
...
+import io.sentry.SentryTracer
+import io.sentry.SpanStatus
+import io.sentry.TransactionContext
+import org.junit.Rule
+import org.junit.Test
+import org.junit.rules.TemporaryFolder
+import java.io.File
+import kotlin.test.assertEquals
+
+class SentryFileOutputStreamTest {
```
marandaneto, line 16: same comments from SentryFileInputStreamTest

sentry/src/test/java/io/sentry/instrumentation/file/SentryFileInputStreamTest.kt
romtsn: But the TemporaryFolder rule will erase the entire folder anyway?

sentry/src/main/java/io/sentry/instrumentation/file/FileInputStreamInitData.java
marandaneto: not really, indeed

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
marandaneto: this method specifically, its the only one that takes a `hub`

sentry/src/main/java/io/sentry/util/StringUtils.java
marandaneto: it is an edge case, iirc Chinese ROMs

sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
marandaneto: gotcha, the method is more a DRY instead of having the very same method twice in both classes.

wondering if it makes sense to pass the `sendDefaultPii` via ctor, instead of the whole `hub`

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
marandaneto: mm, yeah, wondering if it makes sense to keep it closed, easier to debug new issues, and later open it if people need it, its always easier to open, always harder to close it.

sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
romtsn: Yeah good point, we could do that certainly

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
romtsn: Actually, good point. We have a Factory which they still can use to pass their impls, and we will do wrapping and instrumenting

sentry/src/test/java/io/sentry/instrumentation/file/SentryFileInputStreamTest.kt
marandaneto: oh yeah good catch, we dont use TemporaryFolder in that test.

marandaneto (approved):

PR #1832, opened 2021-12-09
Feat: Add locale to device context and deprecate language

marandaneto:
## :scroll: Description
Feat: Add locale to device context and deprecate language
`language` now returns only the language eg `en` instead of `en_US`, since the `language` field was never indexed, its not a breaking change on discover and alerts

`language` was actually the `locale`, and `locale` was missing.
Closes https://github.com/getsentry/sentry-java/issues/1834

unit tests

## :pencil: Checklist

- [X] No breaking changes

## :crystal_ball: Next steps
add Support for https://develop.sentry.dev/sdk/event-payloads/contexts/#culture-context

### DIFF
--- a/sentry-android-core/src/test/java/io/sentry/android/core/DefaultAndroidEventProcessorTest.kt
+++ b/sentry-android-core/src/test/java/io/sentry/android/core/DefaultAndroidEventProcessorTest.kt
@@ -27,6 +27,7 @@ import io.sentry.protocol.SentryTransaction
 import io.sentry.protocol.User
 import io.sentry.test.getCtor
 import org.junit.runner.RunWith
+import java.util.*
 import kotlin.test.BeforeTest
 import kotlin.test.Test
 import kotlin.test.assertEquals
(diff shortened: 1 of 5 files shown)


sentry-android-core/src/test/java/io/sentry/android/core/DefaultAndroidEventProcessorTest.kt
```diff
@@ -376,4 +377,16 @@ class DefaultAndroidEventProcessorTest {
...
     }
+
+    @Test
+    fun `Event sets language and locale`() {
+        val sut = fixture.getSut(context)
+        val locale = Locale.getDefault()
+
+        assertNotNull(sut.process(SentryEvent(), null)) {
+            val device = it.contexts.device!!
+            assertEquals(device.language, locale.language)
```
romtsn, line 388: can we assert against raw string values? because we are using the same function (getDefault) as in production code, so if it changes, our test in theory should fail, but it won't.

I think I'd set the default to `Locale.setDefault(Locale.GERMAN)` for instance, then assert against "de_DE" and so on

romtsn (approved):

sentry/src/main/java/io/sentry/protocol/Device.java
```diff
@@ -93,7 +93,18 @@
   private @Nullable TimeZone timezone;
 
   private @Nullable String id;
-  private @Nullable String language;
+
+  /**
+   * This method returns the language code for this locale, which will either be the empty string or
```
philipphofmann, line 98: `m`: I think it's worth mentioning that `language` is the same as `locale`, but it was a mistake and we can't change it, because it would be a breaking change.

philipphofmann (approved):
One minor comment.

sentry/src/main/java/io/sentry/protocol/Device.java
marandaneto: it returns the language only now, see the PR description

PR #1848, opened 2021-12-16
Fix: Make App start cold/warm visible to Hybrid SDKs

marandaneto:
## :scroll: Description
Fix: Make App start cold/warm visible to Hybrid SDKs

https://github.com/getsentry/sentry-react-native/issues/1968
React native reads the Cold or Warm start from the Android SDK, but the Android SDK only sets it if performance is enabled on the Android SDK as well, by default, performance is not enabled, only on the RN SDK, hence we have to flip the flag anyway, even if performance is disabled on the Android SDK.

## :pencil: Checklist

- [X] No breaking changes

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AppStartState.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AppStartState.java
@@ -44,7 +44,7 @@ public final class AppStartState {
   }
 
   @Nullable
-  synchronized Long getAppStartInterval() {
+  public synchronized Long getAppStartInterval() {
     if (appStartMillis == null || appStartEndMillis == null) {
       return null;
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 * Bump: log4j to 2.16.0 (#1845)
+* Fix: Make App start cold/warm visible to Hybrid SDKs (#)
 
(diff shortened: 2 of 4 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/AppStartState.java
```diff
@@ -44,7 +44,7 @@ void setAppStartEnd(final long appStartEndMillis) {
   }
 
   @Nullable
-  synchronized Long getAppStartInterval() {
+  public synchronized Long getAppStartInterval() {
```
marandaneto, line 47: this needs to be public to be used in hybrid SDKs, e.g.
reason is https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/PerformanceAndroidEventProcessor.java#L63-L64

romtsn (approved):

romtsn:
LGTM, but wouldn't it be better to propagate the performanceEnabled flag from the hybrid SDKs rather? If they are utilising native SDKs for perf anyway

marandaneto:
> LGTM, but wouldn't it be better to propagate the performanceEnabled flag from the hybrid SDKs rather? If they are utilising native SDKs for perf anyway

no, we've decided that users would need to enable that manually if they want to, most hybrid apps don't do "native screens" so it'd not make much sense anyway.

PR #1855, opened 2021-12-20
Fix: Detect App Cold start correctly for Hybrid SDKs

marandaneto:
## :scroll: Description
Fix: Detect App Cold start correctly for Hybrid SDKs

https://github.com/getsentry/sentry-react-native/issues/1968

## :pencil: Checklist

- [X] No breaking changes

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AppStartState.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AppStartState.java
@@ -18,7 +18,7 @@ public final class AppStartState {
   private @Nullable Long appStartEndMillis;
 
   /** The type of App start coldStart=true -> Cold start, coldStart=false -> Warm start */
-  private boolean coldStart;
+  private @Nullable Boolean coldStart = null;
 
   /** appStart as a Date used in the App's Context */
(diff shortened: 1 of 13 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/AppStartState.java
```diff
@@ -18,7 +18,7 @@
   private @Nullable Long appStartEndMillis;
 
   /** The type of App start coldStart=true -> Cold start, coldStart=false -> Warm start */
-  private boolean coldStart;
+  private @Nullable Boolean coldStart = null;
```
romtsn, line 21: we kinda have 3 states now for this: unset, cold, warm - should we introduce an enum rather?

romtsn (approved):

sentry-android-core/src/main/java/io/sentry/android/core/AppStartState.java
marandaneto: yep, `enum` is more expensive but it makes sense, since this is internal, I will make it in another PR to unblock this one and cut a release

PR #1861, opened 2022-01-04
Feat: Relax TransactionNameProvider.

maciejwalkowiak:
## :scroll: Description

Turn TransactionNameProvider into an interface enabling using SentryTracingFilter with other MVC frameworks like Grails or JAX-RS.

- #1859
- https://forum.sentry.io/t/sentry-java-spring-boot-and-grails-performance-tracing-not-working/15765/3

Unit & Integration tests

## :pencil: Checklist

- [ ] No breaking changes

For great majority of users it's not a breaking change. It is breaking only because the `SentryTracingFilter` secondary constructor takes now an interface instead of a class.


marandaneto:
@maciejwalkowiak please merge 6.x.x into your branch, I will be moving 6.0.0 forward.

marandaneto (approved):

maciejwalkowiak:
Since we do not provide official support for Grails 2 (it's like 10 years old?) I can only give some hints:

- include `SentryTracingFilter` in `web.xml` or any other place where you can declare servlet filters
- assuming that it works similar to Grails 4, make sure that it runs after any Grails component that sets controller name on a `HttpServletRequest` attribute

If you are not able to use servlet filters, you will likely need to create custom Grails filter as explained in [Grails 2.2. docs](https://grails.github.io/grails2-doc/2.2.2/guide/single.html#filters) and you can use `SentryTracingFilter` as a reference, as the logic will be likely the same.

Issue #1536, opened 2021-06-16
Remove @Deprecated and @ApiStatus.ScheduledForRemoval fields


marandaneto:
breaking change 6.0.0
do migration docs

marandaneto:
@maciejwalkowiak same question for `SentryUserProviderEventProcessor`.

maciejwalkowiak:
- `SentrySpringRequestListener` - should be removed, `SentrySpringFilter` should be used instead. This only affects users who used `SentrySpringRequestListener` directly, and message "use `SentrySpringFilter` instead` is enough
- `SentryUserProviderEventProcessor` - should be removed, once users start using `SentrySpringFilter` no other action is required
- `SentryProperties#isEnableTracing` - tracing gets enabled either with a presence of `TracesSamplerCallback` set on `SentryOptions` or `tracesSampleRate` set. To read if tracing is enabled use `SentryOptions#isTracingEnabled`

marandaneto:
@maciejwalkowiak thanks

marandaneto:
@maciejwalkowiak related to the removal of `SentryUserProviderEventProcessor`, Can I remove the `SentryUserProvider` and all its usages too? looks like only related to the `SentryUserProviderEventProcessor` but I wanted to confirm anyway.

marandaneto:
@maciejwalkowiak Removing `SentryUserProviderEventProcessor`, tests fail for `SentrySpringIntegrationTest`, I also don't see how `SentrySpringFilter` sets the user to the scope automatically since the tests use the `TestRestTemplate` with basic auth, so the user should be automatically captured by `SentrySpringFilter`, right?

PR #1875, opened 2022-01-18
Ref: Remove deprecated and scheduled fields

marandaneto:
## :scroll: Description
Ref: Remove deprecated and scheduled fields

Closes https://github.com/getsentry/sentry-java/issues/1536

## :pencil: Checklist

- [ ] No breaking changes


sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -22,8 +20,6 @@
   private final @NotNull Span root;
   private final @NotNull List<Span> children = new CopyOnWriteArrayList<>();
   private final @NotNull IHub hub;
-  private final @NotNull Contexts contexts = new Contexts();
-  private @Nullable Request request;
```
marandaneto, line 26: @maciejwalkowiak since those 2 fields should be set on the Scope, I believe nothing else should be changed, is that correct? only the removal of the fields.

sentry/src/test/java/io/sentry/SentryTracerTest.kt
```diff
@@ -143,70 +140,6 @@ class SentryTracerTest {
         verify(fixture.hub).setSpanContext(ex, tracer.root, "name")
     }
 
-    @Test
```
marandaneto, line 146: TODO: maybe the tests from this class should be written by setting those fields to the Scope, so the transaction includes the fields

sentry/src/main/java/io/sentry/SentryTracer.java
maciejwalkowiak: Yes, as far as I can tell these fields should be removed because they are set on the scoped and set on actual transaction in `SentryClient#applyScope`.

romtsn (approved):

PR #1876, opened 2022-01-18
feat: Add breadcrumbs auto-capturing for UI events

### LINKED ISSUE #1412: Add UI events as breadcrumbs for Android
Requires https://github.com/getsentry/sentry-android-gradle-plugin/ and bytecode manipulation

romtsn:
## :scroll: Description

* Adds a custom `GestureListener` and `Window.Callback` that auto-capture breadcrumbs for UI events (clicks, swipes, etc.)

### Current limitations
* onMenuItemClicks are not supported
* Clicks inside DialogFragments are not supported (probably can be an option for FragmentLifecycleIntegration or just a new integration)
* Swipes for ViewPager and ViewPager2 are not supported (a new integration, since those are new deps?)
* RecyclerView is partially supported (the click on an item will be captured, but there's no useful data like which position in the list was selected - again, a new integration?)

Closes #1412 

Cus this looks beautiful
image

Manually

## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next steps
:shipit:


sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -27,6 +27,7 @@
  * Android Options initializer, it reads configurations from AndroidManifest and sets to the
  * SentryOptions. It also adds default values for some fields.
  */
+@SuppressWarnings("Convert2MethodRef")
```
bruno-garcia, line 30: I always wonder what/why are we suppressing things here. There's no spot to add a 'reason', right? 
Would it make sense to add as a comment?

```suggestion
@SuppressWarnings("Convert2MethodRef") // Because of reasons
```

sentry-android-core/src/main/java/io/sentry/android/core/gestures/ViewUtils.java
```diff
@@ -0,0 +1,101 @@
...
+    int h = view.getHeight();
+
+    return !(x < vx || x > vx + w || y < vy || y > vy + h);
+  }
+
+  static boolean isViewTappable(final @NotNull View view) {
+    return view.isClickable() && view.getVisibility() == View.VISIBLE;
+  }
+
+  static boolean isViewScrollable(final @NotNull View view) {
```
romtsn, line 71: in theory we could remove this, and then we don't need to add support for anything that I mentioned in the PR description, but then if you move a finger over a static view (say, textview/button/etc.), then the breadcrumb will be also captured as `swipe`. Open for ideas, but for now don't see any other way.

sentry-android-core/src/main/java/io/sentry/android/core/UserInteractionIntegration.java
```diff
@@ -0,0 +1,183 @@
...
+
+  @Override
+  public void onActivityPaused(@NotNull Activity activity) {
+    if (!isAllActivityCallbacksAvailable) {
+      stopTracking(activity.getWindow());
+    }
+  }
+
+  @Override
+  public void onActivityStopped(@NotNull Activity activity) {
```
bruno-garcia, line 131: Looking at this lifecycle method I assume we have another integration that already hooks into all of these.

I wonder if it wouldn't be more performatic to have a single class that inherits from `Application.ActivityLifecycleCallbacks` and that subscribe to all of these methods. And exposes a publish/subscribe hook where we can plug new integrations.

Please consider that I might be optimizing something here that doesn't require optimization. If having N instances hooked into ActivityLifecycleCallbacks is cheap, than adding the complexity clearly wouldnt' be worth it.

sentry-android-core/src/main/java/io/sentry/android/core/gestures/SentryGestureListener.java
```diff
@@ -0,0 +1,238 @@
...
+    final View decorView = ensureWindowDecorView("onUp");
+    final View scrollTarget = scrollState.targetRef.get();
+    if (decorView == null || scrollTarget == null) {
+      return;
+    }
+
+    if (scrollState.type == null) {
+      options
+        .getLogger()
+        .log(SentryLevel.INFO, "Unable to define scroll type. No breadcrumb captured.");
```
bruno-garcia, line 47: Is this worth an INFO level or it's only useful if we're debugging something (why didn't the crumb got added?)
```suggestion
        .log(SentryLevel.DEBUG, "Unable to define scroll type. No breadcrumb captured.");
```

sentry-android-core/src/main/java/io/sentry/android/core/gestures/SentryGestureListener.java
```diff
@@ -0,0 +1,238 @@
...
+    scrollState.startX = motionEvent.getX();
+    scrollState.startY = motionEvent.getY();
+    return false;
+  }
+
+  @Override
+  public boolean onSingleTapUp(final @Nullable MotionEvent motionEvent) {
+    final View decorView = ensureWindowDecorView("onSingleTapUp");
+    if (decorView == null || motionEvent == null) {
+      return false;
```
bruno-garcia, line 72: should this fallback to the parent implementation instead?

```suggestion
      return super.onSingleTapUp(motionEvent);
```

What does the return type do here?

sentry-android-core/src/main/java/io/sentry/android/core/gestures/SentryGestureListener.java
```diff
@@ -0,0 +1,238 @@
...
+    if (scrollState.type == null) {
+      @SuppressWarnings("Convert2MethodRef")
+      final @Nullable View target =
+        ViewUtils.findTarget(decorView, firstEvent.getX(), firstEvent.getY(),
+          view -> ViewUtils.isViewScrollable(view));
+
+      if (target == null) {
+        options
+          .getLogger()
+          .log(SentryLevel.INFO, "Unable to find scroll target. No breadcrumb captured.");
```
bruno-garcia, line 112: Similar to above, perhaps debug. Or if this seems like a bug, possibly WARN instead?

sentry-android-core/src/main/java/io/sentry/android/core/gestures/SentryGestureListener.java
```diff
@@ -0,0 +1,238 @@
...
+  }
+
+  private void addBreadcrumb(
+    final @NotNull View target,
+    final @NotNull String eventType,
+    final @NotNull Map<String, Object> additionalData
+  ) {
+    @NotNull String className;
+    if (target.getClass().getCanonicalName() != null) {
+      className = target.getClass().getCanonicalName();
```
bruno-garcia, line 156: Is this likely going to be minified? If so, we might need to consider proguard support here with the native team

sentry-android-core/src/main/java/io/sentry/android/core/gestures/SentryGestureListener.java
```diff
@@ -0,0 +1,238 @@
...
+    this.windowRef = windowRef;
+    this.hub = hub;
+    this.options = options;
+  }
+
+  public void onUp(final @NotNull MotionEvent motionEvent) {
+    final View decorView = ensureWindowDecorView("onUp");
+    final View scrollTarget = scrollState.targetRef.get();
+    if (decorView == null || scrollTarget == null) {
+      return;
```
bruno-garcia, line 41: `scrollTarget` will be null if the strong reference count went on to be 0 and GC collected it. In that case, this class becomes useless. Is this a valid scenario? Or is this something that can happen during shutdown or some other use case?

Trying to understand why we're going with WeakRef here

bruno-garcia (commented):
I don't have a lot of background on how this works but left a few thoughts hoping to help out

sentry-android-core/src/main/java/io/sentry/android/core/UserInteractionIntegration.java
romtsn: Yeah, it's just a list that is being iterated whenever a callback needs to fire, so I don't think we can optimize much here

sentry-android-core/src/main/java/io/sentry/android/core/gestures/SentryGestureListener.java
```diff
@@ -0,0 +1,238 @@
...
+    if (scrollState.type == null) {
+      @SuppressWarnings("Convert2MethodRef")
+      final @Nullable View target =
+        ViewUtils.findTarget(decorView, firstEvent.getX(), firstEvent.getY(),
+          view -> ViewUtils.isViewScrollable(view));
+
+      if (target == null) {
+        options
+          .getLogger()
+          .log(SentryLevel.INFO, "Unable to find scroll target. No breadcrumb captured.");
```
romtsn, line 112:
```suggestion
          .log(SentryLevel.DEBUG, "Unable to find scroll target. No breadcrumb captured.");
```

sentry-android-core/src/main/java/io/sentry/android/core/gestures/SentryGestureListener.java
romtsn: Yes, it will be minified for some cases (like AndroidX views), but this is the same problem we have now with the fragment integration, so I guess they should be tackled together

sentry-android-core/src/main/java/io/sentry/android/core/gestures/SentryGestureListener.java
romtsn: There's no parent since it's an interface that we are implementing, so there's no `super` to delegate to. The return type should indicate whether we consumed the event or not, so other listeners in the downstream chain will/will not be called depending on that

sentry-android-core/src/main/java/io/sentry/android/core/gestures/SentryGestureListener.java
romtsn: This is to make sure we are not leaking the underlying activity/fragment. If the ref count becomes 0, this class *should* become useless, as the activity/fragment got destroyed in that case.

sentry-android-core/src/main/java/io/sentry/android/core/UserInteractionIntegration.java
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  private static boolean checkAndroidXAvailability(final @NotNull LoadClass loadClass) {
+    try {
+      loadClass.loadClass("androidx.core.view.GestureDetectorCompat");
+      return true;
+    } catch (ClassNotFoundException ignored) {
+      return false;
+    }
+  }
```
marandaneto, line 53: Is there a way to know if the `androidx.core` package exists instead of a specific class? so we could do this check only once and pass the result along with the integrations.

sentry-android-core/src/main/java/io/sentry/android/core/UserInteractionIntegration.java
```diff
@@ -0,0 +1,183 @@
...
+    final @NotNull Application application,
+    final @NotNull IBuildInfoProvider buildInfoProvider,
+    final @NotNull LoadClass classLoader
+  ) {
+    this.application = Objects.requireNonNull(application, "Application is required");
+    Objects.requireNonNull(buildInfoProvider, "BuildInfoProvider is required");
+
+    isAndroidXAvailable = checkAndroidXAvailability(classLoader);
+    isAllActivityCallbacksAvailable =
+      buildInfoProvider.getSdkInfoVersion() >= Build.VERSION_CODES.Q;
```
marandaneto, line 43: Do we really need this for this integration? that was important for performance since we'd like to start as early as possible, but for the breadcrumbs, maybe not?

sentry-android-core/src/main/java/io/sentry/android/core/UserInteractionIntegration.java
marandaneto: Bruno's idea isn't bad tho, I already thought about it when doing the ActivityLifeCycle integration which does more than one thing, either separating or going with Bruno's approach is fine for me.

sentry-android-core/src/main/java/io/sentry/android/core/gestures/SentryGestureListener.java
marandaneto: The other integrations use `simpleName`, should we keep `getCanonicalName`? Also, I'd cache the `getCanonicalName` in a local field since it does some work and we're calling twice.

sentry-android-core/src/main/java/io/sentry/android/core/gestures/SentryGestureListener.java
```diff
@@ -0,0 +1,237 @@
...
+
+    final Breadcrumb breadcrumb = new Breadcrumb();
+    breadcrumb.setType("user");
+    breadcrumb.setCategory("ui." + eventType);
+    breadcrumb.setData("view.id", ViewUtils.getResourceId(target));
+    breadcrumb.setData("view.class", className);
+    for (final Map.Entry<String, Object> entry : additionalData.entrySet()) {
+      breadcrumb.getData().put(entry.getKey(), entry.getValue());
+    }
+    breadcrumb.setLevel(SentryLevel.INFO);
```
marandaneto, line 168: Maybe we create a `userInteraction` static method to create this breadcrumb that could be used by custom instrumentation too?

sentry-android-core/src/main/java/io/sentry/android/core/gestures/SentryGestureListener.java
```diff
@@ -0,0 +1,237 @@
...
+        .getLogger()
+        .log(SentryLevel.DEBUG, "DecorView is null in " + caller + ". No breadcrumb captured.");
+      return null;
+    }
+    return decorView;
+  }
+  // endregion
+
+  // region scroll logic
+  private static class ScrollState {
```
marandaneto, line 193: Can it be final too?

sentry-android-core/src/main/java/io/sentry/android/core/gestures/ViewUtils.java
```diff
@@ -0,0 +1,101 @@
...
+  static @Nullable View findTarget(
+    final @NotNull View decorView,
+    final float x,
+    final float y,
+    final @NotNull ViewTargetSelector viewTargetSelector
+  ) {
+    Queue<View> queue = new ArrayDeque<>();
+    queue.add(decorView);
+
+    @Nullable View target = null;
```
marandaneto, line 25: I don't recall us using annotations within a method but it's not a bad idea.

sentry-android-core/src/main/java/io/sentry/android/core/gestures/ViewUtils.java
marandaneto: I'm afraid of doing reflection on every scroll call, that might cause junk in the UI

sentry-android-core/src/main/java/io/sentry/android/core/gestures/SentryGestureListener.java
romtsn: I think canonical name makes sense here, what if you have a view that you named the same, but it's your custom one (like `io.sentry.Button`)? simple name wouldn't give you a clue what button is this.

sentry-android-core/src/main/java/io/sentry/android/core/UserInteractionIntegration.java
romtsn: not that I know of. We could just check for some random class from androidx.core, that we know should be there and assume that the entire package exists. But what if there's a version, where one of those classes didn't exist? we'd crash in this case

marandaneto: yeah indeed, guess its better like this for now

sentry-android-core/src/main/java/io/sentry/android/core/gestures/ViewUtils.java
romtsn: yeah I realized it a bit later, that we don't do that, but I guess I will just leave it

sentry-android-core/src/main/java/io/sentry/android/core/UserInteractionIntegration.java
romtsn: at first I thought we need this because of this bug https://issuetracker.google.com/issues/186791590 (requires setting a callback before `onCreate`), but I tried it with just onresume/onpause - seems working, will remove

sentry-android-core/src/main/java/io/sentry/android/core/gestures/ViewUtils.java
romtsn: We're only doing this on the first onScroll occurrence, so should be alright?

```Java
if (scrollState.type == null) {
  @SuppressWarnings("Convert2MethodRef")
  final @Nullable View target =
    ViewUtils.findTarget(decorView, firstEvent.getX(), firstEvent.getY(),
      view -> ViewUtils.isViewScrollable(view));

  scrollState.setTarget(target);
  scrollState.type = "scroll";
}
```

marandaneto: So is this being called only once? during the whole App's lifecycle? by reading the code it looks like its being called quite often, but I didn't execute the code.

romtsn: This is called when a user starts scrolling/swiping everytime. But only once, in the beginning, we identify the targetView and store it until the ACTION_UP is performed (that's when we capture a breadcrumb and nullify the target view ref). Next time, the user starts scrolling again, then we again call this method to identify the view target.

sentry-android-core/src/main/java/io/sentry/android/core/UserInteractionIntegration.java
romtsn: filed #1877

sentry-android-core/src/main/java/io/sentry/android/core/gestures/ViewUtils.java
marandaneto: Is there any other way? Did you check how others do it? If not, can we benchmark with and w/o it? I'm afraid that this would be a problem for low end devices.

romtsn: Yeah others do it similarly. I don't think there's any other way to find a target view, you essentially need to traverse the tree to find out.  

Actually, while writing this, I realized we can improve this for the scrollable case, we can return early if the `viewTargetSelector.select` returned true (so we don't need to traverse children of the ScrollView/RecyclerView, if we just hit the ScrollView). This way I think there won't be many iterations for the while-loop, because the scrollviews are usually somewhere on the top of the view hierarchy.

Another improvement - we can pass the androidXAvailable boolean as a parameter, so we don't check `Class.forName` every time, but somewhere once, when the integration gets installed for instance.

As for benchmarking, we could use [macrobencmark lib](https://github.com/android/performance-samples/blob/main/MacrobenchmarkSample/macrobenchmark/src/main/java/com/example/macrobenchmark/FrameTimingBenchmark.kt), but it supports only API 23 and above, and needs to run on a real device (the emulator won't give us much confidence probably)

marandaneto: I believe that the first two optimizations would do for now.

romtsn:
Ready for final review

sentry/src/main/java/io/sentry/Breadcrumb.java
```diff
@@ -210,6 +211,55 @@ public Breadcrumb(final @NotNull Date timestamp) {
...
+   * additional data like {@code viewId} or {@code viewClass}. By default, the breadcrumb is
+   * captured with {@link SentryLevel.INFO} level.
+   *
+   * @param category - the category, for example "click"
+   * @param viewId - the human-readable view id, for example "button_load"
+   * @param viewClass - the fully qualified class name, for example "android.widget.Button"
+   * @param additionalData - additional properties to be put into the data bag
+   * @return the breadcrumb
+   */
+  public static @NotNull Breadcrumb userInteraction(
```
romtsn, line 242: I actually just realized this is in the core-java package - not sure if it's suitable here with all these viewId/viewClass params? We have other static methods if you look above (`user` and `ui`), but none of them really suited my case, because we don't have a message. Should I change one of the above rather?

build.gradle.kts
```diff
@@ -38,6 +38,11 @@ buildscript {
 }
 
 apiValidation {
+    ignoredPackages.addAll(
+        setOf(
+            "io.sentry.android.core.internal"
```
romtsn, line 43: @marandaneto I moved the internal classes under the `internal` package and excluded them from the Api validation, how does that sound?

marandaneto: mmm, What's the reason to do that? it's fine for me but trying to understand it first.
Now we're using the `@ApiStatus.Internal` annotation + internal package + excluded from apiValidation, sounds like we have to chose a single approach at least.

sentry-android-core/src/test/java/io/sentry/android/core/internal/gestures/SentryGestureListenerClickTest.kt
```diff
@@ -0,0 +1,242 @@
...
+import io.sentry.IHub
+import io.sentry.SentryLevel.INFO
+import io.sentry.android.core.SentryAndroidOptions
+import org.junit.Test
+import java.lang.ref.WeakReference
+import kotlin.test.assertEquals
+
+class SentryGestureListenerClickTest {
+    class Fixture {
+        val window: Window = mock()
```
marandaneto, line 27: I'd prefer to keep using the `mock<T>` syntax

sentry-android-core/src/test/java/io/sentry/android/core/internal/gestures/SentryGestureListenerClickTest.kt
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    @Test
+    fun `ignores invisible or gone views`() {
+        val mockEvent: MotionEvent = mock()
+        val invalidTarget: View = mockView(
+            event = mockEvent,
+            visible = false,
+            clickable = true,
+        )
```
marandaneto, line 115: This apparently repeats in all the tests, we could have a method with optional params that return the object, it'd reduce to 1 liner most probably

sentry/src/main/java/io/sentry/Breadcrumb.java
marandaneto: I think viewId and viewClass would be generic for any App that would have user interaction, so its fine.
Maybe we rename the `category` to `subCategory` since we append to the `ui.` prefix?

build.gradle.kts
romtsn: So, the internal package is needed to exclude from apiValidation conveniently (we just specify a package, instead of a list of classes). 

I thought if the classes are internal, they should not be part of the `.api` files - the `@ApiStatus.Internal` has no effect on this, and is just there to indicate for ourselves/users that the class shouldn't be used outside sentry SDK (though the binary compat validator still checks for those classes)

marandaneto: > So, the internal package is needed to exclude from apiValidation conveniently (we just specify a package, instead of a list of classes).

ok, that makes sense.
maybe @bruno-garcia would like to double-check, but I buy it.

romtsn: @bruno-garcia 10% discount for you to buy it

marandaneto (approved):
LGTM

marandaneto:
@romtsn once this is merged, please update the [6.x.x](https://github.com/getsentry/sentry-java/pull/1791) branch with the main branch, which might have conflicts that you'd solve better.

build.gradle.kts
bruno-garcia: wfm: 💵

PR #1886, opened 2022-01-28
Coroutines exception handler

marandaneto:
Closes https://github.com/getsentry/sentry-java/issues/1524

## :pencil: Checklist

- [ ] No breaking changes

### DIFF
--- a/sentry-kotlin-extensions/src/main/java/io/sentry/kotlin/SentryCoroutineExceptionHandler.kt
+++ b/sentry-kotlin-extensions/src/main/java/io/sentry/kotlin/SentryCoroutineExceptionHandler.kt
@@ -9,6 +9,9 @@ import kotlinx.coroutines.CoroutineExceptionHandler
 import kotlin.coroutines.AbstractCoroutineContextElement
 import kotlin.coroutines.CoroutineContext
 
+/**
+ * Sentry exception handler element for [CoroutineExceptionHandler].
+ */
 public class SentryCoroutineExceptionHandler(private val hub: IHub = HubAdapter.getInstance()) : AbstractCoroutineContextElement(CoroutineExceptionHandler), CoroutineExceptionHandler {
     public constructor() : this(HubAdapter.getInstance())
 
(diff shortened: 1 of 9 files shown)


bruno-garcia (approved):

bruno-garcia:
Would be good to add an example here: https://docs.sentry.io/platforms/java/enriching-events/scopes/#kotlin-coroutines

sentry-kotlin-extensions/src/main/java/io/sentry/kotlin/SentryCoroutineExceptionHandler.kt
```diff
@@ -0,0 +1,29 @@
...
+import kotlin.coroutines.CoroutineContext
+
+/**
+ * Sentry exception handler element for [CoroutineExceptionHandler].
+ */
+@ApiStatus.Experimental
+public class SentryCoroutineExceptionHandler(private val hub: IHub = HubAdapter.getInstance()) : AbstractCoroutineContextElement(CoroutineExceptionHandler), CoroutineExceptionHandler {
+    public constructor() : this(HubAdapter.getInstance())
+
+    override fun handleException(context: CoroutineContext, exception: Throwable) {
```
romtsn, line 20: I'm wondering if should check `if (context instanceof SentryContext)` and use the hub from there?

sentry-kotlin-extensions/src/main/java/io/sentry/kotlin/SentryCoroutineExceptionHandler.kt
```diff
@@ -0,0 +1,29 @@
...
+import kotlinx.coroutines.CoroutineExceptionHandler
+import org.jetbrains.annotations.ApiStatus
+import kotlin.coroutines.AbstractCoroutineContextElement
+import kotlin.coroutines.CoroutineContext
+
+/**
+ * Sentry exception handler element for [CoroutineExceptionHandler].
+ */
+@ApiStatus.Experimental
+public class SentryCoroutineExceptionHandler(private val hub: IHub = HubAdapter.getInstance()) : AbstractCoroutineContextElement(CoroutineExceptionHandler), CoroutineExceptionHandler {
```
romtsn, line 17: Should we make it `open` for inheritance? If I have my custom exception handler I'd want to have mine on top of Sentry's, or we advise delegating?

sentry-kotlin-extensions/src/main/java/io/sentry/kotlin/SentryCoroutineExceptionHandler.kt
marandaneto: `HubAdapter` always forwards to `Sentry` static class, it'd not make any difference.

sentry-kotlin-extensions/src/main/java/io/sentry/kotlin/SentryCoroutineExceptionHandler.kt
marandaneto: Probably a good idea, other integrations had to make it open because of that.

PR #1897, opened 2022-02-03
Feature: Android profiling traces

stefanosiano:
## :scroll: Description
adds Android profiling whenever a transaction starts 
adds a new envelope generation method to send profiling data
Bind traces to transactions
Ignore profiling for other transactions when they run at the same time

We are integrating the Specto features in the Sentry sdk. This pr adds support for profiling Android methods whenever a sentry transaction is started

I still have to make specific tests for it

## :pencil: Checklist

- [ ] No breaking changes

## :crystal_ball: Next steps
Finish profiling item payload
Add tests


sentry-android-core/src/main/java/io/sentry/android/core/AndroidTraceTransactionListener.java
```diff
@@ -0,0 +1,148 @@
...
+          .getLogger()
+          .log(SentryLevel.ERROR, "No profiling traces dir path is defined in options.");
+      return;
+    }
+
+    traceFilesDir = new File(tracesFilesDirPath);
+    // Method trace files are normally deleted at the end of traces, but if that fails
+    // for some reason we try to clear any old files here.
+    FileUtils.deleteRecursively(traceFilesDir);
+    traceFilesDir.mkdirs();
```
marandaneto, line 57: Do that on `Sentry#initConfigurations` only if `isProfilingEnabled` is enabled, otherwise we do IO for nothing.

sentry/src/main/java/io/sentry/Hub.java
```diff
@@ -658,6 +658,14 @@ public void flush(long timeoutMillis) {
               startTimestamp,
               waitForChildren,
               transactionFinishedCallback);
+
+      // todo Should the listener be called only when a transaction is sampled?
```
marandaneto, line 662: probably yes

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
```diff
@@ -243,6 +200,76 @@ public static SentryEnvelopeItem fromAttachment(
...
+
+    final CachedItem cachedItem =
+        new CachedItem(
+            () -> {
+              if (!traceFilePath.isEmpty()) {
+                return readEnvelopePayloadFromFile(traceFilePath, maxTraceSize);
+              }
+              throw new SentryEnvelopeException(
+                  String.format(
+                      "Couldn't attach the trace %s.\nPlease check the trace was written correctly",
```
marandaneto, line 217:
```suggestion
                      "Couldn't attach the trace %s.\nPlease check if the trace was written correctly",
```

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -284,6 +284,21 @@
...
+  private @Nullable String profilingTracesDirPath;
+
+  /** Control if profiling is enabled or not for transactions */
+  private boolean profilingEnabled = false;
+
+  /** Max trace file size in bytes. */
+  private long maxTraceFileSize = 5 * 1024 * 1024;
+
+  /** Interval for profiling traces in milliseconds. Defaults to 300 times per second */
+  private int profilingTracesIntervalMs = 1_000 / 300;
```
marandaneto, line 297: other properties follow the `Millis` suffix convention.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTraceTransactionListener.java
```diff
@@ -0,0 +1,148 @@
...
+    if (intervalMs <= 0) {
+      options
+          .getLogger()
+          .log(SentryLevel.DEBUG, "Profiling trace interval is set to %d milliseconds", intervalMs);
+      return;
+    }
+
+    // We stop the trace after 30 seconds, since such a long trace is very probably a trace
+    // that will never end due to an error
+    SentryExecutors.tracingExecutor.schedule(
```
marandaneto, line 94: We already have an executor `SentryOptions#executorService`, should we maybe try to adapt it?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTraceTransactionListener.java
```diff
@@ -0,0 +1,148 @@
...
+            SentryEnvelope.from(
+                new SentryId(),
+                traceFile.getPath(),
+                traceFile.getName(),
+                options.getSdkVersion(),
+                options.getMaxTraceFileSize(),
+                true);
+        Sentry.getCurrentHub().captureEnvelope(envelope);
+      } catch (IOException e) {
+        options.getLogger().log(SentryLevel.ERROR, "Failed to capture session.", e);
```
marandaneto, line 138:
```suggestion
        options.getLogger().log(SentryLevel.ERROR, "Failed to capture the trace.", e);
```

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTraceTransactionListener.java
```diff
@@ -0,0 +1,148 @@
...
+        return;
```
marandaneto, line 139: If we return that early, we don't delete the file for clean up the `activeTransaction`, is that correct?

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -284,6 +284,21 @@
   /** Controls if the `tracestate` header is attached to envelopes and HTTP client integrations. */
   private boolean traceSampling;
 
+  /** The cache dir. path for caching profiling traces */
+  private @Nullable String profilingTracesDirPath;
```
marandaneto, line 288: Should some of these properties be on `SentryAndroidOptions` or do we intend to expand to Java backend/desktop too?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTraceTransactionListener.java
stefanosiano: yep, fixing it.
Will also make the trace file get deleted if anything goes wrong

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTraceTransactionListener.java
stefanosiano: yeah, it's just that the `SentryOptions#getExecutorService` is not public, and i don't know if i can simply change the visibility without problems

marandaneto: If it works, we can make it public and make with the internal annotation

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -284,6 +284,21 @@
...
+
+  /** Control if profiling is enabled or not for transactions */
+  private boolean profilingEnabled = false;
```
marandaneto, line 291: Please also add the support for this flag to the Android Manifest, look at ManifestMetadataReader

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -1471,6 +1486,96 @@ public void setTraceSampling(boolean traceSampling) {
...
+    return transactionListener;
+  }
+
+  /**
+   * Sets the listener interface to perform operations when a transaction is started or ended.
+   *
+   * @param transactionListener - the listener for operations when a transaction is started or ended
+   */
+  public void setTransactionListener(final @NotNull ITransactionListener transactionListener) {
+    this.transactionListener = transactionListener;
```
marandaneto, line 1558: Do not allow null values, set the NoOp otherwise, see how `setLogger` does

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTraceTransactionListener.java
```diff
@@ -0,0 +1,150 @@
...
+      try {
+        SentryEnvelope envelope =
+            SentryEnvelope.from(
+                new SentryId(),
+                traceFile.getPath(),
+                traceFile.getName(),
+                options.getSdkVersion(),
+                options.getMaxTraceFileSize(),
+                true);
+        hub.captureEnvelope(envelope);
```
marandaneto, line 140: The transaction already calls `captureEnvelope` with the transaction item, ideally, this traceFile would be one more item of that envelope.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTraceTransactionListener.java
```diff
@@ -0,0 +1,150 @@
...
+      } catch (IOException e) {
+        options.getLogger().log(SentryLevel.ERROR, "Failed to capture the trace.", e);
+        // We don't return out of this function here, so we can go on and clean the variables
+      }
+    }
+
+    if (traceFile != null) traceFile.delete();
```
marandaneto, line 147: This is likely deleting the file even before the SDK got the chance to send it over to the network since the EnvelopeItem is read async.

sentry/src/main/java/io/sentry/ITransactionListener.java
```diff
@@ -0,0 +1,11 @@
+package io.sentry;
+
+import org.jetbrains.annotations.ApiStatus;
+
+/** Used for performing operations when a transaction is started or ended. */
+@ApiStatus.Internal
+public interface ITransactionListener {
+  void onTransactionStart(ITransaction transaction);
+
+  void onTransactionEnd(ITransaction transaction);
```
marandaneto, line 10: Since we have the method `transaction.finish()` I'd rather call it `onTransactionFinish`

sentry/src/main/java/io/sentry/util/FileUtils.java
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  /**
+   * If the relative path denotes an absolute path, it is returned back. Otherwise it is returned
+   * the file relative to f For instance, `File.resolve(new File("/foo/bar"), "gav")` is `new
+   * File("/foo/bar/gav")` While `File.resolve(new File("/foo/bar"), "/gav")` is `new File("/gav")`.
+   *
+   * @return concatenated this and [relative] paths, or just [relative] if it's absolute.
+   */
+  public static @Nullable File resolve(@Nullable File f, @Nullable String relative) {
```
marandaneto, line 40: Not entirely sure why we need that, isn't `File(folder, "name")` enough?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTraceTransactionListener.java
```diff
@@ -0,0 +1,150 @@
...
+  @SuppressWarnings("FutureReturnValueIgnored")
+  public synchronized void onTransactionStart(ITransaction transaction) {
+
+    // Debug.startMethodTracingSampling() is only available since Lollipop
+    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) return;
+
+    // Let's be sure to end any running trace
+    if (activeTransaction != null) onTransactionEnd(activeTransaction);
+
+    traceFile = FileUtils.resolve(traceFilesDir, "sentry-" + UUID.randomUUID() + ".trace");
```
marandaneto, line 73: If the profiling has its own folder, we don't need the prefix `sentry-`

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTraceTransactionListener.java
```diff
@@ -0,0 +1,150 @@
...
+
+  private @NotNull final SentryOptions options;
+
+  public AndroidTraceTransactionListener() {
+    this(HubAdapter.getInstance());
+  }
+
+  public AndroidTraceTransactionListener(IHub hub) {
+    this.hub = Objects.requireNonNull(hub, "hub is required");
+    options = hub.getOptions();
```
marandaneto, line 52: if `hub.getOptions` works, you don't need to cache the `options` again

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTraceTransactionListener.java
```diff
@@ -0,0 +1,150 @@
...
+  private boolean startedMethodTracing = false;
+  private @Nullable ITransaction activeTransaction = null;
+
+  private @NotNull final SentryOptions options;
+
+  public AndroidTraceTransactionListener() {
+    this(HubAdapter.getInstance());
+  }
+
+  public AndroidTraceTransactionListener(IHub hub) {
```
marandaneto, line 50: please always use `final ` `@NotNull` or `@Nullable` for ctor params, method params etc otherwise static code analysis get unhappy

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTraceTransactionListener.java
```diff
@@ -0,0 +1,150 @@
...
+
+  @Override
+  @SuppressWarnings("FutureReturnValueIgnored")
+  public synchronized void onTransactionStart(ITransaction transaction) {
+
+    // Debug.startMethodTracingSampling() is only available since Lollipop
+    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) return;
+
+    // Let's be sure to end any running trace
+    if (activeTransaction != null) onTransactionEnd(activeTransaction);
```
marandaneto, line 71: I'd say we should use a boolean to control whether profiling is running or not, if a transaction is already running the profiler, the other ones don't need to, since the profiler is a static method from all threads and not from a specific scope.

stefanosiano: Just thought about a possible issue: what if the second transaction (that is ignored by the profiler) ends before the first one?
Using a boolean I cannot distinguish the transaction that is finishing the profiling.
So, i think I have 3 options:
1) keep a reference to the transaction: this means that when the transaction that started the profiling finishes, then the profiling finishes, too, even if another transaction is running
2) use a counter, increasing when a transaction starts and decreasing when it finishes: this means that the profiling is stopped when there are no active transactions
3) ignore it and use a boolean: whenever a transaction finishes, the profiling finishes, too

I would go with 1 or 2, depending on whether we want the profiling to cover all running transactions or only the starting one.
What do you think @marandaneto  ?

sentry/src/main/java/io/sentry/SentryOptions.java
stefanosiano: Should I add only profiling enabled as option from manifest with something like io.sentry.traces.profiling-enable?
Don't think the others should be set by the manifest (profilingTracesDirPath, maxTraceFileSize, profilingTracesIntervalMillis), right?

marandaneto: Usually, we only add to the AndroidManifest the most important ones, the ones that should be customized the most.
I believe, in this case, only the `profilingEnabled` flag.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTraceTransactionListener.java
marandaneto: When transaction A starts the profiler, the profiler should be stopped once transaction A is finished.
Either a reference or an identifier is needed, indeed.

sentry/src/main/java/io/sentry/util/FileUtils.java
stefanosiano: yeah, it's just a "safer" method we use in the specto sdk (not only for this method), but maybe it's cleaner to use just `new File(folder, name)` as it's enough

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTraceTransactionListener.java
stefanosiano: I'm adding it to the same envelope. Just one thing: 
Can I add 2 fields in the `SentryEnvelopeItemHeader`? We are thinking to add a transaction id and the span id the profiling trace refers to.
Of course, I would add the needed logic in `SentryEnvelopeItemHeaderAdapter`, too

sentry/src/main/java/io/sentry/SentryOptions.java
stefanosiano: Ok, will move the profiling options there

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTraceTransactionListener.java
```diff
@@ -0,0 +1,149 @@
...
+
+    traceFile =
+        traceFilesDir == null ? null : new File(traceFilesDir, UUID.randomUUID() + ".trace");
+
+    if (traceFile == null) {
+      hub.getOptions().getLogger().log(SentryLevel.DEBUG, "Could not create a trace file");
+      return;
+    }
+
+    if (traceFile.exists()) {
```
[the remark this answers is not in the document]
stefanosiano: not needed if we check the directory existence before, so i'll remove it

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTraceTransactionListener.java
```diff
@@ -0,0 +1,149 @@
...
+    activeTransaction = transaction;
+
+    // We stop the trace after 30 seconds, since such a long trace is very probably a trace
+    // that will never end due to an error
+    scheduledFinish =
+        hub.getOptions()
+            .getExecutorService()
+            .schedule(() -> lastProfilingData = onTransactionFinish(transaction), 30_000);
+
+    int intervalUs = (int) MILLISECONDS.toMicros(intervalMillis);
```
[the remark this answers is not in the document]
stefanosiano: Since we don't support changing the options dynamically, it's perfectly fine to move it to the constructor

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTraceTransactionListener.java
marandaneto: This should be discussed with the ingest team, if they say that these fields go into the Item header of the envelope, then sure.

sentry/src/main/java/io/sentry/SentryOptions.java
stefanosiano: I moved `profilingTracesDirPath` and `profilingTracesIntervalMillis` to `SentryAndroidOptions`
I kept `profilingEnabled`, `maxTraceFileSize` and `transactionListener` in `SentryOptions`

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTraceTransactionListener.java
stefanosiano: After several discussions we will go with a json payload with all needed variables in it, without any custom headers

sentry/src/main/java/io/sentry/SentryEnvelopeHeader.java
```diff
@@ -15,6 +15,9 @@
 
   private final @Nullable TraceState trace;
 
+  // todo Should I add a TraceProfilingData object here and in SentryEnvelopeHeaderAdapter to allow
```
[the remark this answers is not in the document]
stefanosiano: After several discussions we will go with a json payload with all needed variables in it, without any custom headers

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTraceTransactionListener.java
```diff
@@ -0,0 +1,171 @@
...
+        traceFile,
+        transaction,
+        Build.VERSION.SDK_INT,
+        Build.MANUFACTURER,
+        Build.MODEL,
+        "buildNumber", // ???
+        "android", // ???
+        Build.VERSION.RELEASE,
+        "versionName", // ???
+        "versionCode", // ???
```
romtsn, line 168: We have https://github.com/getsentry/sentry-java/blob/380ea81a7e8b37d3f033e5b7b97885d06fd96fe6/sentry-android-core/src/main/java/io/sentry/android/core/ContextUtils.java#L37 for this, I think you could do the same for `versionName` easily.

romtsn (approved):
Some minor things, but LGTM 🚀 

The tests are still missing, are you going to add them? Also, if the PR is targeting `main`, I think I'd rather target some integration branch for now, if you want to add more stuff (like tests) in other PRs.

romtsn:
Other question (@bruno-garcia @marandaneto) - are we going to make it part of 6.0.0 or earlier? In the former case we'd need to have custom json serializers for profiling data as we got rid of gson in 6.+

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTraceTransactionListener.java
```diff
@@ -0,0 +1,171 @@
...
+import io.sentry.ProfilingTraceData;
+import io.sentry.SentryLevel;
+import io.sentry.util.Objects;
+import java.io.File;
+import java.util.UUID;
+import java.util.concurrent.Future;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public final class AndroidTraceTransactionListener implements ITransactionListener {
```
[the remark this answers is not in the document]
stefanosiano: yep, oversight 😅

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTraceTransactionListener.java
```diff
@@ -0,0 +1,171 @@
...
+    if (traceFilesDir == null || intervalUs == 0) {
+      return;
+    }
+
+    // If a transaction is currently being profiled, we ignore this call
+    if (activeTransaction != null) {
+      options
+          .getLogger()
+          .log(
+              SentryLevel.WARNING,
```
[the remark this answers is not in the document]
stefanosiano: @bruno-garcia suggested `WARNING`, as the user may wonder why a concurrent transaction is not profiled, and since it's an edge case having a warning looked fine

sentry/src/main/java/io/sentry/ProfilingTraceData.java
```diff
@@ -0,0 +1,168 @@
...
+import org.jetbrains.annotations.Nullable;
+
+@ApiStatus.Internal
+public final class ProfilingTraceData {
+
+  // This field is transient so that it's ignored by Gson
+  private final transient @NotNull File traceFile;
+
+  // Device metadata
+  private final int android_api_level;
```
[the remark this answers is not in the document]
stefanosiano: I ran it, but it didn't care at all. If we get rid of Gson then i can create a custom method that creates a json and use camelCase names for the variables

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTraceTransactionListener.java
stefanosiano: Yeah, i've seen it, but it requires a `PackageInfo`, which requires a `Context`, and i don't have it anywhere.
Would it be a good idea to have these kind of fields (which require a `PackageInfo`) initialized in `AndroidOptionsInitializer.init()` and save them in the `SentryAndroidOptions`?

stefanosiano:
> The tests are still missing, are you going to add them? Also, the PR is targeting `main`, I think I'd rather target some integration branch for now, if you want to add more stuff (like tests) in other PRs.

I was going to add them in this pr, after having everything sorted out.
If i have to target an integration branch, should i create a new one or use an existing one?

sentry/src/main/java/io/sentry/ITransactionListener.java
```diff
@@ -0,0 +1,14 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/** Used for performing operations when a transaction is started or ended. */
+@ApiStatus.Internal
+public interface ITransactionListener {
+  void onTransactionStart(@NotNull ITransaction transaction);
+
+  @Nullable
+  ProfilingTraceData onTransactionFinish(@NotNull ITransaction transaction);
```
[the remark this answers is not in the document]
stefanosiano: yeah, the initial idea evolved a bit.
So we can either
1) change the name to something like ITransactionProfiler
2) make the return type more generic
Which one would be better in your opinion?
I think probably number 1, as this interface was created only for a feature that requires Android specific things

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
```diff
@@ -7,6 +7,7 @@
 import io.sentry.SpanStatus;
 import io.sentry.protocol.SdkVersion;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 /** Sentry SDK options for Android */
 public final class SentryAndroidOptions extends SentryOptions {
```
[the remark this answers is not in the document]
stefanosiano: Yes, we decided to put some options only on `SentryAndroidOptions` and in case will ever add profiling to jvm backend, we'd "move" those options to `SentryOptions`, but it could take a while, as we don't have a current plan on that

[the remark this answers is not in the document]
romtsn: ah sorry I missed the comment thread about it, makes sense!

sentry/src/main/java/io/sentry/ITransactionListener.java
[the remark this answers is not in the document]
romtsn: Yeah sounds good. If we need a listener for some other case in the future, we can make it generic then, for now option 1 is fine 👍

romtsn:
> I was going to add them in this pr, after having everything sorted out. If i have to target an integration branch, should i create a new one or use an existing one?

Existing one you mean the current one from this PR? I think I'd create a new one and point this PR to it

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTraceTransactionListener.java
```diff
@@ -0,0 +1,149 @@
...
+
+    if (scheduledFinish != null) {
+      scheduledFinish.cancel(true);
+    }
+
+    if (traceFile == null || !traceFile.exists()) {
+      hub.getOptions()
+          .getLogger()
+          .log(
+              SentryLevel.DEBUG,
```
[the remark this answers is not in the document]
stefanosiano: yep, even if it's an edge case

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -284,6 +284,15 @@
   /** Controls if the `tracestate` header is attached to envelopes and HTTP client integrations. */
   private boolean traceSampling;
 
+  /** Control if profiling is enabled or not for transactions */
+  private boolean profilingEnabled = true;
```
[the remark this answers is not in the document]
bruno-garcia: Profiling will only start/end when a transaction starts and ends so the end goal is for a user to "opt-in to performance" and we'd enable all features. So from that point of view it looks good to me.
That said, we usually introduce features as an opt-in first, so we can have them go out on main releases without much disruption while we get some customer validation, and then flip the default to `true` (opt-out). e.g: #1914

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTraceTransactionListener.java
```diff
@@ -0,0 +1,149 @@
...
+
+    // We stop the trace after 30 seconds, since such a long trace is very probably a trace
+    // that will never end due to an error
+    scheduledFinish =
+        hub.getOptions()
+            .getExecutorService()
+            .schedule(() -> lastProfilingData = onTransactionFinish(transaction), 30_000);
+
+    int intervalUs = (int) MILLISECONDS.toMicros(intervalMillis);
+    Debug.startMethodTracingSampling(traceFile.getPath(), BUFFER_SIZE_BYTES, intervalUs);
```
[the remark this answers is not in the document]
stefanosiano: The only operation that happens in the main thread (as noted by the strict mode) is the `File.exists()` method, called directly and from the `Debug.startMethodTracingSampling` internally.
The actual writing is handled by the system automatically in the background

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTraceTransactionListener.java
```diff
@@ -0,0 +1,149 @@
...
+        return profilingData;
+      }
+      return null;
+    }
+
+    if (activeTransaction != transaction) {
+      return null;
+    }
+
+    Debug.stopMethodTracing();
```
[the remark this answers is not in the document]
stefanosiano: there is some overhead (and possible frame loss), but that's because of profiling, not i/o writing. Moving things to background threads wouldn't make any difference

sentry/src/main/java/io/sentry/SentryOptions.java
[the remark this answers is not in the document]
stefanosiano: There is some overhead when profiling, even if small. ios will make it disable by default, so we will follow and make it opt-in, too

stefanosiano:
I'm writing a small recap on the open points of this pr
Regarding the options: 
I moved `profilingTracesDirPath` and `profilingTracesIntervalMillis` to `SentryAndroidOptions`
I kept `profilingEnabled`, `maxTraceFileSize` and `transactionListener` in `SentryOptions`

In `ISentryClient`
@marandaneto, @bruno-garcia mentioned the `captureTransaction` could have a Map<String, Object> hint refactor in first, then add the profiling data as a hint.

For `ProfilingTraceData`
This represents a json that will be sent as an envelope item (as the payload). If we are goin to get rid of Gson, as @romtsn said, should i create a custom function to transform it into a json or do you already have an alternative to Gson?

The only thing that I'm missing is how to retrieve versionName and versionCode from the application, since I don't have the Context available to get them. Setting them in the options when they are initialized is a viable way?

The envelope is still rejected because the item type is unknown to relay, yet. The pr to enable it is on the way, though. I manually sent the envelope to the backend guy and everything is as expected, so it should be working fine.
Need to add tests, yet, but will create another pr for them

sentry/src/main/java/io/sentry/ProfilingTraceData.java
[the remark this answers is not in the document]
marandaneto: We got rid of Gson on Java v6 https://github.com/getsentry/sentry-java/tree/6.x.x

marandaneto:
> Other question (@bruno-garcia @marandaneto) - are we going to make it part of 6.0.0 or earlier? In the former case we'd need to have custom json serializers for profiling data as we got rid of gson in 6.+

I don't know, it depends on the deadline of this feature (since no one is working actively on v6), if this PR introduces breaking changes, it needs to go on v6 anyway.
@bruno-garcia ?

sentry/src/main/java/io/sentry/NoOpSentryClient.java
```diff
@@ -48,7 +48,8 @@ public SentryId captureEnvelope(@NotNull SentryEnvelope envelope, @Nullable Obje
       @NotNull SentryTransaction transaction,
       @Nullable TraceState traceState,
       @Nullable Scope scope,
-      @Nullable Object hint) {
+      @Nullable Object hint,
+      @Nullable ProfilingTraceData profilingTraceData) {
```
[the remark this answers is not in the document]
marandaneto: Yep, it could be, https://github.com/getsentry/sentry-java/pull/1929

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTraceTransactionListener.java
marandaneto: We can make a sort of abstraction, either on `SentryAndroidOptions` or this class takes the `PackageInfo` via ctor or something, otherwise it cannot be tested it.

marandaneto:
> I'm writing a small recap on the open points of this pr Regarding the options: I moved `profilingTracesDirPath` and `profilingTracesIntervalMillis` to `SentryAndroidOptions` I kept `profilingEnabled`, `maxTraceFileSize` and `transactionListener` in `SentryOptions`
> 
> In `ISentryClient` @marandaneto, @bruno-garcia mentioned the `captureTransaction` could have a Map<String, Object> hint refactor in first, then add the profiling data as a hint.
> 
> For `ProfilingTraceData` This represents a json that will be sent as an envelope item (as the payload). If we are goin to get rid of Gson, as @romtsn said, should i create a custom function to transform it into a json or do you already have an alternative to Gson?

You'd need to follow the ser/deser format used in there, you can just take a look at any other class being serialized.

> The only thing that I'm missing is how to retrieve versionName and versionCode from the application, since I don't have the Context available to get them. Setting them in the options when they are initialized is a viable way?

Answered on https://github.com/getsentry/sentry-java/pull/1897#discussion_r819451413

> The envelope is still rejected because the item type is unknown to relay, yet. The pr to enable it is on the way, though. I manually sent the envelope to the backend guy and everything is as expected, so it should be working fine. Need to add tests, yet, but will create another pr for them

I see, for that reason, profiling should be disabled by default otherwise it'd be incompatible with older self-hosted versions.
Ideally, this moves to Java/Android v6

The PR is still a Draft and there are conflicts, mind solving them first? thanks.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTraceTransactionListener.java
stefanosiano: I ended up adding the `PackageInfo` via ctor

stefanosiano:
For the serialization, should i rebase this branch to the 6.x.x branch and adapt to the ser/deser used there?

marandaneto:
> For the serialization, should i rebase this branch to the 6.x.x branch and adapt to the ser/deser used there?

I'd rather merge to the feature branch you already have.
Work on tests, etc... so we guarantee that nothing else from the SDK is broken.

bruno-garcia:
> > For the serialization, should i rebase this branch to the 6.x.x branch and adapt to the ser/deser used there?
> 
> I'd rather merge to the feature branch you already have. Work on tests, etc... so we guarantee that nothing else from the SDK is broken.

+1 on merging first. Wrt 6.0 perhaps we need a bigger discussion and commitment from folks and set a timeline (making sure it doesn't negatively affect tracing's goals).

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -249,9 +252,18 @@ private static void readDefaultOptionValues(
...
+      final @NotNull Context context, final @NotNull SentryAndroidOptions options) {
     final File cacheDir = new File(context.getCacheDir(), "sentry");
+    final File profilingTracesDir = new File(cacheDir, "profiling_traces");
     options.setCacheDirPath(cacheDir.getAbsolutePath());
+    options.setProfilingTracesDirPath(profilingTracesDir.getAbsolutePath());
+
+    if (options.isProfilingEnabled()) {
+      // Method trace files are normally deleted at the end of traces, but if that fails for some
+      // reason we try to clear any old files here.
+      FileUtils.deleteRecursively(profilingTracesDir);
```
marandaneto, line 264: This might slow down the start-up of Apps, is it always 1 file, or could be N?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
```diff
@@ -0,0 +1,198 @@
...
+    }
+    intervalUs = (int) MILLISECONDS.toMicros(intervalMillis);
+    traceFilesDir = new File(tracesFilesDirPath);
+  }
+
+  @Override
+  public synchronized void onTransactionStart(@NotNull ITransaction transaction) {
+
+    // Debug.startMethodTracingSampling() is only available since Lollipop
+    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) return;
```
marandaneto, line 75: We have a `BuildInfoProvider` class to make this testable

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
```diff
@@ -0,0 +1,198 @@
...
+      return null;
+    }
+
+    return new ProfilingTraceData(
+        traceFile,
+        transaction,
+        Build.VERSION.SDK_INT,
+        Build.MANUFACTURER,
+        Build.MODEL,
+        Build.VERSION.RELEASE,
```
marandaneto, line 192: Similar to https://github.com/getsentry/sentry-java/pull/1897/files#r822077884

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
```diff
@@ -0,0 +1,198 @@
...
+        options.getProguardUuid(),
+        packageInfo != null ? ContextUtils.getVersionName(packageInfo) : "",
+        packageInfo != null ? ContextUtils.getVersionCode(packageInfo) : "",
```
marandaneto, line 195: We could do a single check for `packageInfo` outside of this block, so we spare one comparison.

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
```diff
@@ -243,6 +204,87 @@ public static SentryEnvelopeItem fromAttachment(
...
+      final long maxTraceFileSize,
+      final @NotNull ISerializer serializer)
+      throws SentryEnvelopeException {
+
+    File traceFile = profilingTraceData.getTraceFile();
+    if (!traceFile.exists()) {
+      return null;
+    }
+
+    // The payload of the profiling item is a json that includes the trace file encoded with base64
```
marandaneto, line 218: Usually, base64 is 1/3 bigger, is there a reason to do so? Envelopes are able to send binary format.

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
```diff
@@ -243,6 +204,87 @@ public static SentryEnvelopeItem fromAttachment(
     return new SentryEnvelopeItem(itemHeader, () -> cachedItem.getBytes());
   }
 
+  public static @Nullable SentryEnvelopeItem fromProfilingTrace(
```
marandaneto, line 207: Apparently, this method is executed on the calling thread, for UI transactions, likely the main thread.
The other methods use a `Callable` which is executed only during the event sending which is performed on background, can we do the same here?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
stefanosiano: I could do a single check for `PackageInfo`, but then i would need two boolean checks for version name and code.
I think a null check is fast enough not to cache this check, but i can do it if you think it's better to do so

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
stefanosiano: We agreed with @phacops, our backend guy, that will expect a base64 string.
Also, the payload of the profile envelope is a json, so a string is a natural fit

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
stefanosiano: ok, i will have another look at it

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
stefanosiano: Also, a transaction can be started (and profiled) at any moment, even during the start-up, so some operation will still be needed to do on the main thread

marandaneto: I see.

> i may add a check to avoid the deletion of an empty folder

That sounds an improvement, I'd do it.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
marandaneto: You'd rather assign to 2 different Strings, they are already with the value or "", this is just a nit.

```
String version = "";
String code "";
if (packageInfo != null) {
  // assign values to both strings
}
```

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
marandaneto: I see, since it's a file, I was expecting to be an attachment that could be sent in bytes.
Reading the code now I understand, the envelope type is `profile`, which is a JSON of the `ProfilingTraceData` class, that contains a `stacktrace` field, and this field has the profiling file content in base64.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
marandaneto: We could also do that in the background, using the `executorService`, but that would require atomic operations since a transaction might start while you are still deleting, your call, happy with the improvement above.
Because the empty folder will be deleted and recreated every time, so we spare some bits for 99% of the use cases.

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
stefanosiano: yep

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
stefanosiano: another thing i could do is to call `listFiles()` in the main thread and then delete only them in the background: since their names are uuids there's no problem on possible conflicts with new profiles

marandaneto: Yeah, sounds better, so let's do this.
Don't delete the folder, and recreate it, if empty.
Just delete the inner files, if any.
List the files within the folder and delete them in the background.

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
stefanosiano: I updated it using the `CachedItem` used in the other methods

marandaneto (approved):

stefanosiano:
I had to add few fields for the profiling envelope, so i made a few changes, but these should be the last one!

PR #1898, opened 2022-02-03
fix: Do not include stacktrace into Timber message

### LINKED ISSUE #1177: Timber: Don't include stack trace frames in message
The Android Timber integration at the time it receives the `message` string from the library, already includes the `exception.toString()` into it:

https://github.com/getsentry/sentry-java/blob/5d8208dba4a7eb32696e38047ab21a4c7d435f3d/sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt#L48-L50

https://github.com/JakeWharton/timber/blob/4fd8e2a09c81931e9c9b004f45aea2937835278c/timber/src/main/java/timber/log/Timber.kt#L158

This becomes very noise in Sentry because the frames already show nicely in the stack trace view, but you also get it duplicated (and raw text) above in the message field.

If possible, override a more 'untreated' method of the timber logging interface, so we can format strings ourselves and not include frames in message.

Another proposed improvement is to log the breadcrumb category as `tag`, instead of `Timber`:

https://github.com/getsentry/sentry-java/blob/5d8208dba4a7eb32696e38047ab21a4c7d435f3d/sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt#L28

Other examples:

## Logback
https://github.com/getsentry/sentry-java/blob/5d8208dba4a7eb32696e38047ab21a4c7d435f3d/sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java#L72-L74

https://github.com/getsentry/sentry-java/blob/5d8208dba4a7eb32696e38047ab21a4c7d435f3d/sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java#L117

## M.E.Logging

https://github.com/getsentry/sentry-dotnet/blob/bfb18ac9dbe32b703886a22c72a0ac61a605539b/src/Sentry.Extensions.Logging/SentryLogger.cs#L68-L72

https://github.com/getsentry/sentry-dotnet/blob/bfb18ac9dbe32b703886a22c72a0ac61a605539b/src/Sentry.Extensions.Logging/SentryLogger.cs#L56

Proposed feature:

Option to group per message template. Override fingerprint so that only the temple is used. So all calls to logger at that location are grouped together regardless of where they came from (stack traces).

romtsn:
## :scroll: Description

* Overridden all Timber.Tree methods to have full control over the log message
* Supply `formatted` and `params` fields if the log-call was made with args
* Do not include stackframes into a message

How this looks now:
image

Grouping seems to work as well:
image

Closes #1177 

## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next steps

1. We can't access tag from Timber because it's [package-private](https://github.com/JakeWharton/timber/blob/84b157605552e7b30f0d20e105d383b48ce8a030/timber/src/main/java/timber/log/Timber.java#L377-L386), so I had to drop it.
2. Since we set the Message ourselves, the stacktraces are not reported as breadcrumbs anymore (previously, Timber default implementation would put a stacktrace into a message variable and we would record a breadcrumb with the exception) - should we still do this?

### DIFF
--- a/sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
+++ b/sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
@@ -199,8 +199,8 @@ class SentryTimberTree(
         val level = getSentryLevel(priority)
         val sentryMessage = Message().apply {
             this.message = message
-            if (message != null && args.isNotEmpty()) {
-                this.formatted = message.format(*args)
+            if (!message.isNullOrEmpty() && args.isNotEmpty()) {
+                this.formatted = message?.format(*args)
             }
             this.params = args.map { it.toString() }
         }
--- a/sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
+++ b/sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
@@ -253,7 +253,9 @@ class SentryTimberTree(
                     category = "Timber"
                     message = msg.message
                 }
-                throwableMsg != null -> Breadcrumb.error(throwableMsg)
+                throwableMsg != null -> Breadcrumb.error(throwableMsg).apply {
+                    category = "exception"
+                }
                 else -> null
             }
 
--- a/sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
+++ b/sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
@@ -192,7 +192,7 @@ class SentryTimberTree(
         vararg args: Any?
     ) {
         if (message.isNullOrEmpty() && throwable == null) {
-            return  // Swallow message if it's null and there's no throwable
+            return // Swallow message if it's null and there's no throwable
         }
 
         val level = getSentryLevel(priority)
--- a/sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
+++ b/sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
@@ -17,42 +17,218 @@ class SentryTimberTree(
     private val minBreadcrumbLevel: SentryLevel
 ) : Timber.Tree() {
 
+    /** Log a verbose message with optional format args. */
+    override fun v(
+        message: String?,
+        vararg args: Any?
+    ) {
+        logWithSentry(Log.VERBOSE, null, message, *args)
+    }
+
+    /** Log a verbose exception and a message with optional format args. */
+    override fun v(
+        t: Throwable?,
+        message: String?,
+        vararg args: Any?
+    ) {
+        logWithSentry(Log.VERBOSE, t, message, *args)
+    }
+
+    /** Log a verbose exception. */
+    override fun v(t: Throwable?) {
+        logWithSentry(Log.VERBOSE, t, null)
+    }
+
+    /** Log a debug message with optional format args. */
+    override fun d(
+        message: String?,
+        vararg args: Any?
+    ) {
+        logWithSentry(Log.DEBUG, null, message, *args)
+    }
+
+    /** Log a debug exception and a message with optional format args. */
+    override fun d(
+        t: Throwable?,
+        message: String?,
+        vararg args: Any?
+    ) {
+        logWithSentry(Log.DEBUG, t, message, *args)
+    }
+
+    /** Log a debug exception. */
+    override fun d(t: Throwable?) {
+        logWithSentry(Log.DEBUG, t, null)
+    }
+
+    /** Log an info message with optional format args. */
+    override fun i(
+        message: String?,
+        vararg args: Any?
+    ) {
+        logWithSentry(Log.INFO, null, message, *args)
+    }
+
+    /** Log an info exception and a message with optional format args. */
+    override fun i(
+        t: Throwable?,
+        message: String?,
+        vararg args: Any?
+    ) {
+        logWithSentry(Log.INFO, t, message, *args)
+    }
+
+    /** Log an info exception. */
+    override fun i(t: Throwable?) {
+        logWithSentry(Log.INFO, t, null)
+    }
+
+    /** Log a warning message with optional format args. */
+    override fun w(
+        message: String?,
+        vararg args: Any?
+    ) {
+        logWithSentry(Log.WARN, null, message, *args)
+    }
+
+    /** Log a warning exception and a message with optional format args. */
+    override fun w(
+        t: Throwable?,
+        message: String?,
+        vararg args: Any?
+    ) {
+        logWithSentry(Log.WARN, t, message, *args)
+    }
+
+    /** Log a warning exception. */
+    override fun w(t: Throwable?) {
+        logWithSentry(Log.WARN, t, null)
+    }
+
+    /** Log an error message with optional format args. */
+    override fun e(
+        message: String?,
+        vararg args: Any?
+    ) {
+        logWithSentry(Log.ERROR, null, message, *args)
+    }
+
+    /** Log an error exception and a message with optional format args. */
+    override fun e(
+        t: Throwable?,
+        message: String?,
+        vararg args: Any?
+    ) {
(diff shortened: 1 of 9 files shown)


sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
```diff
@@ -17,42 +17,218 @@ class SentryTimberTree(
...
+            if (message != null && args.isNotEmpty()) {
+                this.formatted = message.format(*args)
+            }
+            this.params = args.mapNotNull { it.toString() }
+        }
 
-        captureEvent(level, tag, message, throwable)
-        addBreadcrumb(level, message)
+        addBreadcrumb(level, sentryMessage)
+        captureEvent(level, sentryMessage, throwable)
```
marandaneto, line 208: The order here matters, first you capture the event and then add the breadcrumb, otherwise, the crumb is redundant.

marandaneto:
> Since we set the Message ourselves, the stacktraces are not reported as breadcrumbs anymore (previously, Timber default implementation would put a stacktrace into a message variable and we would record a breadcrumb with the exception) - should we still do this?

No, the breadcrumb should not care about the stack trace, if there's an exception, we'll capture an event anyway.

marandaneto:
> We can't access tag from Timber because it's [package-private](https://github.com/JakeWharton/timber/blob/84b157605552e7b30f0d20e105d383b48ce8a030/timber/src/main/java/timber/log/Timber.java#L377-L386), so I had to drop it.

That's a breaking change but not sure if it makes any difference or if somebody uses it as a filter on the Issues page.
@bruno-garcia ?

romtsn:
> That's a breaking change but not sure if it makes any difference or if somebody uses it as a filter on the Issues page. @bruno-garcia ?

Yeah, timber isn't great when it comes to setting tags, because you need to invoke a separate method `Timber.tag` before calling the actual logging function. For `DebugTree` it infers the tag from a class name, in theory we could do the same if desired.

sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
romtsn: hm, actually from my tests, the breadcrumb was captured for the next event, is this how it's designed then?

marandaneto: 1. event
you capture the error/message (no breadcrumb about the error/message itself).
then you add the breadcrumb.

2. event
you capture the error/message (this time, there's a breadcrumb about event 1.)

This should be the way to go.

romtsn: Alright, clear. A follow-up question then: if the event was logged without a message, but just with a throwable (e.g. `Timber.e(Throwable("Test"))`), should we add the breadcrumb or not? 

Because the message is null in this case, so we either do not add it, or we put a stacktrace (or part of it) into the breadcrumb message.

marandaneto: If there's a message, let's use the message, otherwise let's create a message out of the throwable, but not with the stack trace, only with the description of the error, and then we change the type and category of this crumb, following the `error` type here https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/

sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
```diff
@@ -17,42 +17,218 @@ class SentryTimberTree(
...
-     * Captures a Sentry Event if the min. level is equal or higher than the min. required level.
-     */
-    override fun log(priority: Int, tag: String?, message: String, throwable: Throwable?) {
         val level = getSentryLevel(priority)
+        val sentryMessage = Message().apply {
+            this.message = message
+            if (message != null && args.isNotEmpty()) {
+                this.formatted = message.format(*args)
+            }
+            this.params = args.mapNotNull { it.toString() }
```
bruno-garcia, line 204: nit: what about we map `null` to `"null"`? Could "collide" with an actual string that was "null" but the goal is to indicate what values where provided vs what wasn't. (%s, %s, %s = null, "bla", null), args would be : ["bla"] so not sure what position had the value vs null?

Alternatively we can omit args, not sure how much difference is in the UI at the end though

sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
bruno-garcia: +1 on what Manoel said here ^ 🚀

sentry-android-timber/src/test/java/io/sentry/android/timber/SentryTimberTreeTest.kt
```diff
@@ -158,6 +160,9 @@ class SentryTimberTreeTest {
     }
 
     @Test
+    @Ignore(
```
bruno-garcia, line 163: Should we make it public then? Just a getter?
Alternatively for testing purposes we could use reflection

bruno-garcia (approved):
❤️

sentry-android-timber/src/test/java/io/sentry/android/timber/SentryTimberTreeTest.kt
romtsn: I meant it's package-private in timber, we can't really get it, aside from reflection. So we have no way to expose it either. :) It's not only for testing, but also for end users - now when they do `Timber.tag("test"); Timber.e("issue")` we have no way to access this tag (that's also what I was talking about in the PR comments thread)

sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
romtsn: Makes total sense. Actually `mapNotNull` is useless here anyway, because, at first, it maps, and then appends the values that are not-null, but in our case we transform everything `toString()` so there will never be a null value. Will change it to `map` 👍 

I've been following the logback implementation though, and there it's filter not-null values, not sure if should be also changed? https://github.com/getsentry/sentry-java/blob/5d8208dba4a7eb32696e38047ab21a4c7d435f3d/sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java#L100

sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
romtsn: Changed it, had to also additionally set category to `exception`, to align with when an error event happens (otherwise it'd show `generic` category as on the SS below)
image

romtsn:
So the way to go with Timber I think we just do not support tags for now, if we get complaints/issues we can offer a reflection-based support (or automatically infer the tag from the classname). How does that sound?

romtsn:
We could also mention it in the docs and link to an issue on github to vote for

marandaneto:
> We could also mention it in the docs and link to an issue on github to vote for

Let's do this, but let's also add the missing tags feat. on the changelog changes.

sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
```diff
@@ -11,48 +11,225 @@ import timber.log.Timber
...
+        }
 
-    /**
-     * Captures a Sentry Event if the min. level is equal or higher than the min. required level.
-     */
-    override fun log(priority: Int, tag: String?, message: String, throwable: Throwable?) {
         val level = getSentryLevel(priority)
+        val sentryMessage = Message().apply {
+            this.message = message
+            if (message != null && args.isNotEmpty()) {
```
marandaneto, line 202: maybe check if `message.isNullOrEmpty()` as well

romtsn:
Added a breaking change note on the changelog and linked it to the gh issue. Merging it when it's green

PR #1904, opened 2022-02-10
ref: Use perl instead of sed for craft scripts

kamilogorek:
Because sed is sad and it breaks.

#skip-changelog


bruno-garcia (approved):
RDLGTM

PR #1909, opened 2022-02-11
Fix potential memory leaks.

maciejwalkowiak:
## :scroll: Description

Fix potential memory leaks - happening either when Sentry works in servlet environment on application undeployment - or when more than one integration is active.

Running a sample application. I am not sure if there is a way to write a test for that. 

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/sentry-spring/src/main/java/io/sentry/spring/SentryInitBeanPostProcessor.java
+++ b/sentry-spring/src/main/java/io/sentry/spring/SentryInitBeanPostProcessor.java
@@ -2,11 +2,14 @@ package io.sentry.spring;
 
 import com.jakewharton.nopen.annotation.Open;
 import io.sentry.EventProcessor;
+import io.sentry.HubAdapter;
+import io.sentry.IHub;
 import io.sentry.ITransportFactory;
 import io.sentry.Integration;
 import io.sentry.Sentry;
 import io.sentry.SentryOptions;
 import io.sentry.SentryOptions.TracesSamplerCallback;
+import io.sentry.util.Objects;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.springframework.beans.BeansException;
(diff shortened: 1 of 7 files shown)


sentry/src/main/java/io/sentry/HostnameCache.java
```diff
@@ -21,12 +21,18 @@
...
+ * <p>HostnameCache is a singleton and its instance should be obtained through {@link
+ * HostnameCache#getInstance()}.
  */
 final class HostnameCache {
   private static final long HOSTNAME_CACHE_DURATION = TimeUnit.HOURS.toMillis(5);
 
   /** Time before the get hostname operation times out (in ms). */
   private static final long GET_HOSTNAME_TIMEOUT = TimeUnit.SECONDS.toMillis(1);
+
+  @Nullable private static HostnameCache INSTANCE;
```
marandaneto, line 34:
```suggestion
  @NotNull private static final HostnameCache INSTANCE = new HostnameCache();
```

sentry/src/main/java/io/sentry/HostnameCache.java
```diff
@@ -41,7 +47,14 @@
   private final @NotNull ExecutorService executorService =
       Executors.newSingleThreadExecutor(new HostnameCacheThreadFactory());
 
-  public HostnameCache() {
+  static @NotNull HostnameCache getInstance() {
+    if (INSTANCE == null) {
+      INSTANCE = new HostnameCache();
+    }
```
marandaneto, line 53:
```suggestion
```

sentry-spring/src/main/java/io/sentry/spring/SentryInitBeanPostProcessor.java
```diff
@@ -68,4 +73,9 @@ public void setApplicationContext(final @NotNull ApplicationContext applicationC
       throws BeansException {
     this.applicationContext = applicationContext;
   }
+
+  @Override
+  public void destroy() {
+    Sentry.close();
```
marandaneto, line 79: Should we have a ctor that takes an `IHub` and use the default ctor with the `HubAdapter` approach? so we can test the `close` call.

sentry/src/main/java/io/sentry/HostnameCache.java
maciejwalkowiak: This will create `HostnameCache` instance when class is loaded, even if the actual instance has never been requested. As a result there can be nobody responsible for closing the thread pool.

marandaneto: That makes sense, thanks.

marandaneto (approved):

PR #1922, opened 2022-02-27
feat(performance): make ActivityFramesTracker public

ninniuz:
## :scroll: Description
Make `ActivityFramesTracker` and `LoadClass` public.

If the Context used in SentryAndroid.init is not an Application then the ActivityFramesTracker integration is not automatically added to the available ones.
It might be possible that applications need to init Sentry in `Context#attachBaseContext` providing base Context instead of the current `Application`.

With this change, integrators can enable `ActivityFramesTracker` integration during configuration init.

I have created two variants od the sample Android app and verified them both against a personal sentry.io project.

## :pencil: Checklist
- [ ] No breaking changes

## :crystal_ball: Next steps
Maybe add a no-args ctor to `ActivityFramesTracker` which internally provisions an instance of `LoadClass`? It feels a bit awkward to have to create that class in the ctor.


sentry-samples/sentry-samples-android/build.gradle.kts
```diff
@@ -84,6 +84,20 @@ android {
...
+            applicationIdSuffix = ".default"
+            versionNameSuffix = "-default"
+        }
+        create("early") {
+            dimension = "init"
+            applicationIdSuffix = ".early"
+            versionNameSuffix = "-early"
+        }
+    }
+
```
marandaneto, line 100:
```suggestion
    flavorDimensions += "init"
    productFlavors {
        create("default") {
            dimension = "init"
            applicationIdSuffix = ".default"
            versionNameSuffix = "-default"
        }
        create("early") {
            dimension = "init"
            applicationIdSuffix = ".early"
            versionNameSuffix = "-early"
        }
    }

```

I'd rather remove the flavors since it makes the build slower, with more stuff to build and test, per flavor.

sentry-samples/sentry-samples-android/src/early/java/io/sentry/samples/android/MyApplication.java
```diff
@@ -0,0 +1,53 @@
+package io.sentry.samples.android;
```
marandaneto, line 1: I'd rather roll back the sample changes, and focus only on the classes visibility.

marandaneto:
@ninniuz please add an entry to the changelog :)

Thanks for doing this.

marandaneto:
@ninniuz Thanks for doing the PR, closing it in favor of https://github.com/getsentry/sentry-java/pull/1931 anyway since we need it asap.
For some reason, git failed pushing to your branch, so I could address the view myself.

ninniuz:
@marandaneto 👍

Issue #1558, opened 2021-06-24
Make hints `Map<String, Object>` instead of only `Object`


marandaneto:
this would make hints a bit more powerful, Python SDK and JS do the same.

PR #1929, opened 2022-03-01
Ref: Make hints Map<String, Object> instead of only Object

marandaneto:
## :scroll: Description
Ref: Make hints Map<String, Object> instead of only Object

Closes https://github.com/getsentry/sentry-java/issues/1558

Adapted unit tests.

## :pencil: Checklist

- [ ] No breaking changes


sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -123,7 +124,11 @@ private void addBreadcrumb(final @NotNull Activity activity, final @NotNull Stri
       breadcrumb.setData("screen", getActivityName(activity));
       breadcrumb.setCategory("ui.lifecycle");
       breadcrumb.setLevel(SentryLevel.INFO);
-      hub.addBreadcrumb(breadcrumb);
+
+      final Map<String, Object> hintMap = new HashMap<>();
+      hintMap.put("Activity", activity);
```
bruno-garcia, line 129: What about something less likely to be used by a user directly like:

```suggestion
      hintMap.put("Android:Activity", activity);
```

We could use such suffix in all our usages

sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserver.java
```diff
@@ -56,7 +58,11 @@ public void onEvent(int eventType, @Nullable String relativePath) {
     // TODO: Only some event types should be pass through?
 
     final CachedEnvelopeHint hint = new CachedEnvelopeHint(flushTimeoutMillis, logger);
-    envelopeSender.processEnvelopeFile(this.rootPath + File.separator + relativePath, hint);
+
+    final Map<String, Object> hintMap = new HashMap<>();
+    hintMap.put("sentrySdkHint", hint);
```
bruno-garcia, line 63: What about this one for those hints we used type check for logic:

```suggestion
    hintMap.put("Sentry:TypeCheckHint", hint);
```

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpInterceptor.kt
```diff
@@ -59,10 +59,17 @@ class SentryOkHttpInterceptor(
...
-            response?.body?.contentLength().ifHasValidLength {
-                breadcrumb.setData("response_body_size", it)
+
+            val hintsMap = mutableMapOf<String, Any?>("request" to request)
+            response?.let {
+                it.body?.contentLength().ifHasValidLength { responseBodySize ->
+                    breadcrumb.setData("response_body_size", responseBodySize)
+                }
+
+                hintsMap["response"] = it
```
bruno-garcia, line 69:
```suggestion
                hintsMap["OkHttp:Response"] = it
```

bruno-garcia (commented):
I skimmed through it and the only question I have is: Do we plan on documenting these hint keys so folks can find/use them in beforeSend/eventProcessors?

Perhaps it'd be best to have them available in public static String's and use some sort of prefixing. To avoid conflict as more hints are added and to be easier to understand what the value represents.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
marandaneto: The idea is that users see it, and not the opposite, the SDK doesn't need it.
Similar to what's done on JS https://docs.sentry.io/platforms/javascript/configuration/filtering/#hints-for-breadcrumbs

sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserver.java
marandaneto: That makes sense since it's internal

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpInterceptor.kt
marandaneto: Similar to https://github.com/getsentry/sentry-java/pull/1929#discussion_r817034293

marandaneto:
> I skimmed through it and the only question I have is: Do we plan on documenting these hint keys so folks can find/use them in beforeSend/eventProcessors?
> 
> Perhaps it'd be best to have them available in public static String's and use some sort of prefixing. To avoid conflict as more hints are added and to be easier to understand what the value represents.

Yes, the idea is to document the considered public ones, as JS does, https://docs.sentry.io/platforms/javascript/configuration/filtering/#hints-for-breadcrumbs

Some of them are already documented, e.g. `originalException`, `syntheticException` and `event_id`.
I'd document tho that we don't support `originalException` and `event_id` since the Event itself has these properties.

We could make a class with those constants or really just document it via docs.

sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/SentryGestureListener.java
```diff
@@ -163,9 +169,13 @@ private void addBreadcrumb(
       className = target.getClass().getSimpleName();
     }
 
+    final Map<String, Object> hintMap = new HashMap<>();
+    hintMap.put("MotionEvent", motionEvent);
```
romtsn, line 173: can we also put the `target` to the hintMap? Might be useful to retrieve some View's attributes

sentry/src/main/java/io/sentry/TypeCheckHint.java
```diff
@@ -0,0 +1,88 @@
...
+
+import org.jetbrains.annotations.ApiStatus;
+
+/** Constants used for Type Check hints. */
+public final class TypeCheckHint {
+
+  @ApiStatus.Internal public static final String SENTRY_TYPE_CHECK_HINT = "sentry:typeCheckHint";
+
+  /** Used for Synthetic exceptions. */
+  public static final String SENTRY_SYNTHETIC_EXCEPTION = "syntheticException";
```
marandaneto, line 11: Kept as it is since its part of the common docs for hints

sentry/src/main/java/io/sentry/TypeCheckHint.java
```diff
@@ -0,0 +1,88 @@
...
+  public static final String ANDROID_CONFIGURATION = "android:configuration";
+  /** Used for System breadcrumbs. */
+  public static final String ANDROID_INTENT = "android:intent";
+  /** Used for Sensor breadcrumbs. */
+  public static final String ANDROID_SENSOR_EVENT = "android:sensorEvent";
+  /** Used for Gesture breadcrumbs. */
+  public static final String ANDROID_MOTION_EVENT = "android:motionEvent";
+  /** Used for View breadcrumbs. */
+  public static final String ANDROID_VIEW = "android:view";
+  /** Used for View breadcrumbs. */
```
romtsn, line 25:
```suggestion
  /** Used for Fragment breadcrumbs. */
```

romtsn (approved):

PR #1936, opened 2022-03-04
Add Timber and Fragment integrations if they are present on the classpath

romtsn:
## :scroll: Description

* Checks if `sentry-android-timber` and `sentry-android-fragment` are on the classpath and adds the integrations automatically.
* Uses default ctors of those. Note, that for fragment we switched the flag to enable performance by default.

Q1 goal

## :pencil: Checklist

- [x] No breaking changes


audkar:
Are these class names excluded from R8 obfuscation? Otherwise these might not be found in Android Release builds at runtime.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -277,4 +291,23 @@ private static boolean isNdkAvailable(final @NotNull IBuildInfoProvider buildInf
...
     return null;
   }
+
+  private static boolean isIntegrationAvailable(
+      final @NotNull String integrationClassName,
+      final @NotNull SentryOptions options,
+      final @NotNull LoadClass loadClass) {
+    boolean isAvailable;
+    try {
+      loadClass.loadClass(integrationClassName);
```
marandaneto, line 301: What if we refactor the `loadClass`, that catches and logs all exceptions and return the boolean only, this block appears duplicated across the code base.

sentry-android-core/src/test/java/io/sentry/android/core/AndroidOptionsInitializerTest.kt
```diff
@@ -336,6 +337,106 @@ class AndroidOptionsInitializerTest {
         assertNull(actual)
     }
 
+    @Test
+    fun `FragmentLifecycleIntegration added to the integration list if available on classpath`() {
+        val mockContext = ContextUtilsTest.mockMetaData(metaData = createBundleWithDsn())
+        val logger = mock<ILogger>()
+        val sentryOptions = SentryAndroidOptions().apply {
+            setDebug(true)
+        }
```
marandaneto, line 346: Can we refactor this class and add the Fixture pattern, so we improve the DRY?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -155,12 +163,18 @@ private static void installDefaultIntegrations(
           new ActivityLifecycleIntegration(
               (Application) context, buildInfoProvider, activityFramesTracker));
       options.addIntegration(new UserInteractionIntegration((Application) context, loadClass));
+      if (isIntegrationAvailable(SENTRY_FRAGMENT_INTEGRATION_CLASS_NAME, options, loadClass)) {
+        options.addIntegration(new FragmentLifecycleIntegration((Application) context));
```
marandaneto, line 167: What if `FragmentLifecycleIntegration` is already added?
Does this run before or after the `configuration` callback?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -155,12 +163,18 @@ private static void installDefaultIntegrations(
...
     } else {
       options
           .getLogger()
           .log(
               SentryLevel.WARNING,
-              "ActivityLifecycle and UserInteraction Integrations need an Application class to be installed.");
+              "ActivityLifecycle, FragmentLifecycle and UserInteraction Integrations need an Application class to be installed.");
+    }
+    if (isIntegrationAvailable(SENTRY_TIMBER_INTEGRATION_CLASS_NAME, options, loadClass)) {
+      options.addIntegration(new SentryTimberIntegration());
```
marandaneto, line 177: Similar to https://github.com/getsentry/sentry-java/pull/1936/files#r820667921
How do I configure `minEventLevel` and `minBreadcrumbLevel` now?

sentry-android-fragment/src/main/java/io/sentry/android/fragment/FragmentLifecycleIntegration.kt
```diff
@@ -20,7 +20,7 @@ class FragmentLifecycleIntegration(
     Integration,
     Closeable {
 
-    constructor(application: Application) : this(application, true, false)
+    constructor(application: Application) : this(application, true, true)
```
marandaneto, line 23: Can we rather flip this flag during the `FragmentLifecycleIntegration` creation? otherwise, it'd require a major bump I believe, cc @bruno-garcia

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
romtsn: sounds good 👍

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
romtsn: Yeah it runs before the config callback - https://github.com/getsentry/sentry-java/blob/626f92f81b9a2c5c71b50fcffe083b6c0c140285/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java#L74

sentry-android-fragment/src/main/java/io/sentry/android/fragment/FragmentLifecycleIntegration.kt
romtsn: Yeah, sounds like a better way for sure 👍

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
marandaneto: The problem is then if they don't see the changelog/migration, and still add it manually, it will lead to duplicated data, how do we prevent that?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
romtsn: That's what we discussed over slack, right - you can't configure them in the first version. The only way for now would be to remove it in the `init` block and add your own, something like:
```Kotlin
SentryAndroid.init(this) { options ->
  options.integrations.removeIf { it is SentryTimberIntegration }
  options.addIntegration(SentryTimberIntegration(minEventLevel,  minBreadcrumbLevel))
}
```

Later on, we could do this through the gradle plugin and bytecode manipuation (inject params into ctor)

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
romtsn: Good point! In theory we could add a `sanitize` method, which runs after `configuration.configure(options)` and removes duplicate integrations. We'd need some kind of a hint or something for the integration though, to distinguish between user-added and auto-added one. Any other ideas or this sounds fine?

marandaneto: Nope, Some SDKs have a `defaultIntegrations` list instead of using a single one, See https://docs.sentry.io/platforms/javascript/configuration/options/#default-integrations
We don't have that on Java/Android, otherwise, it'd be easy to make a difference.
Worth discussing with @bruno-garcia as well.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
marandaneto: Yes, you are right, forgot about that, my bad.
We still have to address the problem of the user adding the integration by themselves, otherwise, they are unable to use the integration with custom values.
Similar to https://github.com/getsentry/sentry-java/pull/1936#discussion_r820667921

What's about users that don't want to use our plugin? They won't be able to configure custom values.
Maybe it's worth adding a boolean to SentryAndroidOptions, `autoInstallTimber` or something that prevents adding the integration via reflection automatically in case it's false, so they just flip the flag and add themselves, this is enabled by default, obviously.
It's similar to the code snippet above, but instead of removing it, you prevent adding it.

romtsn: hmm, should we maybe just move adding those integrations *after* the configure block was called? This way we could check if the integrations were added by the user with custom values and in case they were not we'd add the default ones?

marandaneto: How would the user remove it then if they don't want to? Maybe we should have a short meeting about it, with @bruno-garcia 
Please take a note of all the edge cases discussed in this PR, so we can go over them.

romtsn: they wouldn't add it on the classpath as a dependency in the first place?

marandaneto: If their App or its transitive dependencies do it, it'll be in the classpath, it's an edge case, but the most common scenario from people migrating from the current behavior to the new one.

romtsn: which new behavior do you mean? Sorry, I guess I'm not getting the problem. If we move adding the integrations logic to post-configure block, this should be fine: If they actually used Timber integration, this wouldn't change anything, if they didn't use it (and don't want to use it) and it was on the classpath, well, that's a good indication that they actually don't need it on the classpath :)

marandaneto: Currently, integrations don't get installed.
New behavior, integrations get installed.

The problem I see doing it after the configure block is, the configure block is meant exactly for that, the user has the chance to fiddle with the options, including integrations.
Bypassing that would be wrong.
Let's see what @bruno-garcia says.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
bruno-garcia: Do we have any integration that legit can be added twice? If not, simplest approach might be to filter out duplicates by type before calling `register`. It's a change in behavior that even if we don't have any integration that can be duplicated, users could rely on it but being honest, it's very unlikely so I think we could go ahead without bumping a major here just adding a note to the changelog.

If we use the last registered integration, we have a path to allow the auto configure work in tandem with the manual one. So we add automatically (with defaults) but if users want to provide options, they can add manually and we'll use that instance instead.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
bruno-garcia: I suggested an approach for the dupe instance + manual configuration in the comment above. Looking forward to see if you think it's a valid approach

sentry-android-fragment/src/main/java/io/sentry/android/fragment/FragmentLifecycleIntegration.kt
bruno-garcia: Good point Manoel!

sentry-android-core/build.gradle.kts
```diff
@@ -49,6 +49,7 @@ android {
 
         // We run a full lint analysis as build part in CI, so skip vital checks for assemble tasks.
         checkReleaseBuilds = false
+        disable += "LogNotTimber"
```
marandaneto, line 52: Wondering why we'd need that.

romtsn: Cause we use android.util.Log in the AndroidLogger and since we add Timber dependency now (compileOnly), it ships with lint rules that are complaining about that. I didn't find a way to exclude lint rules from the dependency

marandaneto: I see, can we disable it from the sample App instead?

romtsn: But it's not a sample app, it's what we use to log SDK messages on Android https://github.com/getsentry/sentry-java/blob/main/sentry-android-core%2Fsrc%2Fmain%2Fjava%2Fio%2Fsentry%2Fandroid%2Fcore%2FAndroidLogger.java

marandaneto: Ah ok, my bad.

romtsn:
@marandaneto addressed everything, PTAL once again

romtsn:
> ```kotlin
> static final String SENTRY_FRAGMENT_INTEGRATION_CLASS_NAME =
>       "io.sentry.android.fragment.FragmentLifecycleIntegration";
> 
>   static final String SENTRY_TIMBER_INTEGRATION_CLASS_NAME =
>       "io.sentry.android.timber.SentryTimberIntegration";
> ```
> 
> Are these class names excluded from R8 obfuscation? Otherwise these might not be found in Android Release builds at runtime.

@audkar thanks for raising the concern. Yes, we ship those modules with proguard rules that keep those classes, see [here](https://github.com/getsentry/sentry-java/blob/530d17dec68a4689d4455941a11f2445d8d10116/sentry-android-fragment/proguard-rules.pro#L1) and [here](https://github.com/getsentry/sentry-java/blob/530d17dec68a4689d4455941a11f2445d8d10116/sentry-android-timber/proguard-rules.pro#L3)

sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesTracker.java
```diff
@@ -25,8 +26,10 @@
   private final @NotNull Map<SentryId, Map<String, @NotNull MeasurementValue>>
       activityMeasurements = new ConcurrentHashMap<>();
 
-  public ActivityFramesTracker(final @NotNull LoadClass loadClass) {
-    androidXAvailable = checkAndroidXAvailability(loadClass);
+  public ActivityFramesTracker(
+      final @NotNull LoadClass loadClass, final @NotNull SentryOptions options) {
```
marandaneto, line 30: This would mean that we can't create this class before initing the SDK which gives us the options, can we pass the logger only?

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
```diff
@@ -95,4 +102,36 @@ public static synchronized void init(
...
+  /**
+   * Deduplicate potentially duplicated Fragment and Timber integrations, which can be added
+   * automatically by our SDK as well as by the user. The user's ones win over ours.
+   *
+   * @param options SentryOptions to retrieve integrations from
+   */
+  private static void deduplicateIntegrations(final @NotNull SentryOptions options) {
+    final List<Integration> timberIntegrations = new ArrayList<>();
+    final List<Integration> fragmentIntegrations = new ArrayList<>();
+    for (final Integration integration : options.getIntegrations()) {
```
marandaneto, line 115: `options.getIntegrations()` is a `CopyOnWriteArrayList`, you can likely use an `Iterator` and remove it directly.

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
```diff
@@ -95,4 +102,36 @@ public static synchronized void init(
...
+      if (integration instanceof FragmentLifecycleIntegration) {
+        fragmentIntegrations.add(integration);
+      } else if (integration instanceof SentryTimberIntegration) {
+        timberIntegrations.add(integration);
+      }
```
marandaneto, line 120: This is working because it is `compileOnly`, if you run this code without the dependencies, it will crash the App.
What if we filter out inside of the `SentryTimberIntegration` and `FragmentLifecycleIntegration` class?
The register method has access to the `options`

sentry-android-core/src/test/java/io/sentry/android/core/AndroidOptionsInitializerTest.kt
```diff
@@ -30,345 +25,313 @@ import kotlin.test.assertTrue
 
 @RunWith(AndroidJUnit4::class)
 class AndroidOptionsInitializerTest {
```
marandaneto, line 27: Nice refactoring.

sentry-android-core/src/test/java/io/sentry/android/core/SentryAndroidTest.kt
```diff
@@ -21,6 +26,28 @@ import kotlin.test.assertTrue
 @RunWith(AndroidJUnit4::class)
 class SentryAndroidTest {
 
+    class Fixture {
```
marandaneto, line 29: Nice refactoring.

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
romtsn: oh, indeed... But how would I filter it out inside the integration that is going to be removed, actually? or you mean we'd just not register the integration if it's a duplicate one?

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
romtsn: But we still need to go through the entire list because we are keeping the latest integrations, I wasn't sure how to make it work with the iterator. Unless we reverse the list

sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesTracker.java
marandaneto: Yes, I'd make it nullable with an extra ctor, but I'd still move to a Logger only, Hybrid SDKs have their own logger, but passing an option sounds wrong in this case.

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
marandaneto: Indeed, that's a good point.
If we use an Iterator and traverse using a reverse other, 2 booleans would do the job, but it's a nit, it'd not make much difference anyway, your call.

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
marandaneto: Good point, didn't think about it, maybe just check if it really errors or not if both integrations are not in the classpath, it was a hunch that it'd crash, maybe for `instanceof` it does not.

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
romtsn: decided to keep it as is for now.

sentry-android-core/src/main/java/io/sentry/android/core/LoadClass.java
```diff
@@ -9,10 +13,34 @@
...
+   * @return a Class<?> if it's available, or null
    */
-  public @NotNull Class<?> loadClass(@NotNull String clazz) throws ClassNotFoundException {
-    return Class.forName(clazz);
+  public @Nullable Class<?> loadClass(final @NotNull String clazz, final @Nullable ILogger logger) {
+    try {
+      return Class.forName(clazz);
+    } catch (ClassNotFoundException e) {
+      if (logger != null) {
+        logger.log(SentryLevel.WARNING, "Failed to load class " + clazz, e);
```
bruno-garcia, line 24: I feel this log as `Warning` with a message saying `Failed to` will be a source of confusion to users.
This would be the right behavior whether we were "really trying to find it" but looking from the point of view of `isClassAvailable`, it sounds like just a check: "is it there? It's OK if not..".

Another way to look at this: Do we expect the class to be found in >95% of cases? I believe for timber for example the answer is _no_ so best to use log level Debug here and let the caller decide whether that's a warning or not.

```suggestion
        logger.log(SentryLevel.DEBUG, "Class not available:" + clazz, e);
```

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
```diff
@@ -95,4 +115,46 @@ public static synchronized void init(
...
     }
   }
+
+  /**
+   * Deduplicate potentially duplicated Fragment and Timber integrations, which can be added
+   * automatically by our SDK as well as by the user. The user's ones win over ours.
+   *
+   * @param options SentryOptions to retrieve integrations from
+   */
+  private static void deduplicateIntegrations(
```
bruno-garcia, line 125: This will be a spot where we'll need to "remember" to check for new integrations we add in the future.
What about we dedupe when iterating the list before calling `register`?

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/MyApplication.java
```diff
@@ -16,12 +13,17 @@ public void onCreate() {
 
     // Example how to initialize the SDK manually which allows access to SentryOptions callbacks.
     // Make sure you disable the auto init via manifest meta-data: io.sentry.auto-init=false
-    SentryAndroid.init(
-        this,
-        options -> {
-          options.addIntegration(new FragmentLifecycleIntegration(MyApplication.this, true, true));
-          options.addIntegration(new SentryTimberIntegration());
-        });
+    // SentryAndroid.init(
```
bruno-garcia, line 16: Is this supposed to be uncommented or deleted before we merge?

sentry-samples/sentry-samples-android/src/main/AndroidManifest.xml
```diff
@@ -64,7 +64,7 @@
         <!--    <meta-data android:name="io.sentry.anr.enable" android:value="false" />-->
 
         <!--    how to disable the auto-init-->
-        <meta-data android:name="io.sentry.auto-init" android:value="false" />
+        <!--    <meta-data android:name="io.sentry.auto-init" android:value="false" />-->
```
bruno-garcia, line 67: This feature conflicts with the auto install? Curious why this change in this PR

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/MyApplication.java
romtsn: This was a suggestion from Manoel https://github.com/getsentry/sentry-java/pull/1936#discussion_r820670053, this way we could show those who are looking at the sample code that it's still possible to manually init the sdk

sentry-samples/sentry-samples-android/src/main/AndroidManifest.xml
romtsn: similar to https://github.com/getsentry/sentry-java/pull/1936#discussion_r826381968

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
romtsn: mm, we could, but what would be the benefit? even if we add a new integration, we do not touch the code around `register`, right? + we couldn't reuse the LoadClass result there which is costly on Android due to reflection.

marandaneto (approved):

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/MyApplication.java
bruno-garcia: Sample code can lode outside of production code. I think this will end up becoming a source of maintenance

romtsn: that is also true for the commented manifest entries now. I think I'd keep it like that, find+replace on text should reveal this code as well :)

Issue #1840, opened 2021-12-11
if `transaction.finish` was called, we should not allow calling `startChild`


marandaneto:
https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/SentryTracer.java#L134-L157

we'd need to check as we do on `finish` https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/SentryTracer.java#L209

@philipphofmann confirm on iOS, this could avoid concurrent problems like https://github.com/getsentry/sentry-dart/issues/662

marandaneto:
@bruno-garcia worth checking .NET too

philipphofmann:
@marandaneto, what do you exactly mean by not allowing to call `startChild`?

marandaneto:
@philipphofmann If the transaction is finished, we should not allow adding one more child to the children list, which means, we've to check if the transaction is already finished before executing the `startChild` method or bail at the very beginning.

philipphofmann:
Yes, but was does not allowing actually do? Does the transaction return a no-op child? Does it throw an exception?

marandaneto:
> Yes, but was does not allowing actually do? Does the transaction return a no-op child? Does it throw an exception?

I'd say it returns a no-op child, but the issue has a `needs discussion` tag, that's open.

marandaneto:
We're gonna return NoOp if the transaction was already finished, similar to implemented Dart solution, We should also log out when that happens, so the user knows why and what to do, which is also done for every setter on the span, e.g. setData, etc...

PR #1940, opened 2022-03-08
Fix: If transaction or span is finished, do not allow to mutate

marandaneto:
## :scroll: Description
Fix: If transaction or span is finished, do not allow to mutate

Closes https://github.com/getsentry/sentry-java/issues/1840

## :pencil: Checklist

- [X] No breaking changes

### DIFF
--- a/sentry/src/main/java/io/sentry/SentryTracer.java
+++ b/sentry/src/main/java/io/sentry/SentryTracer.java
@@ -137,6 +137,10 @@ public final class SentryTracer implements ITransaction {
       final @NotNull String operation,
       final @Nullable String description,
       @Nullable Date timestamp) {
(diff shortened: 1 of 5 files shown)


romtsn (approved):

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -379,6 +415,10 @@ public void setName(@NotNull String name) {
   @Deprecated
   @ApiStatus.ScheduledForRemoval
   public void setRequest(final @Nullable Request request) {
+    if (root.isFinished()) {
+      return;
```
bruno-garcia, line 419: Should we log a warning? It seems like an incorrect instrumentation

PR #1949, opened 2022-03-14
tests for Android profiling

stefanosiano:
## :scroll: Description
We added profiling to the android sdk. Now let's add tests!

Unit tests on everything touched by last pr

## :pencil: Checklist

- [ ] No breaking changes

### DIFF
--- a/Sentry/src/test/java/io/sentry/NoOpTransactionProfilerTest.kt
+++ b/Sentry/src/test/java/io/sentry/NoOpTransactionProfilerTest.kt
@@ -4,10 +4,11 @@ import kotlin.test.Test
 import kotlin.test.assertNull
 
 class NoOpTransactionProfilerTest {
-    private var profiler: NoOpTransactionProfiler = NoOpTransactionProfiler.getInstance()
+    private var profiler = NoOpTransactionProfiler.getInstance()
 
     @Test
-    fun `onTransactionStart is no op`() = profiler.onTransactionStart(NoOpTransaction.getInstance())
+    fun `onTransactionStart does not throw`() =
+        profiler.onTransactionStart(NoOpTransaction.getInstance())
 
     @Test
     fun `onTransactionFinish returns null`() =
--- a/Sentry/src/test/java/io/sentry/util/FileUtilsTest.kt
+++ b/Sentry/src/test/java/io/sentry/util/FileUtilsTest.kt
@@ -1,41 +1,60 @@
 package io.sentry.util
 
-import org.junit.Test
 import java.io.File
 import java.nio.file.Files
+import kotlin.test.Test
 import kotlin.test.assertEquals
 import kotlin.test.assertFalse
 import kotlin.test.assertNull
+import kotlin.test.assertTrue
 
 class FileUtilsTest {
 
     @Test
-    fun `deleteRecursively returns true on non-existing file or null`() {
-        assert(FileUtils.deleteRecursively(null))
-        assert(FileUtils.deleteRecursively(File("")))
+    fun `deleteRecursively returns true on null file`() {
+        assertTrue(FileUtils.deleteRecursively(null))
+    }
+
+    @Test
+    fun `deleteRecursively returns true on non-existing file`() {
+        assertTrue(FileUtils.deleteRecursively(File("")))
     }
 
     @Test
     fun `deleteRecursively deletes a simple file`() {
         val f = Files.createTempFile("here", "test").toFile()
-        assert(f.exists())
-        assert(FileUtils.deleteRecursively(f))
+        assertTrue(f.exists())
+        assertTrue(FileUtils.deleteRecursively(f))
         assertFalse(f.exists())
     }
 
     @Test
     fun `deleteRecursively deletes a folder`() {
-        val d = Files.createTempDirectory("here").toFile()
-        val f = File(d, "test")
-        val d2 = File(d, "dir2")
-        val f2 = File(d2, "test")
-        f.createNewFile()
-        d2.mkdir()
-        f2.createNewFile()
-        assert(d.exists() && d.isDirectory && f.exists() && d2.exists() && d2.isDirectory)
-        assert(f2.exists())
-        assert(FileUtils.deleteRecursively(d))
-        assertFalse(f.exists() || d.exists() || f2.exists() || d2.exists())
+        // Setup vars
+        val rootDir = Files.createTempDirectory("here").toFile()
+        val rootChild = File(rootDir, "test")
+        val innerDir = File(rootDir, "dir2")
+        val innerChild = File(innerDir, "test")
+
+        // Create directories and files
+        rootChild.createNewFile()
+        innerDir.mkdir()
+        innerChild.createNewFile()
+
+        // Assert dirs and files exist
+        assertTrue(rootDir.exists() && rootDir.isDirectory)
+        assertTrue(rootChild.exists())
+        assertTrue(innerDir.exists() && innerDir.isDirectory)
+        assertTrue(innerChild.exists())
+
+        // Assert deletion returns true
+        assertTrue(FileUtils.deleteRecursively(rootDir))
+
+        // Assert dirs and files no longer exist
+        assertFalse(rootChild.exists())
+        assertFalse(rootDir.exists())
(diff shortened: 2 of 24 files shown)


stefanosiano:
I still need to add tests for `AndroidTransactionProfiler` (will be here by tomorrow, I think)
Other than that, i should have tested everything else (and found few details to fix, too)

Sentry/src/test/java/io/sentry/NoOpTransactionProfilerTest.kt
```diff
@@ -0,0 +1,15 @@
+package io.sentry
+
+import kotlin.test.Test
+import kotlin.test.assertNull
+
+class NoOpTransactionProfilerTest {
+    private var profiler: NoOpTransactionProfiler = NoOpTransactionProfiler.getInstance()
```
marandaneto, line 7:
```suggestion
    private val profiler = NoOpTransactionProfiler.getInstance()
```

Sentry/src/test/java/io/sentry/NoOpTransactionProfilerTest.kt
```diff
@@ -0,0 +1,15 @@
...
+
+    @Test
+    fun `onTransactionStart is no op`() = profiler.onTransactionStart(NoOpTransaction.getInstance())
```
marandaneto, line 10: There's no assert here, What does the test do?

Sentry/src/test/java/io/sentry/util/FileUtilsTest.kt
```diff
@@ -0,0 +1,58 @@
+package io.sentry.util
+
+import org.junit.Test
```
marandaneto, line 3:
```suggestion
import kotlin.test.Test
```

Sentry/src/test/java/io/sentry/util/FileUtilsTest.kt
```diff
@@ -0,0 +1,58 @@
...
+import java.io.File
+import java.nio.file.Files
+import kotlin.test.assertEquals
+import kotlin.test.assertFalse
+import kotlin.test.assertNull
+
+class FileUtilsTest {
```
marandaneto, line 10: Please use the `assertX` methods from `kotlin.test` instead of `assert(x)`

Sentry/src/test/java/io/sentry/util/FileUtilsTest.kt
marandaneto: Always use the `kotlin.test` package

Sentry/src/test/java/io/sentry/util/FileUtilsTest.kt
```diff
@@ -0,0 +1,58 @@
...
+
+    @Test
+    fun `deleteRecursively returns true on non-existing file or null`() {
+        assert(FileUtils.deleteRecursively(null))
+        assert(FileUtils.deleteRecursively(File("")))
```
marandaneto, line 15: I'd split up in different tests

Sentry/src/test/java/io/sentry/util/FileUtilsTest.kt
```diff
@@ -0,0 +1,58 @@
...
+        assert(FileUtils.deleteRecursively(f))
+        assertFalse(f.exists())
+    }
+
+    @Test
+    fun `deleteRecursively deletes a folder`() {
+        val d = Files.createTempDirectory("here").toFile()
+        val f = File(d, "test")
+        val d2 = File(d, "dir2")
+        val f2 = File(d2, "test")
```
marandaneto, line 31: I'd give proper names to the variables, and jump a line to show the intention of your tests.

```
// setup vars

// call methods

// assert values
```

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
```diff
@@ -41,7 +43,7 @@
   private @Nullable File traceFile = null;
   private @Nullable File traceFilesDir = null;
   private @Nullable Future<?> scheduledFinish = null;
-  private volatile @Nullable ITransaction activeTransaction = null;
+  @VisibleForTesting volatile @Nullable ITransaction activeTransaction = null;
```
marandaneto, line 46: Can we test the class without `VisibleForTesting`? I see this almost in every class.

Sentry/src/test/java/io/sentry/NoOpTransactionProfilerTest.kt
stefanosiano: yeah, I copied it from other NoOp...Test classes
Changed it to `assertEquals(Unit, profiler.onTransactionStart(NoOpTransaction.getInstance()))`

Sentry/src/test/java/io/sentry/util/FileUtilsTest.kt
stefanosiano: Alright, but can you tell me what's the difference, please?
I saw several tests using `org.junit.Test`, too. Is it a mistake?

marandaneto: Yep, they are all wrong, it's just an Alias to be honest, but I'd rather stick to Kotlin packages, so Kotlin first in tests

marandaneto: Raise a new issue to replace all `org.junit` please, or change in this PR ^^

Sentry/src/test/java/io/sentry/NoOpTransactionProfilerTest.kt
marandaneto: Maybe we rename to `methodX does not throw`, since we cannot assert anything to tell its `no op` or not, right? unless you can.

stefanosiano: yeah, i cannot assert anything, so will do that `methodX does not throw`

Sentry/src/test/java/io/sentry/util/FileUtilsTest.kt
stefanosiano: Same goes with 
`org.junit.Ignore` -> `kotlin.test.Ignore`
`org.junit.Before` -> `kotlin.test.BeforeTest`
`org.junit.After` -> `kotlin.test.AfterTest`
right?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
stefanosiano: Avoiding it would make it harder/less readable to test the code without increasing the visibility of the field.
I realized that when using `org.jetbrains.annotations.VisibleForTesting` Android studio doesn't care whether you use it in tests or outside them, and builds in any case.  
However, using `androidx.annotation.VisibleForTesting` makes Android studio complain if you access that field outside tests.  
Of course, that would work in sentry-android-* only. Would it make any difference?

bruno-garcia:
> ❗ No coverage uploaded for pull request base (feat/profiling/android@6a53471). [Click here to learn what that means](https://docs.codecov.io/docs/error-reference?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=getsentry#section-missing-base-commit).

Odd that codecov didn't work

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
marandaneto: The problem is that we could not use `androidx.annotation` in non-Android projects and that would make 2 different packages of annotations across the project, so we preferred to keep the standard.

You can configure AS to warn in such cases well.

You can also test the behavior of the method without using `VisibleForTesting`, for example, when you finish the transaction 2 times, you can `verify` that the `captureTransaction` method was called only once, so the `activeTransaction` was actually null during the 2nd time, instead of asserting the field directly.

Usually `VisibleForTesting` is only necessary when you are working with static fields or when its technically not possible to test the behaviour, it does not like this use case for this field.

marandaneto (approved):

PR #1951, opened 2022-03-16
fix codecov badge

bruno-garcia:
We've had the wrong badge for over a year :D

Good news, it's not 65% anymore:
image

_#skip-changelog_


marandaneto (approved):
LOL, good catch @bruno-garcia

PR #1959, opened 2022-03-24
all operations involving file reads were moved to background threads

stefanosiano:
## :scroll: Description
All operations involving IO have been moved to the background, as enabling the strict mode on Android was causing crashes

### DIFF
--- a/sentry/src/main/java/io/sentry/ProfilingTraceData.java
+++ b/sentry/src/main/java/io/sentry/ProfilingTraceData.java
@@ -1,9 +1,11 @@
 package io.sentry;
 
 import java.io.File;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
 import java.util.UUID;
+import java.util.concurrent.Callable;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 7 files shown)


sentry/src/main/java/io/sentry/ProfilingTraceData.java
```diff
@@ -189,4 +192,12 @@ public boolean isDevice_is_emulator() {
   public void setSampled_profile(@Nullable String sampledProfile) {
     this.sampled_profile = sampledProfile;
   }
+
+  public void readDeviceCpuFrequencies() {
+    try {
+      this.device_cpu_frequencies = deviceCpuFrequenciesReader.call();
+    } catch (Exception e) {
+      this.device_cpu_frequencies = new ArrayList<>();
```
marandaneto, line 200: Why do we need that? `device_cpu_frequencies` is already created in the field declaration.

stefanosiano: It's just because I cannot leave exception block empty. Since there's no logger to log the exception i don't know if it's fine to call `e.printStackTrace()`

marandaneto: You can, do this:

```
} catch (Throwable ignored) {
      // should never happen
}
```

just name it `ignored` and add a comment.

stefanosiano: I previously did
```
} catch (Throwable ignored) {
}
```
without the comment and it was complaining. I added the comment and now it's fine, thanks!

marandaneto (approved):

PR #1974, opened 2022-04-05
Fix: Add TimberTag

### LINKED ISSUE #1900: Support Timber tags
We have no way to get the tag when inheriting `Timber.Tree`, because it's marked as internal/package-private for consumers https://github.com/JakeWharton/timber/blob/dc45d69b081a03e3fb83e81a7aee0f430f46b4f1/timber/src/main/java/timber/log/Timber.kt#L22. This means, when timber is used like this:

```Kotlin
Timber.tag("Tag")
Timber.e("Issue")
```
the tag will not be reflected on Sentry.

If this gets traction, we could either offer a reflection-based support or automatically infer classname of the caller (that's what `Timber.DebugTree` does by default).

Please, vote on this issue if you'd like us to provide support for Timber tags.

audkar:
## :scroll: Description

Brings back `TimberTag`. Closes #1900

We are relying heavily on `TimberTag` value. One of our dependencies is abusing Timber logging. We were using this tag to filter out these events. This is a blocker to upgrade Sentry for us.

I decided not to use reflection, since retrieving value is not enough. We would also have to write `tag` clearing logic via reflection.

Pros: 
- This approach will not break if internal Timber's tag management logic change in the future.
- No need to write tag clearing logic

Cons:
- Timber will format `message` string, which we don't use (performance)

Unit test

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Fix: bring back support for `Timber.tag` ([#1974](https://github.com/getsentry/sentry-java/pull/1974))
+
 ## 5.7.1
 
 * Fix: Sentry Timber integration does not submit msg.formatted breadcrumbs (#1957)
--- a/sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
+++ b/sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
@@ -17,12 +17,22 @@ class SentryTimberTree(
     private val minEventLevel: SentryLevel,
     private val minBreadcrumbLevel: SentryLevel
 ) : Timber.Tree() {
+    private val pendingTag = ThreadLocal<String?>()
+
+    private fun retrieveTag(): String? {
+        val tag = pendingTag.get()
+        if (tag != null) {
+            this.pendingTag.remove()
+        }
+        return tag
+    }
 
     /** Log a verbose message with optional format args. */
     override fun v(
         message: String?,
         vararg args: Any?
     ) {
+        super.v(message, *args)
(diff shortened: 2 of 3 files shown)


romtsn:
Thanks for the PR!

> Cons:
>
> * Timber will format message string, which we don't use (performance)

Unfortunately, that's why we've made that PR in the first place and gone away from the `log` method of timber. This was resulting in bad user experience, since Timber is adding the stacktrace to the message and this was duplicated on Sentry (since Sentry parses and displays the stacktrace nicely already). 👇 

image

So, I'm unsure if we can accept this solution, @marandaneto thoughts?

audkar:
> Unfortunately, that's why we've made that PR in the first place and gone away from the log method of timber. 

I think you misunderstood me. That message is not going to be logged into Logcat (added to message)

Previously you ware using `message` from this Timber method. After change you made this no-op. This PR doesn't change this behavior. 
```kotlin
 override fun log(
        priority: Int,
        tag: String?,
        message: String,
        t: Throwable?
    ) {
        // no-op as we've overridden all the methods
    }
```
:point_up: this `message` contains merged "message+stacktrace". Downside is that Timber still creates this variable. But it is not logged in Sentry envents.

audkar:
"Changelog" CI Action is failing. Do I need to to something?

romtsn:
> "Changelog" CI Action is failing. Do I need to to something?

Yeah, add a changelog entry mentioning PR id, something like:
```markdown
* Fix: bring back support for `Timber.tag` (#1974)
```

marandaneto:
@romtsn We'd need to update docs, sine we mention the removal of this feature.
@audkar thanks for doing this.

PR #1975, opened 2022-04-05
feat: UI events transactions

### LINKED ISSUE #1811: Timeout for `waitForChildren` feature
Could avoid issues like https://github.com/getsentry/sentry-java/issues/1787

romtsn:
## :scroll: Description

* Adds `idleTimeout` to `SentryTracer` to finish idled transaction
* Creates a new transaction whenever a UI action is performed (click, swipe, scroll).

Q1 Goal
Closes #1811

## :pencil: Checklist

- [ ] No breaking changes


sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -211,30 +246,33 @@ public void finish() {
     this.finish(this.getStatus());
   }
 
+  @SuppressWarnings({"JdkObsolete", "JavaUtilDate"})
   @Override
   public void finish(@Nullable SpanStatus status) {
+    cancelTimer();
```
romtsn, line 252: note - we won't re-schedule the timer when a new span gets added, but rather continuously wait until the last added span finishes, I kinda followed dart implementation here https://github.com/getsentry/sentry-dart/blob/28c88c966e2996591ea4363342080d9a003ad072/dart/lib/src/sentry_tracer.dart#L56-L58. Dunno if that's what we want exactly?

romtsn:
> @romtsn This should target `6.x.x`

Why?

marandaneto:
> > @romtsn This should target `6.x.x`
> 
> Why?

`Hub` interface is a public API and it has changed.

sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
```diff
@@ -54,13 +54,15 @@
...
       "io.sentry.traces.activity.auto-finish.enable";
+  static final String TRACES_UI_ENABLE = "io.sentry.traces.user-interaction.enable";
 
   @ApiStatus.Experimental static final String TRACE_SAMPLING = "io.sentry.traces.trace-sampling";
 
   static final String TRACING_ORIGINS = "io.sentry.traces.tracing-origins";
 
   static final String ATTACH_THREADS = "io.sentry.attach-threads";
   static final String PROGUARD_UUID = "io.sentry.proguard-uuid";
+  static final String IDLE_TIMEOUT = "io.sentry.idle-timeout";
```
marandaneto, line 65:
```suggestion
  static final String IDLE_TIMEOUT = "io.sentry.traces.idle-timeout";
```

sentry-android-core/src/main/java/io/sentry/android/core/UserInteractionIntegration.java
```diff
@@ -55,8 +55,13 @@ private void startTracking(final @Nullable Window window, final @NotNull Context
 
       final SentryGestureListener gestureListener =
           new SentryGestureListener(
-              new WeakReference<>(window), hub, options, isAndroidXScrollViewAvailable);
-      window.setCallback(new SentryWindowCallback(delegate, context, gestureListener, options));
+              new WeakReference<>(window),
+              new WeakReference<>(activity),
```
marandaneto, line 59: We can get the window thru the `activity.window` right, so we don't need to pass both?

sentry-android-core/src/main/java/io/sentry/android/core/UserInteractionIntegration.java
```diff
@@ -119,6 +125,10 @@ public void register(@NotNull IHub hub, @NotNull SentryOptions options) {
             "UserInteractionIntegration enabled: %s",
             this.options.isEnableUserInteractionBreadcrumbs());
 
+    this.performanceEnabled =
+        options.isTracingEnabled()
+            && ((SentryAndroidOptions) options).isEnableUserInteractionTracing();
```
marandaneto, line 130: Why do you need to cast to `SentryAndroidOptions`? `register` already does this casting.

sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/SentryGestureListener.java
```diff
@@ -168,6 +184,58 @@ private void addBreadcrumb(
             eventType, ViewUtils.getResourceId(target), className, additionalData));
   }
 
+  private void startTracing(final @NotNull View target, final @NotNull String eventType) {
+    if (!performanceEnabled) {
+      return;
+    }
```
marandaneto, line 190: Check the options instead?

sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/SentryGestureListener.java
```diff
@@ -168,6 +184,58 @@ private void addBreadcrumb(
...
+    final Activity activity = currentActivity.get();
+    if (activity == null) {
+      options.getLogger().log(SentryLevel.DEBUG, "Activity is null, no transaction captured");
+      return;
+    }
+
+    // as we allow a single UI transaction running on the bound Scope, we finish the previous one
+    if (activeTransaction != null) {
+      activeTransaction.finish();
+    }
```
marandaneto, line 201: Wondering if that makes sense, I'd rather let the current one run, we can discuss that today.

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -298,6 +298,16 @@
...
 
+  /**
+   * The idle time, measured in ms, to wait until the transaction will be finished. The transaction
+   * will use the end timestamp of the last finished span as the endtime for the transaction.
+   *
+   * <p>When set to {@code null} the transaction must be finished manually.
+   *
+   * <p>The default is 3 seconds.
+   */
+  private @Nullable Long idleTimeout = 3000L;
```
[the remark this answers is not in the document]
marandaneto: It's fine, since it can be used by the Java SDK for Desktop as well.

sentry/src/main/java/io/sentry/SentryTracer.java
marandaneto: Good question, let's discuss the trade-offs today.

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -483,4 +527,19 @@ private FinishStatus(final boolean isFinishing, final @Nullable SpanStatus spanS
...
+    @SuppressWarnings({"JdkObsolete", "JavaUtilDate"})
+    @Override
+    public int compare(Span o1, Span o2) {
+      final Date first =
+          Objects.requireNonNull(
+              o1.getTimestamp(), "The span timestamp should not be null after it was finished");
+      final Date second =
+          Objects.requireNonNull(
+              o2.getTimestamp(), "The span timestamp should not be null after it was finished");
+      return first.compareTo(second);
```
marandaneto, line 542: That means it'll throw and we'll lose the whole transaction?

sentry/src/main/java/io/sentry/config/PropertiesProvider.java
```diff
@@ -74,4 +74,16 @@ default Double getDoubleProperty(final @NotNull String property) {
...
+  /**
+   * Resolves a {@link Long} property given by it's name.
+   *
+   * @param property - the property name
+   * @return property value or the default value if not found.
+   */
+  @Nullable
+  default Long getLongProperty(final @NotNull String property) {
+    final String result = getProperty(property);
+    return result != null ? Long.valueOf(result) : null;
```
marandaneto, line 87: That can throw `NumberFormatException`, Maybe we wrap in a try catch?

sentry-android-core/src/main/java/io/sentry/android/core/UserInteractionIntegration.java
```diff
@@ -28,6 +27,7 @@
 
   private final boolean isAndroidXAvailable;
   private final boolean isAndroidXScrollViewAvailable;
+  private boolean performanceEnabled = false;
```
marandaneto, line 30: I don't see this var changing its value, or do I miss something?

sentry/src/main/java/io/sentry/SentryTracer.java
romtsn: Discussed [here](https://www.notion.so/sentry/UI-events-Transactions-26057935c07c43c9981f2f62ac48780c) and agreed to keep it like in the dart SDK.

sentry-android-core/src/main/java/io/sentry/android/core/UserInteractionIntegration.java
romtsn: yup, you're right, can be just activity

marandaneto:
> @marandaneto mm, but only methods marked as `Internal` have changed. Only if we care about ABI, then it makes sense.

Indeed, but I'd rather point to v6 to speed up this major release, by that I mean more people working on the new code base and testing out all the changes.
Unless @bruno-garcia says otherwise.

sentry-android-core/src/main/java/io/sentry/android/core/UserInteractionIntegration.java
romtsn: mm, IDE/build was complaining so I had to do it. But I think I will remove it altogether in favour of your suggestion below (just checking the options directly instead of having this obscure boolean flag)

marandaneto: Likely because you were using `options` instead of `this.options`, after casting.

sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/SentryGestureListener.java
romtsn: yeah, as we discussed If it's the same view -> we keep the transaction running and finish if it's a new view.

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -282,6 +320,12 @@ public void finish(@Nullable SpanStatus status) {
     }
   }
 
+  private @Nullable Date findLatestChildTimestamp() {
+    final List<Span> spans = new ArrayList<>(this.children);
+    Collections.sort(spans, spanByTimestampComparator);
```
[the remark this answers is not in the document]
romtsn: not sure I got it, wdym by  "sorting before traversing"?

sentry/src/main/java/io/sentry/SentryTracer.java
romtsn: if you mean the `requireNonNull` checks above - yes, but this should never happen, because at this point all children are finished and must have the timestamp. Maybe I misunderstood how it works though, so lemme know if you see otherwise.

romtsn:
> > @marandaneto mm, but only methods marked as `Internal` have changed. Only if we care about ABI, then it makes sense.
> 
> Indeed, but I'd rather point to v6 to speed up this major release, by that I mean more people working on the new code base and testing out all the changes. Unless @bruno-garcia says otherwise.

sounds good, have no problem with targeting 6.x.x actually!

sentry-android-core/src/main/java/io/sentry/android/core/UserInteractionIntegration.java
romtsn: indeed

sentry/src/main/java/io/sentry/SentryTracer.java
marandaneto: That makes sense, but I'd still apply defensive programming here, rather bailing if one of them is null than throwing an exception.

sentry/src/main/java/io/sentry/SentryTracer.java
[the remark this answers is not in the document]
marandaneto: I meant, what's more expensive?
Sorting the list using a comparator as you did or just traversing the array and doing an `if` check it in a for loop?
I believe the comparator will be more expensive since it needs to move all the items around.

sentry/src/main/java/io/sentry/config/PropertiesProvider.java
marandaneto: If possible yes, log out if there's a logger instance available and return null.

sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/SentryGestureListener.java
```diff
@@ -178,12 +191,128 @@ private void addBreadcrumb(
...
+        options
+            .getLogger()
+            .log(
+                SentryLevel.DEBUG,
+                "The view with id: "
+                    + viewId
+                    + " already has an ongoing transaction assigned. Rescheduling finish");
+
+        final Long idleTimeout = options.getIdleTimeout();
+        if (idleTimeout != null) {
```
romtsn, line 235: Here perhaps we'll add a check to do this only for scrolls/swipes

sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/SentryGestureListener.java
```diff
@@ -178,12 +191,128 @@ private void addBreadcrumb(
...
+          // reschedule the finish task for the idle transaction, so it keeps running for the same
+          // view
+          activeTransaction.scheduleFinish(idleTimeout);
+        }
+        return;
+      } else {
+        // as we allow a single UI transaction running on the bound Scope, we finish the previous
+        // one, if it's a new view
+        activeTransaction.finish();
+        activeTransaction = null;
```
romtsn, line 245: Here maybe, we'd need to remove the transaction from scope, in case `activeTransaction` has an unfinished child, `finish` wouldn't do anything, hence the new transaction wouldn't be added to Scope, since there's still an ongoing one.

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -52,29 +81,88 @@ public SentryTracer(
...
+  @TestOnly
+  void scheduleFinish(final @NotNull Long idleTimeout, final @Nullable CountDownLatch latch) {
+    cancelTimer();
+    timerTask =
+        new TimerTask() {
+          @Override
+          public void run() {
+            final SpanStatus status = getStatus();
+            finish((status != null) ? status : SpanStatus.OK);
+            if (latch != null) {
```
romtsn, line 147: I'm not sure if it's good idea to have this in prod code, but idk how to test it otherwise?

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -139,6 +227,10 @@ private ISpan createChild(
 
     Objects.requireNonNull(parentSpanId, "parentSpanId is required");
     Objects.requireNonNull(operation, "operation is required");
+    if (idleTimeout != null) {
+      // reset the timer when a new child gets added to extend the idle transaction
```
romtsn, line 231: here, perhaps we'd need to change status to `FinishStatus.notFinished()`, in case a new child was added AFTER the idle timeout, but we are still waiting for another child to finish. This is needed if we decide to reset the timer, no matter what, even if the status is already Finishing

romtsn: We've decided to reset the timer when the child has finished, rather than when it was created, so this avoids the problem entirely

sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/SentryGestureListener.java
romtsn: We've decided to not distinguish between clicks and scrolls/swipes, so we'll just extend the transaction time for any kind of user interaction for the sake of simplicity

romtsn:
@marandaneto @philipphofmann @adinauer  this is ready for review PTAL

philipphofmann (commented):
Overall the PR looks great. Thanks for all the effort. Maybe I missed it, but I'm not sure if the code handles the following [scenario](https://www.notion.so/sentry/UI-Event-Transactions-26057935c07c43c9981f2f62ac48780c#58a563eb3c4c4b45a3adb5cc4dc17396): 
```
Scenario: Ongoing UI event transaction
   Given an ongoing UI event transaction
   When the SDK creates a new screen load transaction
   Then the SDK finishes the previous transactions
   And removes it from the scope
   And sets the status to canceled
   And waits for its children to finish
```

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ Starting with version `6.6.0` of the `sentry`, [Sentry's version >= v21.9.0](htt
 * Fix: Profiling rate decreased from 300hz to 100hz; fixed profiling traces folder creation on manual sdk init (#1997)
 * Fix: Allow disabling sending of client reports via Android Manifest and external options (#2007)
 * Ref: Upgrade Spring Boot dependency to 2.5.13 (#2011)
+* Feat: Automatically create transactions for UI events (#1975)
```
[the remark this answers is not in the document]
romtsn: I think I'd rather keep it like it is now, I think it's a bit more informative when skimming through the changelog.

sentry/src/test/java/io/sentry/SentryTracerTest.kt
```diff
@@ -562,4 +566,102 @@ class SentryTracerTest {
...
+    }
+
+    @Test
+    fun `when a child is finished and the transaction is idle, resets the timer`() {
+        val transaction = fixture.getSut(waitForChildren = true, idleTimeout = 3000)
+
+        val initialTime = transaction.timerTask!!.scheduledExecutionTime()
+
+        val span = transaction.startChild("op")
+        Thread.sleep(10)
```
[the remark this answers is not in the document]
romtsn: mm, we need a sleep here for sure, otherwise the code might be too quick to execute, so there's no actual delay in between 2 timer schedules. but I believe we can make just 1ms and it still should work + the impact is gonna be almost unnoticeable

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
```diff
@@ -87,6 +87,9 @@ public final class SentryAndroidOptions extends SentryOptions {
   /** Interval for profiling traces in milliseconds. Defaults to 100 times per second */
   private int profilingTracesIntervalMillis = 1_000 / 100;
 
+  /** Enables the Auto instrumentation for user interaction tracing. */
+  private boolean enableUserInteractionTracing = false;
```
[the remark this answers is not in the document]
brustolin: Yes

CHANGELOG.md
marandaneto, line 10: I guess you need to merge 6.x.x into your branch

sentry/src/main/java/io/sentry/SentryTracer.java
marandaneto: You can implement the `latch` logic in the testing.

sentry/src/test/java/io/sentry/SentryTracerTest.kt
```diff
@@ -562,4 +566,102 @@ class SentryTracerTest {
             anyOrNull()
         )
     }
+
+    @Test
+    fun `when initialized without idleTimeout, does not schedule finish timer`() {
```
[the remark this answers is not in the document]
romtsn: But we said we cancel the timer when a new child is added and reset it when the last child is finished, so there's technically no idle timeout while there's an ongoing child span.

Even though, this has been missing in my code, so I'll update that, thanks for the point.

sentry/src/main/java/io/sentry/SentryTracer.java
romtsn: Do you have an example? I'm not sure I got what you mean

romtsn:
> Overall the PR looks great. Thanks for all the effort. Maybe I missed it, but I'm not sure if the code handles the following [scenario](https://www.notion.so/sentry/UI-Event-Transactions-26057935c07c43c9981f2f62ac48780c#58a563eb3c4c4b45a3adb5cc4dc17396):
> 
> ```
> Scenario: Ongoing UI event transaction
>    Given an ongoing UI event transaction
>    When the SDK creates a new screen load transaction
>    Then the SDK finishes the previous transactions
>    And removes it from the scope
>    And sets the status to canceled
>    And waits for its children to finish
> ```

Yeah, this is done here -> https://github.com/getsentry/sentry-java/blob/f67519e541ebc61d6ddacc1027ac1b4b934f80bd/sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/SentryGestureListener.java#L265-L276

With the only difference that we do this when the current screen goes inactive (onPause). This makes sense for us, because we use android's WindowCallback, which will get detached as soon as the activity/screen goes inactive, therefore it wouldn't make sense for us to wait until the new screen transaction, but rather finish it here.

sentry/src/main/java/io/sentry/SentryTracer.java
romtsn: yes, exactly, it's only there for testing purpose.

marandaneto: So then you can have your own implementation of `CountDownLatch` on tests, e.g. `AppLifecycleIntegrationTest` has that, or just search for `CountDownLatch` and `org.awaitility.kotlin.await`, both should give you ideas.

romtsn: dang, I was absolutely unaware about such thing as awaitility, thanks 👍  Refactored it to be like LifecycleWatcherTest

sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/SentryGestureListener.java
```diff
@@ -178,12 +192,131 @@ private void addBreadcrumb(
...
+        // one, if it's a new view
+        stopTracing(SpanStatus.OK);
+      }
+    }
+
+    // we can only bind to the scope if there's no running transaction
+    final String name = getActivityName(activity) + "." + viewId;
+    final String op = UI_ACTION + "." + eventType;
+    final ITransaction transaction =
+        hub.startTransaction(name, op, true, options.getIdleTimeout(), true);
```
marandaneto, line 253: Since we have the instance of the `activity` here, we should consider adding support for slow and frozen frames as well.
https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java#L164-L175=
See the usage of `activityFramesTracker`.

romtsn: We haven't planned for that, but I'm fine doing it since it's a quick one, @philipphofmann @brustolin is it also feasible for you?

marandaneto: Your call, file an issue otherwise, so we don't forget.

sentry/src/test/java/io/sentry/SentryTracerTest.kt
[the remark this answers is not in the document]
philipphofmann: Yes, makes sense. Thanks for the clarification.

sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/SentryGestureListener.java
philipphofmann: We do this anyways for every transaction https://github.com/getsentry/sentry-cocoa/blob/e5a906bf07e0793d1d538ac33e2869efd0b668be/Sources/Sentry/SentryTracer.m#L99-L105

romtsn: ah I didn't know this, that's neat. I will do it as part of this PR then, np

sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesTracker.java
```diff
@@ -55,9 +55,13 @@ public synchronized void addActivity(final @NotNull Activity activity) {
...
   }
 
+  public synchronized void setMetrics(final @NotNull SentryId sentryId) {
+    setMetrics(null, sentryId);
+  }
+
   @SuppressWarnings("NullAway")
   public synchronized void setMetrics(
-      final @NotNull Activity activity, final @NotNull SentryId sentryId) {
+      final @Nullable Activity activity, final @NotNull SentryId sentryId) {
```
marandaneto, line 64: Why is `Activity` nullable? The frames tracker needs the activity.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesTracker.java
```diff
@@ -68,7 +72,11 @@ public synchronized void setMetrics(
 
     SparseIntArray[] framesRates = null;
     try {
-      framesRates = frameMetricsAggregator.remove(activity);
+      if (activity != null) {
+        framesRates = frameMetricsAggregator.remove(activity);
+      } else {
+        framesRates = frameMetricsAggregator.getMetrics();
```
marandaneto, line 78: That will return values for all activities, but our transaction is bound to activity x, also, this won't remove/reset the values for that specific activity.

sentry-android-core/src/main/java/io/sentry/android/core/UserInteractionIntegration.java
```diff
@@ -82,16 +89,23 @@ private void stopTracking(final @Nullable Window window) {
...
   @Override
   public void onActivityResumed(@NotNull Activity activity) {
-    startTracking(activity.getWindow(), activity);
+    startTracking(activity);
   }
 
   @Override
   public void onActivityPaused(@NotNull Activity activity) {
-    stopTracking(activity.getWindow());
+    stopTracking(activity);
```
marandaneto, line 108: Shouldn't it be when the transaction is finished? `ActivityLifecycleIntegration` only does on destroyed, iirc.

sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/SentryGestureListener.java
```diff
@@ -18,25 +24,35 @@
...
-  private final @NotNull WeakReference<Window> windowRef;
+  public static final String UI_ACTION = "ui.action";
+
+  private final @NotNull WeakReference<Activity> activityRef;
   private final @NotNull IHub hub;
   private final @NotNull SentryAndroidOptions options;
   private final boolean isAndroidXAvailable;
+  private final @NotNull ActivityFramesTracker activityFramesTracker;
+
+  private @NotNull WeakReference<View> activeView = new WeakReference<>(null);
```
marandaneto, line 40: Shouldn't be initing once there's a view? initing with `null` is suboptimal maybe?

sentry-android-core/src/main/java/io/sentry/android/core/UserInteractionIntegration.java
romtsn: do you mean when the activity is finished? Nope, because the WindowCallback/GestureListener only makes sense when the activity is interact-able, otherwise we remove it in onPause and re-attach onResume, so it makes sense to finish the UI transaction here as well, as the activity is not interact-able anyway?

marandaneto: Yeah makes sense, my bad.

sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/SentryGestureListener.java
romtsn: mm, I just wanted to avoid null check for the weakref itself, but yeah would be better to init it when there's a view, true 👍

romtsn:
Revert frame metrics as we'll do it as part of https://github.com/getsentry/team-mobile/issues/2

PR #1991, opened 2022-04-18
feat: ndk scope sync on by default

bruno-garcia:
Turn scope sync on by default


bruno-garcia:
I missed this:

>    * Enable the Java to NDK Scope sync. The default value for sentry-java is disabled and enabled
   * for sentry-android.

PR #1997, opened 2022-04-20
Feat/profiling fixes

stefanosiano:
## :scroll: Description

reduced profiling sampling rate from 300hz to 100hz
profiling trace folder is now created regardless of profiling enabled option

Manual initialisation happens after profiling traces folder creation, so we were losing profiling when it was set manually, as the traces folder was not being created

Running a sample with strict mode did not complain about folder creation

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
@@ -14,7 +14,6 @@ import io.sentry.SendFireAndForgetOutboxSender;
 import io.sentry.SentryLevel;
 import io.sentry.android.fragment.FragmentLifecycleIntegration;
 import io.sentry.android.timber.SentryTimberIntegration;
-import io.sentry.util.FileUtils;
 import io.sentry.util.Objects;
 import java.io.BufferedInputStream;
 import java.io.File;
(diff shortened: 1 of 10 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -305,23 +305,23 @@ private static void initializeCacheDirs(
...
-                if (oldTracesDirContent == null) return;
-                // Method trace files are normally deleted at the end of traces, but if that fails
-                // for some
-                // reason we try to clear any old files here.
-                for (File f : oldTracesDirContent) {
-                  FileUtils.deleteRecursively(f);
-                }
-              });
-    }
+    // We don't have the options set by code (manual initialisation) at this point, so we have no
```
marandaneto, line 308: That's why we do the file handling in the Sentry init class
https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/Sentry.java#L222-L229=
Can we move this logic in there? so we don't need create a folder unnecessarily

stefanosiano: alright, but i should move `profilingTracesDirPath` from `SentryAndroidOptions` to `SentryOptions` (`profilingEnabled` is already in `SentryOptions`).
Is it fine?

sentry-android-core/src/test/java/io/sentry/android/core/AndroidTransactionProfilerTest.kt
```diff
@@ -46,6 +49,7 @@ class AndroidTransactionProfilerTest {
         context = ApplicationProvider.getApplicationContext()
         file = context.cacheDir
         AndroidOptionsInitializer.init(fixture.options, createMockContext())
+        Sentry.init(fixture.options)
```
marandaneto, line 52: Not all the tests require `Sentry.init`, so I'd rather do in a specific test and also call `Sentry.close` after.

stefanosiano: yeah, it's needed to generate the profiling trace folder. Was missing the `Sentry.close` part 😅

marandaneto:
@romtsn would you like to do a final review? :)

sentry/src/main/java/io/sentry/Sentry.java
```diff
@@ -231,6 +233,28 @@ private static boolean initConfigurations(final @NotNull SentryOptions options)
       options.setEnvelopeDiskCache(EnvelopeCache.create(options));
     }
 
+    String profilingTracesDirPath = options.getProfilingTracesDirPath();
```
romtsn, line 236:
```suggestion
    final String profilingTracesDirPath = options.getProfilingTracesDirPath();
```

romtsn (approved):

Issue #1647, opened 2021-08-13
Add Radiography integration


romtsn:
Add optional [Radiography](https://github.com/square/radiography) integration to automatically capture view hierarchy in case of a crash. Can be sent as a text file via an [attachment](https://docs.sentry.io/platforms/android/enriching-events/attachments/).

The benefit is that it also supports Jetpack Compose.

marandaneto:
This is partially blocked by https://github.com/getsentry/sentry-cocoa/issues/1772
Ideally, we'd do a quick investigation for Cocoa support, so we know if it's possible and how the widget tree would look like,

markushi:
View Hierarchy is now collected without radiography.

PR #1998, opened 2022-04-20
PoC radiography integration

marandaneto:
_#skip-changelog_

That's just a PoC of radiography as an attachment.
Ideally Sentry UI would render that nicely.

It should be a separate package since it needs to add the `com.squareup.radiography:radiography` dependency.
We can check at runtime if it's available in the classpath and add the integration automatically.
The package should also be installed automatically with the SAGP if `radiography` is in the classpath.
Support for Jetpack compose depends on `androidx.compose.ui:ui-tooling` in the classpath, that can also be added automatic if compose is in the classpath.

Closes https://github.com/getsentry/sentry-java/issues/1647

### DIFF
--- a/buildSrc/src/main/java/Config.kt
+++ b/buildSrc/src/main/java/Config.kt
@@ -10,7 +10,7 @@ object Config {
     val kotlinCompatibleLanguageVersion = "1.4"
 
     object BuildPlugins {
-        val androidGradle = "com.android.tools.build:gradle:7.1.2"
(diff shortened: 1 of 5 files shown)


marandaneto:
radiography

That's what it'd look like.

marandaneto:
Ideally, the file format would match with the radiography on iOS, so Sentry UI could prettify and offer features like collapsing, etc...

romtsn:
Closing in favour of #2440

PR #2009, opened 2022-04-26
AndroidTransactionProfiler is now initialized the first time a transa…

stefanosiano:
## :scroll: Description
AndroidTransactionProfiler is now initialized the first time a transaction is started, since in the constructor we don't have the options, yet

When `AndroidTransactionProfiler` is created and set in the options, in the `AndroidOptionsInitializer`, we don't know what options were set through `SentryAndroid.init()` yet, as the profiler is created right before it.
We can set the internal variables of the profiler that depends on the options on the first profile, as we would be sure to already have all options set up.

I'm building some ui tests (that will be released in another pr)

## :pencil: Checklist

- [x] No breaking changes


marandaneto:
@stefanosiano would you mind using the PR template to explain why this is needed?

marandaneto:
@stefanosiano We always add a test for a regression, would you mind adding it? thanks.

marandaneto (approved):

PR #2013, opened 2022-04-28
Tests/android UI tests

stefanosiano:
## :scroll: Description
Add some ui tests using Google's Espresso for Android.
There are currently two modules, grouped under `sentry-android-ui-tests`:
* benchmark
This module checks the difference between two operations and check they don't pass a specified threshold.
It's better to keep it separate from the other tests, as it requires a specific test runner (`androidx.benchmark.junit4.AndroidBenchmarkRunner`) and the application shouldn't be debuggable.
* end2end
It contains all end2end tests that we are interested in. Right now it contains a mock relay server that allows us to check the envelope that was sent to relay.

We need a benchmark to be sure that profiling overhead is under a certain threshold.
We also needed to start using some ui tests, as they can check things that Unit tests cannot do.

🤣 

## :pencil: Checklist

- [ ] No breaking changes

## :crystal_ball: Next steps
* Comments and javadoc on end2end module
* Clean the end2end tests 
* Maybe remove some details from the Benchmark tests as they are already in the end2end (appContext, testRunner)?
* Apply detekt on test modules?


marandaneto:
@stefanosiano drop us a line once this is ready to be reviewed.

What's about addressing the 2 items from the next steps in this PR:

- Maybe remove some details from the Benchmark tests as they are already in the end2end (appContext, testRunner)?
- Apply detekt on test modules?

 We try to keep tests as clean as possible as well, so we run detekt and friends to all of our tests as well for every other module.

marandaneto:
@stefanosiano drop us a line once this is ready to be reviewed.

What's about addressing the 2 items from the next steps in this PR:

- Maybe remove some details from the Benchmark tests as they are already in the end2end (appContext, testRunner)?
- Apply detekt on test modules?

 We try to keep tests as clean as possible as well, so we run detekt and friends to all of our tests as well for every other module.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Tests: added Android UI tests (#2013)
```
marandaneto, line 5: We don't need a changelog entry, this is only for user-facing changes.

marandaneto, line 5: We don't need a changelog entry, this is only for user-facing changes.

stefanosiano:
@marandaneto
Now the pr is ready to review. 
I just need to add saucelabs, but it has nothing to do with actual code, so it's fine to review now.

marandaneto:
> @marandaneto Now the pr is ready to review. I just need to add saucelabs, but it has nothing to do with actual code, so it's fine to review now.

Cool, I'll let this one to @romtsn , Thanks.

romtsn:
@stefanosiano I'm happy to review it, but could you give more context on this please? Was it aligned with someone, did you have any discussions on how this will look like? (e.g. is this just a PoC for profiling or are we going to reuse it later for other benchmarks like app startup time and so on). 

It's a bit hard to review a big PR without much context.

Also, regarding the e2e tests, I believe we actually wanted to run a real/test relay in a docker container alongside the test app and have assertions on their site, so I'm not sure if the current approach with the MockWebServer would actually be valuable, since we just essentially test serialization logic (which is already tested anyway). cc @bruno-garcia

stefanosiano:
> @stefanosiano I'm happy to review it, but could you give more context on this please? Was it aligned with someone, did you have any discussions on how this will look like? (e.g. is this just a PoC for profiling or are we going to reuse it later for other benchmarks like app startup time and so on).

This is mostly how it was done in Specto. We are going to have a call tomorrow morning (european timezone) to discuss on how to use the values we get from these tests. We can add you to it if you want.
The idea is to add generic benchmark tests to the sentry-uitest-android-benchmark module.
And add generic end2end tests to the sentry-uitest-android module.

> It's a bit hard to review a big PR without much context.

I can understand, so we can have a call and go through it together, if it can help

> Also, regarding the e2e tests, I believe we actually wanted to run a real/test relay in a docker container alongside the test app and have assertions on their site, so I'm not sure if the current approach with the MockWebServer would actually be valuable, since we just essentially test serialization logic (which is already tested anyway). cc @bruno-garcia

These tests can also be used to check caching logic, mock errors on the server to check the behaviour, etc.
Also, changing the dsn we can change the relay server we send data to, so you can still use it to test a real relay server.

bruno-garcia:
> Also, regarding the e2e tests, I believe we actually wanted to run a real/test relay in a docker container alongside the test app and have assertions on their site, so I'm not sure if the current approach with the MockWebServer would actually be valuable, since we just essentially test serialization logic (which is already tested anyway). cc @bruno-garcia

@romtsn  That might come eventually but we've waited for a few years so best we figure something out to add some level of confidence here. [Unity repo for example](https://github.com/getsentry/sentry-unity/actions/runs/2302145995) built a matrix to test things in real devices and it asserts the payload on mock servers. Not doing benchmarking or validating things work with Relay but it has caught a lot of problems already at the PR level:

image

> These tests can also be used to check caching logic, mock errors on the server to check the behaviour, etc.
Also, changing the dsn we can change the relay server we send data to, so you can still use it to test a real relay server.

@stefanosiano  Yeah that makes sense.

sentry-uitest/sentry-uitest-android-benchmark/src/androidTest/java/io/sentry/uitest/android/benchmark/SentryBenchmarkTest.kt
```diff
@@ -0,0 +1,135 @@
...
+
+        assertTrue { comparisonResult.durationIncrease >= -1 }
+        assertTrue { comparisonResult.durationIncrease < 1 }
+        assertTrue { comparisonResult.cpuTimeIncrease >= -1 }
+        assertTrue { comparisonResult.cpuTimeIncrease < 1 }
+        // The fps decrease is skipped for the moment, due to approximation:
+        // if an operation runs at 59.49 fps and the other at 59.51, they are considered 59 and 60 fps respectively.
+        // Their difference would be 1 / 60 * 100 = 1.66666%
+        // On slow devices, a difference of 1 fps on an average of 20 fps would account for 5% decrease.
+        // On even slower devices the difference would be even higher. Let's skip for now, as it's not important anyway.
```
romtsn, line 62: Didn't we say that the fps is the most important metric? 😜

sentry-uitest/sentry-uitest-android-benchmark/src/androidTest/java/io/sentry/uitest/android/benchmark/SentryBenchmarkTest.kt
```diff
@@ -0,0 +1,135 @@
...
+    fun benchmarkSameOperation() {
+
+        // We compare two operation that are the same. We expect the increases to be negligible, as the results
+        // should be very similar.
+        val op1 = BenchmarkOperation(runner, getOperation(runner) { null })
+        val op2 = BenchmarkOperation(runner, getOperation(runner) { null })
+        val comparisonResult = BenchmarkOperation.compare(op1, "Op1", op2, "Op2")
+
+        assertTrue { comparisonResult.durationIncrease >= -1 }
+        assertTrue { comparisonResult.durationIncrease < 1 }
```
romtsn, line 55:
```suggestion
        assertTrue { comparisonResult.durationIncrease in -1..1 }
```
I think it's more readable and also for other assertions

sentry-uitest/sentry-uitest-android-benchmark/src/androidTest/java/io/sentry/uitest/android/benchmark/SentryBenchmarkTest.kt
```diff
@@ -0,0 +1,135 @@
...
+    private fun getOperation(runner: AndroidJUnitRunner, transactionBuilder: () -> ITransaction?): () -> Unit = {
+        var transaction: ITransaction? = null
+        // Launch the sentry-uitest-android-benchmark activity
+        val benchmarkScenario = launchActivity<BenchmarkActivity>()
+        // Starts a transaction (it can be null, but we still runOnMainSync to make operations as similar as possible)
+        runner.runOnMainSync {
+            transaction = transactionBuilder()
+        }
+        // Just swipe the list some times: this is the benchmarked operation
+        repeat(2) {
```
romtsn, line 121: maybe extract this into a private method (or can be even a local function), like:
```kotlin
        val swipe = fun(times: Int) {
            repeat(times) {
		onView
     		onIdle
	    }
        }
```

sentry-uitest/sentry-uitest-android-benchmark/src/androidTest/java/io/sentry/uitest/android/benchmark/SentryBenchmarkTest.kt
```diff
@@ -0,0 +1,135 @@
...
+        assertTrue { comparisonResult.fpsDecrease < 5.0 }
+        assertTrue { comparisonResult.droppedFramesIncrease >= 0 }
+        assertTrue { comparisonResult.droppedFramesIncrease < 5.0 }
+    }
+
+    /**
+     * Operation that will be compared: it launches [BenchmarkActivity], swipe the list and closes it.
+     * The [transactionBuilder] is used to create the transaction before the swipes.
+     */
+    private fun getOperation(runner: AndroidJUnitRunner, transactionBuilder: () -> ITransaction?): () -> Unit = {
```
romtsn, line 112:
```suggestion
    private fun getOperation(runner: AndroidJUnitRunner, transactionBuilder: () -> ITransaction? = { null }): () -> Unit = {
```
so we don't need to specify it everytime

sentry-uitest/sentry-uitest-android-benchmark/src/androidTest/java/io/sentry/uitest/android/benchmark/SentryBenchmarkTest.kt
```diff
@@ -0,0 +1,135 @@
...
+    fun cleanup() {
+        IdlingRegistry.getInstance().unregister(BenchmarkActivity.scrollingIdlingResource)
+    }
+
+    @Test
+    fun benchmarkSameOperation() {
+
+        // We compare two operation that are the same. We expect the increases to be negligible, as the results
+        // should be very similar.
+        val op1 = BenchmarkOperation(runner, getOperation(runner) { null })
```
romtsn, line 50: can we just pass a `Choreographer` instance instead of a runner, since it's the same choreographer instance for all operations anyway, right?

sentry-uitest/sentry-uitest-android/src/main/java/io/sentry/uitest/android/utils/BooleanIdlingResource.kt
```diff
@@ -0,0 +1,30 @@
+package io.sentry.uitest.android.utils
+
+import androidx.test.espresso.IdlingResource
+import java.util.concurrent.atomic.AtomicBoolean
+
+/** Idling resource based on a boolean flag. */
+class BooleanIdlingResource(private val name: String) : IdlingResource {
```
romtsn, line 7: I think this one is unused in this module?

romtsn (commented):
Thanks for the PR, looks nice! I'm almost through, but I will come back later to finish the review for e2e tests.

sentry-uitest/sentry-uitest-android-benchmark/src/androidTest/java/io/sentry/uitest/android/benchmark/SentryBenchmarkTest.kt
stefanosiano: The most important is the dropped frames, not the fps.
The fps can be the same (as they don't count "partial frames") while the dropped frames consider the time a frame took over the maximum threshold for a fluid application (60 fpr = around 17 milliseconds per frame). 
So, rather than just saying "we dropped 1 frame", we say e.g. "we dropped 0.3 frames", in case a frame took 5 milliseconds more than allowed.

sentry-uitest/sentry-uitest-android/src/main/java/io/sentry/uitest/android/utils/BooleanIdlingResource.kt
stefanosiano: Not at the moment, but since new tests will be added it's something that could make sense to keep.
Can remove it, then just copy again from the benchmark module in case it's needed. Or just keep it here 🤔

romtsn: Maybe put it into `sentry-test-support` so it can be shared across modules?

sentry-uitest/sentry-uitest-android-benchmark/build.gradle.kts
```diff
@@ -0,0 +1,131 @@
...
+            keyPassword = "android"
+        }
+    }
+
+    testBuildType = System.getProperty("testBuildType", "debug")
+
+    buildTypes {
+        getByName("debug") {
+            isDebuggable = false
+            isMinifyEnabled = false
```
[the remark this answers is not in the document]
stefanosiano: Yeah, i was having some issues with it, but solved it now (had to avoid shrinking as it strips away things from the test apk even with `keep` rules)

sentry-uitest/sentry-uitest-android/src/main/java/io/sentry/uitest/android/utils/BooleanIdlingResource.kt
stefanosiano: BooleanIdlingResource extends Espresso's `IdlingResource`
Is it fine to add an Android specific class to `sentry-test-support`, as it's used by plain java sdk, too?
If yes, i'll add it there

romtsn: mmm, I'd say it's fine, since it's a test dependency anyway.. but maybe @marandaneto has a different opinion - we could create a new module `sentry-android-test-support` in the worst case.

sentry-uitest/sentry-uitest-android-benchmark/src/androidTest/java/io/sentry/uitest/android/benchmark/SentryBenchmarkTest.kt
```diff
@@ -0,0 +1,135 @@
...
+import io.sentry.android.core.SentryAndroid
+import io.sentry.uitest.android.benchmark.util.BenchmarkOperation
+import org.junit.Test
+import org.junit.runner.RunWith
+import kotlin.test.AfterTest
+import kotlin.test.BeforeTest
+import kotlin.test.assertTrue
+
+@RunWith(AndroidJUnit4::class)
+class SentryBenchmarkTest {
```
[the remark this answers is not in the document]
stefanosiano: Had a look at it, but:
1) `FrameTimingMetric` and its `frameOverrunMs` require api 31 (or 29 depending on which docs page you open...), and we want to measure against several Android versions
2) Macrobenchmark allows to print the benchmark data to console or to a json, but it's not returned in the code. Comparing two operations and make assertions on them would require a custom gradle plugin, which cannot be used in Saucelabs (not sure if you can use it in firebase), or reading the json results and compare them
Sadly it doesn't have a built-in method to compare two or more operations.
Maybe @indragiek could say something more on it?

sentry-uitest/sentry-uitest-android-benchmark/src/androidTest/java/io/sentry/uitest/android/benchmark/SentryBenchmarkTest.kt
```diff
@@ -0,0 +1,137 @@
...
+            }
+        )
+        val comparisonResult = BenchmarkOperation.compare(
+            benchmarkOperationNoTransaction,
+            "NoTransaction",
+            benchmarkOperationProfiled,
+            "ProfiledTransaction"
+        )
+
+        assertTrue { comparisonResult.durationIncrease in -1F..5F }
```
romtsn, line 100: I think it should be in the range of 0..5, right?

stefanosiano: yeah, but on high end devices with several threads available, a single thread can manage profiling by itself, and while the cpu usage increases, the other metrics could remain the same.
I was thinking about just setting upper thresholds for that.
Of course, on real devices of SauceLabs, since they are slower, we expect all the metrics to increase, even if not much

stefanosiano: Will put back the range 0..5 👍

sentry-uitest/sentry-uitest-android-benchmark/src/androidTest/java/io/sentry/uitest/android/benchmark/SentryBenchmarkTest.kt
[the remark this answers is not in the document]
romtsn: ok, sounds good. I guess we'll still use macrobenchmark for app startup time, but your approach is useful for when we need to benchmark different options of the sdk.

sentry-uitest/sentry-uitest-android-benchmark/build.gradle.kts
```diff
@@ -0,0 +1,132 @@
...
+    kotlin("android")
+    id(Config.QualityPlugins.errorProne)
+    id(Config.QualityPlugins.gradleVersions)
+    id(Config.QualityPlugins.detektPlugin)
+}
+
+configure<JavaPluginExtension> {
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
+}
```
marandaneto, line 16: Do we need that? this is the default version anyway.

sentry-uitest/sentry-uitest-android-benchmark/build.gradle.kts
```diff
@@ -0,0 +1,132 @@
...
+    compileSdk = Config.Android.compileSdkVersion
+
+    defaultConfig {
+        applicationId = "io.sentry.uitest.android.benchmark"
+        minSdk = Config.Android.minSdkVersionNdk
+        targetSdk = Config.Android.targetSdkVersion
+        versionCode = 1
+        versionName = "1.0.0"
+
+        testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
```
marandaneto, line 28: Let's extract this to a const in the Config class, there are more usages across packages

sentry-uitest/sentry-uitest-android-benchmark/build.gradle.kts
```diff
@@ -0,0 +1,132 @@
...
+            keyPassword = "android"
+        }
+    }
+
+    testBuildType = System.getProperty("testBuildType", "debug")
+
+    buildTypes {
+        getByName("debug") {
+            isDebuggable = false
+            isMinifyEnabled = true
```
marandaneto, line 55:
```suggestion
```

sentry-uitest/sentry-uitest-android-benchmark/build.gradle.kts
```diff
@@ -0,0 +1,132 @@
...
+            isMinifyEnabled = true
+            signingConfig = signingConfigs.getByName("debug")
+            proguardFiles(getDefaultProguardFile("proguard-android-optimize.txt"), "benchmark-proguard-rules.pro")
+        }
+        getByName("release") {
+            isDebuggable = false
```
marandaneto, line 61:
```suggestion
```

This is the default value I believe already

sentry-uitest/sentry-uitest-android-benchmark/build.gradle.kts
```diff
@@ -0,0 +1,132 @@
...
+        checkDependencies = true
+
+        // We run a full lint analysis as build part in CI, so skip vital checks for assemble tasks.
+        checkReleaseBuilds = false
+    }
+
+    variantFilter {
+        if (Config.Android.shouldSkipDebugVariant(buildType.name)) {
+            ignore = true
+        }
```
marandaneto, line 84: Do we need the `test` variant for your tests on CI?

sentry-uitest/sentry-uitest-android-benchmark/src/androidTest/java/io/sentry/uitest/android/benchmark/SentryBenchmarkTest.kt
```diff
@@ -0,0 +1,137 @@
...
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.test.platform.app.InstrumentationRegistry
+import androidx.test.runner.AndroidJUnitRunner
+import io.sentry.ITransaction
+import io.sentry.Sentry
+import io.sentry.SentryOptions
+import io.sentry.android.core.SentryAndroid
+import io.sentry.uitest.android.benchmark.util.BenchmarkOperation
+import org.junit.Test
+import org.junit.runner.RunWith
```
marandaneto, line 22: Lets stick to `kotlin.test` package.

sentry-uitest/sentry-uitest-android-benchmark/src/androidTest/java/io/sentry/uitest/android/benchmark/SentryBenchmarkTest.kt
```diff
@@ -0,0 +1,137 @@
...
+        }
+    }
+
+    @AfterTest
+    fun cleanup() {
+        IdlingRegistry.getInstance().unregister(BenchmarkActivity.scrollingIdlingResource)
+    }
+
+    @Test
+    fun benchmarkSameOperation() {
```
marandaneto, line 52: Let's name the methods using the `like this`, so we keep the pattern with the other tests

sentry-uitest/sentry-uitest-android-benchmark/src/androidTest/java/io/sentry/uitest/android/benchmark/SentryBenchmarkTest.kt
```diff
@@ -0,0 +1,137 @@
...
+
+        // We compare two operation that are the same. We expect the increases to be negligible, as the results
+        // should be very similar.
+        val op1 = BenchmarkOperation(choreographer, getOperation(runner))
+        val op2 = BenchmarkOperation(choreographer, getOperation(runner))
+        val comparisonResult = BenchmarkOperation.compare(op1, "Op1", op2, "Op2")
+
+        assertTrue { comparisonResult.durationIncrease in -1F..1F }
+        assertTrue { comparisonResult.cpuTimeIncrease in -1F..1F }
```
marandaneto, line 61: we could use `assertTrue(result)` instead of a function, it's cheaper.

sentry-uitest/sentry-uitest-android-benchmark/src/androidTest/java/io/sentry/uitest/android/benchmark/SentryBenchmarkTest.kt
```diff
@@ -0,0 +1,137 @@
...
+        // The fps decrease is skipped for the moment, due to approximation:
+        // if an operation runs at 59.49 fps and the other at 59.51, they are considered 59 and 60 fps respectively.
+        // Their difference would be 1 / 60 * 100 = 1.66666%
+        // On slow devices, a difference of 1 fps on an average of 20 fps would account for 5% decrease.
+        // On even slower devices the difference would be even higher. Let's skip for now, as it's not important anyway.
+//        assertTrue { comparisonResult.fpsDecrease in -2F..2F }
```
marandaneto, line 67: Should we remove or uncomment?

sentry-uitest/sentry-uitest-android-benchmark/src/androidTest/java/io/sentry/uitest/android/benchmark/SentryBenchmarkTest.kt
```diff
@@ -0,0 +1,137 @@
...
+
+@RunWith(AndroidJUnit4::class)
+class SentryBenchmarkTest {
+
+    private lateinit var runner: AndroidJUnitRunner
+    private lateinit var context: Context
+    private lateinit var choreographer: Choreographer
+
+    @BeforeTest
+    fun setUp() {
```
marandaneto, line 35: Should likely call `Sentry.close` first since we use the `SentryAndroid.init` in tests?

sentry-uitest/sentry-uitest-android-benchmark/src/androidTest/java/io/sentry/uitest/android/benchmark/SentryBenchmarkTest.kt
```diff
@@ -0,0 +1,137 @@
...
+
+    @Test
+    fun benchmarkProfiledTransaction() {
+
+        runner.runOnMainSync {
+            SentryAndroid.init(context) { options: SentryOptions ->
+                options.dsn = "https://key@uri/1234567"
+                options.isEnableAutoSessionTracking = false
+                options.tracesSampleRate = 1.0
+                options.isTraceSampling = true
```
marandaneto, line 79: Do we need that? all you need here is `tracesSampleRate` only.

sentry-uitest/sentry-uitest-android-benchmark/src/androidTest/java/io/sentry/uitest/android/benchmark/util/BenchmarkOperation.kt
```diff
@@ -0,0 +1,129 @@
...
+
+// 60 FPS is the recommended target: https://www.youtube.com/watch?v=CaMTIgxCSqU
+private const val FRAME_DURATION_60FPS_NS: Double = 1_000_000_000 / 60.0
+
+/**
+ * Class that allows to sentry-uitest-android-benchmark some operations.
+ * Create two [BenchmarkOperation] objects and compare them using [BenchmarkOperation.compare] to get
+ * a [BenchmarkResult] with relative measured overheads.
+ */
+
```
marandaneto, line 16:
```suggestion
```

sentry-uitest/sentry-uitest-android-benchmark/src/androidTest/java/io/sentry/uitest/android/benchmark/util/BenchmarkOperation.kt
```diff
@@ -0,0 +1,129 @@
...
+        operationName
+    )
+
+    /**
+     * Helps ensure that operations don't impact one another.
+     * Doesn't appear to currently have an impact on the benchmark.
+     */
+    private fun isolate() {
+        Thread.sleep(500)
+        Runtime.getRuntime().gc()
```
marandaneto, line 111: Curious why we need to call `gc`.

sentry-uitest/sentry-uitest-android-benchmark/src/androidTest/java/io/sentry/uitest/android/benchmark/util/BenchmarkOperationResult.kt
```diff
@@ -0,0 +1,103 @@
...
+) {
+    /**
+     * Compare two [BenchmarkOperationResult], calculating increases of each parameter in percentage.
+     */
+    fun compare(other: BenchmarkOperationResult): BenchmarkResult {
+
+        // Measure average duration
+        val durationDiffNanos = avgDurationNanos - other.avgDurationNanos
+        val durationIncreasePercentage = durationDiffNanos * 100.0 / other.avgDurationNanos
+        println(
```
marandaneto, line 21: Why do we have all of this `println` if this is going to run on CI/device farm? do we output this log somewhere?

sentry-uitest/sentry-uitest-android-benchmark/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,25 @@
...
+      Important: disable debugging for accurate performance results
+
+      In a com.android.library project, this flag must be disabled from this
+      manifest, as it is not possible to override this flag from Gradle.
+    -->
+    <application
+      android:theme="@style/Theme.AppCompat"
+      android:label="Sentry Benchmark"
+      android:icon="@android:mipmap/sym_def_app_icon"
+      android:debuggable="false"
```
marandaneto, line 16: Isn't `android:debuggable` already set in the Gradle file anyway?

sentry-uitest/sentry-uitest-android-benchmark/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,25 @@
...
+      tools:ignore="HardcodedDebugMode">
+
+      <activity android:name=".BenchmarkActivity" />
+
+      <meta-data android:name="io.sentry.auto-init" android:value="false" />
```
marandaneto, line 21: Maybe remove the content providers instead? https://docs.sentry.io/platforms/android/configuration/manual-init/ it's more efficient.

sentry-uitest/sentry-uitest-android-benchmark/src/main/java/io/sentry/uitest/android/benchmark/BenchmarkActivity.kt
```diff
@@ -0,0 +1,67 @@
...
+     * This number can be increased to mimic busier applications.
+     */
+    private val backgroundThreadPoolSize = 1
+    private val executor: ExecutorService = Executors.newFixedThreadPool(backgroundThreadPoolSize)
+    private var resumed = false
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+
+        setContentView(R.layout.activity_benchmark)
```
marandaneto, line 31: We can use `buildFeatures.viewBinding`, which makes the code much simpler.

sentry-uitest/sentry-uitest-android/build.gradle.kts
```diff
@@ -0,0 +1,128 @@
+import io.gitlab.arturbosch.detekt.Detekt
```
marandaneto, line 1: See feedback from the other Gradle file.

sentry-uitest/sentry-uitest-android-benchmark/build.gradle.kts
stefanosiano: Maybe not, but I've seen it in all the modules, so I added it here, too, at least to be consistent with the others.
Should i remove it?

sentry-uitest/sentry-uitest-android/src/androidTest/java/io/sentry/uitest/android/EnvelopeTests.kt
```diff
@@ -0,0 +1,61 @@
...
+            assertNoOtherEnvelopes()
+            assertNoOtherRequests()
+        }
+    }
+
+    @Test
+    fun checkEnvelopeProfiledTransaction() {
+
+        initSentry(true) { options: SentryOptions ->
+            options.isEnableAutoSessionTracking = false
```
marandaneto, line 39: See feedback from the other init block

sentry-uitest/sentry-uitest-android/src/androidTest/java/io/sentry/uitest/android/mockservers/EnvelopeAsserter.kt
```diff
@@ -0,0 +1,35 @@
...
+ */
+class EnvelopeAsserter(val envelope: SentryEnvelope, val response: MockResponse) {
+    /** List of items to assert. */
+    private val unassertedItems = envelope.items.toMutableList()
+
+    /**
+     * Asserts an envelope item that can be deserialized as an instance of [clazz] exists and returns the first one.
+     * The asserted item is then removed from internal list of unasserted items.
+     */
+    fun <T> assertItem(clazz: Class<T>): T {
```
marandaneto, line 19: We have the project `sentry-test-support` that does something similar, maybe move this in there, and avoid using `Sentry.getCurrentHub()`, you could simple create a new `JsonSerializer`

sentry-uitest/sentry-uitest-android/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,20 @@
...
+    android:theme="@style/Theme.AppCompat"
+    android:label="Sentry End2End Tests"
+    android:icon="@android:mipmap/sym_def_app_icon"
+    android:usesCleartextTraffic="true"
+    tools:targetApi="m">
+<!--  android:usesCleartextTraffic="true" is needed to send data to http://localhost... (the mock relay server)  -->
+
+    <activity android:name=".EmptyActivity" />
+
+    <meta-data android:name="io.sentry.auto-init" android:value="false" />
```
marandaneto, line 16: See the review from the other manifest file.

sentry-uitest/sentry-uitest-android/src/main/java/io/sentry/uitest/android/utils/BooleanIdlingResource.kt
marandaneto: Since it's testing only it's fine, the question is if the `java` tolling won't complain when trying to build it.
Otherwise, we can create a `sentry-android-test-support` as @romtsn said.

settings.gradle.kts
```diff
@@ -36,7 +36,9 @@ include(
     "sentry-samples:sentry-samples-spring",
     "sentry-samples:sentry-samples-spring-boot",
     "sentry-samples:sentry-samples-spring-boot-webflux",
-    "sentry-samples:sentry-samples-netflix-dgs"
+    "sentry-samples:sentry-samples-netflix-dgs",
```
marandaneto, line 39: I don't see this project in this PR?

sentry-uitest/sentry-uitest-android-benchmark/benchmark-proguard-rules.pro
```diff
@@ -0,0 +1,35 @@
+# Add project specific ProGuard rules here.
```
marandaneto, line 1: Maybe we rename the `sentry-uitest` folder to something more generic, since we have benchmarking and UI tests.
Maybe `sentry-android-integration-tests` or something else? since more benchmarking will come and not necessarily mean there's UI in there.

sentry-uitest/sentry-uitest-android-benchmark/src/androidTest/java/io/sentry/uitest/android/benchmark/SentryBenchmarkTest.kt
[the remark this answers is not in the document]
marandaneto: Since `Macrobenchmark` is the latest tooling and has a rich file as output, parsing the JSON file and reading it in the test wouldn't be problematic IMO.
Not telling you to change everything, but I'd go this way if I'd have to choose.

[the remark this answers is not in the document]
marandaneto: API level is a problem indeed, but isn't it about compiling the App and setting the right `compileSdk` and `targetSdk`? or do you mean that tests won't run on older devices?

sentry-uitest/sentry-uitest-android/src/androidTest/java/io/sentry/uitest/android/mockservers/EnvelopeAsserter.kt
romtsn:
```suggestion
    inline fun <reified T> assertItem(): T {
```
Also, since we are in Kotlin, can we make it inline+reified, so we can use it like `assertItem<SentryEvent>()` and use `T::class.java` instead of `clazz`

sentry-uitest/sentry-uitest-android/src/androidTest/java/io/sentry/uitest/android/BaseUiTest.kt
```diff
@@ -0,0 +1,82 @@
...
+        runner = InstrumentationRegistry.getInstrumentation() as AndroidJUnitRunner
+        context = ApplicationProvider.getApplicationContext()
+        context.cacheDir.deleteRecursively()
+        relay.start()
+        mockDsn = relay.createMockDsn()
+    }
+
+    @AfterTest
+    fun baseFinish() {
+        if (relayWaitForRequests) {
```
romtsn, line 49: do we really need this if-check and the boolean field? Wouldn't `unregister` do nothing if the resource was never registered?

settings.gradle.kts
romtsn: I guess it's a diff caused by the new comma

sentry-uitest/sentry-uitest-android-benchmark/build.gradle.kts
marandaneto: They are only on Java projects, not on the Android ones.

sentry-uitest/sentry-uitest-android-benchmark/build.gradle.kts
stefanosiano: We want development version to be the same as release, so that we can run a benchmark locally with correct results.
Proguard is used in release and can affect performances, so we want to enable it in debug, too

sentry-uitest/sentry-uitest-android-benchmark/build.gradle.kts
stefanosiano: We will need `release` variant only. 
Also, if `debug` variant is built on ci, everything breaks, as it expects its dependency modules (like sentry-android) to be `debug`, too, but they also skip the debug variant build on ci.

sentry-uitest/sentry-uitest-android-benchmark/build.gradle.kts
marandaneto: I know but its duplicated, you have 2 of them, isMinifyEnabled L55 and L56

sentry-uitest/sentry-uitest-android-benchmark/src/androidTest/java/io/sentry/uitest/android/benchmark/SentryBenchmarkTest.kt
stefanosiano: Yep, but directly in the test that inits the sdk (the other doesn't do it)

sentry-uitest/sentry-uitest-android-benchmark/build.gradle.kts
stefanosiano: Didn't realize it 😓

sentry-uitest/sentry-uitest-android-benchmark/src/androidTest/java/io/sentry/uitest/android/benchmark/SentryBenchmarkTest.kt
stefanosiano: do you mean naming the method e.g. ``` `benchmark same operation` ```?
It's not possible here due to `Identifier not allowed in Android projects`
Or do you mean something else?

sentry-uitest/sentry-uitest-android-benchmark/src/main/AndroidManifest.xml
stefanosiano: yes, initially the module was an android library, not an application, so i had to put debuggable flag in the manifest (gradle doesn't accept it for libraries).
I forgot to remove it

sentry-uitest/sentry-uitest-android-benchmark/src/androidTest/java/io/sentry/uitest/android/benchmark/SentryBenchmarkTest.kt
stefanosiano: I updated the comment and made it a 2 lines comment, just in case checking the test someone asks why we didn't check the fps decrease

stefanosiano: And removed the assertion

sentry-uitest/sentry-uitest-android-benchmark/build.gradle.kts
marandaneto: Then I don't see the need of configuring `getByName("debug")` in the first place.

sentry-uitest/sentry-uitest-android-benchmark/src/androidTest/java/io/sentry/uitest/android/benchmark/SentryBenchmarkTest.kt
marandaneto: Yes, the naming convention.
I just tested it and it worked on sentry-android-core.

sentry-uitest/sentry-uitest-android-benchmark/build.gradle.kts
stefanosiano: just because doing so, you don't have to manually select the `release` variant of the other modules if you want to run the benchmarks locally (android studio complains about `variants conflict`), and then reselect the `debug` variant if you want to work on other modules.
Or am I missing something?

sentry-uitest/sentry-uitest-android-benchmark/src/androidTest/java/io/sentry/uitest/android/benchmark/SentryBenchmarkTest.kt
stefanosiano: sentry-android-core is defined in gradle witht the plugin `com.android.library`.
When using the plugin `com.android.application` you cannot do that.
The `com.android.library` doesn't have the `isDebuggable` flag, and it seems it ignores what's in the Android manifest.

I found a workaround, though. I can change it to `com.android.library`, but i will have to add `(this as com.android.build.gradle.internal.dsl.BuildType).isDebuggable = false` in the `buildTypes` to make it not debuggable (just tested it and it's working fine).

stefanosiano: i tried setting the project as a `com.android.library`, as `sentry-android-core`, to check if it changes anything, but the method identifiers cannot contain spaces. 
It seems this behaviour involves classes in the `androidTest` folder

sentry-uitest/sentry-uitest-android-benchmark/src/androidTest/java/io/sentry/uitest/android/benchmark/util/BenchmarkOperation.kt
stefanosiano: We try to reduce the impact a benchmark run can have on the following one. We cannot clear the caches, but at least we can ask the system to destroy the objects of the previous run, and avoid they get reused.
Also, multiple runs of different operations could fill the ram, or in any case increase its usage, distorting the tests.
We try to run each test in the same conditions, as much as we can

sentry-uitest/sentry-uitest-android-benchmark/src/androidTest/java/io/sentry/uitest/android/benchmark/SentryBenchmarkTest.kt
marandaneto: It could be, thanks.
https://github.com/android/performance-samples/blob/main/MicrobenchmarkSample/microbenchmark/src/androidTest/java/com/example/benchmark/ViewInflateBenchmark.kt

sentry-uitest/sentry-uitest-android-benchmark/build.gradle.kts
marandaneto: If AS complaints about `variants conflict`, all good, I never experienced this error so I was not sure if it was needed.

sentry-uitest/sentry-uitest-android-benchmark/src/androidTest/java/io/sentry/uitest/android/benchmark/util/BenchmarkOperationResult.kt
stefanosiano: I removed few prints and made them comments. The idea is to start by printing the values, so that we can read them locally or from the logs on Saucelabs, and later on find a way to push these values somewhere else

sentry-uitest/sentry-uitest-android/src/main/java/io/sentry/uitest/android/utils/BooleanIdlingResource.kt
stefanosiano: I tried to create a new module `sentry-android-test-support`
There are several things, though. The most important thing is that the `BooleanIdlingResource` is needed as a normal dependency, not a test dependency. I think creating a whole module just for one class is an overkill.
So i make `sentry-uitest-android-benchmark` depending on `sentry-uitest-android`, so that only `sentry-uitest-android` contains BooleanIdlingResource
Is it fine? we could change it later on, if more things get added

romtsn:
@stefanosiano I've already approved, so nothing is blocking you from merging really ;)

stefanosiano:
@romtsn yeah, I know. Just wanted a last confirmation :)

PR #2016, opened 2022-05-03
Feat: Allow setting SDK info (name & version) in manifest

vaind:
## :scroll: Description
Enables users to override the SDK info

This allows other SDKs that use sentry-cocoa internally to identify the events. Enables https://github.com/getsentry/sentry-unity/issues/616

Added a test case

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+* Feat: Allow setting SDK info (name & version) in manifest (#2016)
 
 ## 6.0.0-beta.3
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,8 @@ Breaking changes:
     * Remove `IUnknownPropertiesConsumer`
 * Bump: Kotlin to 1.5 and compatibility to 1.4 for sentry-android-timber (#1815)
 
+* Feat: Allow setting SDK info (name & version) in manifest (#2016)
+
 ## 5.7.3
 
 * Fix: Sentry Timber integration throws an exception when using args (#1986)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,8 +84,6 @@ Breaking changes:
     * Remove `IUnknownPropertiesConsumer`
 * Bump: Kotlin to 1.5 and compatibility to 1.4 for sentry-android-timber (#1815)
 
-* Feat: Allow setting SDK info (name & version) in manifest (#2016)
-
 ## 5.7.3
 
 * Fix: Sentry Timber integration throws an exception when using args (#1986)
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
@@ -6,6 +6,7 @@ import android.content.pm.PackageManager;
 import android.os.Bundle;
 import io.sentry.ILogger;
 import io.sentry.SentryLevel;
+import io.sentry.protocol.SdkVersion;
 import io.sentry.util.Objects;
 import java.util.Arrays;
 import java.util.List;
@@ -31,6 +32,8 @@ final class ManifestMetadataReader {
   static final String NDK_SCOPE_SYNC_ENABLE = "io.sentry.ndk.scope-sync.enable";
   static final String RELEASE = "io.sentry.release";
   static final String ENVIRONMENT = "io.sentry.environment";
+  static final String SDK_NAME = "io.sentry.sdk.name";
+  static final String SDK_VERSION = "io.sentry.sdk.version";
 
   // TODO: remove on 6.x in favor of SESSION_AUTO_TRACKING_ENABLE
   static final String SESSION_TRACKING_ENABLE = "io.sentry.session-tracking.enable";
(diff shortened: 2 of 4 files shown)


bruno-garcia (approved):
LGTM. Lets see what @marandaneto and @romtsn think about this

sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
```diff
@@ -227,6 +230,22 @@ static void applyMetadata(
 
         options.setProguardUuid(
             readString(metadata, logger, PROGUARD_UUID, options.getProguardUuid()));
+
+        SdkVersion sdkInfo = options.getSdkVersion();
```
romtsn, line 234:
```suggestion
        final SdkVersion sdkInfo = options.getSdkVersion();
```

sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
```diff
@@ -227,6 +230,22 @@ static void applyMetadata(
...
+        if (sdkInfo != null) {
+          final String sdkName = readString(metadata, logger, SDK_NAME, sdkInfo.getName());
+          if (sdkName != null) {
```
romtsn, line 237: I think these checks are redundant, because `sdkInfo.getName()/sdkInfo.getVersion()` are non-nullable, hence the default value is always non-nullable

sentry-android-core/src/test/java/io/sentry/android/core/ManifestMetadataReaderTest.kt
```diff
@@ -548,6 +548,66 @@ class ManifestMetadataReaderTest {
...
+
+        // Act
+        ManifestMetadataReader.applyMetadata(context, fixture.options)
+
+        // Assert
+        assertEquals(expectedValue, fixture.options.sdkVersion?.name)
+    }
+
+    @Test
+    fun `applyMetadata does not override SDK name from options`() {
```
romtsn, line 567: the PR is failing because these tests are failing - I'm not sure they make sense, because the manifest metadata is actually overriding what was set in the Options ctor, right?

sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
vaind: the problem is with readString, returning a nullable type - should I instead create a non-nullable return type override?

romtsn: mm, you could do that, but it's a nitpick rather - it's also fine like that. So, as you prefer :)

vaind: added the override

sentry-android-core/src/test/java/io/sentry/android/core/ManifestMetadataReaderTest.kt
vaind: yeah, they don't really make sense, I was copying it from one other option but I see it's only used there (sample rate). Removed the test cases

sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
vaind: changed in another commit

romtsn (approved):

marandaneto:
@vaind please target the `6.x.x` branch, we are only fixing bugs on the current stable version -> `main`.

sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
```diff
@@ -227,6 +230,17 @@ static void applyMetadata(
...
+
+        final SdkVersion sdkInfo = options.getSdkVersion();
+        if (sdkInfo != null) {
+          sdkInfo.setName(readStringNotNull(metadata, logger, SDK_NAME, sdkInfo.getName()));
+          sdkInfo.setVersion(
+              readStringNotNull(metadata, logger, SDK_VERSION, sdkInfo.getVersion()));
+          options.setSdkVersion(sdkInfo);
+        } else {
+          // Should always be set by the Options constructor but let's error out if that changes.
+          throw new IllegalArgumentException("Property options.sdkVersion must not be null.");
```
marandaneto, line 242: This would crash the SDK and we don't want to do that, degrade gracefully.
I'd rather create a new `SdkVersion` instance and set it.

vaind:
> @vaind please target the `6.x.x` branch, we are only fixing bugs on the current stable version -> `main`.

done

sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
vaind: OK, changed

CHANGELOG.md
```diff
@@ -80,6 +80,8 @@ Breaking changes:
     * Remove `IUnknownPropertiesConsumer`
 * Bump: Kotlin to 1.5 and compatibility to 1.4 for sentry-android-timber (#1815)
 
+* Feat: Allow setting SDK info (name & version) in manifest (#2016)
```
marandaneto, line 83: Move it under `unreleased`.

marandaneto: This is duplicated now, changelog entry is in both places

vaind: sorry about the mess caused by the conflicts while rebasing. removed the latter entry.

Issue #1976, opened 2022-04-06
Fix Proguard/R8 rules


Thomas-Vos:
### Integration

sentry-android

### Build System

Gradle

### AGP Version

all

### Proguard

Enabled

### Version

6.0.0-alpha.4

### Steps to Reproduce

The Proguard/R8 rules keep far too many classes. See for example the following file: 

https://github.com/getsentry/sentry-java/blob/6.x.x/sentry-android-core/proguard-rules.pro

Lots of those rules seem to be needed for GSON, but GSON was removed in 6.x.x. So the rules can be reduced.

Specifically the following:

https://github.com/getsentry/sentry-java/blob/4414ba0f40bd97fa8a1e8879d8dd6b387b4af86b/sentry-android-core/proguard-rules.pro#L15-L18

This should either be changed to keep specific classes, or just removed completely as GSON is no longer used.

Same for the other rules, some of them seem like they were added for GSON but never removed.

### Expected Result

Remove a lot of the rules

### Actual Result

Too many classes are kept

marandaneto:
@Thomas-Vos indeed, it's in our backlog to find out what really needs to be excluded from minification or not, e.g., every public API class, interfaces, classes used in reflection, etc.

Thomas-Vos:
I am using Sentry in a JVM desktop app with R8 obfuscation. With the alpha version mentioned I did not need to add any additional rules for Sentry.

marandaneto:
> I am using Sentry in a JVM desktop app with R8 obfuscation. With the alpha version mentioned I did not need to add any additional rules for Sentry.

Yep, nothing needs to be added, it's about removing the not needed rules anymore.

Thomas-Vos:
Yes, what I meant was that I did not need to add any rules at all. The rules in the Android module were not included in my project as I only depend on the Java dependency.

philipphofmann:
Was fixed with #2031

PR #2031, opened 2022-05-10
Allow optimization and obfuscation of the SDK by reducing proguard rules

marandaneto:
Closes https://github.com/getsentry/sentry-java/issues/1976


sentry-android-ndk/proguard-rules.pro
```diff
@@ -1,6 +1,10 @@
 ##---------------Begin: proguard configuration for NDK  ----------
 
--keep class io.sentry.android.ndk.** { *; }
+# The Android SDK checks at runtime if this class is available via Class.forName
+-keep class io.sentry.android.ndk.SentryNdk { *; }
+# The JNI layer uses this classes through reflection
```
marandaneto, line 5: Good catch, do we need to keep classes with `native` methods such as `NativeScope`?

marandaneto:
@romtsn what's about `-renamesourcefileattribute SourceFile`? it's been removed but it was in the official docs.

romtsn:
> @romtsn what's about `-renamesourcefileattribute SourceFile`? it's been removed but it was in the official docs.

Sorry I've been meaning to leave some comments here but got lost in testing the changes 😄 

So I thought that we probably don't need `-renamesourcefileattribute` because our sdks are open-source, and there's no reason to hide the names of the files - I think this is something the consumers can decide whether they want to hide filenames from stacktraces or not and add this on themselves. This doesn't break de-obfuscation or anything, just there for security reasons (to make it harder to find out what files are causing problems)

sentry-android-ndk/proguard-rules.pro
romtsn: We already do this here below:
https://github.com/getsentry/sentry-java/blob/426d9be10fb3b28f16e73000834a686e1d755ae3/sentry-android-ndk/proguard-rules.pro#L10-L12

marandaneto: oh yeah indeed.

romtsn (approved):

PaulWoitaschek:
This crashes on release builds on AGP 7.2.1 due to the NdkIntegration using reflection.

marandaneto:
> This crashes on release builds on AGP 7.2.1 due to the NdkIntegration using reflection.

Thanks for the report, would you mind raising a new issue with more context and stack trace? Thanks.

romtsn:
Yeah we'd need more details - we explicitly keep `SentryNdk` class, so unless there's a rule which removes it, it should work fine (also, tried to reproduce - no luck)

PaulWoitaschek:
A regular bug report will come next week, just wanted to let you know this as this seems critical.

But it happens when calling Sentriy.init:

```kotlin
fun initSentry(application: Application) {
  val options = Sentry.OptionsConfiguration<SentryAndroidOptions> { options ->
    options.apply {
      isEnableNdk = false // <- this fixes the crash
      ...
    }
  }
  SentryAndroid.init(application, NoOpLogger.getInstance(), options)
}
```

I only have the backtrace:
```shell
  #00  pc 0000000000021d94  /system/lib64/libc.so (abort+124)
  #00  pc 0000000000470460  /system/lib64/libart.so (art::Runtime::Abort(char const*)+1320)
  #00  pc 0000000000008d2c  /system/lib64/libbase.so (android::base::LogMessage::~LogMessage()+724)
  #00  pc 00000000004a1cec  /system/lib64/libart.so (art::Thread::AssertNoPendingExceptionForNewException(char const*) const+1212)
  #00  pc 00000000004a5b04  /system/lib64/libart.so (art::Thread::ThrowNewExceptionF(char const*, char const*, ...)+180)
  #00  pc 000000000038ebe8  /system/lib64/libart.so (art::FindMethodID(art::ScopedObjectAccess&, _jclass*, char const*, char const*, bool)+640)
  #00  pc 0000000000339e0c  /system/lib64/libart.so (art::JNI::GetMethodID(_JNIEnv*, _jclass*, char const*, char const*)+652)
  #00  pc 0000000000002340  /data/app/com.yazio.android-cP3b_Tvq9c6BduqhlFfooQ==/split_config.arm64_v8a.apk!lib/arm64-v8a/libsentry-android.so (offset 0x3a3000) (Java_io_sentry_android_ndk_SentryNdk_initSentryNative+128)
  #00  pc 00000000000b78ac  /data/app/com.yazio.android-cP3b_Tvq9c6BduqhlFfooQ==/oat/arm64/base.odex (offset 0xb7000) (io.sentry.android.ndk.SentryNdk.initSentryNative [DEDUPED]+172)
  #00  pc 0000000000562a4c  /system/lib64/libart.so (art_quick_invoke_static_stub+604)
  #00  pc 00000000000d0160  /system/lib64/libart.so (art::ArtMethod::Invoke(art::Thread*, unsigned int*, unsigned int, art::JValue*, char const*)+232)
  #00  pc 00000000002865e8  /system/lib64/libart.so (art::interpreter::ArtInterpreterToCompiledCodeBridge(art::Thread*, art::ArtMethod*, art::ShadowFrame*, unsigned short, art::JValue*)+344)
  #00  pc 00000000002805a4  /system/lib64/libart.so (bool art::interpreter::DoCall<false, false>(art::ArtMethod*, art::Thread*, art::ShadowFrame&, art::Instruction const*, unsigned short, art::JValue*)+948)
  #00  pc 0000000000532edc  /system/lib64/libart.so (MterpInvokeStatic+204)
  #00  pc 0000000000554f14  /system/lib64/libart.so (ExecuteMterpImpl+14612)
  #00  pc 0000000001265ac6  /data/app/com.yazio.android-cP3b_Tvq9c6BduqhlFfooQ==/oat/arm64/base.vdex (io.sentry.android.ndk.SentryNdk.init+14)
```
I could swear that earlier this day I saw a proper stacktrace complaining that one of the SentryNdk functions was missing.

PR #2033, opened 2022-05-10
Bump/native 0.4.17

vaind:
## :scroll: Description

Update the native dependency to the latest version

## :green_heart: How did you test it?
In the Unity SDK CI: https://github.com/getsentry/sentry-unity/actions/runs/2305224087

## :pencil: Checklist

- [x] I reviewed the submitted code
- [ ] I added tests to verify the changes
- [ ] I updated the docs if needed
- [x] No breaking changes

PR #2035, opened 2022-05-10
Feat: Android native SDK name override

vaind:

Enables integrations, e.g. sentry-unity SDK, to override the native SDK name reported in errors

not sure how to test this except manually - are there some integration tests already that this could be included in?

## :pencil: Checklist

- [x] No breaking changes
- [ ] depends on #2033 to actually make a difference

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Features
 
 - Allow setting SDK info (name & version) in manifest ([#2016](https://github.com/getsentry/sentry-java/pull/2016))
+- Allow setting native Android SDK name during build ([#2035](https://github.com/getsentry/sentry-java/pull/2035))
 
 ## 6.0.0-beta.3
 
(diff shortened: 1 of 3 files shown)


romtsn:
Hm, sorry if I missed something, but aren't these lines doing exactly that already? 

https://github.com/getsentry/sentry-native/blob/bd56100922e1359c08f512e84ab975972788dc98/include/sentry.h#L26-L32

vaind:
> Hm, sorry if I missed something, but aren't these lines doing exactly that already?
> 
> https://github.com/getsentry/sentry-native/blob/bd56100922e1359c08f512e84ab975972788dc98/include/sentry.h#L26-L32

Yes, that's the automatic behaviour. This PR enables integrating SDKs to set e.g. sentry.android.native.**unity**

romtsn:
Ah ok, this makes sense 👍

romtsn (approved):

Issue #2026, opened 2022-05-09
Isolate cached events with hashed DSN folder


marandaneto:
### Description

See issue https://github.com/getsentry/sentry-cocoa/issues/1812 for context.
`SentryOptions#cacheDirPath` should be isolated per DSN.
We could hash the DSN and make it a folder, so events from DSN X don't leak to DSN Y.
It's a breaking change so likely do it on v7.

Update the specs as well to consider the isolation https://develop.sentry.dev/sdk/features/#buffer-to-disk

https://github.com/getsentry/sentry-cocoa/blob/e254b733f7706d4c04c02093f5ad0715ced92443/Sources/Sentry/SentryDsn.m#L31-L41

bruno-garcia:
It's out on 6.0.0-rc.1

PR #2038, opened 2022-05-11
Isolate cached events with hashed dsn subfolder

marandaneto:
## :scroll: Description
Isolate cached events with hashed dsn subfolder

Closes https://github.com/getsentry/sentry-java/issues/2026

## :pencil: Checklist

- [ ] No breaking changes

## :crystal_ball: Next steps
dd to the migration page.


sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -374,8 +380,46 @@ public void addIntegration(@NotNull Integration integration) {
...
+    dsnHash = calculateHashedDsn(this.dsn);
+  }
+
+  private @Nullable String calculateHashedDsn(final @Nullable String dsn) {
+    if (dsn == null || dsn.isEmpty()) {
+      return null;
+    }
+
+    try {
+      // getInstance() method is called with algorithm SHA-1
```
marandaneto, line 395: Similar to https://github.com/getsentry/sentry-cocoa/blob/e254b733f7706d4c04c02093f5ad0715ced92443/Sources/Sentry/SentryDsn.m#L31-L41 I believe

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -1486,16 +1534,11 @@ public void setProfilingEnabled(boolean profilingEnabled) {
...
-  }
-
-  /**
-   * Sets the profiling traces dir. path
-   *
-   * @param profilingTracesDirPath the profiling traces dir. path
-   */
-  public void setProfilingTracesDirPath(@Nullable String profilingTracesDirPath) {
-    this.profilingTracesDirPath = profilingTracesDirPath;
+    final String cacheDirPath = getCacheDirPath();
```
marandaneto, line 1537: `setProfilingTracesDirPath` is removed but `getProfilingTracesDirPath` still returns the same value, similar to `getOutboxPath` since its always a subdirectory.

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -374,8 +380,46 @@ public void addIntegration(@NotNull Integration integration) {
...
+      // Convert message digest into hex value
+      String hashtext = no.toString(16);
+
+      // Add preceding 0s to make it 32 bit
+      while (hashtext.length() < 32) {
+        hashtext = "0" + hashtext;
+      }
+
+      // return the HashText
+      return hashtext;
```
marandaneto, line 415: I could make a StringBuilder

stefanosiano (approved):
Looks good to me.
The only thing is whether we should cache the profilingTraceDirPath and outboxPath in `SentryOptions` instead of evaluating the cache file every time.
But they should already be cached by the sdk components that use them, so i think it's fine.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -299,9 +299,7 @@ private static void readDefaultOptionValues(
   private static void initializeCacheDirs(
       final @NotNull Context context, final @NotNull SentryAndroidOptions options) {
     final File cacheDir = new File(context.getCacheDir(), "sentry");
```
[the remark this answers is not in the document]
bruno-garcia: The hashed path doesn't need to be here, we can let the user define the "base path" and we always append the "basepath/hashed_dsn/stuff"

sentry/src/main/java/io/sentry/SentryOptions.java
bruno-garcia: yeah lets use the builder to avoid an allocation for each concatenation

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -374,8 +380,48 @@ public void addIntegration(@NotNull Integration integration) {
...
+  }
+
+  private @Nullable String calculateHashedDsn(final @Nullable String dsn) {
+    if (dsn == null || dsn.isEmpty()) {
+      return null;
+    }
+
+    try {
+      // getInstance() method is called with algorithm SHA-1
+      final MessageDigest md = MessageDigest.getInstance("SHA-1");
```
bruno-garcia, line 396: is SHA 256 available in all versions we support? If possible prefer that instead

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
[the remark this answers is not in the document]
marandaneto: That's how it is already.

marandaneto:
> Looks good to me. The only thing is whether we should cache the profilingTraceDirPath and outboxPath in `SentryOptions` instead of evaluating the cache file every time. But they should already be cached by the sdk components that use them, so i think it's fine.

It's only being used when creating the integrations, and it does not do IO, the File class just makes sure that it has the right file separator, etc.

sentry/src/main/java/io/sentry/SentryOptions.java
marandaneto: SHA-1 is there from API 1 https://developer.android.com/reference/java/security/MessageDigest
So leaving as it is, the `catch` is needed anyway tho.

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -597,7 +600,11 @@ public void setBeforeBreadcrumb(@Nullable BeforeBreadcrumbCallback beforeBreadcr
    * @return the cache dir. path or null if not set
    */
   public @Nullable String getCacheDirPath() {
-    return cacheDirPath;
+    if (cacheDirPath == null || cacheDirPath.isEmpty()) {
+      return null;
+    }
+
+    return dsnHash != null ? new File(cacheDirPath, dsnHash).getAbsolutePath() : cacheDirPath;
```
bruno-garcia, line 607: in what case can we have dsnHash here? I understand it's nullable but is it a valid execution path to get this thing called before we get a hold of a dsn? Would it be fair to return `null` or log with the diagnostic logger in that case? Just a way for us to find out this is happening

sentry/src/main/java/io/sentry/util/StringUtils.java
```diff
@@ -72,4 +80,50 @@ private StringUtils() {}
...
+      final byte[] messageDigest = md.digest(str.getBytes(UTF_8));
+
+      // Convert byte array into signum representation
+      final BigInteger no = new BigInteger(1, messageDigest);
+
+      // Convert message digest into hex value
+      final StringBuilder stringBuilder = new StringBuilder(no.toString(16));
+
+      // Add preceding 0s to make it 32 bit
+      while (stringBuilder.length() < 32) {
```
bruno-garcia, line 113: Do we need a 32 character long string?
I don't believe we're reducing the risk of collision with this so maybe we can go without this altogether. (seems like this came [from here](https://www.geeksforgeeks.org/sha-512-hash-in-java/), they likely needed a fixed size, but we don't).

sentry/src/main/java/io/sentry/util/StringUtils.java
```diff
@@ -72,4 +80,50 @@ private StringUtils() {}
...
+        stringBuilder.insert(0, "0");
+      }
+
+      // return the HashText
+      return stringBuilder.toString();
+    }
+
+    // For specifying wrong message digest algorithms
```
bruno-garcia, line 121: this is here because of checked exceptions or do we know there are some Androids out there without SHA1?

When can SHA1 be "wrong"?

romtsn (approved):

sentry/src/main/java/io/sentry/util/StringUtils.java
marandaneto: Already answered on slack.

sentry/src/main/java/io/sentry/SentryOptions.java
marandaneto: in case `dsn` is empty and SDK is disabled, `dsnHash` is null, SDK is NoOp anyway but we have to double-check due to NPE static code analysis.

sentry/src/main/java/io/sentry/util/StringUtils.java
marandaneto: Yep, no need to be fixed size, but iOS also uses something similar https://github.com/getsentry/sentry-cocoa/blob/e254b733f7706d4c04c02093f5ad0715ced92443/Sources/Sentry/SentryDsn.m
`CC_SHA1_DIGEST_LENGTH * 2` so it might be fixed size as well?

bruno-garcia: In iOS is just looks like it's allocating a string twice the size the number of bytes used in the digest creation.
It's a copy paste from:https://stackoverflow.com/questions/7570377/creating-sha1-hash-from-nsstring

See: 
image

bruno-garcia (approved):

bruno-garcia:
Other than prepending the `0`'s it LGTM

marandaneto:
> Other than prepending the `0`'s it LGTM

@adinauer or @romtsn in case you have the time to finish up this PR and merge it, it should be just 2 min :) thanks so much.

sentry/src/main/java/io/sentry/util/StringUtils.java
adinauer: Updated

PR #2042, opened 2022-05-12
Update Gradle to 7.4.2 and AGP to 7.2

marandaneto:
## :pencil: Checklist

- [ ] No breaking changes


romtsn (approved):

marandaneto:
@romtsn you likely wanna upgrade the SAGP as well https://github.com/getsentry/sentry-android-gradle-plugin/blob/main/.github/workflows/test-publish-dry-run.yaml#L32-L37=

romtsn:
Yeah for sure, but I don't think we can depend on the new 7.2 as a dependency yet... they changed the instrumentation API, so what we are using is now deprecated, maybe we'd need to do a version check and use proper apis based on that.

PR #2105, opened 2022-06-16
ci: auto-update gradle wrapper

vaind:
note: currently not verified because the wrapper is already up to date

#skip-changelog


.github/workflows/update-deps.yml
```diff
@@ -0,0 +1,21 @@
...
+    branches:
+      - main
+
+jobs:
+  gradle-wrapper:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: gradle-update/update-gradle-wrapper-action@981c551b17dbcf1940b1b4435afdb79babb7c13a
```
bruno-garcia, line 18: There's even a built-in action? 👀

bruno-garcia:
It'll fail on changelog though every time

philipphofmann:
Can't you achieve the same with [dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file)?

vaind:
> Can't you achieve the same with [dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file)?

I'm not aware of doing that with dependabot

marandaneto:
@vaind The problem I see is that there's a compat. matrix between Gradle and AGP https://developer.android.com/studio/releases/gradle-plugin#updating-gradle
If we bump Gradle automatically and not AGP, it could lead to either compile or runtime errors, this still has to be checked manually, I'm not sure if we should merge this.
@bruno-garcia opinions?

bruno-garcia:
> @vaind The problem I see is that there's a compat. matrix between Gradle and AGP [developer.android.com/studio/releases/gradle-plugin#updating-gradle](https://developer.android.com/studio/releases/gradle-plugin#updating-gradle) If we bump Gradle automatically and not AGP, it could lead to either compile or runtime errors, this still has to be checked manually, I'm not sure if we should merge this. @bruno-garcia opinions?

@marandaneto Ivan has been working on "custom" automation based on our needs. Including adding the changlog entry as per our format, example: https://github.com/getsentry/sentry-unity/pull/820

Could you please write down what needs to be done in order to have a successful bump PR?
The automation won't try to make the build pass in case there are breaking changes, but at least it automates opening up the PR and bumping the right files, and we can just pull that branch and fix breaking changes if needed.
The goal is to remove the need to keep up with version updates from AGP and Gradle, well, anything we can automate to free devs from the burden of bump PRs.

marandaneto:
@bruno-garcia I agree with the effort, I'm just worried about runtime errors, usually, I go thru the AGP changelog and check things manually and test out if there's something I'm not 100% sure.
AGP changelog: eg https://developer.android.com/studio/releases/gradle-plugin#7-2-0
To upgrade you just need to: `./gradlew wrapper --gradle-version=$version`
Commit the files and let CI to pass.
The AGP vs Gradle compatibility is here as well https://developer.android.com/studio/releases/gradle-plugin#updating-gradle

bruno-garcia:
> @bruno-garcia I agree with the effort, I'm just worried about runtime errors, usually, I go thru the AGP changelog and check things manually and test out if there's something I'm not 100% sure. 

What manual tests do you do? I'm looking for ways to automate as much as possible so we can reduce time spent in the manual validations.

AGP changelog: eg [developer.android.com/studio/releases/gradle-plugin#7-2-0](https://developer.android.com/studio/releases/gradle-plugin#7-2-0) To upgrade you just need to: `./gradlew wrapper --gradle-version=$version` Commit the files and let CI to pass.

Great! Thanks. cc @vaind 

The AGP vs Gradle compatibility is here as well [developer.android.com/studio/releases/gradle-plugin#updating-gradle](https://developer.android.com/studio/releases/gradle-plugin#updating-gradle)

Should we link to this in the PR body?

This automation would save us from having to keep track when there's a new version, creating a branch, making the bump and opening the PR. You'd already have something to start from in case manual intervention was required which is a huge step forward IMO

marandaneto:
@bruno-garcia 

> This automation would save us from having to keep track when there's a new version, creating a branch, making the bump and opening the PR. You'd already have something to start from in case manual intervention was required which is a huge step forward IMO

Agreed, as long as somebody reads the AGP's changelog and figures out if something has to be tested manually, that's all we need.
I just don't want to blindly merge & release a new version, what usually I test is depending on what I read on the changelog, that's not really helpful but I can't tell anything otherwise.

.github/workflows/update-deps.yml
```diff
@@ -0,0 +1,22 @@
...
+        with:
+          # TODO custom token can be used in order for CI to get triggered
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          labels: dependencies
```
marandaneto, line 22: Use the new token that allows CI to run.

vaind, line 22:
```suggestion
          repo-token: ${{ secrets.CI_DEPLOY_KEY }}
          labels: dependencies
```

marandaneto:
@vaind 

> Error: ❌ token *** is not a legal HTTP header value

CI is broken because of that, not sure if it's an issue with the action or our automation.

vaind:
> @vaind
> 
> > Error: ❌ token *** is not a legal HTTP header value
> 
> CI is broken because of that, not sure if it's an issue with the action or our automation.

https://github.com/getsentry/sentry-java/pull/2156

Issue #1850, opened 2021-12-16
Update SentryApolloInterceptor for Apollo v3


apramana:
[Apollo client v3](https://github.com/apollographql/apollo-kotlin) has been promoted to GA and there is a [new API for interceptors](https://github.com/apollographql/apollo-kotlin/blob/main/apollo-runtime/src/commonMain/kotlin/com/apollographql/apollo3/interceptor/ApolloInterceptor.kt).

PR #2109, opened 2022-06-19
Feat/apollo v3

lbloder:
## :scroll: Description
Add tracing support for Apollo v3.

Fixes https://github.com/getsentry/sentry-java/issues/1850

## :pencil: Checklist

 - https://github.com/getsentry/sentry-docs/pull/5289
- [ ] No breaking changes


buildSrc/src/main/java/Config.kt
```diff
@@ -132,6 +134,7 @@ object Config {
         val awaitility = "org.awaitility:awaitility-kotlin:4.1.1"
         val mockWebserver = "com.squareup.okhttp3:mockwebserver:${Libs.okHttpVersion}"
         val mockWebserver3 = "com.squareup.okhttp3:mockwebserver:3.14.9"
+        val mockWebserver4 = "com.squareup.okhttp3:mockwebserver:4.9.3"
```
marandaneto, line 137: Since it's a dev dep, can we update `sentry-apollo` to use the v4 as well instead?

buildSrc/src/main/java/Config.kt
```diff
@@ -104,6 +104,8 @@ object Config {
         val graphQlJava = "com.graphql-java:graphql-java:17.3"
 
         val kotlinReflect = "org.jetbrains.kotlin:kotlin-reflect"
+
+        val apollo3 = "com.apollographql.apollo3:apollo-runtime:3.3.0"
```
marandaneto, line 108: Should we call `apolloKotlin` instead?

> Apollo Kotlin (formerly known as Apollo Android) is a GraphQL client that generates Kotlin and Java models from GraphQL queries.

buildSrc/src/main/java/Config.kt
lbloder: Sounds good, will do

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3HttpInterceptor.kt
```diff
@@ -0,0 +1,162 @@
...
+import io.sentry.IHub
+import io.sentry.ISpan
+import io.sentry.SentryLevel
+import io.sentry.SpanStatus
+import io.sentry.TracingOrigins
+import io.sentry.TypeCheckHint
+import okio.Buffer
+import java.io.InputStreamReader
+
+class SentryApollo3HttpInterceptor(private val hub: IHub = HubAdapter.getInstance(), private val beforeSpan: BeforeSpanCallback? = null) :
```
marandaneto, line 20: We need the ctor with only hub and only `BeforeSpanCallback` as well due to Java compatibility

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3HttpInterceptor.kt
```diff
@@ -0,0 +1,162 @@
...
+        if (this != null && this != -1L) {
+            fn.invoke(this)
+        }
+    }
+
+    private fun startChild(request: HttpRequest, activeSpan: ISpan): ISpan {
+        val buffer = Buffer()
+        request.body?.writeTo(buffer)
+
+        val reader = InputStreamReader(buffer.inputStream())
```
marandaneto, line 74: We need to take into consideration `SentryOptions#maxRequestBodySize`

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3HttpInterceptor.kt
```diff
@@ -0,0 +1,162 @@
...
+
+        val content = try {
+            hub.options.serializer.deserialize(
+                reader,
+                ApolloRequestBodyContent::class.java,
+                ApolloRequestBodyContent.Deserializer
+            )
+        } catch (e: Exception) {
```
marandaneto, line 82: You likely always want to use a `Throwable`

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3HttpInterceptor.kt
```diff
@@ -0,0 +1,162 @@
...
+                span.status = SpanStatus.fromHttpStatusCode(httpResponse.statusCode, SpanStatus.UNKNOWN)
+                finish(span, modifiedRequest, httpResponse)
+
+                httpResponse
+            } catch (e: Exception) {
+                when (e) {
+                    is ApolloNetworkException -> span.status = SpanStatus.INTERNAL_ERROR
+                    else -> SpanStatus.UNKNOWN
+                }
+                finish(span, modifiedRequest)
```
marandaneto, line 58: Wouldn't be better to use a `finally` block that calls `finish`? otherwise, we duplicate code.

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3HttpInterceptor.kt
```diff
@@ -0,0 +1,162 @@
...
+            return try {
+                val httpResponse = chain.proceed(modifiedRequest)
+                span.status = SpanStatus.fromHttpStatusCode(httpResponse.statusCode, SpanStatus.UNKNOWN)
+                finish(span, modifiedRequest, httpResponse)
+
+                httpResponse
+            } catch (e: Exception) {
+                when (e) {
+                    is ApolloNetworkException -> span.status = SpanStatus.INTERNAL_ERROR
+                    else -> SpanStatus.UNKNOWN
```
marandaneto, line 56: That looks wrong, every error would be `UNKNOWN` if not thrown by network issues, the `finally` block would be better cus there might be an error but at least we have a HTTP status code?

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3HttpInterceptor.kt
```diff
@@ -0,0 +1,162 @@
...
+        request.body?.writeTo(buffer)
+
+        val reader = InputStreamReader(buffer.inputStream())
+
+        val content = try {
+            hub.options.serializer.deserialize(
+                reader,
+                ApolloRequestBodyContent::class.java,
+                ApolloRequestBodyContent.Deserializer
+            )
```
marandaneto, line 81: Maybe we just copy the body array and let the serialization happen during the event processing which is in a background thread? otherwise, this would affect the app's performance in the calling thread, it could be the main thread, `CachedBodyHttpServletRequest` does something similar I guess.

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3HttpInterceptor.kt
```diff
@@ -0,0 +1,162 @@
...
+                it.set(TypeCheckHint.APOLLO_RESPONSE, httpResponse)
+            }
+
+            hub.addBreadcrumb(breadcrumb, hint)
+        }
+    }
+
+    private fun parseOperationType(content: ApolloRequestBodyContent?): String {
+        val operationPart = content?.query?.takeWhile { !it.isWhitespace() }
+        return KNOWN_OPERATIONS.firstOrNull { it == operationPart } ?: "unknown"
```
marandaneto, line 142: `SentryApolloInterceptor` fall back to the class simple name, ideally we'd follow the same pattern, `unknown` isn't useful.

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3HttpInterceptor.kt
lbloder: Will double check but, afaik we get a httpResponse for all Status Codes including 4xx and 5xx in the HttpInterceptor

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3HttpInterceptor.kt
lbloder: We don't have access to the High-Level Operation. So we can't use the class simple name here.
In Apollo3 they split up the interceptors into Application and HttpInterceptors.

I started with the ApplicationInterceptor but ran into a few issues that could be resolved by using the HttpInterceptor instead. For example we didn't have access to request/response body to add their sizes to the breadcrumb. Also there was no way to get the request url.

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3HttpInterceptor.kt
lbloder: The request/response bodies are not sent to Sentry. Only used for the Hints like in the original SentryApolloInterceptor. Should we send the bodies to the server?

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3HttpInterceptor.kt
marandaneto: Gotcha, I'd use the interceptor that makes more sense, give us richer values, even if that means dropping request/response size, the `url` is important only if it has something useful.

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3HttpInterceptor.kt
lbloder: By default the calling thread is a Background thread, ApolloClient uses Dispatcher.IO, unless explicitly defined otherwise.
The calling thread itself, however, is blocked, so it I'll have a look at using an `EventProcessor`

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3HttpInterceptor.kt
lbloder: Is it the two calls to `finish` that are bugging you?

`finish(span, modifiedRequest, httpResponse)` // Success i.e. there is a response from the server
`finish(span, modifiedRequest)` // Error i.e. no response from the server, or a local exception like no connectivity

The alternative would look like this:
```
var response: HttpResponse? = null

response = try {
    chain.proceed(modifiedRequest)
} catch (e: Throwable) {
    when (e) {
        is ApolloNetworkException -> span.status = SpanStatus.INTERNAL_ERROR
        else -> SpanStatus.UNKNOWN
    }
    throw e
} finally {
    span.status = SpanStatus.fromHttpStatusCode(response?.statusCode, SpanStatus.UNKNOWN)
    finish(span, modifiedRequest, response)
}

return response!!
```

Which, tbh, I don't like much. Especially since we need to force unwrap `response` at the end.

wdyt?

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3HttpInterceptor.kt
lbloder: With the `url` we can check against `TraceOrigins` and decide whether to set the trace/baggage headers.

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3HttpInterceptor.kt
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+        val operationId = request.headers.firstOrNull { it.name == "X-APOLLO-OPERATION-ID" }?.value ?: "unknown"
+
+        return span.apply {
+            setData("operationId", operationId)
+        }
+    }
+
+    private fun createSpanFromBodyContent(activeSpan: ISpan, request: HttpRequest, content: ApolloRequestBodyContent): ISpan {
```
adinauer, line 101: Is it worth deserializing the body for retrieving operation and variables? Should this be an opt-in?

Also is there a PII concern with variables?

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3HttpInterceptor.kt
marandaneto: Nope, that's not what I meant, unless this depends on `request.body`, if the graph query and operation are in the `body`, then we need it.
`ApolloInterceptor` offers access such as `request.operation.name()` but not sure if there's everything either.
I guess you both are on top of this, I was just wondering if the deserialization was necessary, if it is, well... nothing to do.

lbloder:
New approach:
- Add a custom `RequestComposer` that sets internal headers that can then be read by the interceptor to get information about the call (e.g. variables). The interceptor removes these headers afterwards.
- When these headers are not present, no information about the `operationType` or `variables` are added.

Drawbacks:
- Users need to add both our `SentryApollo3RequestComposer` by using the `httpTransport` builder AND the `SentryApollo3HttpInterceptor` to their `ApolloClient` configuration.

```
ApolloClient.builder()
    .networkTransport(
        HttpNetworkTransport.Builder()
            .httpRequestComposer(SentryApollo3RequestComposer("urll"))
            .addInterceptor(SentryApollo3HttpInterceptor())
            .build()
    ).build()
```

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3HttpInterceptor.kt
```diff
@@ -46,69 +47,64 @@ class SentryApollo3HttpInterceptor(private val hub: IHub = HubAdapter.getInstanc
...
-    private fun createSpanFromBodyContent(activeSpan: ISpan, request: HttpRequest, content: ApolloRequestBodyContent): ISpan {
-        val operationType = parseOperationType(content)
-        val operationName = content.operationName ?: request.headers.firstOrNull { it.name == "X-APOLLO-OPERATION-NAME" }?.value ?: "unknown"
-        val description = "$operationType $operationName"
-
-        return activeSpan.startChild(operationName, description).apply {
-            setData("variables", content.variables.toString())
-        }
+    private fun operationNameFromHeaders(request: HttpRequest): String? {
+        return request.valueForHeader(SENTRY_APOLLO_3_OPERATION_NAME) ?: request.valueForHeader("X-APOLLO-OPERATION-NAME")
```
adinauer, line 102: Can we add a comment somewhere to document that apollo headers are not present for batched operations.

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3HttpInterceptor.kt
```diff
@@ -119,34 +115,44 @@ class SentryApollo3HttpInterceptor(private val hub: IHub = HubAdapter.getInstanc
         }
         newSpan.finish()
```
adinauer, line 116: `BeforeSpanCallback` may return `null`, why does this even compile?

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3RequestComposer.kt
```diff
@@ -0,0 +1,39 @@
...
+    private val defaultHttpRequestComposer = DefaultHttpRequestComposer(url)
+
+    override fun <D : Operation.Data> compose(apolloRequest: ApolloRequest<D>): HttpRequest {
+        val httpRequest = defaultHttpRequestComposer.compose(apolloRequest)
+        val builder = httpRequest.newBuilder()
+            .addHeader(SentryApollo3HttpInterceptor.SENTRY_APOLLO_3_OPERATION_TYPE, operationType(apolloRequest))
+            .addHeader(SentryApollo3HttpInterceptor.SENTRY_APOLLO_3_OPERATION_NAME, apolloRequest.operation.name())
+
+        apolloRequest.scalarAdapters?.let {
+            builder.addHeader(SentryApollo3HttpInterceptor.SENTRY_APOLLO_3_VARIABLES, apolloRequest.operation.variables(it).valueMap.toString())
```
adinauer, line 24: Will this blow up if there's long variables? Do we need to truncate / skip if too long?

sentry-apollo-3/src/test/java/io/sentry/apollo3/SentryApollo3InterceptorWithComposerTest.kt
```diff
@@ -0,0 +1,232 @@
...
+            anyOrNull()
+        )
+    }
+
+    @Test
+    fun `when there is no active span, does not add sentry trace header to the request`() {
+        executeQuery(isSpanActive = false)
+
+        val recorderRequest = fixture.server.takeRequest()
+        assertNull(recorderRequest.headers[SentryTraceHeader.SENTRY_TRACE_HEADER])
```
adinauer, line 136: Can we add baggage as well please? Same below.

adinauer (changes requested):
Nicely done 👏  only a few more things, otherwise LGTM.

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3HttpInterceptor.kt
adinauer: Devs can always use callbacks we provide to read request / response from hints and attach them if they want.

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3HttpInterceptor.kt
lbloder: The BeforeSpanCallback is currently implemented as returning a NonNull Value. Same as in ApolloInterceptor V2. Returning null is dropping the span, correct? I think we should open an issue for V2 then.

sentry/src/main/java/io/sentry/ISerializer.java
```diff
@@ -12,6 +12,9 @@
 public interface ISerializer {
   <T> @Nullable T deserialize(@NotNull Reader reader, @NotNull Class<T> clazz);
 
+  <T> @Nullable T deserialize(
```
lbloder, line 15: @adinauer Since we don't do serialization inside the interceptor anymore, this change could be reverted. Or is this something we want to keep?

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3RequestComposer.kt
lbloder: Hmm, good point... Don't think it would blow up because header sizes are usually limited by the server side. However, in order to keep our request the Sentry Server small we should truncate/skip.
Is that something we could/should use `SentryOptions#maxRequestBodySize` for?

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3HttpInterceptor.kt
lbloder: `RequestHeaders` are present on batched operations as well. However, the `ResponseHeaders` are stripped in batched operations.

Working around it by trying the header first and if that doesn't have a valid size, I'm querying the buffer.

```
            httpResponse.headersContentLength().ifHasValidLength { contentLength ->
                breadcrumb.setData("response_body_size", contentLength)
            }

            if (!breadcrumb.data.containsKey("response_body_size")) {
                httpResponse.body?.buffer?.size?.ifHasValidLength { contentLength ->
                    breadcrumb.setData("response_body_size", contentLength)
                }
            }
```

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3HttpInterceptor.kt
marandaneto: The request body should be by default depending on the configuration -> https://docs.sentry.io/platforms/java/configuration/options/#max-request-body-size
This is part of our spec, let's open a separate issue and track it somewhere to implement in this interceptor and in the missing ones, not a blocker for this PR since the apollo v2 does not have it either.

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3RequestComposer.kt
marandaneto: `maxRequestBodySize` is about the body only, headers are added as it is, IIRC this should be guarded by `defaultPii`

sentry-apollo-3/src/test/java/io/sentry/apollo3/SentryApollo3InterceptorWithComposerTest.kt
marandaneto: If it takes too much time, we can unblock folks first with apollo v3 and add the baggage next.

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3RequestComposer.kt
lbloder: @marandaneto Actually these headers are meant to be passed internally only in order to work around the need to deserialize.
They are added by the `SentryApollo3RequestComposer` and then read by the `SentryApollo3HttpInterceptor` and then removed. Only if people use the `RequestComposer` without the interceptor will the headers be sent over the wire. But then they don't get any use out of it as the `Interceptor` is the one to send info to our Backend.

I think Alex's question was about size limits for headers in general.

As for my idea with the `maxRequestBodySize`: Depending on the Apollo Operation the variables could get quite big and would therefore increase our payload sent to the Backend. Should we set a fixed sizeLimit or just send the variables as they are?

We thought about guarding with `defaultPii` but decided against it, correct @adinauer ?

sentry-apollo-3/src/test/java/io/sentry/apollo3/SentryApollo3InterceptorWithComposerTest.kt
lbloder: Added in the `SentryApollo3InterceptorTest`

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3RequestComposer.kt
marandaneto: I'd just send the variables as they are.
We don't do any trimming on the client, the server drops if it exceeds and reports to the `client_reports` feature.
They can use the filtering callbacks to drop stuff in case they need to.

@lbloder using the request's header for this workaround could be problematic, in case this is a request for a 3rd party and somebody isn't using the interceptor, we can be leaking some PII data, I guess?
If there's no way around getting everything we need either with both interceptors, I'd rather go with the option that has more useful data and open an issue on the `apollo3` repo with a feature request explaining our use case.
Maybe @bruno-garcia has opinions here.

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3HttpInterceptor.kt
adinauer: Opened https://github.com/getsentry/sentry-java/issues/2148 to track

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3HttpInterceptor.kt
adinauer: Ah oversight on my part, didn't see there's multiple interfaces with different signatures. Created https://github.com/getsentry/sentry-java/issues/2149 to changes this in v2 to allow returning `null`.

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3RequestComposer.kt
adinauer: > We thought about guarding with defaultPii but decided against it, correct @adinauer ?

Yes, was decided internally to not have this behind PII flag as devs can strip the variables in callbacks if they want to.

> @lbloder using the request's header for this workaround could be problematic, in case this is a request for a 3rd party and somebody isn't using the interceptor, we can be leaking some PII data, I guess?

We'd only be sending the data to the server where the variables are sent to anyways.

sentry/src/main/java/io/sentry/ISerializer.java
adinauer: I'd remove for now, as it seems easy to add back when needed.

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3RequestComposer.kt
marandaneto: @adinauer about the PII flags, headers usually are a source of tokens, etc, which is different than some other stuff such as navigation args, if you all have considered this use case and agreed with it, I'd double check tho, I recall using this PII guard for headers elsewhere and this was a must.

marandaneto: @adinauer 

> We'd only be sending the data to the server where the variables are sent to anyways.

Headers are not compressed by default on normal SSL iirc, not sure if this is a concern.
Please ping @bruno-garcia if needed.

adinauer: We're not just taking all headers but only selected one's. In `SentryRequestResolver` for Spring we take all headers but exclude certain ones we deem sensitive.

https://github.com/getsentry/sentry-java/blob/413862e108edbf58285cf3b5c42de134c96a3bbe/sentry-spring/src/main/java/io/sentry/spring/SentryRequestResolver.java#L20-L21

adinauer: > Headers are not compressed by default on normal SSL iirc, not sure if this is a concern.

They should be encrypted if that's what you're saying. As for compression I don't know but it'd only happen if they misconfigure anyways.

lbloder: > Right now, we don't send the Request or its Headers to our Backend in the Interceptors, neither are we doing this in e.g. SentryOkHttpInterceptor nor in the SentryFeignClient if I'm not mistaken. @marandaneto You said this is in our spec? Does this apply to mobile as well? Should we open issues for the other integrations as well?

Sorry, just found your original comment.  

As for my workaround with using headers to pass data between `Interceptors`, I don't see a problem if our headers are leaked, except that the request is getting bigger, because the headers we add only contain info that is in the request body anyways.

In terms of sending the whole request to the server, then I would definitely do it like in the `SentryRequestResolver` that @adinauer linked

marandaneto: If we are not sending headers in this PR to Sentry, no need to worry, but if we do, we should guard against defaultPii, See https://github.com/getsentry/sentry-dart/blob/d75a8c698a4942816336f3bae9a50c87fa6be0a8/dio/lib/src/dio_event_processor.dart#L117

lbloder: Made it easier to use with an extension function that adds both interceptors. Also this should make it harder to misconfigure,

The only headers we would send, and only if the `SentryApollo3HttpInterceptor` isn't added, are those that contain `variables` `operationName` and `operationType` which are all present in the request body anyways.

I would split sending the `request` and its `headers` to a separate issue for the integrations that are not sending those yet. I.e. The Apollo Integration v2, the OkHttp Integration, the FeignClient and maybe others, will look through them and create issues for those that are not sending the `request` and its `headers` to our backend.

Wdyt?

.github/ISSUE_TEMPLATE/bug_report_java.yml
```diff
@@ -12,6 +12,7 @@ body:
         - sentry-jul
         - sentry-jdbc
         - sentry-apollo
+        - sentry-apollo-3
```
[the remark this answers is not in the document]
lbloder: @marandaneto should I wait with adapting the `craft.yml` until the release registry PR is approved?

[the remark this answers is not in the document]
marandaneto: I'd comment it out, for now, https://github.com/getsentry/sentry-release-registry/pull/78 isn't done yet and 6.2.0 is already outdated.

bruno-garcia:
fyi @marandaneto is OOO until the 20th, and @romtsn will be away for 60 days from the 20th.

@adinauer might be able to help with reviews here or get a hold of folks internally

sentry-apollo-3/src/main/java/io/sentry/apollo3/SentryApollo3RequestComposer.kt
adinauer: We have https://github.com/getsentry/sentry-java/issues/2148 to track.

adinauer (approved):
LGTM thanks for doing this. Just need to wait for https://github.com/getsentry/sentry-release-registry/pull/78 to be merged or comment out changes to `.craft.yml`.

Issue #2096, opened 2022-06-14
Jetpack Compose Navigation support


romtsn:
### Description

Add support for Jetpack Compose [Navigation library](https://developer.android.com/jetpack/compose/navigation)

- [x] Collect breadcrumbs for navigation/route changes, capturing from/to arguments as `data` (similar to [flutter](https://github.com/getsentry/sentry-dart/blob/main/flutter/lib/src/navigation/sentry_navigator_observer.dart)) #2116 and #2121 
- [x] Capture transactions for navigation/route changes (transactions should be idle) #2137 
- [x] Update [release registry](https://github.com/getsentry/sentry-release-registry) and .craft.yml 
	- [x] Craft https://github.com/getsentry/sentry-java/pull/2138
	- [x] Release registry https://github.com/getsentry/sentry-release-registry/pull/76 
- [ ] Update docs for:
	- [ ] New Navigation integration https://github.com/getsentry/sentry-docs/pull/5229
	- [ ] New Compose integration (android + desktop)
	- [ ] Migration docs (Kotlin bump to 1.6.10) https://github.com/getsentry/sentry-docs/pull/5219

romtsn:
It'd be awesome to expand this to support Compose desktop/web as well, but the Jetpack Navigation lib only supports Android at the moment (see issue for tracking it https://github.com/JetBrains/compose-jb/issues/85), so the best we can have now is just UI breadcrumbs/tracing for compose desktop.

PR #2116, opened 2022-06-20
feat: Add sentry-android-navigation module

romtsn:
_#skip-changelog_

## :scroll: Description

* Adds a new `sentry-android-navigation` module to support androidx.navigation library (only breadcrumbs for now)

This is part of the Jetpack Compose Navigation support.

Part of #2096 

## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next steps
* Tracing will be done in a separate PR


sentry-android-navigation/src/main/java/io/sentry/android/navigation/SentryNavigationListener.kt
```diff
@@ -0,0 +1,57 @@
...
+            val from = fromRef?.get()?.route
+            from?.let { data["from"] = it }
+            fromArgs?.let { args ->
+                data["from_arguments"] = args.keySet().filter {
+                    it != NavController.KEY_DEEP_LINK_INTENT // there's a lot of unrelated stuff
+                }.associateWith { args[it] }
+            }
+            val to = destination.route
+            to?.let { data["to"] = it }
+            // TODO: should this be hidden behind Pii flag?
```
romtsn, line 45: I guess we should do it, because the user may pass username + password between composables/screens and this might be a problem?

sentry-android-navigation/src/main/java/io/sentry/android/navigation/SentryNavigationListener.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    private fun addBreadcrumb(destination: NavDestination, arguments: Bundle?) {
+        val breadcrumb = Breadcrumb().apply {
+            type = "navigation"
+            category = "navigation"
+
+            val from = fromRef?.get()?.route
+            from?.let { data["from"] = it }
+            fromArgs?.let { args ->
+                data["from_arguments"] = args.keySet().filter {
```
romtsn, line 39: I think we should not send an empty map here, otherwise it doesn't look nice, wdyt?
image

sentry-android-navigation/src/main/java/io/sentry/android/navigation/SentryNavigationListener.kt
philipphofmann: Better to be safe than sorry. People could pass random PII things as arguments. This also applies for the other arguments.

philipphofmann (commented):
This looks incredibly simple.

sentry-android-navigation/src/main/java/io/sentry/android/navigation/SentryNavigationListener.kt
marandaneto: We don't use `defaultPii` for such things, not on Flutter nor React Native, the reason is to reproduce issues you need this data and it's not related to PII (most likely), if there's PII and they don't want to, they can strip out using `beforeBreadcrumb`, or `beforeSend` or even on the server-side, maybe not even using it at all.

sentry-android-navigation/src/main/java/io/sentry/android/navigation/SentryNavigationListener.kt
```diff
@@ -0,0 +1,57 @@
...
+    ) {
+        addBreadcrumb(destination, arguments)
+        fromRef = WeakReference(destination)
+        fromArgs = arguments
+    }
+
+    private fun addBreadcrumb(destination: NavDestination, arguments: Bundle?) {
+        val breadcrumb = Breadcrumb().apply {
+            type = "navigation"
+            category = "navigation"
```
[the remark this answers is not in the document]
romtsn: category is the first time we use, but the type is already used for activity/fragment/app lifecycle events.

Yes, we do have it on other platforms already for a while (js, flutter) and also here https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/ (look for navigation)

sentry-android-navigation/src/main/java/io/sentry/android/navigation/SentryNavigationListener.kt
```diff
@@ -0,0 +1,57 @@
...
+
+            val from = fromRef?.get()?.route
+            from?.let { data["from"] = it }
+            fromArgs?.let { args ->
+                data["from_arguments"] = args.keySet().filter {
+                    it != NavController.KEY_DEEP_LINK_INTENT // there's a lot of unrelated stuff
+                }.associateWith { args[it] }
+            }
+            val to = destination.route
+            to?.let { data["to"] = it }
```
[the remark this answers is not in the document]
romtsn: Because of the above, let's stick to the docs and other sdks

romtsn:
> This looks incredibly simple.

Is it good or bad? 🙈 haha

romtsn:
@adinauer add you to keep you in the loop, feel free to delegate to Lukas if you think it's better

sentry-android-navigation/src/main/java/io/sentry/android/navigation/SentryNavigationListener.kt
```diff
@@ -0,0 +1,57 @@
...
+import io.sentry.Breadcrumb
+import io.sentry.Hint
+import io.sentry.HubAdapter
+import io.sentry.IHub
+import io.sentry.SentryLevel.INFO
+import io.sentry.TypeCheckHint
+import java.lang.ref.WeakReference
+
+class SentryNavigationListener(
+    private val hub: IHub = HubAdapter.getInstance()
```
marandaneto, line 15: We need an empty ctor to be Java friendly as well.

romtsn: I think @JvmOverloads should do? https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-jvm-overloads/

marandaneto: If it generates without any params as well, that should work, we could even use this annotation within the other ctors, IIRC we've not used this annotation yet due to likely not knowing about it, good catch.

sentry-android-navigation/build.gradle.kts
```diff
@@ -0,0 +1,86 @@
...
+}
+
+kotlin {
+    explicitApi()
+}
+
+dependencies {
+    api(projects.sentry)
+
+    implementation(Config.Libs.navigationRuntime)
```
marandaneto, line 70: Quick question, What's about the other 2.x versions? are they all compatible with the public APIs that we use?
I'm wondering if we actually should start using `compileOnly`

sentry-android-navigation/src/main/java/io/sentry/android/navigation/SentryNavigationListener.kt
romtsn: I think it should generate yeah, will verify it

romtsn: Yeah, it does 
image

romtsn: filed https://github.com/getsentry/sentry-java/issues/2124

sentry-android-navigation/build.gradle.kts
romtsn: Yeah looks like it's compatible all the way to version 1.0.0-alpha10 actually. We could use `compileOnly` yes, I think it'd be the right approach, rather than silently updating user's dependency tree to our version (if they have an older one).

marandaneto: Maybe start using `compileOnly` for this one and raise an issue for the other dependencies?
We could even have a matrix test in the future that runs with multiple versions of the transitive dependencies, similar to SAGP.

bruno-garcia (approved):
Before we can release this we need to add to .craft.yml in this repo, and https://github.com/getsentry/sentry-release-registry

sentry-android-navigation/src/main/java/io/sentry/android/navigation/SentryNavigationListener.kt
```diff
@@ -0,0 +1,64 @@
...
+import io.sentry.SentryLevel.INFO
+import io.sentry.TypeCheckHint
+import java.lang.ref.WeakReference
+
+class SentryNavigationListener @JvmOverloads constructor(
+    private val hub: IHub = HubAdapter.getInstance()
+) : NavController.OnDestinationChangedListener {
+
+    private var previousDestinationRef: WeakReference<NavDestination>? = null
+    private var previousArgs: Bundle? = null
```
adinauer, line 19: Should the `Bundle` also be wrapped with `WeakReference`?

adinauer (approved):
Only one question, rest LGTM!

sentry-android-navigation/src/main/java/io/sentry/android/navigation/SentryNavigationListener.kt
romtsn: nope, because it doesn't hold any reference to activity/context and it's just an android-specific version of `Map`, so we are good here

sentry-android-navigation/src/test/java/io/sentry/android/navigation/SentryNavigationListenerTest.kt
```diff
@@ -0,0 +1,117 @@
...
+import io.sentry.Breadcrumb
+import io.sentry.IHub
+import io.sentry.SentryLevel
+import org.junit.runner.RunWith
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNull
+
+@RunWith(AndroidJUnit4::class)
+class SentryNavigationListenerTest {
```
philipphofmann, line 21: Nice tests 👌

philipphofmann (approved):
LGTM 🚀

sentry-android-navigation/src/main/java/io/sentry/android/navigation/SentryNavigationListener.kt
bruno-garcia: > We don't use defaultPii for such things, not on Flutter nor React Native, the reason is to reproduce issues you need this data and it's not related to PII (most likely), if there's PII and they don't want to, they can strip out using beforeBreadcrumb, or beforeSend or even on the server-side, maybe not even using it at all.

Agreed, but this has the assumption it's unlikely to have PII. Roman mentioned there could be login/password. 
If so, I'd suggest adding behind `sendDefaultPii`. We haven't used much for the lack of use cases, but the feature is there.

PR #2117, opened 2022-06-20
New sample with min api level

marandaneto:
Fixes https://github.com/getsentry/sentry-java/issues/1403

## :pencil: Checklist

- [ ] No breaking changes


sentry-android-core/lint-baseline.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<issues format="6" by="lint 7.2.0" type="baseline" client="gradle" dependencies="true" name="AGP (7.2.0)" variant="all" version="7.2.0">
+
+    <issue
+        id="TrulyRandom"
```
marandaneto, line 5: Needed because we can't use `SuppressLint` on Java projects, and it generates 1 per project.

sentry-android-okhttp/lint-baseline.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
marandaneto, line 1: No issue for this module, but for some reason, it requires the baseline, likely a bug.

sentry-samples/sentry-samples-android-minsdk/build.gradle.kts
```diff
@@ -0,0 +1,90 @@
...
+    }
+}
+
+dependencies {
+    implementation(fileTree(mapOf("dir" to "libs", "include" to listOf("*.jar"))))
+
+    implementation(projects.sentryAndroid)
+    implementation(projects.sentryAndroidFragment)
+    implementation(projects.sentryAndroidTimber)
+    implementation(projects.sentryApollo)
```
marandaneto, line 87: So all projects with lower API is here

sentry/src/main/java/io/sentry/Baggage.java
```diff
@@ -15,7 +16,7 @@
 @ApiStatus.Experimental
 public final class Baggage {
 
-  static final @NotNull String CHARSET = "UTF-8";
+  static final @NotNull String CHARSET = StandardCharsets.UTF_8.toString();
```
marandaneto, line 19: For some reason this is not being detected, I guess AGP regressed here.

> Warning: Lint will treat :sentry as an external dependency and not analyze it.
>  * Recommended Action: Apply the 'com.android.lint' plugin to java library project :sentry. to enable lint to analyze those sources.

The linter is able to analyze and check everything but does not consider the `minApi` apparently.

marandaneto: I should revert this change and merge it anyway since I found out other issues with the linter.

sentry-apollo/build.gradle.kts
```diff
@@ -8,6 +8,7 @@ plugins {
     id(Config.QualityPlugins.errorProne)
     id(Config.QualityPlugins.gradleVersions)
     id(Config.BuildPlugins.buildConfig) version Config.BuildPlugins.buildConfigVersion
+    id(Config.QualityPlugins.androidLint)
```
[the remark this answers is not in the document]
adinauer: Apollo is also available for non Android, not sure how we would add it

sentry/src/main/java/io/sentry/Baggage.java
adinauer: So this means we'll have to be careful and check manually in Android Studio if there's any minApi breakages we introduced?

sentry-apollo/build.gradle.kts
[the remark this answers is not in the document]
marandaneto: @adinauer what do you mean? This is the Apollo module and I'm adding the `lint` to it, so it's alright.

sentry/src/main/java/io/sentry/Baggage.java
marandaneto: If we don't find a fix for this, yes, which sucks, this used to work.
@romtsn ideas?

sentry-apollo/build.gradle.kts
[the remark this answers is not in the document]
adinauer: Yes I was trying to say that we probably can't just add the android plugin to it to gain `android {}` in there as this can also be used by non Android applications.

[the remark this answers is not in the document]
marandaneto: I managed to config using `configure<Lint>` but it still does not detect old APIs compatibility.

sentry/src/main/java/io/sentry/Baggage.java
romtsn: answered in https://github.com/getsentry/sentry-java/pull/2117/files#r902318872

marandaneto: @romtsn where do you usually file your AGP bugs? I've created a repro https://github.com/marandaneto/NewApiLintTest

sentry-apollo/build.gradle.kts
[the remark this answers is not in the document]
adinauer: Could we simply run grep on the logs instead and check for minApi warnings?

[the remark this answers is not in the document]
romtsn: But which logs do you want to run grep on? You need to be able to run lint first :)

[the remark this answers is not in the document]
adinauer: I thought it works but we can't customize it to fail the build for minSdk violations.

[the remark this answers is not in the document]
marandaneto: the Linter runs already, but it does not check for `newApi`.

sentry/src/main/java/io/sentry/Baggage.java
marandaneto: This used to work in the past, so I'd consider either a bug or a non-documented breaking change.

marandaneto: Yeah, that's strange, I'm pretty sure that it worked in the past when developing sentry-android v2 unless I'm mistaken, I don't recall leaking NewApi tho so I strongly think that it worked in the past.
We could try this workaround, it's a good idea.
I just quickly tried and it requires a bit of more work, I'd stop here tho, feel free to add this to our backlog and eventually prioritize finishing up.

romtsn: Sounds good, @adinauer or I could take it from there, thanks 👍

sentry-apollo/build.gradle.kts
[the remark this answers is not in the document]
marandaneto: https://github.com/open-toast/gummy-bears
This can help
have you used/tried this @romtsn @markushi ?

bruno-garcia:
👋

marandaneto:
Outdated, a new PR would make more sense.

PR #2121, opened 2022-06-21
feat: Add compose module (navigation)

romtsn:
_#skip-changelog_

## :scroll: Description

* Adds Jetpack compose module (multiplatform setup to later support Compose/Desktop)
* Adds support for instrumenting NavController in compose

Part of #2096

## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next step


buildSrc/src/main/java/Config.kt
```diff
@@ -19,14 +21,16 @@ object Config {
         val grettyVersion = "4.0.0"
         val gradleMavenPublishPlugin = "com.vanniktech:gradle-maven-publish-plugin:0.18.0"
         val dokkaPlugin = "org.jetbrains.dokka:dokka-gradle-plugin:$kotlinVersion"
+        val composeGradlePlugin = "org.jetbrains.compose:compose-gradle-plugin:$composeVersion"
     }
 
     object Android {
-        private val sdkVersion = 31
```
romtsn, line 25: We need this for compose latest version

buildSrc/src/main/java/Config.kt
```diff
@@ -1,14 +1,16 @@
 import java.math.BigDecimal
 
 object Config {
-    val kotlinVersion = "1.5.31"
```
romtsn, line 4: We alos need this to support compose latest version. @marandaneto do you see any problems with bumping this for the whole monorepo? Kotlin now is at 1.7.

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryNavigationIntegration.kt
```diff
@@ -0,0 +1,52 @@
...
+
+    fun dispose() {
+        navController.removeOnDestinationChangedListener(navListener)
+    }
+}
+
+// As described in https://developer.android.com/codelabs/jetpack-compose-advanced-state-side-effects#6
+@Composable
+@NonRestartableComposable
+public fun NavHostController.withObservableEffect(): NavHostController {
```
romtsn, line 39: naming is hard, open for suggestions here

buildSrc/src/main/java/Config.kt
```diff
@@ -1,14 +1,16 @@
...
+    val kotlinVersion = "1.6.10"
```
romtsn, line 4: We also need this to support compose latest version. @marandaneto do you see any problems with bumping this for the whole monorepo? Kotlin now is at 1.7.

buildSrc/src/main/java/Config.kt
```diff
@@ -19,14 +21,16 @@ object Config {
         val grettyVersion = "4.0.0"
         val gradleMavenPublishPlugin = "com.vanniktech:gradle-maven-publish-plugin:0.18.0"
         val dokkaPlugin = "org.jetbrains.dokka:dokka-gradle-plugin:$kotlinVersion"
+        val composeGradlePlugin = "org.jetbrains.compose:compose-gradle-plugin:$composeVersion"
     }
 
     object Android {
-        private val sdkVersion = 31
+        private val sdkVersion = 32
```
romtsn, line 28: We need this for compose latest version

buildSrc/src/main/java/Config.kt
marandaneto: We still compile with `kotlinCompatibleLanguageVersion = "1.4"`.
Is that still possible with Kotlin 1.6, right?
We can consider using `1.5` for the `kotlinCompatibleLanguageVersion`.
Do we know the usage of Kotlin versions?
It's a matter of back-compatibility here.

buildSrc/src/main/java/Config.kt
marandaneto: Is that stable already? if so, no problem.
If it's not yet, we can consider using 32 only for the compose packages.

buildSrc/src/main/java/Config.kt
```diff
@@ -104,8 +108,11 @@ object Config {
         val graphQlJava = "com.graphql-java:graphql-java:17.3"
 
         val kotlinReflect = "org.jetbrains.kotlin:kotlin-reflect"
+        val kotlinStdLib = "org.jetbrains.kotlin:kotlin-stdlib"
 
         val navigationRuntime = "androidx.navigation:navigation-runtime:2.4.2"
+        // compose deps
+        val composeNavigation = "androidx.navigation:navigation-compose:2.4.2"
```
marandaneto, line 115: If they follow the same versioning, we can extract `2.4.2` to a const.

sentry-compose/build.gradle.kts
```diff
@@ -0,0 +1,121 @@
...
+import io.gitlab.arturbosch.detekt.extensions.DetektExtension
+
+plugins {
+    kotlin("multiplatform")
+    id("com.android.library")
+    id("org.jetbrains.compose")
+    jacoco
+    id(Config.QualityPlugins.gradleVersions)
+    id(Config.QualityPlugins.detektPlugin)
+    `maven-publish`
```
marandaneto, line 11: isn't `com.vanniktech:gradle-maven-publish-plugin` applied to every module already?

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryNavigationIntegration.kt
```diff
@@ -0,0 +1,52 @@
...
+import androidx.navigation.NavHostController
+import io.sentry.HubAdapter
+import io.sentry.IHub
+import io.sentry.android.navigation.SentryNavigationListener
+
+internal class SentryLifecycleObserver(
+    private val navController: NavController,
+    private val hub: IHub = HubAdapter.getInstance(),
+    private val navListener: NavController.OnDestinationChangedListener =
+        SentryNavigationListener(hub)
```
marandaneto, line 20: Optional ctors as well or the overload annotation.

buildSrc/src/main/java/Config.kt
romtsn: good call, yeah they do

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryNavigationIntegration.kt
romtsn: mm I don't think it's needed here, cause this class is internal and is only supposed to be used in kotlin-only codebases, since compose is a kotlin framework

buildSrc/src/main/java/Config.kt
romtsn: Yeah it is https://developer.android.com/studio/releases/platforms#12

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryNavigationIntegration.kt
romtsn: Compose has a concept of [side-effects](https://developer.android.com/jetpack/compose/side-effects) and we are using one of them here to attach the navigation observer (DisposableEffect).

sentry-compose/build.gradle.kts
romtsn: The problem was that it wasn't publishing correct artifacts when I tested it with mavenLocal. I think it still publishes correctly to mavenCentral though, I've seen someone is using `maven-publish` exclusively for mavenLocal, I guess I will leave a comment there

buildSrc/src/main/java/Config.kt
romtsn: Oh yeah, I think we can still keep it at 1.4 actually, it's still possible with 1.6.

The language adoption is not available publicly unfortunately.

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryNavigationIntegration.kt
marandaneto: So the problem is that, if they look at optional params, they will think, do I need a `hub`? do I need a `navListener`? even with default parameters, ideally, they'd not even think about it, that's why I believe having fewer options the better, we just need everything to make it testable.

romtsn: But they are not going to use this class at all, because it's marked as internal :) it's only us who's using this class

sentry-compose/build.gradle.kts
romtsn: I've just compared the `pom` file with the okhttp one and I'd say it looks correct. I think the plugin is smart enough to pick up values like developer name, license, etc.
```
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <!-- This module was also published with a richer model, Gradle metadata,  -->
  <!-- which should be used instead. Do not delete the following line which  -->
  <!-- is to indicate to Gradle or any Gradle module metadata file consumer  -->
  <!-- that they should prefer consuming it instead. -->
  <!-- do_not_remove: published-with-gradle-metadata -->
  <modelVersion>4.0.0</modelVersion>
  <groupId>io.sentry</groupId>
  <artifactId>sentry-compose-android</artifactId>
  <version>6.1.1</version>
  <packaging>aar</packaging>
  <name>Sentry SDK</name>
  <description>SDK for sentry.io</description>
  <url>https://github.com/getsentry/sentry-java</url>
  <licenses>
    <license>
      <name>MIT</name>
      <url>http://www.opensource.org/licenses/mit-license.php</url>
    </license>
  </licenses>
  <developers>
    <developer>
      <id>getsentry</id>
      <name>Sentry Team and Contributors</name>
      <url>https://github.com/getsentry/</url>
    </developer>
  </developers>
  <scm>
    <connection>scm:git:git://github.com/getsentry/sentry-java.git</connection>
    <developerConnection>scm:git:ssh://git@github.com/getsentry/sentry-java.git</developerConnection>
    <url>https://github.com/getsentry/sentry-java</url>
  </scm>
  <dependencies>
    <dependency>
      <groupId>androidx.compose.runtime</groupId>
      <artifactId>runtime</artifactId>
      <version>1.1.0</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>androidx.compose.ui</groupId>
      <artifactId>ui</artifactId>
      <version>1.1.0</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>io.sentry</groupId>
      <artifactId>sentry</artifactId>
      <version>6.1.1</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>io.sentry</groupId>
      <artifactId>sentry-android-navigation</artifactId>
      <version>6.1.1</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>androidx.navigation</groupId>
      <artifactId>navigation-compose</artifactId>
      <version>2.4.2</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>org.jetbrains.kotlin</groupId>
      <artifactId>kotlin-stdlib</artifactId>
      <version>1.6.10</version>
      <scope>runtime</scope>
      <exclusions>
        <exclusion>
          <artifactId>kotlin-stdlib-common</artifactId>
          <groupId>org.jetbrains.kotlin</groupId>
        </exclusion>
      </exclusions>
    </dependency>
    <dependency>
      <groupId>androidx.lifecycle</groupId>
      <artifactId>lifecycle-common-java8</artifactId>
      <version>2.2.0</version>
      <scope>runtime</scope>
      <exclusions>
        <exclusion>
          <artifactId>kotlin-stdlib-common</artifactId>
          <groupId>org.jetbrains.kotlin</groupId>
        </exclusion>
      </exclusions>
    </dependency>
  </dependencies>
</project>

```

marandaneto: Cool, `gradle-maven-publish-plugin` is doing its work as well then.

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryNavigationIntegration.kt
marandaneto: Gotcha,  Since it's an extension method of `NavHostController`, I'd still add the `Sentry` within its name to be really explicit.

buildSrc/src/main/java/Config.kt
romtsn: will do 👍

PR #2137, opened 2022-06-24
Feat/androidx navigation tracing

romtsn:
_#skip-changelog_

## :scroll: Description

* Adds tracing support for Jetpack Compose Navigation
* Uses idle transactions (similar to UI events)

image

Part of #2096 

## :pencil: Checklist

- [x] No breaking changes


build.gradle.kts
```diff
@@ -89,6 +90,14 @@ allprojects {
 }
 
 subprojects {
+    plugins.withId(Config.QualityPlugins.detektPlugin) {
+        configure<DetektExtension> {
+            buildUponDefaultConfig = true
+            allRules = true
+            config.setFrom("${rootProject.rootDir}/detekt.yml")
+        }
+    }
```
marandaneto, line 99: Nice DRY

detekt.yml
```diff
@@ -0,0 +1,3 @@
+style:
+  ReturnCount:
+    max: 3
```
marandaneto, line 3: What does that mean?

romtsn: This means we can write functions like this:
```
fun method() {
  if (condition1) return
  if (condition2) return
  if (condition3) return
}
```

The default number is 2 and I think it's pretty restrictive. Tbh I'm fine with any number of early returns, especially in void methods; on a private project I'd just disable this rule altogether I guess.

sentry-android-navigation/src/main/java/io/sentry/android/navigation/SentryNavigationListener.kt
```diff
@@ -61,4 +76,77 @@ class SentryNavigationListener @JvmOverloads constructor(
...
+                "Destination id cannot be retrieved from Resources, no transaction captured."
+            )
+            return
+        }
+
+        // we add '/' to the name to match dart and web pattern
+        name = "/" + name.substringBefore('/') // strip out arguments from the tx name
+
+        val transaction =
+            hub.startTransaction(name, NAVIGATION_OP, true, hub.options.idleTimeout, true)
```
adinauer, line 118: This means we're now creating a new thread on every navigation right? Does that mean we should increase the priority of https://github.com/getsentry/sentry-java/issues/2130 ?

sentry-android-navigation/src/test/java/io/sentry/android/navigation/SentryNavigationListenerTest.kt
```diff
@@ -106,14 +163,181 @@ class SentryNavigationListenerTest {
...
+    }
+
+    @Test
+    fun `onDestinationChanged starts tracing with destination id if route is not available`() {
+        val sut = fixture.getSut(toRoute = null, hasViewIdInRes = true)
+
+        sut.onDestinationChanged(fixture.navController, fixture.destination, null)
+
+        verify(fixture.hub).startTransaction(
+            check { assertEquals("/id", it) },
```
adinauer, line 279: Should we use something a little more unique instead of `id` as destination ID, so it's easier to find where this is set in the test?

sentry-android-navigation/src/test/java/io/sentry/android/navigation/SentryNavigationListenerTest.kt
```diff
@@ -106,14 +163,181 @@ class SentryNavigationListenerTest {
...
+    fun `onDestinationChanged finishes previous navigation transaction before starting a new one`() {
+        val sut = fixture.getSut()
+
+        sut.onDestinationChanged(fixture.navController, fixture.destination, null)
+        sut.onDestinationChanged(fixture.navController, fixture.destination, null)
+
+        assertEquals(true, fixture.transaction.isFinished)
+        val captor = argumentCaptor<IWithTransaction>()
+        verify(fixture.scope, times(4)).withTransaction(captor.capture())
+        // 1st time - bind to scope, 2nd time - in SentryTracer when finish, 3rd time - in the nav listener
```
adinauer, line 339: what's the 4th one?

adinauer (approved):
Mostly LGTM just some questions.

sentry-android-navigation/src/main/java/io/sentry/android/navigation/SentryNavigationListener.kt
```diff
@@ -61,4 +76,77 @@ class SentryNavigationListener @JvmOverloads constructor(
...
+        } catch (e: NotFoundException) {
+            hub.options.logger.log(
+                DEBUG,
+                "Destination id cannot be retrieved from Resources, no transaction captured."
+            )
+            return
+        }
+
+        // we add '/' to the name to match dart and web pattern
+        name = "/" + name.substringBefore('/') // strip out arguments from the tx name
```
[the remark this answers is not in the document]
marandaneto: On Flutter we have a special case if its "/" only, we call "/ (root)" I guess.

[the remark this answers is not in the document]
romtsn: nope, the route will always start with a name of the route - they strip the slash out from my tests

sentry-android-navigation/src/test/java/io/sentry/android/navigation/SentryNavigationListenerTest.kt
romtsn: It's again bind to scope, but for the 2nd transaction since we call `onDestinationChanged` 2 times. I think it's irrelevant for this test though, since we are interested only in the 3rd one (though I agree the test is a bit brittle)

sentry-android-navigation/src/main/java/io/sentry/android/navigation/SentryNavigationListener.kt
romtsn: yeah we do, but I don't think it's a decider for this issue tbh. The UI transactions is more of a problem than nav events, since you usually navigate less than interacting with screen. I think the usual usage pattern is like going through 5 screens maybe and that's it (open slack, check channel, check threads) that's it kinda :D

adinauer: If you take a mail app for example where you quickly navigate into a mail and back out, there's quite a few transactions.

sentry-android-navigation/src/test/java/io/sentry/android/navigation/SentryNavigationListenerTest.kt
romtsn: mm, I'm not sure I got the problem - what is easier to find?

sentry-android-navigation/src/main/java/io/sentry/android/navigation/SentryNavigationListener.kt
romtsn: good example, but with your recent fix it would cancel the timer after each navigation event and kill the thread, right? Regardless, I think we should do #2130 rather sooner than later, though I actually don't see a good solution for it right away (which would work both for Android and Java desktop apps)

adinauer: I'll bump prio on it and see if I can get to it sometime.

sentry-android-navigation/src/test/java/io/sentry/android/navigation/SentryNavigationListenerTest.kt
adinauer:
```suggestion
            check { assertEquals("/destination-id-1", it) },
```

Maybe like this. This isn't really important though.

romtsn: oh yeah we can do that

PR #2138, opened 2022-06-24
Add new modules to craft.yml

romtsn:
_#skip-changelog_

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -41,3 +41,5 @@ targets:
       maven:io.sentry:sentry-apollo:
       maven:io.sentry:sentry-jdbc:
       maven:io.sentry:sentry-graphql:
+      maven:io.sentry:sentry-android-navigation:
+      maven:io.sentry:sentry-compose:


bruno-garcia (approved):

.craft.yml
```diff
@@ -41,3 +41,5 @@ targets:
       maven:io.sentry:sentry-apollo:
       maven:io.sentry:sentry-jdbc:
       maven:io.sentry:sentry-graphql:
+      maven:io.sentry:sentry-android-navigation:
+      maven:io.sentry:sentry-compose:
```
marandaneto, line 45: You have to add them first (manually) to https://github.com/getsentry/sentry-release-registry
otherwise, releases will fail.

romtsn: shouldn't we first push this to maven central and then merge the release registry? or it's other way around?

marandaneto: It is the other way around, but if you merge this, you won't be able to release (push to maven central), because the release registry does not exist, craft will fail for non-preview releases.

romtsn: Alright, thanks for explaining 👍 will make sure the release registry is out first

PR #2152, opened 2022-07-04
moved average benchmark results to p90 results

stefanosiano:
## :scroll: Description

1. Moved benchmark results from average to p90 results
2. Increased the measured iterations from 15 to 20 by default
3. Profiling benchmark now prints raw values to the console, to later read them from the log file
4. Removed testOrchestrator option from saucelabs config and gradle
5. Benchmarks in SauceLabs will now run on 2 devices with Android 12, 3 with Android 11, 2 with Android 10
6. We will just assert on the cpu overhead, not on duration or dropped frames anymore, even if we are printing them
7. Added a test to send profiles to a Sentry project (dogfooding test)

1. To get better results from the benchmark, the p90 is a better metric than the average.
2. We decided to move to 20 measured iteration to get reliable results.
3. Since we want benchmark numbers to show on docs, we are going to print them to the log file. This is a quick solution to allow us to download the log, parse it, and get all the raw numbers we want to feed our systems. Later on, we will stop printing these values, and will send them to an endpoint. 
4. I realized the tests were not working on SauceLabs (even if they were suceeding) due to an issue with androidx test orchestrator. Removing it fixed the problem.
5. The devices on SauceLbas to run the benchmarks on have been updated. Now we test devices for Android 10, 11 and 12. Doing so, we cover more than 70% of currently used devices, based on https://gs.statcounter.com/os-version-market-share/android/mobile-tablet/worldwide.
6. Duration changes can get pretty unpredictable, due to system things, like thermal throttling, power saving, etc. Cpu overhead is, instead, pretty consistent across all devices, and that's the most important thing at the moment.
7. The backend needs some instrumented profiles 

#skip-changelog

I ran the saucelabs commands manually, and the tests ran perfectly fine. 🥳 

## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next steps
We will send the benchmark data to a specific endpoint, but this will be done in another pr later.
I just need to check the log files and if we can get all the values just fine.


.sauce/sentry-uitest-android-benchmark.yml
```diff
@@ -21,27 +21,25 @@ suites:
...
-#    testOptions:
-#      useTestOrchestrator: true
+  - name: "Android 10 (api 29)"
+    devices:
+      - id: Google_Pixel_4_XL_real_us1 # Google Pixel 4 XL - api 29 (10)
+      - id: Motorola_Moto_G_Power_real_us # Motorola Moto G Power (2021) - api 29 (10)
 
-#      - id: Google_Pixel_4_XL_real_us1 # Google Pixel 4 XL - api 29 (10)
-#      - id: Motorola_Moto_G_Power_real_us # Motorola Moto G Power (2021) - api 29 (10)
+# At the time of writing, the market share per android version is:
```
philipphofmann, line 38: `l`: You could add the date, so you don't have to check git blame.

sentry-android-integration-tests/sentry-uitest-android-benchmark/build.gradle.kts
```diff
@@ -26,6 +26,7 @@ android {
         // https://developer.android.com/training/testing/instrumented-tests/androidx-test-libraries/runner#enable-gradle
         // This doesn't work on some devices with Android 11+. Clearing package data resets permissions.
         // Check the readme for more info.
+        // Test orchestrator was removed due to issues with SauceLabs
```
philipphofmann, line 29: Important comment thanks.

philipphofmann (commented):
Gave this a quick pass. What's missing to finish this PR, @stefanosiano?

stefanosiano:
> Gave this a quick pass. What's missing to finish this PR, @stefanosiano?

Just a few changes on how to calculate the p90 values, as benchmarks are failing because of it 😅 
And a change for refresh rate to avoid crashes on Android 11/12 that i didn't realize before 😅

stefanosiano:
The only thing is that benchmarks take around 1 hour to run. 
Are we fine with that?

philipphofmann:
> The only thing is that benchmarks take around 1 hour to run. Are we fine with that?

Faster would be better, of course. What's the bottleneck? How could we reduce this?

sentry-android-integration-tests/sentry-uitest-android/src/androidTest/java/io/sentry/uitest/android/EnvelopeTests.kt
```diff
@@ -56,4 +62,31 @@ class EnvelopeTests : BaseUiTest() {
...
+            options.tracesSampleRate = 1.0
+            options.isProfilingEnabled = true
+        }
+
+        val transaction = Sentry.startTransaction("e2etests", "testProfile")
+        val benchmarkScenario = launchActivity<ProfilingSampleActivity>()
+        swipeList(10)
+        benchmarkScenario.moveToState(Lifecycle.State.DESTROYED)
+        transaction.finish()
+        Thread.sleep(100)
```
philipphofmann, line 81: `l`: Why do we need these `sleep`? Can you add the info as a comment, please?

sentry-android-integration-tests/sentry-uitest-android/src/main/java/io/sentry/uitest/android/ProfilingSampleActivity.kt
```diff
@@ -0,0 +1,109 @@
...
+    companion object {
+
+        /** The activity will set this when scrolling. */
+        val scrollingIdlingResource = CountingIdlingResource("sentry-uitest-android-profilingSampleActivityScrolling")
+    }
+
+    private lateinit var binding: ActivityProfilingSampleBinding
+    private val backgroundThreadPoolSize = 2
+    private val executor: ExecutorService = Executors.newFixedThreadPool(backgroundThreadPoolSize)
+    private var resumed = false
```
philipphofmann, line 30: `l`: Why do we need that property?

philipphofmann (approved):
LGTM, but ideally, we reduce the 1h duration.

sentry-android-integration-tests/sentry-uitest-android/src/main/java/io/sentry/uitest/android/ProfilingSampleActivity.kt
stefanosiano: it was a flag to stop the function after killing the activity, and i forgot to put it in fibonacci. Not a real issue, as the test is ran only once and the process is killed when the test ends. But i'm going to add it anyway

stefanosiano:
> LGTM, but ideally, we reduce the 1h duration.

There is [this pr](https://github.com/getsentry/sentry-java/pull/2166), with a slack thread on that and some consideration on [this notion doc](https://www.notion.so/sentry/Benchmarking-74d150a7eb7e46c4bff9901037cc4e1d)

sentry-android-integration-tests/sentry-uitest-android/src/androidTest/java/io/sentry/uitest/android/EnvelopeTests.kt
stefanosiano: useless, actually. i removed it

.sauce/sentry-uitest-android-benchmark.yml
```diff
@@ -21,27 +21,25 @@ suites:
...
 
   - name: "Android 11 (api 30)"
     devices:
-      - id: OnePlus_9_Pro_real_us # OnePlus 9 Pro - api 30 (11)
-      - id: Google_Pixel_2_real_us # Google Pixel 2 - api 30 (11)
-    testOptions:
-      useTestOrchestrator: true
+      - id: OnePlus_9_Pro_real_us # OnePlus 9 Pro - api 30 (11) - high end
+      - id: Google_Pixel_4_real_us # Google Pixel 4 - api 30 (11) - mid end
+      - id: Google_Pixel_2_real_us # Google Pixel 2 - api 30 (11) - low end
```
romtsn, line 31: I'm not sure if Pixel 2 is really a low-end device, but that's alright I guess

stefanosiano: It's one of the lowest-end provided by saucelabs for android 11. The other at ~same level should be Samsung Galaxy A50, but the pixel 2 is a 5 years old device, so it's fine for benchmarks

sentry-android-integration-tests/sentry-uitest-android-benchmark/src/androidTest/java/io/sentry/uitest/android/benchmark/util/BenchmarkComparisonResult.kt
```diff
@@ -0,0 +1,153 @@
...
+) {
+
+    /**
+     * Prints the raw results of all runs of the comparison.
+     * Each printed line is prefixed by [prefix], to allow parsers to easily parse log files to read raw values.
+     */
+    fun printAllRuns(prefix: String) {
+        repeat(iterations) { index ->
+
+            println("$prefix ==================== Iteration $index ====================")
```
romtsn, line 58: Just for me to understand - this is gonna be printed into Logcat and later you pull this through adb probably, and parse, correct?

PR #2166, opened 2022-07-12
Reduced benchmarks

stefanosiano:
## :scroll: Description
Reduced benchmarks to only run on android/java packages (not on integrations)
On every pull request the benchmarks will be run on one device only.
The full benchmark device suite will be tested nightly.

#skip-changelog

The benchmarks currently take ~1 hour to run on saucelabs, more if a device times out. The situation is going to get worse as more tests are added.
To alleviate the situation we run the tests only on one device, so that this workflow should take ~10-15 minutes (plus build time). Also, running benchmarks on just one device reduces the risk of timeouts from saucelabs.

## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next steps
We will monitor the situation and will make further changes in the future if needed, but at the moment these changes should be enough


.github/workflows/integration-tests-benchmarks.yml
```diff
@@ -4,13 +4,26 @@ on:
...
+      - '**/.github/**'
 
 jobs:
   test:
     name: Benchmarks
     runs-on: ubuntu-latest
 
     steps:
+      - name: Check pr keyword
+        uses: JJ/github-pr-contains-action@releases/v4
```
romtsn, line 22: could you pin it to a specific commit hash, so it's reproducible in case of problems?

romtsn (approved):

.github/workflows/integration-tests-benchmarks.yml
stefanosiano: i'm just making some tests, it's not meant to be merged right now
it seems the workflow fails instead of being cancelled, so maybe it's not a viable option (or can we "force" a merge in case a check fails?)

romtsn: Can we do similar but for comments maybe? https://stackoverflow.com/a/59775665

stefanosiano: i removed the logic to skip based on the pull request description.
We will consider to add a check on the git commit message, in case benchmarks are still taking a lot to run

.github/workflows/integration-tests-benchmarks.yml
```diff
@@ -1,9 +1,18 @@
 name: "Integration Tests - Benchmarks"
 on:
+  schedule:
```
romtsn, line 3: Just thinking if it's necessary, because they will anyway run on every merge/push to `main`? If there were no changes on main and a cron job kicks in, it's still useless probably, because there were nothing new on main?

romtsn:
Just one small comment, otherwise LGTM 🚀 You already got my approval anyway

.github/workflows/integration-tests-benchmarks.yml
stefanosiano: you may be right on this
let's remove pushing to main: we may have multiple pushes to main daily, so we could end up with a lot of tests running on saucelabs, risking timeouts and slowdowns. Also, we would still run the single benchmark for every pr, so running only nightly is enough at the moment. We can still change it later

romtsn: Yeah, sounds good 👍 just need some kind of reminder on slack or somewhere to check the benchmarks from the nightly builds

romtsn: We don't need to do it now, but would be cool to either post on slack or automatically create an issue if there were regressions after the nightly build

romtsn: actually I guess it's pretty straightforward - if a test run failed, we just use some action to create an issue and that's it :D most important is that there shouldn't be flaky tests to not overwhelm us with issues haha

PR #2175, opened 2022-07-15
feat: Make upstream dependencies compileOnly in integrations

### LINKED ISSUE #2125: Switch to compileOnly dependencies for integrations
### Description

At the moment we declare all integration dependencies (e.g. okhttp for sentry-okhttp-intreceptor) as `implementation`. This means we are potentially overriding user-defined dependencies, if they use an older version of okhttp. We should switch those dependencies to be `compileOnly` so at runtime we'd actually use what user has defined in their deps. 

We should cautious about incompatible versions though, at best we could define version constraints for the min. supported version, at worst have a note for each integration/troubleshooting page on the docs

romtsn:
## :scroll: Description

Switch to compileOnly upstream dependencies in integrations, so we do not override user-defined version and use what they actually have on their classpath.

Closes #2125 

Tested manually here and on SAGP

## :pencil: Checklist

- [x] I updated the docs if needed https://github.com/getsentry/sentry-docs/pull/5298
- [ ] No breaking changes

## :crystal_ball: Next steps
@adinauer I'm not sure if you want to do this for backend things, I think it can also be applicable there? We could create another issue, for now I'm afraid of breaking something there

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
@@ -33,7 +33,7 @@ public final class SentryAndroid {
 
   private static final String TIMBER_CLASS_NAME = "timber.log.Timber";
   private static final String FRAGMENT_CLASS_NAME =
-    "androidx.fragment.app.FragmentManager$FragmentLifecycleCallbacks";
+      "androidx.fragment.app.FragmentManager$FragmentLifecycleCallbacks";
 
   private SentryAndroid() {}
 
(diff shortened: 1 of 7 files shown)


bruno-garcia (approved):
LGTM. Maybe @adinauer or @marandaneto want to 👀 ?

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
```diff
@@ -84,10 +88,17 @@ public static synchronized void init(
           OptionsContainer.create(SentryAndroidOptions.class),
           options -> {
             final LoadClass classLoader = new LoadClass();
+            final boolean isTimberUpstreamAvailable =
+                classLoader.isClassAvailable(TIMBER_CLASS_NAME, options);
+            final boolean isFragmentUpstreamAvailable =
+                classLoader.isClassAvailable(FRAGMENT_CLASS_NAME, options);
             final boolean isFragmentAvailable =
-                classLoader.isClassAvailable(SENTRY_FRAGMENT_INTEGRATION_CLASS_NAME, options);
+                (isFragmentUpstreamAvailable
```
philipphofmann, line 96: `m`: What error message does the user get if any timber or fragment is not available? Will they know from the error message how to resolve the issue?

philipphofmann (approved):
LGTM, but maybe we should add more detail to the error messages somewhere.

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
romtsn: It will say "Class not available: Fragment..." with a debug log level. That's fine, because it's part of the auto-install feature and it's not supposed to be always installed, there's no action for the user to take.

philipphofmann (approved):
LGTM 🚀

adinauer (approved):
LGTM

PR #2184, opened 2022-07-25
Add profilesSampleRate and profileSampler for Android sdk

stefanosiano:
## :scroll: Description
I added the `profilesSampleRate` and `profileSampler` options for the Android sdk.
The `profilingEnabled` options has been deprecated, and its usage changed to "setProfilesSampleRate(1.0 or null)". Also, it is overridden by the `profilesSampleRate` option.

We are reaching the profiling open beta stage. 
We want to provide users a way to reduce the number of profiles sent to Sentry, in the same way we do with transactions.
To be consistent with transactions, we will provide the `profilesSampleRate` and `profileSampler` options.

I added some unit tests 

## :pencil: Checklist

- [x] No breaking changes


sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -1500,16 +1511,65 @@ public void setTransactionProfiler(final @Nullable ITransactionProfiler transact
    * @return if profiling is enabled for transactions.
    */
   public boolean isProfilingEnabled() {
-    return profilingEnabled;
+    return (getProfilesSampleRate() != null && getProfilesSampleRate() > 0)
```
marandaneto, line 1514: We don't check `> 0` for `isTracingEnabled` at least.
Do we need it here?

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -1500,16 +1511,65 @@ public void setTransactionProfiler(final @Nullable ITransactionProfiler transact
...
+        || getProfilesSampler() != null;
   }
 
   /**
    * Sets whether profiling is enabled for transactions.
    *
+   * @deprecated use {{@link SentryOptions#setProfilesSampleRate(Double)} }
    * @param profilingEnabled - whether profiling is enabled for transactions
    */
+  @Deprecated
```
marandaneto, line 1524: I'd add a note in the changelog about this deprecation.

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -295,7 +295,11 @@ public void finish(@Nullable SpanStatus status) {
       if (isSampled == null) {
         isSampled = false;
       }
-      if (hub.getOptions().isProfilingEnabled() && isSampled) {
+      Boolean isProfileSampled = isProfileSampled();
+      if (isProfileSampled == null) {
+        isProfileSampled = false;
+      }
```
marandaneto, line 301: This and `isSampled` can likely be simplified with `Boolean.TRUE.equals(value)`

sentry/src/main/java/io/sentry/TracesSamplingDecision.java
```diff
@@ -9,14 +9,27 @@ public final class TracesSamplingDecision {
 
   private final @NotNull Boolean sampled;
   private final @Nullable Double sampleRate;
+  private final @NotNull Boolean profileSampled;
+  private final @Nullable Double profileSampleRate;
+
+  public TracesSamplingDecision(@NotNull Boolean sampled) {
```
marandaneto, line 15:
```suggestion
  public TracesSamplingDecision(final @NotNull Boolean sampled) {
```

Please add all the missing `final` to this PR, and missing `@Nullable` and @NotNull`.

sentry/src/main/java/io/sentry/util/SampleRateUtil.java
```diff
@@ -15,7 +15,7 @@ public static boolean isValidSampleRate(@Nullable Double sampleRate, boolean all
       return allowNull;
     }
 
-    return !(sampleRate.isNaN() || (sampleRate > 1.0 || sampleRate <= 0.0));
+    return !(sampleRate.isNaN() || sampleRate > 1.0 || sampleRate <= 0.0);
```
marandaneto, line 18: Apparently a new private method that takes a Double and do this check would avoid code duplication, See this method and `isValidTracesSampleRate`.

sentry/src/main/java/io/sentry/util/SampleRateUtil.java
```diff
@@ -28,6 +28,19 @@ public static boolean isValidTracesSampleRate(
...
-    return !(tracesSampleRate.isNaN() || (tracesSampleRate > 1.0 || tracesSampleRate < 0.0));
+    return !(tracesSampleRate.isNaN() || tracesSampleRate > 1.0 || tracesSampleRate < 0.0);
+  }
+
+  public static boolean isValidProfilesSampleRate(@Nullable Double profilesSampleRate) {
+    return isValidProfilesSampleRate(profilesSampleRate, true);
+  }
+
+  public static boolean isValidProfilesSampleRate(
+      @Nullable Double profilesSampleRate, boolean allowNull) {
```
marandaneto, line 39: Why do we need `allowNull`?

sentry/src/test/java/io/sentry/HubTest.kt
```diff
@@ -1213,19 +1213,24 @@ class HubTest {
     @Test
     fun `when startTransaction and profiling is enabled, transaction is profiled only if sampled`() {
         val mockTransactionProfiler = mock<ITransactionProfiler>()
+        val unsampledHub = generateHub {
```
marandaneto, line 1216: Why did we rename this?

sentry/src/test/java/io/sentry/TracesSamplerTest.kt
```diff
@@ -136,23 +243,55 @@ class TracesSamplerTest {
...
         val samplingDecision =
             sampler.sample(SamplingContext(transactionContextNotSampled, CustomSamplingContext()))
         assertFalse(samplingDecision.sampled)
         assertNull(samplingDecision.sampleRate)
+        assertFalse(samplingDecision.profileSampled)
+        assertNull(samplingDecision.profileSampleRate)
 
         val transactionContextSampled = TransactionContext("name", "op")
-        transactionContextSampled.sampled = true
+        transactionContextSampled.setSampled(true, true)
```
marandaneto, line 280: Isn't that a breaking change in the public API?

sentry/src/main/java/io/sentry/SentryOptions.java
stefanosiano: well, if the sample rate is 0 and there's no sampler set, we'd still initialize the profile, even if in reality it will never be used

marandaneto: Yeah for profiling maybe it's better to keep with this check.
For transactions, we actually need it, because 0 is enabled and it propagates transactions that come from the backend and inherits the sampling decision from the parent.
If sampling should be the case, then removing the > 0 check makes sense.

sentry/src/main/java/io/sentry/util/SampleRateUtil.java
stefanosiano: may i rename `isValidTracesSampleRate` to `isValidTracesOrProfilesSampleRate` and use just one function, as they do basically the same thing?

sentry/src/main/java/io/sentry/SentryOptions.java
stefanosiano: the logic of profiles sampling mimics the transaction sampling, including the parent decision. But we don't get profiles from the backend. Should we consider the case for consistency? Initializing the profiler is quick anyway, and the option already says that `null` should be used to disable it

sentry/src/main/java/io/sentry/util/SampleRateUtil.java
stefanosiano: was for consistency, but it actually doesn't make any sense.
removing it

sentry/src/test/java/io/sentry/TracesSamplerTest.kt
stefanosiano: not really, as profile sampled defaults to false so 
```
        transactionContextNotSampled.sampled = false
        transactionContextNotSampled.setSampled(false, false) 
```
are the same, like
```
        transactionContextSampled.sampled = true
        transactionContextSampled.setSampled(true, false)
```
the `setSampled(..., true)` is the real change, but it's an addition, rather than a change. 
I can change `setSampled(false, false)` to `sampled = false` in the test for clarity

sentry/src/main/java/io/sentry/SentryOptions.java
marandaneto: That's a business decision, so I can't tell you what makes more sense for profiling.

sentry/src/main/java/io/sentry/util/SampleRateUtil.java
marandaneto: `isValidSampleRate` should be enough IMO, as long as both functions do the same.

sentry/src/test/java/io/sentry/TracesSamplerTest.kt
marandaneto: If `setSampled` is a new API, its ok.
I thought that we replaced a public API to another.

sentry/src/main/java/io/sentry/util/SampleRateUtil.java
stefanosiano: yeah, but `isValidSampleRate` already exists and does the same check, expect the rate cannot be 0.0

sentry/src/test/java/io/sentry/HubTest.kt
stefanosiano: i wanted to let the hub make decisions based on the options, but it would make more sense to simply change the `TracesSamplingDecision()` call

sentry/src/main/java/io/sentry/util/SampleRateUtil.java
marandaneto: Then your call.

PR #2192, opened 2022-08-01
Add profiling activity to Android sample app

stefanosiano:
## :scroll: Description
Few additions to the Android sample app:

- Added profiling activity with size printed on screen
- Enabled profiling
- auto user interaction instrumentation should be disabled when testing the manual profile (creates concurrent transactions not currently supported by profiling)

There was no clear way to answer the "How big is a profile?" question.
Now we can have a way to know, depending on length and number of working threads, how big is the profiling trace file, how big is the created envelope item, and how big is the actual data sent to Sentry (the item compressed with gzip).
Also, we can now feed different profiles by running the sample app and editing the dsn.
#skip-changelog

No way 😄 

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/sentry-samples/sentry-samples-android/src/main/AndroidManifest.xml
+++ b/sentry-samples/sentry-samples-android/src/main/AndroidManifest.xml
@@ -54,6 +54,10 @@
             android:exported="false" />
 
       <activity
+            android:name=".ProfilingActivity"
+            android:exported="false" />
+
+      <activity
             android:name=".compose.ComposeActivity"
             android:exported="false" />
 
@@ -100,7 +104,7 @@
         <meta-data android:name="io.sentry.traces.sample-rate" android:value="1.0" />
 
         <!--    how to enable profiling when starting transactions -->
-        <!--    <meta-data android:name="io.sentry.traces.profiling.sample-rate" android:value="1.0" />-->
+        <meta-data android:name="io.sentry.traces.profiling.sample-rate" android:value="1.0" />
 
         <!--    how to disable the Activity auto instrumentation for tracing-->
         <!--    <meta-data android:name="io.sentry.traces.activity.enable" android:value="false" />-->
@@ -109,7 +113,7 @@
         <meta-data android:name="io.sentry.traces.activity.auto-finish.enable" android:value="false" />
 
         <!-- how to enable the UI auto instrumentation for tracing -->
-        <meta-data android:name="io.sentry.traces.user-interaction.enable" android:value="true" />
+<!--        <meta-data android:name="io.sentry.traces.user-interaction.enable" android:value="true" />-->
 
         <!--    how to enable and set a sampleRate (anything between 0.01 and 1.0), it's disabled by default-->
         <!--    <meta-data android:name="io.sentry.sample-rate" android:value="0.5" />-->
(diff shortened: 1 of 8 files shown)


sentry-samples/sentry-samples-android/src/main/AndroidManifest.xml
```diff
@@ -109,7 +113,7 @@
         <meta-data android:name="io.sentry.traces.activity.auto-finish.enable" android:value="false" />
 
         <!-- how to enable the UI auto instrumentation for tracing -->
-        <meta-data android:name="io.sentry.traces.user-interaction.enable" android:value="true" />
+<!--        <meta-data android:name="io.sentry.traces.user-interaction.enable" android:value="true" />-->
```
marandaneto, line 116: I understand the problem, but disabling another feature isn't the best solution.
Maybe we create a new sample app for profiling as suggested before?

marandaneto, line 116: Can you elaborate on:

> Disabled auto user interaction instrumentation (creates concurrent transactions not currently supported by profiling)

I don't follow, there should be a single transaction in the scope, this is like this by design, so I'm not sure what you mean by concurrent transactions.
Also, disabling another feature isn't the best idea, either when testing this feature, we manually disable and enable it again (not coming the changes).
We can also create a new flavor that enables and disables certain features.
We can also create a new sample app for profiling as suggested before as well if needed.

stefanosiano: The thing is that when i added an activity where i click on a button and it starts a transaction to be profiled.
But when i click on the button an automatic transaction is started first, so that one is profiled. When the manual transaction is started, the automatic one hasn't finished, yet, so the profiler ignores the manual transaction, since it doesn't have the notion of scope.
We're currently working on allowing profiling to work in these cases.
We can reenable the user-interaction option, disabling it when testing the profiler.
I'm not sure a new sample app just for profiling is worth it, as we can find some edge cases we didn't consider when using other options of the sdk, too.

bruno-garcia: Is it possible this uncovered a scenario we need to address? Should the old transaction be completed (and so profiling ended) when the new starts? Or should the new transaction not start because we already have one on going. We only create ui event transactions if none exists?

stefanosiano: We create it only if none exists, but by default it's automatically stopped after 3 seconds. This case is a transaction manually started in those 3 seconds.
we discussed this scenario, and we decided to make profiling work with multiple transactions [here](https://www.notion.so/sentry/Profiling-concurrent-transactions-37da1625332e40799dc0e0bae90ced63).
We should support it anyway, as this has less attrition for the user, and it's more important in the backend clients, where it's easy to have multiple concurrent transactions, possibly spawn over multiple threads.

bruno-garcia: I'd say let's move ahead with this then and we can track this on an issue? Would you please open an issue?

bruno-garcia (approved):

sentry-samples/sentry-samples-android/src/main/AndroidManifest.xml
stefanosiano: yeah, we already created a [doc](https://www.notion.so/sentry/Profiling-concurrent-transactions-37da1625332e40799dc0e0bae90ced63)
Anyway, i just created [this issue](https://github.com/getsentry/sentry-java/issues/2209) and i'll track it in the new pr i'll do

PR #2195, opened 2022-08-02
fix: update-gradle.sh

vaind:
Ensures updater workflow doesn't fail because `./gradlew --version` reports a different version than the tag on github.com/gradle/gradle, when the tag ends with a zero.

#skip-changelog

### DIFF
--- a/scripts/update-gradle.sh
+++ b/scripts/update-gradle.sh
@@ -7,8 +7,14 @@ case $1 in
 get-version)
     # `./gradlew` shows some info on the first run, breaking the parsing in the next step.
     # Therefore, we run it once without checking any output.
-    ./gradlew --version > /dev/null
+    ./gradlew --version >/dev/null
     version="$(./gradlew --version | sed -E -n 's/.*Gradle +([0-9.]+).*/\1/p')"
+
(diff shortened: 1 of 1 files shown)


adinauer (approved):

scripts/update-gradle.sh
```diff
@@ -7,8 +7,14 @@ case $1 in
...
     # `./gradlew` shows some info on the first run, breaking the parsing in the next step.
     # Therefore, we run it once without checking any output.
-    ./gradlew --version > /dev/null
+    ./gradlew --version >/dev/null
     version="$(./gradlew --version | sed -E -n 's/.*Gradle +([0-9.]+).*/\1/p')"
+
+    # Add trailing ".0" - gradlew outputs '7.1' instead of '7.1.0'
+    if [[ "$version" =~ ^[0-9]\.[0-9]$ ]]; then
+        version="$version.0"
+    fi
```
marandaneto, line 16: Looking at https://github.com/gradle/gradle/releases/tag/v7.5.0
The tag is `7.5.0` but the command to update is:
`./gradlew wrapper --gradle-version=7.5`
Wouldn't that break our auto-updater?

vaind: That's already trimmed below in set-version

PR #2228, opened 2022-08-30
fix: AbstractMethodError when getting Lifecycle

bitsandfoxes:
## :scroll: Description
This came up through Unity. In some cases, Unity projects resolve their own external dependencies via adding and including `arr` files. This seems to lead to issues during runtime when the AppLifecycleIntegration tries to add its observer and results in an `AbstractMethodError`.

To deal with the `AbstractMethodError` gracefully, we can catch the error instead and bail before we create the watcher.

```
2022/08/22 15:26:47.122 30530 30530 Error AndroidRuntime FATAL EXCEPTION: main
2022/08/22 15:26:47.122 30530 30530 Error AndroidRuntime Process: studio.banditos.banditos5, PID: 30530
2022/08/22 15:26:47.122 30530 30530 Error AndroidRuntime java.lang.AbstractMethodError: abstract method "void androidx.lifecycle.DefaultLifecycleObserver.onCreate(androidx.lifecycle.LifecycleOwner)"
2022/08/22 15:26:47.122 30530 30530 Error AndroidRuntime 	at androidx.lifecycle.FullLifecycleObserverAdapter.onStateChanged(FullLifecycleObserverAdapter.java:36)
2022/08/22 15:26:47.122 30530 30530 Error AndroidRuntime 	at androidx.lifecycle.LifecycleRegistry$ObserverWithState.dispatchEvent(LifecycleRegistry.java:354)
2022/08/22 15:26:47.122 30530 30530 Error AndroidRuntime 	at androidx.lifecycle.LifecycleRegistry.addObserver(LifecycleRegistry.java:196)
2022/08/22 15:26:47.122 30530 30530 Error AndroidRuntime 	at io.sentry.android.core.AppLifecycleIntegration.addObserver(AppLifecycleIntegration.java:92)
2022/08/22 15:26:47.122 30530 30530 Error AndroidRuntime 	at io.sentry.android.core.AppLifecycleIntegration.register(AppLifecycleIntegration.java:60)
2022/08/22 15:26:47.122 30530 30530 Error AndroidRuntime 	at io.sentry.Sentry.init(Sentry.java:189)
2022/08/22 15:26:47.122 30530 30530 Error AndroidRuntime 	at io.sentry.Sentry.init(Sentry.java:118)
2022/08/22 15:26:47.122 30530 30530 Error AndroidRuntime 	at io.sentry.android.core.SentryAndroid.init(SentryAndroid.java:87)
2022/08/22 15:26:47.122 30530 30530 Error AndroidRuntime 	at io.sentry.android.core.SentryAndroid.init(SentryAndroid.java:56)
2022/08/22 15:26:47.122 30530 30530 Error AndroidRuntime 	at io.sentry.android.core.SentryInitProvider.onCreate(SentryInitProvider.java:27)
2022/08/22 15:26:47.122 30530 30530 Error AndroidRuntime 	at android.content.ContentProvider.attachInfo(ContentProvider.java:2429)
2022/08/22 15:26:47.122 30530 30530 Error AndroidRuntime 	at android.content.ContentProvider.attachInfo(ContentProvider.java:2399)
2022/08/22 15:26:47.122 30530 30530 Error AndroidRuntime 	at io.sentry.android.core.SentryInitProvider.attachInfo(SentryInitProvider.java:44)
2022/08/22 15:26:47.122 30530 30530 Error AndroidRuntime 	at android.app.ActivityThread.installProvider(ActivityThread.java:8210)
2022/08/22 15:26:47.122 30530 30530 Error AndroidRuntime 	at android.app.ActivityThread.installContentProviders(ActivityThread.java:7746)
2022/08/22 15:26:47.122 30530 30530 Error AndroidRuntime 	at android.app.ActivityThread.handleBindApplication(ActivityThread.java:7566)
2022/08/22 15:26:47.122 30530 30530 Error AndroidRuntime 	at android.app.ActivityThread.access$1500(ActivityThread.java:301)
2022/08/22 15:26:47.122 30530 30530 Error AndroidRuntime 	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2177)
2022/08/22 15:26:47.122 30530 30530 Error AndroidRuntime 	at android.os.Handler.dispatchMessage(Handler.java:106)
2022/08/22 15:26:47.122 30530 30530 Error AndroidRuntime 	at android.os.Looper.loop(Looper.java:246)
2022/08/22 15:26:47.122 30530 30530 Error AndroidRuntime 	at android.app.ActivityThread.main(ActivityThread.java:8653)
2022/08/22 15:26:47.122 30530 30530 Error AndroidRuntime 	at java.lang.reflect.Method.invoke(Native Method)
2022/08/22 15:26:47.122 30530 30530 Error AndroidRuntime 	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:602)
2022/08/22 15:26:47.122 30530 30530 Error AndroidRuntime 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1130)
```

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AppLifecycleIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AppLifecycleIntegration.java
@@ -1,5 +1,6 @@
 package io.sentry.android.core;
 
+import androidx.lifecycle.Lifecycle;
 import androidx.lifecycle.ProcessLifecycleOwner;
 import io.sentry.IHub;
 import io.sentry.Integration;
@@ -56,6 +57,7 @@ public final class AppLifecycleIntegration implements Integration, Closeable {
       try {
         Class.forName("androidx.lifecycle.DefaultLifecycleObserver");
         Class.forName("androidx.lifecycle.ProcessLifecycleOwner");
+
         if (MainThreadChecker.isMainThread()) {
           addObserver(hub);
         } else {
@@ -83,13 +85,26 @@ public final class AppLifecycleIntegration implements Integration, Closeable {
     if (this.options == null) {
       return;
     }
+
+    Lifecycle lifeCycle;
+
+    try
+    {
+      lifeCycle = ProcessLifecycleOwner.get().getLifecycle();
+    } catch (AbstractMethodError e) {
+      options
+        .getLogger()
+        .log(SentryLevel.ERROR, "AppLifecycleIntegration failed to get Lifecycle.", e);
+      return;
+    }
+
     watcher =
-        new LifecycleWatcher(
-            hub,
-            this.options.getSessionTrackingIntervalMillis(),
-            this.options.isEnableAutoSessionTracking(),
-            this.options.isEnableAppLifecycleBreadcrumbs());
-    ProcessLifecycleOwner.get().getLifecycle().addObserver(watcher);
+      new LifecycleWatcher(
+        hub,
+        this.options.getSessionTrackingIntervalMillis(),
+        this.options.isEnableAutoSessionTracking(),
+        this.options.isEnableAppLifecycleBreadcrumbs());
+    lifeCycle.addObserver(watcher);
     options.getLogger().log(SentryLevel.DEBUG, "AppLifecycleIntegration installed.");
   }
 
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AppLifecycleIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AppLifecycleIntegration.java
@@ -85,25 +85,28 @@ public final class AppLifecycleIntegration implements Integration, Closeable {
       return;
     }
 
-    Lifecycle lifeCycle;
+    watcher =
+      new LifecycleWatcher(
+        hub,
+        this.options.getSessionTrackingIntervalMillis(),
+        this.options.isEnableAutoSessionTracking(),
+        this.options.isEnableAppLifecycleBreadcrumbs());
 
     try {
-      lifeCycle = ProcessLifecycleOwner.get().getLifecycle();
-    } catch (AbstractMethodError e) {
+      ProcessLifecycleOwner.get().getLifecycle().addObserver(watcher);
+      options.getLogger().log(SentryLevel.DEBUG, "AppLifecycleIntegration installed.");
+    } catch (Throwable e) {
+      // This is to handle a potential 'AbstractMethodError' gracefully. The error is triggered in
+      // connection with conflicting dependencies of the androidx.lifecycle.
+      // //See the issue here: https://github.com/getsentry/sentry-java/pull/2228
+      watcher = null;
       options
           .getLogger()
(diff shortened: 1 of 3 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/AppLifecycleIntegration.java
```diff
@@ -83,13 +84,25 @@ private void addObserver(final @NotNull IHub hub) {
...
+    }
+
     watcher =
         new LifecycleWatcher(
             hub,
             this.options.getSessionTrackingIntervalMillis(),
             this.options.isEnableAutoSessionTracking(),
             this.options.isEnableAppLifecycleBreadcrumbs());
-    ProcessLifecycleOwner.get().getLifecycle().addObserver(watcher);
+    lifeCycle.addObserver(watcher);
```
marandaneto, line 105: Actually based on the stack trace, `addObserver` is throwing and not the `ProcessLifecycleOwner.get()`
I'd rather wrap the whole L92 within a try/catch, catching a `Throwable` even and not a specific  `AbstractMethodError`, and add a comment about the `AbstractMethodError` problem.

marandaneto:
@bitsandfoxes revert the sentry-native changes.

sentry-android-core/src/main/java/io/sentry/android/core/AppLifecycleIntegration.java
bitsandfoxes: Can I just set the `watcher = null` to avoid removing it during close?

bruno-garcia: It threw where he got the guard unless I'm missing something: addObserver(AppLifecycleIntegration.java:92
Original line 92

bruno-garcia (commented):
Please rever the submodule changes

philipphofmann:
@bitsandfoxes, I think we discussed yesterday in a call that we don't need this change because the root cause is some weird transient dependency handling of Unity. If the same error happens on Android, we want the SDK to crash instead of failing silently to let the user know something is wrong with the dependencies. Furthermore, we concluded that this error is an edge case on Android. Should we close this?

marandaneto:
> @bitsandfoxes, I think we discussed yesterday in a call that we don't need this change because the root cause is some weird transient dependency handling of Unity. If the same error happens on Android, we want the SDK to crash instead of failing silently to let the user know something is wrong with the dependencies. Furthermore, we concluded that this error is an edge case on Android. Should we close this?

I think this quick workaround makes sense so that the SDK won't install `AppLifecycleIntegration` but everything else should work, it's the 2nd time this issue comes up, so apparently, it was not a one-off.

marandaneto (approved):

philipphofmann:
> I think this quick workaround makes sense so that the SDK won't install AppLifecycleIntegration but everything else should work, it's the 2nd time this issue comes up, so apparently, it was not a one-off.

Fine with me as long as we don't silently fail. Logging the error should be enough.

PR #2235, opened 2022-09-07
add cpu start/end timestamps to concurrent profiling transactions

stefanosiano:
## :scroll: Description
add cpu start/end timestamps to transactions in profile envelope payloads
added ui test to check times are different

#skip-changelog

Profiling traces have a clock field to indicate how timestamps were captured in the trace itself and it can be dual/cpu/wall/global time.
If the timestamps are cpu time, there's no way to filter the trace based on start/end timestamps, as the payload only sends wall time.

I added a ui test to check correctness of cpu timestamps

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
@@ -9,6 +9,7 @@ import android.content.Context;
 import android.content.pm.PackageInfo;
 import android.os.Build;
 import android.os.Debug;
+import android.os.Process;
 import android.os.SystemClock;
 import io.sentry.ITransaction;
 import io.sentry.ITransactionProfiler;
@@ -54,6 +55,7 @@ final class AndroidTransactionProfiler implements ITransactionProfiler {
   private final @NotNull BuildInfoProvider buildInfoProvider;
   private final @Nullable PackageInfo packageInfo;
   private long transactionStartNanos = 0;
+  private long profileStartCpuMillis = 0;
   private boolean isInitialized = false;
   private int transactionsCounter = 0;
   private final @NotNull Map<String, ProfilingTransactionData> transactionMap = new HashMap<>();
@@ -141,15 +143,17 @@ final class AndroidTransactionProfiler implements ITransactionProfiler {
                   PROFILING_TIMEOUT_MILLIS);
 
       transactionStartNanos = SystemClock.elapsedRealtimeNanos();
+      profileStartCpuMillis = Process.getElapsedCpuTime();
 
       ProfilingTransactionData transactionData =
-          new ProfilingTransactionData(transaction, transactionStartNanos);
+          new ProfilingTransactionData(transaction, transactionStartNanos, profileStartCpuMillis);
       transactionMap.put(transaction.getEventId().toString(), transactionData);
 
       Debug.startMethodTracingSampling(traceFile.getPath(), BUFFER_SIZE_BYTES, intervalUs);
     } else {
       ProfilingTransactionData transactionData =
-          new ProfilingTransactionData(transaction, SystemClock.elapsedRealtimeNanos());
+          new ProfilingTransactionData(
+              transaction, SystemClock.elapsedRealtimeNanos(), Process.getElapsedCpuTime());
       transactionMap.put(transaction.getEventId().toString(), transactionData);
     }
     options
@@ -230,13 +234,18 @@ final class AndroidTransactionProfiler implements ITransactionProfiler {
       ProfilingTransactionData transactionData =
           transactionMap.get(transaction.getEventId().toString());
       if (transactionData != null) {
-        transactionData.notifyFinish(SystemClock.elapsedRealtimeNanos(), transactionStartNanos);
+        transactionData.notifyFinish(
+            SystemClock.elapsedRealtimeNanos(),
+            transactionStartNanos,
+            Process.getElapsedCpuTime(),
+            profileStartCpuMillis);
       }
       return null;
     }
 
     Debug.stopMethodTracing();
     long transactionEndNanos = SystemClock.elapsedRealtimeNanos();
+    long transactionEndCpuMillis = Process.getElapsedCpuTime();
     long transactionDurationNanos = transactionEndNanos - transactionStartNanos;
 
     List<ProfilingTransactionData> transactionList = new ArrayList<>(transactionMap.values());
@@ -270,7 +279,11 @@ final class AndroidTransactionProfiler implements ITransactionProfiler {
     // We notify all transactions data that all transactions finished.
     // Some may not have been really finished, in case of a timeout
     for (ProfilingTransactionData t : transactionList) {
-      t.notifyFinish(transactionEndNanos, transactionStartNanos);
+      t.notifyFinish(
+          transactionEndNanos,
+          transactionStartNanos,
+          transactionEndCpuMillis,
+          profileStartCpuMillis);
     }
 
     // cpu max frequencies are read with a lambda because reading files is involved, so it will be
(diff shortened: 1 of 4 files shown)


sentry-android-integration-tests/sentry-uitest-android/src/androidTest/java/io/sentry/uitest/android/EnvelopeTests.kt
```diff
@@ -84,10 +84,14 @@ class EnvelopeTests : BaseUiTest() {
         relayIdlingResource.increment()
         relayIdlingResource.increment()
         val transaction = Sentry.startTransaction("e2etests", "test1")
+        Thread.sleep(5)
```
bruno-garcia, line 87: Do we rely on this pause bring 5 seconds? If it were 1, or 500ms, would it still validate what it needs?
I worry of things getting flaky so checking on the need of exact value, but also on making the test slower than it needs to be

sentry/api/sentry.api
```diff
@@ -798,16 +798,18 @@ public final class io/sentry/ProfilingTraceData$JsonKeys {
...
 	public fun getName ()Ljava/lang/String;
+	public fun getRelativeEndCpuMs ()Ljava/lang/Long;
 	public fun getRelativeEndNs ()Ljava/lang/Long;
+	public fun getRelativeStartCpuMs ()Ljava/lang/Long;
 	public fun getRelativeStartNs ()Ljava/lang/Long;
 	public fun getTraceId ()Ljava/lang/String;
 	public fun getUnknown ()Ljava/util/Map;
 	public fun hashCode ()I
-	public fun notifyFinish (Ljava/lang/Long;Ljava/lang/Long;)V
+	public fun notifyFinish (Ljava/lang/Long;Ljava/lang/Long;Ljava/lang/Long;Ljava/lang/Long;)V
```
bruno-garcia, line 812: Technically a breaking change, should we annotate this method or type saying it's internal?

sentry-android-integration-tests/sentry-uitest-android/src/androidTest/java/io/sentry/uitest/android/EnvelopeTests.kt
stefanosiano: it's 5 milliseconds, and it's added to avoid that multiple transactions have the same cpu timestamp, as they are collected with milliseconds precision

stefanosiano: I'll add a comment to make it clear

stefanosiano: Ok, i removed the sleep altogether :D

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
```diff
@@ -230,13 +234,18 @@ public synchronized void onTransactionStart(@NotNull ITransaction transaction) {
       ProfilingTransactionData transactionData =
           transactionMap.get(transaction.getEventId().toString());
       if (transactionData != null) {
-        transactionData.notifyFinish(SystemClock.elapsedRealtimeNanos(), transactionStartNanos);
+        transactionData.notifyFinish(
+            SystemClock.elapsedRealtimeNanos(),
+            transactionStartNanos,
+            Process.getElapsedCpuTime(),
```
philipphofmann, line 240: @adinauer, I think this is what we also need to measure a transaction's CPU overhead. @stefanosiano, please correct me if I'm wrong. 

[SystemClock.elapsedRealtimeNanos()](https://developer.android.com/reference/android/os/SystemClock#elapsedRealtimeNanos())
> return the time since the system was booted, and include deep sleep. This clock is guaranteed to be monotonic, and continues to tick even when the CPU is in power saving modes, so is the recommend basis for general purpose interval timing.

[Process.getElapsedCpuTime()](https://developer.android.com/reference/android/os/Process#getElapsedCpuTime())

> Returns elapsed milliseconds of the time this process has run.

So you can calculate the CPU overhead by checking the duration of the transaction with `SystemClock.elapsedRealtimeNanos()` and comparing it with the milliseconds of the process has actually run `Process.getElapsedCpuTime()`.

How expensive are the API calls, @stefanosiano?

stefanosiano: The apis should be quick, even if i don't have any number to tell you.
I'm not sure on what you mean by cpu overhead. I mean, if a transaction lasts longer, both system clock and cpu time increase.
The difference between them should be `SystemClock.elapsedRealtimeNanos()` starts since system boot, while `Process.getElapsedCpuTime()` since the process start. But during a transaction they should increase in the same way (except for deep sleep, but that's not important here now, imho).
Am i missing something?

stefanosiano:
@philipphofmann @bruno-garcia may i consider this pr reviewed and merge it?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
philipphofmann: Hmm, maybe I still don't get the actual difference between [SystemClock.elapsedRealtimeNanos()](https://developer.android.com/reference/android/os/SystemClock#elapsedRealtimeNanos()) and [Process.getElapsedCpuTime()](https://developer.android.com/reference/android/os/Process#getElapsedCpuTime()). Nevermind, @adinauer is looking into this right now and he will know.

sentry-android-integration-tests/sentry-uitest-android/src/androidTest/java/io/sentry/uitest/android/EnvelopeTests.kt
```diff
@@ -115,13 +117,19 @@ class EnvelopeTests : BaseUiTest() {
...
                 assertNotEquals(startTimes[0], startTimes[1])
                 assertNotEquals(startTimes[0], startTimes[2])
                 assertNotEquals(startTimes[1], startTimes[2])
                 assertNotEquals(endTimes[0], endTimes[1])
                 assertNotEquals(endTimes[0], endTimes[2])
                 assertNotEquals(endTimes[1], endTimes[2])
 
+                // The cpu timestamps shouldn't be all the same
+                assertFalse(startCpuTimes[0] == startCpuTimes[1] && startCpuTimes[1] == startCpuTimes[2])
+                assertFalse(endCpuTimes[0] == endCpuTimes[1] && endCpuTimes[1] == endCpuTimes[2])
```
philipphofmann, line 131: `m`: I think we could do a more extensive check here. We could verify that one is bigger than the other, and we could also verify that they are in a certain time range.

sentry/src/main/java/io/sentry/ProfilingTransactionData.java
```diff
@@ -94,6 +112,8 @@ public boolean equals(Object o) {
         && traceId.equals(that.traceId)
         && name.equals(that.name)
         && relativeStartNs.equals(that.relativeStartNs)
+        && relativeStartCpuMs.equals(that.relativeStartCpuMs)
+        && Objects.equals(relativeEndCpuMs, that.relativeEndCpuMs)
```
philipphofmann, line 116: `h`: `hashCode` also needs an update. We could add some unit tests for equals and hashCode to ensure we don't make the same mistake in the future.

philipphofmann (commented):
One important comment about the hashCode to address, please.

sentry-android-integration-tests/sentry-uitest-android/src/androidTest/java/io/sentry/uitest/android/EnvelopeTests.kt
stefanosiano: some timestamps could be the same, so we should check if one is <= than the other.
But then, the transaction list is not ordered, due to how they are stored (we use a map instead of a list to be able to quickly check the existence of a certain id)
So i should order the list and then check the each timestamp is lower or equals to the following one.
I'll add a comment on it to explain it

philipphofmann (approved):
LGTM

PR #2245, opened 2022-09-14
meta(gha): Add stale bot

philipphofmann:
#skip-changelog

Copied from https://github.com/getsentry/sentry-cocoa/blob/master/.github/workflows/stale.yml

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,47 @@
+name: 'close stale issues/PRs'
+on:
+  schedule:
+    - cron: '0 0 * * *'
+  workflow_dispatch:
+jobs:
+  stale:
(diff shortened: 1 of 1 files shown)


philipphofmann:
Came up during our sync meeting that @adinauer would like to have it. I don't really mind.

vaind:
Rather than copying the workflow (and maintaining/fixing both), maybe it's a good candidate to move to https://github.com/getsentry/github-workflows ?

philipphofmann:
> Rather than copying the workflow (and maintaining/fixing both), maybe it's a good candidate to move to https://github.com/getsentry/github-workflows ?

Definitely, @vaind, do you have the bandwidth to do that? I would love to replace it with a reusable workflow.

PR #794, opened 2019-11-18
adds support for logback encoders (the message field is encoded).

jeacott1:
adds support for logback encoders (the message field is encoded)
adds support for global tags and extras (not threadbound). tags and extras are
 distinguished by the existing mdcTags property.

basic usage:
encoders

```
    <appender name="jsonSentryAppender" class="io.sentry.logback.SentryAppender">
        <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
            <level>WARN</level>
        </filter>
        <encoder
            class="net.logstash.logback.encoder.LoggingEventCompositeJsonEncoder">
           <providers>
               <timestamp />
                <version />
                <logLevel />
                ...
```

basic usage:
global entries as tags or extras

```
  LoggerContext context = (LoggerContext)LoggerFactory.getILoggerFactory();
  context.putProperty("global", "value");
```

PR #2246, opened 2022-09-16
Add support for using `Encoder` with `logback.SentryAppender`

norwood:
## :scroll: Description
this adds support for logback encoders in the sentry appender

this was previously added in https://github.com/getsentry/sentry-java/pull/794/files and was mentioned here https://github.com/getsentry/sentry-java/issues/2217

added a test to the test suite
tested against my own sentry instance that formatted messages go through

## :pencil: Checklist

- [ ] No breaking changes


sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
```diff
@@ -137,6 +140,13 @@ protected void append(@NotNull ILoggingEvent eventObject) {
     return event;
   }
 
+  private String formatted(@NotNull ILoggingEvent loggingEvent) {
+    if (encoder != null) {
+      return new String(encoder.encode(loggingEvent), StandardCharsets.UTF_8);
```
adinauer, line 145: Would it make sense to wrap this with a try/catch just in case a faulty encoder has been configured so we still get the formatted message as fallback?

adinauer (commented):
Thank you for the contribution @norwood 🙏 and sorry for taking so long - we've been pretty busy. Left one question regarding try/catch otherwise this LGTM.

adinauer (approved):

jeacott1:
Is there still a means to add global MDC tags?

adinauer:
@jeacott1 you can set a list of MDC tags that Sentry will pick up and convert to tags as shown in this sample:
https://github.com/getsentry/sentry-java/blob/c1b740d9b4d5665d44550acf71043591e696e377/sentry-samples/sentry-samples-logback/src/main/resources/logback.xml#L14-L15

I looks like we never documented this feature, created https://github.com/getsentry/sentry-java/issues/2351 to do so.

Is the feature not working for you or are you?

jeacott1:
@adinauer the option used to exist due to my lost commit https://github.com/getsentry/sentry-java/pull/794/files
I didn't know there was now an alternative.
thanks!

transducer:
@norwood thanks for reply. It looks like this:

```xml
<appender name="SENTRY" class="io.sentry.logback.SentryAppender">
  <encoder class="ch.qos.logback.core.encoder.LayoutWrappingEncoder">
    <layout class="com.baeldung.logback.MaskingPatternLayout">
      <maskPattern>\"SSN\"\s*:\s*\"(.*?)\"</maskPattern>
    </layout>
  </encoder>
  <options>
    <dsn>***</dsn>
  </options>
</appender>
```

I thought this would be similar to `setEncoder`. But it results in empty events like this:

Screenshot 2023-06-21 at 20 03 04

I guess then I have to figure out how to manually configure Sentry (with Java code?).

adinauer:
@transducer can you please check for processing errors in the Sentry UI? If there's none, you could turn on `debug` and check the JSON that gets sent to Sentry.

transducer:
@adinauer I think something else went wrong in my configuration; locally I missed passing a pattern in the XML, thus the whole of Logback was not properly instantiated and onload also an ERROR was visible from Logback ("PatternLayout("null") - Empty or null pattern"). With proper configuration, it does work! Thanks a lot for the help both.

transducer:
@adinauer it seems breadcrumbs and messages are encoded, but exception messages are not. Any ideas how exception messages could also be using the encoder? In our use case of scrubbing sensitive information, sometimes config (like a database spec) ends up in the message, we want to scrub that.

adinauer:
@transducer can you please specify where exactly you see the message you're referring to? Is this on Sentry UI or on the JSON you can see in debug log?

I'm assuming the problem is this line: https://github.com/getsentry/sentry-java/blob/4d55d5104a79fb58042ce2c5084024450c01b976/sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java#L102

Just trying to confirm before digging deeper. As a workaround for now you should be able to filter data from errors in `beforeSend`.

transducer:
@adinauer it is in the Sentry UI when logging an uncaught exception. It shows the exception message which is not encoded above the stacktrace. Other log messages and breadcrumbs are encoded.

adinauer:
@transducer thanks. Can you confirm the workaround does fix your problem?

adinauer:
@transducer we've created https://github.com/getsentry/sentry-java/issues/2973 to track. We want to discuss this a bit more internally before implementing. Will update there.

PR #2248, opened 2022-09-17
Fix file descriptor leak in FileIO instrumentation

ntoskrnl:
## :scroll: Description

There is a bug in FileIO tracing that causes leak in file descriptors.
The bug affects Android and Java.

The context and all the information is [here](https://github.com/getsentry/sentry-android-gradle-plugin/issues/375). 

When we create `new SentryFileInputStream(file)` or `new SentryFileOutputStream(file)`, 2 file descriptors are opened. But when we close the stream, only one is closed. The cause is in [this comment](https://github.com/getsentry/sentry-android-gradle-plugin/issues/375#issuecomment-1250127522).

```java

// write into several files with SentryFileOutputStream, close and delete them.
ExecutorService executor = Executors.newSingleThreadExecutor();
for (int i = 0; i < 10; i++) {
  final String name = "my-test-file-" + i + ".bin";
  executor.submit(
    () -> {
      File file = new File(name);
      try (FileOutputStream out = new SentryFileOutputStream(file, false)) {
        out.write("Hello world!".getBytes(StandardCharsets.UTF_8));
        out.flush();
        Thread.sleep(50);
      } catch (Exception error) {
        // ignore
      }
      file.delete();
    }
  );
}

// run "lsof" and check if they are still open
executor.submit(
  () -> {
    try {
      StringBuilder output = new StringBuilder();
      Process process = Runtime.getRuntime().exec("lsof");
      BufferedReader reader = new BufferedReader(new InputStreamReader(process.getInputStream()));
      while (true) {
        String line = reader.readLine();
        if (line == null) break;
        if (line.contains("my-test-file-")) {
          output.append(line);
          output.append("\n");
        }
      }
      System.out.println("> slof \n" + output);
      process.destroy();
    } catch (IOException e) {
      // ignore
    }
  }
);

executor.shutdown();
```

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
+++ b/sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
@@ -54,7 +54,7 @@ public final class SentryFileInputStream extends FileInputStream {
 
   private SentryFileInputStream(final @NotNull FileInputStreamInitData data)
       throws FileNotFoundException {
-    super(data.file);
+    super(getFileDescriptor(data.delegate));
     spanManager = new FileIOSpanManager(data.span, data.file, data.isSendDefaultPii);
     delegate = data.delegate;
   }
@@ -112,6 +112,14 @@ public final class SentryFileInputStream extends FileInputStream {
   @Override
   public void close() throws IOException {
     spanManager.finish(delegate);
+  }
+
+  private static FileDescriptor getFileDescriptor(FileInputStream stream) throws FileNotFoundException {
+    try {
+      return stream.getFD();
+    } catch (IOException error) {
+      throw new FileNotFoundException("No file descriptor");
+    }
   }
 
   public static final class Factory {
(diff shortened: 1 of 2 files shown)


adinauer (approved):
LGTM. Thank you very much for fixing this bug!

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
```diff
@@ -114,6 +114,15 @@ public void close() throws IOException {
     spanManager.finish(delegate);
   }
 
+  private static FileDescriptor getFileDescriptor(FileInputStream stream)
```
marandaneto, line 117:
```suggestion
  private static FileDescriptor getFileDescriptor(final @NotNull FileInputStream stream)
```

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileOutputStream.java
```diff
@@ -120,6 +120,15 @@ public void close() throws IOException {
     spanManager.finish(delegate);
   }
 
+  private static FileDescriptor getFileDescriptor(FileOutputStream stream)
```
marandaneto, line 123:
```suggestion
  private static FileDescriptor getFileDescriptor(final @NotNull FileOutputStream stream)
```

Issue #2133, opened 2022-06-24
Switch from empty AndroidManifest to namespace


romtsn:
### Description

AGP will not allow using `package` attribute on an empty AndroidManifest starting from [version 8.0](https://issuetracker.google.com/issues/200682321) (we use it a lot, since we publish android libraries without UI). We should specify a `namespace` attribute in the `android { }` Gradle DSL instead.

Some automation script on how to do it:
https://twitter.com/JakeWharton/status/1540075477108490246?s=20&t=VN5FPQPh_Nf87VXw6038ng 

This will also avoid the manifest merging task altogether

philipphofmann:
@romtsnl, can you maybe add the most important info for this issue here and mention the link as a resource? The tweet could disappear and then the info is lost.

stefanosiano:
merged into master

PR #2251, opened 2022-09-21
Upgrade to Android 13 sdk

stefanosiano:
## :scroll: Description
This pr is still a draft: to work with the new agp version, the latest version of Android Studio is needed. It is already out in the stable version, but we should be sure everyone updated before merging this pr into master.

upgraded Android Gradle Plugin to 7.3.0
upgraded target and compile sdk of sample app to 33 (android 13)
updated getPackageInfo and getApplicationInfo deprecated methods

#skip-changelog

We want to make sure the Android sdk can be built without issues when targeting the latest Android version, currently 13.
So we upgrade our sample app to sdk 33 and fix any deprecation issues, other than testing the app is working fine.

Just running the app locally. We want to run ui tests on Android 13 devices, but we are going to do it when Saucelabs will have working devices with Android 13. There is a [separate pr](https://github.com/getsentry/sentry-java/pull/2252) for this.

## :pencil: Checklist

- [x] No breaking changes


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Upgrade to Android 13 sdk ([#2251](https://github.com/getsentry/sentry-java/pull/2251))
```
marandaneto, line 7: That's not a user-facing change, so not needed

sentry-android-core/src/main/java/io/sentry/android/core/ContextUtils.java
```diff
@@ -28,16 +30,46 @@ static PackageInfo getPackageInfo(final @NotNull Context context, final @NotNull
    * @return the Application's PackageInfo if possible, or null
    */
   @Nullable
+  @SuppressWarnings("deprecation")
   static PackageInfo getPackageInfo(
       final @NotNull Context context, final int flags, final @NotNull ILogger logger) {
     try {
-      return context.getPackageManager().getPackageInfo(context.getPackageName(), flags);
+      if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
```
marandaneto, line 37: There's a class that replaces `Build.VERSION` in the SKD to make it testable.

romtsn:
@stefanosiano since you are on it already, could you also tackle https://github.com/getsentry/sentry-java/issues/2133? It's about switching from AndroidManifest to namespace attribute in build.gradle. I planned to do it when updating to AGP 7.3.0, but you've already done it here.

Otherwise the PR LGTM 👍

romtsn (approved):

stefanosiano:
@romtsn sure, i'll put it here, too

stefanosiano:
@romtsn i did the change. Of course, i didn't make the change to agp 8
would you give another look?
Also, should i merge this pr now, or should i wait a little bit until we are sure everyone updated android studio?

romtsn:
Looks good 👍 I'd merge it now, since it's just a few of us who uses AS and shouldn't be a problem to update. Since AS Dolphin is already stable, I think it's an easy update.

sentry-android-core/src/main/java/io/sentry/android/core/ContextUtils.java
stefanosiano: @marandaneto I created an istance of `BuildInfoProvider` inside these methods. 
I couldn't pass directly the instance we create at the beginning of the init flow.
Is it fine, since it wasn't created anywhere else?

PR #2253, opened 2022-09-21
ci: use reusable Danger workflow

vaind:
Reuses a common Danger workflow. Compared to the current version in this repo, it also recognizes "ci:" & "chore(deps):" title prefixes which causes it to skip the changelog check.

Additionally, new checks that will later be added to the shared dangerfile will automatically propagate to all the repos.


philipphofmann (approved):
LGTM 🥳

PR #2276, opened 2022-10-05
ci: Update update-deps.yml

vaind:
fixes regression from #2267


adinauer (approved):
LGTM. Should the same fix be applied to dart as well (see https://github.com/getsentry/sentry-dart/blob/e390cedf364c65e61623aaffbdb9359bd3e090bf/.github/workflows/update-deps.yml)?

philipphofmann (approved):
LGTM

PR #2290, opened 2022-10-11
Add captureProfile method to hub and client

stefanosiano:
## :scroll: Description
I added the method `captureProfile` to `IHub` and `ISentryClient`, so the profiles can be sent in separate envelopes instead of being attached to transactions.
Since the methods that send transaction and profile together were already exposed e.g. in `HubAdapter` or `SentryClient`, they were deprecated to avoid breaking changes.

We want to separate Profiling and Performance. 
The first step to do so is to send separate envelopes for them.
This will also allows to send timed out profiles without waiting for a transaction to be attached to (more on this on another pr)

I added few unit tests for the new methods and added one to check the deprecated method calls the new ones.

## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next steps
The profiler will send the profile envelope by itself, instead of relying on `SentryTracer`, but will be done in another pr


sentry/src/test/java/io/sentry/SentryClientTest.kt
```diff
@@ -1038,39 +1039,55 @@ class SentryClientTest {
...
     @Test
     fun `when captureTransaction with attachments and profiling data`() {
         val transaction = SentryTransaction(fixture.sentryTracer)
-        fixture.getSut().captureTransaction(transaction, null, createScopeWithAttachments(), null, fixture.profilingTraceData)
-        verifyAttachmentsInEnvelope(transaction.eventId)
-        verifyProfilingTraceInEnvelope(transaction.eventId)
+        val client = fixture.getSut()
+        client.captureTransaction(transaction, null, createScopeWithAttachments(), null)
+        client.captureProfile(fixture.profilingTraceData)
+        verifyProfilingTraceInEnvelope(SentryId(fixture.profilingTraceData.profileId))
```
philipphofmann, line 1051: `m`: I think we should split that test case up into two test cases. Previously, we had one call on the client. Now we have two and only verify for the profiling trace in the envelope. 

Something like
```Kotlin
client.captureTransaction(transaction, null, createScopeWithAttachments(), null)
verifyAttachmentsInEnvelope(transaction.eventId)
```
and
```Kotlin
client.captureProfile(fixture.profilingTraceData)
verifyProfilingTraceInEnvelope(SentryId(fixture.profilingTraceData.profileId))
```

That also applies to some of the other test cases.

sentry/src/test/java/io/sentry/HubTest.kt
```diff
@@ -1312,6 +1281,61 @@ class HubTest {
...
+        val contexts = TransactionContext("name", "op", TracesSamplingDecision(false, null, true, null))
+        val transaction = hub.startTransaction(contexts)
+        transaction.finish()
+        verify(mockClient, never()).captureProfile(any())
+
+        // Transaction is sampled, so it should be profiled
+        val sampledContexts = TransactionContext("name", "op", TracesSamplingDecision(true, null, true, null))
+        val sampledTransaction = hub.startTransaction(sampledContexts)
+        sampledTransaction.finish()
+        verify(mockClient, never()).captureProfile(any())
```
philipphofmann, line 1320: `m`: Shouldn't this be
```suggestion
        verify(mockClient).captureProfile(any())
```

philipphofmann (approved):
Just some comments about the tests, apart from that LGTM

sentry/src/test/java/io/sentry/SentryClientTest.kt
stefanosiano: yep. Actually, there were already those tests, so i just removed the transactions from the captureProfile tests

PR #2298, opened 2022-10-14
Profile envelopes sent directly from profiler

stefanosiano:
## :scroll: Description
The `AndroidTransactionProfiler` now directly sends profile envelopes. This means we don't have to pass the `ProfileTraceData` anymore anyway. 
So we could drop the old timed out profiles handling, as they are directly sent when they time out

This change allows profiles to be sent independently from transactions.
It also fixes and simplify timed out profiles handling. Previously we had to wait for a transaction to finish to attach the timed out profile to it and send the envelope. Now the envelope is sent as soon as the profile times out.

I added a ui test to check the profile is sent on timeout without finishing the transaction
I added few unit tests to check the `captureProfile` method of the hub is called by the profiler itself.

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
@@ -12,12 +12,15 @@ import android.os.Debug;
 import android.os.Process;
 import android.os.SystemClock;
 import io.sentry.HubAdapter;
+import io.sentry.IHub;
 import io.sentry.ITransaction;
 import io.sentry.ITransactionProfiler;
 import io.sentry.ProfilingTraceData;
 import io.sentry.ProfilingTransactionData;
+import io.sentry.SentryEnvelope;
 import io.sentry.SentryLevel;
 import io.sentry.android.core.internal.util.CpuInfoUtils;
+import io.sentry.exception.SentryEnvelopeException;
 import io.sentry.util.Objects;
 import java.io.File;
 import java.util.ArrayList;
@@ -50,6 +53,7 @@ final class AndroidTransactionProfiler implements ITransactionProfiler {
   private @Nullable Future<?> scheduledFinish = null;
   private final @NotNull Context context;
   private final @NotNull SentryAndroidOptions options;
+  private final @NotNull IHub hub;
   private final @NotNull BuildInfoProvider buildInfoProvider;
   private final @Nullable PackageInfo packageInfo;
   private long transactionStartNanos = 0;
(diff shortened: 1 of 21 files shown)


stefanosiano:
Do I need a changelog entry? The `ITransactionProfiler` class is internal, and so should be all the changes in this pr

markushi:
> Do I need a changelog entry? The `ITransactionProfiler` class is internal, and so should be all the changes in this pr

Good question! @romtsn how do we usually deal with internal changes?

markushi:
Looks good to me! 👍

romtsn:
I guess the changelog is not only for the users but also for us, right? So I'd probably add an entry there. Also, this sounds like a behavior change (profiles are sent without transactions now), which makes it reasonable to add to the changelog again.

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -366,10 +365,6 @@ public void finish(@Nullable SpanStatus status) {
       }
 
       transaction.getMeasurements().putAll(measurements);
-
-      if (profilingTraceData != null) {
```
romtsn, line 370: there was no test for this previously? Sounds like it should fail

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
```diff
@@ -287,26 +259,29 @@ public synchronized void onTransactionStart(@NotNull ITransaction transaction) {
...
+            totalMem,
+            options.getProguardUuid(),
+            versionName,
+            versionCode,
+            options.getEnvironment(),
+            isTimeout
+                ? ProfilingTraceData.TRUNCATION_REASON_TIMEOUT
+                : ProfilingTraceData.TRUNCATION_REASON_NORMAL);
+
+    HubAdapter.getInstance().captureProfile(profilingTraceData);
```
romtsn, line 284: How about we move this into the constructor? Then we should be also able to test it properly without overriding `Sentry.setCurrentHub`

Perhaps, we could even make `AndroidTransactionProfiler` an integration, but I'm not sure if it was ever discussed before @marandaneto

adinauer (approved):
LGTM, agree with roman on the changelog. I too think it should be listed.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
marandaneto: Maybe create an envelope out of the `profilingTraceData` and call `captureEnvelope` instead?
Your call here @romtsn 
I don't recall any discussion related to `AndroidTransactionProfiler` as an integration.

sentry/src/main/java/io/sentry/SentryTracer.java
stefanosiano: previously it was captured with the transaction, and the only required parameter was the transaction (the others could be null)
then I switched to send the profile directly using `hub.captureProfile` and added the null check
now i moved the `captureProfile` call into the Profiler directly
the test check if finishing a transaction calls the `profiler.onTransactionFinish` method, and the profiler tests check if the `ISentryClient.captureProfile` is called

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
stefanosiano: Out of curiosity, why is it preferred to create an Envelope and call `captureEnvelope` instead of calling `captureProfile`/`captureTransaction`/`captureWhatever`?
Is it for consistency or are there other reasons? @marandaneto 
I can quickly create the envelope and drop the `captureProfile` methods, if it's better

marandaneto: Because this expands the public API and is not needed.

stefanosiano: @adinauer  Actually, `HubAdapter` is public, and it has a `captureProfile` method

marandaneto: `HubAdapter` is just to make Sentry static class testable.

stefanosiano: That's fine, but `HubAdapter` isn't marked as internal, so technically there's nothing preventing users to actually use it.
Just wanted to make sure we are aware of this

stefanosiano: @marandaneto i made the change. Would you take another look before merging?

romtsn (approved):
@stefanosiano I'm planning to cut a release today, would you like to merge it so it gets in?

Issue #1734, opened 2021-09-22
Make lifecycle breadcrumbs logged by sentry-android-fragment integration customizable


mateuszkwiecinski:
## Context
I tried to add `FragmentLifecycleIntegration` to my app, and I quickly reached default breadcrumbs limit. Almost all events came from `FragmentLifecycleIntegration` which decreased number of visible user interactions to last 2-3 screens before crash. 
I could increase breadcrumbs limit, but I concluded I'd prefer see less events logged by `FragmentLifecycleIntegration`.

## Proposal
The simples thing I can suggest is to introduce api like:
```
FragmentLifecycleIntegration(
	application = application,
	loggedLifecycleEvents = setOf(Lifecycle.State.RESUMED, Lifecycle.State.PAUSED),
)
```
which would skip events non-listed in new parameter.

marandaneto:
@mateuszkwiecinski thanks for reporting, such grain fined level makes sense, let's see if people upvote.

a workaround, for now, is using the `BeforeBreadcrumb` callback to filter breadcrumbs out https://docs.sentry.io/platforms/android/enriching-events/breadcrumbs/#customize-breadcrumbs
or obviously, increasing the number of crumbs.

romtsn:
Closes by #2299

PR #2299, opened 2022-10-14
Fix: Make lifecycle breadcrumbs logged in fragments customizable (#1734)

ILikeYourHat:
## :scroll: Description

Added a way for logging only a subset of a fragment lifecycle events through breadcrumbs.

Issue: https://github.com/getsentry/sentry-java/issues/1734

This change also introduces an API incompatibility for `SentryFragmentLifecycleCallbacks`. This is intentional: in my opinion, this class should have never been a part of public API. It is used only internally in `FragmentLifecycleIntegration`, and I can't think of a reason to use it in a different way. Feel free to discuss if you disagree, I am open to reverting this and providing full API compatibility if it's really needed.

Changed tests to include this new behavior.

## :pencil: Checklist
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,12 +7,22 @@
 - Ensure potential callback exceptions are caught #2123 ([#2291](https://github.com/getsentry/sentry-java/pull/2291))
 - Remove verbose FrameMetricsAggregator failure logging ([#2293](https://github.com/getsentry/sentry-java/pull/2293))
 - Ignore broken regex for tracePropagationTarget ([#2288](https://github.com/getsentry/sentry-java/pull/2288))
+- No longer serialize static fields; use toString as fallback ([#2309](https://github.com/getsentry/sentry-java/pull/2309))
+- Fix `SentryFileWriter`/`SentryFileOutputStream` append overwrites file contents ([#2304](https://github.com/getsentry/sentry-java/pull/2304))
+- Respect incoming parent sampled decision when continuing a trace ([#2311](https://github.com/getsentry/sentry-java/pull/2311))
 
 ### Features
 
+- Profile envelopes are sent directly from profiler ([#2298](https://github.com/getsentry/sentry-java/pull/2298))
 - Add support for using Encoder with logback.SentryAppender ([#2246](https://github.com/getsentry/sentry-java/pull/2246))
-- Add captureProfile method to hub and client ([#2290](https://github.com/getsentry/sentry-java/pull/2290))
 - Report Startup Crashes ([#2277](https://github.com/getsentry/sentry-java/pull/2277))
+- HTTP Client errors for OkHttp ([#2287](https://github.com/getsentry/sentry-java/pull/2287))
+
+### Dependencies
+
+- Bump Native SDK from v0.5.0 to v0.5.1 ([#2306](https://github.com/getsentry/sentry-java/pull/2306))
+  - [changelog](https://github.com/getsentry/sentry-native/blob/master/CHANGELOG.md#051)
+  - [diff](https://github.com/getsentry/sentry-native/compare/0.5.0...0.5.1)
 
 ## 6.5.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Add support for using Encoder with logback.SentryAppender ([#2246](https://github.com/getsentry/sentry-java/pull/2246))
 - Report Startup Crashes ([#2277](https://github.com/getsentry/sentry-java/pull/2277))
 - HTTP Client errors for OkHttp ([#2287](https://github.com/getsentry/sentry-java/pull/2287))
+- Customizable fragment lifecycle breadcrumbs ([#2299](https://github.com/getsentry/sentry-java/pull/2299))
 
 ### Dependencies
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Features
+
+- Customizable fragment lifecycle breadcrumbs ([#2299](https://github.com/getsentry/sentry-java/pull/2299))
+
+## 6.6.0
+
 ### Fixes
 
 - Ensure potential callback exceptions are caught #2123 ([#2291](https://github.com/getsentry/sentry-java/pull/2291))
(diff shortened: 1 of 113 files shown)


sentry-android-fragment/api/sentry-android-fragment.api
```diff
@@ -20,28 +21,19 @@ public final class io/sentry/android/fragment/FragmentLifecycleIntegration : and
 	public fun register (Lio/sentry/IHub;Lio/sentry/SentryOptions;)V
 }
 
-public final class io/sentry/android/fragment/SentryFragmentLifecycleCallbacks : androidx/fragment/app/FragmentManager$FragmentLifecycleCallbacks {
```
romtsn, line 23: I understand your reasoning behind this, but we'd not like to introduce breaking API changes outside of major releases. So I'd suggest to revert this to not delay this PR until the next major release.

Also, it is possible that someone has used the `SentryFragmentLifecycleCallback` class directly, to manually call `registerFragmentLifecycleCallbacks` on the fragment manager, so we'd be breaking those.

sentry-android-fragment/src/main/java/io/sentry/android/fragment/FragmentLifecycleIntegration.kt
```diff
@@ -9,18 +9,35 @@ import io.sentry.IHub
 import io.sentry.Integration
 import io.sentry.SentryLevel.DEBUG
 import io.sentry.SentryOptions
+import io.sentry.android.fragment.internal.SentryFragmentLifecycleCallbacks
 import java.io.Closeable
 
 class FragmentLifecycleIntegration(
     private val application: Application,
-    private val enableFragmentLifecycleBreadcrumbs: Boolean,
+    private val loggedFragmentLifecycleStates: Set<FragmentLifecycleState>,
```
romtsn, line 17: how about we rename it to something more descriptive, e.g. `createFragmentLifecycleBreadcrumbsFor` or something similar. 

I think it's important to keep the `Breadcrumb` part in the variable name, because `logged` can be confusing, as in, we are logging it with an internal logger.

sentry-android-fragment/src/main/java/io/sentry/android/fragment/internal/SentryFragmentLifecycleCallbacks.kt
```diff
@@ -14,24 +14,16 @@ import io.sentry.ISpan
 import io.sentry.SentryLevel.INFO
 import io.sentry.SpanStatus
 import io.sentry.TypeCheckHint.ANDROID_FRAGMENT
+import io.sentry.android.fragment.FragmentLifecycleState
 import java.util.WeakHashMap
 
 @Suppress("TooManyFunctions")
-class SentryFragmentLifecycleCallbacks(
+internal class SentryFragmentLifecycleCallbacks(
```
romtsn, line 21: If we are going to revert the `internal` part, we should also keep the old constructor, similar how we did it with `FragmentLifecycleIntegration`

romtsn (changes requested):
Looks good overall, besides some suggestions. Thank you a lot for doing that 👍 

Also, please add a changelog entry, so the CI passes, you can look up at other PRs on how we do that

sentry-android-fragment/src/main/java/io/sentry/android/fragment/FragmentLifecycleIntegration.kt
ILikeYourHat: Good point. Imo the best possible name would be `enabledFragmentLifecycleBreadcrumbs` (note the `d` in `enabled`). But this name is too similar to the existing `enableFragmentLifecycleBreadcrumbs`, so we must pick something else. What do you think about `filterFragmentLifecycleBreadcrumbs`?

sentry-android-fragment/api/sentry-android-fragment.api
ILikeYourHat: Ok 👍 Do you want to include this change in the next major release? I can prepare PR for it after merging this one

ILikeYourHat:
> We'd also have to update the documentation for the integration https://docs.sentry.io/platforms/android/configuration/integrations/fragment/, but we can take care of that if you don't have time.

No problem, I missed it previously

sentry-android-fragment/api/sentry-android-fragment.api
```diff
@@ -20,15 +21,33 @@ public final class io/sentry/android/fragment/FragmentLifecycleIntegration : and
...
+	public static fun values ()[Lio/sentry/android/fragment/FragmentLifecycleState;
+}
+
 public final class io/sentry/android/fragment/SentryFragmentLifecycleCallbacks : androidx/fragment/app/FragmentManager$FragmentLifecycleCallbacks {
 	public static final field Companion Lio/sentry/android/fragment/SentryFragmentLifecycleCallbacks$Companion;
 	public static final field FRAGMENT_LOAD_OP Ljava/lang/String;
+	public fun <init> (Lio/sentry/IHub;Ljava/util/Set;Z)V
+	public synthetic fun <init> (Lio/sentry/IHub;Ljava/util/Set;ZILkotlin/jvm/internal/DefaultConstructorMarker;)V
 	public fun <init> (Lio/sentry/IHub;ZZ)V
-	public synthetic fun <init> (Lio/sentry/IHub;ZZILkotlin/jvm/internal/DefaultConstructorMarker;)V
```
ILikeYourHat, line 27: This is the only "breaking" change I must leave to keep the code looking sane. Instead of `(IHub, Boolean,Boolean)` the default constructor is now `(IHub,Set,Boolean)`. Of course the old one is also present, but as a secondary constructor. Is this ok for you?

sentry-android-fragment/api/sentry-android-fragment.api
romtsn: hmm, yeah we can do that, though I'm not sure when is the next major, could very well be in a few months 😅

sentry-android-fragment/api/sentry-android-fragment.api
romtsn: yeah, as long as we preserve the old ctor, it's fine 👍

ILikeYourHat:
Do you need anything else from me about this PR? :)

romtsn:
> Do you need anything else from me about this PR? :)

Nope, all good, just haven't had time to merge it. Will do so today. Thanks a lot for your contribution, appreciated 👍

CHANGELOG.md
```diff
@@ -24,6 +26,7 @@
 - HTTP Client errors for OkHttp ([#2287](https://github.com/getsentry/sentry-java/pull/2287))
 - Add option to enable or disable Frame Tracking ([#2314](https://github.com/getsentry/sentry-java/pull/2314))
 
+
```
romtsn, line 29:
```suggestion
```

romtsn (approved):

Issue #2319, opened 2022-10-24
Jetpack Compose Navigation instrumentation support


markushi:
In order to auto-instrument navigation for Jetpack Compose we want to re-route the original `rememberNavController` method call to our sentry specific implementation. This way the required bytecode changes can be kept to a minimum.

philipphofmann:
@markushi, the issue has no description. Maybe it got lost?

PR #2320, opened 2022-10-24
Overload withSentryObservableEffect to provide compose instrumentation hook

markushi:
## :scroll: Description
Overload withSentryObservableEffect to provide compose instrumentation hook

In order to auto-instrument navigation in Jetpack Compose we simply want to adapt the original `rememberNavController` call with our sentry specific navigation listeners. This way the required bytecode changes can be kept to a minimum.

Related to [#2319](https://github.com/getsentry/sentry-java/issues/2319)

Right now just manually, need to investigate how all this Android specifics can be tested.

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryNavigationIntegration.kt
+++ b/sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryNavigationIntegration.kt
@@ -10,7 +10,9 @@ import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.LifecycleEventObserver
 import androidx.lifecycle.LifecycleOwner
 import androidx.navigation.NavController
+import androidx.navigation.NavDestination
 import androidx.navigation.NavHostController
+import androidx.navigation.Navigator
 import io.sentry.Breadcrumb
 import io.sentry.ITransaction
 import io.sentry.SentryOptions
@@ -72,4 +74,10 @@ public fun NavHostController.withSentryObservableEffect(
         }
     }
     return this
+}
+
+@Composable
+public fun rememberNavController(vararg navigators: Navigator<out NavDestination>): NavHostController {
+    return androidx.navigation.compose.rememberNavController(navigators = navigators)
+        .withSentryObservableEffect()
 }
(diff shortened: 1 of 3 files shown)


sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryNavigationIntegration.kt
```diff
@@ -73,3 +75,9 @@ public fun NavHostController.withSentryObservableEffect(
     }
     return this
 }
+
+@Composable
+public fun rememberNavController(vararg navigators: Navigator<out NavDestination>): NavHostController {
```
romtsn, line 80: How about we have a singleton object, like `SentryComposeInstrumentation` where we'd put this method? Now it has the same signature as the original `rememberNavController`, i think the users might get confused when picking the right one from autocomplete.

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryNavigationIntegration.kt
```diff
@@ -73,3 +73,15 @@ public fun NavHostController.withSentryObservableEffect(
     }
     return this
 }
+
+/**
+ * A [DisposableEffect] that captures a [Breadcrumb] and starts an [ITransaction] and sends
+ * them to Sentry for every navigation event when being attached to the respective [NavHostController].
+ */
+@Composable
+public fun NavHostController.withSentryObservableEffect(): NavHostController {
```
romtsn, line 82: I'm still thinking if we should change it to `internal fun`, so those who manually instrument do not get confused between 2 methods? It should still work in bytecode, since it's effectively public, but for the user-written code the kotlin compiler will fail, can we try that?

Or in general having some way to hide it would be great, either via @Deprecated or @OptIn annotations, if the above approach doesn't work

romtsn (approved):

romtsn:
one small remark, but otherwise lgtm, needs a changelog tho

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryNavigationIntegration.kt
markushi: `internal` worked, looking good now!

PR #2321, opened 2022-10-24
Skip saucelabs for community contributions

romtsn:
_#skip-changelog_

E.g. this PR has failed because there are no secrets available in the contributor's repo. https://github.com/getsentry/sentry-java/pull/2299

@vaind do you know if there's a better way to do it?


philipphofmann:
Not running SauceLabs for contributors at all is not great cause SauceLabs is there to find bugs. I'm not sure right now how to fix this either.

vaind:
Running CI with secrets involved on forks is indeed problematic. As described in [this article](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/), there should be an option to use `workflow_run` trigger which would allow us to use secrets within defined within the main repo with artefacts built by a workflow running on a fork. However, this seems a bit more complicated so not sure it's worth it. 

FWIW in dart I've changed the CI to run on push instead of pull_request: https://github.com/getsentry/sentry-dart/pull/1075/files

.github/workflows/integration-tests-ui.yml
```diff
@@ -36,3 +40,4 @@ jobs:
           sauce-username: ${{ secrets.SAUCE_USERNAME }}
           sauce-access-key: ${{ secrets.SAUCE_ACCESS_KEY }}
           config-file: .sauce/sentry-uitest-android-ui.yml
+        if: "${{ env.SAUCE_USERNAME != '' }}"
```
romtsn, line 43:
```suggestion
        if: env.SAUCE_USERNAME !=  null
```

.github/workflows/integration-tests-benchmarks.yml
```diff
@@ -82,3 +91,4 @@ jobs:
           config: sentry-android-integration-tests/metrics-test.yml
           sauce-user: ${{ secrets.SAUCE_USERNAME }}
           sauce-key: ${{ secrets.SAUCE_ACCESS_KEY }}
+        if: "${{ env.SAUCE_USERNAME != '' }}"
```
romtsn, line 94:
```suggestion
        if: env.SAUCE_USERNAME != null      
```

.github/workflows/integration-tests-ui.yml
```diff
@@ -36,3 +40,4 @@ jobs:
           sauce-username: ${{ secrets.SAUCE_USERNAME }}
           sauce-access-key: ${{ secrets.SAUCE_ACCESS_KEY }}
           config-file: .sauce/sentry-uitest-android-ui.yml
+        if: env.SAUCE_USERNAME !=  null
```
markushi:
```suggestion
        if: env.SAUCE_USERNAME != null
```

markushi (approved):

philipphofmann:
> FWIW in dart I've changed the CI to run on push instead of pull_request: https://github.com/getsentry/sentry-dart/pull/1075/files

@vaind, what's the benefit of doing that?

vaind:
> @vaind, what's the benefit of doing that?

not much in terms of benefits - it's just another way of doing the same, i.e. not running when there's no secret. Maybe a slight benefit would be it would run (and fail) on this repo branches where the secret is supposed to be defined, as opposed of the approach in this PR where if someone were to delete the secret, the workflow would just silently skip running tests.

PR #2324, opened 2022-11-02
Populate event.modules with dependencies metadata

### LINKED ISSUE #1470: Populate event modules with dependencies metadata
https://developer.android.com/studio/releases/#4-0-0-dependency-metadata

AGP 4.x creates a file with all the dependencies and their transitive dependencies.

If we decide to read the local file at compile-time, we'd need to use a Gradle plugin.

<project>/<module>/build/outputs/sdk-dependencies/release/sdkDependency.txt

if `includeInApk` is enabled, the same content is shipped with the APK.

> The data is compressed, encrypted by a Google Play signing key, and stored in the signing block of your release app

Because it's encrypted by a Google Play signing key, I don't think we can read at runtime.

romtsn:
## :scroll: Description

* Adds platform-specific dependencies readers for Java and Android, that parse `sentry-external-modules.txt` file ingested by the gradle plugin into the final jar/apk
* Sends those dependencies as part of every SentryEvent

Closes #1470 
Relates to https://github.com/getsentry/sentry-android-gradle-plugin/pull/396
Part of https://github.com/getsentry/team-mobile/issues/54

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/internal/modules/AssetsModulesLoader.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/internal/modules/AssetsModulesLoader.java
@@ -1,0 +1,72 @@
+package io.sentry.android.core.internal.modules;
+
+import android.content.Context;
+import io.sentry.ILogger;
+import io.sentry.SentryLevel;
+import io.sentry.internal.modules.IModulesLoader;
+import java.io.BufferedReader;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.lang.ref.WeakReference;
+import java.nio.charset.Charset;
+import java.util.Map;
+import java.util.TreeMap;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+@ApiStatus.Internal
+public final class AssetsModulesLoader implements IModulesLoader {
+
+  private @Nullable Map<String, String> cachedModules = null;
+  private final @NotNull ILogger logger;
+  private final @NotNull WeakReference<Context> contextRef;
+
+  public AssetsModulesLoader(final @NotNull Context context, final @NotNull ILogger logger) {
+    this.logger = logger;
+    this.contextRef = new WeakReference<>(context);
+  }
+
+  @Override public @Nullable Map<String, String> getOrLoadModules() {
+    if (cachedModules != null) {
+      return cachedModules;
+    }
+    cachedModules = extractModules();
+    return cachedModules;
+  }
+
+  @SuppressWarnings("CharsetObjectCanBeUsed")
+  private Map<String, String> extractModules() {
+    final Map<String, String> modules = new TreeMap<>();
+    try {
+      final Context context = contextRef.get();
+      if (context == null) {
+        return modules;
+      }
+
+      try (final BufferedReader reader =
+             new BufferedReader(new InputStreamReader(context.getAssets().open("sentry-external-modules.txt"), Charset.forName("UTF-8")))) {
+        String module = reader.readLine();
+        while (module != null) {
+          int sep = module.lastIndexOf(':');
+          final String group = module.substring(0, sep);
+          final String version = module.substring(sep + 1);
+          modules.put(group, version);
+          module = reader.readLine();
+        }
+        logger.log(SentryLevel.DEBUG, "Extracted %d modules from resources.", modules.size());
+      } catch (FileNotFoundException e) {
+        logger.log(SentryLevel.INFO, "sentry-external-modules.txt file was not found.");
+      } catch (IOException e) {
+        logger.log(SentryLevel.ERROR, "Error extracting modules.", e);
+      } catch (RuntimeException e) {
+        logger.log(SentryLevel.ERROR, "sentry-external-modules.txt file is malformed.", e);
+      }
+    } catch (SecurityException e) {
+      logger.log(SentryLevel.INFO, "Access to resources denied.", e);
+    }
+    return modules;
+  }
+}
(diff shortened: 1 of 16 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/internal/modules/AssetsModulesLoader.java
```diff
@@ -0,0 +1,44 @@
...
+import java.lang.ref.WeakReference;
+import java.util.Map;
+import java.util.TreeMap;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+@ApiStatus.Internal
+public final class AssetsModulesLoader extends ModulesLoader {
+
+  private final @NotNull WeakReference<Context> contextRef;
```
markushi, line 19: `l`: is there a reason why you're using a `WeakReference` here? As far as I can see `AndroidOptionsInitializer` ensures that context is an application context already.

sentry-android-core/src/test/java/io/sentry/android/core/internal/modules/AssetsModulesLoaderTest.kt
```diff
@@ -0,0 +1,104 @@
...
+        )
+        // the context only called once when there's no in-memory cache
+        verify(fixture.context, times(1)).assets
+    }
+
+    @Test
+    fun `when file does not exist, swallows exception and returns empty map`() {
+        val sut = fixture.getSut(throws = true)
+
+        assertEquals(emptyMap(), sut.orLoadModules)
```
markushi, line 90: `ultra-l`: Could be `assertTrue(sut.orLoadModules.isEmpty())` as well right?

sentry/src/test/java/io/sentry/internal/modules/ResourcesModulesLoaderTest.kt
```diff
@@ -0,0 +1,96 @@
...
+        // then call it second time
+        assertEquals(
+            mapOf(
+                "com.squareup.okhttp3:okhttp" to "3.14.9",
+                "com.squareup.okio:okio" to "1.17.2"
+            ),
+            sut.orLoadModules
+        )
+        // the classloader only called once when there's no in-memory cache
+        verify(fixture.classLoader, times(1)).getResourceAsStream(any())
```
markushi, line 75: `l`: The `times(1)` could be omitted as it's the standard behavior

markushi (approved):
Nice one, LGTM!

sentry/src/main/java/io/sentry/MainEventProcessor.java
```diff
@@ -90,6 +91,13 @@ private void setDebugMeta(final @NotNull SentryEvent event) {
     }
   }
 
+  private void setModules(final @NotNull SentryEvent event) {
+    final Map<String, String> modules = event.getModules();
+    if (modules == null) {
+      event.setModules(options.getModulesLoader().getOrLoadModules());
+    }
+  }
```
marandaneto, line 99: What if we append the items if there are some already?

sentry-android-core/src/main/java/io/sentry/android/core/internal/modules/AssetsModulesLoader.java
```diff
@@ -0,0 +1,44 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.lang.ref.WeakReference;
+import java.util.Map;
+import java.util.TreeMap;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+@ApiStatus.Internal
+public final class AssetsModulesLoader extends ModulesLoader {
```
marandaneto, line 17: L: Method `getProguardUUID` also uses `context.getAssets()`, maybe we move that method to this modules loader as well?
Maybe we rename the interface to something more generic, like AssetsLoader instead? having methods for the modules and proguard.

sentry/src/main/java/io/sentry/MainEventProcessor.java
romtsn: yeah we could do that, what would be the usecase tho? Hybrid SDKs setting their packages and we're appending to them? 

If we do that, can the MainEventProcessor run twice for the same event, so we're potentially doubling the modules list in that case?

sentry-android-core/src/main/java/io/sentry/android/core/internal/modules/AssetsModulesLoader.java
romtsn: hmm, not sure, the only reason I created this class is because plain java also needs a ModulesLoader, hence different implementations. I think there's also not much in common for proguarduuid and modules besides opening a stream and catching a few exceptions.

If there's a new thing which we have to read from the assets I'll think about how to better generalize it, but for now I'd just keep it like this

sentry-android-core/src/main/java/io/sentry/android/core/internal/modules/AssetsModulesLoader.java
romtsn: No real reason, just extra-safety I guess 🤣

sentry/src/main/java/io/sentry/MainEventProcessor.java
marandaneto: `modules` is part of the public API, and people can raise events with their own modules, the current behavior is to keep theirs or ours, I'd prefer to send them both (Imagine hybrid SDKs here with JS modules + Native modules, that is awesome and right now it won't be possible).
Event processors run only once per event, so I'm not aware of how this could happen.

sentry-android-core/src/main/java/io/sentry/android/core/internal/modules/AssetsModulesLoader.java
marandaneto: Makes sense, your call to refactor this later if needed.

sentry/src/main/java/io/sentry/MainEventProcessor.java
romtsn: Cool, thanks, I'll change it then 👍

PR #2329, opened 2022-11-03
Call setSustainedPerformanceMode for better benchmarks

stefanosiano:
## :scroll: Description
Added window.setSustainedPerformanceMode(true) on benchmark activity, to get better results. 
Reduced espresso default timeout to 10 seconds
Updated list of devices for benchmarks

#skip-changelog

`setSustainedPerformanceMode` is one of the things the android benchmark runner (which we cannot use due to problems on Saucelabs) does. It tries to limit the thermal throttling by reducing the overall performance. https://developer.android.com/reference/kotlin/androidx/benchmark/junit4/AndroidBenchmarkRunner#clock-stability
This will result in more predictable benchmarks, as currently there are few runs failing.
Saucelabs devices for benchmarks were updated: the Google Pixel 3 was replaced because consistently timing out

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/.sauce/sentry-uitest-android-benchmark.yml
+++ b/.sauce/sentry-uitest-android-benchmark.yml
@@ -22,7 +22,7 @@ suites:
   - name: "Android 12 (api 31)"
     devices:
       - id: Google_Pixel_6_Pro_real_us # Google Pixel 6 Pro - api 31 (12) - high end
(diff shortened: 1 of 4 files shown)


romtsn (approved):

markushi (approved):
Nice!

PR #2331, opened 2022-11-04
Add dependabot for github-actions

### LINKED ISSUE #2296: Use dependabot to upgrade gh-actions versions
### Description

We should use dependabot to track gh-actions version changes and update them automatically. Prior art: https://github.com/getsentry/sentry-react-native/blob/main/.github/dependabot.yml

romtsn:
_#skip-changelog_

Closes #2296


markushi (approved):

philipphofmann:
It works 💪💯

PR #2343, opened 2022-11-08
VendorId should be a String

markushi:
## :scroll: Description
`gpu.vendorId` was defined as an Int, should be a String.

This is a breaking change but I'm wondering if anyone is using it. We could introduce a `@Deprecated void setVendorId(int vendorId) { setVendorId(String.valueOf(vendorId)); }` to minimize the impact, but I'm not sure if it's worth it.

Mobile Issue: https://github.com/getsentry/team-mobile/issues/67
Dart: https://github.com/getsentry/sentry-dart/pull/1112

## :pencil: Checklist

- [ ] No breaking changes

### DIFF
--- a/sentry/src/main/java/io/sentry/protocol/Gpu.java
+++ b/sentry/src/main/java/io/sentry/protocol/Gpu.java
@@ -22,7 +22,7 @@ public final class Gpu implements JsonUnknown, JsonSerializable {
   /** The PCI identifier of the graphics device. */
   private @Nullable Integer id;
   /** The PCI vendor identifier of the graphics device. */
-  private @Nullable Integer vendorId;
+  private @Nullable String vendorId;
   /** The vendor name as reported by the graphics device. */
   private @Nullable String vendorName;
   /** The total GPU memory available in Megabytes. */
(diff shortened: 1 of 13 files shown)


romtsn:
I'm fine with the breaking change not being in major, @marandaneto what do you say?

romtsn (approved):

sentry/src/main/java/io/sentry/protocol/Gpu.java
```diff
@@ -74,11 +74,11 @@ public void setId(Integer id) {
     this.id = id;
   }
 
-  public @Nullable Integer getVendorId() {
+  public @Nullable String getVendorId() {
     return vendorId;
   }
 
-  public void setVendorId(Integer vendorId) {
+  public void setVendorId(String vendorId) {
```
marandaneto, line 81:
```suggestion
  public void setVendorId(@Nullable String vendorId) {
```

marandaneto:
> This is a breaking change but I'm wondering if anyone is using it. We could introduce a `@Deprecated void setVendorId(int vendorId) { setVendorId(String.valueOf(vendorId)); }` to minimize the impact, but I'm not sure if it's worth it.

I like the idea:

> This is a breaking change but I'm wondering if anyone is using it. We could introduce a @Deprecated void setVendorId(int vendorId) { setVendorId(String.valueOf(vendorId)); } to minimize the impact, but I'm not sure if it's worth it.

But I dont think anyone is using it directly, Dart has no overloads, so I considered a bug fix because right now it does not work anyway.

PR #2360, opened 2022-11-14
Make possible screenshots on hybrid sdks (react-native)

krystofwoldrich:
## :scroll: Description
To make screenshots work on RN we need to set the current activity for the ScreenshotEventProcessor when initializing the Android SDK otherwise [it will be null](https://github.com/getsentry/sentry-react-native/issues/2362#issuecomment-1193763553).

We will be sending the event with a screenshot therefore we need to ensure another screenshot won't be added by the Android SDK.

- https://github.com/getsentry/team-mobile/issues/11
- https://github.com/getsentry/sentry-react-native/pull/2610

## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next steps
#skip-changelog


sentry-android-core/src/main/java/io/sentry/android/core/ScreenshotEventProcessor.java
```diff
@@ -51,10 +52,17 @@ public ScreenshotEventProcessor(
     application.registerActivityLifecycleCallbacks(this);
   }
 
+  public void setCurrentActivity(@NonNull Activity activity) {
+    if (currentActivity != null && currentActivity.get() == activity) {
+      return;
+    }
+    currentActivity = new WeakReference<>(activity);
+  }
```
marandaneto, line 60: L: any reason to move the code?

sentry-android-core/src/main/java/io/sentry/android/core/ScreenshotEventProcessor.java
```diff
@@ -51,10 +52,17 @@ public ScreenshotEventProcessor(
...
+
   @SuppressWarnings("NullAway")
   @Override
   public @NotNull SentryEvent process(final @NotNull SentryEvent event, @NotNull Hint hint) {
-    if (!lifecycleCallbackInstalled) {
+    if (!lifecycleCallbackInstalled || !HintUtils.shouldApplyScopeData(hint)) {
```
marandaneto, line 65: H: `EnvelopeFileObserver` uses the `CachedEnvelopeHint` hint so this won't work, we need to extend the current hint type for Hybrid SDKs most likely.
The goal here is that at least for screenshots, screenshots should not be taken (again) for Hybrid SDKs.

sentry-android-core/src/main/java/io/sentry/android/core/ScreenshotEventProcessor.java
krystofwoldrich: I don't know how is the lint set in this project, but I saw everywhere the public methods are before private.

sentry-android-core/src/main/java/io/sentry/android/core/ScreenshotEventProcessor.java
krystofwoldrich: Okay got it, that was my goal, I've just misunderstood the function.

sentry-android-core/src/main/java/io/sentry/android/core/ScreenshotEventProcessor.java
```diff
@@ -51,10 +54,86 @@ public ScreenshotEventProcessor(
     application.registerActivityLifecycleCallbacks(this);
   }
 
+  public static ScreenshotEventProcessor createInstance(
```
marandaneto, line 57:
```suggestion
  public static @NotNull ScreenshotEventProcessor getInstance(
```

H: Always use the nullability annotations, better Kotlin interoperability.
M: Stick to the convention of `getInstance` naming.

sentry-android-core/src/main/java/io/sentry/android/core/ScreenshotEventProcessor.java
```diff
@@ -51,10 +54,86 @@ public ScreenshotEventProcessor(
...
+      final @NotNull Application application,
+      final @NotNull SentryAndroidOptions options,
+      final @NotNull BuildInfoProvider buildInfoProvider) {
+    ScreenshotEventProcessor.instance =
+        new ScreenshotEventProcessor(application, options, buildInfoProvider);
+    return ScreenshotEventProcessor.instance;
```
marandaneto, line 63: H: if `instance` isn't null, return itself.

sentry-android-core/src/main/java/io/sentry/android/core/ScreenshotEventProcessor.java
```diff
@@ -33,6 +34,8 @@
 public final class ScreenshotEventProcessor
     implements EventProcessor, Application.ActivityLifecycleCallbacks, Closeable {
 
+  private static @NotNull ScreenshotEventProcessor instance;
```
marandaneto, line 37: H: if `createInstance` was not called, this returns null, so it should be nullable or we should not use a singleton pattern

sentry-android-core/src/main/java/io/sentry/android/core/ScreenshotEventProcessor.java
```diff
@@ -51,10 +54,86 @@ public ScreenshotEventProcessor(
...
+  public static ScreenshotEventProcessor createInstance(
+      final @NotNull Application application,
+      final @NotNull SentryAndroidOptions options,
+      final @NotNull BuildInfoProvider buildInfoProvider) {
+    ScreenshotEventProcessor.instance =
+        new ScreenshotEventProcessor(application, options, buildInfoProvider);
+    return ScreenshotEventProcessor.instance;
+  }
+
+  public static ScreenshotEventProcessor getInstance() {
```
marandaneto, line 66: Similar to https://github.com/getsentry/sentry-java/pull/2360/files#r1023634311

sentry-android-core/src/main/java/io/sentry/android/core/ScreenshotEventProcessor.java
```diff
@@ -51,10 +54,86 @@ public ScreenshotEventProcessor(
...
+      this.options.getLogger().log(SentryLevel.ERROR, "Taking screenshot failed.", e);
+    }
+    return null;
+  }
+
   @SuppressWarnings("NullAway")
   @Override
   public @NotNull SentryEvent process(final @NotNull SentryEvent event, @NotNull Hint hint) {
-    if (!lifecycleCallbackInstalled) {
+    if (HintUtils.getIsFromHybridSdk(hint)
```
marandaneto, line 133: M: I'd move the `getIsFromHybridSdk` later otherwise Native events pay the price every time even if the integration isn't installed nor its an error event.

sentry/src/main/java/io/sentry/util/HintUtils.java
```diff
@@ -16,6 +19,24 @@ public final class HintUtils {
 
   private HintUtils() {}
 
+  @ApiStatus.Internal
```
marandaneto, line 22: M: The class is already `@ApiStatus.Internal`, no need to add to every method.

sentry/src/main/java/io/sentry/util/HintUtils.java
```diff
@@ -16,6 +19,24 @@ public final class HintUtils {
...
+  public static void setIsFromHybridSdk(Hint hint, String sdkName) {
+    if (sdkName.startsWith(SENTRY_REACT_NATIVE_SDK_NAME)
+        || sdkName.startsWith(SENTRY_DART_SDK_NAME)) {
```
marandaneto, line 25: L: Other Hybrid SDKs would have the same issue I suspect, but we can expand this later.

sentry/src/main/java/io/sentry/util/HintUtils.java
```diff
@@ -16,6 +19,24 @@ public final class HintUtils {
...
+    }
+  }
+
+  @ApiStatus.Internal
+  public static void setIsFromHybridSdk(Hint hint, boolean isFromHybridSdk) {
+    hint.set(SENTRY_IS_FROM_HYBRID_SDK, isFromHybridSdk);
+  }
+
+  @ApiStatus.Internal
+  public static boolean getIsFromHybridSdk(Hint hint) {
```
marandaneto, line 36:
```suggestion
  public static boolean isFromHybridSdk(Hint hint) {
```

marandaneto:
Missing a few things, nullability annotations, naming convention, and tests.

sentry-android-core/src/main/java/io/sentry/android/core/ScreenshotEventProcessor.java
```diff
@@ -35,26 +30,22 @@
...
   public ScreenshotEventProcessor(
-      final @NotNull Application application,
-      final @NotNull SentryAndroidOptions options,
-      final @NotNull BuildInfoProvider buildInfoProvider) {
+      final @NotNull Application application, final @NotNull SentryAndroidOptions options) {
     this.application = Objects.requireNonNull(application, "Application is required");
     this.options = Objects.requireNonNull(options, "SentryAndroidOptions is required");
-    this.buildInfoProvider =
-        Objects.requireNonNull(buildInfoProvider, "BuildInfoProvider is required");
+    this.currentActivityHolder = CurrentActivityHolder.getInstance();
```
marandaneto, line 40: H: You should always access `CurrentActivityHolder.getInstance()` directly.

sentry-android-core/src/main/java/io/sentry/android/core/CurrentActivityHolder.java
```diff
@@ -0,0 +1,42 @@
...
+
+  private @Nullable WeakReference<Activity> currentActivity;
+
+  public static CurrentActivityHolder getInstance() {
+    if (instance != null) {
+      return instance;
+    }
+    instance = new CurrentActivityHolder();
+    return instance;
+  }
```
marandaneto, line 22: H: Check how other `getInstance()` is done, this implementation is not multithreaded safe.

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/ActivityUtils.java
```diff
@@ -0,0 +1,20 @@
...
+import androidx.annotation.Nullable;
+import org.jetbrains.annotations.ApiStatus;
+
+@ApiStatus.Internal
+public class ActivityUtils {
+  public static boolean isActivityValid(@Nullable Activity activity) {
+    if (activity == null) {
+      return false;
+    }
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR1) {
```
marandaneto, line 14: H: The `buildInfoProvider` has been removed and now this cannot be unit tested.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidLogger.java
```diff
@@ -3,12 +3,22 @@
...
 
-final class AndroidLogger implements ILogger {
+@ApiStatus.Internal
+public final class AndroidLogger implements ILogger {
 
-  private static final String tag = "Sentry";
+  private final String tag;
+
+  public AndroidLogger(){
+    tag = "Sentry";
```
marandaneto, line 16:
```suggestion
    this("Sentry");
```

sentry-android-core/src/main/java/io/sentry/android/core/AndroidLogger.java
```diff
@@ -3,12 +3,22 @@
...
+import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-final class AndroidLogger implements ILogger {
+@ApiStatus.Internal
+public final class AndroidLogger implements ILogger {
 
-  private static final String tag = "Sentry";
+  private final String tag;
```
marandaneto, line 13:
```suggestion
  private final @NotNull String tag;
```

M: double check all the missing nullability annotations

sentry-android-core/src/main/java/io/sentry/android/core/AndroidLogger.java
```diff
@@ -3,12 +3,22 @@
...
+
+  public AndroidLogger() {
+    this("Sentry");
+  }
+
+  public AndroidLogger(@NotNull String tag) {
```
marandaneto, line 19:
```suggestion
  public AndroidLogger(final @NotNull String tag) {
```

M: Always make fields `final` when possible, there are more cases in this PR

sentry-android-core/src/main/java/io/sentry/android/core/ScreenshotEventProcessor.java
```diff
@@ -69,60 +63,28 @@ public ScreenshotEventProcessor(
 
       return event;
     }
+    if (CurrentActivityHolder.getInstance().getActivity() == null
```
marandaneto, line 66:
```suggestion
    final Activity activity = CurrentActivityHolder.getInstance().getActivity()
    if (activity == null
```

M: Now you can reuse the `activity` object which lives only in the scope of this method.

sentry-android-core/src/main/java/io/sentry/android/core/ScreenshotEventProcessor.java
```diff
@@ -69,60 +63,28 @@ public ScreenshotEventProcessor(
...
+        takeScreenshot(
+            CurrentActivityHolder.getInstance().getActivity(),
+            options.getLogger(),
+            buildInfoProvider);
+    if (screenshot == null) {
+      return event;
     }
 
+    hint.setScreenshot(Attachment.fromScreenshot(screenshot));
+    hint.set(ANDROID_ACTIVITY, CurrentActivityHolder.getInstance().getActivity());
```
marandaneto, line 81:
```suggestion
    hint.set(ANDROID_ACTIVITY, activity);
```

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/ScreenshotUtils.java
```diff
@@ -0,0 +1,80 @@
...
+import io.sentry.SentryLevel;
+import io.sentry.android.core.BuildInfoProvider;
+import java.io.ByteArrayOutputStream;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+@ApiStatus.Internal
+public class ScreenshotUtils {
+  public static @Nullable byte[] takeScreenshot(
+      final @Nullable Activity activity,
```
marandaneto, line 20: L: I guess we can do the `activity` check before and make it `NotNull` here.

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/ScreenshotUtils.java
```diff
@@ -0,0 +1,80 @@
...
+      // screenshot png is around ~100-150 kb
+      return byteArrayOutputStream.toByteArray();
+    } catch (Throwable e) {
+      logger.log(SentryLevel.ERROR, "Taking screenshot failed.", e);
+    }
+    return null;
+  }
+
+  @SuppressLint("NewApi")
+  public static boolean isActivityValid(
```
marandaneto, line 69:
```suggestion
  private static boolean isActivityValid(
```

sentry/src/main/java/io/sentry/util/HintUtils.java
```diff
@@ -16,6 +19,21 @@ public final class HintUtils {
...
   private HintUtils() {}
 
+  public static void setIsFromHybridSdk(@NotNull Hint hint, @NotNull String sdkName) {
+    if (sdkName.startsWith(SENTRY_REACT_NATIVE_SDK_NAME)
+        || sdkName.startsWith(SENTRY_DART_SDK_NAME)) {
+      HintUtils.setIsFromHybridSdk(hint, true);
+    }
+  }
+
+  public static void setIsFromHybridSdk(@NotNull Hint hint, boolean isFromHybridSdk) {
```
marandaneto, line 29: M: This can be private or inlined with the other `setIsFromHybridSdk` method, I don't see it being called from somewhere else.

sentry/src/main/java/io/sentry/util/HintUtils.java
krystofwoldrich: I've changed it a bit because we can consider all the SDKs based on JS, Dotnet, or Dart them hybrid then the check can be more universal.

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/ScreenshotUtils.java
```diff
@@ -0,0 +1,73 @@
...
+
+    try {
+      // ARGB_8888 -> This configuration is very flexible and offers the best quality
+      final Bitmap bitmap =
+          Bitmap.createBitmap(view.getWidth(), view.getHeight(), Bitmap.Config.ARGB_8888);
+
+      final Canvas canvas = new Canvas(bitmap);
+      view.draw(canvas);
+
+      final ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
```
marandaneto, line 45: M: Likely an improvement I forgot to do before:
`ByteArrayOutputStream` can be created within a try with resources so it auto closes the stream, right now if it fails for some reason, the stream is kept open.
Example:
`try (final ByteArrayOutputStream stream = new ByteArrayOutputStream();) {}`

marandaneto:
Left the last comment but all is good on my side.
I will let the maintainers do the final approval.

marandaneto (approved):

romtsn:
it's just flaky

sentry/src/main/java/io/sentry/util/HintUtils.java
```diff
@@ -16,6 +20,18 @@ public final class HintUtils {
 
   private HintUtils() {}
 
+  public static void setIsFromHybridSdk(final @NotNull Hint hint, final @NotNull String sdkName) {
```
romtsn, line 23: `l`: should these 2 new methods be also marked as `@Internal`?

sentry-android-core/src/main/java/io/sentry/android/core/CurrentActivityHolder.java
```diff
@@ -0,0 +1,39 @@
...
+import androidx.annotation.Nullable;
+import java.lang.ref.WeakReference;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+@ApiStatus.Internal
+public class CurrentActivityHolder {
+
+  private static final @NotNull CurrentActivityHolder instance = new CurrentActivityHolder();
+
```
romtsn, line 14: `m`: can we also add a private constructor here so it's not possible to instantiate the `CurrentActivityHolder` and we are always forced to use the singleton instance?
```suggestion

    private CurrentActivityHolder() { }
```

romtsn (approved):
lgtm 👍 can you also please add a changelog entry - I understand that it's more-or-less an internal change, but if something goes wrong and this PR causes bugs, we can quickly look at the past releases and see if there was something suspicious in the changelog.

sentry/src/main/java/io/sentry/util/HintUtils.java
marandaneto: The class is already internal.

marandaneto:
> lgtm 👍 can you also please add a changelog entry - I understand that it's more-or-less an internal change, but if something goes wrong and this PR causes bugs, we can quickly look at the past releases and see if there was something suspicious in the changelog.

Changelog is meant for user-facing changes, I'd not add it but your call :)

sentry/src/main/java/io/sentry/util/HintUtils.java
romtsn: then we should probably remove it from the other methods to keep it aligned

Issue #1828, opened 2021-12-06
Align `getSpan` method from Scope on Android/Java and iOS


marandaneto:
related to https://github.com/getsentry/sentry-java/issues/1827 and https://github.com/getsentry/sentry-java/issues/1822

Since they have different behavior on both platforms.
on iOS it always returns the root transaction.
on Android/Java it returns the latest active span.

marandaneto:
If hub is `globalMode` enabled, we always return the root transaction.

Doing this we don't need to do https://github.com/getsentry/sentry-java/issues/1827

This is going to break the relationship for the DB auto instrumentation, We have to create the parent span, hold the reference and create the child from it, which requires changes on the Plugin.

We should talk about the thread-local storage before going forward.

romtsn:
Closed by #2855

PR #2369, opened 2022-11-16
Add ttid span to ActivityLifecycleIntegration

stefanosiano:
## :scroll: Description
Added ttid (time to initial display) span to ActivityLifecycleIntegration

We want to add the ttid info to the transactions, so we are adding a span to the automatic activity tracking transactions

Added unit test to check ttid span is created and cancelled in case activity finishes early

## :pencil: Checklist

- [x] No breaking changes


sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -198,8 +205,24 @@ private void startTracing(final @NotNull Activity activity) {
         appStartSpan =
             transaction.startChild(
                 getAppStartOp(coldStart), getAppStartDesc(coldStart), appStartTime);
+        // The first activity ttidSpan should start at the same time as the app start time
+        ttidSpan = transaction.startChild("TTID", activityName + ".ttid", appStartTime);
```
romtsn, line 209: I guess we'd need to raise it with the TSC, because this is can affect performance issues and so on and we have a documented list of span ops https://develop.sentry.dev/sdk/performance/span-operations/. You said you're on PTO next week? I can bring it up instead, if you want me to.

romtsn:
I've seen Firebase is taking a slightly different approach using `OnDrawListener` and posting at front of the queue from within the `onDraw` method - do you think it can be more precise than our approach? Links for reference

https://github.com/firebase/firebase-android-sdk/blob/e0cfdd60b8c7204805f78874892a03dccbe174af/firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java#L319

https://github.com/firebase/firebase-android-sdk/blob/e0cfdd60b8c7204805f78874892a03dccbe174af/firebase-perf/src/main/java/com/google/firebase/perf/util/FirstDrawDoneListener.java

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
stefanosiano: yes, please
thanks

stefanosiano:
I'll try to implement it in the same way just to compare the values and see, thanks for posting it

stefanosiano:
Firebase implementation logs a lot of things, but the approach on using `OnDrawListener` is more reliable.
While timings are the same as this implementation most of the times, there are few cases when using Firebase's approach is unsurprisingly more correct.
I'm making this pr a draft and will switch implementation taking inspiration from Firebase

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/FirstDrawDoneListener.java
```diff
@@ -0,0 +1,95 @@
...
+import android.os.Handler;
+import android.os.Looper;
+import android.view.View;
+import android.view.ViewTreeObserver;
+import androidx.annotation.RequiresApi;
+import io.sentry.android.core.BuildInfoProvider;
+import java.util.concurrent.atomic.AtomicReference;
+import org.jetbrains.annotations.NotNull;
+
+/**
```
romtsn, line 14: `m:` I think we should put a comment here that we adapted it from Firebase. SOmething like "Adapted from <link> under Apache 2.0 license` or so

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/MainActivity.java
```diff
@@ -206,7 +205,7 @@ protected void onResume() {
     final ISpan span = Sentry.getSpan();
     if (span != null) {
       span.setMeasurement("screen_load_count", screenLoadCount, new MeasurementUnit.Custom("test"));
-      span.finish(SpanStatus.OK);
+      // span.finish(SpanStatus.OK);
```
stefanosiano, line 208: @romtsn I found that using `Sentry.getSpan()` and then `span.finish()` is finishing the ttid span
Is it something to be worried about? Would a user be already using something like this?

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+- Add ttid span to ActivityLifecycleIntegration ([#2369](https://github.com/getsentry/sentry-java/pull/2369))
```
romtsn, line 7: Maybe we could expand this after renaming the op name 
```suggestion
- Add time-to-initial-display span to Activity transactions ([#2369](https://github.com/getsentry/sentry-java/pull/2369))
```

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -250,18 +276,25 @@ private boolean isRunningTransaction(final @NotNull Activity activity) {
...
+      final @Nullable ITransaction transaction, final @Nullable ISpan ttidSpan) {
     if (transaction != null) {
       // if io.sentry.traces.activity.auto-finish.enable is disabled, transaction may be already
       // finished manually when this method is called.
       if (transaction.isFinished()) {
         return;
       }
 
+      // in case the ttidSpan isn't completed yet, we finish it as cancelled to avoid memory leak
+      if (ttidSpan != null && !ttidSpan.isFinished()) {
```
romtsn, line 294: `l`: we're repeating this code 3 times, maybe worth to extract into a method?

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/MainActivity.java
romtsn: yes, it's possible that the users already do this to finish the activity transaction whenever they think it's suitable for them. 

Once we implement https://github.com/getsentry/sentry-java/issues/1828, this should not matter anymore, but until then, we should probably mention this in the docs + migration docs, that calling span.finish in onResume will change how ttid span is measured.

romtsn (approved):
Looks good to me overall, great job 👍 One more thing I'd like to ask you - could you enable [auto-finish ](https://github.com/getsentry/sentry-java/blob/da3486184e591b3981d8dbdb258c099914dc4541/sentry-samples/sentry-samples-android/src/main/AndroidManifest.xml#L112)for activity transactions and test if it works fine with the ttid spans? I think it shouldn't change anything but just for the sake of correctness.

romtsn:
> I think it shouldn't change anything but just for the sake of correctness.

Actually after looking at this one more time, I think it will affect the ttid span, because currently we're finishing it as cancelled inside the `stopTracing` method, which is also called when auto-finish is enabled. 

Perhaps we should only finish the ttid span as cancelled in `onActivityDestroyed`, otherwise we should just finish the activity transaction, and keep the ttid span running until it finishes itself

stefanosiano:
> Perhaps we should only finish the ttid span as cancelled in onActivityDestroyed, otherwise we should just finish the activity transaction, and keep the ttid span running until it finishes itself

Yep, auto-finish stops the ttid. Let's finish the span only in `onActivityDestroyed` then. Thanks for the catch

romtsn:
> What about the stopPreviousTransactions()? It makes sense to cancel the ttid span in that case, right?

yeah I think it makes sense, otherwise it will run indefinitely right? if we switched the screens in the meantime

stefanosiano:
> yeah I think it makes sense, otherwise it will run indefinitely right? if we switched the screens in the meantime

It would end when the previous activity is destroyed, but if an activity is started without destroying the previous one, then it would run indefinitely, yeah

PR #2371, opened 2022-11-16
Update profiling release field

stefanosiano:
## :scroll: Description
put the release in the profiling payload
#skip-changelog

We were previously using the version name and version code. We moved to using the release to be aligned with the other events

Unit test

## :pencil: Checklist

- [x] No breaking changes


sentry/src/main/java/io/sentry/ProfilingTraceData.java
```diff
@@ -533,16 +521,10 @@ public static final class Deserializer implements JsonDeserializer<ProfilingTrac
...
             }
             break;
-          case JsonKeys.VERSION_NAME:
+          case JsonKeys.RELEASE:
             String versionName = reader.nextStringOrNull();
             if (versionName != null) {
-              data.versionName = versionName;
-            }
-            break;
-          case JsonKeys.VERSION_CODE:
```
romtsn, line 542: I'm thinking of a case where the user had some unsent envelopes and updated the SDK in the meantime - in this case, we would only send the `version_name` in the old format, but without the `version_code` - would it be fine for backend? Or should we keep the `version_code` deserialization for backwards compatibility?

stefanosiano: it's an edge case, but yeah, i can keep the deserialization and just init the value on new envelopes as empty strings

romtsn (approved):

Issue #2373, opened 2022-11-17
Envelope with offline storage can't reupload


wangbax:
### Integration

sentry-android

### Build System

Gradle

### AGP Version

7.0.2

### Proguard

Disabled

### Version

6.7.0

### Steps to Reproduce

1. close network
2. open app
3. close app
4. reopen app

### Expected Result

Adding all envelope to offline storage: `/data/user/0/[package]/cache/sentry/[dsn]/[id].envelope`

### Actual Result

Adding all envelope to offline storage: `/data/user/0/[package]/cache/sentry/[id].envelope`

<img width="900" alt="image" src="https://user-images.githubusercontent.com/5964735/202350469-aedd868b-3971-4e82-b75f-41ef4d5e950b.png">

wangbax:
This issues actually is related to [new AndroidEnvelopeCache(options)](https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java#L136) #2277 

<img width="902" alt="image" src="https://user-images.githubusercontent.com/5964735/202351021-be71e683-facc-4c41-9a6b-8566547e2e74.png">
<img width="1150" alt="image" src="https://user-images.githubusercontent.com/5964735/202351272-9133d429-ea0c-4804-9e46-11c065b982e2.png">

My solution: 
``` kotlin
SentryAndroid.init(context) { options ->
      options.apply {
        this.dsn = dsn
        // after change dsn, try reset disk cache
        setEnvelopeDiskCache(AndroidEnvelopeCache(options))
      }
    }
```

Hope can fix this issue least version. ☺️

markushi:
FYI: The fix will be part of the `6.8.0` release.

PR #2377, opened 2022-11-17
Fix ensure all options are processed before integrations are loaded

markushi:
## :scroll: Description
Right now some integrations are created between default/auto-init and the custom `SentryOptions.configure` callback. If an integration copies references from `SentryOptions` it can hold a stale configuration, as the  `SentryOptions.configure` block can change any option afterwards.

This is a quick fix, I guess in the long run we should take a look into streamlining the overall SDK init process to avoid these kind of issues.

This PR splits up the `AndroidOptionsInitializer.init` method into two parts:
- `loadDefaultAndMetadataOptions()`
- `initializeIntegrationsAndProcessors()`

This allows us to execute execute the custom `OptionsConfiguration.configure()` callback in-between, which ensures all options are set before any integrations and processors are created.

Fixes https://github.com/getsentry/sentry-java/issues/2373

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
@@ -7,6 +7,7 @@ import android.content.Context;
 import android.content.pm.PackageInfo;
 import android.content.res.AssetManager;
 import android.os.Build;
+import androidx.annotation.NonNull;
 import io.sentry.ILogger;
 import io.sentry.SendFireAndForgetEnvelopeSender;
 import io.sentry.SendFireAndForgetOutboxSender;
@@ -24,6 +25,7 @@ import java.io.InputStream;
 import java.util.Properties;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import org.jetbrains.annotations.TestOnly;
 
 /**
  * Android Options initializer, it reads configurations from AndroidManifest and sets to the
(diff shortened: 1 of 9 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -135,6 +80,28 @@ static void init(
     initializeCacheDirs(context, options);
     options.setEnvelopeDiskCache(new AndroidEnvelopeCache(options));
 
+    readDefaultOptionValues(options, context, buildInfoProvider);
+  }
+
+  @TestOnly
+  static void initializeIntegrationsAndProcessors(
+      @NonNull SentryAndroidOptions options, @NonNull Context context) {
```
romtsn, line 88: can we make the params `final` (and also in the other method overload)

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
```diff
@@ -102,11 +102,24 @@ public static synchronized void init(
...
-                options, context, logger, isFragmentAvailable, isTimberAvailable);
+            final BuildInfoProvider buildInfoProvider = new BuildInfoProvider(logger);
+            final LoadClass loadClass = new LoadClass();
+
+            AndroidOptionsInitializer.loadDefaultAndMetadataOptions(
+                options, context, logger, buildInfoProvider);
+
             configuration.configure(options);
-            deduplicateIntegrations(options, isFragmentAvailable, isTimberAvailable);
             resetEnvelopeCacheIfNeeded(options);
```
romtsn, line 112: I guess we don't need this one anymore, if we set the AndroidEnvelopeCache after the user has configured the options

markushi: In terms of integrations we still do, as both Fragment and Timber integration could be duplicates, and since there are tests for this I know we could still have this situation.
But I agree, since the user can add the integrations first we could only add the Sentry-default ones in case they don't exist, making this method obsolete

romtsn: no, I meant `resetEnvelopeCacheIfNeeded`, the deduplicate method we still need, yes :)

markushi: Oh right, good point 👍

PR #2379, opened 2022-11-19
Fix R8 config

### LINKED ISSUE #2362: Use Class#getCanonicalName to log Fragment name in breadcrumbs
### Problem Statement

Currently, in case of an obfuscated Android app, it is impossible to determine to which `Fragment` ui breadcrumbs from `SentryFragmentLifecycleCallbacks` refer to.

This is because `SentryFragmentLifecycleCallbacks` retrieves the fragment name using the following function

```
private fun getFragmentName(fragment: Fragment): String {
  return fragment.javaClass.simpleName
}
```

`Class#getSimpleName` does not allow us manually deobfuscate `Fragment` names thus making those breadcrumbs basically useless.

### Solution Brainstorm

Replace `Class#getSimpleName` with `Class#getCanonicalName`

```
private fun getFragmentName(fragment: Fragment): String {
  return fragment.javaClass.canonicalName
}
```

romtsn:
## :scroll: Description

While using minified builds I noticed that timber and fragment integrations do not get autoinstalled. Turns out it's because we're checking for presence of upstream dependencies classes (e.g. `timber.log.Timber`), but the get obfuscated too indeed. So this PR will keep their names so we can successfully check them. Along the way fixed another 2 issues:
* Use canonicalName instead of simpleName in Fragment Integration for better de-obfuscation
* Keep `io.sentry` package name to filter ouе our frames from stacktraces when sending the crashes. This adds ~5kb for the final apk size, but I guess this is not a big price.
 
Closes #2362 

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Use `canonicalName` in Fragment Integration for better de-obfuscation ([#2379](https://github.com/getsentry/sentry-java/pull/2379))
+- Fix Timber and Fragment integrations auto-installation for obfuscated builds ([#2379](https://github.com/getsentry/sentry-java/pull/2379))
+
 ## 6.8.0
 
 ### Fixes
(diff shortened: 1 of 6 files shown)


markushi (approved):
Well spotted, nice!

PR #2382, opened 2022-11-21
File I/O on main thread

romtsn:
## :scroll: Description

* Adds necessary data to File I/O spans for creating performance issues out of them
	* Adds `blocked_ui_thread` boolean flag indicating whether the span has run its entire lifespan on the main thread
	* Adds `call_stack` array which contains stack frames leading to the File I/O span for proper grouping of the issues
* Attaches `debug_meta` on the `SentryBaseEvent` level now, to also send it with transactions and not only events.

Q4 Goal
rfc: https://github.com/getsentry/rfcs/pull/36

Manually, automated tests will follow

## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next steps


sentry/src/main/java/io/sentry/util/CollectionUtils.java
```diff
@@ -120,6 +120,23 @@ public static int size(final @NotNull Iterable<?> data) {
...
 
+  /**
+   * Returns a new list which entries match a predicate specified by a parameter.
+   *
+   * @param predicate - the predicate
+   * @return a new list
+   */
+  public static @NotNull <T> List<T> filterListEntries(
+      final @NotNull List<T> list, final @NotNull Predicate<T> predicate) {
+    final List<T> filteredList = new ArrayList<>();
```
markushi, line 131: `l`: given that we usually just just remove a few items it could make sense to init the new list with the size of the original one, like this
```suggestion
    final List<T> filteredList = new ArrayList<>(list.size());
```

sentry/src/main/java/io/sentry/util/thread/MainThreadChecker.java
```diff
@@ -0,0 +1,28 @@
...
+ * Class that checks if a given thread is the Main/UI thread. The Main thread is denoted by the
+ * thread, where the `Sentry.init` method was called. If it was inited from a background thread,
+ * then this class is wrong, but it's a best effort.
+ *
+ * <p>We're gonna educate people through the docs.
+ */
+@ApiStatus.Internal
+public final class MainThreadChecker implements IMainThreadChecker {
+
+  private static final long mainThreadId = Thread.currentThread().getId();
```
markushi, line 15: `m`: I think `mainThreadId` will only be set the first time accessing `MainThreadChecker. getInstance()` thus could be wrong / depending on the first caller.

sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
```diff
@@ -30,10 +34,23 @@ final class FileIOSpanManager {
   FileIOSpanManager(
       final @Nullable ISpan currentSpan,
       final @Nullable File file,
-      final boolean isSendDefaultPii) {
+      final @NotNull SentryOptions options) {
+    this(
+        currentSpan,
+        file,
+        options,
+        new SentryStackTraceFactory(options.getInAppExcludes(), options.getInAppExcludes()));
```
markushi, line 42: `h`: 
```suggestion
        new SentryStackTraceFactory(options.getInAppExcludes(), options.getInAppIncludes()));
```

markushi (changes requested):
Nicely done, please have a look at my comments 🚀

sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java
romtsn: lol, nice catch! will add a test for this

sentry/src/main/java/io/sentry/util/thread/MainThreadChecker.java
romtsn: yep, but we are accessing it from Sentry.init only, so most of the time it should be correct (though not guaranteed that Sentry.init cannot be called from a background thread). There's also a doc comment to this class, check it out :) 

Or do you see a better way of identifying the main thread for Java desktop apps?

markushi: Ah my bad, looks like I skimmed over that comment part. I don't know of any better way, so all good 👍

markushi (approved):
Nice! :shipit:

PR #2390, opened 2022-11-25
Provide automatic breadcrumbs and transactions for click/scroll events for Compose 

markushi:
## :scroll: Description
Provide automatic breadcrumbs and transactions for click/scroll events on Composable UI widgets.

## :pencil: Checklist

- [x] I updated the docs if needed: https://github.com/getsentry/sentry-docs/pull/5893
- [x] No breaking changes


romtsn (commented):
as discussed internally let's try to use our SentryGestureListener for that so we still add value even for those who don't use SAGP for auto-instrumentation

sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/AndroidComposeViewUtils.java
```diff
@@ -0,0 +1,102 @@
...
+import androidx.compose.ui.semantics.SemanticsProperties;
+import io.sentry.util.Objects;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Queue;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+@SuppressWarnings("KotlinInternalInJava")
+final class AndroidComposeViewUtils {
```
markushi, line 20: @romtsn this is the secret sauce for Compose. Once we're happy with what we're doing here I suggest we switch to using reflection or add some safe guards around this.

markushi: Ideally this lives in our compose lib as well - unfortunately only `AndroidComposeView` exposes the root layout node. Maybe it's fine to add a `compileOnly` dependency?

sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/ViewUtils.java
```diff
@@ -58,9 +129,15 @@ final class ViewUtils {
           }
         }
       }
-    }
 
-    return target;
+      if (AndroidComposeViewUtils.isComposeView(view)) {
```
romtsn, line 133: How about we make some abstraction here like 
```
interface TouchTargetFinder {
  UiElement findTarget(...);
}
```
then we could have a list of those in SentryOptions, and register them upon initialization. And then we could iterate through the list (should be ordered) and use the first found target.

Then we could have a default implementation for traditional views in `-core`, and have a `compileOnly` dependecy on `-compose` package as you suggested, and then check at runtime if the compose implementation is present - if yes, we'd add it to the beginning of the list of touchTargetFinders (or whatever the name is). And then we could just iterate through the list in `SentryGestureListener`. How does that sound?

markushi:
@romtsn looking good now from my perspective. A few points I'm not happy about right now:
- As the `GestureTargetLocator` needs to live within `sentry` to make is accessible to both `sentry-compose` and the android module it doesn't know anything about Android or Compose, thus the `.locate()` method takes the "root view" of type `Object` as an argument
- The tests are very basic right now and only check the breadcrumbs, but I really tried to keeps things simple. Also it's using "Espresso" for testing the interactions, as the built-in Compose testing lib is not simulating real touch events

sentry-android-integration-tests/sentry-uitest-android/build.gradle.kts
```diff
@@ -14,7 +14,7 @@ android {
     namespace = "io.sentry.uitest.android"
 
     defaultConfig {
-        minSdk = Config.Android.minSdkVersionNdk
+        minSdk = 21
```
romtsn, line 17: I think we also have `Config.Android.minSdkVersionCompose` for that, perhaps we could use it here

sentry-android-integration-tests/sentry-uitest-android/build.gradle.kts
```diff
@@ -33,6 +33,11 @@ android {
         // Determines whether to support View Binding.
         // Note that the viewBinding.enabled property is now deprecated.
         viewBinding = true
+        compose = true
+    }
+
+    composeOptions {
+        kotlinCompilerExtensionVersion = Config.composeVersion
```
romtsn, line 40: Compose is at `1.3.2` already and we use `1.1.1` - we talked about setting up a CI test matrix already, no need to do it now, but could we at least manually smoke-test if our approach still works with 1.3.2?

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
```diff
@@ -133,6 +130,8 @@ public final class SentryAndroidOptions extends SentryOptions {
 
   private boolean enableFramesTracking = true;
 
+  private final @NotNull List<GestureTargetLocator> gestureTargetLocators = new ArrayList<>();
```
romtsn, line 133: `m`: Just thinking if it should be part of `SentryOptions` instead of `SentryAndroidOptions`, since `GestureTargetLocator` lives there? Also, if we ever do UI event transactions/breadcrumbs for compose-desktop, it also needs to be there

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
```diff
@@ -372,6 +355,15 @@ public void setEnableFramesTracking(boolean enableFramesTracking) {
     this.enableFramesTracking = enableFramesTracking;
   }
 
+  public List<GestureTargetLocator> getGestureTargetLocators() {
+    return gestureTargetLocators;
+  }
+
+  public void setGestureTargetLocators(@NotNull final List<GestureTargetLocator> evaluators) {
```
romtsn, line 362: `l`: maybe rename parameter to `locators`?

sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/AndroidViewGestureTargetLocator.java
```diff
@@ -0,0 +1,80 @@
...
+import android.widget.ScrollView;
+import androidx.core.view.ScrollingView;
+import io.sentry.internal.gestures.GestureTargetLocator;
+import io.sentry.internal.gestures.UiElement;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+@ApiStatus.Internal
+public class AndroidViewGestureTargetLocator implements GestureTargetLocator {
```
romtsn, line 14: `l`: should be `final`

sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/AndroidViewGestureTargetLocator.java
```diff
@@ -0,0 +1,80 @@
...
+public final class AndroidViewGestureTargetLocator implements GestureTargetLocator {
+
+  private final boolean isAndroidXAvailable;
+  private final int[] coordinates = new int[2];
+
+  public AndroidViewGestureTargetLocator(final boolean isAndroidXAvailable) {
+    this.isAndroidXAvailable = isAndroidXAvailable;
+  }
+
+  static boolean isViewTappable(final @NotNull View view) {
```
romtsn, line 23: `l`: Can all these static methods be private now?

sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/AndroidViewGestureTargetLocator.java
```diff
@@ -0,0 +1,80 @@
...
+
+  private static boolean isJetpackScrollingView(
+      final @NotNull View view, final boolean isAndroidXAvailable) {
+    if (!isAndroidXAvailable) {
+      return false;
+    }
+    return ScrollingView.class.isAssignableFrom(view.getClass());
+  }
+
+  static boolean touchWithinBounds(
```
romtsn, line 42: `l`: do we still need to pass the `coords` parameter or can just use `coordinates` directly since it's a class member now?

sentry-android-integration-tests/sentry-uitest-android/build.gradle.kts
markushi: Just tested with `1.3.2` - works fine ✅

sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/AndroidViewGestureTargetLocator.java
```diff
@@ -0,0 +1,80 @@
...
+        return createUiElement(view);
+      } else if (targetType == UiElement.Type.SCROLLABLE
+          && isViewScrollable(view, isAndroidXAvailable)) {
+        return createUiElement(view);
+      }
+    }
+    return null;
+  }
+
+  private UiElement createUiElement(@NotNull View targetView) {
```
romtsn, line 72:
```suggestion
  private UiElement createUiElement(final @NotNull View targetView) {
```

sentry/src/main/java/io/sentry/internal/gestures/UiElement.java
```diff
@@ -0,0 +1,59 @@
+package io.sentry.internal.gestures;
+
+import java.lang.ref.WeakReference;
+import java.util.Objects;
```
romtsn, line 4: `h`: I think we cannot use `java.util.Objects` because it's available on Android API 19+, and our minSdkVersion is still 14. So we should use `io.sentry.util.Objects` and add missing methods there.

Also while typing I just realized that we have used it in other places already, so we should fix that there as well, but not in this PR :)

markushi:
As agreed let's get rid of the reflection parts and use java instead. This requires us to use shadow due to https://youtrack.jetbrains.com/issue/KT-30878.

sentry-compose-helper/src/main/java/io/sentry/compose/gestures/ComposeGestureTargetLocator.java
```diff
@@ -0,0 +1,106 @@
...
+    while (!queue.isEmpty()) {
+      final @Nullable LayoutNode node = queue.poll();
+      if (node == null) {
+        continue;
+      }
+
+      final boolean isPlaced = node.isPlaced();
+      final boolean inBounds = layoutNodeBoundsContain(node, x, y);
+
+      if (node.isPlaced() && layoutNodeBoundsContain(node, x, y)) {
```
romtsn, line 43: should we use the variables from above? :)

sentry-compose-helper/src/main/java/io/sentry/compose/gestures/ComposeGestureTargetLocator.java
```diff
@@ -0,0 +1,106 @@
...
+
+        final List<ModifierInfo> modifiers = node.getModifierInfo();
+        for (ModifierInfo modifierInfo : modifiers) {
+          if (modifierInfo.getModifier() instanceof SemanticsModifier) {
+            final SemanticsModifier semanticsModifierCore =
+                (SemanticsModifier) modifierInfo.getModifier();
+            final SemanticsConfiguration semanticsConfiguration =
+                semanticsModifierCore.getSemanticsConfiguration();
+            for (Map.Entry<? extends SemanticsPropertyKey<?>, ?> entry : semanticsConfiguration) {
+              final @Nullable String key = entry.getKey().getName();
```
romtsn, line 56: I thought we are not allowed to pass nullable values into `switch`? I guess it will throw an NPE if the key is null?

sentry-compose-helper/build.gradle.kts
```diff
@@ -0,0 +1,40 @@
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
+plugins {
+    `java-library`
+    kotlin("jvm")
```
romtsn, line 5: do we still need the kotlin plugin here, even if we don't use it?

sentry-compose-helper/build.gradle.kts
```diff
@@ -0,0 +1,40 @@
...
+    targetCompatibility = JavaVersion.VERSION_1_8
+}
+
+tasks.withType<KotlinCompile>().configureEach {
+    kotlinOptions.jvmTarget = JavaVersion.VERSION_1_8.toString()
+}
+
+dependencies {
+    api(projects.sentry)
+    api(compose.runtime)
```
romtsn, line 23: I guess the compose deps can be `implementation` since we're not exposing compose-related stuff?

romtsn (approved):

romtsn:
great stuff 🚀 I think now it looks more fail-safe and more performant than it was with reflection, etc. Thanks for bearing with back and forths :)

romtsn:
ah, yeah, one more thing I forgot - could you please add a `README.md` to the `sentry-compose-helper` module explaining why do we need in the first place and why it's not published as a separate module? 🙏 thanks

sentry-compose-helper/src/main/java/io/sentry/compose/gestures/ComposeGestureTargetLocator.java
markushi: Oh no, they slipped in for some UI test debugging I did earlier, let me remove them

PR #2407, opened 2022-12-01
Clear window reference only on activity stop in profileMeasurements collector

stefanosiano:
## :scroll: Description
the current window reference in SentryFrameMetricsCollector is cleared only after the activity stops, and no more when the profiler stops

We need the reference to the current window to attach or detach frameMetrics listeners. Currently, when the profiler stops we clear this reference. This means that if the profiler is restarted in the same activity, we cannot attach the frameMetrics listener because we don't have the reference to the window.
So we are having profiles without measurements, especially when auto instrumentation starts (and stops) profiler when an activity is started

Added a unit test

## :pencil: Checklist

- [x] No breaking changes


sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
```diff
@@ -147,9 +150,6 @@ public void stopCollection(final @Nullable String listenerId) {
 
   @SuppressLint("NewApi")
   private void clearCurrentWindow(final @NotNull Window window) {
```
markushi, line 152:
```suggestion
  private void stopFrameMetricsTrackingForWindow(final @NotNull Window window) {
```
`m`: since we're not clearing the window in here anymore, I guess it makes sense to rename this method accordingly, how about `stopFrameMetricsTrackingForWindow`?

markushi (approved):
LGTM, please have a look at my comment

PR #2412, opened 2022-12-02
Updated ProfileMeasurementValue types

stefanosiano:
## :scroll: Description
Changed ProfileMeasurementValue types: timestamp is now a string and value is a double
Added a check against negative timestamps

Relay added a stronger type check on the profile measurements. This caused profiles to be rejected. While the issue has already been addressed in relay, it's better to update the schema to what relay expects.

Updated unit test

## :pencil: Checklist

- [x] No breaking changes


markushi (approved):
Looks good to me! Do we need to migrate other components (e.g. hybrid SDKs) or wait for a backend release (e.g. on relay) before this can be merged?

stefanosiano:
> Looks good to me! Do we need to migrate other components (e.g. hybrid SDKs) or wait for a backend release (e.g. on relay) before this can be merged?

nope: this was the expected format from the beginning, but relay was previously more "permissive" and profiles were ingested.
This change is because relay was updated to check strictly the types, but that check was temporary reverted to allow the sdk to update the code

PR #2422, opened 2022-12-07
Fix Android ui flaky tests checking order of envelopes

stefanosiano:
## :scroll: Description
ui tests now search the envelope among all the ones sent to the mock relay server and make assertion on them, instead of relying on the order the envelopes were created
added androidx test orchestrator to saucelabs configuration with "clearPackageData" option enabled 
updated androidx orchestrator dependency
renamed transactions in tests to be more useful

#skip-changelog

The ui tests were flaky, as sometimes (rarely) the envelopes were not sent in the creation order, causing https://github.com/getsentry/sentry-java/issues/2396

## :pencil: Checklist

- [] I added tests to verify the changes
- [] I updated the docs if needed
- [x] No breaking changes

### DIFF
--- a/.sauce/sentry-uitest-android-benchmark-lite.yml
+++ b/.sauce/sentry-uitest-android-benchmark-lite.yml
@@ -19,6 +19,9 @@ espresso:
 suites:
 
   - name: "Android 11 (api 30)"
+    testOptions:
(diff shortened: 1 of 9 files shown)


adinauer (approved):
🚀

PR #2430, opened 2022-12-15
Fix compose-helper module is exposed and gets published

markushi:
## :scroll: Description
Right now the `compose-helper` gradle module gets published as an maven artifact. That's not what we want, as it's just an interium solution until the mentioned kmp bug mentioned in the readme is fixed.

## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next steps

#skip-changelog


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
```
romtsn, line 5: I think we don't need a changelog for that, because the "broken" version hasn't been published yet, so no user-facing changes. So I'd just add _#skip-changelog_ to the PR description and remove it

sentry-compose/build.gradle.kts
```diff
@@ -159,3 +161,30 @@ dependencies {
 tasks.withType<LibraryAarJarsTask> {
     mainScopeClassFiles.setFrom(embedComposeHelperConfig)
 }
+
+configure<PublishingExtension> {
```
romtsn, line 165:
```suggestion

// we embed the sentry-compose-helper classes to the same .jar above
// so we need to exclude the dependency from the .pom publication and .module metadata
configure<PublishingExtension> {
```

romtsn (approved):

CHANGELOG.md
markushi: Sounds good 👍

PR #2431, opened 2022-12-15
Revert sourcesets config to produce same modules as before

markushi:
## :scroll: Description
This will produce artifacts for `sentry-compose`, `sentry-compose-android`, `sentry-compose-desktop`

## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next steps

#skip-changelog

### DIFF
--- a/sentry-compose/build.gradle.kts
+++ b/sentry-compose/build.gradle.kts
@@ -21,7 +21,7 @@ kotlin {
     android {
         publishLibraryVariants("release")
     }
-    jvm() {
+    jvm("desktop") {
         compilations.all {
             kotlinOptions.jvmTarget = JavaVersion.VERSION_1_8.toString()
         }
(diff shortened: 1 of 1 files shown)


kahest (approved):

sentry-compose/build.gradle.kts
```diff
@@ -42,21 +42,12 @@ kotlin {
...
 
                 implementation(Config.Libs.kotlinStdLib)
-            }
-        }
-
-        val jvmMain by getting {
-            dependencies {
-                api(projects.sentry)
-                implementation(Config.Libs.kotlinStdLib)
                 api(projects.sentryComposeHelper)
```
romtsn, line 45: hm, I'm a bit surprised it's gonna work, because it contains only java files and we include it into kotlin-common-main. My idea was to just publish another artifact (`-jvm.jar`), but if this works, then it's good I guess :)

markushi: I just double checked the artifacts when publishing locally:
- `sentry-compose.jar` is empty [like before](https://mvnrepository.com/artifact/io.sentry/sentry-compose/6.9.2)
- `sentry-compose-android.jar` and `sentry-compose-desktop.jar` both contain the relevant class files, including the ones from `sentry-helper`

I'm still not happy with this solutions, as it feels like too many workarounds stacked on each other - but I would release it for now and check on how this can be improved in the next iteration.

PR #2432, opened 2022-12-15
add time-to-full-display span to activity transactions

stefanosiano:
## :scroll: Description
added ttfd span
added a FullyDrawnReporter into the options
added a new api `Sentry.reportFullyDrawn()`

We want to allow the user to report when the ui is fully loaded for them, e.g. after retrieving some data from some api call and fill their own ui
More info can be found on [this rfc](https://github.com/getsentry/rfcs/pull/46) 

Unit tests

## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next steps
Update docs


sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
```diff
@@ -37,6 +39,9 @@ public final class SentryAndroid {
   private static final String FRAGMENT_CLASS_NAME =
       "androidx.fragment.app.FragmentManager$FragmentLifecycleCallbacks";
 
+  private static final @NotNull FullyDrawnReporter fullyDrawnReporter =
+      FullyDrawnReporter.getInstance();
```
stefanosiano, line 43: I don't really like this one. What would be the best way to use this `FullyDrawnReporter`?
We cannot add it to the hub, as it takes the Activity as parameter and Java doesn't have Android dependencies.
Maybe we can add the `FullyDrawnReporter` to the `SentryAndroidOptions`?

sentry/src/main/java/io/sentry/Sentry.java
```diff
@@ -844,6 +844,18 @@ public static void endSession() {
...
+  /**
+   * Report a screen has been fully loaded. That means all data needed by the UI was loaded. If
+   * time-to-full-display tracing {{@link SentryOptions#isEnableTimeToFullDisplayTracing()} } is
+   * disabled this call is ignored.
+   *
+   * <p>This method is safe to be called multiple times. If the time-to-full-display span is already
+   * finished, this call will be ignored.
+   */
+  public static void reportFullDisplayed() {
+    getCurrentHub().reportFullyDrawn();
```
romtsn, line 856: `m`: should we align the naming, i.e. rename everything from `fullyDrawn` to `fullDisplayed`?

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -399,6 +399,13 @@ public class SentryOptions {
   private @NotNull TransactionPerformanceCollector transactionPerformanceCollector =
       NoOpTransactionPerformanceCollector.getInstance();
 
+  /** Enables the time-to-full-display spans in automatic ui transactions. */
```
romtsn, line 402:
```suggestion
  /** Enables the time-to-full-display spans in activity transactions. */
```
there are multiple occurences of "automatic ui transactions" in this file, so I'd also fix the wording there. (I believe we can omit the "automatic" part as well, because "activity transaction" is already self-descriptive, we always assume they are automatic when talking about them)

sentry/src/main/java/io/sentry/FullyDisplayedReporter.java
```diff
@@ -0,0 +1,42 @@
...
+      final @NotNull Iterator<FullyDrawnReporterListener> listenerIterator = listeners.iterator();
+      while (listenerIterator.hasNext()) {
+        listenerIterator.next().onFullyDrawn();
+        listenerIterator.remove();
+      }
+    }
+  }
+
+  @ApiStatus.Internal
+  public interface FullyDrawnReporterListener {
```
romtsn, line 39: `m`: should we align the naming in this class as well (i.e. `FullDisplayedReporterListener`)?

sentry/src/main/java/io/sentry/FullyDisplayedReporter.java
```diff
@@ -0,0 +1,42 @@
...
+  private final @NotNull List<FullyDrawnReporterListener> listeners = new ArrayList<>();
+
+  private FullyDisplayedReporter() {}
+
+  public static @NotNull FullyDisplayedReporter getInstance() {
+    return instance;
+  }
+
+  public void registerFullyDrawnListener(final @NotNull FullyDrawnReporterListener listener) {
+    synchronized (listeners) {
```
romtsn, line 23: `m`: I think we could use `CopyOnWriteArrayList` and would not need the `synchronized` block. I think it's also cheaper to iterate over the concurrent collection, rather than locking it.

sentry/src/main/java/io/sentry/FullyDisplayedReporter.java
```diff
@@ -0,0 +1,42 @@
+package io.sentry;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+@ApiStatus.Internal
+public final class FullyDisplayedReporter {
```
romtsn, line 10:
```suggestion
public final class FullDisplayedReporter {
```
`l`: for the sake of aligning

sentry-samples/sentry-samples-android/src/main/AndroidManifest.xml
```diff
@@ -114,6 +114,9 @@
         <!-- how to enable the UI auto instrumentation for tracing -->
         <meta-data android:name="io.sentry.traces.user-interaction.enable" android:value="true" />
 
+        <!-- how to enable the time-to-full-display tracing in UI auto instrumentation -->
```
romtsn, line 117:
```suggestion
        <!-- how to enable the time-to-full-display tracing in activity auto instrumentation -->
```

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Features
 
+- Add time-to-full-display span to Activity automatic transactions ([#2432](https://github.com/getsentry/sentry-java/pull/2432))
```
```suggestion
- Add time-to-full-display span to Activity auto-instrumentation ([#2432](https://github.com/getsentry/sentry-java/pull/2432))
```

sentry/src/main/java/io/sentry/Sentry.java
stefanosiano: yeah, i renamed the "external" API and forgot to update the hub 😅

sentry/src/main/java/io/sentry/SentryOptions.java
stefanosiano: The thing is that this is Java, not Android SDK. Is it fine to talk explicitly about Activities?
What about `ui transactions` without automatic?

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/SecondActivity.kt
```diff
@@ -76,6 +76,7 @@ class SecondActivity : AppCompatActivity() {
 
                 showText(true, "error: ${t.message}")
 
+                Sentry.reportFullDisplayed()
                 // I opt out enableActivityLifecycleTracingAutoFinish so I know best when to end my transaction
                 // be sure to finish all your spans before this
                 val transaction = Sentry.getSpan()
```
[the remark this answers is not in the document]
stefanosiano: yeah, that should work, too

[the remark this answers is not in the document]
romtsn: Shall we try to enable it back and remove `Sentry.getSpan.finish()` to showcase the `.reportFullDisplayed` functionality?

romtsn:
So, I quickly checked how the new Activity transactions look, and I've spotted 2 problems there:
image

* First, we should probably align the AppStart span with the TTID span, because TTID shows you the truly correct time it took to display the first frame, as mentioned in the [docs](https://developer.android.com/topic/performance/vitals/launch-time#av). We can tackle this separately (perhaps, we could even deprecate/remove the AppStart span in favor of the TTID span, but that's a separate discussion).
* Second, in my mind the TTFD span should've actually taken longer than the TTID span. I quickly checked the [documentation](https://developer.android.com/reference/android/app/Activity#reportFullyDrawn()) around `reportFullyDrawn`, and they actually cap TTFD to the TTID time, if TTFD takes less time than TTID. From their docs:
> If this method is called before the activity's window is first drawn and displayed as measured by the system, the reported time here will be shifted to the system measured time.

I did some quick tests, and, indeed, if I call `Activity.reportFullyDrawn()` inside `onResume` it shows the same time as TTID (`Displayed` in the logcat):
image


However, If I post an async runnable to call `reportFullyDrawn()` after 2 seconds, it shows correct time after approximately 2 seconds:
image

So, I believe we should align this behavior with how Google does it (and also document it for other SDKs). Sorry for the long-read. 🤣

stefanosiano:
> First, we should probably align the AppStart span with the TTID span, because TTID shows you the truly correct time it took to display the first frame, as mentioned in the [docs](https://developer.android.com/topic/performance/vitals/launch-time#av). We can tackle this separately (perhaps, we could even deprecate/remove the AppStart span in favor of the TTID span, but that's a separate discussion).  

We can think about it for sure

> Second, in my mind the TTFD span should've actually taken longer than the TTID span. I quickly checked the [documentation](https://developer.android.com/reference/android/app/Activity#reportFullyDrawn()) around reportFullyDrawn, and they actually cap TTFD to the TTID time, if TTFD takes less time than TTID. From their docs:

I was thinking the same, as that's what Firebase is doing, too.  
Maybe I should do another pr for that? That would also mean adding a new (internal) API to the span to change its endTimestamp after it's finished

romtsn (approved):

romtsn:
> I was thinking the same, as that's what Firebase is doing, too.
Maybe I should do another pr for that? That would also mean adding a new (internal) API to the span to change its endTimestamp after it's finished

Sounds good, I just approved the PR, feel free to merge and open a new one

PR #2434, opened 2022-12-16
Disable Android concurrent profiling

stefanosiano:
## :scroll: Description
SentryTracer sends profile inside transaction envelope
AndroidTransactionProfiler now profiles only one transaction at a time 
Removed @Deprecated annotations to methods in SentryClient to capture a profile with a transaction 
Added methods in Hub to capture a profile with a transaction

Concurrent profiling create issues with dynamic sampling. We have to send the profile in the same envelope of the transaction to have dynamic sampling work with both of them.
Since we don't have control over the Android profiler internals, we currently cannot adapt the profiler to create a profile for each concurrent transaction, so we are disabling it. 

Unit and ui tests

## :pencil: Checklist

- [x] No breaking changes


stefanosiano:
just realized a possible race condition when starting/finishing a transaction. Gonna make it a draft and fix it

sentry-android-integration-tests/sentry-uitest-android/src/androidTest/java/io/sentry/uitest/android/EnvelopeTests.kt
```diff
@@ -233,7 +145,7 @@ class EnvelopeTests : BaseUiTest() {
                 assertEquals("emptyProfileTransaction", transactionItem.transaction)
             }
             // The profile failed to be sent. Trying to read the envelope from the data transmitted throws an exception
-            assertFailsWith<IllegalArgumentException> { assertFirstEnvelope {} }
+            // assertFailsWith<IllegalArgumentException> { assertFirstEnvelope {} }
```
markushi, line 148: `l` if this is not relevant anymore, I'd remove the line

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
```diff
@@ -242,28 +246,50 @@ public void onFrameMetricCollected(
...
     // onTransactionStart() is only available since Lollipop
     // and SystemClock.elapsedRealtimeNanos() since Jelly Bean
-    if (buildInfoProvider.getSdkInfoVersion() < Build.VERSION_CODES.LOLLIPOP) return;
+    if (buildInfoProvider.getSdkInfoVersion() < Build.VERSION_CODES.LOLLIPOP) return null;
 
-    // Transaction finished, but it's not in the current profile. We can skip it
+    final ProfilingTraceData profilingData = timedOutProfilingData;
+
+    // Transaction finished, but it's not in the current profile
     if (!transactionMap.containsKey(transaction.getEventId().toString())) {
```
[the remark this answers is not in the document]
stefanosiano: we don't need it anymore, so i'm removing it. Thanks for the catch

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
```diff
@@ -229,41 +232,75 @@ public void onFrameMetricCollected(
...
   @Override
-  public synchronized void onTransactionFinish(final @NotNull ITransaction transaction) {
-    options.getExecutorService().submit(() -> onTransactionFinish(transaction, false));
+  public @Nullable synchronized ProfilingTraceData onTransactionFinish(
+      final @NotNull ITransaction transaction) {
+    try {
+      return options
+          .getExecutorService()
+          .submit(() -> onTransactionFinish(transaction, false))
+          .get();
```
romtsn, line 255: hmm, why do we need to use ExecutorService here if we are anyway awaiting for it in a blocking way?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
```diff
@@ -229,41 +232,75 @@ public void onFrameMetricCollected(
...
+    } catch (ExecutionException e) {
+      e.printStackTrace();
```
romtsn, line 257: Can we use `options.getLogger()` here (and in the catch below) with an ERROR or WARNING log level?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
stefanosiano: to avoid calling `onTransactionStart` in the executor service thread and `onTransactionFinish` in a different thread

romtsn (approved):

PR #2436, opened 2022-12-19
Replace java.util.Objects with io.sentry.util.Objects

romtsn:
## :scroll: Description

Since our minSdk is 16 (14 in fact, but nobody is using that already), we cannot use `java.util.Objects` because it's available only from API 19.

## :pencil: Checklist

- [x] No breaking changes


markushi (approved):
LGTM

PR #2440, opened 2022-12-22
Add Android View Hierarchy support

markushi:
## :scroll: Description
Add an event processor which captures the View Hierarchy.

https://github.com/getsentry/team-mobile/issues/64

Some refactoring was required:
Created at `CurrentActivityIntegration` which watches for activity lifecycle updates and updates the `CurrentActivityHolder` singleton - this was previously managed by the `ScreenshotEventProcessor`.
Now both `ScreenshotEventProcessor` and `ViewHierarchyEventProcessor` use the `CurrentActivityHolder` to retrieve the current activity.

Open issues:
- [x] https://github.com/getsentry/relay/issues/1704 needs fixed first before we can use the proper `attachment_type`

Unit tests

## :pencil: Checklist

- [X] No breaking changes


sentry/src/main/java/io/sentry/Attachment.java
```diff
@@ -230,4 +284,31 @@ boolean isAddToTransactions() {
...
     return new Attachment(screenshotBytes, "screenshot.png", "image/png", false);
   }
+
+  /**
+   * Creates a new View Hierarchy Attachment
+   *
+   * @param viewHierarchy the View Hierarchy
+   * @return the Attachment
+   */
+  public static @Nullable Attachment fromViewHierarchy(final ViewHierarchy viewHierarchy) {
```
markushi, line 294: @romtsn I really don't like this part
* this will do the serialization on the main thread
* it's accessing `Sentry.getCurrentHub().getOptions().getSerializer()` which makes it hard to test

Now I'm wondering: Would it make sense to refactor the `Attachment` class and provide an option for having serialize-able payload?
e.g.
````
abstract class Attachment {}

class RawAttachment extends Attachment {
  private final byte[] bytes;
}

class JsonAttachment extends Attachment {
  private final JsonSerializable serializable;
}
````

Comment (romtsn) [reply sentry/src/main/java/io/sentry/Attachment.java:294]: re. options - I think we can just pass them as a parameter from the event processor, right? then it'd be testable.

re. serialization on main thread - I think we can offload it to another thread, but should still block and wait for it, because we're doing it only for errored events and the app is gonna crash anyway? That's what we do for screenshots as well. But I'm wondering how other SDKs did that? @marandaneto @brustolin 

Imo we should block here, because the app may crash, and if we don't wait for it, then we'll lose the attachment

Comment (marandaneto) [reply sentry/src/main/java/io/sentry/Attachment.java:294]: @markushi I do the serialization on the background, you can do the same on Android, attachments already have a callback that is only called during the event sending (and reads the bytes async).
You can check the Dart PR or the Android code.

Comment (markushi) [reply sentry/src/main/java/io/sentry/Attachment.java]: I moved the serialization part to the`ViewHierarchyProcessor`, looking way better now in terms of testability.
re. serialization on main thread: You're right, view hierarchies are only generated when the app crashes, thus it shouldn't be an issue. (On a another note: The Android -> Native scope sync is performing serialization on the main thread as well, maybe that's something we should fix).

File: sentry-android-core/src/main/java/io/sentry/android/core/ViewHierarchyEventProcessor.java [hunk 2]
```diff
@@ -0,0 +1,137 @@
...
+
+    try {
+      final @NotNull ViewHierarchy viewHierarchy = snapshotViewHierarchy(activity);
+      @SuppressWarnings("CharsetObjectCanBeUsed")
+      final Charset UTF8 = Charset.forName("UTF-8");
+
+      try (final ByteArrayOutputStream stream = new ByteArrayOutputStream();
+          final Writer writer = new BufferedWriter(new OutputStreamWriter(stream, UTF8))) {
+
+        options.getSerializer().serialize(viewHierarchy, writer);
```
marandaneto, line 60: This can be done out of the main thread, See this comment https://github.com/getsentry/sentry-java/pull/2440#discussion_r1055409486

marandaneto: M: The charset can be created as a static field. https://github.com/getsentry/sentry-java/blob/f55020d90a282dc59a50203aead122e4d9e2c694/sentry/src/main/java/io/sentry/cache/CacheStrategy.java#L36-L37
So you don't need to create a new one every time.

marandaneto: L: you can also fix it here -> https://github.com/getsentry/sentry-java/blob/f55020d90a282dc59a50203aead122e4d9e2c694/sentry/src/main/java/io/sentry/internal/modules/ModulesLoader.java#L42

sentry-android-core/src/main/java/io/sentry/android/core/ViewHierarchyEventProcessor.java
```diff
@@ -0,0 +1,137 @@
...
+
+    return event;
+  }
+
+  @NotNull
+  public static ViewHierarchy snapshotViewHierarchy(Activity activity) {
+    final List<ViewHierarchyNode> windows = new ArrayList<>();
+    final ViewHierarchy viewHierarchy = new ViewHierarchy("android_view_system", windows);
+
+    final @Nullable View decorView = activity.getWindow().peekDecorView();
```
marandaneto, line 77: M: `activity.getWindow()` can return null, a null check is cheaper than throwing the exception.

sentry-android-core/src/main/java/io/sentry/android/core/ViewHierarchyEventProcessor.java
```diff
@@ -0,0 +1,137 @@
...
+    @Nullable String className = view.getClass().getCanonicalName();
+    if (className == null) {
+      className = view.getClass().getSimpleName();
+    }
+    node.setType(className);
+
+    try {
+      final String identifier = ViewUtils.getResourceId(view);
+      node.setIdentifier(identifier);
+    } catch (Exception e) {
```
marandaneto, line 125:
```suggestion
    } catch (Throwable e) {
```

sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/ViewUtils.java
```diff
@@ -85,7 +86,8 @@ static String getResourceIdWithFallback(final @NotNull View view) {
    * @return human-readable view id
    * @throws Resources.NotFoundException in case the view id was not found
    */
-  static String getResourceId(final @NotNull View view) throws Resources.NotFoundException {
+  @ApiStatus.Internal
```
marandaneto, line 89: L: I'd rather add `@ApiStatus.Internal` to the whole class.

marandaneto: THere's also a utils package btw.

marandaneto:
M: A few methods/params/fields are missing the nullability annotations or `final` when they don't change its values.

sentry-android-core/src/main/java/io/sentry/android/core/ViewHierarchyEventProcessor.java
markushi: I moved the serialization to a background thread now and added a simple time guard (1s) to ensure it doesn't take too long - because otherwise it could potentially trigger an ANR, killing the app and losing the whole crash event.
The snapshotting of the view hierarchy itself is still done on the main thread, as one shouldn't touch UI widgets from a non-ui thread.

sentry/src/main/java/io/sentry/Hint.java
```diff
@@ -117,6 +118,14 @@ public void setScreenshot(@Nullable Attachment screenshot) {
     return screenshot;
   }
 
+  public void setViewHierarchy(@Nullable Attachment viewHierarchy) {
```
romtsn, line 121:
```suggestion
  public void setViewHierarchy(final @Nullable Attachment viewHierarchy) {
```

sentry/src/main/java/io/sentry/protocol/ViewHierarchy.java
```diff
@@ -0,0 +1,108 @@
...
+    public static final String RENDERING_SYSTEM = "rendering_system";
+    public static final String WINDOWS = "windows";
+  }
+
+  private final @Nullable String renderingSystem;
+  private final @Nullable List<ViewHierarchyNode> windows;
+  private @Nullable Map<String, Object> unknown;
+
+  public ViewHierarchy(
+      @Nullable String renderingSystem, @Nullable List<ViewHierarchyNode> windows) {
```
romtsn, line 29:
```suggestion
      final @Nullable String renderingSystem, final @Nullable List<ViewHierarchyNode> windows) {
```

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Add Android View Hierarchy support ([#2440](https://github.com/getsentry/sentry-java/pull/2440))
```
```suggestion
- Attach View Hierarchy to the errored/crashed events ([#2440](https://github.com/getsentry/sentry-java/pull/2440))
```

maybe this gives a little bit more idea what is this all about for those who read

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -209,19 +209,21 @@ private static void installDefaultIntegrations(
...
           new ActivityLifecycleIntegration(
               (Application) context, buildInfoProvider, activityFramesTracker));
+      options.addIntegration(new CurrentActivityIntegration((Application) context));
       options.addIntegration(new UserInteractionIntegration((Application) context, loadClass));
       if (isFragmentAvailable) {
         options.addIntegration(new FragmentLifecycleIntegration((Application) context, true, true));
       }
-      options.addEventProcessor(
-          new ScreenshotEventProcessor((Application) context, options, buildInfoProvider));
+      options.addEventProcessor(new ScreenshotEventProcessor(options, buildInfoProvider));
```
romtsn, line 217: `m`: we could actually move the `ScreenshotEventProcessor` out of this if-condition and add it where we add the view hierarchy one.

romtsn: btw, can we actually move these two from this method, because it's called `installDefaultIntegrations` - maybe we could just move it to the `initializeIntegrationsAndProcessors` method, where we add other event processors?

sentry-android-core/src/main/java/io/sentry/android/core/ViewHierarchyEventProcessor.java
```diff
@@ -0,0 +1,169 @@
...
+      return event;
+    }
+
+    final @Nullable Window window = activity.getWindow();
+    if (window == null) {
+      return event;
+    }
+
+    final @Nullable View decorView = window.peekDecorView();
+    if (decorView == null) {
```
romtsn, line 67: `l`: Not sure if we should add log statements for all of the ifs above? Would be easier to investigate potential issues of why the view hierarchy hasn't been captured by just enabling the debug log

markushi: Good point, I've added some logs - but at an INFO level as it could be completely fine that there's not activity or window active at the moment the crash happened.

sentry/src/main/java/io/sentry/Attachment.java
brustolin: For cocoa, if it's a crash we serialize it in the crash thread, which is not the main. And if is an error capture we serialize in the main thread otherwise there is a bunch of warnings that we are calling UI methods in a background thread.

sentry-android-core/src/main/java/io/sentry/android/core/ViewHierarchyEventProcessor.java
marandaneto: I'd do it differently, see how this example is done:
`final Attachment image = new Attachment(imageFile.getAbsolutePath(), "sentry.png", "image/png");`
The file is read async when its being sent, so you can do the same by doing:
`new Attachment(imageFile.getAbsolutePath())`, you just need to do something similar to `SentryEnvelopeItem`, which has the `CachedItem` callback, so you do the serialization there.
Maybe pair up with @romtsn about that, or we can jump on a call and I show you around, also the Dart PR does exactly that, you can also take a look in there.
The current approach has the overhead of one more thread and still makes it a blocking call.

markushi: Ah alright, yeah that's roughly what I initially [proposed here](https://github.com/getsentry/sentry-java/pull/2440#discussion_r1055320512) to postpone serialization. I've added a `bytesFactory` field to the `Attachment` class now, which gets consumed on demand in `SentryEnvelopeItem / CachedItem`.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -210,19 +210,20 @@ private static void installDefaultIntegrations(
...
```
romtsn, line 218: can we also move the `ScreenshotEventProcessor` btw? It doesn't need to be here anymore I believe

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
```diff
@@ -171,25 +171,20 @@ public static SentryEnvelopeItem fromAttachment(
...
-                      String.format(
-                          "Dropping attachment with filename '%s', because the "
-                              + "size of the passed bytes with %d bytes is bigger "
-                              + "than the maximum allowed attachment size of "
-                              + "%d bytes.",
-                          attachment.getFilename(),
-                          attachment.getBytes().length,
-                          maxAttachmentSize));
-                }
+                ensureAttachmentSizeLimit(
```
romtsn, line 174: `l`: I think we could extract `attachment.getBytes()` into a variable and then reuse it when ensuring the size limit and returning, but it's a micro-optimization

romtsn (approved):
Very nice, I like how straightforward the ViewHierarchyProcessor looks 👍 🚀

sentry-android-core/src/main/java/io/sentry/android/core/ViewHierarchyEventProcessor.java
```diff
@@ -73,39 +71,30 @@ public ViewHierarchyEventProcessor(final @NotNull SentryAndroidOptions options)
...
-
-        final Attachment attachment = Attachment.fromViewHierarchy(stream.toByteArray());
-        hint.setViewHierarchy(attachment);
-      }
-    } catch (Throwable t) {
-      options.getLogger().log(SentryLevel.ERROR, "Could not snapshot ViewHierarchy", t);
-    }
+  private void attachViewHierarchy(@NonNull ViewHierarchy viewHierarchy, @NonNull Hint hint) {
+    hint.setViewHierarchy(
+        Attachment.fromViewHierarchy(
```
marandaneto, line 83: M: Since `Attachment.fromViewHierarchy` is a public API, why don't we allow `Attachment.fromViewHierarchy(viewHierarchy)` instead of the callable? so the user does not need to handle all the serialization bits of a format they don't understand, but rather creating the `ViewHierarchy` object, a much user-friendlier API, that's how it is on Dart.

sentry/src/main/java/io/sentry/Attachment.java
```diff
@@ -282,12 +322,12 @@ boolean isAddToTransactions() {
   /**
    * Creates a new View Hierarchy Attachment
    *
-   * @param viewHierarchyBytes the serialized View Hierarchy
+   * @param bytesFactory the serialized View Hierarchy
    * @return the Attachment
    */
-  public static @NotNull Attachment fromViewHierarchy(final byte[] viewHierarchyBytes) {
+  public static @NotNull Attachment fromViewHierarchy(final Callable<byte[]> bytesFactory) {
```
marandaneto, line 328: M: Since Attachment.fromViewHierarchy is a public API, why don't we allow Attachment.fromViewHierarchy(viewHierarchy) instead of the callable? so the user does not need to handle all the serialization bits of a format they don't understand, but rather creating the ViewHierarchy object, a much user-friendlier API, that's how it is on Dart.

See example:

```
SentryAttachment.fromViewHierarchy(SentryViewHierarchy sentryViewHierarchy)
      : this.fromLoader(
          loader: () => Uint8List.fromList(
              utf8JsonEncoder.convert(sentryViewHierarchy.toJson())),
          filename: 'view-hierarchy.json',
          contentType: 'application/json',
        );
```

markushi: I agree it would be a more user-friendly API, but `Attachment` is just a POJO holding some data. So if we would provide a `ViewHierarchy` as an argument, `Attachment` would need to take care of serializing and thus pull in quite some dependencies, e.g. by calling `Sentry.getCurrentHub().getOptions().getSerializer().serialize(...)`. I aimed to be consistent with the existing `public static @NotNull Attachment fromScreenshot(final byte[] screenshotBytes) ` signature here as well.
But I'd be happy to improve the API on a next iteration for both ViewHierarchy and Screenshots.

marandaneto: @markushi that is just an example, the serialization can happen in the `SentryEnvelopeItem` as it is for many other POJOs.
Doing it later would require a breaking change, the current implementation is different from how it's already being done for other features.
IMO, this would be better addressed now.

sentry-android-core/src/main/java/io/sentry/android/core/ViewHierarchyEventProcessor.java
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    final @Nullable Window window = activity.getWindow();
+    if (window == null) {
+      options.getLogger().log(SentryLevel.INFO, "Missing window for view hierarchy snapshot.");
+      return event;
+    }
+
+    final @Nullable View decorView = window.peekDecorView();
+    options.getLogger().log(SentryLevel.INFO, "Missing decor view for view hierarchy snapshot.");
```
krystofwoldrich, line 65: This log should be inside of the if below. Otherwise, it will be always logged.

markushi: nice catch, thanks!

sentry/src/main/java/io/sentry/Attachment.java
markushi: Alright, now I see - thanks! That's actually pretty much what I initially proposed, but I overlooked the responsibilities of `SentryEnvelopeItem` 😅  I've now delegated the serialization to `SentryEnvelopeItem`, looking better now.

sentry/src/main/java/io/sentry/SentryClient.java
```diff
@@ -278,7 +278,11 @@ private boolean shouldSendSessionUpdateForDroppedEvent(
     if (attachments != null) {
       for (final Attachment attachment : attachments) {
         final SentryEnvelopeItem attachmentItem =
-            SentryEnvelopeItem.fromAttachment(attachment, options.getMaxAttachmentSize());
+            SentryEnvelopeItem.fromAttachment(
```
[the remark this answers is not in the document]
markushi: I tried to stay consistent here with other `SentryEnvelopeItem.fromXYZ()` methods which all take at least the `ISerializer` as a parameter. If this ever gets a public API, I agree - then options should be used instead. I'll keep it like this for now, as it makes it a bit easier to write tests.

krystofwoldrich:
I've tried the snapshotViewHierarchy on RN. It works but I get a lot "invalid id" error logs. Any ideas about what it could cause?

```java
logger.log(SentryLevel.INFO, "snapshot viewHierarchy");
final @NotNull ViewHierarchy viewHierarchy = ViewHierarchyEventProcessor.snapshotViewHierarchy(decorView);
```

```logcat
2023-01-11 12:50:10.687 10492-10550 RNSentry                io.sentry.sample                     I  snapshot viewHierarchy
2023-01-11 12:50:10.688 10492-10550 o.sentry.sample         io.sentry.sample                     E  No package ID ff found for ID 0xffffffff.
2023-01-11 12:50:10.688 10492-10550 o.sentry.sample         io.sentry.sample                     E  No package ID ff found for ID 0xffffffff.
2023-01-11 12:50:10.688 10492-10550 o.sentry.sample         io.sentry.sample                     E  No package ID ff found for ID 0xffffffff.
2023-01-11 12:50:10.689 10492-10550 o.sentry.sample         io.sentry.sample                     E  Invalid ID 0x00000001.
2023-01-11 12:50:10.689 10492-10550 o.sentry.sample         io.sentry.sample                     E  Invalid ID 0x00000107.
2023-01-11 12:50:10.689 10492-10550 o.sentry.sample         io.sentry.sample                     E  Invalid ID 0x00000103.
2023-01-11 12:50:10.689 10492-10550 o.sentry.sample         io.sentry.sample                     E  Invalid ID 0x00000101.
2023-01-11 12:50:10.689 10492-10550 o.sentry.sample         io.sentry.sample                     E  Invalid ID 0x000000ff.
2023-01-11 12:50:10.689 10492-10550 o.sentry.sample         io.sentry.sample                     E  Invalid ID 0x00000003.
2023-01-11 12:50:10.689 10492-10550 o.sentry.sample         io.sentry.sample                     E  Invalid ID 0x000000f9.
2023-01-11 12:50:10.689 10492-10550 o.sentry.sample         io.sentry.sample                     E  Invalid ID 0x000000f7.
2023-01-11 12:50:10.689 10492-10550 o.sentry.sample         io.sentry.sample                     E  Invalid ID 0x000000f5.
2023-01-11 12:50:10.689 10492-10550 o.sentry.sample         io.sentry.sample                     E  Invalid ID 0x000000f3.
2023-01-11 12:50:10.689 10492-10550 o.sentry.sample         io.sentry.sample                     E  Invalid ID 0x000000d1.
2023-01-11 12:50:10.689 10492-10550 o.sentry.sample         io.sentry.sample                     E  Invalid ID 0x000000cf.
2023-01-11 12:50:10.690 10492-10550 o.sentry.sample         io.sentry.sample                     E  Invalid ID 0x00000009.
2023-01-11 12:50:10.690 10492-10550 o.sentry.sample         io.sentry.sample                     E  Invalid ID 0x000000cd.
2023-01-11 12:50:10.690 10492-10550 o.sentry.sample         io.sentry.sample                     E  Invalid ID 0x0000000f.
2023-01-11 12:50:10.690 10492-10550 o.sentry.sample         io.sentry.sample                     E  Invalid ID 0x00000013.
2023-01-11 12:50:10.690 10492-10550 o.sentry.sample         io.sentry.sample                     E  Invalid ID 0x00000019.
2023-01-11 12:50:10.690 10492-10550 o.sentry.sample         io.sentry.sample                     E  Invalid ID 0x00000021.
2023-01-11 12:50:10.690 10492-10550 o.sentry.sample         io.sentry.sample                     E  Invalid ID 0x0000001d.
2023-01-11 12:50:10.690 10492-10550 o.sentry.sample         io.sentry.sample                     E  Invalid ID 0x00000027.
2023-01-11 12:50:10.690 10492-10550 o.sentry.sample         io.sentry.sample                     E  Invalid ID 0x00000025.
2023-01-11 12:50:10.691 10492-10550 o.sentry.sample         io.sentry.sample                     E  Invalid ID 0x000000a3.
2023-01-11 12:50:10.691 10492-10550 o.sentry.sample         io.sentry.sample                     E  Invalid ID 0x0000002f.
2023-01-11 12:50:10.691 10492-10550 o.sentry.sample         io.sentry.sample                     E  Invalid ID 0x0000002d.
```

markushi:
> I've tried the snapshotViewHierarchy on RN. It works but I get a lot "invalid id" error logs. Any ideas about what it could cause?
I think I've found the culprit which was happening when converting view IDs to their resource names (e.g. button_login), [a fix is already push](https://github.com/getsentry/sentry-java/pull/2440/commits/3d22c9f78d1b000b52d623a69864cc75579a861d), could you give it another try?

krystofwoldrich:
> > I've tried the snapshotViewHierarchy on RN. It works but I get a lot "invalid id" error logs. Any ideas about what it could cause?
> 
> @krystofwoldrich I think I've found the culprit which was happening when converting view IDs to their resource names (e.g. button_login), [a fix is already pushed](https://github.com/getsentry/sentry-java/pull/2440/commits/3d22c9f78d1b000b52d623a69864cc75579a861d), could you give it another try?

Thanks. I will try.

krystofwoldrich (approved):
LGTM. Works great with RN.

PR #2445, opened 2022-12-28
Collect memory usage in transactions

stefanosiano:
## :scroll: Description
Added a MemoryCollector, with different implementations for Java and Android, set in SentryOptions
Added TransactionPerformanceCollector, that uses the MemoryCollector, to collect performance data while a transaction is running.
Collected data is a list of objects containing a timestamp, java heap memory usage and native memory usage (only on Android)

We want to build a POC to get the memory usage data as a time series while a transaction runs.

Unit tests

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidMemoryCollector.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidMemoryCollector.java
@@ -1,0 +1,17 @@
+package io.sentry.android.core;
+
+import android.os.Debug;
+import io.sentry.IMemoryCollector;
+import io.sentry.MemoryCollectionData;
+import org.jetbrains.annotations.ApiStatus;
+
+@ApiStatus.Internal
+public class AndroidMemoryCollector implements IMemoryCollector {
+  @Override
+  public MemoryCollectionData collect() {
+    long now = System.currentTimeMillis();
+    long usedMemory = Runtime.getRuntime().totalMemory() - Runtime.getRuntime().freeMemory();
+    long usedNativeMemory = Debug.getNativeHeapSize() - Debug.getNativeHeapFreeSize();
+    return new MemoryCollectionData(now, usedMemory, usedNativeMemory);
+  }
+}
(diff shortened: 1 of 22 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/AndroidMemoryCollector.java
```diff
@@ -0,0 +1,17 @@
...
+import android.os.Debug;
+import io.sentry.IMemoryCollector;
+import io.sentry.MemoryCollectionData;
+import org.jetbrains.annotations.ApiStatus;
+
+@ApiStatus.Internal
+public class AndroidMemoryCollector implements IMemoryCollector {
+  @Override
+  public MemoryCollectionData collect() {
+    long now = System.currentTimeMillis();
```
stefanosiano, line 12: Is `System.currentTimeMillis()` good enough as timestamp? Or should I go with `DateUtils.getCurrentDateTime().getTime()`?

sentry/src/main/java/io/sentry/TransactionPerformanceCollector.java
```diff
@@ -0,0 +1,84 @@
...
+          },
+          0,
+          TRANSACTION_COLLECTION_INTERVAL_MILLIS);
+    }
+  }
+
+  public @Unmodifiable List<MemoryCollectionData> stop(final @NotNull ITransaction transaction) {
+    synchronized (memoryMap) {
+      List<MemoryCollectionData> memoryData = memoryMap.remove(transaction.getEventId().toString());
+      if (memoryMap.isEmpty() && isStarted.get()) {
```
markushi, line 76: `m`: You can utilize `compareAndSet()` to read and update the AtomicBoolean at the same time
```suggestion
      if (memoryMap.isEmpty() && isStarted.compareAndSet(true, false)) {
```

sentry/src/main/java/io/sentry/TransactionPerformanceCollector.java
```diff
@@ -0,0 +1,84 @@
...
+        timer.cancel();
+        timer = new Timer();
+        isStarted.set(false);
```
markushi, line 79:
```suggestion
```

markushi (approved):
Nice one, looks good! Please have a look at my comments

PR #2447, opened 2022-12-30
Send transaction memory stats in profile payload

stefanosiano:
## :scroll: Description
Added list of MemoryCollectionData to Android profiles payload as measurements

We want to add more data to profiles, so we are including the memory statistics we get from transactions to profile payloads. 
Also, at the moment there's no easy way to send the data in the transaction, due to transaction metrics not supporting lists of data. We want to have some data and eventually show it in the Sentry dashboard. The quickest way is to send them in the profiles.

Unit tests

## :pencil: Checklist

- [X] I updated the docs if needed
- [X] No breaking changes


romtsn:
will take a look after the cpu PR is merged, otherwise the diff contains the changes from there as well

sentry-android-core/src/main/java/io/sentry/android/core/AndroidCpuCollector.java
```diff
@@ -0,0 +1,115 @@
...
+    lastCpuNanos = readTotalCpuNanos();
+  }
+
+  @SuppressLint("NewApi")
+  @Override
+  public @Nullable CpuCollectionData collect() {
+    if (buildInfoProvider.getSdkInfoVersion() < Build.VERSION_CODES.LOLLIPOP || !isEnabled) {
+      return null;
+    }
+    long nowNanos = SystemClock.elapsedRealtimeNanos();
```
markushi, line 73: `l`: could be marked as `final`
```suggestion
    final long nowNanos = SystemClock.elapsedRealtimeNanos();
```

sentry-android-core/src/main/java/io/sentry/android/core/AndroidCpuCollector.java
```diff
@@ -0,0 +1,115 @@
...
+    } catch (IOException e) {
+      // If an error occurs when reading the file, we avoid reading it again until the setup method
+      // is called again
+      isEnabled = false;
+      logger.log(
+          SentryLevel.ERROR, "Unable to read /proc/self/stat file. Disabling cpu collection.", e);
+    }
+    if (stat != null) {
+      stat = stat.trim();
+      String[] stats = stat.split("[\n\t\r ]");
```
markushi, line 102: `l`: I'm wondering if we should put this block inside the above `try .. catch` as well? E.g. `parseLong` may throws a `NumberFormatException`

sentry/src/main/java/io/sentry/TransactionPerformanceCollector.java
```diff
@@ -32,64 +30,83 @@ public TransactionPerformanceCollector(final @NotNull SentryOptions options) {
...
+                }
+                synchronized (timerLock) {
+                  for (PerformanceCollectionData data : performanceDataMap.values()) {
+                    data.addData(memoryData, cpuData);
                   }
                 }
               }
             },
-            0,
+            TRANSACTION_COLLECTION_INTERVAL_MILLIS,
```
markushi, line 99: `m`: any reason you're providing a delay parameter here?

markushi (commented):
Nicely done, looking good! I might be missing something but as far as I can see the CPU data is not serialized / added as a measurement to the profile protocol, right? I'm looking at `AndroidTransactionProfiler.putPerformanceCollectionDataInMeasurements`

stefanosiano:
> Nicely done, looking good! I might be missing something but as far as I can see the CPU data is not serialized / added as a measurement to the profile protocol, right? I'm looking at `AndroidTransactionProfiler.putPerformanceCollectionDataInMeasurements`

I'm doing another pr for sending the cpu data, so yeah, they're not added to the profile, yet

sentry/src/main/java/io/sentry/TransactionPerformanceCollector.java
stefanosiano: so we let some time pass between setup() and collect() calls.
This way ICollectors that collect average stats based on time intervals, like AndroidCpuCollector, can have an actual time interval to evaluate.
Just added a comment to clarify it

romtsn (approved):

PR #2459, opened 2023-01-06
Fix lazy initialization causing long startup times

wangbax:
## :scroll: Description

The app needs a sentry to be initialized after the privacy confirmation pop-up box is confirmed, so as to avoid collecting user information in advance.

When using performance monitoring, if the initialization is delayed for some reasons (such as the privacy confirmation popup), the startup time will no longer be accurate (although there is a 60s limit).

1. Open app
2. Show privacy confirmation dialog
3. Wait 30s
4. Click to confirm，init Sentry，jump to MainActivity

## :pencil: Checklist

- [x] No breaking changes

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AppStartState.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AppStartState.java
@@ -1,6 +1,7 @@
 package io.sentry.android.core;
 
 import android.os.SystemClock;
+import io.sentry.DateUtils;
 import java.util.Date;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
@@ -25,6 +26,11 @@ public final class AppStartState {
 
   /** appStart as a Date used in the App's Context */
   private @Nullable Date appStartTime;
+
+  /**
+   * first activity create time
+   */
+  private Long firstActivityCreatedMillis;
 
   private AppStartState() {}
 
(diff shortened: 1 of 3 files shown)


markushi:
Thanks for opening this PR! Could you give us some more details about your use case?
As you mentioned you initialise the Sentry SDK after a privacy confirmation popup.
* Did you manually disable the [SentryPerformanceProvider](https://docs.sentry.io/platforms/android/performance/instrumentation/automatic-instrumentation/#app-start-instrumentation)? It should be running automatically, even if the SDK has not been initialized yet.

wangbax:
> * Did you manually disable the [SentryPerformanceProvider](https://docs.sentry.io/platforms/android/performance/instrumentation/automatic-instrumentation/#app-start-instrumentation)? It should be running automatically, even if the SDK has not been initialized yet.

thank you for your reply. 😊

The detailed steps are:
1. Set `<meta-data android:name="io.sentry.auto-init" android:value="false" />`
2. In Application `onCreate()`, judge whether the user has confirmed that he has show the privacy dialog, if true, initialize Sentry, otherwise skip
3. In SplashActivity (launcher page), judge whether the user has confirmed that he has read the privacy pop-up window. If true, he will enter the HomeActivity (main page). Otherwise, a comfirm dialog will show up. After the confirmation button of the dialog is clicked, Sentry will be initialized
4. Check startup time

If the `SentryPerformanceProvider` is blocked, the startup time of the application can only start from the initialization of Sentry, and the time from Application creation to onCreate() will be lost.

romtsn:
ah, so your issue is that we have a limit of 60 seconds for the app startup time, but in your case it actually can take longer, right? What if we just make that 60 seconds threshold configurable instead? So you could set it to 5 minutes (or whatever you think we'll be the longest time your app is gonna show the confirmation dialog)?

wangbax:
actually, i need statistics **real** `app_start_cold` time.

image

markushi:
@wangbax thanks for clarification! We have found a fix which works across a wider set of edge cases, so I'll close this PR in favor of https://github.com/getsentry/sentry-java/pull/2519. The issues itself is tracked as https://github.com/getsentry/sentry-java/issues/2518, feel free to follow along and comment.
Thanks again for bringing this up! 👍

PR #2462, opened 2023-01-10
Add cpu usage collection

stefanosiano:
## :scroll: Description
added ICpuCollector to collect cpu usage info while transactions are running
TransactionPerformanceCollector now returns PerformanceCollectionData, which contains both memory and cpu usage data

We want to track a time series of cpu percentage usage while a transaction runs, to provide more info to the user.

Unit tests

## :pencil: Checklist

- [x] No breaking changes


sentry/src/main/java/io/sentry/ICpuCollector.java
```diff
@@ -0,0 +1,15 @@
+package io.sentry;
+
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.Nullable;
+
+/** Used for collecting data about cpu load when a transaction is active. */
+@ApiStatus.Internal
+public interface ICpuCollector {
```
romtsn, line 8: `m`: just thinking, what if we make a generic `interface ICollector<T>` instead and make `setup` method for MemoryCollector a no-op? Then we could have a list of `ICollector<T>` in SentryOptions instead of two separate property for memory and cpu (and this is also extendable to other collectors in the future).

sentry-android-core/src/main/java/io/sentry/android/core/AndroidCpuCollector.java
```diff
@@ -0,0 +1,115 @@
...
+  /** Read the /proc/self/stat file and parses the result. */
+  private long readTotalCpuNanos() {
+    String stat = null;
+    try {
+      stat = FileUtils.readText(selfStat);
+    } catch (IOException e) {
+      // If an error occurs when reading the file, we avoid reading it again until the setup method
+      // is called again
+      isEnabled = false;
+      logger.log(
```
romtsn, line 97: `m`: Should we change the log level to `warning` perhaps? It's not some critical functionality of the SDK, so probably fine to only log it when `debug` is enabled

sentry/src/main/java/io/sentry/ICpuCollector.java
stefanosiano: yeah, I thought about it, too
agree, let's do it

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -162,7 +162,10 @@ static void initializeIntegrationsAndProcessors(
       options.setGestureTargetLocators(gestureTargetLocators);
     }
     options.setMainThreadChecker(AndroidMainThreadChecker.getInstance());
-    options.setMemoryCollector(new AndroidMemoryCollector());
+    if (options.getCollectors().isEmpty()) {
```
romtsn, line 165: `l`: just a thought - since this a list, should we just add to it no matter what? Even if the users added their own collectors, we could still add ours? 

Not applicable for `JavaMemoryCollector` though, since it'd be duplicating the values of `AndroidMemoryCollector` kinda

sentry/src/main/java/io/sentry/ICollector.java
```diff
@@ -0,0 +1,14 @@
+package io.sentry;
+
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+/** Used for collecting data about cpu load when a transaction is active. */
```
romtsn, line 6:
```suggestion
/** Used for collecting vitals (memory, cpu, etc.) when a transaction is active. */
```

sentry/src/main/java/io/sentry/ICollector.java
```diff
@@ -0,0 +1,14 @@
...
+@ApiStatus.Internal
+public interface ICollector {
+
+  void setup();
+
+  /** Used for collecting data about cpu load when a transaction is active. */
```
romtsn, line 12:
```suggestion
```

sentry/src/main/java/io/sentry/PerformanceCollectionData.java
```diff
@@ -0,0 +1,53 @@
...
+  public void addCpuData(final @Nullable CpuCollectionData cpuCollectionData) {
+    if (cpuCollectionData != null) {
+      uncommittedCpuData = cpuCollectionData;
+    }
+  }
+
+  /** Save any uncommitted data. */
+  public void commitData() {
+    if (uncommittedMemoryData != null) {
+      memoryData.add(uncommittedMemoryData);
```
romtsn, line 39: `m`: should we nullify `uncommittedMemoryData` and `uncommittedCpuData` after commiting them here? Otherwise if you call `commitData` 2 times in a row with the same data, it'd duplicate?

romtsn (approved):
LGTM 🚀

sentry/src/main/java/io/sentry/TransactionPerformanceCollector.java
```diff
@@ -11,85 +10,83 @@
...
     }
   }
 
-  public @Unmodifiable @Nullable List<MemoryCollectionData> stop(
-      final @NotNull ITransaction transaction) {
+  public @Nullable PerformanceCollectionData stop(final @NotNull ITransaction transaction) {
     synchronized (timerLock) {
-      List<MemoryCollectionData> memoryData = memoryMap.remove(transaction.getEventId().toString());
-      if (memoryMap.isEmpty() && isStarted.getAndSet(false)) {
+      PerformanceCollectionData memoryData =
```
romtsn, line 87:
```suggestion
      PerformanceCollectionData data =
```

romtsn: also needs a change when returning

sentry/src/main/java/io/sentry/PerformanceCollectionData.java
stefanosiano: damn right. Nice catch!
Just added an unit test for it, too

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
stefanosiano: Actually, only the last added Collectors would work, so we'd overwrite user's defined collectors.
Maybe we could set them as first in the list, using something like `collectors.add(0, collector)`
However, at the moment the collector is marked `@Internal`. 
Should we cover this case anyway?

PR #2464, opened 2023-01-10
Use a single transactionPerfomanceCollector

stefanosiano:
## :scroll: Description
the TransactionPerformanceCollector instance is now put in the options, instead of being created in every Hub

In the backend we create a Hub for every thread that access a Hub. 
We were creating an instance of TransactionPerformanceCollector in the Hub, creating an additional thread (since we use a Timer in the collector) per each hub. 
We want to avoid creating too many useless threads, so we are now using a single instance that is saved and retrieved directly from the options

Unit test

## :pencil: Checklist

- [x] No breaking changes


romtsn (approved):
thanks for the fix 👍

PR #2466, opened 2023-01-11
Fix potential Privilege Escalation via Content Provider (CVE-2018-9492)

### LINKED ISSUE #2460: Security Vulnerability: Privilege Escalation via Content Provider (CVE-2018-9492)
### Integration

sentry-android

### Build System

Gradle

### AGP Version

7.3.0

### Proguard

Enabled

### Version

6.6.0

### Steps to Reproduce

1. Integrate Android sentry into an application
2. Scan the app using [Secure theorem](https://www.securetheorem.com/) for security vulnerabilities.

### Expected Result

A security issue should not exist.

### Actual Result

[Secure theorem](https://www.securetheorem.com/) is flagging a security issue "Privilege Escalation via Content Provider" ([CVE-2018-9492](https://www.cvedetails.com/cve/CVE-2018-9492/)).

### More info

We are currently using sentry version [6.6.0](https://github.com/getsentry/sentry-java/releases/tag/6.6.0), which was release on Oct 24, 2022. 

We routinely upgrade our sentry version, so we will be upgrading to the latest; ([6.11.0](https://github.com/getsentry/sentry-java/releases/tag/6.11.0)). However, I doubt that the security issue is fixed in more recent versions, given that the security issue has existed for over 2 years now on our app and I don't see any issues related to this being fixed in the [sentry issues page](https://github.com/getsentry/sentry-java/issues?q=is%3Aissue+). We are just now being forced to resolve this 3rd party security issue due to our company's efforts in increasing app security. Hence, me creating this issue 😅 

### About the security vulnerability

Here are all the _copy-paste_ information we have about the security issue from secure theorem.

Note that we cannot fix the issue by setting our app's `minSdkVersion` to at least 29. Therefore, we need this security issue to be fixed within sentry itself 🙏 

#### AFFECTED CODE

Screenshot 2023-01-09 at 9 55 33 AM

#### DESCRIPTION

Unpatched versions of Android 8 and 9 and vulnerable to [CVE-2018-9492](https://www.cvedetails.com/cve/CVE-2018-9492/): Privilege Escalation Vulnerability in [ActivityManagerService.java](https://android.googlesource.com/platform/frameworks/base/+/962fb40991f15be4f688d960aa00073683ebdd20). This vulnerability allows any third party malicious App to access private data of a victim App without any user interaction.

Any third party App can acquire read/write access to any ContentProvider in the victim App. It doesn't matter if the ContentProvider is exported/non-exported or if the ContentProvider is protected with permissions.

[Security patches](https://source.android.com/security/bulletin/2018-10-01) were released by Google in 2018 to mitigate this vulnerability. Majority of the devices should be patched by now. However, it is possible that some device manufacturers may not have pushed these updates to the users or that some users may have chosen to not install the security updates, leaving a very small percentage of users vulnerable to this issue.

#### RECOMMENDATION

This issue is a vulnerability in the Android OS and can't be exploited on a patched version of Android. However, if you are concerned about unpatched devices, the following options can be used to mitigate the vulnerability:

Option 1: Validate the App requesting for access to the ContentProvider.

Get the calling App's package name via the [getCallingPackage()](https://developer.android.com/reference/android/content/ContentProvider#getCallingPackage()) API. You can use this information to validate that the recipient is allowed to receive the data.

Option 2: Set minSdkVersion to at least 29 for your application.

This vulnerability affects unpatched versions of Android 8 and 9 (API level 26 to 28), by setting the minSdkVersion to 29 your App will install only on devices running Android 10 and above. However, please note that doing so will cause your App not to work on a majority of Android devices.

#### SECURE CODE

```kotlin
class TestProvider extends android.content.ContentProvider {
    @Override
    public boolean onCreate() {
        return false;
    }

    private void securityCheck(Uri uri)  {
        String callingPackage = this.getCallingPackage();
        if(callingPackage != null && (callingPackage.equals("com.trusted.app"))) {
            return;
        }
       SecurityException exception = new SecurityException("Provider does not allow granting of Uri permissions");
       throw exception;
    }

    @Nullable
    @Override
    public Cursor query(@NonNull Uri uri, @Nullable String[] projection, @Nullable String selection, @Nullable String[] selectionArgs, @Nullable String sortOrder) {
        securityCheck(uri);
        return null;
    }

}
```

vestrel00:
## :exclamation: Problem

Apps that use [Secure theorem](https://www.securetheorem.com/) to scan for security vulnerabilities are being flagged for "Privilege Escalation via Content Provider" ([CVE-2018-9492](https://www.cvedetails.com/cve/CVE-2018-9492/)). Devices on Android API 26, 27, and 28 are susceptible to attackers.

The most alarming part of this vulnerability is that it allows attackers to use content providers (even those that do not provide any actual data or handling) to break open the entire system (within or outside the bounds of the application)...

Screen Shot 2023-01-11 at 10 56 00 AM

The affected, insecure code are in the `query` functions in the following content providers...

- [`io.sentry.android.core.SentryInitProvider`](https://github.com/getsentry/sentry-java/blob/6.11.0/sentry-android-core/src/main/java/io/sentry/android/core/SentryInitProvider.java#L48)
- [`io.sentry.android.core.SentryPerformanceProvider`](https://github.com/getsentry/sentry-java/blob/6.11.0/sentry-android-core/src/main/java/io/sentry/android/core/SentryPerformanceProvider.java#L75)

The above content providers do not provide any content but may still be used by an attacker. 

Users of Sentry Android can work around this issue by [removing the content providers](https://docs.sentry.io/platforms/android/configuration/manual-init/) and using manual init. However, this would render apps using auto-init vulnerable.

For more info, initial discussion is at https://github.com/getsentry/sentry-java/issues/2460

## :bulb: Solution

Apply the suggested "secure code" from secure theorem;

```java
class TestProvider extends android.content.ContentProvider {

    private void securityCheck(Uri uri)  {
        String callingPackage = this.getCallingPackage();
        if(callingPackage != null && (callingPackage.equals("com.trusted.app"))) {
            return;
        }
       SecurityException exception = new SecurityException("Provider does not allow granting of Uri permissions");
       throw exception;
    }

    @Nullable
    @Override
    public Cursor query(@NonNull Uri uri, @Nullable String[] projection, @Nullable String selection, @Nullable String[] selectionArgs, @Nullable String sortOrder) {
        securityCheck(uri);
        return null;
    }
}
```

## :green_heart: Testing

I added unit tests for the security checker SRP class. 

I do not actually know how to perform the attack, so I'm not able to verify that this works against attackers. However, it is safe to assume that it will work given that the attacker will have a different package than the app package.

I also played around with the Android sample to make sure that it still works and not just crashing lol.

## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next steps

1. Either merge and include this in the next Sentry release OR create a one-off-tag/snapshot that contains these changes (not released to the general public).
    - Note that Secure Theorem scans our apps from the Google Play Store, which requires signed builds that we can only generate in our remote CI pipelines. Therefore, I will need an actual Sentry release as I am unable to use a locally-built Sentry binary. Otherwise, I would have to spend countless hours trying to figure out how to get the app binaries scanned by Secure Theorem, which I just don't have the time to do.
2. Once a Sentry tag is available with these changes, I will upgrade our apps to that Sentry version and check if Secure Theorem considers the vulnerability to be fixed.
    - Even if the Secure Theorem scanner still considers the vulnerability to be open (maybe because it expects 100% copy-paste of their suggested code), we know for sure that the vulnerability is fixed and we can close #2460 as 100% "false positive".
3. If a snapshot was used in step 1, then we should make sure to include these changes in a Sentry release that is available to the general public.


sentry-android-core/src/test/java/io/sentry/android/core/internal/util/PrivilegeEscalationViaContentProviderCheckerTest.kt
```diff
@@ -0,0 +1,105 @@
...
+import android.os.Build
+import io.sentry.android.core.BuildInfoProvider
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.verifyNoInteractions
+import org.mockito.kotlin.whenever
+import kotlin.test.Test
+import kotlin.test.assertNotNull
+import kotlin.test.assertNull
+
+class PrivilegeEscalationViaContentProviderCheckerTest {
```
vestrel00, line 14: I used one of the existing tests to make sure I'm following the code style/convention.

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/PrivilegeEscalationViaContentProviderChecker.java
```diff
@@ -0,0 +1,63 @@
...
+import io.sentry.android.core.BuildInfoProvider;
+
+/**
+ * Protects against "Privilege Escalation via Content Provider" (CVE-2018-9492).
+ * <p>
+ * See https://www.cvedetails.com/cve/CVE-2018-9492/
+ * and https://github.com/getsentry/sentry-java/issues/2460
+ */
+@ApiStatus.Internal
+public final class PrivilegeEscalationViaContentProviderChecker {
```
vestrel00, line 21: I created this SRP class to make it testable. Hopefully, Secure Theorem will recognize that the vulnerability is fixed even though I did not 100% copy-paste their suggested code.

sentry-android-core/src/test/java/io/sentry/android/core/internal/util/PrivilegeEscalationViaContentProviderCheckerTest.kt
```diff
@@ -0,0 +1,105 @@
...
+            whenever(buildInfoProvider.sdkInfoVersion).thenReturn(sdkVersion)
+
+            return PrivilegeEscalationViaContentProviderChecker(buildInfoProvider)
+        }
+    }
+
+    private val fixture = Fixture()
+
+    @Test
+    fun `When sdk version is less than vulnerable versions, security check is not performed`() {
```
artour-bakiev, line 31: Minor stuff, it appears the Sentry team [ is utilising](https://github.com/getsentry/sentry-java/blob/1e1ab7f5a4a1586225c60ae343015e012649ef26/sentry-android-core/src/test/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollectorTest.kt#L91) lowercase symbols in their test methods.

vestrel00: I think they use both upper and lowercase then because I copy-pasted-and-modified their existing `RootCheckterTest.kt` file for this; https://github.com/getsentry/sentry-java/blob/6.11.0/sentry-android-core/src/test/java/io/sentry/android/core/internal/util/RootCheckerTest.kt#L48

All tests in that file are in the format "When ..., ...."

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/PrivilegeEscalationViaContentProviderChecker.java
```diff
@@ -0,0 +1,63 @@
...
+
+  private final @NotNull BuildInfoProvider buildInfoProvider;
+
+  public PrivilegeEscalationViaContentProviderChecker() {
+    this(new BuildInfoProvider(NoOpLogger.getInstance()));
+  }
+
+  public PrivilegeEscalationViaContentProviderChecker(
+    final @NotNull BuildInfoProvider buildInfoProvider
```
artour-bakiev, line 30: Minor stuff, `final` specified for a parameter.

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/PrivilegeEscalationViaContentProviderChecker.java
artour-bakiev: Does it make sense to fetch a snapshot into the product and run the analyser against it?

vestrel00: @artour-bakiev, I think this is what I'm suggesting in the **Next steps** in the PR description. Or do you mean something else?

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/PrivilegeEscalationViaContentProviderChecker.java
vestrel00: I was just following the style they had been using in `RootChecker.kt`. 

I copy-pasted-and-modified their existing `RootCheckter.kt` file for this; https://github.com/getsentry/sentry-java/blob/6.6.0/sentry-android-core/src/main/java/io/sentry/android/core/internal/util/RootChecker.java#L36

Parameters in that file are all final.

-----

Please keep in mind that this code is in Java where parameters are not final and can be reassigned.

java
 
In Kotlin, parameters are implicitly final (val) and cannot be reassigned.

kotlin

I'm guessing that this is why the Sentry team is prepending parameters in Java with `final`. It is good practice to make sure something is immutable when it does not need to be mutable. Makes everything safer. Kotlin just does these things for us and sometimes we forget that we have to do it ourselves in Java 😁

sentry-android-core/src/main/java/io/sentry/android/core/SentryInitProvider.java
```diff
@@ -51,6 +53,7 @@ public void attachInfo(@NotNull Context context, @NotNull ProviderInfo info) {
       @Nullable String s,
       @Nullable String[] strings1,
       @Nullable String s1) {
+    new PrivilegeEscalationViaContentProviderChecker().securityCheck(this);
```
artour-bakiev, line 56: It seems we should guard every public (`ContentProvider`) method.

vestrel00: I think that the vulnerability is specific to read operations though right? This means we only need to have this in the `query` method as suggested by Secure Theorem.

I'd like to avoid unnecessary code additions / changes.

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/PrivilegeEscalationViaContentProviderChecker.java
```diff
@@ -0,0 +1,63 @@
...
+   * This should be invoked regardless of whether there is data to query or not. The attack is
+   * not contained to the specific provider but rather the entire system.
+   * <p>
+   * This blocks the attacker by only allowing the app itself (not other apps) to "query" the
+   * provider.
+   * <p>
+   * The vulnerability is specific to un-patched versions of Android 8 and 9 (API 26 to 28).
+   * Therefore, this security check is limited to those versions to mitigate risk of regression.
+   */
+  @TargetApi(Build.VERSION_CODES.KITKAT) // Required for ContentProvider#getCallingPackage()
```
artour-bakiev, line 50: Minor, but it seems a bit confusing. It looks like we can safely remove `@TargetApi` annotation since `getCallingPackage()` is only called if the device's SDK version is `>= 0`.

vestrel00: No because I am using `buildInfoProvider.getSdkInfoVersion()` to retrieve the sdkInt. The compiler or lint will complain because we are not using `Build.VERSION.SDK_INT` directly here.

You can try removing `@TargetApi` and you will see that you will get a compile or lint error.

I am using `buildInfoProvider.getSdkInfoVersion()` because it seems to be used throughout the codebase AND it allows me to mock the sdkInt in the unit tests.

marandaneto: Since the method is going to be called, you can ignore the linter with the Supress annotation.

sentry-android-core/src/test/java/io/sentry/android/core/internal/util/PrivilegeEscalationViaContentProviderCheckerTest.kt
```diff
@@ -0,0 +1,105 @@
...
+        contentProvider.mockPackages(null)
+
+        var securityException: SecurityException? = null
+        try {
+            fixture.getSut().securityCheck(contentProvider)
+        } catch (se: SecurityException) {
+            securityException = se
+        }
+
+        assertNotNull(securityException)
```
marandaneto, line 61: Kotlin test has an `assertThrows<T> {}` function.

sentry-android-core/src/test/java/io/sentry/android/core/internal/util/PrivilegeEscalationViaContentProviderCheckerTest.kt
```diff
@@ -0,0 +1,105 @@
...
+        verifyNoInteractions(contentProvider)
+    }
+
+    @Test
+    fun `When calling package is null, security check exception is thrown`() {
+        val contentProvider = mock<ContentProvider>()
+
+        contentProvider.mockPackages(null)
+
+        var securityException: SecurityException? = null
```
markushi, line 54: I think you can simplify this by using assertThrows
```suggestion
        assertThrows<SecurityException> { fixture.getSut().securityCheck(contentProvider) }
```

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/PrivilegeEscalationViaContentProviderChecker.java
markushi, line 21: It could make sense to give this utility class a more generic name, so we could extends it's functionality in case there will be future CVEs.
```suggestion
public final class ContentProviderSecurityChecker {
```

markushi (commented):
Looks good to me, thanks for your contribution!
Please have a look at my comments.
Regarding testability: I think we should be able to provide you with a public snapshot once the code is merged, so you can create builds in your CI and have them verified.

sentry-android-core/src/main/java/io/sentry/android/core/SentryInitProvider.java
vestrel00: @markushi, it's not necessary to override the other can-be-overloaded `query` functions because they all ultimately just call this abstract function subclass implementation.

only-need-to-check-the-abstract-query-function

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/PrivilegeEscalationViaContentProviderChecker.java
vestrel00: I would typically no here and drop the ol' [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it). I don't see future CVEs popping up specifically for ContentProviders when we've been scanning our apps with Secure Theorem for the past 2+ years and this is the only complaint we got so far.

However, I do like the generic name you suggested more in terms of length and how less intimidating it is 😃 

Sure! I'll rename to `ContentProviderSecurityChecker` !

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/PrivilegeEscalationViaContentProviderChecker.java
vestrel00: Good point!!! This method can be called even on prior versions to KitKat since I'm not doing any version checks at the call sites.

This is embarrassing 😓 I'll suppress instead!

Nice catch, thanks!

sentry-android-core/src/test/java/io/sentry/android/core/internal/util/PrivilegeEscalationViaContentProviderCheckerTest.kt
vestrel00: Great callout! I'm an idiot 🤣 I'll change 🔥

sentry-android-core/src/test/java/io/sentry/android/core/internal/util/PrivilegeEscalationViaContentProviderCheckerTest.kt
vestrel00: Yep! @markushi pointed out the same thing https://github.com/getsentry/sentry-java/pull/2466/files#r1069057680

sentry-android-core/src/test/java/io/sentry/android/core/internal/util/PrivilegeEscalationViaContentProviderCheckerTest.kt
vestrel00: Hmm... Wait a sec... I know I initially tried `assertThrows` but no code suggestion was popping up for it. 

Looking for all instances in java or Kotlin files...

Screen Shot 2023-01-13 at 7 22 20 AM

Seems like it requires a dependency currently only used in `sentry-spring` and `sentry-spring-jakarta`.

Screen Shot 2023-01-13 at 7 25 39 AM

I'd like to avoid changing any build configurations in this PR and since I'm only using this `try-catch` approach in two places in this one file, maybe we can let this slide? If you feel strongly about using `assertThrows`, you can make changes after the PR has been merged 🙏

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/PrivilegeEscalationViaContentProviderChecker.java
vestrel00: Done! https://github.com/getsentry/sentry-java/pull/2466/commits/d8272e20ec841e7621ad14bdbdfa5cd502042d7b

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/PrivilegeEscalationViaContentProviderChecker.java
vestrel00: Done! https://github.com/getsentry/sentry-java/pull/2466/commits/17521025dd027c24721d518bcbbd3dab7f6bdaf6

vestrel00:
@markushi @marandaneto, I finished addressing all comments. Please let me know what else I need to do 🙏

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/PrivilegeEscalationViaContentProviderChecker.java
artour-bakiev: I thought about running the datatheorem check **before** landing the PR into main branch. But perhaps it was intended, sorry if it was :)

sentry-android-core/src/main/java/io/sentry/android/core/SentryInitProvider.java
artour-bakiev: > I think that the vulnerability is specific to read operations though right?

Mmmm, I'm not completely sure since I still can't understand the attack. But according to https://www.cvedetails.com/cve/CVE-2018-9492/ the issue happens because:
> In checkGrantUriPermissionLocked of ActivityManagerService.java, there is a possible permissions bypass

It may mean that any public ContentProvider method potentially has the vulnerability (because the `checkGrantUriPermissionLocked` should happen every time the user addresses a `ContentProvider`).

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/PrivilegeEscalationViaContentProviderChecker.java
artour-bakiev: All good then, mate. A `final` param seems to be a bit confusing but it's perhaps just for me.

sentry-android-core/src/main/java/io/sentry/android/core/SentryInitProvider.java
vestrel00: @artour-bakiev, the query function is the only one the Secure Theorem is suggesting we fix. If Secure Theorem thinks we should fix any interaction with the `ContentProvider`, then they would probably suggest to add the check in `onCreate` instead to cover everything.

Let's keep this check in just the `query` function. It should still cover the vast majority of attackers, if not all.

sentry-android-core/src/test/java/io/sentry/android/core/internal/util/ContentProviderSecurityCheckerTest.kt
```diff
@@ -0,0 +1,107 @@
+package io.sentry.android.core.internal.util
+
+import android.content.ContentProvider
+import android.content.Context
+import android.os.Build
+import io.sentry.android.core.BuildInfoProvider
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.verifyNoInteractions
+import org.mockito.kotlin.whenever
+import kotlin.test.Test
```
markushi, line 10:
```suggestion
import kotlin.test.Test
import kotlin.test.assertFailsWith
```

sentry-android-core/src/test/java/io/sentry/android/core/internal/util/ContentProviderSecurityCheckerTest.kt
```diff
@@ -0,0 +1,107 @@
...
+        contentProvider.mockPackages(null)
+
+        var securityException: SecurityException? = null
+        try {
+            fixture.getSut().checkPrivilegeEscalation(contentProvider)
+        } catch (se: SecurityException) {
+            securityException = se
+        }
+
+        assertNotNull(securityException)
```
markushi, line 63:
```suggestion
        assertFailsWith<SecurityException> {
            fixture.getSut().checkPrivilegeEscalation(contentProvider)
        }
```

sentry-android-core/src/test/java/io/sentry/android/core/internal/util/ContentProviderSecurityCheckerTest.kt
```diff
@@ -0,0 +1,107 @@
...
+        contentProvider.mockPackages("{$APP_PACKAGE}.attacker")
+
+        var securityException: SecurityException? = null
+        try {
+            fixture.getSut().checkPrivilegeEscalation(contentProvider)
+        } catch (se: SecurityException) {
+            securityException = se
+        }
+
+        assertNotNull(securityException)
```
markushi, line 79:
```suggestion
        assertFailsWith<SecurityException> {
            fixture.getSut().checkPrivilegeEscalation(contentProvider)
        }
```

markushi (commented):
Thanks again for iterating on this PR! Regarding the exception handling we're already using `assertFailsWith<>()` from the kotlin test library, which is already a dependency and should work out of the box (as opposed to `assertThrows()`).

As a last step please run `make format` in a terminal on the project root folder to ensure all code is formatted.

sentry-android-core/src/test/java/io/sentry/android/core/internal/util/ContentProviderSecurityCheckerTest.kt
vestrel00: @markushi, ahh! This is what I've been looking for! Thanks!

I will use `assertFailsWith` 🔥

sentry-android-core/src/test/java/io/sentry/android/core/internal/util/ContentProviderSecurityCheckerTest.kt
vestrel00: Done! https://github.com/getsentry/sentry-java/pull/2466/commits/b7a40d7c982af216508034960545e7402ec6b3f5

sentry-android-core/src/test/java/io/sentry/android/core/internal/util/ContentProviderSecurityCheckerTest.kt
vestrel00: Done! https://github.com/getsentry/sentry-java/pull/2466/commits/b7a40d7c982af216508034960545e7402ec6b3f5

sentry-android-core/src/test/java/io/sentry/android/core/internal/util/ContentProviderSecurityCheckerTest.kt
vestrel00: Done! https://github.com/getsentry/sentry-java/pull/2466/commits/b7a40d7c982af216508034960545e7402ec6b3f5

vestrel00:
Okay, I addressed all PR comments again. I also ran `make format` so that spotless can format my code properly. Plus, added the fi to the CHANGELOG because I saw a build job was failing because of it.

Please review again and hopefully merge if all builds pass 🤞 🙏

sentry-android-core/src/main/java/io/sentry/android/core/SentryInitProvider.java
```diff
@@ -51,6 +52,7 @@ public void attachInfo(@NotNull Context context, @NotNull ProviderInfo info) {
       @Nullable String s,
       @Nullable String[] strings1,
       @Nullable String s1) {
+    new ContentProviderSecurityChecker().checkPrivilegeEscalation(this);
```
romtsn, line 55: nit: could we just create an instance and keep it as a class member instead of creating a new instance of the SecurityChecker every time the method is called? Same for `SentryPerformanceProvider`.

romtsn (approved):

sentry-android-core/src/main/java/io/sentry/android/core/SentryInitProvider.java
vestrel00: We could. It's what I would have done if the call-sites were using dependency injection or if we had to use an instance of this class in at least one other function. It's not used anywhere else, nor is this function really invoked (it's only invoke if an attacker tries the exploit) so performance shouldn't be an issue.

I was originally writing a static function but ended up with an SRP class instead for testability and future-proofing in case it needs to be mocked at call-sites.

markushi (approved):

sentry-android-core/src/main/java/io/sentry/android/core/SentryInitProvider.java
artour-bakiev: The datatheorem's recommendation seems to be straight forward:
> Option 1: Validate the App requesting for access to the ContentProvider.

Which seems to be any access to the ContentProvider including `update`, `insert` and `update` methods.

vestrel00: @artour-bakiev, sorry I don't mean to ignore your warnings. From my experience, Data/Secure Theorem typically provides suggested "secure code" samples that are complete. So I'm thinking that guarding query should be enough because that's what they suggested.

artour-bakiev: No worries, mate. Let's have a look at the dt scan with v 6.12.0.

vestrel00: If the DT scan still flags the issue, I will update the Sentry implementation to guard `insert`, `update`, and `delete` functions too.

We might have to wait 2-4 weeks for the next Sentry release but I don't think there is a rush at this point 😄

vestrel00: Okay, these changes did not pass the Secure/Data Theorem scan...

However, this does not mean that the vulnerability was not actually fixed. Anyways, as agreed upon, I will create another PR to apply the security checks in `insert`, `update`, and `delete` functions.

If it still does not pass the scan, then there is nothing else we can do. It just means that the scanner wants an exact copy-paste of their suggested code 🤷

vestrel00: Created another PR; https://github.com/getsentry/sentry-java/pull/2482

PR #2473, opened 2023-01-16
Update benchmark cpu approach

stefanosiano:
## :scroll: Description
Cpu usage in benchmark has now been converted by reading the /proc/self/stat/ file.
Updated devices used in Saucelabs for benchmarking.
Decreased acceptance threshold from 5.5% to 5%

#skip-changelog

Previous approach was reading global cpu time to make comparisons.
New approach reads /proc/self/stat/ file, which returns the cpu time spent by the application only, giving more accurate results.
Also, the `sustainedPerformanceMode` has been enabled in the benchmarks, which prevents thermal throttling, increasing accuracy even more.

## :pencil: Checklist

- [x] No breaking changes


sentry-android-integration-tests/sentry-uitest-android-benchmark/src/androidTest/java/io/sentry/uitest/android/benchmark/util/BenchmarkComparisonResult.kt
```diff
@@ -114,7 +114,7 @@ internal data class BenchmarkSummaryResult(
      * Increase of cpu time in milliseconds.
```
romtsn, line 114:
```suggestion
     * Increase of cpu time in nanoseconds.
```

romtsn (approved):
Great stuff, especially the sustainedPerformanceMode flag 🚀

PR #2482, opened 2023-01-18
Expand fix for potential Privilege Escalation via Content Provider (CVE-2018-9492)

vestrel00:
## :exclamation: Problem

The changes contained in https://github.com/getsentry/sentry-java/pull/2466 that were merged and released in [6.12.0](https://github.com/getsentry/sentry-java/releases/tag/6.12.0) were not enough to pass [Data/Secure theorem](https://www.securetheorem.com/) security scans.

This is either due to...

- The security scans expecting exact copy-paste of their suggested code 
- OR it expects the security checks to also be applied to `insert`, `update`, and `delete` ContentProvider functions
    - The security check was, and is currently, only applied to the `query` function.

## :bulb: Solution

In order to completely fix the **potential** security vulnerability described in https://github.com/getsentry/sentry-java/issues/2460, we should also apply the security checks to the `insert`, `update`, and `delete` functions of `SentryInitProvider` and `SentryPerformanceProvider`.

This may not result in passing [Data/Secure theorem](https://www.securetheorem.com/) security scans in case it is expecting copy-paste code of the suggest code. However, we will be 100% sure that we are protected no matter what scanners say 😎 

Thanks to my colleague @artour-bakiev for being [diligent in making sure that I do the right thing](https://github.com/getsentry/sentry-java/pull/2466#discussion_r1071721924) for everyone ❤️ 

I smoke tested the Android sample to make sure that it still works and is not just crashing (lol).

I do not actually know how to perform the attack, so I'm not able to verify that this works against attackers. However, it is safe to assume that it will work given that the attacker will have a different package than the app package.

## :pencil: Checklist

- [x] No breaking changes

## :crystal_ball: Next steps

There are no next steps after this. If the next Sentry version that contains these changes still do no pass the [Data/Secure theorem](https://www.securetheorem.com/) security scans, then we know that they are expecting copy-paste of their suggested code, which we are just not going to do 😁 

Note that there is no urgency in getting this change out in a new Sentry version. We can wait a whole month. No rush. Whenever the next Sentry version is released is kewl!

### DIFF
--- a/sentry-android-core/api/sentry-android-core.api
+++ b/sentry-android-core/api/sentry-android-core.api
@@ -189,24 +189,18 @@ public final class io/sentry/android/core/SentryAndroidOptions : io/sentry/Sentr
 	public fun setProfilingTracesIntervalMillis (I)V
 }
 
-public final class io/sentry/android/core/SentryInitProvider : android/content/ContentProvider {
+public final class io/sentry/android/core/SentryInitProvider {
 	public fun <init> ()V
 	public fun attachInfo (Landroid/content/Context;Landroid/content/pm/ProviderInfo;)V
-	public fun delete (Landroid/net/Uri;Ljava/lang/String;[Ljava/lang/String;)I
 	public fun getType (Landroid/net/Uri;)Ljava/lang/String;
-	public fun insert (Landroid/net/Uri;Landroid/content/ContentValues;)Landroid/net/Uri;
 	public fun onCreate ()Z
-	public fun query (Landroid/net/Uri;[Ljava/lang/String;Ljava/lang/String;[Ljava/lang/String;Ljava/lang/String;)Landroid/database/Cursor;
 	public fun shutdown ()V
-	public fun update (Landroid/net/Uri;Landroid/content/ContentValues;Ljava/lang/String;[Ljava/lang/String;)I
 }
 
-public final class io/sentry/android/core/SentryPerformanceProvider : android/content/ContentProvider, android/app/Application$ActivityLifecycleCallbacks {
+public final class io/sentry/android/core/SentryPerformanceProvider : android/app/Application$ActivityLifecycleCallbacks {
 	public fun <init> ()V
 	public fun attachInfo (Landroid/content/Context;Landroid/content/pm/ProviderInfo;)V
-	public fun delete (Landroid/net/Uri;Ljava/lang/String;[Ljava/lang/String;)I
 	public fun getType (Landroid/net/Uri;)Ljava/lang/String;
-	public fun insert (Landroid/net/Uri;Landroid/content/ContentValues;)Landroid/net/Uri;
 	public fun onActivityCreated (Landroid/app/Activity;Landroid/os/Bundle;)V
 	public fun onActivityDestroyed (Landroid/app/Activity;)V
 	public fun onActivityPaused (Landroid/app/Activity;)V
@@ -215,8 +209,6 @@ public final class io/sentry/android/core/SentryPerformanceProvider : android/co
 	public fun onActivityStarted (Landroid/app/Activity;)V
 	public fun onActivityStopped (Landroid/app/Activity;)V
 	public fun onCreate ()Z
-	public fun query (Landroid/net/Uri;[Ljava/lang/String;Ljava/lang/String;[Ljava/lang/String;Ljava/lang/String;)Landroid/database/Cursor;
-	public fun update (Landroid/net/Uri;Landroid/content/ContentValues;Ljava/lang/String;[Ljava/lang/String;)I
 }
 
 public final class io/sentry/android/core/SystemEventsBreadcrumbsIntegration : io/sentry/Integration, java/io/Closeable {
--- a/sentry-android-core/src/main/java/io/sentry/android/core/EmptySecureContentProvider.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/EmptySecureContentProvider.java
@@ -1,0 +1,56 @@
+package io.sentry.android.core;
+
+import android.content.ContentProvider;
+import android.content.ContentValues;
+import android.database.Cursor;
+import android.net.Uri;
+import io.sentry.android.core.internal.util.ContentProviderSecurityChecker;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * A ContentProvider that does NOT store or provide any data for read or write operations.
+ *
+ * <p>This does not allow for overriding the abstract query, insert, update, and delete operations
+ * of the {@link ContentProvider}. Additionally, those functions are secure.
+ */
+@ApiStatus.Internal
+abstract class EmptySecureContentProvider extends ContentProvider {
+
+  private final ContentProviderSecurityChecker securityChecker =
+      new ContentProviderSecurityChecker();
+
+  @Override
+  public final @Nullable Cursor query(
+      @NotNull Uri uri,
+      @Nullable String[] strings,
+      @Nullable String s,
+      @Nullable String[] strings1,
+      @Nullable String s1) {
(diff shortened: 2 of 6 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/EmptySecureContentProvider.java
```diff
@@ -0,0 +1,56 @@
...
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * A ContentProvider that does NOT store or provide any data for read or write operations.
+ *
+ * <p>This does not allow for overriding the abstract query, insert, update, and delete operations
+ * of the {@link ContentProvider}. Additionally, those functions are secure.
+ */
+@ApiStatus.Internal
+abstract class EmptySecureContentProvider extends ContentProvider {
```
vestrel00, line 19: This PR comes with this additional benefit of making it clear to contributors and consumers that Sentry does not actually have any "real" ContentProviders =)

sentry-android-core/src/main/java/io/sentry/android/core/EmptySecureContentProvider.java
```diff
@@ -0,0 +1,56 @@
...
+      new ContentProviderSecurityChecker();
+
+  @Override
+  public final @Nullable Cursor query(
+      @NotNull Uri uri,
+      @Nullable String[] strings,
+      @Nullable String s,
+      @Nullable String[] strings1,
+      @Nullable String s1) {
+    securityChecker.checkPrivilegeEscalation(this);
```
vestrel00, line 31: Note that there is no need to add this check to to other overridable `query` functions because they all ultimately use this function.

See my comment in my previous PR; https://github.com/getsentry/sentry-java/pull/2466#discussion_r1069676327

sentry-android-core/src/main/java/io/sentry/android/core/EmptySecureContentProvider.java
```diff
@@ -0,0 +1,56 @@
...
+ * of the {@link ContentProvider}. Additionally, those functions are secure.
+ */
+@ApiStatus.Internal
+abstract class EmptySecureContentProvider extends ContentProvider {
+
+  private final ContentProviderSecurityChecker securityChecker =
+      new ContentProviderSecurityChecker();
+
+  @Override
+  public final @Nullable Cursor query(
```
vestrel00, line 25: Added the `final` modifier here to make sure that this cannot be overridden further by subclasses. This ensures that the contract described in the class name and class doc is upheld at the compiler level!

sentry-android-core/src/main/java/io/sentry/android/core/EmptySecureContentProvider.java
```diff
@@ -0,0 +1,56 @@
...
+
+  @Override
+  public final @Nullable Uri insert(@NotNull Uri uri, @Nullable ContentValues contentValues) {
+    securityChecker.checkPrivilegeEscalation(this);
+    return null;
+  }
+
+  @Override
+  public final int delete(@NotNull Uri uri, @Nullable String s, @Nullable String[] strings) {
+    securityChecker.checkPrivilegeEscalation(this);
```
vestrel00, line 43: Note that there is no need to add this check to to other overridable `insert` functions because they all ultimately use this function.

See my similar comment in my previous PR; https://github.com/getsentry/sentry-java/pull/2466#discussion_r1069676327

sentry-android-core/src/main/java/io/sentry/android/core/EmptySecureContentProvider.java
```diff
@@ -0,0 +1,56 @@
...
+ * A ContentProvider that does NOT store or provide any data for read or write operations.
+ *
+ * <p>This does not allow for overriding the abstract query, insert, update, and delete operations
+ * of the {@link ContentProvider}. Additionally, those functions are secure.
+ */
+@ApiStatus.Internal
+abstract class EmptySecureContentProvider extends ContentProvider {
+
+  private final ContentProviderSecurityChecker securityChecker =
+      new ContentProviderSecurityChecker();
```
vestrel00, line 22: You will notice that I am not initializing this with a reference to `this` ContentProvider. If I did, the code would be a tiny bit shorter at function call-sites;

```java
securityChecker.checkPrivilegeEscalation();
```

It looks nice and is okay because there is no harm in saving a hard reference to the ContentProvider as long as we don't leak it. However, this would come at the price of having to assign the value of this in `onCreate`, which would require it being annotated with `@CallSuper`, and this not being `final`...

```java
EmptySecureContentProvider {

  private ContentProviderSecurityChecker securityChecker;

  @CallSuper
  @Override
  public boolean onCreate() {
    securityChecker = new ContentProviderSecurityChecker(this);
    return false;
  }
}
```

So.. I think that we don't need to go through all that trouble just so that we don't have to pas in a reference to the ContentProvider when calling `securityChecker.checkPrivilegeEscalation`.

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/ContentProviderSecurityChecker.java
```diff
@@ -30,20 +29,21 @@ public ContentProviderSecurityChecker(final @NotNull BuildInfoProvider buildInfo
    * <p>See https://www.cvedetails.com/cve/CVE-2018-9492/ and
    * https://github.com/getsentry/sentry-java/issues/2460
    *
-   * <p>Call this function in the {@link ContentProvider#query(Uri, String[], String, String[],
-   * String)} function.
+   * <p>Call this function in the {@link ContentProvider}'s implementations of the abstract
+   * functions; query, insert, update, and delete.
```
vestrel00, line 33: Updated documentation.

sentry-android-core/api/sentry-android-core.api
```diff
@@ -189,24 +189,18 @@ public final class io/sentry/android/core/SentryAndroidOptions : io/sentry/Sentr
 	public fun setProfilingTracesIntervalMillis (I)V
 }
 
-public final class io/sentry/android/core/SentryInitProvider : android/content/ContentProvider {
+public final class io/sentry/android/core/SentryInitProvider {
```
vestrel00, line 192: I ran `./gradlew apiDump` to update this file. This should be safe and are NOT breaking changes because these APIs are internal and nothing really changed.

markushi (approved):
LGTM!

vestrel00:
Just realized how many typos I made in this PR 🤦

Anyways,  I will fix the merge conflicts tomorrow. Or feel free to do it if you plan to merge this now.

sentry-android-core/src/main/java/io/sentry/android/core/SentryInitProvider.java
```diff
@@ -45,38 +41,8 @@ public void attachInfo(@NotNull Context context, @NotNull ProviderInfo info) {
...
-      @Nullable String s,
-      @Nullable String[] strings1,
-      @Nullable String s1) {
-    new ContentProviderSecurityChecker().checkPrivilegeEscalation(this);
-    return null;
-  }
-
   @Override
   public @Nullable String getType(@NotNull Uri uri) {
     return null;
```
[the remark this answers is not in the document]
vestrel00: No. According to the [ContentProvider.getType](https://developer.android.com/reference/android/content/ContentProvider#getType(android.net.Uri)), documentation...

> Note that there are no permissions needed for an application to access this information; if your content provider requires read and/or write permissions, or is not exported, all applications can still call this method regardless of their access permissions. This allows them to retrieve the MIME type for a URI when dispatching intents.

There is no "privilege escalation" exploit here because there is no privilege required to use this function at all.  This is public function that is able to be used without any permissions even if not exported.

I would really like to keep my code changes minimal so as to minimize the risk of regression 🙏

vestrel00:
I fixed the merge conflicts. Feel free to merge whenever 🙏

markushi (approved):

markushi:
FYI: This fix will be part of the `6.13.0` release, which should be available within the next hours.

PR #2507, opened 2023-02-01
Add capabilities to track jetpack compose composition/rendering time

markushi:
## :scroll: Description
Add capabilities to track compose composition/rendering time. This requires some changes to our span interface. 
Since composition and rendering has no clear start and end, we'd need to have spans
* which get removed from the transaction in case they have no children
* which auto-finish whenever their root transaction finishes
* which trim their start/end based on start/end of their children

In terms of performance monitoring we want to track 
* Composition Time
* Rendering Time

## :pencil: Checklist

- [ ] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled
- [ ] No breaking changes


sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -343,6 +397,44 @@ public void finish(@Nullable SpanStatus status) {
   @Override
   @ApiStatus.Internal
   public void finish(@Nullable SpanStatus status, @Nullable SentryDate finishDate) {
+
+    // auto-finish any open spans
+    for (Span span : getSpans()) {
```
romtsn, line 402: `h`: as discussed internally, maybe let's try to create a new type for this kind of spans instead of introducing `SpanOptions` and packing everything into `SentryTracer`. There may be 2 different ways to solve this going forward:
* try to have `IdleSpan` with `idleTimeout` which finishes itself after some timeout. This would require changing Activity txs to `trimEnd = true` to have the correct end timestamp for them
* try to have `IdleSpan` without the idle timeout, so we could just iterate here the same way we do now, and check if the span is instance of `IdleSpan`, and force-finish it. It would then do the trimStart/trimEnd logic inside its own `finish` method.

sentry/src/main/java/io/sentry/Span.java
```diff
@@ -317,4 +340,17 @@ public boolean isNoOp() {
   void setSpanFinishedCallback(final @Nullable SpanFinishedCallback callback) {
     this.spanFinishedCallback = callback;
   }
+
+  public void setStartDate(SentryDate date) {
```
romtsn, line 344: `l`: some `final @Nullable/@NotNull` annotations are missing

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/compose/ComposeActivity.kt
```diff
@@ -39,52 +41,49 @@ class ComposeActivity : ComponentActivity() {
         super.onCreate(savedInstanceState)
 
         setContent {
-            val navController = rememberNavController().withSentryObservableEffect()
+            val navController = rememberNavController()
```
markushi, line 44:
```suggestion
            val navController = rememberNavController().withSentryObservableEffect()
```

sentry/src/main/java/io/sentry/SentryTracer.java
markushi: @romtsn Sounds good, but I'm wondering how far I should go here? Spans right now are a half-baked POJOs, e.g. a span can not directly access it's children. On the other hand a span has access to it's root transaction.

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
```diff
@@ -0,0 +1,71 @@
...
+    content: @Composable BoxScope.() -> Unit
+) {
+    val parentCompositionSpan = localSentryCompositionParentSpan.current
+    val parentRenderingSpan = localSentryRenderingParentSpan.current
+    val compositionSpan = parentCompositionSpan.item?.startChild(OP_COMPOSE, tag)
+    val firstRendered = remember { ImmutableHolder(false) }
+
+    Box(
+        modifier = modifier
+            .testTag(tag)
```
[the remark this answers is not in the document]
markushi: yeah I'm not sure if we should do that or not, kinda sneaky to add those tags right?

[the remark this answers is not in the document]
romtsn: I guess it's fine to add them to bring more value with auto-instrumentation out of the box, but maybe let's have a parameter for that, something like `enableUserInteractionTracing: Boolean = true`, which is turned on by default, when you're wrapping with `SentryTraced`?

sentry/src/main/java/io/sentry/SentryTracer.java
romtsn: hmm, good point. Would you mind asking about that in `team-client-infra` to see if anyone has concerns if we keep track of child spans in Span and not only in SentryTracer? For me it would be fine, but maybe there's something we should align on, or at least document.

Also, there's a [SpanRecorder](https://github.com/getsentry/sentry-javascript/blob/f8047f66ac8abf64f992b5d681007383ea175276/packages/tracing/src/span.ts#L18-L40) class available on JS. Not sure if it's exactly what we want to achieve, but maybe we could reuse some bits.

sentry/api/sentry.api
```diff
@@ -1742,7 +1750,7 @@ public final class io/sentry/SentryTraceHeader {
 
 public final class io/sentry/SentryTracer : io/sentry/ITransaction {
 	public fun <init> (Lio/sentry/TransactionContext;Lio/sentry/IHub;)V
-	public fun <init> (Lio/sentry/TransactionContext;Lio/sentry/IHub;ZLio/sentry/TransactionFinishedCallback;)V
+	public fun <init> (Lio/sentry/TransactionContext;Lio/sentry/IHub;Lio/sentry/TransactionOptions;Lio/sentry/TransactionFinishedCallback;)V
```
romtsn, line 1753: `m`: technically a breaking change, but I doubt anyone ever used that. Should we remove it altogether and just use the other ctor in the ActivityLifecycleIntegrationTest?

sentry/src/main/java/io/sentry/Span.java
```diff
@@ -317,4 +366,42 @@ public boolean isNoOp() {
...
+  }
+
+  /**
+   * Updates the end date of the span. Note: This will only update the end date if the span is
+   * already finished.
+   *
+   * @param date the end date to set
+   * @return true if the end date was updated, false otherwise
+   */
+  public boolean updateEndDate(@NotNull SentryDate date) {
```
romtsn, line 381: `m`: I think we have to merge against `main`, because Stefano's changes landed, and this is actually part of the ISpan interface there

sentry/src/main/java/io/sentry/Span.java
```diff
@@ -317,4 +366,42 @@ public boolean isNoOp() {
   void setSpanFinishedCallback(final @Nullable SpanFinishedCallback callback) {
     this.spanFinishedCallback = callback;
   }
+
+  public void setStartDate(@NotNull SentryDate date) {
```
romtsn, line 370: `l`: should we rename it to `updateStartDate` for the sake of alignment?

sentry/src/main/java/io/sentry/Span.java
```diff
@@ -163,6 +188,30 @@ public void finish(final @Nullable SpanStatus status, final @Nullable SentryDate
 
     this.context.setStatus(status);
     this.timestamp = timestamp == null ? hub.getOptions().getDateProvider().now() : timestamp;
+    if (options.isTrimStart() || options.isTrimEnd()) {
+      @Nullable SentryDate minChildStart = null;
+      @Nullable SentryDate maxChildEnd = null;
+
+      final @NotNull List<Span> children = getChildren();
+      for (final Span child : children) {
+        if (child.getParentSpanId() != null && child.getParentSpanId().equals(getSpanId())) {
```
romtsn, line 197: `m`: we don't need this check, because we already do it in `getChildren`, right?

sentry/src/main/java/io/sentry/Span.java
```diff
@@ -317,4 +366,42 @@ public boolean isNoOp() {
...
+  public boolean updateEndDate(@NotNull SentryDate date) {
+    if (this.timestamp != null) {
+      this.timestamp = date;
+      return true;
+    }
+    return false;
+  }
+
+  @NotNull
+  public SpanOptions getOptions() {
```
romtsn, line 390: `m`: maybe we should make this method package-private? Don't think we need to expose it
```suggestion
  SpanOptions getOptions() {
```

sentry/src/main/java/io/sentry/Span.java
```diff
@@ -163,6 +188,30 @@ public void finish(final @Nullable SpanStatus status, final @Nullable SentryDate
 
     this.context.setStatus(status);
     this.timestamp = timestamp == null ? hub.getOptions().getDateProvider().now() : timestamp;
+    if (options.isTrimStart() || options.isTrimEnd()) {
```
romtsn, line 191: `m`: I'd also like to have it tested if possible

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -345,8 +352,30 @@ public void finish(@Nullable SpanStatus status) {
...
+
+    // if it's not set -> fallback to the current time
+    if (finishTimestamp == null) {
+      finishTimestamp = hub.getOptions().getDateProvider().now();
+    }
+
+    // auto-finish any idle spans first
+    for (final Span span : children) {
+      if (span.getOptions().isIdle()) {
+        span.finish(getSpanContext().status, finishTimestamp);
```
romtsn, line 372: hrm, is there some idea behind not using the `status` parameter, which we are finishing this transaction with?

romtsn (approved):
lgtm, great stuff! 👍 

left some comments. Also there's one more point from the previous iteration - to surround the `testTag` with a boolean flag, which is turned on by default.

sentry/src/main/java/io/sentry/Span.java
markushi: There are some tests within `SentryTracer`, but let me add some to `Span` as well 👍

sentry/api/sentry.api
markushi: I think that's fine as `SentryTracer` is `ApiStatus.Internal`

sentry/src/main/java/io/sentry/Span.java
markushi: Also made it `private` since it's only accessed within `Span`

sentry/src/main/java/io/sentry/Span.java
```diff
@@ -163,6 +188,28 @@ public void finish(final @Nullable SpanStatus status, final @Nullable SentryDate
...
+        if (maxChildEnd == null
+            || (child.getFinishDate() != null && child.getFinishDate().isAfter(maxChildEnd))) {
+          maxChildEnd = child.getFinishDate();
+        }
+      }
+      if (options.isTrimStart() && minChildStart != null) {
+        updateStartDate(minChildStart);
+      }
+      if (options.isTrimEnd() && maxChildEnd != null) {
+        updateEndDate(maxChildEnd);
```
romtsn, line 209: just thinking if we should check here whether the current end timestamp is already less than the `maxChildEnd`? otherwise we're extending and not trimming kinda 😅 shouldn't happen I guess, but just for safety

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -345,8 +352,30 @@ public void finish(@Nullable SpanStatus status) {
   @Override
   @ApiStatus.Internal
   public void finish(@Nullable SpanStatus status, @Nullable SentryDate finishDate) {
```
romtsn, line 354: @adinauer wanna take a look at this, if it has any effect for backend? I don't think it does, but just to make sure we do not introduce some new potential memory leak or something

romtsn:
a couple of minor things, but looks good after the final pass! :shipit:

adinauer (approved):
Didn't see any obvious problems.

Issue #2518, opened 2023-02-06
App Start spans are not properly generated if SDK init is deferred


markushi:
### Integration

sentry-android

### Build System

Gradle

### AGP Version

7.4.0

### Proguard

Enabled

### Version

latest

### Steps to Reproduce

1. Launch App
2. Provide privacy confirmation dialog
3. Init the Sentry SDK

### Expected Result

App Start span is properly tracked from "App start" -> "first activity resume", even if the SDK is not initialized yet.

### Actual Result

The reported app start span is tracked from "App start" -> "first activity resume after SDK init".

### Related issues:
https://github.com/getsentry/sentry-java/pull/2459

PR #2519, opened 2023-02-06
Fix app start span end-time is wrong if SDK init is deferred

markushi:
This way the appstart start/end times are properly tracked, even if the
SDK is not initialized yet. In case of a deferred SDK init, the app
start span will be attached to the first tracked Activity after SDK
init.

Fixes https://github.com/getsentry/sentry-java/issues/2518

Unit tests

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
@@ -210,6 +210,14 @@ public final class ActivityLifecycleIntegration
                 getAppStartDesc(coldStart),
                 appStartTime,
                 Instrumenter.SENTRY);
+
+        // in case we already have an end time, e.g. due to deferred SDK init we can already finish
+        // the span
+        final SentryDate appStartEndTime = AppStartState.getInstance().getAppStartEndTime();
+        if (appStartEndTime != null) {
+          appStartSpan.finish(SpanStatus.OK, appStartEndTime);
+        }
+
         // The first activity ttidSpan should start at the same time as the app start time
         ttidSpanMap.put(
             activity,
@@ -328,26 +336,16 @@ public final class ActivityLifecycleIntegration
   @SuppressLint("NewApi")
   @Override
   public synchronized void onActivityResumed(final @NotNull Activity activity) {
+    // finishes app start span
+    @Nullable final SentryDate appStartEndTime = AppStartState.getInstance().getAppStartEndTime();
+    if (appStartSpan != null
+        && !appStartSpan.isFinished()
+        && performanceEnabled
+        && appStartEndTime != null) {
+      appStartSpan.finish(SpanStatus.OK, appStartEndTime);
+    }
+
     if (!firstActivityResumed) {
-
-      // we only finish the app start if the process is of foregroundImportance
-      if (foregroundImportance) {
-        // sets App start as finished when the very first activity calls onResume
-        AppStartState.getInstance().setAppStartEnd();
-      } else {
-        if (options != null) {
-          options
-              .getLogger()
-              .log(
-                  SentryLevel.DEBUG,
-                  "App Start won't be reported because Process wasn't of foregroundImportance.");
-        }
-      }
-
-      // finishes app start span
-      if (performanceEnabled && appStartSpan != null) {
-        appStartSpan.finish();
-      }
       firstActivityResumed = true;
     }
 
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
@@ -56,8 +56,7 @@ public final class ActivityLifecycleIntegration
   private boolean isAllActivityCallbacksAvailable;
 
   private boolean firstActivityCreated = false;
-  private boolean firstActivityResumed = false;
-  private boolean foregroundImportance = false;
+  private final boolean foregroundImportance;
 
   private @Nullable ISpan appStartSpan;
   private final @NotNull WeakHashMap<Activity, ISpan> ttidSpanMap = new WeakHashMap<>();
@@ -211,12 +210,9 @@ public final class ActivityLifecycleIntegration
                 appStartTime,
                 Instrumenter.SENTRY);
 
-        // in case we already have an end time, e.g. due to deferred SDK init we can already finish
-        // the span
-        final SentryDate appStartEndTime = AppStartState.getInstance().getAppStartEndTime();
-        if (appStartEndTime != null) {
-          appStartSpan.finish(SpanStatus.OK, appStartEndTime);
-        }
+        // in case there's already an end time (e.g. due to deferred SDK init)
+        // we can finish the app-start span
+        finishAppStartSpan();
 
         // The first activity ttidSpan should start at the same time as the app start time
         ttidSpanMap.put(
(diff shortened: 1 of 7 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -328,26 +336,16 @@ public synchronized void onActivityStarted(final @NotNull Activity activity) {
...
+        && performanceEnabled
+        && appStartEndTime != null) {
+      appStartSpan.finish(SpanStatus.OK, appStartEndTime);
+    }
 
-      // finishes app start span
-      if (performanceEnabled && appStartSpan != null) {
-        appStartSpan.finish();
-      }
+    if (!firstActivityResumed) {
```
romtsn, line 348: `m`: I believe we don't need this flag anymore? It's not used anywhere else I think

sentry-android-core/src/main/java/io/sentry/android/core/SentryPerformanceProvider.java
```diff
@@ -102,7 +101,16 @@ public void onActivityCreated(@NotNull Activity activity, @Nullable Bundle saved
   public void onActivityStarted(@NotNull Activity activity) {}
 
   @Override
-  public void onActivityResumed(@NotNull Activity activity) {}
+  public void onActivityResumed(@NotNull Activity activity) {
+    if (!firstActivityResumed) {
```
romtsn, line 105: `m`: hmm, do we still need to check for `foregroundImportance` here actually?

romtsn (approved):
lgtm 👍 a couple of minor comments

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -328,26 +336,16 @@ public synchronized void onActivityStarted(final @NotNull Activity activity) {
...
-          options
-              .getLogger()
-              .log(
-                  SentryLevel.DEBUG,
-                  "App Start won't be reported because Process wasn't of foregroundImportance.");
-        }
-      }
+    // finishes app start span
+    @Nullable final SentryDate appStartEndTime = AppStartState.getInstance().getAppStartEndTime();
+    if (appStartSpan != null
```
romtsn, line 341: `h`: actually, I just realized - if the users have removed/disabled the `SentryPerformanceProvider` via manifest, we'd never finish the app start span, because the `appStartEndTime` is not gonna be ever set. If you look at `SentryAndroid` - we set the app start times there to mitigate the removal of the provider. So I believe we should also have a fallback here in case `appStartEndTime` is not available

romtsn:
@marandaneto would you like to take a look at this? Was there a reason why we have not capture the app start span end time in the same provider where we capture the start time?

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
markushi: true, let me remove that

sentry-android-core/src/main/java/io/sentry/android/core/SentryPerformanceProvider.java
markushi: The `foregroundImportance` check is still done inside the `ActivityLifecycleIntegration`, as it's also used for creating the transaction as well.
So, yes the end-time in `AppStartState` is still set, but it's ignored in the `ActivityLifecycleIntegration`, so I guess it's fine.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -210,6 +210,14 @@ private void startTracing(final @NotNull Activity activity) {
                 getAppStartDesc(coldStart),
                 appStartTime,
                 Instrumenter.SENTRY);
+
+        // in case we already have an end time, e.g. due to deferred SDK init we can already finish
+        // the span
+        final SentryDate appStartEndTime = AppStartState.getInstance().getAppStartEndTime();
+        if (appStartEndTime != null) {
+          appStartSpan.finish(SpanStatus.OK, appStartEndTime);
```
marandaneto, line 218: M: `status` should only be set in case there's none otherwise you overwrite values set by the user, similar case on L345

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
romtsn: yea, but that's what I'm saying - in [SentryAndroid](https://github.com/getsentry/sentry-java/blob/f6a135d1ef81f9cd70d72608e8beb8c2103f69bd/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java#L84-L86) we actually set the app start times again for the case when the provider is disabled, so it will be set still :)

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -328,26 +336,16 @@ public synchronized void onActivityStarted(final @NotNull Activity activity) {
...
-        if (options != null) {
-          options
-              .getLogger()
-              .log(
-                  SentryLevel.DEBUG,
-                  "App Start won't be reported because Process wasn't of foregroundImportance.");
-        }
-      }
+    // finishes app start span
+    @Nullable final SentryDate appStartEndTime = AppStartState.getInstance().getAppStartEndTime();
```
marandaneto, line 340: L: This block and L216 do pretty much the same, maybe extracting to a method would allow DRY.

sentry-android-core/src/main/java/io/sentry/android/core/SentryPerformanceProvider.java
marandaneto: `foregroundImportance` is about if the App received a notification and no screen was shown or something similar, so the app start is inaccurate because we end the app start only when there's a visible screen, I'd say unless you fix this problem using another technique, this should stay.

marandaneto:
> @marandaneto would you like to take a look at this? Was there a reason why we have not capture the app start span end time in the same provider where we capture the start time?

Most likely because we offer the possibility of no content providers (docs tell how to remove them).
Also because Hybrid SDKs finish the app start by themselves (checking their own screen visibility).
The current approach might be a problem for the later, it has to be checked.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -328,28 +332,17 @@ public synchronized void onActivityStarted(final @NotNull Activity activity) {
...
+    // app start span
+    @Nullable final SentryDate appStartStartTime = AppStartState.getInstance().getAppStartTime();
+    @Nullable final SentryDate appStartEndTime = AppStartState.getInstance().getAppStartEndTime();
+    // in case the SentryPerformanceProvider is disabled it does not set the app start times,
+    // and we need to set the end time manually here,
+    // the start time gets set manually in SentryAndroid.init()
+    if (appStartStartTime != null && appStartEndTime == null) {
+      AppStartState.getInstance().setAppStartEnd();
     }
+    finishAppStartSpan();
```
romtsn, line 345: hm, do we still need to wrap it with the `firstActivityResumed` flag, for the case when `SentryPerformanceProvider` is disabled? or he `appStartSpan.isFinished()` check should do?

markushi: yeah I'm ensuring the `appStartSpan` is only finished once, but let me add another test for that to ensure this behavior

PR #2537, opened 2023-02-14
add enableReadForegroundImportance

fzyzcjy:
## :scroll: Description

Close #2187 

Please see description there :)

May not have much time so changing code to this PR maybe slow 

## :bulb: Motivation and Context

Please see description there :)

## :green_heart: How did you test it?

Manual testing - this is just a simple flag.
Not sure how to write unit tests after looking at existing codebase. Will add unit tests if provided some hints.

## :pencil: Checklist

- [x] I reviewed the submitted code.
- [ ] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [x] I updated the docs if needed.
- [ ] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

PR #2543, opened 2023-02-14
Fix invalid session creation when app is launched in background

markushi:
## :scroll: Description
Skip session creation if the app is not started in foreground.

This is a follow-up based on some customer feedback, specifically seeing session creations spikes for scheduled background jobs (e.g. a nightly sync).

Likely fixes https://github.com/getsentry/sentry-java/discussions/2512 

## :pencil: Checklist

### DIFF
--- a/sentry/src/main/java/io/sentry/SentryClient.java
+++ b/sentry/src/main/java/io/sentry/SentryClient.java
@@ -479,13 +479,6 @@ public final class SentryClient implements ISentryClient {
       return;
     }
 
-    if (session.isAppInForeground() != null && !session.isAppInForeground()) {
-      options
-          .getLogger()
-          .log(SentryLevel.WARNING, "Sessions can't be captured because app was in background.");
-      return;
-    }
-
     SentryEnvelope envelope;
     try {
       envelope = SentryEnvelope.from(options.getSerializer(), session, options.getSdkVersion());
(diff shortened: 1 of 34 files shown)


markushi:
@romtsn as an alternative we could keep the old logic (always create a session) and attach a foreground flag once the first Activity is launched ~ `session.setIsForeground(true|false)`. When `SentryClient.captureSession()` is called, we could then drop non-foreground sessions. This way we don't have to check for foreground importance which might causes issues in third-party app stores (see https://github.com/getsentry/sentry-java/pull/2537)

romtsn:
> @romtsn as an alternative we could keep the old logic (always create a session) and attach a foreground flag once the first Activity is launched ~ `session.setIsForeground(true|false)`. When `SentryClient.captureSession()` is called, we could then drop non-foreground sessions. This way we don't have to check for foreground importance which might causes issues in third-party app stores (see #2537)

I think that makes sense, would you mind changing it then?

sentry/src/main/java/io/sentry/SentryClient.java
```diff
@@ -479,6 +479,13 @@ public void captureSession(final @NotNull Session session, final @Nullable Hint
       return;
     }
 
+    if (session.isAppInForeground() != null && !session.isAppInForeground()) {
```
stefanosiano, line 482: Maybe replace this with
`if (!Boolean.TRUE.equals(session.isAppInForeground()))`?

romtsn (approved):

romtsn:
:shipit:

PR #2550, opened 2023-02-17
Adjust time-to-full-display span if reportFullDisplayed is called too early

stefanosiano:
## :scroll: Description
`time-to-full-display` end timestamp is now adjusted to the `time-to-initial-display` end timestamp if called too early
added updateEndDate method to ISpan

If the user calls the manual API `Sentry.reportFullDisplayed` too early, we would show the `time-to-full-display` span to be shorter than our automatically calculated `time-to-initial-display`, which would be pretty confusing.
This adjustment is done also by the [Android system](https://developer.android.com/reference/android/app/Activity#reportFullyDrawn()): `If this method is called before the activity's window is first drawn and displayed as measured by the system, the reported time here will be shifted to the system measured time.`
It's also done by the Firebase SDK.

Unit tests

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Features
 
+- Adjust time-to-full-display span if reportFullDisplayed is called too early ([#2550](https://github.com/getsentry/sentry-java/pull/2550))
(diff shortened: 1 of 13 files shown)


markushi (approved):
Nice one, looks good!

PR #2573, opened 2023-02-27
Make 0 sample rate a valid rate

romtsn:
## :scroll: Description

Allow 0.0 sample rate to be set (though the effect would be as setting sample rate to null).

To align with other SDKs (RN, Cocoa, Flutter)

Automated tests

## :pencil: Checklist

### DIFF
--- a/sentry/src/test/java/io/sentry/SentryOptionsTest.kt
+++ b/sentry/src/test/java/io/sentry/SentryOptionsTest.kt
@@ -92,11 +92,6 @@ class SentryOptionsTest {
     }
 
     @Test
-    fun `when setSampling is set to exactly 0, setter throws`() {
-        assertFailsWith<IllegalArgumentException> { SentryOptions().sampleRate = 0.0 }
-    }
-
-    @Test
     fun `when setTracesSampleRate is set to exactly 0, value is set`() {
         val options = SentryOptions().apply {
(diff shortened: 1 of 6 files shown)


markushi (approved):
LGTM! [The docs](https://docs.sentry.io/platforms/android/configuration/options/#sample-rate) already mention "a value between 0.0 and 1.0", so no need to update anything there.

adinauer (approved):
LGTM

sentry/src/test/java/io/sentry/SentryOptionsTest.kt
```diff
@@ -91,11 +91,6 @@ class SentryOptionsTest {
         assertFailsWith<IllegalArgumentException> { SentryOptions().sampleRate = -0.0000000000001 }
     }
 
-    @Test
-    fun `when setSampling is set to exactly 0, setter throws`() {
```
stefanosiano, line 95: maybe add a test `when setSampleRate is set to exactly 0, value is set` like it's done for `setTracesSampleRate`?

romtsn: We already test this as part of `SampleRateUtilsTest`, and in the options we just use the same logic, so I'd keep it like that if you don't mind

PR #2584, opened 2023-03-03
Fixed timestamps of slow and frozen frames

stefanosiano:
## :scroll: Description
Fixed timestamps of slow and frozen frames with inspiration from JankStats library https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:metrics/metrics-performance/src/main/java/androidx/metrics/performance/JankStatsApi24Impl.kt
Frame start timestamp is found using reflection
Only time spent in cpu is counted for frame duration 

One of the reasons of the wrong timestamps is the internal timestamp is not exposed, so we are now reading it through reflection.
Another reason is vsync, that was taken into account by the internal timestamp, resulting in a frame being drawn in the future, with other frames being created in the meantime.
Also, we are now using only the time spent by the cpu in the main thread for frame creation, without considering the time spent in the render thread by the gpu.

The reported timestamp of some frames were overlapping, causing confusion

## :pencil: Checklist


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Fixed timestamps of slow and frozen frames ([#2584](https://github.com/getsentry/sentry-java/pull/2584))
```
```suggestion
- Fixed timestamps of slow and frozen frames for profiles ([#2584](https://github.com/getsentry/sentry-java/pull/2584))
```

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
```diff
@@ -83,18 +89,83 @@ public SentryFrameMetricsCollector(
...
+            l.onFrameMetricCollected(lastFrameEndNanos, cpuDuration, refreshRate);
           }
         };
   }
 
+  /**
+   * Return the internal timestamp in the choreographer of the last frame start timestamp through
+   * reflection.
+   */
+  private long getFrameStartTimestamp() {
```
romtsn, line 139: `m`: I've also seen JankStats uses a different implementation to get the frame start time from [API 26 and above](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:metrics/metrics-performance/src/main/java/androidx/metrics/performance/JankStatsApi26Impl.kt).

Should we consider this in our implementation as well? Not sure if it's more accurate, maybe needs some testing (or maybe it just avoids reflection on new Android versions, which is also good for us?).

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
```diff
@@ -83,18 +89,83 @@ public SentryFrameMetricsCollector(
...
+          final long now = System.nanoTime();
+          final float refreshRate =
               buildInfoProvider.getSdkInfoVersion() >= Build.VERSION_CODES.R
                   ? window.getContext().getDisplay().getRefreshRate()
                   : window.getWindowManager().getDefaultDisplay().getRefreshRate();
+
+          final long cpuDuration = getFrameCpuDuration(frameMetrics);
+          long startTime = getFrameStartTimestamp();
+          // If we couldn't get the timestamp through reflection, we use current time
+          if (startTime == -1) {
```
romtsn, line 117: `m`: would it be possible to test this new logic (and everything below that)? I see we test that we access Choreographer, but this would also worth to test I think.

stefanosiano: added the new tests

PR #2605, opened 2023-03-15
draft: Unity backwards compatibility issue

bitsandfoxes:
fix 4 unity


romtsn:
@bitsandfoxes closing this as I assume we're gonna tackle this by changing how unity is consuming android artifacts

PR #2609, opened 2023-03-16
Fix session tracking

romtsn:
## :scroll: Description

We use `withScope` to start a session, but `withScope` drops the scope immediately after the callback, so technically we were starting a new session in every onStart, but never finishing it. 

I switched to `configureScope` to just set the new `lastUpdatedSession` value, if it was started in `.init`, and moved starting session code back out of the callback as it was before.

Also added an integration test that was failing before the fix.

Customer issue

Manually and automated

## :pencil: Checklist

## :crystal_ball: Next steps
:shipit:


romtsn:
At first, I got rid of `getSession` and switched to using `withSession` callback, but then realized that it might be stuck in a lock, if another thread is calling `captureEvent`/`captureException` in the meantime, so I guess it's better to not use that method here on the main thread, since we're anyway just reading the start time, which is not gonna change ever.

markushi (approved):
Nice one, lgtm!

PR #2610, opened 2023-03-17
Fix wrong default environment in Session

### LINKED ISSUE #2381: sessions sent without an environment when environment option is not set - Android
### Description

Similar to https://github.com/getsentry/sentry-cocoa/issues/2411

If the environment is not set for the SDK errors will default to production. However, sessions are sent without an environment.

Expected outcome: sessions would be sent with an environment
Actual outcome: sessions are sent without an environment set

Steps To Reproduce

1. set the environment and release options
2. Send an error event and session
3. Observe that session and error event with the correct environment
4. repeat without setting the environment option
5. Observe the error event with an environment of production
6. Observe the session with no environment

lbloder:
## :scroll: Description
The environment getter in `SentryOptions` now returns "production" as default if not set.

Fixes #2381

## :pencil: Checklist

### DIFF
--- a/sentry/src/main/java/io/sentry/SentryOptions.java
+++ b/sentry/src/main/java/io/sentry/SentryOptions.java
@@ -777,7 +777,7 @@ public class SentryOptions {
   /**
    * Returns the environment
    *
-   * @return the environment or null if not set
+   * @return the environment or 'production' if not set
    */
   public @Nullable String getEnvironment() {
     return environment != null ? environment : DEFAULT_ENVIRONMENT;


sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -773,8 +779,8 @@ public void setRelease(@Nullable String release) {
    *
    * @return the environment or null if not set
```
romtsn, line 780:
```suggestion
   * @return the environment or 'production' if not set
```

romtsn (approved):

adinauer (approved):

Issue #1620, opened 2021-07-24
Sentry Android does not capture LogCat logs


fzyzcjy:
_Platform:_
- [x]  Android -> If yes, which Device API (and compileSdkVersion/targetSdkVersion/Build tools) version?
- [ ]  Java -> If yes, which Java (and sourceCompatibility/targetCompatibility) version?
- [ ]  Kotlin -> If yes, which Kotlin (and jvmTarget) version?
- [ ]  NDK -> If yes, which NDK/CMake version?
- [ ]  React-Native -> If yes, which version?
- [ ]  Timber -> If yes, which version?
- [ ]  Log4j2 -> If yes, which version?
- [ ]  Logback -> If yes, which version?
- [ ]  Spring -> If yes, which version?

_IDE:_
- [x]  Android Studio -> If yes, which version?
- [ ]  IntelliJ -> If yes, which version?
- [ ]  Other -> If yes, which one?

_Build system:_
- [x]  Gradle -> If yes, which version?
- [ ]  Buck -> If yes, which version?
- [ ]  Bazel -> If yes, which version?
- [ ]  Maven -> If yes, which version?
- [ ]  Other -> If yes, which one?

_Android Gradle Plugin:_
- [ ]  Yes -> If yes, which version?
- [x]  No

_Sentry Android Gradle Plugin:_
- [ ]  Yes -> If yes, which version?
- [x]  No

_Proguard/R8:_
- [x]  Enabled
- [ ]  Disabled

_Platform installed with:_
- [ ] JCenter
- [ ] Bintray
- [ ] Maven Central
- [ ] Manually

The version of the SDK:
**3.x.x**

---
I have the following issue:

The description goes here ...

**Steps to reproduce:**
- run sample

**Actual result:**
Sentry Android does not capture LogCat logs

**Expected result:**
when having things like `Log.i(...)`, I hope sentry can automatically capture it.
thank you!

bruno-garcia:
Thanks for raising @fzyzcjy. Last we looked at this it wasn't as straight forward as something like Timber. We'll do another investigation. Do you have suggestions on how to implement this and/or is willing to make a PR to add support?

fzyzcjy:
@bruno-garcia I also do not have quite good solutions. If there is an idea I can make a PR. In addition, it is not possible to let every library use Timber. For example, one third party library may print out some Log.w or Log.e which I want to capture and recognize (because it can be the cause of some bugs), but it is not possible now :(

[...]

PR #2620, opened 2023-03-23
feat: logcat integration

buenaflor:
## :scroll: Description
Implements a `SentryLogcatAdapter` that will replace `Log` calls via bytecode manipulation in the Sentry AGP.

Ref: #1620 

This adapter is needed by the AGP. See: https://github.com/getsentry/sentry-android-gradle-plugin/pull/455

 - Manual testing
 - Jadx
 
## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/SentryLogcatAdapter.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/SentryLogcatAdapter.java
@@ -1,11 +1,20 @@
 package io.sentry.android.core;
 
 import android.util.Log;
+
 import io.sentry.Breadcrumb;
 import io.sentry.Sentry;
 import io.sentry.SentryLevel;
+
+import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 
+/**
+ * This class replaces {@link android.util.Log} with its own implementations which creates a
+ * {@link io.sentry.Breadcrumb} for each log. It only replaces log functions that meet a minimum level
+ * set by the user, set through the Sentry Android Gradle Plugin.
+ */
+@ApiStatus.Internal
 public class SentryLogcatAdapter {
 
   private static void addAsBreadcrumb(@NotNull String tag, @NotNull SentryLevel level, String msg) {
@@ -13,12 +22,12 @@ public class SentryLogcatAdapter {
   }
 
   private static void addAsBreadcrumb(
-      @NotNull String tag, @NotNull SentryLevel level, Throwable tr) {
+    @NotNull String tag, @NotNull SentryLevel level, Throwable tr) {
     addAsBreadcrumb(tag, level, null, tr);
   }
 
   private static void addAsBreadcrumb(
-      @NotNull String tag, @NotNull SentryLevel level, String msg, Throwable tr) {
+    @NotNull String tag, @NotNull SentryLevel level, String msg, Throwable tr) {
     Breadcrumb breadcrumb = new Breadcrumb();
     breadcrumb.setCategory("log");
     breadcrumb.setMessage(msg);
--- a/sentry-android-core/src/main/java/io/sentry/android/core/SentryLogcatAdapter.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/SentryLogcatAdapter.java
@@ -1,18 +1,16 @@
 package io.sentry.android.core;
 
 import android.util.Log;
-
 import io.sentry.Breadcrumb;
 import io.sentry.Sentry;
 import io.sentry.SentryLevel;
-
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 
 /**
- * This class replaces {@link android.util.Log} with its own implementations which creates a
- * {@link io.sentry.Breadcrumb} for each log. It only replaces log functions that meet a minimum level
- * set by the user on the Sentry Android Gradle Plugin.
+ * This class replaces {@link android.util.Log} with its own implementations which creates a {@link
+ * io.sentry.Breadcrumb} for each log. It only replaces log functions that meet a minimum level set
+ * by the user on the Sentry Android Gradle Plugin.
  */
 @ApiStatus.Internal
 public final class SentryLogcatAdapter {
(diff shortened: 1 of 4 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/SentryLogcatAdapter.java
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  private static void addAsBreadcrumb(
+      @NotNull String tag, @NotNull SentryLevel level, Throwable tr) {
+    addAsBreadcrumb(tag, level, null, tr);
+  }
+
+  private static void addAsBreadcrumb(
+      @NotNull String tag, @NotNull SentryLevel level, String msg, Throwable tr) {
+    Breadcrumb breadcrumb = new Breadcrumb();
```
markushi, line 29: We usually like to use the `final` keyword to indicate that objects will not be re-assigned. 
```suggestion
      @NotNull final String tag, @NotNull final SentryLevel level, final String msg, final Throwable tr) {
    final Breadcrumb breadcrumb = new Breadcrumb();
```

sentry-android-core/src/main/java/io/sentry/android/core/SentryLogcatAdapter.java
```diff
@@ -0,0 +1,109 @@
...
+    breadcrumb.setMessage(msg);
+    breadcrumb.setLevel(level);
+    breadcrumb.setData("tag", tag);
+    if (tr != null) {
+      breadcrumb.setData("throwable", tr.getMessage());
+    }
+    Sentry.addBreadcrumb(breadcrumb);
+  }
+
+  public static int v(String tag, @NotNull String msg) {
```
markushi, line 40: I think it probably makes sense to have `tag`, `msg` and `throwable` marked as `@Nullable` everwhere
image

sentry-android-core/src/main/java/io/sentry/android/core/SentryLogcatAdapter.java
```diff
@@ -0,0 +1,109 @@
...
+
+  private static void addAsBreadcrumb(
+      @NotNull String tag, @NotNull SentryLevel level, String msg, Throwable tr) {
+    Breadcrumb breadcrumb = new Breadcrumb();
+    breadcrumb.setCategory("log");
+    breadcrumb.setMessage(msg);
+    breadcrumb.setLevel(level);
+    breadcrumb.setData("tag", tag);
+    if (tr != null) {
+      breadcrumb.setData("throwable", tr.getMessage());
```
markushi, line 35: I think in the first version it's fine to just have the throwable message inside the breadcrumb, @romtsn what do you think?

markushi (commented):
Looking good, please check my comments

markushi:
I think it could make sense to add an integration test here as well.
> Given an initialized Sentry SDK
> When SentryLogcatAdapter.x is called with params y,z
> Then a breadcrumb should be captured with y,z

sentry-android-core/src/main/java/io/sentry/android/core/SentryLogcatAdapter.java
```diff
@@ -0,0 +1,109 @@
...
+import io.sentry.SentryLevel;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * This class replaces {@link android.util.Log} with its own implementations which creates a {@link
+ * io.sentry.Breadcrumb} for each log. It only replaces log functions that meet a minimum level set
+ * by the user on the Sentry Android Gradle Plugin.
+ */
+@ApiStatus.Internal
```
romtsn, line 15: Dunno if it makes sense to make it public, people might want to still use it instead of `android.util.Log`, but without the gradle plugin, what do you think @markushi @buenaflor ?

sentry-android-core/src/main/java/io/sentry/android/core/SentryLogcatAdapter.java
```diff
@@ -0,0 +1,109 @@
...
+
+  private static void addAsBreadcrumb(
+      @NotNull String tag, @NotNull SentryLevel level, Throwable tr) {
+    addAsBreadcrumb(tag, level, null, tr);
+  }
+
+  private static void addAsBreadcrumb(
+      @NotNull String tag, @NotNull SentryLevel level, String msg, Throwable tr) {
+    Breadcrumb breadcrumb = new Breadcrumb();
+    breadcrumb.setCategory("log");
```
romtsn, line 30:
```suggestion
    breadcrumb.setCategory("Logcat");
```
would it make sense to have Logcat here? Similar to how we set [Timber](https://github.com/getsentry/sentry-java/blob/d81684e2d1825564da8d0e52454c0e3ede5cd514/sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt#L290)

sentry-android-core/src/main/java/io/sentry/android/core/SentryLogcatAdapter.java
markushi: IMHO I wouldn't 😅 Since the config for `minLogLevel` lives within the gradle plugin, people calling the `SentryLogcatAdapter` will always generate breadcrumbs, even for e.g. verbose logging

sentry-android-core/src/main/java/io/sentry/android/core/SentryLogcatAdapter.java
romtsn: yeah, I think that's fine

sentry-android-core/src/test/java/io/sentry/android/core/SentryLogcatAdapterTest.kt
```diff
@@ -0,0 +1,184 @@
...
+import io.sentry.SentryLevel
+import io.sentry.SentryOptions
+import org.junit.runner.RunWith
+import java.lang.RuntimeException
+import kotlin.test.BeforeTest
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+@RunWith(AndroidJUnit4::class)
+class SentryLogcatAdapterTest {
```
markushi, line 16: Nice!

markushi (approved):
LGTM!

Issue #2601, opened 2023-03-13
Crash when closing sentry


bmarty:
### Integration

sentry-android

### Build System

Gradle

### AGP Version

7.6

### Proguard

Enabled

### Version

6.15.0

### Steps to Reproduce

Call `Sentry.close()`

### Expected Result

No crash

### Actual Result

The application crashes.

Here is the stack trace:

```
Thread: main, Exception: java.lang.NullPointerException: Parameter specified as non-null is null: method androidx.lifecycle.LifecycleRegistry.removeObserver, parameter observer
at androidx.lifecycle.LifecycleRegistry.removeObserver(LifecycleRegistry.kt:4)
at io.sentry.android.core.AppLifecycleIntegration$$ExternalSyntheticLambda0.run(R8$$SyntheticClass:12)
at android.os.Handler.handleCallback(Handler.java:883)
at android.os.Handler.dispatchMessage(Handler.java:100)
at android.os.Looper.loop(Looper.java:237)
at android.app.ActivityThread.main(ActivityThread.java:8167)
at java.lang.reflect.Method.invoke(Native Method)
at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:496)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1100)
```

I believe that the issue is coming from [these lines](https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/AppLifecycleIntegration.java#L112-L131)

When `removeObserver` is invoked, the `watcher` is null.

This is not new code, but the crash is occurring only now, I am not sure why.

Also this does not crash on every device. Crash on Android 10.

jmartinesp:
When I tried to debug the issue I confirmed that this [`watcher = null`](https://github.com/getsentry/sentry-java/blob/469420402e4717906783a744f8e2fb156168ec25/sentry-android-core/src/main/java/io/sentry/android/core/AppLifecycleIntegration.java#L126) line gets called before the runnable in [`handler.post(() -> removeObserver());`](https://github.com/getsentry/sentry-java/blob/469420402e4717906783a744f8e2fb156168ec25/sentry-android-core/src/main/java/io/sentry/android/core/AppLifecycleIntegration.java#L124), so at the `removeObserver` call, watcher is already null.

romtsn:
Hey, apologies for the crash, yeah that indeed can be a problem when we use handler, we'll look into this!

[...]

PR #2625, opened 2023-03-27
Fix missing null check when removing lifecycle observer

markushi:
## :scroll: Description
If `Sentry.close` is called off the main thread, the lifecycle watcher field is set to `null`, while de-registering is posted on the main thread, leading to a NPE in combination with some versions of android x lifecycle lib.

Fixes https://github.com/getsentry/sentry-java/issues/2601

Manual testing, the overhead of running a single test against different version of the androidx lifecycle library didn't seem to justify the efforts.

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AppLifecycleIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AppLifecycleIntegration.java
@@ -15,7 +15,7 @@ import org.jetbrains.annotations.TestOnly;
 
 public final class AppLifecycleIntegration implements Integration, Closeable {
 
-  @TestOnly @Nullable LifecycleWatcher watcher;
+  @TestOnly @Nullable volatile LifecycleWatcher watcher;
 
   private @Nullable SentryAndroidOptions options;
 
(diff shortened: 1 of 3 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/AppLifecycleIntegration.java
```diff
@@ -110,23 +110,27 @@ private void addObserver(final @NotNull IHub hub) {
   }
 
   private void removeObserver() {
-    ProcessLifecycleOwner.get().getLifecycle().removeObserver(watcher);
+    final @Nullable LifecycleWatcher watcherRef = watcher;
```
stefanosiano, line 113: Is the `watcherRef` necessary? An `if` on the `watcher` should be enough, shouldn't it?

stefanosiano (commented):
Can we also add a test to verify the change? like verifying the watcher is not null after the `close` function is called in a background thread and it's null only after the main thread executes?

sentry-android-core/src/main/java/io/sentry/android/core/AppLifecycleIntegration.java
markushi: Yes, as watcher could be set to `null` on another thread

sentry-android-core/src/test/java/io/sentry/android/core/AppLifecycleIntegrationTest.kt
```diff
@@ -94,4 +96,23 @@ class AppLifecycleIntegrationTest {
...
+
+        assertNotNull(sut.watcher)
+
+        Thread {
+            sut.close()
+            latch.countDown()
+        }.start()
+
+        latch.await()
+        InstrumentationRegistry.getInstrumentation().waitForIdleSync()
```
stefanosiano, line 115: For the main looper it should be possible to use `shadowOf(Looper.getMainLooper()).idle()` instead of creating a thread manually. It would be easier for readability if it works

stefanosiano (approved):
just a minor comment, if possible.
Looks good for everything else!

PR #2631, opened 2023-03-29
Trim time-to-full-display span if reportFullyDisplayed API is never called

stefanosiano:
## :scroll: Description
ttfd span is trimmed if `reportFullyDisplayed` API is never called
exposed `getFinishDate` internally in ISpan

Currently if the user enables the option to track the fullDisplay, the transaction keeps running and if the manual API is never called, the `time-to-full-display` span would always last 30 seconds, increasing the transaction duration.
This change aims to mitigate this issue, so that if no other spans are present, the total transaction duration would be until the `time-to-initial-display` span is finished.

Unit test

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
@@ -546,8 +546,8 @@ public final class ActivityLifecycleIntegration
     }
     ttfdSpan.setDescription(getExceededTtfdDesc(ttfdSpan));
     // We set the end timestamp of the ttfd span to be equal to the ttid span. This way,
-    SentryDate ttidEndDate = ttidSpan != null ? ttidSpan.getFinishDate() : null;
-    SentryDate ttfdEndDate = ttidEndDate != null ? ttidEndDate : ttfdSpan.getStartDate();
+    final @Nullable SentryDate ttidEndDate = ttidSpan != null ? ttidSpan.getFinishDate() : null;
+    final @NotNull SentryDate ttfdEndDate = ttidEndDate != null ? ttidEndDate : ttfdSpan.getStartDate();
     finishSpan(ttfdSpan, ttfdEndDate, SpanStatus.DEADLINE_EXCEEDED);
   }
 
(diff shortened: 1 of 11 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -531,6 +540,17 @@ private void onFullFrameDrawn() {
...
   }
 
+  private void finishExceededTtfdSpan(final @Nullable ISpan ttidSpan) {
+    if (ttfdSpan == null) {
+      return;
+    }
+    ttfdSpan.setDescription(getExceededTtfdDesc(ttfdSpan));
+    // We set the end timestamp of the ttfd span to be equal to the ttid span. This way,
+    SentryDate ttidEndDate = ttidSpan != null ? ttidSpan.getFinishDate() : null;
+    SentryDate ttfdEndDate = ttidEndDate != null ? ttidEndDate : ttfdSpan.getStartDate();
```
markushi, line 550:
```suggestion
    final SentryDate ttidEndDate = ttidSpan != null ? ttidSpan.getFinishDate() : null;
    final SentryDate ttfdEndDate = ttidEndDate != null ? ttidEndDate : ttfdSpan.getStartDate();
```

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -531,6 +540,17 @@ private void onFullFrameDrawn() {
     cancelTtfdAutoClose();
   }
 
+  private void finishExceededTtfdSpan(final @Nullable ISpan ttidSpan) {
+    if (ttfdSpan == null) {
+      return;
+    }
+    ttfdSpan.setDescription(getExceededTtfdDesc(ttfdSpan));
```
markushi, line 547: `low`: I guess it's never happening but in case `finishExceededTtfdSpan` get's called mutliple times, we'd add the `"-exceeded"` multiple times here.

markushi (approved):
LGTM!

PR #2158, opened 2022-07-08
OkHttp span details

marandaneto:
## :scroll: Description
Screenshot 2022-07-08 at 09 23 53

## :bulb: Motivation and Context
More HTTP details for the OkHttp integration.
Context: https://medium.com/okcredit/how-okcredit-android-app-boosted-network-performance-by-30-84109080c065

## :pencil: Checklist

- [ ] I reviewed the submitted code
- [ ] I added tests to verify the changes
- [ ] I updated the docs if needed
- [ ] No breaking changes

PR #2659, opened 2023-04-17
Add OkHttp event spans

stefanosiano:
## :scroll: Description
added SentryOkHttpEventListener that creates a span on every OkHttp event
added SentryOkHttpEvent

We are increasing the information shown to the user by adding spans for OkHttp events, including dns, https setup, proxies, request headers/body and response headers/body

Unit test

## :pencil: Checklist


stefanosiano:
I just need a confirmation if some data could be seen as PII

marandaneto:
@stefanosiano should we close this then? https://github.com/getsentry/sentry-java/pull/2158

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEvent.kt
```diff
@@ -0,0 +1,138 @@
...
+        callRootSpan?.setData("filtered_url", trimmedUrl)
+        callRootSpan?.setData("host", host)
+        callRootSpan?.setData("path", encodedPath)
+        callRootSpan?.setData("method", method)
+        callRootSpan?.setData("success", true)
+    }
+
+    private fun trimUrl(url: String): String {
+        // Remove any uuid from the url and replace it with a "*"
+        val uuidRegex = Regex("[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}")
```
marandaneto, line 51: M: You can make this static and lazy loaded, `Regex` is CPU expensive.

marandaneto:
@stefanosiano can you attach a screenshot of the result?

stefanosiano:
> @stefanosiano should we close this then? #2158

I think so. I wanted to use and expand that, but there were several differences, so I ended up creating a whole new pr

> @stefanosiano can you attach a screenshot of the result?

Added them in the description

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/GithubAPI.kt
```diff
@@ -1,21 +1,23 @@
...
 import retrofit2.converter.gson.GsonConverterFactory
 
 object GithubAPI {
 
-    private val client = OkHttpClient.Builder().addInterceptor(
-        SentryOkHttpInterceptor(
-            captureFailedRequests = true,
-            failedRequestStatusCodes = listOf(
-                HttpStatusCodeRange(400, 599)
+    private val client = OkHttpClient.Builder().eventListener(SentryOkHttpEventListener())
```
romtsn, line 12: `m`: I think we should create a follow-up task to auto-instrument okhttp and the event listener through our gradle plugin, wdyt? Could you create it please?

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Features
 
+- Add OkHttp event spans ([#2659](https://github.com/getsentry/sentry-java/pull/2659))
```
romtsn, line 7: `m`: Could make the changelog more descriptive? Maybe something like

```suggestion
- Instrument http requests more granularly for OkHttp integration ([#2659](https://github.com/getsentry/sentry-java/pull/2659))
  - Create spans for time spent on:
  	- DNS resolution
  	- Connecting
  	- Requesting headers
  	- ...
```

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/GithubAPI.kt
stefanosiano: Yeah, I was already thinking about it.
The only thing is that there can only be one eventListener, so we should be careful on how to automatically add it.
Definitely doable, I'll open an issue there

romtsn: hrm, does that mean if someone already has a custom event listener, they won't be able to use ours? is it possible for our intereceptor to accept the other listener as a param to delegate all the calls to it?

CHANGELOG.md
marandaneto: M: Also add code snippets on how to set up the listener.

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEvent.kt
```diff
@@ -0,0 +1,139 @@
...
+        val host: String = call.request().url.host
+        val encodedPath: String = call.request().url.encodedPath
+        val method: String = call.request().method
+
+        // We start the call span that will contain all the others
+        callRootSpan = hub.span?.startChild("http.client", "$method $url")
+
+        // We setup a breadcrumb with all meaningful data
+        breadcrumb = Breadcrumb.http(url, method)
+        breadcrumb.setData("url", url)
```
romtsn, line 35: `m`: I don't remember by heart if Relay scrubs PII from the `data` bag, but i'd double-check with the ingest team or somewhere develop docs maybe

romtsn (commented):
After a quick glance looks good, but I'll defer to @markushi for proper review

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/GithubAPI.kt
stefanosiano: yeah, we could make it. Should I add it in this pr?

romtsn: I'd do it, yeah

CHANGELOG.md
stefanosiano: Code snippet is in the Javadoc of the Listener. Is it enough there, or should I add it in the changelog, too?

stefanosiano:
Let's update the span data keys to align with https://develop.sentry.dev/sdk/performance/span-data-conventions/

marandaneto:
@stefanosiano can we actually add the children's span nested to its parents?
For example, `secureConnect` is a child of `connect`, etc.
Since you have the reference of each of them, it's possible, I recall @bruno-garcia asking this as well when I did the first PoC.

stefanosiano:
> @stefanosiano can we actually add the children's span nested to its parents? For example, `secureConnect` is a child of `connect`, etc. Since you have the reference of each of them, it's possible, I recall @bruno-garcia asking this as well when I did the first PoC.

Sure, i'll add it

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEvent.kt
```diff
@@ -0,0 +1,158 @@
...
+
+        // We start the call span that will contain all the others
+        callRootSpan = hub.span?.startChild("http.client", "$method $url")
+
+        urlDetails.applyToSpan(callRootSpan)
+
+        // We setup a breadcrumb with all meaningful data
+        breadcrumb = Breadcrumb.http(url, method)
+        breadcrumb.setData("url", url)
+        breadcrumb.setData("filtered_url", trimmedUrl)
```
stefanosiano, line 45: should we keep this `filtered_url` or should we just put this value in `url`?

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEvent.kt
```diff
@@ -0,0 +1,158 @@
...
+        breadcrumb.setData("host", host)
+        breadcrumb.setData("path", encodedPath)
+        breadcrumb.setData("http.method", method)
```
romtsn, line 48: `m`: should we just change the key here instead? Or is Sentry frontend not aware of this key and won't properly format the description? https://github.com/getsentry/sentry-java/blob/e9bec6ec2c5092673e12b21b008c3abf8e3cccc1/sentry/src/main/java/io/sentry/Breadcrumb.java#L163

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEvent.kt
```diff
@@ -0,0 +1,158 @@
...
+        if (protocolName != null) {
+            breadcrumb.setData("protocol", protocolName)
+            callRootSpan?.setData("protocol", protocolName)
+        }
+    }
+
+    fun setRequestBodySize(byteCount: Long) {
+        if (byteCount > -1) {
+            breadcrumb.setData("http.request_content_length", byteCount)
+            callRootSpan?.setData("http.request_content_length", byteCount)
```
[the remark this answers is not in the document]
romtsn: yeah, I think it's fine. But maybe we should also update the span data conventions docs with what we have added in this PR?

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEvent.kt
```diff
@@ -0,0 +1,158 @@
...
+        val parentSpan = when (event) {
+            // PROXY_SELECT, DNS, CONNECT and CONNECTION are not children of one another
+            SECURE_CONNECT_EVENT -> eventSpans[CONNECT_EVENT]
+            REQUEST_HEADERS_EVENT -> eventSpans[CONNECTION_EVENT]
+            REQUEST_BODY_EVENT -> eventSpans[CONNECTION_EVENT]
+            RESPONSE_HEADERS_EVENT -> eventSpans[CONNECTION_EVENT]
+            RESPONSE_BODY_EVENT -> eventSpans[CONNECTION_EVENT]
+            else -> callRootSpan
+        } ?: callRootSpan
+        val span = parentSpan?.startChild("http.client", event) ?: return
```
romtsn, line 129: `m`: I think we should probably change the span op to something different (maybe `http.client.details`)?. I would assume we have a lot on the product side that is built specifically around `http.client` span op, and this might mess something up when these new spans do not adhere to the `http.client` spans structure (like method + url in the description, for example).

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEvent.kt
romtsn: Also, just my personal preference, but I'm not sure if this looks aligned - I guess I would either prepend every key with `http.` here or just remove the `http.` prefix altogether (we don't have breadcrumb data conventions I believe). Do you think that makes sense?
image

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEvent.kt
```diff
@@ -0,0 +1,158 @@
...
+import io.sentry.android.okhttp.SentryOkHttpEventListener.Companion.RESPONSE_HEADERS_EVENT
+import io.sentry.android.okhttp.SentryOkHttpEventListener.Companion.SECURE_CONNECT_EVENT
+import io.sentry.util.UrlUtils
+import okhttp3.Call
+import okhttp3.Response
+import java.net.URL
+
+private val uuidRegex by lazy { Regex("[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}") }
+
+internal class SentryOkHttpEvent(private val hub: IHub, private val call: Call) {
```
romtsn, line 23: nit: do we still have to hold reference to the `call` variable? We're only using it in the `init` block, so I think it's not necessary
```suggestion
internal class SentryOkHttpEvent(private val hub: IHub, call: Call) {
```

romtsn, line 23: nit: we could just pass a request here, right, we don't need the entire call?

romtsn (approved):

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEvent.kt
stefanosiano: yep

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEvent.kt
stefanosiano: the difference is basically removing any uuid from the url. 
I copied it from an old pr from Manoel, and thought it could make sense to add it here, but now I'm not so sure. No one complained, so i guess there's no concern of PII here, right?
Guess I can remove it in this case

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEvent.kt
stefanosiano: makes sense 👍

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEvent.kt
stefanosiano: data convention should apply to span only, so we can remove the `http.` prefix, to make it consistent to the previous implementation (the `SentryOkHttpInterceptor`)

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEvent.kt
romtsn: I'm fine with keeping it as well, but I'd then just put it into `url` - otherwise I don't see how we're preventing from leaking PII if we're still reporting the plain url alongside the filtered one :) Also, instead of just removing the uuid, I'd replace it with `[Filtered]` as we do for other stuff.

But I'm leaning towards just keeping it as is and not stripping the uuid

marandaneto: Same as https://github.com/getsentry/sentry-java/pull/2659#discussion_r1175044364
We don't need `filtered_url` since the `url` will be split out in different metadata and Relay will strip out what is PII, check RFC and Dart PR as an example.

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEvent.kt
```diff
@@ -0,0 +1,142 @@
...
+        val parentSpan = when (event) {
+            // PROXY_SELECT, DNS, CONNECT and CONNECTION are not children of one another
+            SECURE_CONNECT_EVENT -> eventSpans[CONNECT_EVENT]
+            REQUEST_HEADERS_EVENT -> eventSpans[CONNECTION_EVENT]
+            REQUEST_BODY_EVENT -> eventSpans[CONNECTION_EVENT]
+            RESPONSE_HEADERS_EVENT -> eventSpans[CONNECTION_EVENT]
+            RESPONSE_BODY_EVENT -> eventSpans[CONNECTION_EVENT]
+            else -> callRootSpan
+        } ?: callRootSpan
+        val span = parentSpan?.startChild("http.client.details", event) ?: return
```
marandaneto, line 113: What does `details` mean here? it's not part of the spec right.

stefanosiano: `http.client` is the operation of the "root" span of the http call only.
Based on [this comment](https://github.com/getsentry/sentry-java/pull/2659#discussion_r1196974000) we thought it makes sense to change the operation of the sub-spans in `http.client.details`

marandaneto: It's just an assumption, though I'd rather ask if that is the case before doing the change (because in the end, they are all part of the HTTP call anyway), the spec is there to be followed, if indeed we need a new one, let's add it to the spec.
Also `details` is too generic, I'd rather go with `.dns`, `.connect`, etc, because that would make sense for pretty much every platform.

romtsn: Not an assumption, here you can see a bunch of entries for `http.client` in particular for starfish https://github.com/search?q=repo%3Agetsentry%2Fsentry%20%22http.client%22&type=code

E.g. this one doesn't account that there's something different in the description https://github.com/getsentry/sentry/blob/67c64a7a7743b384b6c5301a50ccd01b312b2aa0/static/app/components/events/interfaces/spans/utils.tsx#L288-L302

romtsn: also, checked python,php,ruby - all of them adhere to "method + url" pattern in the description, so I guess we should be cautious here. And also we should update the spec with that, if we're relying on that pattern on the product side

marandaneto: I understand that Sentry itself relies on the `http.client` op, but my question is, adding more spans with that op which is an inner span of an HTTP call anyway (which is also a http.client op), why would that be a problem? that is my question?
I'd not change things just because we are assuming that we will have a problem.
And again, not against adding a suffix, I just think that `details` is too generic as well since we clearly know what is a connect, a dsn, a proxy, etc.

romtsn: the problem would be that these new `http.client` spans do not contain `method + url` in their description, as far as I found out we rely on that on the product side

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEvent.kt
```diff
@@ -0,0 +1,142 @@
...
+
+        urlDetails.applyToSpan(callRootSpan)
+
+        // We setup a breadcrumb with all meaningful data
+        breadcrumb = Breadcrumb.http(url, method)
+        breadcrumb.setData("url", url)
+        breadcrumb.setData("host", host)
+        breadcrumb.setData("path", encodedPath)
+        breadcrumb.setData("method", method)
+        breadcrumb.setData("success", true)
```
marandaneto, line 44: https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/
`status_code` already implies the `success`

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEvent.kt
marandaneto: Why do we set `url` again?
`Breadcrumb.http` already does it.

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEvent.kt
```diff
@@ -0,0 +1,142 @@
...
+        callRootSpan = hub.span?.startChild("http.client", "$method $url")
+
+        urlDetails.applyToSpan(callRootSpan)
+
+        // We setup a breadcrumb with all meaningful data
+        breadcrumb = Breadcrumb.http(url, method)
+        breadcrumb.setData("url", url)
+        breadcrumb.setData("host", host)
+        breadcrumb.setData("path", encodedPath)
+        breadcrumb.setData("method", method)
```
marandaneto, line 43: Same as https://github.com/getsentry/sentry-java/pull/2659/files#r1202080681

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEvent.kt
```diff
@@ -0,0 +1,142 @@
...
+    /**
+     * Sets the [Response] that will be sent in the breadcrumb [Hint].
+     * Also, it sets the protocol and status code in the breadcrumb and the call root span.
+     */
+    fun setResponse(response: Response) {
+        this.response = response
+        breadcrumb.setData("protocol", response.protocol.name)
+        breadcrumb.setData("status_code", response.code)
+        callRootSpan?.setData("protocol", response.protocol.name)
+        callRootSpan?.setData("status_code", response.code)
```
marandaneto, line 63: The key is different https://github.com/getsentry/team-mobile/issues/121

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEvent.kt
```diff
@@ -0,0 +1,142 @@
...
+
+    /**
+     * Sets the success flag in the breadcrumb and the call root span to false.
+     * Also sets the [errorMessage] if not null.
+     */
+    fun setError(errorMessage: String?) {
+        breadcrumb.setData("success", false)
+        callRootSpan?.setData("success", false)
+        if (errorMessage != null) {
+            breadcrumb.setData("error_message", errorMessage)
```
marandaneto, line 96: This is probably the `reason` field? https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEvent.kt
```diff
@@ -0,0 +1,142 @@
...
+import io.sentry.android.okhttp.SentryOkHttpEventListener.Companion.REQUEST_HEADERS_EVENT
+import io.sentry.android.okhttp.SentryOkHttpEventListener.Companion.RESPONSE_BODY_EVENT
+import io.sentry.android.okhttp.SentryOkHttpEventListener.Companion.RESPONSE_HEADERS_EVENT
+import io.sentry.android.okhttp.SentryOkHttpEventListener.Companion.SECURE_CONNECT_EVENT
+import io.sentry.util.UrlUtils
+import okhttp3.Request
+import okhttp3.Response
+
+internal class SentryOkHttpEvent(private val hub: IHub, private val request: Request) {
+    private val eventSpans: MutableMap<String, ISpan> = HashMap()
```
marandaneto, line 21: M: Should this be thread-safe? can it be called from multiple threads?

marandaneto (dismissed):
@stefanosiano I've left a few names, there's a spec for crumbs and spans, let's use the very same keys as per spec.

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEvent.kt
stefanosiano: I'm not sure. `reason` is a `Text that describes the status code`. I think it refers to things like `Not found` or `Bad Method`, rather than the error that caused that status code.
This `error_message` is the actual message returned by the Exception

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEvent.kt
stefanosiano: it would be ok to change the op to `http.client.dns`, `http.client.connection`, etc. 
But what would we put in the description in that case? Are we fine with having a repetition?
E.g. op = `http.client.connection`, description = `connection`

marandaneto: `description` is an optional field, if there's nothing useful (specific to the action), just leave it, if there's something useful, then we put something.

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEvent.kt
marandaneto: True, makes sense, lets keep it.

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEvent.kt
stefanosiano: This is called by the event listener only, and I think each "event" call is done in a specific thread per `Call`. However, using a `ConcurrentHashMap` makes no harm here

stefanosiano:
@marandaneto would you take another look?

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEventListener.kt
```diff
@@ -0,0 +1,370 @@
...
+        internal const val DNS_EVENT = "dns"
+        internal const val SECURE_CONNECT_EVENT = "secureConnect"
+        internal const val CONNECT_EVENT = "connect"
+        internal const val CONNECTION_EVENT = "connection"
+        internal const val REQUEST_HEADERS_EVENT = "requestHeaders"
+        internal const val REQUEST_BODY_EVENT = "requestBody"
+        internal const val RESPONSE_HEADERS_EVENT = "responseHeaders"
+        internal const val RESPONSE_BODY_EVENT = "responseBody"
+
+        internal val eventMap: MutableMap<Call, SentryOkHttpEvent> = HashMap()
```
marandaneto, line 59:
```suggestion
        internal val eventMap: MutableMap<Call, SentryOkHttpEvent> = ConcurrentHashMap()
```
Do we need thread safe?

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEventListener.kt
```diff
@@ -0,0 +1,370 @@
...
+ * ```
+ * val client = OkHttpClient.Builder()
+ *     .eventListener(SentryOkHttpEventListener(myEventListener))
+ *     .addInterceptor(SentryOkHttpInterceptor())
+ *     .build()
+ * ```
+ */
+@Suppress("TooManyFunctions")
+class SentryOkHttpEventListener(
+    private val hub: IHub = HubAdapter.getInstance(),
```
marandaneto, line 42: Java will be forced to set `hub`, make another ctor that does not require it.

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEvent.kt
```diff
@@ -0,0 +1,134 @@
...
+        callRootSpan?.setData("http.method", method)
+    }
+
+    /**
+     * Sets the [Response] that will be sent in the breadcrumb [Hint].
+     * Also, it sets the protocol and status code in the breadcrumb and the call root span.
+     */
+    fun setResponse(response: Response) {
+        this.response = response
+        breadcrumb.setData("protocol", response.protocol.name)
```
marandaneto, line 57: String keys could be constants if they are repeated.
https://en.wikipedia.org/wiki/Defensive_programming

sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEventListener.kt
```diff
@@ -0,0 +1,370 @@
...
+    private val originalEventListenerCreator: ((call: Call) -> EventListener)? = null
```
marandaneto, line 43: Why is it called `original`?

stefanosiano: This is the listener that was (optionally) originally set by the user. We could name it in another way, like `delegateEventListener` if that's preferable

markushi:
@marandaneto could you unblock this PR - we want to merge and ship this today 😊

marandaneto:
> happy

It's not blocked, https://github.com/getsentry/sentry-java/pull/2659#pullrequestreview-1443264718
Only one approval is needed.

romtsn:
> > happy
> 
> It's not blocked, [#2659 (review)](https://github.com/getsentry/sentry-java/pull/2659#pullrequestreview-1443264718) Only one approval is needed.

I think you have to approve it actually
image

PR #2667, opened 2023-04-21
fix(ci): Move aar maven fix out of buildSrc

romtsn:
_#skip-changelog_

## :scroll: Description

A fix for the previous PR that seems to broke CI https://github.com/getsentry/sentry-java/pull/2641. Probably because the classpath from buildSrc was leaking into the main classpath and some AGP stuff broke.


build.gradle.kts
```diff
@@ -224,3 +224,70 @@ gradle.taskGraph.whenReady {
...
+    "sentry-android-navigation",
+    "sentry-android-okhttp",
+    "sentry-android-timber",
+    "sentry-compose-android"
+)
+
+private val androidXLibs = listOf(
+    "androidx.core:core",
+    "androidx.lifecycle:lifecycle-process",
+    "androidx.lifecycle:lifecycle-common-java8"
```
markushi, line 241: would it make sense to just check for the group `androidx` here?

markushi (approved):

build.gradle.kts
romtsn: we could, but I just spotted that `androidx.lifecycle` are pure `jars` and not `aars`, so this is already incorrect 🤣 I guess we'll have to maintain the hardcoded list here, but I doubt we'll depend on many more androidx libs (and to be correct, `androidx.core` should be `compileOnly`, I just didn't dare to change it without testing which I don't have time for :D). 

Let's look out for androidx dependencies in the future and try to make them compileOnly rather than maintaining this list.

PR #2677, opened 2023-04-26
Fix remove saucelab config name attribute, as it does not exist anymore

markushi:
Use `build` attribute instead which describes which module was used to run the tests.

#skip-changelog


stefanosiano:
Would it be possible to do something like [this sample](https://github.com/saucelabs/saucectl-testcafe-example/blob/main/.sauce/config.yml#L11), if it's worth it? wdyt?

stefanosiano (approved):

PR #2679, opened 2023-04-27
Fix crash on double SDK init

stefanosiano:
## :scroll: Description
all calls to ISentryExecutorService are now wrapped in a try catch block
closing the hub will close the profiler and the performance collector

Calling `init()` multiple times closes any previous instance of the SDK. 
After closing, any call to the executor service which was previously scheduled resulted in a crash due to `RejectedExecutionException`.
Fixes https://github.com/getsentry/sentry-java/issues/2493

Unit and ui test

## :pencil: Checklist

### DIFF
--- a/sentry/src/main/java/io/sentry/Hub.java
+++ b/sentry/src/main/java/io/sentry/Hub.java
@@ -339,6 +339,21 @@ public final class Hub implements IHub {
             ((Closeable) integration).close();
           }
         }
+
+        withScope(
+            scope -> {
+              ITransaction transaction = scope.getTransaction();
+              if (transaction != null) {
+                for (Span span : transaction.getSpans()) {
+                  span.setSpanFinishedCallback(null);
+                  span.finish(SpanStatus.CANCELLED);
+                }
+                transaction.finish(SpanStatus.CANCELLED);
+              }
+              scope.clear();
+            });
+        options.getTransactionProfiler().close();
+        options.getTransactionPerformanceCollector().close();
         options.getExecutorService().close(options.getShutdownTimeoutMillis());
 
         // Close the top-most client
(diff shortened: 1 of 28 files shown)


sentry/src/main/java/io/sentry/Hub.java
```diff
@@ -339,6 +339,21 @@ public void close() {
...
+
+        withScope(
+            scope -> {
+              ITransaction transaction = scope.getTransaction();
+              if (transaction != null) {
+                for (Span span : transaction.getSpans()) {
+                  span.setSpanFinishedCallback(null);
+                  span.finish(SpanStatus.CANCELLED);
+                }
+                transaction.finish(SpanStatus.CANCELLED);
```
markushi, line 351: You can use `ITransaction.forceFinish()` here instead as it will basically do the same.

markushi (changes requested):
Looking good already, please have a look at my comments!

sentry/src/main/java/io/sentry/Hub.java
stefanosiano: Always learning new things 😅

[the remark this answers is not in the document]
stefanosiano: we are dropping canceling the transaction for the moment, as we have to be careful on the backend use case.
Worst case we will show some errors in the logs, pointing at a possible `Sentry.close()` previous call.
In another pr we will reevaluate cancelling the current transactions.

markushi (approved):
Nice, looks good to me!

sentry/src/main/java/io/sentry/Hub.java
```diff
@@ -339,6 +339,10 @@ public void close() {
             ((Closeable) integration).close();
           }
         }
+
+        withScope(scope -> scope.clear());
```
romtsn, line 343: late to the party, but this is not gonna do much, because `withScope` will duplicate the scope, and then we're trying to clear the duplicated scope, but the main scope stays untouched. This should be changed to `configureScope`. I think we need a custom lint rule to never use withScope on mobile 😅 this is the second time we introduced something like this (this one is not as severe though).

Issue #2649, opened 2023-04-12
In situations where the network is slow, ANR often occurs


yaohuzhou:
### Integration

sentry-android

### Build System

Gradle

### AGP Version

7.3.1

### Proguard

Disabled

### Version

6.17.0

### Steps to Reproduce

**Reproduction steps:**

**1. Initialize SDK.**

`class TestApplication : Application() {

    override fun onCreate() {
        super.onCreate()
        SentryAndroid.init(this) { options ->
            options.isDebug = BuildConfig.DEBUG
            options.dsn = "https://examplePublicKey@o0.ingest.sentry.io/0"
            options.isEnableSystemEventBreadcrumbs = false
            options.isEnableUncaughtExceptionHandler = true
            options.isEnableNdk = true
            options.isAnrEnabled = true
            options.maxRequestBodySize = SentryOptions.RequestSize.MEDIUM
            options.isAnrReportInDebug = false
            options.tracesSampleRate = 1.0
            options.profilesSampleRate = 1.0
        }
    }
}`
 
**2.Simulate a weak network environment using Charles proxy.**
Throttle Settings-> Bandwidth (kbps): Download = 1 Upload = 1
Start Throttling

**3.Repeat opening the app multiple times and wait for about 20 seconds.**

**4.ANR**

![image](https://user-images.githubusercontent.com/19381111/231422524-7a5474d9-d0d6-4e06-ba3a-25af1f78fe4a.png)

### Expected Result

There will be no ANR. 

### Actual Result

ANR often occurs.

![image](https://user-images.githubusercontent.com/19381111/231422524-7a5474d9-d0d6-4e06-ba3a-25af1f78fe4a.png)

yaohuzhou:
The bug can be reproduced using the demo.
[Demo.zip](https://github.com/getsentry/sentry-java/files/11210237/Demo.zip)

Charles Throttle Settings
![image](https://user-images.githubusercontent.com/19381111/231430627-206b6713-a50a-4ceb-ad79-305d662711c2.png)

romtsn:
@yaohuzhou hi, thanks for the report. Can you share the full stacktrace/thread dump of the ANR please? It's not clear why you think our SDK is the cause.

yaohuzhou:
收到

yaohuzhou:
> @yaohuzhou hi, thanks for the report. Can you share the full stacktrace/thread dump of the ANR please? It's not clear why you think our SDK is the cause.
[...]

PR #2691, opened 2023-05-03
Android Profiler on calling thread

stefanosiano:
## :scroll: Description
AndroidTransactionProfiler now starts/stops profiles without offloading to the executorService
removed any file checks due to the code running possibly on the main thread
wrapped Debug methods into try/catch to compensate lack of file checks

If the executor service is very busy, the profiler would start with some delay, and when finishing it would wait for the executor to finish, possibly causing ANRs.
fixes https://github.com/getsentry/sentry-java/issues/2649

Unit test

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixes
 
+- Android Profiler on calling thread ([#2691](https://github.com/getsentry/sentry-java/pull/2691))
 - Use `configureScope` instead of `withScope` in `Hub.close()`. This ensures that the main scope releases the in-memory data when closing a hub instance. ([#2688](https://github.com/getsentry/sentry-java/pull/2688))
 
 ### Dependencies
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
@@ -32,7 +32,6 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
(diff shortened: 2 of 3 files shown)


romtsn (approved):
LGTM, but in the original issue this was somehow related to slow network connection. Have you managed to reproduce it? I wasn't sure how the slow connectivity can be the reason here, because our HttpConnection uses a different thread pool (also single-threaded), so in theory they should not conflict with SentryExecutorService

stefanosiano:
> LGTM, but in the original issue this was somehow related to slow network connection. Have you managed to reproduce it? I wasn't sure how the slow connectivity can be the reason here, because our HttpConnection uses a different thread pool (also single-threaded), so in theory they should not conflict with SentryExecutorService

The `SentryExecutorService` is used by `SendCachedEnvelopeIntegration` and `SendCachedEnvelopeFireAndForgetIntegration`.
I think in case of slow connection envelopes are cached to disk and sent, but there is a wait in the [processing of the envelope file](https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/EnvelopeSender.java#L70)
Does it make sense or am i missing something?
At the moment the `SentryExecutorService` is not used for any critical things anymore (except scheduling auto finish of profiling, ttfd and cpu/mem collection, but that's more to cover edge cases), but we could think about using an executor only for the envelope handling.

romtsn:
@stefanosiano ah gotcha, that makes sense, yes! And then we were blocking on the profiler, although the executor was already busy. I think it's fine to use the executor as long as we don't call blocking `.get` on the main thread (except the startup crash case, where we block for 5s max), so this looks good

helloword1:
When will the new version be released to fix the issue?

PR #2695, opened 2023-05-04
Add support for Sentry Kotlin Compiler Plugin

markushi:
## :scroll: Description
Counterpart for Kotlin Compiler Plugin

- Adds compose support for automatic user interactions tracking
- Add View Hierarchy support

## :pencil: Checklist


sentry-android-core/src/main/java/io/sentry/android/core/ViewHierarchyEventProcessor.java
```diff
@@ -89,7 +93,9 @@ public static byte[] snapshotViewHierarchyAsData(
 
   @Nullable
   public static ViewHierarchy snapshotViewHierarchy(
-      @Nullable Activity activity, @NotNull ILogger logger) {
+      final @Nullable Activity activity,
+      final @NotNull ILogger logger,
+      final @NotNull List<ViewHierarchyExporter> exporters) {
```
romtsn, line 98: `m`: again this is a breaking change for RN - probably we should have overloads here rather? https://github.com/getsentry/sentry-react-native/blob/2d064085c1bb8e07ee516ecc6711232f42fbc5cc/android/src/main/java/io/sentry/react/RNSentryModuleImpl.java#L411

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
         - Auxiliary information (such as current memory load) at the time of ANR event.
       - If you would like us to provide support for the old approach working alongside the new one on Android 11 and above (e.g. for raising events for slow code on main thread), consider upvoting [this issue](https://github.com/getsentry/sentry-java/issues/2693).
     - The old watchdog implementation will continue working for older API versions (Android < 11)
+- Add support for Sentry Kotlin Compiler Plugin ([#2695](https://github.com/getsentry/sentry-java/pull/2695))
```
romtsn, line 17: `m`: Now it's the time for a detailed changelog :)

sentry-compose-helper/src/jvmMain/java/io/sentry/compose/viewhierarchy/ComposeViewHierarchyExporter.java
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  @Override
+  public boolean export(@NotNull final ViewHierarchyNode parent, @NotNull final Object element) {
+
+    if (!(element instanceof Owner)) {
+      return false;
+    }
+
+    final SentryComposeHelper composeHelper = new SentryComposeHelper(logger);
```
romtsn, line 37: `l`: I think this can be a class property, right?

sentry-compose-helper/src/jvmTest/java/io/sentry/compose/viewhierarchy/ComposeViewHierarchyExporterTest.java
romtsn, line 1: `l`: we usually tend to write tests in Kotlin, can we convert this one as well?

romtsn (approved):
Code-wise this looks good, but I have few rather conceptual things:

* When I tried integrating kotlin-compiler-plugin into the sample, I've faced a compilation error, because our sample uses Kotlin 1.8, and the compiler plugin seems to support only 1.8.20 onwards (after updating the kotlin version, it worked). The error was the same one as https://github.com/drewhamilton/Poko/issues/122. The question is if we want to support lower version of Kotlin, or we just say we support the latest one onwards?
* I thought of moving `SentryModifier` and `SentryComposeTracing` to `commonMain` if possible, but I guess we could do it later when we start looking into compose-multiplatform support
* Last thing - I've tried integrating the compiler plugin into our sample, but unfortunately I don't think it brings much value in the current state (unless I misunderstood how it all should work in the end). E.g. for our sample on the screenshot I've removed all of the custom `SentryTraced` wrappers and just left the compiler plugin do its job and it only inferred one tag (Landing). I guess we should rethink our instrumentation a bit for the view hierarchy case 😅 
image

markushi:
> Code-wise this looks good, but I have few rather conceptual things:
> 
> * When I tried integrating kotlin-compiler-plugin into the sample, I've faced a compilation error, because our sample uses Kotlin 1.8, and the compiler plugin seems to support only 1.8.20 onwards (after updating the kotlin version, it worked). The error was the same one as [Kotlin 1.8.20 support drewhamilton/Poko#122](https://github.com/drewhamilton/Poko/issues/122). The question is if we want to support lower version of Kotlin, or we just say we support the latest one onwards?
> * I thought of moving `SentryModifier` and `SentryComposeTracing` to `commonMain` if possible, but I guess we could do it later when we start looking into compose-multiplatform support
> * Last thing - I've tried integrating the compiler plugin into our sample, but unfortunately I don't think it brings much value in the current state (unless I misunderstood how it all should work in the end). E.g. for our sample on the screenshot I've removed all of the custom `SentryTraced` wrappers and just left the compiler plugin do its job and it only inferred one tag (Landing). I guess we should rethink our instrumentation a bit for the view hierarchy case 😅
> 
> image

1) Yes, this is a known issue. The only alternative would be to release different versions of our kotlin compiler plugin and have a compatibility matrix, just like [Jetpack Compose has](https://developer.android.com/jetpack/androidx/releases/compose-kotlin), but as of now this would also mean having different versions of our AGP as well. I'd stick to the `1.8.20` requirement.
2) Sounds good!
3) This looks odd, there should definitely be more metadata in there. Let me have a closer look at this.

sentry-compose-helper/src/jvmTest/java/io/sentry/compose/viewhierarchy/ComposeViewHierarchyExporterTest.java
markushi: Unfortunately we can't, as we're accessing classes marked as Kotlin internal 😉

romtsn: ahh, makes sense indeed 👍

sentry-compose-helper/src/jvmMain/java/io/sentry/compose/viewhierarchy/ComposeViewHierarchyExporter.java
romtsn: but calling it here everytime would mean we'll actually do the reflection bits potentially in a loop (addChildren), right? And also for each error where we capture a view hierarchy this also will be called every time. 

We could also extract it and pass it as a ctor parameter to both ComposeViewHierarchyExporter and ComposeGestureTargetLocator, hence instantiating it only once and for all.

markushi: No, not really - this is guarded by the if check beforehand. It will only run once per AndroidComposeView. Since we setup the options on app startup, passing it via ctor also means performing reflection on SDK init - and I wouldn't do that. But if you want we could init lazy init it within the `export` call

romtsn: yeah I think lazy init would do. There's still the case that we'll instantiate this every time there's an error and we capture the view hierarchy in compose right?

markushi: I've added a lazy-init mechanism now, only on the first time when VH is accessed `SentryComposeHelper` is create, following calls will just re-use the same instance then.

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Features
+- Add support for Sentry Kotlin Compiler Plugin ([#2695](https://github.com/getsentry/sentry-java/pull/2695))
+  - In conjunction with our sentry-kotlin-compiler-plugin we improved Jetpack Compose support for
```
romtsn, line 7: nit: Ideally here would be also a link to sentry-kotlin-compiler-plugin docs, but we can update it later ;)

tfcporciuncula: I'd really appreciate a link to the compiler plugin docs as I've been meaning to try these new features and haven't been able to understand how to use the compiler plugin.

romtsn: @tfcporciuncula they are in the making, but soon will be there :) https://github.com/getsentry/sentry-docs/pull/7026

PR #2701, opened 2023-05-05
Open up some transaction methods

### LINKED ISSUE #2640: Open up Transaction Factory methods
### Problem Statement

Perhaps there is a workaround, but I would like to create a slight variation of [ActivityLifecycleIntegration](https://github.com/getsentry/sentry-java/blob/9a46975a2e61106bc75c4c2075ffcefa3f8fea41/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java).

It seems that this integration is the way Sentry creates the App Start Span that times app start up before the first activity is created and the integration callback is issued.  This integration calls the method below on the Hub to use TransactionOption to set the start time of the transaction to that of the App Start Span. 

```kotlin
@ApiStatus.Internal
  @NotNull
  ITransaction startTransaction(
      final @NotNull TransactionContext transactionContext,
      final @NotNull TransactionOptions transactionOptions);
```

### Solution Brainstorm

Rather than the App Start Span being attached to a transaction for `ui.load`, and thus attached to the first activity, I'd like it to be attached to a transaction that I dictate.   I currently have a scoped transaction that I think will "block" the ui.load one from posting so I can't use this integration as-is (unless I missed something).

As you can see above, the @ApiStatus of the method is internal (as is TransactionOptions) and I don't want to depend on internal APIs.  Is there a workaround you could suggest or do you have any interesting in publicizing these methods and types?  Or perhaps could we separate out the integrations?

lbloder:
## :scroll: Description
Open up some transaction methods, so that they can be used from outside sentry
Add new `startChild` method to set timestamp

Fixes #2640 

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
         - Auxiliary information (such as current memory load) at the time of ANR event.
       - If you would like us to provide support for the old approach working alongside the new one on Android 11 and above (e.g. for raising events for slow code on main thread), consider upvoting [this issue](https://github.com/getsentry/sentry-java/issues/2693).
     - The old watchdog implementation will continue working for older API versions (Android < 11)
+- Open up transaction methods ([#2701](https://github.com/getsentry/sentry-java/pull/2701))
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,22 @@
 
 ## Unreleased
 
+### Features
+
+- Add Screenshot and ViewHierarchy to integrations list ([#2698](https://github.com/getsentry/sentry-java/pull/2698))
+- New ANR detection based on [ApplicationExitInfo API](https://developer.android.com/reference/android/app/ApplicationExitInfo) ([#2697](https://github.com/getsentry/sentry-java/pull/2697))
+    - This implementation completely replaces the old one (based on a watchdog) on devices running Android 11 and above:
+      - New implementation provides more precise ANR events/ANR rate detection as well as system thread dump information. The new implementation reports ANRs exactly as Google Play Console, without producing false positives or missing important background ANR events.
+      - However, despite producing many false positives, the old implementation is capable of better enriching ANR errors (which is not available with the new implementation), for example:
+        - Capturing screenshots at the time of ANR event;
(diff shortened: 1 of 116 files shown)


sentry/src/main/java/io/sentry/ITransaction.java
```diff
@@ -41,6 +41,10 @@ public interface ITransaction extends ISpan {
   @TestOnly
   List<Span> getSpans();
 
+  @NotNull
+  ISpan startChild(
```
romtsn, line 45: `l`: can we add some javadoc since it's public?

sentry/src/main/java/io/sentry/IHub.java
```diff
@@ -494,7 +494,6 @@ ITransaction startTransaction(
     return startTransaction(name, operation, null);
   }
 
-  @ApiStatus.Internal
   @NotNull
   ITransaction startTransaction(
```
romtsn, line 498: `l`: same here, some javadoc would be nice

romtsn (approved):

lbloder:
Hey @romtsn,
Thanks for reviewing.

I missed removing the `@ApiStatus.Internal` annotation from the `TransactionOptions` class, which I have done now and also added javadoc comments to it. Do you mind having another quick glance?

Thanks

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
         - Auxiliary information (such as current memory load) at the time of ANR event.
       - If you would like us to provide support for the old approach working alongside the new one on Android 11 and above (e.g. for raising events for slow code on main thread), consider upvoting [this issue](https://github.com/getsentry/sentry-java/issues/2693).
     - The old watchdog implementation will continue working for older API versions (Android < 11)
+- Open up transaction methods ([#2701](https://github.com/getsentry/sentry-java/pull/2701))
```
romtsn, line 17:
```suggestion
- Open up `TransactionOptions`, `ITransaction` and `IHub` methods allowing consumers modify start/end timestamp of transactions and spans ([#2701](https://github.com/getsentry/sentry-java/pull/2701))
```

romtsn:
looks good, thanks @lbloder !

Issue #2674, opened 2023-04-24
Sentry screenshotting seems to cause crashes with Compose


smartfuse:
### Integration

sentry-android

### Build System

Gradle

### AGP Version

7.4.0

### Proguard

Disabled

### Version

6.12.1

### Steps to Reproduce

1. Enable Sentry Android screenshots in an app with Compose views
2. When a non-fatal occurs on a screen with Compose views, a Compose crash has a high likelihood of happening.
3. Crashes are usually Segfaults or Compose inconsistency crashes. It seems that Sentry is attempting to take a screenshot of a Compose view while it's detached from the view hierarchy which Compose doesn't seem to support.
4. When we turned off screenshots, non-fatals were still properly reported but the crashes stopped occurring.

<img width="1092" alt="Screen Shot 2023-04-17 at 3 08 14 PM" src="https://user-images.githubusercontent.com/5400589/234077621-e8f27dd2-a1d1-4036-ab81-fcaac70cb43a.png">

### Expected Result

App crash rate should not increase when screenshots are turned on.

### Actual Result

App crash rate is high since every non-fatal triggers a potential crash.

markushi:
@smartfuse Thanks for reporting! Could you provide us with the following details:
* Do you use [any of our Jetpack Compose features](https://docs.sentry.io/platforms/android/configuration/integrations/jetpack-compose/)?
* Could you provide some details about your non-fatal errors? Any chance those are happening on a background thread?

helloword1:
I have the same problem。

markushi:
@smartfuse any chance you could take a look at my questions above?

We're having some discussions around this topic (see similar issue: https://github.com/getsentry/sentry-java/issues/2665) and would love to have some more context around this.

smartfuse:
@markushi Sorry for the late response. It looks like notifications aren't quite working for my personal Github account so I didn't see your follow-up.

Some more details:
* We are not using any Sentry Compose features
* Crashes look something like this (usually Compose complaining about an invalid state change or attempting to measure a disposed Composable, we also had Segfaults for libart and libhwui). These crashes would happen on the main thread.

<img width="300px" alt="Screen Shot 2023-05-08 at 7 41 41 PM" src="https://user-images.githubusercontent.com/5400589/236980422-f6f70901-f744-4327-b58f-7a9ec4ea808e.png">
[...]

Issue #2515, opened 2023-02-03
Suggestions for internals refactoring


marandaneto:
### Description

Some places use `System.currentTimeMillis()` directly (such as `AndroidMemoryCollector`) instead of `CurrentDateProvider` which makes it untestable.

`DefaultAndroidEventProcessor` already reads memory data, but now we also have `AndroidMemoryCollector`, maybe we could refactor `DefaultAndroidEventProcessor` to use `AndroidMemoryCollector`.

`CurrentActivityHolder` holds the current activity thru the `CurrentActivityIntegration`, maybe we could refactor other integrations such as `ActivityLifecycleIntegration` to use the holder as well.

marandaneto:
`Looper.getMainLooper()` Also being used directly instead of `MainLooperHandler` for testability.

marandaneto:
@romtsn @markushi added to milestone v7, let me know if that's fine.
I am afraid that we will keep using the static methods making the code difficult to be tested.

markushi:
Let's remove the milestone and re-priortize this.

markushi:
Let's check if there are any breaking API changes due to the mentioned refactorings.

romtsn:
Let's call each other out when we review PRs and try to fix this as we go

markushi:
> Let's call each other out when we review PRs and try to fix this as we go

Let's close this one for now and keep that in mind.

PR #2712, opened 2023-05-16
Ensure screenshots and view hierarchies are captured on the main thread

markushi:
## :scroll: Description

Right now screenshots/VH are created on the calling thread. According [to the docs](https://developer.android.com/guide/components/processes-and-threads#Threads) it's not safe to modify or just access the UI toolkit from a background thread.

Jetpack Compose specifically may makes things worse, as `View.draw()` [might initiates a layout pass](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/ui/ui/src/androidMain/kotlin/androidx/compose/ui/platform/AndroidComposeView.android.kt;l=1064;drc=63f7e4ae6366ee195b0586229a292abfb7739710), causing modifications on the calling thread.

Fixes https://github.com/getsentry/sentry-java/issues/2674

Add unit tests

## :pencil: Checklist


stefanosiano (approved):

sentry-android-core/src/main/java/io/sentry/android/core/ViewHierarchyEventProcessor.java
```diff
@@ -108,12 +117,28 @@ public static ViewHierarchy snapshotViewHierarchy(
...
 
     try {
-      final @NotNull ViewHierarchy viewHierarchy = snapshotViewHierarchy(decorView);
-      return viewHierarchy;
+      if (mainThreadChecker.isMainThread()) {
+        return snapshotViewHierarchy(decorView);
+      } else {
+        final CountDownLatch latch = new CountDownLatch(1);
+        final AtomicReference<ViewHierarchy> viewHierarchy = new AtomicReference<>(null);
+        activity.runOnUiThread(
```
romtsn, line 125: `l`: potentially we could do `handler.postAtFrontOfQueue` instead to capture the VH/SS as soon as possible, but I don't know what could be the potential side effects of this. WDYT?

romtsn (approved):
Looks good, but it's a breaking change for react native. @krystofwoldrich is that fine for you to just adapt it with the new Android SDK version, or do we have to take a less invasive path (e.g. introducing another method overload to avoid the breaking change)?

sentry-android-core/src/main/java/io/sentry/android/core/ViewHierarchyEventProcessor.java
markushi: Hmm that's a good one. [Reading up on the docs](https://developer.android.com/reference/android/os/Handler#postAtFrontOfQueue(java.lang.Runnable)), I wouldn't do it - at least not within this PR - maybe as a follow up investigation, as it could be relevant for other parts within our code as well?

marandaneto: Avoid `activity.runOnUiThread` in favor of `MainLooperHandler` for testability.
See the previous comment related to this issue as well:
https://github.com/getsentry/sentry-java/pull/2281#discussion_r990935530
Added a new comment here https://github.com/getsentry/sentry-java/issues/2515#issuecomment-1556754682

PR #2731, opened 2023-05-24
Fix AndroidProfiler and SentryTracer crashes

romtsn:
## :scroll: Description

A customer reported these 2 crashes internally, so this PR simply fixes/swallows them. Most likely they were caused by calling `Sentry.close()` in the middle of processing transactions.

*  The `synchronized` in `AndroidTransactionProfiler` should actually fix the root cause, but I've also added the null check just to make sure
* For SentryTracer fix we just swallow the exception and finish the transaction immediately if we failed to schedule the finish timer

image
image

Manually and automated

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
@@ -283,7 +283,7 @@ final class AndroidTransactionProfiler implements ITransactionProfiler {
   }
 
   @SuppressLint("NewApi")
-  private @Nullable ProfilingTraceData onTransactionFinish(
+  private @Nullable synchronized ProfilingTraceData onTransactionFinish(
       final @NotNull ITransaction transaction,
       final boolean isTimeout,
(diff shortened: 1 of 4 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
```diff
@@ -283,7 +283,7 @@ public void onFrameMetricCollected(
   }
 
   @SuppressLint("NewApi")
-  private @Nullable ProfilingTraceData onTransactionFinish(
+  private @Nullable synchronized ProfilingTraceData onTransactionFinish(
```
stefanosiano, line 286: This is called only in the other `onTransactionFinish`, which is already `synchronized`, so i'm not sure it makes a difference here, but it doesn't harm, either

romtsn: mm, not really, it's also called in `close()` and when scheduling the finish timeout, so likely one of those was nullifying the current transaction data

stefanosiano (approved):

PR #2757, opened 2023-05-31
Improve Android specifics for ecosystem 

markushi:
## :scroll: Description

This PR bundles a few Android specific improvements based on some SDK feedback
* Enable non-transitive R class
* create empty `res/public.xml`, automatically marking all resources as private
* create empty `proguard-rules.pro` file for `sentry-android` meta package

## :pencil: Checklist

## :crystal_ball: Next steps

#skip-changelog

### DIFF
--- a/sentry-kotlin-extensions/src/main/res/values/public.xml
+++ b/sentry-kotlin-extensions/src/main/res/values/public.xml
@@ -1,0 +1,2 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources></resources>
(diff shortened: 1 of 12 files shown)


stefanosiano:
[docs](https://developer.android.com/studio/projects/android-library.html#PrivateResources) say that you should add an empty `<public />` tag in the `resource` tag in `public.xml `

sentry-kotlin-extensions/src/main/res/values/public.xml
romtsn, line 1: `l`: I think sentry-kotlin-extensions is published as a jar, so this one doesn't make any effect here?

Issue #2713, opened 2023-05-17
Improve Screenshots/ViewHierarchy mechanism


markushi:
### Problem Statement

Right now there's little control over capturing Screenshots/VH. This can have a negative impact on the app performance itself (capturing many screenshots, contribution to quota) as well as the usability of the feature itself (no control when a screenshot should be taken or not).

### Solution Brainstorm

* Add debouncing mechansim
Capturing many subsequent screenshots may has a negative performance impact (both CPU and memory). 

* Introduce BeforeCaptureScreenshotCalback/BeforeCaptureViewHierarchyCallback
We want to provide more controls over when and what events for a screenshot should be taken. Thus providing a hook *before* the screenshot is taken makes most sense here.

Issue #2665, opened 2023-04-20
attach screenshots only when crash happens


kasogg:
### Problem Statement

Is there a way to attach screenshots only in crash issues? Custom exception do not need to attach screenshots.

### Solution Brainstorm
Add an option in SentryAndroidOptions.

_No response_

hu441444238xin:
My question is similar




Can the screenshot function be reported based on custom scenarios?Not necessarily

romtsn:
For the moment, you can drop screenshots for non-crashed events via `beforeSend`:
```kotlin
SentryAndroid.init(context) { options ->
    options.beforeSend = SentryOptions.BeforeSendCallback { event, hint ->
        if (!event.isCrashed) {
            hint.screenshot = null
        }
        event       
    }
}
```

We're discussing internally whether we want to introduce a new option for capturing screenshots for handled events, but no decision yet.

@hu441444238xin what custom scenarios do you have in mind? Would like to better understand them before suggesting anything.

markushi:
Hi @kasogg @hu441444238xin I have good news for you: We've added support for customizing the capture behavior with the upcoming `6.24.0` release of our SDK. [The docs also illustrate](https://github.com/getsentry/sentry-docs/pull/7239/files) on how your use-case can be supported.

hu441444238xin:
Happy



发自我的iPhone


------------------ Original ------------------
From: Markus Hintersteiner ***@***.***&gt;
Date: Thu,Jun 22,2023 3:39 PM
To: getsentry/sentry-java ***@***.***&gt;
Cc: feiwang ***@***.***&gt;, Mention ***@***.***&gt;
Subject: Re: [getsentry/sentry-java] attach screenshots only when crashhappens (Issue #2665)





 
Hi @kasogg @hu441444238xin I have good news for you: We've added support for customizing the capture behavior with the upcoming 6.24.0 release of our SDK. The docs also illustrate on how your use-case can be supported.

PR #2773, opened 2023-06-06
Add debouncing mechanism and before-capture callbacks for screenshots/vh

markushi:
## :scroll: Description

As discussed we want to debounce the capturing of screenshots / view hierarchies. As of now I hard-coded the debouncing time to 2 seconds, thus in worst case capturing is happening every 2 seconds.

On top of that a new `beforeCapture` callback is introduced, to allow even more customization and eventually overrule the debouncing mechanism. This callback can also be used to prevent capturing for non-fatal events.

Fixes https://github.com/getsentry/sentry-java/issues/2713 and https://github.com/getsentry/sentry-java/issues/2665

## :pencil: Checklist


sentry-android-core/src/main/java/io/sentry/android/core/ScreenshotEventProcessor.java
```diff
@@ -52,6 +59,16 @@ public ScreenshotEventProcessor(
       return event;
     }
 
+    final boolean shouldDebounce = debouncer.checkForDebounce();
```
stefanosiano, line 62: I think everything is fine here
If you could just add a comment to explain the order of the calls it would be even better 😄

sentry-android-core/src/test/java/io/sentry/android/core/internal/util/DebouncerTest.kt
```diff
@@ -0,0 +1,62 @@
...
+        assertTrue(debouncer.checkForDebounce())
+    }
+
+    @Test
+    fun `Debouncer should not signal debounce if the second invocation is late enough`() {
+        fixture.currentTimeMs = 1000
+        val debouncer = fixture.getDebouncer(3000)
+        assertFalse(debouncer.checkForDebounce())
+
+        fixture.currentTimeMs = 3001
```
stefanosiano, line 59: this test is like the previous one
guess you meant this?

```
fixture.currentTimeMs = 4001
assertFalse(debouncer.checkForDebounce())
```

sentry-android-core/src/test/java/io/sentry/android/core/ViewHierarchyEventProcessorTest.kt
```diff
@@ -315,6 +315,72 @@ class ViewHierarchyEventProcessorTest {
...
+        val (_, hint) = fixture.process(
+            true,
+            SentryEvent().apply {
+                exceptions = listOf(SentryException())
+            }
+        )
+
+        assertNotNull(hint.viewHierarchy)
+    }
+
```
stefanosiano, line 383: could you add a last test `when capture callback returns true, debounce flag is overwritten`?
basically a test where the debounce flag is false and the callback returns true, so the view hierarchy should be attached anyway

sentry-android-core/src/test/java/io/sentry/android/core/ScreenshotEventProcessorTest.kt
```diff
@@ -200,5 +202,81 @@ class ScreenshotEventProcessorTest {
...
+
+        val event = SentryEvent().apply {
+            exceptions = listOf(SentryException())
+        }
+        val hint = Hint()
+
+        processor.process(event, hint)
+        assertNotNull(hint.screenshot)
+    }
+
```
stefanosiano, line 280: could you add a last test `when capture callback returns true, debounce flag is overwritten`?
basically a test where the debounce flag is false and the callback returns true, so the view screenshot should be attached anyway

sentry-android-core/src/main/java/io/sentry/android/core/ViewHierarchyEventProcessor.java
```diff
@@ -59,6 +66,16 @@ public ViewHierarchyEventProcessor(final @NotNull SentryAndroidOptions options)
       return event;
     }
 
+    final boolean shouldDebounce = debouncer.checkForDebounce();
```
stefanosiano, line 69: [same comment](https://github.com/getsentry/sentry-java/pull/2773/files#r1219084571) here

sentry-android-core/src/test/java/io/sentry/android/core/ViewHierarchyEventProcessorTest.kt
markushi: Oh yeah I missed that, thanks! This test case actually caught an edge case 🙇

stefanosiano (approved):

sentry-android-core/src/main/java/io/sentry/android/core/ScreenshotEventProcessor.java
```diff
@@ -52,6 +59,19 @@ public ScreenshotEventProcessor(
       return event;
     }
 
+    // skip capturing in case of debouncing (=too many frequent capture requests)
+    // the BeforeCaptureCallback may overrules the debouncing decision
+    final boolean shouldDebounce = debouncer.checkForDebounce();
+    final @Nullable SentryAndroidOptions.BeforeCaptureCallback beforeCaptureCallback =
+        options.getBeforeScreenshotCaptureCallback();
+    if (beforeCaptureCallback != null) {
```
romtsn, line 67: just checking if that's what we want - if the `beforeCaptureCallback` is set, we will *never* debounce (or rather - we completely move this responsibility to our users). Maybe it's worth to give more details in the javadoc at least? (or maybe docs, not sure).

romtsn (approved):

sentry-android-core/src/main/java/io/sentry/android/core/ScreenshotEventProcessor.java
markushi: Yes, exactly that's intentional - the user should have full control over this - let me update the docs then.

sentry-android-core/src/main/java/io/sentry/android/core/ScreenshotEventProcessor.java
```diff
@@ -52,6 +59,19 @@ public ScreenshotEventProcessor(
...
+      if (!beforeCaptureCallback.execute(event, hint, shouldDebounce)) {
```
krystofwoldrich, line 68: In this case, the `shouldDebounce` works as a recommendation for the user and they can decide if they will respect it or not, right?

Issue #2767, opened 2023-06-05
Enable `captureFailedRequests` by default


marandaneto:
### Description

https://docs.sentry.io/platforms/android/configuration/integrations/okhttp/#http-client-errors

Flutter and iOS are enabled by default already.
Requires major bump.
Update docs.

PR #2794, opened 2023-06-22
Capture failed HTTP requests by default

markushi:

Fixes https://github.com/getsentry/sentry-java/issues/2767

## :pencil: Checklist

- [x] I updated the docs if needed: https://github.com/getsentry/sentry-docs/pull/7238


romtsn (approved):

romtsn:
Btw, I think we should do this for apollo-3 as well? @marandaneto what's your take? or is it still too early for enabling it by default?

marandaneto:
> Btw, I think we should do this for apollo-3 as well? @marandaneto what's your take? or is it still too early for enabling it by default?

That we can't because GraphQL errors include request/response by default, it should be opt-in.

Issue #2808, opened 2023-06-30
Provide APIs to retrieve scope data for Hybrid SDKs


markushi:
### Problem Statement

Main discussion is happening here: https://github.com/getsentry/team-mobile/issues/11

Similar to cocoa we want to provide APIs to sync the current scope down to our Hybrid SDKs. As this only needs to be done on-demand, a pull-based approach should be sufficient.

### Solution Brainstorm

~ `ScopeUtil.serialize(scope: Scope)`

PR #2814, opened 2023-07-03
Add API to retrieve Android scope data

markushi:
## :scroll: Description
Fixes https://github.com/getsentry/sentry-java/issues/2808

More discussion on this one: https://github.com/getsentry/team-mobile/issues/11

Enable pulling scope data via `InternalSentrySdk.serializeScope(scope: Scope) : Map<String, Any>`

In order to not re-implement all our protocol serialization logic again, I extracted some interfaces to abstract away the underlying mechanism (Classic `JsonObjectWriter` + new `MapObjectWriter`).

I tried to keep the changes to a minimum, but there are still a few rough edges:
- The `ObjectWriter` interface methods all throw `IOException`s because the underlying Gson impl does it
- A few `@SuppressWarnings("unchecked")` in `MapObjectWriter` due to the dynamic nature of `Map<String, Any>`
- Some `*ObjectWriter.serialize Date/TimeZone/Collection` code is duplicate

## :pencil: Checklist

## :crystal_ball: Next steps

#skip-changelog


sentry/src/main/java/io/sentry/ScopeUtil.java
```diff
@@ -0,0 +1,48 @@
...
+    try {
+      serialize(writer, options, "user", scope.getUser());
+      serialize(writer, options, "contexts", scope.getContexts());
+      serialize(writer, options, "tags", scope.getTags());
+      serialize(writer, options, "extras", scope.getExtras());
+      serialize(writer, options, "fingerprint", scope.getFingerprint());
+      serialize(writer, options, "level", scope.getLevel());
+      serialize(writer, options, "breadcrumbs", scope.getBreadcrumbs());
+    } catch (Exception e) {
+      options.getLogger().log(SentryLevel.ERROR, "Could not serialize scope");
```
markushi, line 33: @krystofwoldrich if serializing the scope fails would you rather prefer to get an exception or an empty `Map`?

krystofwoldrich:
RN implementation using the serializer is here https://github.com/getsentry/sentry-react-native/pull/3170.

I've noticed there is very little data returned from the Android scope, only Breadcrumbs and items synced from RN to Android are present (redux state and routing navigation tag).

```json
{
  "breadcrumbs": [
    {
      "category": "app.lifecycle",
      "level": "info",
      "timestamp": "2023-07-10T07:11:12.875Z",
      "type": "session",
      "data": {
        "state": "start"
      }
    }
  ],
  "contexts": {
    "state": {
      "state": {
        "value": {
          "counter": 0
        },
        "type": "redux"
      }
    }
  },
  "fingerprint": [],
  "user": null,
  "extras": {},
  "tags": {
    "routing.route.name": "Home"
  },
  "level": null
}
```

marandaneto:
@krystofwoldrich there's a lot of data missing, `InternalSentrySdk.serializeScope` is just one part of the data you need.
For example, all the data set on `DefaultAndroidEventProcessor`, `MainEventProcessor`, etc, and every data that is set when `captureEvent` is captured should be returned as well.
I don't think that `InternalSentrySdk.serializeScope` is a good naming as well since we need more data than just the scope's data.

krystofwoldrich:
@marandaneto Yes, I agree,

I think something like `InternalSentrySdk.getEventScope`, `InternalSentrySdk.getFullScope`, or `InternalSentrySdk.getAndroidScope` would make sense.

sentry-android-core/src/main/java/io/sentry/android/core/ContextUtils.java
```diff
@@ -202,20 +233,10 @@ static boolean isForegroundImportance(final @NotNull Context context) {
...
-          sideLoadedInfo.put("isSideLoaded", "false");
-          // could be amazon, google play etc
-          sideLoadedInfo.put("installerStore", installerPackageName);
-        } else {
-          // if it's installed via adb, system apps or untrusted sources
-          sideLoadedInfo.put("isSideLoaded", "true");
-        }
-
-        return sideLoadedInfo;
+        return new SideLoadedInfo(installerPackageName == null, installerPackageName);
```
stefanosiano, line 239: You can put the check inside the constructor, so that you can pass only one parameter

sentry-android-core/src/main/java/io/sentry/android/core/ContextUtils.java
```diff
@@ -15,17 +16,47 @@
...
 
-final class ContextUtils {
+@ApiStatus.Internal
+public final class ContextUtils {
+
+  static class SideLoadedInfo {
+    private final boolean isSideLoaded;
+    private final @Nullable String installerStore;
+
+    public SideLoadedInfo(boolean isSideLoaded, @Nullable String installerStore) {
```
stefanosiano, line 38: missing `final` keywords

sentry-android-core/src/main/java/io/sentry/android/core/Installation.java
```diff
@@ -12,7 +12,7 @@
 import org.jetbrains.annotations.Nullable;
 import org.jetbrains.annotations.TestOnly;
 
-final class Installation {
+public final class Installation {
```
stefanosiano, line 15: why making this class public? Do we want to expose it to the users?

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
```diff
@@ -0,0 +1,96 @@
...
+import io.sentry.protocol.User;
+import io.sentry.util.MapObjectWriter;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicReference;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+@ApiStatus.Internal
```
stefanosiano, line 23: Maybe just add a comment/Javadoc for the reason of this class, since it's not used in the SDK directly? even something like "This is going to be used by hybrid SDKs to get the objects they need" should suffice

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    final SentryEnvelope repackagedEnvelope =
+        new SentryEnvelope(envelope.getHeader(), envelopeItems);
+    return hub.captureEnvelope(repackagedEnvelope);
+  }
+
+  @Nullable
+  private static Session updateSession(
+      @NotNull IHub hub,
```
stefanosiano, line 163: again final keywords 😄

sentry/src/main/java/io/sentry/protocol/Geo.java
```diff
@@ -136,7 +136,7 @@ public void setUnknown(@Nullable Map<String, Object> unknown) {
   }
 
   @Override
-  public void serialize(JsonObjectWriter writer, ILogger logger) throws IOException {
+  public void serialize(@NotNull ObjectWriter writer, ILogger logger) throws IOException {
```
stefanosiano, line 139: `@NotNull ILogger logger`

sentry-android-core/src/main/java/io/sentry/android/core/Installation.java
markushi: Oh that's an oversight due to some refactoring, let me revert that 👍

stefanosiano (approved):
Looks good!

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
```diff
@@ -0,0 +1,202 @@
...
+@ApiStatus.Internal
+public final class InternalSentrySdk {
+
+  /**
+   * @return a copy of the current hub's topmost scope, or null in case the hub is disabled
+   */
+  @Nullable
+  public static Scope getCurrentScope() {
+    final @NotNull AtomicReference<Scope> scopeRef = new AtomicReference<>();
+    HubAdapter.getInstance().withScope(scopeRef::set);
```
marandaneto, line 44: @markushi See https://github.com/getsentry/sentry-java/blob/beba01476ab71745b0cae821bed705d582374161/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java#L46-L47
This has an issue with some specific use cases, please avoid method references.

marandaneto: Also check if the source code has any more references to `::`.

marandaneto, line 44: Also, `withScope` has side effects and isn't intended to be used on Android, since it's a single hub/scope.
It's being even removed from a few SDKs in favor of the scope lambda -> `captureException(scope {})`
This might have side effects, I know @romtsn dislike this API and already had issues with it as well.

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  @Nullable
+  private static Session updateSession(
+      final @NotNull IHub hub,
+      final @NotNull SentryOptions options,
+      final @Nullable Session.State status,
+      final boolean crashedOrErrored) {
+    final @NotNull AtomicReference<Session> sessionRef = new AtomicReference<>();
+    hub.withScope(
```
marandaneto, line 184: See https://github.com/getsentry/sentry-java/pull/2814/files#r1266374699

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
```diff
@@ -0,0 +1,202 @@
...
+   */
+  public static SentryId captureEnvelope(final @NotNull byte[] envelopeData) throws Exception {
+    final @NotNull IHub hub = HubAdapter.getInstance();
+    final @NotNull SentryOptions options = hub.getOptions();
+
+    final @NotNull ISerializer serializer = options.getSerializer();
+    final @Nullable SentryEnvelope envelope =
+        options.getEnvelopeReader().read(new ByteArrayInputStream(envelopeData));
+    if (envelope == null) {
+      throw new IllegalArgumentException("Envelope could not be read");
```
marandaneto, line 143: M: I'd rather prefer that the Java SDK swallow the exception and log it out, and return an empty ID, otherwise handling this use case has to be duplicated across different hybrid SDKs.

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
markushi: Are you sure this is a problem? I see more than 50 occurrences in our code. Most of them are backend-related, but there are also some which do get executed on Android devices (e.g. `JsonObjectDeserializer`)

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
```diff
@@ -0,0 +1,202 @@
...
+      }
+
+      writer.name("user").value(logger, scope.getUser());
+      writer.name("contexts").value(logger, scope.getContexts());
+      writer.name("tags").value(logger, scope.getTags());
+      writer.name("extras").value(logger, scope.getExtras());
+      writer.name("fingerprint").value(logger, scope.getFingerprint());
+      writer.name("level").value(logger, scope.getLevel());
+      writer.name("breadcrumbs").value(logger, scope.getBreadcrumbs());
+    } catch (Exception e) {
```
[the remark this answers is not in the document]
markushi: Happy to change that but it will make the output ambiguous: E.g. `null` is returned when there's no scope or in case of an error.

sentry/src/main/java/io/sentry/JsonObjectWriter.java
```diff
@@ -6,18 +6,85 @@
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-public final class JsonObjectWriter extends JsonWriter {
+public final class JsonObjectWriter implements ObjectWriter {
```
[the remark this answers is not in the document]
markushi: This is a required change in order to serialize protocol classes to either `Json` or `Map<>`. IMHO it's an oversight that this class is not marked as apistatus.internal. If we consider this a breaking change we should talk to Karl/Krystof as this PR is blocking the RN profiling feature.

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
markushi: This is intentional, as it returns a safe copy of the current scope. I already talked to Roman about this, this is probably the only good use-case for using `withScope` 😅

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
markushi: Fixed here: https://github.com/getsentry/sentry-java/pull/2846

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
marandaneto: yes, I've written that comment myself, its a problem on Android only during desugaring, mainly because of older AGP versions (Hi RN and Flutter).
If there are more on Android code, they should be removed, that's why this comment https://github.com/getsentry/sentry-java/pull/2814#discussion_r1266363020

marandaneto: @markushi can you raise a new issue for replacing all `::` references that run on Android code?

marandaneto: H: `withScope` isn't thread-safe because it's meant to run in a single thread.
Android is single hub/scope.
If another thread set something on the scope while the `push` has been executed but not the `pop` yet, the data is set in the wrong scope, IMO this is still a bug, and `withScope` should never be used in Android.
This API was a mistake and is already deprecated in a few SDKs as mentioned, in favor of the `scope` lambda.

sentry/src/main/java/io/sentry/JsonObjectWriter.java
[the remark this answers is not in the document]
marandaneto: I agree its an oversight, but the class/interface was used in pretty much every class of the protocol.

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
marandaneto: `JsonObjectDeserializer` has references to `::` not sure if more.

markushi: :+1: created an issue here, https://github.com/getsentry/sentry-java/issues/2848

romtsn: Following this logic, we should also get rid of all `default` methods in interfaces, because this is also part of the desugaring. We use those quite heavily. The fact, that nobody has reported any problems so far kinda confirms that there's noone of our customers on AGP lower than 3.0. Is still worth considering this limitation?
image

marandaneto: @romtsn no, we decided and use Java 8 features, the min. version of AGP isn't a problem here, the method reference was a bug o the AGP desugaring and not "not supported" if not version >= X.
If `default` has a bug in one of the AGP versions that are above the min. required, then yes, we'd get rid of `default` methods for compatibility, which is not the case.

romtsn: Ok, what was the version where the bug has been fixed? If it's 3.0.1 I'd say we are still good to not consider it

marandaneto: Don't recall, but why do we even bother if we can keep 100% compatibility by having just a few chars more? `() -> {}`

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
marandaneto: @markushi the usage of `withScope` is still there, this should be replaced by `configureScope`

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
marandaneto: @markushi the usage of `withScope` is still there, mind sharing a commit link or PR?

krystofwoldrich: I thought[ this PR](https://github.com/getsentry/sentry-java/pull/2846/files#diff-8a43e80beabd704bc935bf3f79710e3934bbe99dffca6b8f4a3e3a008c4f4cc5R45) solved that, but looking closer at the code the `withScope` is used.
- https://github.com/getsentry/sentry-java/pull/2846

sentry/src/main/java/io/sentry/JsonObjectWriter.java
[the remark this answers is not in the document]
bruno-garcia: Broke the C# binding somehow: https://github.com/getsentry/sentry-dotnet/pull/2498/files
cc @bitsandfoxes

Issue #2816, opened 2023-07-03
ConcurrentModificationException in SentryFrameMetricsCollector 


markushi:
### Description

SDK Console reports a crash around our metrics collector

````java
java.util.ConcurrentModificationException
io.sentry.android.core.internal.util.SentryFrameMetricsCollector.lambda$new$2
````
Looks like the `listenerMap` is not correctly synchronized:
https://github.com/getsentry/sentry-java/blob/4e809acf9363b52dff9739ce3c903b13af468900/sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java#L37C7-L37C7

PR #2823, opened 2023-07-06
Fix concurrent access to frameMetrics listener

stefanosiano:
## :scroll: Description
changed metrics listener map from HashMap to ConcurrentHashMap

Fixes https://github.com/getsentry/sentry-java/issues/2816

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
@@ -19,11 +19,11 @@ import io.sentry.android.core.BuildInfoProvider;
 import io.sentry.util.Objects;
 import java.lang.ref.WeakReference;
 import java.lang.reflect.Field;
-import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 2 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
```diff
@@ -34,8 +35,8 @@ public final class SentryFrameMetricsCollector implements Application.ActivityLi
   private final @NotNull SentryOptions options;
   private @Nullable Handler handler;
   private @Nullable WeakReference<Window> currentWindow;
-  private final @NotNull HashMap<String, FrameMetricsCollectorListener> listenerMap =
-      new HashMap<>();
+  private final @NotNull Map<String, FrameMetricsCollectorListener> listenerMap =
```
markushi, line 38: `h`: I see that `trackedWindows` is being access via `stopCollection()` as well - I think this needs to be thread-safe too right?

markushi (changes requested):

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
stefanosiano: yeah, nice catch

markushi (approved):
LGTM, Nice!

PR #2845, opened 2023-07-17
Deduplicate events happening in multiple threads simultaneously

### LINKED ISSUE #2827: Duplicate crashes per session in OOM situations
### Description

Per design, once a session experiences a crash, the session needs to be finished and a new session created. In some OOM edge cases, the SDK might send report multiple duplicate crashes for the same session before a new session is started.

Details to follow

romtsn:
## :scroll: Description

Introduce new DeduplicateMultithreadedEventProcessor which keeps an in-memory cache of crashed events in a form of map/dict (crash type - thread id). If a crash type has already been processed once and coming in again from a different thread, the processor will drop the event. 

This can be the case for OutOfMemoryError, when multiple threads are trying to allocate new memory and therefore crashing within the same millisecond from multiple threads. When this happens, the events are almost equal, except the thread id and timestamp. We've seen this for one of our customers and also were somewhat able to reproduce it locally.

Closes #2827 

Manually and automated

## :pencil: Checklist

### DIFF
--- a/sentry/src/main/java/io/sentry/UncaughtExceptionHandlerIntegration.java
+++ b/sentry/src/main/java/io/sentry/UncaughtExceptionHandlerIntegration.java
@@ -3,6 +3,7 @@ package io.sentry;
 import com.jakewharton.nopen.annotation.Open;
 import io.sentry.exception.ExceptionMechanismException;
 import io.sentry.hints.BlockingFlushHint;
+import io.sentry.hints.EventDropReason;
 import io.sentry.hints.SessionEnd;
 import io.sentry.protocol.Mechanism;
 import io.sentry.protocol.SentryId;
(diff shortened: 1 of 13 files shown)


sentry/src/main/java/io/sentry/UncaughtExceptionHandlerIntegration.java
```diff
@@ -98,7 +99,11 @@ public void uncaughtException(Thread thread, Throwable thrown) {
 
         final @NotNull SentryId sentryId = hub.captureEvent(event, hint);
         final boolean isEventDropped = sentryId.equals(SentryId.EMPTY_ID);
-        if (!isEventDropped) {
+        final EventDropReason eventDropReason = HintUtils.getEventDropReason(hint);
+        // in case the event has been dropped by multithreaded deduplicator, the other threads will
+        // crash the app without a chance to persist the main event so we have to special-case this
+        if (!isEventDropped
```
romtsn, line 105: The change is needed because otherwise the other crashing threads will not hold a lock anymore and will crash the app immediately, not giving a chance to persist the crash that was the first one (the one which we're going to deduplicate).

markushi (approved):
Nice one, LGTM!

PR #2846, opened 2023-07-18
Fix ensure session is updated using configureScope

markushi:
## :scroll: Description
Use configureScope instead of withScope, as it doesn't (partially) deep-copy the scope.
Doesn't really seem to be a problem as scope.session is never deep-copied.

## :pencil: Checklist

## :crystal_ball: Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
 - Fix ANRv2 thread dump parsing for native-only threads ([#2839](https://github.com/getsentry/sentry-java/pull/2839))
 - Derive `TracingContext` values from event for ANRv2 events ([#2839](https://github.com/getsentry/sentry-java/pull/2839))
 
+### Features
+- (Internal) Extend APIs for hybrid SDKs ([#2814](https://github.com/getsentry/sentry-java/pull/2814), [#2846](https://github.com/getsentry/sentry-java/pull/2846))
+
 ## 6.25.2
 
 ### Fixes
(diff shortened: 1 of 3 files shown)


romtsn (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@
 - Fix ANRv2 thread dump parsing for native-only threads ([#2839](https://github.com/getsentry/sentry-java/pull/2839))
 - Derive `TracingContext` values from event for ANRv2 events ([#2839](https://github.com/getsentry/sentry-java/pull/2839))
 
+### Features
```
romtsn, line 10: nit: usually we tend to put `### Features` above `### Fixes`

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
```diff
@@ -124,54 +124,62 @@ public static Map<String, Object> serializeScope(
...
    *
    * @param envelopeData the serialized envelope data
-   * @return The Id (SentryId object) of the event
-   * @throws Exception In case the provided envelope could not be parsed / is invalid
+   * @return The Id (SentryId object) of the event, or null in case the envelope could not be
+   *     captured
    */
-  public static SentryId captureEnvelope(final @NotNull byte[] envelopeData) throws Exception {
+  @Nullable
+  public static SentryId captureEnvelope(final @NotNull byte[] envelopeData) {
```
marandaneto, line 138: The return should not be nullable, but rather `SentryId.EMPTY_ID` as every other capture method, so the caller does not have to handle NPEs.

PR #2725, opened 2023-05-22
Add custom maven target for unity

romtsn:
_#skip-changelog_

PR #2847, opened 2023-07-18
Make SDK unity-compatible

romtsn:
## :scroll: Description

Since sentry-unity consumes Android SDK artifacts decoupled from each other (e.g. `sentry.jar` and `sentry-android-core.aar` do not know about each other), some versions of Unity cannot properly desugar `default` methods in interfaces during build time. 

This results in runtime errors when a `default` method from `sentry.jar` is called in `sentry-android-core.aar`. We've tried to work this around by properly consuming android artifacts using local maven (https://github.com/getsentry/sentry-java/pull/2725), but this turned out to be a nightmare to support, because we modify gradle build files via regexes on sentry-unity.

This PR simply converts all `default` methods that are used in sentry-android-core to be non-default (meaning proper implementations in each interface implementor). 

@bitsandfoxes mental well-being

Manually and automated

## :pencil: Checklist


romtsn:
The only problem is that this change is binary-incompatible, so if someone is using an older integration version (e.g. `sentry-android-timber:6.24.0`), and we ship this but they don't update and align the sentry versions, it will crash at runtime. So I'm leaning towards making it a part of v7.0.0 (this shouldn't take longer than 2-3 weeks @bitsandfoxes I hope this should be fine for you?).

stefanosiano (approved):
looks great!
just a note, though. The public interfaces that were changed are technically a breaking change. A note in the changelog is enough, as it will be part of v7 anyway

markushi (approved):

PR #2849, opened 2023-07-19
Do not send session updates for terminated sessions

romtsn:
## :scroll: Description

If the session is in terminal state (`!= Status.Ok`), we should not send any new session updates, because there's no protection for that server-side, so potentially we were reporting incorrect crash-free rate. Follow-up of #2845  

Internal customer report

Manually and automated

## :pencil: Checklist

### DIFF
--- a/sentry/src/main/java/io/sentry/SentryClient.java
+++ b/sentry/src/main/java/io/sentry/SentryClient.java
@@ -137,7 +137,10 @@ public final class SentryClient implements ISentryClient {
     @Nullable Session session = null;
 
     if (event != null) {
-      session = updateSessionData(event, hint, scope);
+      // https://develop.sentry.dev/sdk/sessions/#terminal-session-states
+      if (sessionBeforeUpdate == null || !sessionBeforeUpdate.isTerminated()) {
+        session = updateSessionData(event, hint, scope);
+      }
 
       if (!sample()) {
         options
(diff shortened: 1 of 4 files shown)


stefanosiano (approved):

sentry/src/main/java/io/sentry/SentryClient.java
```diff
@@ -490,9 +493,8 @@ Session updateSessionData(
                     }
 
                     if (session.update(status, userAgent, crashedOrErrored, abnormalMechanism)) {
-                      // if we have an uncaughtExceptionHint we can end the session.
-                      if (HintUtils.hasType(
-                          hint, UncaughtExceptionHandlerIntegration.UncaughtExceptionHint.class)) {
+                      // if session terminated we can end it.
+                      if (session.isTerminated()) {
```
romtsn, line 497: @marandaneto @krystofwoldrich this is kind of behavioral change for the Hybrid SDKs (or rather a fix), because previously we'd only end the session if a hybrid app crashed in the Android layer, but would ignore fatals from the Hybrid layer. Just wanna check if you're fine with that?

marandaneto: So Flutter never crashes, and when it does, it's on the Java or C layer.
React native throws on the Java/C layer as well.
Maybe take a look at `EnvelopeCache`, `PreviousSessionFinalizer`, and `InternalSentrySdk`.
Multiple classes are calling `session.update` now, wondering if more changes are required.
For example, Hybrid SDKs will call `captureEnvelope` in the future after the changes for enriching events with Android context on the Hybrid layer.
So technically this method isn't even executed here, unless its thrown in the native layer.
Hybrid SDKs most of the time crash via C as well, which relies on `EnvelopeCache` ending the session.
`InternalSentrySdk` also does that now, and not sure about `PreviousSessionFinalizer`.
The change looks fine but I'd double-check the implementation on the other side.

romtsn: Yeah `InternalSentrySdk` takes care of that as well now https://github.com/getsentry/sentry-java/blob/0bff5c1be7cb8741972c7498e90df21e40f3da0c/sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java#L192-L207

if they all crash in the native layer, then we are all good I believe. thanks for checking 👍

PR #2851, opened 2023-07-19
Measure AppStart time till First Draw 

### LINKED ISSUE #2627: Report AppStart time till First Frame Drawn and not onResume
### Description

onResume isn't correct according to https://developer.android.com/topic/performance/vitals/launch-time#cold

image

Potentially we should also change when we start measuring app start time (first class load) https://github.com/square/papa/blob/26f21379a8efb14b292c1db5188d4a3be6a48872/papa/src/main/java/papa/internal/Perfs.kt#L300-L359

Relates to #2177 and https://github.com/getsentry/sentry-cocoa/issues/2840.

romtsn:
## :scroll: Description

* Moves setting AppStart end time to First Draw listener in both `ActivityLifecycleIntegration `and `SentryPerformanceProvider`

image

Closes #2627 

Manually and automated

## :pencil: Checklist


stefanosiano (approved):
everything looks good
i was thinking to add the considerations in https://blog.p-y.wtf/tracking-android-app-launch-in-production#heading-cold-launch-1, since sometime (rarely) the start timestamp we get is wrong and we are changing the related code
We can do that in another pr, too

PR #2852, opened 2023-07-19
Do not overwrite UI transaction status if set by the user

### LINKED ISSUE #2378: Do not overwrite existing status of UI event transactions
### Description

We should not force set the status when finishing UI event transactions, but instead check if the existing status is null and only set the status in that case, otherwise we might be overriding what the user set https://github.com/getsentry/sentry-java/blob/84f0ef1abb55bccd90315665ff4e307c5f71886c/sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/SentryGestureListener.java#L277

romtsn:
## :scroll: Description

Do not overwrite UI transaction status if set by the user

Closes #2378 

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ Breaking changes:
 
 ### Fixes
 
+- Do not overwrite UI transaction status if set by the user ([#2852](https://github.com/getsentry/sentry-java/pull/2852))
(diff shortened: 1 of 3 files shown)


markushi (approved):
LGTM!

PR #2855, opened 2023-07-20
Return transaction for Sentry.getSpan when global hub mode is enabled

markushi:
## :scroll: Description

Fixes https://github.com/getsentry/sentry-java/issues/1828

Just like Cocoa, always return the root transaction instead of the last active span. This is only applied if globalHubMode is enabled.

## :pencil: Checklist


romtsn:
I think simply changing this on the `Sentry.java` level is probably not enough, because we are using `Hub.getSpan()` in our own auto-instrumentation, and this will still be returning the latest span and not the root one.
image

Also, we've talked about introducing another method `getLatestSpan` or something like that, which we can use in our DB instrumentation to not break the parent-child relationship for Room+SQLite spans, can we also do that in this PR please?

markushi:
> I think simply changing this on the `Sentry.java` level is probably not enough, because we are using `Hub.getSpan()` in our own auto-instrumentation, and those will still be returning the latest span and not the root one. image
As of now globalhub mode is not stored anywhere but `Sentry.java` so the hub itself doesn't know if it's global or not - let me look into this.


> Also, we've talked about introducing another method `getLatestSpan` or something like that, which we can use in our DB instrumentation to not break the parent-child relationship for Room+SQLite spans, can we also do that in this PR please?
Yes, we thought about this, on it...

romtsn:
> As of now globalhub mode is not stored anywhere but Sentry.java so the hub itself doesn't know if it's global or not - let me look into this.

We could just use `hub.getTransaction()` in our integrations maybe, where applicable

romtsn (approved):
LGTM, but I think we also have to do it for Apollo, at least on Android? Is the apollo integration used server-side as well @adinauer? But we have the same problem for the okhttp integration now anyway, if someone is using it on the server alongside the Java SDK

adinauer:
> Is the apollo integration used server-side as well

It's a client only, we have support for `graphql-java` on server side which is used by Netflix DGS as well as `spring-graphql`.

Or do you mean can this be used in a backend application? That's a yes.

Not really sure about details here, maybe we can do a quick call if you need my input here so I can better grasp implications.

markushi:
Let's change the `Sentry.getSpan()` behavior for Android only. In case this comes up for Desktop/Cli let's deal with this at a later point.

sentry/src/main/java/io/sentry/Sentry.java
```diff
@@ -918,10 +919,16 @@ public static void endSession() {
...
    * Gets the current active transaction or span.
    *
-   * @return the active span or null when no active transaction is running
+   * @return the active span or null when no active transaction is running. In case of
+   *     globalHubMode=true, always the active transaction is returned, rather than the last active
+   *     span.
    */
   public static @Nullable ISpan getSpan() {
-    return getCurrentHub().getSpan();
+    if (globalHubMode && Platform.isAndroid()) {
```
romtsn, line 927: but this is still not the only thing we want, right? In integrations like okhttp, where we use `hub.getTransaction` we should use either `getTransaction` or `getSpan` depending on whether it's running on Android or not

romtsn:
@markushi I think the PR is still missing `SentryApolloInterceptor` and `SentryApollo3Interceptor` from the list 😅 
image

romtsn (approved):
looks good now, thanks for addressing back-and-forths 🎉

PR #2858, opened 2023-07-20
Reduce flush timeout to 4s on Android to avoid ANRs

### LINKED ISSUE #2732: Reduce `flushTimeoutMillis` to 3 or 4 seconds for Android to avoid ANRs
### Description

At the moment, we hold the lock for 15seconds (default for Java backend, which we inherited from on Android), which obviously will result in ANR on Android. I don't think this happens often when someone is using only Sentry for crash reporting, but when used in conjunction with Crashlytics for example, this can happen pretty easily. But still we could be more conservative here and reduce it to something like Crashlytics (3-4s), this can prevent some of those reported ANRs.

Originally reported in #2719

markushi:
## :scroll: Description
Reduce flush timeout to 4s on Android to avoid ANRs

[Fixes #2732](https://github.com/getsentry/sentry-java/issues/2732)

## :pencil: Checklist

### DIFF
--- a/sentry/src/main/java/io/sentry/cache/EnvelopeCache.java
+++ b/sentry/src/main/java/io/sentry/cache/EnvelopeCache.java
@@ -67,6 +67,8 @@ public class EnvelopeCache extends CacheStrategy implements IEnvelopeCache {
 
   public static final String STARTUP_CRASH_MARKER_FILE = "startup_crash";
 
+  private static final long SESSION_FLUSH_DISK_TIMEOUT_MS = 15000;
+
   private final CountDownLatch previousSessionLatch;
 
   private final @NotNull Map<SentryEnvelope, String> fileNameMap = new WeakHashMap<>();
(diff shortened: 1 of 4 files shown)


stefanosiano (approved):

romtsn:
@markushi there's one place where I'd like to keep the default 15s for flush timeout and it's this https://github.com/getsentry/sentry-java/blob/f60279b3bb7fa8e599cbbc951afd706637d17c5c/sentry/src/main/java/io/sentry/cache/EnvelopeCache.java#L430-L438

when we're waiting for the previous unfinished session to be flushed to disk. Would it be possible to keep the 15s value there (we could just have a constant actually)?

sentry/src/main/java/io/sentry/cache/EnvelopeCache.java
```diff
@@ -429,7 +431,9 @@ public void discard(final @NotNull SentryEnvelope envelope) {
   /** Awaits until the previous session (if any) is flushed to its own file. */
   public boolean waitPreviousSessionFlush() {
     try {
-      return previousSessionLatch.await(options.getFlushTimeoutMillis(), TimeUnit.MILLISECONDS);
+      // use fixed timeout instead of configurable options.getFlushTimeoutMillis() to ensure there's
+      // enough time to flush the session to disk
```
markushi, line 435: @romtsn if you've better reasoning here, please adapt the comment 😊

romtsn: yeah, it's good!

romtsn (approved):

Issue #2730, opened 2023-05-24
Refrain from overwriting the span status for unfinished spans


markushi:
### Problem Statement

As discussed in https://github.com/getsentry/sentry-java/issues/1690 unfinished spans should be sent, but their status should not be set to `deadline_exceeded`.

### Solution Brainstorm

Don't overwrite the span status.

markushi:
Let's keep self-hosted in mind here, and add corresponding information to the release.

stefanosiano:
the corresponding [pr](https://github.com/getsentry/relay/pull/1690) in relay was in Dec 2022

PR #2859, opened 2023-07-20
Fix don't overwrite the span status of unfinished spans

markushi:
## :scroll: Description
Span status should remain untouched when being force-finished, instead of setting `DEADLINE_EXCEEDED`.

Fixes https://github.com/getsentry/sentry-java/issues/2730

## :pencil: Checklist


sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -198,7 +198,7 @@ public void finish(
         if (!child.isFinished()) {
           child.setSpanFinishedCallback(
               null); // reset the callback, as we're already in the finish method
-          child.finish(SpanStatus.DEADLINE_EXCEEDED, finishTimestamp);
+          child.finish(child.getStatus(), finishTimestamp);
```
romtsn, line 201: hm, but as I understand, we don't even need this logic anymore at all (iterating through children and finishing them), relay will just do that for us?

romtsn: image

markushi: Oh right, good catch. Also: the status only gets set if there's no end timestamp: https://github.com/getsentry/relay/pull/1690/files#diff-7f49c53574a8b3eef93c44e6df72f927d591b4f00ebf791af696ea6eae77fb2bR280-R283
Let me fix that

romtsn (approved):
LGTM, but only one concern - we're not executing this block for JPC stuff, do we still need to handle it (at least the start date), in case of parent transaction is aborted?

https://github.com/getsentry/sentry-java/blob/695d3a3acd91e7759da0699f9b5e004d945de53b/sentry/src/main/java/io/sentry/Span.java#L193-L222

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ Breaking changes:
 
 Breaking changes:
 - Move enableNdk from SentryOptions to SentryAndroidOptions ([#2793](https://github.com/getsentry/sentry-java/pull/2793))
+- Fix don't overwrite the span status of unfinished spans ([#2859](https://github.com/getsentry/sentry-java/pull/2859))
```
romtsn, line 16: Also, should we mention min support for self-hosted? as per https://github.com/getsentry/sentry-java/issues/2730#issuecomment-1569843157

stefanosiano (approved):
good!

Issue #1454, opened 2021-05-03
Codecov stats are missing for Android modules


marandaneto:
https://app.codecov.io/gh/getsentry/sentry-java

markushi:
@marandaneto Let's re-priortize this in the Android sync and decouple it from the milestone.

markushi:
To clarify: this is about the android modules (like `sentry-android`, `sentry-android-okhttp`) not being part of the code stats. [This guide should get us started](https://about.codecov.io/blog/code-coverage-for-android-development-using-kotlin-jacoco-github-actions-and-codecov/).

markushi:
Same applies for our SAGP repo: https://github.com/getsentry/sentry-android-gradle-plugin/issues/87

romtsn:
Closed by #2868

PR #2868, opened 2023-07-27
Codecov reports for Android modules

buenaflor:
Ref: #1454 

#skip-changelog

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -97,19 +97,29 @@ allprojects {
 }
 
 subprojects {
-    if (name.contains("sentry-android") && !name.contains("sentry-android-integration-tests")) {
+    if (name.contains("sentry-android") || name.equals("sentry-compose") && !name.contains("sentry-android-integration-tests")) {
         val androidJacocoTaskName = "androidJacocoTestReport"
 
         tasks.create(androidJacocoTaskName, JacocoReport::class) {
             dependsOn("testReleaseUnitTest")
 
             reports {
-                html.required.set(false)
+                html.required.set(true)
                 xml.required.set(true)
             }
 
+            val classesTree = fileTree("$buildDir/intermediates/javac/release").setExcludes(
+                listOf(
+                    "**/R.class",
+                    "**/R$*.class",
+                    "**/BuildConfig.*",
+                    "**/Manifest*.*",
+                    "**/*Test*.*",
+                    "android/**/*.*"
+                )
+            )
             executionData.setFrom("$buildDir/outputs/unit_test_code_coverage/releaseUnitTest/testReleaseUnitTest.exec")
-            classDirectories.setFrom("$buildDir/tmp/kotlin-classes/releaseUnitTest")
+            classDirectories.setFrom(classesTree)
             sourceDirectories.setFrom("$projectDir}/src/main/java")
 
             doLast {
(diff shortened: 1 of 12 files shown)


romtsn:
I'm a bit confused, because we're running test coverage on `src/test/java` and not `src/main/java`, though it should be the other way around.

image

Also, `sentry-compose` module is still missing from the coverage it looks like

buenaflor:
@romtsn should be fixed now, also `sentry-compose` is also now included

romtsn:
Looks better already, but still something is off. For example, `ActivityLifecycleIntegration` appears to be completely [uncovered](https://app.codecov.io/gh/getsentry/sentry-java/blob/chore%2Fcodecov-android/sentry-android-core%2Fsrc%2Fmain%2Fjava%2Fio%2Fsentry%2Fandroid%2Fcore%2FActivityLifecycleIntegration.java), although we have a pretty extensive test [covering it](https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/test/java/io/sentry/android/core/ActivityLifecycleIntegrationTest.kt)

markushi (approved):

buenaflor:
Seems to be resolved now, the percentage drops a bit but the coverage looks good so far

build.gradle.kts
```diff
@@ -97,6 +98,32 @@ allprojects {
...
+        "sentry-android-navigation",
+        "sentry-android-ndk",
+        "sentry-android-okhttp",
+        "sentry-android-sqlite",
+        "sentry-android-timber",
+        "sentry-compose"
+    )
+    if (jacocoAndroidModules.contains(name)) {
+        afterEvaluate {
+            tasks.getByName("jacocoTestReport") {
```
romtsn, line 113: you could also use the default Gradle's `Copy` task, which should be even pretty straightforward https://docs.gradle.org/current/dsl/org.gradle.api.tasks.Copy.html

romtsn:
`sentry-compose` is missing again though, probably because the plugin is not aware of `androidMain`? 
image

romtsn:
We might need to look into https://kotlin.github.io/kotlinx-kover/gradle-plugin/ for KMP modules

sentry-android-core/build.gradle.kts
```diff
@@ -64,9 +65,14 @@ android {
     }
 }
 
+jacoco {
+    toolVersion = "0.8.10"
```
marandaneto, line 69: M: Move it to Config.kt, DRY

sentry-android-core/build.gradle.kts
```diff
@@ -64,9 +65,14 @@ android {
...
+}
+
 tasks.withType<Test> {
     configure<JacocoTaskExtension> {
-        isIncludeNoLocationClasses = false
+        isIncludeNoLocationClasses = true
+        excludes = listOf("jdk.internal.*")
```
marandaneto, line 75: M: If its always the same, can also be moved to Config.kt

buenaflor:
@romtsn 
`sentry-compose` is now included again, does the coverage look correct? (it shows similar results with Kover)

As for Kover: after testing it out, Codecov never could parse the kover.xml report apparently so I chose to implement it manually through jacoco with a custom task.

romtsn (approved):
LGTM 👍

PR #2872, opened 2023-07-28
propagate okhttp status to parent spans

stefanosiano:
## :scroll: Description
status of OkHttp calls now gets propagated to parent span
when OkHttp call finishes, status of unfinished spans is not overridden and is set to INTERNAL_ERROR

OkHttp calls are grouped in a tree of parent-children. If a child span fails with status `internal_error`, the parent span has a different status, but it should be `internal_error`, too.
Fixes https://github.com/sentry-demos/android/issues/60 

Unit test

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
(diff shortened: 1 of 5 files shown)


markushi (approved):
Nice, looks good to me!

Issue #912, opened 2020-04-16
Sending cached events periodically


nominalista:
Because some of the events may be not be sent immediately after they are created, it would be great if there was an option to send these events periodically, even if the app is not restarted. 

Is there any method in SDK that I can call and try to send cached events? I would use it as a temporary solution along with e.g. `WorkManager`.

jadbaz:
@marandaneto, how is possible to use the integration periodically?

marandaneto:
this issue isn't Android specific, hence removing the Android label

marandaneto:
@jadbaz this feature isn't supported yet, I'd prefer to not share a hacky solution with flawns, please upvote the issue, and let's see if we can get prioritized.

jadbaz:
Hello,
Just checking up if this is on the roadmap at some point

AshStefanOltmann:
I think this is really important for client apps and I'm also waiting for this.

bruno-garcia:
Relates to: https://github.com/getsentry/sentry-dotnet/issues/1504

marandaneto:
It'd benefit when you have an unstable connection, or no connection at all.
Events are always sent on restart anyway.
Also when you never close/kill the app very often, the OS does it anyways when memory is needed.

The next step would be to implement an improvement already made by the Sentry Cocoa SDK, which is sending every pending envelope once an envelope is captured.
So if you capture an envelope and you have no connection, the envelope is cached, if you capture another envelope and there's connection, both envelopes would be sent.

How does that sound?

vibin:
@marandaneto FWIW, I like that Sentry doesn't use WorkManager/AlarmManager, etc. to auto-initialise background jobs for sending cached events – because, I have a usecase where I initialise Sentry conditionally (only if a feature flag is on).

I see that currently `SendCachedEnvelopeFireAndForgetIntegration` is the class that is responsible for pushing cached events on app initialisation, so it's working well for my usecase

farfromrefug:
I am very interested with such a feature. One thing that could be done to help apps is to implement the `io.sentry.Sentry.flush` method.  Wondering if this could be done by using this https://github.com/getsentry/sentry-java/blob/469420402e4717906783a744f8e2fb156168ec25/sentry/src/main/java/io/sentry/SendCachedEnvelopeFireAndForgetIntegration.java

farfromrefug:
[...]

PR #2910, opened 2023-09-01
Observe network state to upload any unsent envelopes

markushi:
## :scroll: Description
Adds a `ConnectionStatusProvider` mechanic, so we can observe changes to the connection state.
API wise this is already complete, only some more tests for `SendCachedEnvelopeFireAndForgetIntegration` are required.

Partially implements https://github.com/getsentry/sentry-java/issues/912

Added unit tests.

## :pencil: Checklist


romtsn:
Note that on Android we use [SendCachedEnvelopeIntegration](https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/SendCachedEnvelopeIntegration.java), so the change has likely to be implemented for both `SendCachedEnvelopeIntegration` and `SendCachedEnvelopeFireAndForgetIntegration`.

The former one is android-only as we also check for startup crashes there and block in case there are any. The latter one is jvm/backend only, and has to be manually installed by our users. We could also try to combine those two into one, but I remember I didn't want to do that, because on pure jvm apps we don't really need it, see https://github.com/getsentry/sentry-java/pull/2277#issuecomment-1277655397

sentry/src/main/java/io/sentry/EnvelopeSender.java
```diff
@@ -53,6 +57,14 @@ protected void processFile(final @NotNull File file, final @NotNull Hint hint) {
       return;
     }
 
+    if (envelopeCache.containsFile(file)) {
```
romtsn, line 60: `m`: I think we should do the same in `OutboxSender` as we also install the second instance of `SendCachedEnvelopeIntegration` for sending cached files from the outbox folder

romtsn: Also, a comment would be nice to mention why this is needed

sentry/src/main/java/io/sentry/transport/RateLimiter.java
```diff
@@ -87,6 +87,11 @@ public RateLimiter(final @NotNull SentryOptions options) {
     return envelope;
   }
 
+  @Nullable
+  public Date getRateLimitedUntilFor(final @NotNull String itemType) {
```
romtsn, line 91: This is probably from a different PR?

sentry/src/main/java/io/sentry/cache/EnvelopeCache.java
```diff
@@ -416,6 +425,11 @@ public void discard(final @NotNull SentryEnvelope envelope) {
     return ret.iterator();
   }
 
+  @Override
+  public synchronized boolean containsFile(@NotNull File file) {
```
romtsn, line 429: `m`: how about we rename this to something like `hasAlreadyProcessedFile(file)` to make it a bit less confusing on the calling side (where we actually processed cached files, but at the same time early-returning if it's already cached 😅 )

romtsn (approved):
A few minor things:
* Left some comments what I think make sense to change a bit
* A few tests missing for the `sentry` module
* A few `final` keywords missing

Otherwise looks great, I think this is a crucial improvement 🚀

adinauer (approved):
Gave this a quick look, LGTM.

sentry-android-core/src/main/java/io/sentry/android/core/SendCachedEnvelopeIntegration.java
```diff
@@ -29,7 +39,8 @@ public SendCachedEnvelopeIntegration(
   @Override
   public void register(@NotNull IHub hub, @NotNull SentryOptions options) {
     Objects.requireNonNull(hub, "Hub is required");
-    final SentryAndroidOptions androidOptions =
+    this.hub = hub;
```
stefanosiano, line 42: you can merge the assignment
```suggestion
    this.hub = Objects.requireNonNull(hub, "Hub is required");
```

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/AndroidConnectionStatusProvider.java
```diff
@@ -7,40 +7,48 @@
...
-  public enum Status {
-    CONNECTED,
-    NOT_CONNECTED,
-    NO_PERMISSION,
-    UNKNOWN
-  }
+  private final @NotNull Context context;
+  private final @NotNull ILogger logger;
+  private final @NotNull BuildInfoProvider buildInfoProvider;
+  private final @NotNull Map<IConnectionStatusObserver, ConnectivityManager.NetworkCallback>
```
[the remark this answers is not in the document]
stefanosiano: not required from what i can see, but that definitely doesn't hurt, especially for future changes

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/AndroidConnectionStatusProvider.java
```diff
@@ -50,6 +58,50 @@ private ConnectivityChecker() {}
...
 
+  @Override
+  public @Nullable String getConnectionType() {
+    return getConnectionType(context, logger, buildInfoProvider);
+  }
+
+  @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
+  @Override
+  public boolean addConnectionStatusObserver(@NotNull IConnectionStatusObserver observer) {
+    final ConnectivityManager.NetworkCallback callback =
```
stefanosiano, line 69: what about adding a guard here for min api?
like
```
if (buildInfoProvider.getSdkInfoVersion() < Build.VERSION_CODES.LOLLIPOP) {
    logger.log(SentryLevel.DEBUG, "addConnectionStatusObserver need Android LOLLIPOP+.");
    return false;
}
```

sentry/src/main/java/io/sentry/SendCachedEnvelopeFireAndForgetIntegration.java
```diff
@@ -54,20 +60,59 @@ public SendCachedEnvelopeFireAndForgetIntegration(
...
 
-  @SuppressWarnings("FutureReturnValueIgnored")
   @Override
-  public final void register(final @NotNull IHub hub, final @NotNull SentryOptions options) {
+  public void register(final @NotNull IHub hub, final @NotNull SentryOptions options) {
     Objects.requireNonNull(hub, "Hub is required");
     Objects.requireNonNull(options, "SentryOptions is required");
 
+    this.hub = hub;
+    this.options = options;
```
stefanosiano, line 69: you can merge assignments here

stefanosiano (approved):
looks great, just fix the comments left and we're good to go!

dave-t-tran:
@romtsn I noticed these code changes aren't in https://github.com/getsentry/sentry-java/releases/tag/7.0.0-rc.1. Are these changes slated for a future 7.x release?

romtsn:
@dave-t-tran they've been part of https://github.com/getsentry/sentry-java/releases/tag/7.0.0-beta.1 already, so the rc.1 also contains them

Issue #2908, opened 2023-08-30
OkHttp not closing on responses not closed


stefanosiano:
### Description

We automatically instrument rest calls done through OkHttp, using its `EventListener`.
The `EventListener`'s `callEnd` method is called only after the `Response` is consumed. That is, after any of `Response.close()`, `Response.body.text()` or `Response.body.bytes()` is called.
So, when the user performs a request and doesn't read/close the response, we don't automatically finish the `connection` span, neither the `http.client` span. This means that we end up with never ending transactions, in case of idle transactions.
This applies to both GET and POST requests.

v7 will have a final timeout that would mitigate this issue, but we should find a way to automatically stop those spans, even if the user doesn't close the response.

markushi:
According to the docs [Response](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-response/) is closable, so in theory it should auto-close itself.

markushi:
Let's try to remove the `Response` from the hint and see if the body closes automatically or not.

PR #2923, opened 2023-09-06
Add OkHttp span auto-close when response body is not read

stefanosiano:
## :scroll: Description
added 500 milliseconds timeout after responseHeaders to finish the call root span if the response body is not read
added scheduleFinish to SentryOkHttpEvent

The OkHttp span could be running forever in case a request is made and the response is never read, as the listener doesn't receive the corresponding calls.
Fixes https://github.com/getsentry/sentry-java/issues/2908

Unit

## :pencil: Checklist

### DIFF
--- a/sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEvent.kt
+++ b/sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEvent.kt
@@ -4,6 +4,7 @@ import io.sentry.Breadcrumb
 import io.sentry.Hint
 import io.sentry.IHub
 import io.sentry.ISpan
+import io.sentry.SentryDate
 import io.sentry.SpanDataConvention
 import io.sentry.SpanStatus
 import io.sentry.TypeCheckHint
@@ -19,9 +20,11 @@ import okhttp3.Request
 import okhttp3.Response
 import java.util.Locale
 import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.atomic.AtomicBoolean
 
 private const val PROTOCOL_KEY = "protocol"
 private const val ERROR_MESSAGE_KEY = "error_message"
+private const val RESPONSE_BODY_TIMEOUT_MILLIS = 500L
 internal const val TRACE_ORIGIN = "auto.http.okhttp"
 
 internal class SentryOkHttpEvent(private val hub: IHub, private val request: Request) {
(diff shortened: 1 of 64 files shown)


sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEvent.kt
```diff
@@ -146,4 +159,14 @@ internal class SentryOkHttpEvent(private val hub: IHub, private val request: Req
         RESPONSE_BODY_EVENT -> eventSpans[CONNECTION_EVENT]
         else -> callRootSpan
     } ?: callRootSpan
+
+    fun scheduleFinish(timestamp: SentryDate) {
+        hub.options.executorService.schedule({
```
markushi, line 164: I think this needs to be wrapped in a try/catch - in case the SDK / executor service is in the process of shutting down

markushi (approved):
Looks good to me, please have a look at my comment!

I played around with the `Response` class myself yesterday, and it seems like there's no better option for us - apart from removing the feature altogether. From the docs:
> Response bodies must be [closed](https://square.github.io/okhttp/3.x/okhttp/okhttp3/ResponseBody.html) and may be consumed only once.

So the issue we're seeing here is a misuse of OkHttp itself, which in turn breaks our instrumentation 😅

In the long term I think we should think about having multiple ExecutorServices, as it's handling a lot of async timeout related tasks by now, and features will start to delay each other, ultimately making it harder to have a predictable SDK. cc @romtsn

romtsn:
> So the issue we're seeing here is a misuse of OkHttp itself, which in turn breaks our instrumentation 😅

My opinion - we should not handle this case at all, as it will be anyway a problem on the consumer's side if they are not closing response bodies (regardless if sentry is in the app or not). I'd rather us create an event in this case and warn the users they have leaked resources (but this is probably a bigger initiative to detect or `Closeable`s that are not closed)

stefanosiano:
> > So the issue we're seeing here is a misuse of OkHttp itself, which in turn breaks our instrumentation 😅
> 
> My opinion - we should not handle this case at all, as it will be anyway a problem on the consumer's side if they are not closing response bodies (regardless if sentry is in the app or not). I'd rather us create an event in this case and warn the users they have leaked resources (but this is probably a bigger initiative to detect or `Closeable`s that are not closed)

i agree to some extent. It's true that is user's fault, but not all calls require reading the response body. 
Think about a background task that sends some unimportant data, and you don't care if the server received it or not. You simply don't read the response. This is just an example coming from personal experience.
In that case we could have a transaction running forever, and that would break other features of the SDK. The philosophy should be "Even if the user fails, we shouldn't fail with him"

Issue #2917, opened 2023-09-05
Improve Screenshot/ViewHierarchy debouncing default behavior


kahest:
### Description

Currently, capture events for Screenshots/ViewHierarchies are debounced based on time proximity with no consideration of the number of events per time frame. While this can be overridden using `BeforeCaptureCallback` to provide completely custom logic, the default behavior could be improved.

Suggestion is to introduce a threshold for number of events within a certain time frame. Below the threshold, no debouncing is applide by default.

PR #2945, opened 2023-09-20
improve default debouncing mechanism

stefanosiano:
## :scroll: Description
improve default SS/VH debouncing mechanism: after 3 times occurring in a time frame, the next occurrence is debounced
added execution counter to Debouncer

Closes https://github.com/getsentry/sentry-java/issues/2917

Unit test

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/internal/util/Debouncer.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/internal/util/Debouncer.java
@@ -1,11 +1,10 @@
 package io.sentry.android.core.internal.util;
 
 import io.sentry.transport.ICurrentDateProvider;
-import org.jetbrains.annotations.ApiStatus;
-import org.jetbrains.annotations.NotNull;
-
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
+import org.jetbrains.annotations.ApiStatus;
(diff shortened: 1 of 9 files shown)


stefanosiano:
There is also another way: debounce randomly after the first event, with a hard limit of debouncing in a specific time frame.
In other words, allow SS/VH (and increase the counter) randomly after the first event, and not on every event.

stefanosiano:
> There is also another way: debounce randomly after the first event, with a hard limit of debouncing in a specific time frame. In other words, allow SS/VH (and increase the counter) randomly after the first event, and not on every event.

Thinking more about it, having random number would make it possible to have events with View Hierarchy, but without Screeshot and viceversa. Probably good enough as is

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/Debouncer.java
```diff
@@ -10,12 +10,18 @@ public class Debouncer {
 
   private final long waitTimeMs;
   private final @NotNull ICurrentDateProvider timeProvider;
+  private int executions = 0;
```
romtsn, line 13: `m`: I feel like this should be atomic (and also `lastExecutionTime` fwiw), because `checkForDebounce` can be called from multiple threads simultaneously in theory

romtsn (approved):
LGTM, besides one comment wrt concurrency

PR #2961, opened 2023-09-29
SentryOkHttpEvent report exceptions only on the call root span

stefanosiano:
## :scroll: Description
SentryOkHttpEvent now shows exceptions only on the call root span

Having the expection shown in the inner spans of the http call could be confusing, as in https://github.com/sentry-demos/android/issues/60

Unit

## :pencil: Checklist

### DIFF
--- a/sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEvent.kt
+++ b/sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpEvent.kt
@@ -29,6 +29,7 @@ private const val ERROR_MESSAGE_KEY = "error_message"
 private const val RESPONSE_BODY_TIMEOUT_MILLIS = 500L
 internal const val TRACE_ORIGIN = "auto.http.okhttp"
 
+@Suppress("TooManyFunctions")
 internal class SentryOkHttpEvent(private val hub: IHub, private val request: Request) {
     private val eventSpans: MutableMap<String, ISpan> = ConcurrentHashMap()
     private val breadcrumb: Breadcrumb
@@ -119,8 +120,10 @@ internal class SentryOkHttpEvent(private val hub: IHub, private val request: Req
         val span = eventSpans[event] ?: return null
         val parentSpan = findParentSpan(event)
         beforeFinish?.invoke(span)
+        moveThrowableToRootSpan(span)
         if (parentSpan != null && parentSpan != callRootSpan) {
             beforeFinish?.invoke(parentSpan)
+            moveThrowableToRootSpan(parentSpan)
         }
         callRootSpan?.let { beforeFinish?.invoke(it) }
         span.finish()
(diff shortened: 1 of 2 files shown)


stefanosiano:
This is not going to work with errors captured by the [interceptor](https://github.com/getsentry/sentry-java/blob/fix/okhttp_exception_root_only/sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpInterceptor.kt#L117), correct @romtsn @markushi ?
From my understanding we are not setting the throwable to the span explicitly 😕

romtsn:
@stefanosiano have you got a link to a sample event after this change?

stefanosiano:
> @stefanosiano have you got a link to a sample event after this change?

yep. [This is before](https://sentry-sdks.sentry.io/performance/sentry-android:54fc72880e8d42e384528745f98cedf9/?project=5428559&query=&referrer=performance-transaction-summary&showTransactions=recent&statsPeriod=1h&transaction=MainActivity&unselectedSeries=p100%28%29&unselectedSeries=avg%28%29) (the version in master)
[This is after](https://sentry-sdks.sentry.io/performance/sentry-android:530e485e9e9347fd9ca6d9d3bd877d01/?project=5428559&query=&referrer=performance-transaction-summary&showTransactions=recent&statsPeriod=1h&transaction=MainActivity&unselectedSeries=p100%28%29&unselectedSeries=avg%28%29).
Sadly i found an edge case: manually starting a span while reading the response (which the demo app does) leads to [this](https://sentry-sdks.sentry.io/performance/sentry-android:c7e03e19f9ab4b0fa0ceb295bf03cd3b/?project=5428559&query=&referrer=performance-transaction-summary&showTransactions=recent&statsPeriod=1h&transaction=MainActivity&unselectedSeries=p100%28%29&unselectedSeries=avg%28%29)
I have to think about fixing that edge case, but it could be done on a separate PR

stefanosiano:
Actually, the edge case happens also when using the Interceptor only, even if the EventListener collects more spans for a few more milliseconds than the Interceptor, so it's a little "worse"

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 - Add `CheckInUtils.withCheckIn` which abstracts away some of the manual check-ins complexity ([#2959](https://github.com/getsentry/sentry-java/pull/2959))
 
+### Fixes
+
+- SentryOkHttpEvent report exceptions only on the call root span ([#2961](https://github.com/getsentry/sentry-java/pull/2961))
```
markushi, line 11: It's actually HTTP errors we're catching here, how about naming it this way? 
```suggestion
- Always attach `SentryOkHttpEvent` errors to the call root span ([#2961](https://github.com/getsentry/sentry-java/pull/2961))
```

markushi (approved):
LGTM!

CHANGELOG.md
stefanosiano: Actually it's both
any exception caught by the listener is now associated to the call root span only, and the `SentryHttpClientException` is also associated to the call root span only.
What about "Always attach `OkHttp` errors and `Http Client Errors` only to call root span"?

PR #2968, opened 2023-10-05
Fix json parsing of nullable/empty fields

### LINKED ISSUE #2151: `JsonObjectReader#nextList` ideally is renamed to `nextListOrNull` in case the list is empty or invalid
### Description

```
{
  "breadcrumbs": []
}
```

Currently, an empty list that is deserialized via `nextList` throws if the list is empty or invalid.
The logger is `Failed to deserialize object in list.`
The SDK degrades gracefully but we should not set the error level to `ERROR` in this case, either we do `DEBUG` or not log at all if the list is empty.
Users might understand that there's something wrong but there's not.

romtsn:
## :scroll: Description

* Renames `nextList` to `nestListOrNull` to actually match what the method does
* Checks if there's any object in a collection before trying to parse it (which prevents the "Failed to deserilize object in list" log message)
* If a date can't be parsed as ISO timestamp, attempts to parse it as millis silently, without printing a log message
* If `op` is not defined as part of `SpanContext`, fallback to an empty string. This field is actually optional in the spec, but we made it non-optional, as we also use the same structure across the project everywhere, where the `op` field is actually required. So we just duct-tape it for hybrid SDKs here. See https://develop.sentry.dev/sdk/event-payloads/contexts/#trace-context

Closes #2151 
Closes https://github.com/getsentry/sentry-capacitor/issues/450

Manually + automated

## :pencil: Checklist

### DIFF
--- a/sentry/api/sentry.api
+++ b/sentry/api/sentry.api
@@ -770,7 +770,7 @@ public final class io/sentry/JsonObjectReader : io/sentry/vendor/gson/stream/Jso
 	public fun nextFloat ()Ljava/lang/Float;
 	public fun nextFloatOrNull ()Ljava/lang/Float;
 	public fun nextIntegerOrNull ()Ljava/lang/Integer;
-	public fun nextList (Lio/sentry/ILogger;Lio/sentry/JsonDeserializer;)Ljava/util/List;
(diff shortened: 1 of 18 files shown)


markushi (approved):
Nice one, looks good to me!

stefanosiano (approved):

Issue #2912, opened 2023-09-04
Crash in SentryFrameMetricsCollector


markushi:
### Description

Reported by Play Console

````
Exception java.lang.RuntimeException:
  at android.view.DisplayEventReceiver.nativeInit
  at android.view.DisplayEventReceiver.<init> (DisplayEventReceiver.java:113)
  at android.view.Choreographer$FrameDisplayEventReceiver.<init> (Choreographer.java:1214)
  at android.view.Choreographer.<init> (Choreographer.java:291)
  at android.view.Choreographer.<init>
  at android.view.Choreographer$1.initialValue (Choreographer.java:124)
  at android.view.Choreographer$1.initialValue (Choreographer.java:117)
  at java.lang.ThreadLocal.setInitialValue (ThreadLocal.java:180)
  at java.lang.ThreadLocal.get (ThreadLocal.java:170)
  at android.view.Choreographer.getInstance (Choreographer.java:319)
  at io.sentry.android.core.internal.util.SentryFrameMetricsCollector.lambda$new$1 (SentryFrameMetricsCollector.java:96)
  at android.os.Handler.handleCallback (Handler.java:942)
  at android.os.Handler.dispatchMessage (Handler.java:99)
````

romtsn:
Let's just add a try-catch there as the underlying problem is somewhere in the guts of Choreographer

PR #2970, opened 2023-10-05
Fix crash accessing Choreographer instance

stefanosiano:
## :scroll: Description
choreographer instance is now retrieved in a try/catch block, due to a crash reported by Google Play SDK console

Fixes a crash. Closes https://github.com/getsentry/sentry-java/issues/2912

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
@@ -94,7 +94,20 @@ public final class SentryFrameMetricsCollector implements Application.ActivityLi
     // https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:metrics/metrics-performance/src/main/java/androidx/metrics/performance/JankStatsApi24Impl.kt
 
     // The Choreographer instance must be accessed on the main thread
-    new Handler(Looper.getMainLooper()).post(() -> choreographer = Choreographer.getInstance());
+    new Handler(Looper.getMainLooper())
+        .post(
+            () -> {
+              try {
+                choreographer = Choreographer.getInstance();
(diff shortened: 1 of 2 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
```diff
@@ -94,7 +94,20 @@ public SentryFrameMetricsCollector(
     // https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:metrics/metrics-performance/src/main/java/androidx/metrics/performance/JankStatsApi24Impl.kt
 
     // The Choreographer instance must be accessed on the main thread
-    new Handler(Looper.getMainLooper()).post(() -> choreographer = Choreographer.getInstance());
+    new Handler(Looper.getMainLooper())
+        .post(
+            () -> {
+              try {
+                choreographer = Choreographer.getInstance();
```
markushi, line 101: Since choregrapher is accessed via a handler outside the main thread I think it's a good idea to mark the field as `volatile` as well.

markushi (approved):
LGTM, please check my comment!

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
stefanosiano: I'm not sure, since it's only set once and read multiple times. We'd lose memory cache for that.
In the worst case now, when it's read it's still null, so the frame start will fallback to [now - frameDuration](https://github.com/getsentry/sentry-java/blob/bb0533f5d889d1aa7b81436d167fe06de3ca5935/sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java#L118-L120)

Issue #2986, opened 2023-10-15
`Failed to initialize SystemEventsBreadcrumbsIntegration` when starting sentry


ln-12:
### Integration

sentry-android

### Build System

Gradle

### AGP Version

8.1.0

### Proguard

Enabled

### Version

6.31.0

### Steps to Reproduce

Currently, when I run sentry on a device with Android 14, the `SystemEventsBreadcrumbsIntegration` cannot be initialized. From the stack trace I can see that the error is caused in [this line](https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/Sentry.java#L237). Starting from Android 14, it is necessary to explicitly set the export behavior for runtime-registered broadcasts as described [here](https://developer.android.com/about/versions/14/behavior-changes-14).

### Expected Result

No crash when starting sentry.

### Actual Result

The following error is thrown:
```
Failed to initialize SystemEventsBreadcrumbsIntegration.
java.lang.SecurityException: my.app.debug: One of RECEIVER_EXPORTED or RECEIVER_NOT_EXPORTED should be specified when a receiver isn't being registered exclusively for system broadcasts
	at android.os.Parcel.createExceptionOrNull(Parcel.java:3057)
	at android.os.Parcel.createException(Parcel.java:3041)
	at android.os.Parcel.readException(Parcel.java:3024)
	at android.os.Parcel.readException(Parcel.java:2966)
	at android.app.IActivityManager$Stub$Proxy.registerReceiverWithFeature(IActivityManager.java:5684)
	at java.lang.reflect.Method.invoke(Native Method)
	at leakcanary.ServiceWatcher$install$4$2.invoke(ServiceWatcher.kt:93)
	at java.lang.reflect.Proxy.invoke(Proxy.java:1006)
	at $Proxy3.registerReceiverWithFeature(Unknown Source)
	at android.app.ContextImpl.registerReceiverInternal(ContextImpl.java:1852)
	at android.app.ContextImpl.registerReceiver(ContextImpl.java:1792)
	at android.app.ContextImpl.registerReceiver(ContextImpl.java:1780)
	at android.content.ContextWrapper.registerReceiver(ContextWrapper.java:755)
	at io.sentry.android.core.SystemEventsBreadcrumbsIntegration.register(SystemEventsBreadcrumbsIntegration.java:102)
	at io.sentry.Sentry.init(Sentry.java:237)
	at io.sentry.Sentry.init(Sentry.java:147)
[...]

PR #2990, opened 2023-10-16
Set export flag on Android receivers on API 33+

stefanosiano:
## :scroll: Description
registerReceiver now sets the RECEIVER_NOT_EXPORTED flag on API 33+
updated SDK target version to 34

Starting from Android 14, it is necessary to explicitly set the export behavior for runtime-registered broadcasts as described [here](https://developer.android.com/about/versions/14/behavior-changes-14#runtime-receivers-exported).
Fixes https://github.com/getsentry/sentry-java/issues/2986

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ContextUtils.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ContextUtils.java
@@ -2,7 +2,7 @@ package io.sentry.android.core;
 
 import static android.app.ActivityManager.RunningAppProcessInfo.IMPORTANCE_FOREGROUND;
 import static android.content.Context.ACTIVITY_SERVICE;
-import static android.content.Context.RECEIVER_NOT_EXPORTED;
+import static android.content.Context.RECEIVER_EXPORTED;
 import static android.content.pm.PackageInfo.REQUESTED_PERMISSION_GRANTED;
 
 import android.annotation.SuppressLint;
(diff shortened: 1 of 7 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/ContextUtils.java
```diff
@@ -346,6 +351,29 @@ static boolean isForegroundImportance(final @NotNull Context context) {
...
+
+  /** Register a not exported BroadcastReceiver, independently from platform version. */
+  @SuppressLint({"NewApi", "UnspecifiedRegisterReceiverFlag"})
+  static @Nullable Intent registerReceiver(
+      final @NotNull Context context,
+      final @NotNull BuildInfoProvider buildInfoProvider,
+      final @Nullable BroadcastReceiver receiver,
+      final @NotNull IntentFilter filter) {
+    if (buildInfoProvider.getSdkInfoVersion() >= Build.VERSION_CODES.TIRAMISU) {
+      return context.registerReceiver(receiver, filter, RECEIVER_NOT_EXPORTED);
```
romtsn, line 371: `h`: hm, as I understood we should set it to `RECEIVER_EXPORTED` if we're listening for any system events. From https://developer.android.com/guide/components/broadcasts#context-registered-receivers:

>  If this receiver is listening for broadcasts sent from the system or from other apps—even other apps that you own—use the RECEIVER_EXPORTED flag

sentry-android-core/src/test/java/io/sentry/android/core/ContextUtilsUnitTests.kt
```diff
@@ -161,4 +166,28 @@ class ContextUtilsUnitTests {
...
+        val receiver = mock<BroadcastReceiver>()
+        val filter = mock<IntentFilter>()
+        val context = mock<Context>()
+        whenever(buildInfo.sdkInfoVersion).thenReturn(Build.VERSION_CODES.S)
+        ContextUtils.registerReceiver(context, buildInfo, receiver, filter)
+        verify(context).registerReceiver(eq(receiver), eq(filter))
+    }
+/*
+    context.registerReceiver(BroadcastReceiver, IntentFilter, int) throws with NoSuchMethod exception in tests
+    Re-enable this test when possible
```
romtsn, line 182: this is probably because Robolectric is running it with API 32 or lower by default? You can probably try to set the SDK version for this specific test:
```
@Config(sdk=[33])
@Test
...
```

romtsn (approved):
LGTM, besides a couple of comments 👍

sentry-android-core/src/main/java/io/sentry/android/core/ContextUtils.java
stefanosiano: yes, exactly 😅

sentry-android-core/src/test/java/io/sentry/android/core/ContextUtilsUnitTests.kt
stefanosiano: Was looking for this, thanks!

Issue #2997, opened 2023-10-19
[Android] Tag all spans with main thread flag


markushi:
See span data conventions https://develop.sentry.dev/sdk/performance/span-data-conventions/#thread.

PR #2998, opened 2023-10-19
Tag all spans with main thread flag

markushi:
## :scroll: Description
Whenever SentryTracer creates a span, it immediately sets the main thread flag.

Fixes https://github.com/getsentry/sentry-java/issues/2997

Open questions: How do deal with our existing main thread flagging.
E.g. SentryFileInputStream sets the flag on `.close()` and right now would overwrite the flag - IMHO that's fine.

Add Unit test

## :pencil: Checklist

### DIFF
--- a/sentry/src/main/java/io/sentry/SentryTracer.java
+++ b/sentry/src/main/java/io/sentry/SentryTracer.java
@@ -7,7 +7,6 @@ import io.sentry.protocol.SentryTransaction;
 import io.sentry.protocol.TransactionNameSource;
 import io.sentry.protocol.User;
 import io.sentry.util.Objects;
-import io.sentry.util.thread.MainThreadChecker;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.ListIterator;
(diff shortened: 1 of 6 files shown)


sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -384,6 +385,8 @@ private ISpan createChild(
               }
             });
     span.setDescription(description);
+    span.setData(
```
romtsn, line 388: Did we agree to use the same `blocked_main_thread` key? I'd rather us choose something different, because we don't know for sure if it blocked the main thread, we know that it was just started there? Also, if we pick a different key, the current behavior for File I/O or DB spans will remain the same with `blocked_main_thread`.

markushi: Nope, this wasn't discussed yet - thanks for bringing this up. @shruthilayaj I think Roman has some good points here, would it make sense for you as well to have a separate flag for main thread flagging?

shruthilayaj: Yes, it makes sense to have these as separate flags!

shruthilayaj: 👍 
just to confirm, sending `thread.id` and `thread.name` in addition to `started_on_main_thread` flag right?

romtsn: mm, since in the RFC we say it's the thread the span originated from, I think `thread.name: main` would be the same thing pretty much, so we don't need both flags?
image

shruthilayaj: yeah makes sense, and will thread name always be standardized to `main` if it started on the main thread regardless of the SDK it's coming from?

romtsn: I'd start with just `main`, I guess we can expand it later. We can ensure that it's called `main` from the SDKs side

romtsn (approved):

Issue #2905, opened 2023-08-24
The application crashes with StrictMode error after sentry initialization on android api 31, 32, 33.


AntonVoronkin:
### Integration

sentry-android

### Build System

Gradle

### AGP Version

7.4.1

### Proguard

Enabled

### Version

6.27.0

### Steps to Reproduce

**The application crashes with StrictMode error after sentry initialization on android api 31, 32, 33.**
**StrictMode** policy for detect the SQL leaked, the closable objects leaked and detecting network operations.

Apparently, the issue is related to "android.appcompat:appcompat-resources:1.4.2 and android.appcompat:appcompat:1.4.2" entries present in the sentry-external-modules.txt file.

Sentry plugin version: 3.11.1; sentry version: 6.27.0 (6.28.0 dont' fix this problem); targetAndroidSdkVersion 31; minAndroidSdkVersion 24

Error example: 

```
StrictMode policy violation: android.os.strictmode.LeakedClosableViolation: A resource was acquired at attached stack trace but never released. See java.io.Closeable for information on avoiding resource leaks.
                                                                                                    	at android.os.StrictMode$AndroidCloseGuardReporter.report(StrictMode.java:1987)
                                                                                                    	at dalvik.system.CloseGuard.warnIfOpen(CloseGuard.java:336)
                                                                                                    	at sun.nio.ch.FileChannelImpl.finalize(FileChannelImpl.java:175)
                                                                                                    	at java.lang.Daemons$FinalizerDaemon.doFinalize(Daemons.java:319)
                                                                                                    	at java.lang.Daemons$FinalizerDaemon.runInternal(Daemons.java:306)
                                                                                                    	at java.lang.Daemons$Daemon.run(Daemons.java:140)
                                                                                                    	at java.lang.Thread.run(Thread.java:1012)
                                                                                                    Caused by: java.lang.Throwable: Explicit termination method 'close' not called
                                                                                                    	at dalvik.system.CloseGuard.openWithCallSite(CloseGuard.java:288)
                                                                                                    	at dalvik.system.CloseGuard.open(CloseGuard.java:257)
                                                                                                    	at sun.nio.ch.FileChannelImpl.<init>(FileChannelImpl.java:110)
                                                                                                    	at sun.nio.ch.FileChannelImpl.open(FileChannelImpl.java:120)
                                                                                                    	at java.io.FileInputStream.getChannel(FileInputStream.java:475)
[...]

PR #3000, opened 2023-10-19
Make sure to close all Closeable resources

stefanosiano:
## :scroll: Description
changed all stream operations to try-with-resources blocks
added super.close() to SentryFileInputStream and SentryFileOutputStream

Relates to https://github.com/getsentry/sentry-java/issues/2905
It is not fixed entirely, though. We have to investigate the native layer: removing it makes this issue disappear.

## :pencil: Checklist

### DIFF
--- a/sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
+++ b/sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
@@ -112,6 +112,7 @@ public final class SentryFileInputStream extends FileInputStream {
   @Override
   public void close() throws IOException {
     spanManager.finish(delegate);
+    super.close();
   }
 
   private static FileDescriptor getFileDescriptor(final @NotNull FileInputStream stream)
--- a/sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
+++ b/sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
@@ -112,7 +112,6 @@ public final class SentryFileInputStream extends FileInputStream {
   @Override
   public void close() throws IOException {
     spanManager.finish(delegate);
-    super.close();
   }
 
   private static FileDescriptor getFileDescriptor(final @NotNull FileInputStream stream)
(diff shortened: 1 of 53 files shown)


sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
```diff
@@ -112,6 +112,7 @@ public long skip(final long n) throws IOException {
   @Override
   public void close() throws IOException {
     spanManager.finish(delegate);
+    super.close();
```
markushi, line 115: I don't think we need to do this, as SentryFileInputStream isn't backed by any real inputstream, but rather the delegate is used - but it shouldn't hurt, right @romtsn ?

sentry/src/main/java/io/sentry/internal/debugmeta/ResourcesDebugMetaLoader.java
```diff
@@ -30,19 +30,23 @@ public ResourcesDebugMetaLoader(final @NotNull ILogger logger) {
...
-    } else {
-      try (final InputStream is = new BufferedInputStream(debugMetaStream)) {
-        final Properties properties = new Properties();
-        properties.load(is);
-        return properties;
-      } catch (IOException e) {
-        logger.log(SentryLevel.ERROR, e, "Failed to load %s", DEBUG_META_PROPERTIES_FILENAME);
-      } catch (RuntimeException e) {
-        logger.log(SentryLevel.ERROR, e, "%s file is malformed.", DEBUG_META_PROPERTIES_FILENAME);
+    try (InputStream debugMetaStream =
```
markushi, line 33: nit-picking: could be marked as final

markushi (approved):
LGTM, nice work! Please have a look at my comments.

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
romtsn: Yes, I wanted to leave the same comment - I'd rather remove these `super` calls, not sure if they will mess up something, but since we're using a delegate, it should be fine. E.g. apache-commons also don't call super, but just delegate https://github.com/apache/commons-io/blob/43d3ac6cbdb419a1929870b04be181da20be7f03/src/main/java/org/apache/commons/io/input/ProxyInputStream.java#L119-L122

stefanosiano: agree, was just thinking if strict mode could have a check on those close() functions, but it seems it doesn't.
Looks safe to remove it

Issue #2978, opened 2023-10-10
Jetpack Compose user interaction integration is broken


audkar:
### Integration

sentry-android

### Build System

Gradle

### AGP Version

8.1

### Proguard

Disabled

### Version

6.30.0

### Steps to Reproduce

1. Setup project according to https://docs.sentry.io/platforms/android/configuration/integrations/jetpack-compose/#user-interactions
2. Add any `clickable`/`scrollable` Composable
```kotlin
@Composable
fun MainScreen() {
        Button(
            modifier = Modifier.testTag("btn"),
            onClick = { },
        ) { Text(text = "The button") }
}
```
3. Interact with the Composable
4. Observe Logcat output. e.g. `Unable to find click target. No breadcrumb captured.`

This can be reproduced on the `sentry-samples-android` module in this repo on the `main` branch. "Open Compose Activity"

I think it's related to https://github.com/getsentry/sentry-java/blob/a0e3c8b1903e6d38912f2a592c812b61b10a95d4/sentry-compose-helper/src/jvmMain/java/io/sentry/compose/gestures/ComposeGestureTargetLocator.java#L73 changed type of Semantic nodes

### Expected Result

Click properly registered for Compose component

### Actual Result

Target UI component is not detected. Logcat output: `Unable to find click target. No breadcrumb captured.`

audkar:
Are there any tests that cover the real behavior of Compose user interaction tracking? I wanted to quickly check if I could fix this problem, but could not find any tests I could use for this investigation.

markushi:
Thanks for reporting!
Yes, we do have some basic automated UI tests for this: https://github.com/getsentry/sentry-java/blob/main/sentry-android-integration-tests/sentry-uitest-android/src/androidTest/java/io/sentry/uitest/android/UserInteractionTests.kt

Could you share which version of Jetpack Compose you're using? Especially the compose compiler plugin and the `compose-ui` pakcage.

audkar:
compiler - `androidx.compose.compiler:compiler:1.5.3`
ui  from BOM - `androidx.compose:compose-bom:2023.09.02` `1.5.2`

audkar:
Do you have any insight into what the problem might be? Are we using the Sentry Compose integration incorrectly, is it a bug, does Sentry not support stable Android Compose, or something else?

audkar:
Did you have a change to :eyes: this? We are kind of seriously blocked by this.

markushi:
[...]

PR #3010, opened 2023-10-23
Fix user interaction tracking not working for Jetpack Compose 1.5+

markushi:
## :scroll: Description
Extends the behaviour for determining if a LayoutNode is clickable/scrollable. This change should be backwards compatible as the old logic still stays intact for older versions of Jetpack Compose.

Fixes https://github.com/getsentry/sentry-java/issues/2978

Manual tests, we definitely should setup a test matrix for different Jetpack Compose versions.


romtsn (approved):
> Manual tests, we definitely should setup a test matrix for different Jetpack Compose versions.

💯 I think our agp-matrix workflow already runs compose instrumentation tests, so we should just add another dimension there. Will be hard to keep it in sync though

PR #3016, opened 2023-10-30
chore(profiling): Create independent Profiler for Hybrid SDKs

krystofwoldrich:
## :scroll: Description

This PR extracts Android Profiling and Profile Metrics to separate class that doesn't contain any transaction logic.

This is preparation for mixed profiles in Hybrid SDKs, where the Hybrid SDKs manage the transactions.

Changes that have been done for mixed profiler for RN iOS
- https://github.com/getsentry/sentry-react-native/pull/3349

## :pencil: Checklist

## :crystal_ball: Next steps
- [x] Add tests


krystofwoldrich:
@stefanosiano Is `Integration Tests / Ui tests` flaky or did I cause the flakiness with my changes?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidProfiler.java
```diff
@@ -0,0 +1,360 @@
...
+    this.intervalUs = intervalUs;
+    this.options = Objects.requireNonNull(sentryAndroidOptions, "SentryAndroidOptions is required");
+    this.frameMetricsCollector =
+        Objects.requireNonNull(frameMetricsCollector, "SentryFrameMetricsCollector is required");
+    this.buildInfoProvider =
+        Objects.requireNonNull(buildInfoProvider, "The BuildInfoProvider is required.");
+  }
+
+  @SuppressLint("NewApi")
+  public @Nullable ProfileStartData start() {
```
markushi, line 110: I guess we can't always be certain on which thread profiling is started / stopped?
How about marking start(), endCollection() and close() `synchronized`? This way e.g.
 *  `isRunning` would at least be guarded and parallel calls to e.g. `start()` wouldn't cause any invalid state.
 * `endCollection()` wouldn't be executed in parallel in case it's called by scheduleFinish + manually at the same time

krystofwoldrich: Make sense to me, I've updated the methods.

markushi (approved):

stefanosiano:
> @stefanosiano Is `Integration Tests / Ui tests` flaky or did I cause the flakiness with my changes?

I'm currently already investigating this
you didn't cause anything, don't worry 😅

PR #3036, opened 2023-11-07
Reduce main thread work on init

stefanosiano:
## :scroll: Description
Moved old profiles deletion to the background
moved some Integration.register() to the background using executorService
added tests

This is an attempt to improve SDK startup timing.
Relates to https://github.com/getsentry/sentry-java/issues/2222

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Fixes
 
 - Reduce main thread work on init ([#3036](https://github.com/getsentry/sentry-java/pull/3036))
-- Move Integrations registration to background on init ([#3043](https://github.com/getsentry/sentry-java/pull/3043))
+  - Move Integrations registration to background on init ([#3043](https://github.com/getsentry/sentry-java/pull/3043))
 
 ## 6.34.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 ### Fixes
 
 - Reduce main thread work on init ([#3036](https://github.com/getsentry/sentry-java/pull/3036))
+- Move Integrations registration to background on init ([#3043](https://github.com/getsentry/sentry-java/pull/3043))
 -  Fix SIGSEV, SIGABRT and SIGBUS crashes happening after/around the August Google Play System update, see [#2955](https://github.com/getsentry/sentry-java/issues/2955) for more details (fix provided by Native SDK bump)
 - Ensure DSN uses http/https protocol ([#3044](https://github.com/getsentry/sentry-java/pull/3044))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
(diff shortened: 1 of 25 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserverIntegration.java
```diff
@@ -37,28 +37,42 @@ public final void register(final @NotNull IHub hub, final @NotNull SentryOptions
...
       } catch (Throwable e) {
-        // it could throw eg NoSuchFileException or NullPointerException
-        options
-            .getLogger()
-            .log(SentryLevel.ERROR, "Failed to initialize EnvelopeFileObserverIntegration.", e);
+        logger.log(
+            SentryLevel.DEBUG,
+            "Failed to start EnvelopeFileObserverIntegration on executor thread. Starting on the calling thread.",
+            e);
+        startOutboxSender(hub, options, path);
```
markushi, line 47: `m`: I'm not sure if this is a good idea. I guess it can happen that the SDK is closed directly after init, e.g. due to double init or maybe a remote feature flag to disable the SDK. In those scenarios we shouldn't fallback. Thus maybe it's simpler to remove this logic and just log a warning.

sentry-android-core/src/main/java/io/sentry/android/core/EnvelopeFileObserverIntegration.java
```diff
@@ -37,28 +37,42 @@ public final void register(final @NotNull IHub hub, final @NotNull SentryOptions
...
       }
     }
   }
 
+  private void startOutboxSender(
```
markushi, line 52: `m`: Probably makes sense to synchronize this method and `.close()` to ensure no race conditions when accessing the observer field.
We probably even need a `isClosed` field in case the executor service is running `startOutboxSender()` after `close()` is called.

sentry/src/main/java/io/sentry/Sentry.java
```diff
@@ -335,18 +335,21 @@ private static boolean initConfigurations(final @NotNull SentryOptions options)
...
       profilingTracesDir.mkdirs();
-      final File[] oldTracesDirContent = profilingTracesDir.listFiles();
+      final long timestamp = System.currentTimeMillis();
 
       try {
         options
             .getExecutorService()
             .submit(
                 () -> {
+                  final File[] oldTracesDirContent = profilingTracesDir.listFiles();
```
markushi, line 345: Nice!

markushi (changes requested):
Looking good, I left a few comments which need to be addressed before merging. 🚀

markushi (approved):
LGTM!

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 
 ### Fixes
 
+- Reduce main thread work on init ([#3036](https://github.com/getsentry/sentry-java/pull/3036))
```
romtsn, line 13: This need a merge with `main` and moving the changelog entry under Unreleased

romtsn (approved):

PR #3074, opened 2023-12-01
Revamp CI setup

romtsn:
_#skip-changelog_

* Speed up some tests depending on timeouts
* Use gradle-build-action for most of the jobs
* Split up PR and release workflows, for releases we only build distribution artifacts now, and for PRs we only run tests and lint
* Enable gradle caching to benefit form gradle-build-action
* Only run tasks that we actually need for workflows
* Parallelize tests
* Remove unnecessary tests (e.g. SdkBenchmark init test, already done as part of the app-metrics job)
* Potentially fix an issue when we didn't set the `release` and other fields correctly for ANRv2


.github/workflows/agp-matrix.yml
```diff
@@ -73,6 +63,7 @@ jobs:
           disable-animations: true
           disable-spellchecker: true
           target: 'aosp_atd'
+          arch: x86
```
markushi, line 66: any specific reason we need `x86` here?

romtsn: yeah, it's faster 👀 or actually that is the default anyway so we could remove it https://github.com/ReactiveCircus/android-emulator-runner#configurations

.github/workflows/agp-matrix.yml
```diff
@@ -4,25 +4,20 @@ on:
...
 
+jobs:
   agp-matrix-compatibility:
     timeout-minutes: 30
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        agp: [ '8.0.0','8.1.0-alpha11' ]
+        agp: [ '8.0.0','8.1.4','8.2.0','8.3.0-beta01' ]
```
stefanosiano, line 20: May we skip some intermediate versions, like the `8.1.4` and perhaps `8.2.0`?

.github/workflows/integration-tests-benchmarks.yml
```diff
@@ -77,6 +79,9 @@ jobs:
           java-version: '17'
           distribution: 'temurin'
 
+      - name: Setup Gradle
```
stefanosiano, line 82: Do we need this again?

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
   - If you are setting global job listeners please also add `SentryJobListener`
 - Ensure serialVersionUID of Exception classes are unique ([#3115](https://github.com/getsentry/sentry-java/pull/3115))
 - Get rid of "is not eligible for getting processed by all BeanPostProcessors" warnings in Spring Boot ([#3108](https://github.com/getsentry/sentry-java/pull/3108))
+- Fix missing `release` and other fields for ANRs reported with `mechanism:AppExitInfo` ([#3074](https://github.com/getsentry/sentry-java/pull/3074))
```
stefanosiano, line 16: is this change wanted?

.github/workflows/agp-matrix.yml
romtsn: yeah, I thought about that, but AGP sometimes introduces a change, and then fixes it in the following version, so I think it's good if we are aware if it's broken in e.g. `8.1.4`,  but fixed in `8.2.0`, it will save us time investigating issues. Also, they run in parallel, so shouldn't be noticable, but I would keep them at 4-5 max.

.github/workflows/integration-tests-benchmarks.yml
romtsn: yep, it's a different job that runs in parellel with the one above, so it also has to go through the full setup afaik

CHANGELOG.md
romtsn: Yeah, I accidentally fixed this issue, because we had a flaky failing test, so I think it's fine if it goes into this PR. Or do you think we need a separate PR for that?

stefanosiano: nah, it's fine, just wanted to be sure it wasn't a reminiscence of another WIP PR 👍

sentry-android-integration-tests/sentry-uitest-android/src/androidTest/java/io/sentry/uitest/android/EnvelopeTests.kt
```diff
@@ -174,6 +177,12 @@ class EnvelopeTests : BaseUiTest() {
 
     @Test
     fun checkTimedOutProfile() {
+        Assume.assumeFalse(
```
stefanosiano, line 180: did you try with the latest changes from master, too?

.github/workflows/integration-tests-benchmarks.yml
stefanosiano: ah yeah, i didn't read the lines in between 😅

stefanosiano (approved):
nice!

sentry-android-integration-tests/sentry-uitest-android/src/androidTest/java/io/sentry/uitest/android/EnvelopeTests.kt
romtsn: you mean your fixes for the envelopes with 0-byte bodies? yeah I tried that, still happening

adinauer (approved):
LGTM

PR #3087, opened 2023-12-06
Send breadcrumbs and client error even without transactions

stefanosiano:
## :scroll: Description
moved breadcrumb sent before span null check in SentryOkHttpEvent
added client error reporting if call root span is null

OkHttp breadcrumb and client error were not sent in case the span was not created, like when no transaction is running.
Fixes https://github.com/getsentry/sentry-android-gradle-plugin/issues/606

Unit tests

## :pencil: Checklist


romtsn (approved):
nice!

PR #3094, opened 2023-12-12
Fix Ui tests receiving empty requests

stefanosiano:
## :scroll: Description
added a check to avoid processing requests with 0 bytes bodies in ui tests (probably useless)
added a check to drop duplicated envelopes

#skip-changelog

Some envelopes were sent twice, corrupting the counter in `MockRelay` and making tests fail

## :pencil: Checklist

### DIFF
--- a/sentry-android-integration-tests/sentry-uitest-android/src/androidTest/java/io/sentry/uitest/android/BaseUiTest.kt
+++ b/sentry-android-integration-tests/sentry-uitest-android/src/androidTest/java/io/sentry/uitest/android/BaseUiTest.kt
@@ -25,6 +25,7 @@ abstract class BaseUiTest {
     /** Application context for the current test. */
     protected lateinit var context: Context
 
+    // The mockDsn cannot be changed. If a custom dsn needs to be used, it can be set in the options as usual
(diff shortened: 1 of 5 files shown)


romtsn:
@stefanosiano it looks like the requests are actually not empty-bodied, but they look super weird, cause they are not errors or transactions really https://github.com/getsentry/sentry-java/actions/runs/7219916222/job/19671823552#step:8:411

stefanosiano:
> @stefanosiano it looks like the requests are actually not empty-bodied, but they look super weird, cause they are not errors or transactions really https://github.com/getsentry/sentry-java/actions/runs/7219916222/job/19671823552#step:8:411

That should be the captured message event, just truncated in that log
The problem was that a test could send the same envelope twice, corrupting the `IdlingResource` counter, but it's not logged as failing test (maybe due to timing?)
So subsequent tests fail because the counter gets always corrupted, since it's not reset.
I think on Saucelabs it happened very rarely because we have a TestOrchestrator that kills the process and cleans everything on every test, while in the AGP matrix it seems it's not run, and our cleanup methods are not enough

romtsn (approved):
🚀

Issue #3099, opened 2023-12-14
[Starfish] Add frame delay for transactions and spans


markushi:
```[tasklist]
### Tasks
- [ ] https://github.com/getsentry/sentry-java/pull/3100
- [ ] https://github.com/getsentry/sentry-java/pull/3102
- [ ] https://github.com/getsentry/sentry-java/issues/3183
```

PR #3100, opened 2023-12-14
Calculate frame delay, slow+frozen frames inside SentryFrameMetricsCollector

markushi:
Do it in one place, so we can later use it for both the AndroidProfiler as well as txns.

## :scroll: Description
Mainly moved some logic from `AndroidProfiler` to `SentryFrameMetricsAggregator`.
Added some logic to calculate the frame delay, as well as exposing frame start/end times.

Partly fixes https://github.com/getsentry/sentry-java/issues/3099

Added unit tests.

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
@@ -146,13 +146,13 @@ public final class SentryFrameMetricsCollector implements Application.ActivityLi
                   : window.getWindowManager().getDefaultDisplay().getRefreshRate();
 
           final long expectedFrameDuration = (long) (oneSecondInNanos / refreshRate);
-          final long totalFrameDuration = frameMetrics.getMetric(FrameMetrics.TOTAL_DURATION);
+
+          final long cpuDuration = getFrameCpuDuration(frameMetrics);
 
           // if totalDurationNanos is smaller than expectedFrameTimeNanos,
           // it means that the frame was drawn within it's time budget, thus 0 delay
-          final long delayNanos = Math.max(0, totalFrameDuration - expectedFrameDuration);
+          final long delayNanos = Math.max(0, cpuDuration - expectedFrameDuration);
 
-          final long cpuDuration = getFrameCpuDuration(frameMetrics);
           long startTime = getFrameStartTimestamp(frameMetrics);
           // If we couldn't get the timestamp through reflection, we use current time
           if (startTime < 0) {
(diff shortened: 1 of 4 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
```diff
@@ -155,8 +167,21 @@ public SentryFrameMetricsCollector(
           lastFrameStartNanos = startTime;
           lastFrameEndNanos = startTime + cpuDuration;
 
+          // Most frames take just a few nanoseconds longer than the optimal calculated
+          // duration.
+          // Therefore we subtract one, because otherwise almost all frames would be slow.
+          final boolean isSlow = cpuDuration > oneSecondInNanos / (refreshRate - 1);
```
markushi, line 173: @stefanosiano I moved this over from `AndroidProfiler` but I'm not sure if this is correct:
We're using the `cpuDuration` here, shouldn't it be `totalFrameduration`?

stefanosiano: cpu duration is fine, we're doing the same of google's JankStats
the cpuDuration `Return time spent on the main thread (cpu) for frame creation. It doesn't consider time spent on the render thread (gpu).`

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
```diff
@@ -140,6 +145,13 @@ public SentryFrameMetricsCollector(
                   ? window.getContext().getDisplay().getRefreshRate()
                   : window.getWindowManager().getDefaultDisplay().getRefreshRate();
 
+          final long expectedFrameDuration = (long) (oneSecondInNanos / refreshRate);
+          final long totalFrameDuration = frameMetrics.getMetric(FrameMetrics.TOTAL_DURATION);
+
+          // if totalDurationNanos is smaller than expectedFrameTimeNanos,
+          // it means that the frame was drawn within it's time budget, thus 0 delay
+          final long delayNanos = Math.max(0, totalFrameDuration - expectedFrameDuration);
```
stefanosiano, line 153: i guess we want to use the `cpuDuration` here, as well
[JankStats's uiDuration](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:metrics/metrics-performance/src/main/java/androidx/metrics/performance/JankStatsApi24Impl.kt;l=74-82) is our cpuDuration, and they use it to calculate jank

stefanosiano (changes requested):
just a change in the cpuDuration vs totalDuration
Everything else is good!

stefanosiano (approved):

PR #3101, opened 2023-12-15
Startup Profiling 1 - Decouple Profiler from Transaction

stefanosiano:
## :scroll: Description
removed the need of a transaction to start a profile in AndroidTransactionProfiler, allowing to bind one later
added a constructor to AndroidTransactionProfiler without options
added options as param to AndroidTransactionProfiler.onTransactionFinish
some cleanup in the profiler

For the startup profiling feature, we need to create a profiler instance before the SDK is initialized, which means we don't have the options and cannot start any transaction, yet.
This change allows to create an instance of the profiler with only the parameters needed, without passing the whole SentryOptions object, and also allows to run the profiler before the transaction, and bind a transaction later on.

Unit tests

## :pencil: Checklist


stefanosiano:
Possible breaking changes:
`options.setTransactionProfiler` now works only if no profiler was set, however this API was never meant to be used by the user
`AndroidTransactionProfiler` has now deprecated constructors, which could be removed completely if they are not used by Hybrid SDKs
`AndroidTransactionProfiler` methods (internal) have changed. This doesn't affect users, but can affect Hybrid SDKs.
@krystofwoldrich Can you confirm these are not actually breaking changes for hybrid SDKs?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
```diff
@@ -153,29 +162,43 @@ private boolean onFirstTransactionStarted(final @NotNull ITransaction transactio
...
-
-    currentProfilingTransactionData =
-        new ProfilingTransactionData(transaction, transactionStartNanos, profileStartCpuMillis);
     return true;
   }
 
+  @Override
+  public synchronized void bindTransaction(final @NotNull ITransaction transaction) {
+    // If the profiler is running, but no profilingTransactionData is set, we bind it here
+    if (transactionsCounter > 0 && currentProfilingTransactionData == null) {
```
markushi, line 173: nit: Given that we're anyway inside `bindTransaction` already, do we need to check the `transactionsCounter` here?

markushi (approved):
Looking good! Nice clean cut 🥷 Just left a minor nit

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
stefanosiano: yep, that `transactionsCounter` checks if the profiler is running.
Without it, you could bind the profiler to a transaction even if it's not running. In that case, `transactionsCounter` wouldn't be reset, so when the profiler starts for another transaction, it would keep the reference to the old one

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Features
 
 - Handle `monitor`/`check_in` in client reports and rate limiter ([#3096](https://github.com/getsentry/sentry-java/pull/3096))
+- Startup profiling 1 - Decouple Profiler from Transaction ([#3101](https://github.com/getsentry/sentry-java/pull/3101))
```
romtsn, line 8: So do you wanna ship this independently from the other PRs? If not, I'd just stack them and merge with only a single changelog entry

romtsn (approved):

PR #3102, opened 2023-12-15
Extend performance collector APIs

markushi:
## :scroll: Description
Similar to collecting cpu and memory stats while a transaction is running, we also want to collect slow and frozen frames. But instead of pulling for data in an interval we already have callbacks by the system which delivers "continuous" data. 

This PR mainly opens up the interfaces so we can have polling aka snapshot collectors (cpu, mem) as well as continuous ones (not part in this PR, but later used for slow+frozen frames as well as frame delay).

Partially fixes https://github.com/getsentry/sentry-java/issues/3099

Added unit tests

## :pencil: Checklist


sentry/src/main/java/io/sentry/DefaultTransactionPerformanceCollector.java
```diff
@@ -102,34 +123,54 @@ public void run() {
...
   public void close() {
-    performanceDataMap.clear();
     options
         .getLogger()
         .log(SentryLevel.DEBUG, "stop collecting all performance info for transactions");
+
+    performanceDataMap.clear();
+    for (final @NotNull IPerformanceContinuousCollector collector : continuousCollectors) {
+      collector.clear();
+    }
```
stefanosiano, line 173: you should also reset the `hasNoCollectors` flag here

stefanosiano (approved):
Just add the flag clear (if needed) and it's good to me!

PR #3111, opened 2023-12-19
Collect slow and frozen frames for spans

markushi:
By utilizing our existing `SentryFrameMetricsCollector` via a new performance collector:`SpanFrameMetricsCollector`

## :scroll: Description
Fixes https://github.com/getsentry/sentry-java/issues/3106

Based on https://github.com/getsentry/sentry-java/pull/3102, which needs to merged first.

If perf-v2 is enabled, the `ActivityFramesTracker` will no-op and instead `SpanFrameMetricsCollector` will start collecting the data on a span level (including the top level transaction `ISpan`).

Top level epic: https://github.com/getsentry/team-mobile/issues/156

Added unit tests.

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
@@ -59,19 +59,6 @@ final class AndroidTransactionProfiler implements ITransactionProfiler {
       final @NotNull Context context,
       final @NotNull SentryAndroidOptions sentryAndroidOptions,
       final @NotNull BuildInfoProvider buildInfoProvider,
-      final @NotNull SentryFrameMetricsCollector frameMetricsCollector) {
-    this(
-        context,
-        sentryAndroidOptions,
-        buildInfoProvider,
-        frameMetricsCollector,
-        HubAdapter.getInstance());
-  }
-
-  public AndroidTransactionProfiler(
-      final @NotNull Context context,
-      final @NotNull SentryAndroidOptions sentryAndroidOptions,
-      final @NotNull BuildInfoProvider buildInfoProvider,
       final @NotNull SentryFrameMetricsCollector frameMetricsCollector,
       final @NotNull IHub hub) {
     this.context = Objects.requireNonNull(context, "The application context is required");
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
@@ -44,6 +44,20 @@ final class AndroidTransactionProfiler implements ITransactionProfiler {
   public AndroidTransactionProfiler(
       final @NotNull Context context,
       final @NotNull SentryAndroidOptions sentryAndroidOptions,
+      final @NotNull BuildInfoProvider buildInfoProvider) {
+    this(
+        context,
+        sentryAndroidOptions,
+        buildInfoProvider,
+        Objects.requireNonNull(
+            sentryAndroidOptions.getFrameMetricsCollector(),
+            "options.getFrameMetricsCollector cannot be null."),
+        HubAdapter.getInstance());
+  }
+
+  public AndroidTransactionProfiler(
+      final @NotNull Context context,
+      final @NotNull SentryAndroidOptions sentryAndroidOptions,
       final @NotNull BuildInfoProvider buildInfoProvider,
       final @NotNull SentryFrameMetricsCollector frameMetricsCollector) {
     this(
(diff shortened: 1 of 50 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/FrameMetrics.java
```diff
@@ -0,0 +1,102 @@
...
+    normalFrameCount = 0;
+
+    slowFrameCount = 0;
+    slowFrameDuration = 0;
+
+    frozenFrameCount = 0;
+    frozenFrameDuration = 0;
+  }
+
+  public FrameMetrics duplicate() {
```
romtsn, line 80: nit: we usually call it `clone()` in other places

sentry-android-core/src/main/java/io/sentry/android/core/FrameMetrics.java
```diff
@@ -0,0 +1,102 @@
...
+    return new FrameMetrics(
+        normalFrameCount, slowFrameCount, slowFrameDuration, frozenFrameCount, frozenFrameDuration);
+  }
+
+  /**
+   * @param other the other frame metrics to compare to, usually the older one
+   * @return the difference between two frame metrics (this minus other)
+   */
+  public FrameMetrics diffTo(FrameMetrics other) {
```
romtsn, line 89:
```suggestion
  public FrameMetrics diffTo(final @NotNull FrameMetrics other) {
```

sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
```diff
@@ -0,0 +1,127 @@
...
+public class SpanFrameMetricsCollector
+    implements IPerformanceContinuousCollector,
+        SentryFrameMetricsCollector.FrameMetricsCollectorListener {
+  private @NotNull final Object lock = new Object();
+  private @Nullable final SentryFrameMetricsCollector frameMetricsCollector;
+  private @Nullable volatile String listenerId;
+  private @NotNull final Map<SpanId, FrameMetrics> metricsAtSpanStart;
+
+  private @NotNull final FrameMetrics currentFrameMetrics;
+  float lastRefreshRate = 0;
```
romtsn, line 26: I guess this should be private, but also didn't see where it's used - is this gonna be in another PR?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
```diff
@@ -41,6 +41,20 @@ final class AndroidTransactionProfiler implements ITransactionProfiler {
   private long transactionStartNanos;
   private long profileStartCpuMillis;
 
+  public AndroidTransactionProfiler(
```
stefanosiano, line 44: is this ctor needed?
I'm going to create an instance of the profiler without the `SentryAndroidOptions` (for startup profiling), so i'd like to avoid adding other constructors with options as params, if possible

sentry-android-core/src/main/java/io/sentry/android/core/FrameMetrics.java
markushi: Yeah, that's on purpose. I didn't want to interfere with Object.clone() / the clonable interface here
image

sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
markushi: Oh true, we can remove this within this PR - thanks!

sentry-android-core/src/main/java/io/sentry/android/core/FrameMetrics.java
```diff
@@ -0,0 +1,102 @@
+package io.sentry.android.core;
+
+import org.jetbrains.annotations.ApiStatus;
+
+@ApiStatus.Internal
+final class FrameMetrics {
```
stefanosiano, line 6: Can you rename it to `SentryFrameMetrics`, to distinguish it from the Android's one?

sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
```diff
@@ -0,0 +1,127 @@
...
+      long durationNanos,
+      long delayNanos,
+      boolean isSlow,
+      boolean isFrozen,
+      float refreshRate) {
+
+    lastRefreshRate = (int) (refreshRate * 100) / 100F;
+
+    if (isFrozen) {
+      currentFrameMetrics.addFrozenFrame(durationNanos);
```
stefanosiano, line 120: should these be `delayNanos`? 
Otherwise you add the full frame duration as slow frame

markushi: True, we should collect the delay instead of the total duration 👍

Issue #2902, opened 2023-08-21
UserFeedback envelope not cached


MarcinGawel-TomTom:
### Integration

sentry-android

### Build System

Gradle

### AGP Version

7.4.2

### Proguard

Disabled

### Version

6.28.0

### Steps to Reproduce

With internet connection disabled:
```
val sentryId = Sentry.captureMessage("message")

val userFeedBack = UserFeedback(sentryId).apply {
    comments = "comment"
    this.name = "reporter name"
    this.email = "email@gmail.com"
}

Sentry.captureUserFeedback(userFeedBack)
```

### Expected Result

captureMessage envelope and captureUserFeedback should be stored locally in cache if internet connection is not available.

### Actual Result

Envelope from captureMessage (containing some attachments) is stored in the cache. Envelope from captureUserFeedback (with comment, name and email) is **not** added to the cache. Looks like EnvelopeCache.getEnvelopeFile() is using the same file for both envelopes.

markushi:
@MarcinGawel-TomTom thanks for reporting, we'll look into this!

PR #3116, opened 2023-12-21
Fix UserFeedback disk cache name conflicts with linked events

markushi:
## :scroll: Description

Until now we used the [event-id (if available) for determining the file name for the disk cache](https://github.com/getsentry/sentry-java/blob/320aa5a6d9cce82a630c1a6a6928d210a3ae443b/sentry/src/main/java/io/sentry/cache/EnvelopeCache.java#L369-L371). But this fails for UserFeedback because it’s event-id can be identical to the event-id of a linked (Crash) Event.

Cocoa is using a [unique file name based on a counter + random UUID](https://github.com/getsentry/sentry-cocoa/blob/3297d6e29cd37151f68e9c605aea8b44e5e432cb/Sources/Sentry/SentryFileManager.m#L260C2-L262), which sounds reasonable, so I resorted to always use as random UUID as well.

Fixes https://github.com/getsentry/sentry-java/issues/2902

Added unit tests

## :pencil: Checklist


stefanosiano (approved):

romtsn (approved):
LGTM, but I only see one potential issue with this: when we are sending envelopes on app start (in DirectoryProcessor), there's a case when rate limiting gets active for an envelope and we're discarding it from cache here: https://github.com/getsentry/sentry-java/blob/937879e7bffe6eceeb2ed339971da869fd0df542/sentry/src/main/java/io/sentry/transport/AsyncHttpTransport.java#L89-L93

With this change, the `.discard` method will not do anything, because it will not find the correct envelope file name, but will just generate a new one. I don't know if we should account for this case, but I guess we should? Or we could actually just remove this check I guess, because #2937 is doing all that already? And we anyway delete the envelope file in proccessFile here: https://github.com/getsentry/sentry-java/blob/937879e7bffe6eceeb2ed339971da869fd0df542/sentry/src/main/java/io/sentry/EnvelopeSender.java#L91-L106

markushi:
> LGTM, but I only see one potential issue with this: when we are sending envelopes on app start (in DirectoryProcessor), there's a case when rate limiting gets active for an envelope and we're discarding it from cache here:
> 
> https://github.com/getsentry/sentry-java/blob/937879e7bffe6eceeb2ed339971da869fd0df542/sentry/src/main/java/io/sentry/transport/AsyncHttpTransport.java#L89-L93
> 
> With this change, the `.discard` method will not do anything, because it will not find the correct envelope file name, but will just generate a new one. I don't know if we should account for this case, but I guess we should? Or we could actually just remove this check I guess, because #2937 is doing all that already? And we anyway delete the envelope file in proccessFile here:
> 
> https://github.com/getsentry/sentry-java/blob/937879e7bffe6eceeb2ed339971da869fd0df542/sentry/src/main/java/io/sentry/EnvelopeSender.java#L91-L106

Yeah, good callout!
But doesn't the `DirectoryProcessor` set the `SendCachedEnvelopeHint` in order to avoid re-caching the same envelope? So `cached` would be false here, which is fine:
https://github.com/getsentry/sentry-java/blob/937879e7bffe6eceeb2ed339971da869fd0df542/sentry/src/main/java/io/sentry/transport/AsyncHttpTransport.java#L89-L93
But I think I'm missing something, as where would the file from cache be finally deleted?

romtsn:
> So cached would be false here, which is fine:

It'll be `true` actually, because `SendCachedEnvelopeHint` implements `Cached`:
https://github.com/getsentry/sentry-java/blob/2465853df157c4ad881166804873989c47c10f4d/sentry/src/main/java/io/sentry/transport/AsyncHttpTransport.java#L81-L85

> @romtsn But I think I'm missing something, as where would the file from cache be finally deleted?

It's done here https://github.com/getsentry/sentry-java/blob/2465853df157c4ad881166804873989c47c10f4d/sentry/src/main/java/io/sentry/EnvelopeSender.java#L97-L99

and looking at this we probably don't even need that `envelopeCache.discard` call, because the envelope is anyway gonna be delete here^

PR #3121, opened 2023-12-29
Startup Profiling 2 - Add options and sampling logic

stefanosiano:
## :scroll: Description
followup of https://github.com/getsentry/sentry-java/pull/3101/
added SentryStartupProfilingOptions class with Json ser/deser
added startupProfilingConfigFile deletion and creation on init
added sampling decision on SDK init with isForNextStartup flag set to true
added SentryOptions.getCacheDirPathWithoutDsn for startupProfiling config file

Here is the decision tree. 
The only missing part is actually starting the profiler and bind it to the first transaction, which will come in the next PR.
Note: the auto instrumentation flag is available on `SentryAndroidOptions`, so it will be evaluated (in the next PR) before starting the startup profiler
photo_2023-12-15 14 57 18

For the startup profiling feature, we need to add an option to enable it and we also need to respect the sampling decisions of the user.
Then, we need to save this info to disk, as the startup profiler will not have access to the `SentryOptions` object, as it will start before the SDK is initialized.

Unit tests

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Handle `monitor`/`check_in` in client reports and rate limiter ([#3096](https://github.com/getsentry/sentry-java/pull/3096))
 - Startup profiling 1 - Decouple Profiler from Transaction ([#3101](https://github.com/getsentry/sentry-java/pull/3101))
 - Startup profiling 2 - Add options and sampling logic ([#3121](https://github.com/getsentry/sentry-java/pull/3121))
+- Startup Profiling 3 - Add ContentProvider and start profile ([#3128](https://github.com/getsentry/sentry-java/pull/3128))
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Handle `monitor`/`check_in` in client reports and rate limiter ([#3096](https://github.com/getsentry/sentry-java/pull/3096))
 - Startup profiling 1 - Decouple Profiler from Transaction ([#3101](https://github.com/getsentry/sentry-java/pull/3101))
+- Startup profiling 2 - Add options and sampling logic ([#3121](https://github.com/getsentry/sentry-java/pull/3121))
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,13 @@
 ### Features
 
 - Handle `monitor`/`check_in` in client reports and rate limiter ([#3096](https://github.com/getsentry/sentry-java/pull/3096))
-- Startup profiling 1 - Decouple Profiler from Transaction ([#3101](https://github.com/getsentry/sentry-java/pull/3101))
-- Startup profiling 2 - Add options and sampling logic ([#3121](https://github.com/getsentry/sentry-java/pull/3121))
-- Startup Profiling 3 - Add ContentProvider and start profile ([#3128](https://github.com/getsentry/sentry-java/pull/3128))
+- Added Startup profiling
+  - This depends on the new option `io.sentry.profiling.enable-startup`, other than the already existing `io.sentry.traces.profiling.sample-rate`.
+  - Sampler functions can check the new `isForNextStartup` flag, to adjust startup profiling sampling programmatically.
+  Relevant PRs:
+  - Decouple Profiler from Transaction ([#3101](https://github.com/getsentry/sentry-java/pull/3101))
+  - Add options and sampling logic ([#3121](https://github.com/getsentry/sentry-java/pull/3121))
+  - Add ContentProvider and start profile ([#3128](https://github.com/getsentry/sentry-java/pull/3128))
 
 ### Fixes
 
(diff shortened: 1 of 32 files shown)


sentry/src/main/java/io/sentry/SentryStartupProfilingOptions.java
```diff
@@ -0,0 +1,146 @@
...
+import io.sentry.vendor.gson.stream.JsonToken;
+import java.io.IOException;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+final class SentryStartupProfilingOptions implements JsonUnknown, JsonSerializable {
+
+  boolean profileSampled;
```
romtsn, line 12: should these be final?

sentry/src/main/java/io/sentry/Sentry.java
```diff
@@ -242,6 +246,58 @@ private static synchronized void init(
...
     finalizePreviousSession(options, HubAdapter.getInstance());
+
+    handleStartupProfilingConfig(options, options.getExecutorService());
+  }
+
+  @SuppressWarnings("FutureReturnValueIgnored")
+  private static void handleStartupProfilingConfig(
+      final @NotNull SentryOptions options,
+      final @NotNull ISentryExecutorService sentryExecutorService) {
+    sentryExecutorService.submit(
```
romtsn, line 257: `m`: I think you also need to wrap `submit` into a try-catch, in case e.g. RejectedExecutionException

sentry/src/main/java/io/sentry/Sentry.java
```diff
@@ -242,6 +246,58 @@ private static synchronized void init(
...
+                      "Tracing is disabled and startup profiling will not start.");
+              return;
+            }
+            try {
+              if (startupProfilingConfigFile.createNewFile()) {
+                final @NotNull TracesSamplingDecision startupSamplingDecision =
+                    sampleStartupProfiling(options);
+                final @NotNull SentryStartupProfilingOptions startupProfilingOptions =
+                    new SentryStartupProfilingOptions(options, startupSamplingDecision);
+                try (Writer fileWriter = new SentryFileWriter(startupProfilingConfigFile)) {
```
romtsn, line 282:
```suggestion
                try (Writer fileWriter = new FileWriter(startupProfilingConfigFile)) {
```
SentryFileWriter is specific for performance instrumentation, I guess we don't want to create spans for our own file operations :)

sentry/src/main/java/io/sentry/Sentry.java
```diff
@@ -242,6 +246,58 @@ private static synchronized void init(
...
+                  .getLogger()
+                  .log(SentryLevel.ERROR, "Unable to create startup profiling config file. ", e);
+            }
+          }
+        });
+  }
+
+  private static @NotNull TracesSamplingDecision sampleStartupProfiling(
+      final @NotNull SentryOptions options) {
+    TransactionContext startupTransactionContext = new TransactionContext("ui.load", "");
```
romtsn, line 297: hm, should `ui.load` be an operation name, and not a transaction name? In that case it should be the 2nd parameter then. Also I don't know how the frontend UI will look like for this, but don't we also need a meaningful transaction name?

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Handle `monitor`/`check_in` in client reports and rate limiter ([#3096](https://github.com/getsentry/sentry-java/pull/3096))
 - Startup profiling 1 - Decouple Profiler from Transaction ([#3101](https://github.com/getsentry/sentry-java/pull/3101))
+- Startup profiling 2 - Add options and sampling logic ([#3121](https://github.com/getsentry/sentry-java/pull/3121))
```
romtsn, line 9: As mentioned in the other PR, maybe let's squash these changelogs into one?

romtsn (approved):
A few points, but LGTM otherwise

sentry/src/main/java/io/sentry/Sentry.java
romtsn: ok, I've just checked the 3rd PR, I guess it doesn't really matter what we use here, because the profile is attached to the first activity tx anyway, right?

romtsn: ok, I see you've changed it there, resolving this

sentry/src/main/java/io/sentry/Sentry.java
romtsn: Also addressed in the 3rd PR, will resolve

stefanosiano: yep, i forgot it, but did it on the third PR

sentry/src/main/java/io/sentry/SentryStartupProfilingOptions.java
stefanosiano: nope, as these are read and deserialized from json (L100)

sentry/src/main/java/io/sentry/Sentry.java
stefanosiano: yep. Also, the name and operation will be "app.launch" and "profile" (in third PR), which doesn't really matter, as the user should just check the "isForNextStartup" flag.
But with those values we aligned Android and Cocoa

stefanosiano:
@romtsn may you have a last look here and [here](https://github.com/getsentry/sentry-java/pull/3101/), as well, before merging? 😄

romtsn:
> @romtsn may you have a last look here and [here](https://github.com/getsentry/sentry-java/pull/3101/), as well, before merging? 😄

yeah looks great, thanks for the thorough changelog!

PR #3128, opened 2024-01-04
Startup Profiling 3 - Add ContentProvider and start profile

stefanosiano:
## :scroll: Description
Added SentryStartupProfilingProvider, which reads the config file from disk, sample the trace and profile, and starts the profiler. Then it stores the startup profiler and sampling decision in AppStartMetrics:
- The sampling decision is inherited by the first activity transaction, if available
- The startup profiler is reused in AndroidOptionsInitializer, instead of creating a new one, if available
Startup profile is bound to the startup transaction:
- Added ITransactionProfiler.isRunning to check if profiler is already running
Added io.sentry.profiling.enable-startup manifest option (forgot to do it in previous PR)
Moved profilingTracesHz from SentryAndroidOptions to SentryOptions (needed to get access in `Sentry.init` to save profiling config to disk)

[Example profile](https://sentry-sdks.sentry.io/profiling/profile/sentry-android/90f8246b4888462c8cfd30b626feba89/flamegraph/?colorCoding=by+system+vs+application+frame&fov=0%2C9%2C3151194112%2C8&query=&sorting=call+order&tid=29955&view=top+down): we can see the 98ms of our `SentryInitProvider` 😅 

This is the last PR to enable startup profiling

Unit tests

## :pencil: Checklist

## :crystal_ball: Next steps
Docs


sentry-android-core/src/main/AndroidManifest.xml
```diff
@@ -8,6 +8,12 @@
       android:authorities="${applicationId}.SentryInitProvider"
       android:exported="false"/>
 
+    <provider
```
romtsn, line 11: `m`: I'm wondering if we could consolidate this under `SentryPerformanceProvider`? Usually introducing new content providers will add to init time, so if we can avoid introducing one, that'd be great

sentry/src/main/java/io/sentry/Hub.java
```diff
@@ -700,12 +700,21 @@ public void flush(long timeoutMillis) {
...
+      final @NotNull TransactionOptions transactionOptions,
+      final boolean isStartupTransaction) {
+    return createTransaction(transactionContext, transactionOptions, isStartupTransaction);
   }
 
   private @NotNull ITransaction createTransaction(
       final @NotNull TransactionContext transactionContext,
-      final @NotNull TransactionOptions transactionOptions) {
+      final @NotNull TransactionOptions transactionOptions,
+      final boolean isStartupTransaction) {
```
romtsn, line 717: `m`: can we put this boolean inside `TransactionOptions`? Ideally we'd keep the number of `startTransaction` overloads at a minimum whenever possible

sentry/src/main/java/io/sentry/Sentry.java
```diff
@@ -254,47 +258,56 @@ private static synchronized void init(
...
               }
-            } catch (IOException e) {
-              options
-                  .getLogger()
-                  .log(SentryLevel.ERROR, "Unable to create startup profiling config file. ", e);
             }
-          }
-        });
+          });
+    } catch (RejectedExecutionException e) {
```
romtsn, line 298:
```suggestion
    } catch (Throwable e) {
```
let's make it bulletproof to everything?

sentry/src/main/java/io/sentry/Sentry.java
```diff
@@ -254,47 +258,56 @@ private static synchronized void init(
...
+                        "Tracing is disabled and startup profiling will not start.");
+                return;
+              }
+              try {
+                if (startupProfilingConfigFile.createNewFile()) {
+                  final @NotNull TracesSamplingDecision startupSamplingDecision =
+                      sampleStartupProfiling(options);
+                  final @NotNull SentryStartupProfilingOptions startupProfilingOptions =
+                      new SentryStartupProfilingOptions(options, startupSamplingDecision);
+                  try (Writer fileWriter = new SentryFileWriter(startupProfilingConfigFile)) {
```
romtsn, line 287:
```suggestion
                  try (FileOutputStream stream = new FileOutputStream(startupProfilingConfigFile)) {
```

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
```diff
@@ -307,6 +312,10 @@ public void close() {
           true,
           null,
           HubAdapter.getInstance().getOptions());
+    } else {
```
romtsn, line 315:
```suggestion
    } else if (transactionsCounter != 0) {
```
should we check for it here?

sentry/src/main/java/io/sentry/Hub.java
stefanosiano: We can, the only thing is that `TransactionOption` is public and the we'd have to add the get method to `ITransaction`.
If we are fine with that, it's totally doable

sentry-android-core/src/main/AndroidManifest.xml
stefanosiano: sure, we can do it

sentry-android-core/src/main/java/io/sentry/android/core/SentryStartupProfilingProvider.java
```diff
@@ -0,0 +1,147 @@
...
+
+    final @NotNull File cacheDir = AndroidOptionsInitializer.getCacheDir(context);
+    final @NotNull File configFile = new File(cacheDir, STARTUP_PROFILING_CONFIG_FILE_NAME);
+
+    // No config exists: startup profiling is not enabled
+    if (!configFile.exists() || !configFile.canRead()) {
+      return false;
+    }
+
+    try (final @NotNull Reader fileReader = new SentryFileReader(configFile)) {
```
romtsn, line 69:
```suggestion
    try (final @NotNull FileInputStream stream = new FileInputStream(configFile)) {
```

sentry-android-core/src/main/java/io/sentry/android/core/SentryStartupProfilingProvider.java
```diff
@@ -0,0 +1,147 @@
...
+              profilingOptions.isProfilingEnabled(),
+              profilingOptions.getProfilingTracesHz(),
+              new SentryExecutorService());
+      AppStartMetrics.getInstance().setStartupProfiler(startupProfiler);
+      startupProfiler.start();
+
+    } catch (FileNotFoundException e) {
+      logger.log(SentryLevel.ERROR, "Startup profiling config file not found. ", e);
+      return false;
+    } catch (IOException e) {
```
romtsn, line 117:
```suggestion
    } catch (Throwable e) {
```
to be on the safe side

sentry-android-core/src/main/java/io/sentry/android/core/SentryStartupProfilingProvider.java
```diff
@@ -0,0 +1,147 @@
...
+          new TracesSamplingDecision(
+              profilingOptions.isTraceSampled(),
+              profilingOptions.getTraceSampleRate(),
+              profilingOptions.isProfileSampled(),
+              profilingOptions.getProfileSampleRate());
+      // We store any sampling decision, so we can respect it when the first transaction starts
+      AppStartMetrics.getInstance().setStartupSamplingDecision(startupSamplingDecision);
+
+      if (!(startupSamplingDecision.getProfileSampled() && startupSamplingDecision.getSampled())) {
+        return false;
```
romtsn, line 96: nit: maybe add a log here as well?

sentry-android-core/src/main/java/io/sentry/android/core/SentryStartupProfilingProvider.java
```diff
@@ -0,0 +1,147 @@
...
+          new AndroidTransactionProfiler(
+              context.getApplicationContext(),
+              buildInfoProvider,
+              new SentryFrameMetricsCollector(
+                  context.getApplicationContext(), logger, buildInfoProvider),
+              logger,
+              profilingOptions.getProfilingTracesDirPath(),
+              profilingOptions.isProfilingEnabled(),
+              profilingOptions.getProfilingTracesHz(),
+              new SentryExecutorService());
```
romtsn, line 110: so this means we'll have 2 SentryExecutorServices now when startup profiling is enabled, right? I guess this should be fine, we'll have 2 threads with the same name, but that shouldn't be a problem https://stackoverflow.com/a/4978149

sentry-android-core/src/main/java/io/sentry/android/core/SentryStartupProfilingProvider.java
stefanosiano: i don't know, we don't log when a transaction or profile is not sampled in general, but a debug log shouldn't be annoying 👍

sentry/src/main/java/io/sentry/Hub.java
romtsn: why would we need a new method in ITransaction?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -145,10 +148,23 @@ static void initializeIntegrationsAndProcessors(
...
-        new SentryFrameMetricsCollector(context, options, buildInfoProvider);
-    options.setTransactionProfiler(
-        new AndroidTransactionProfiler(context, options, buildInfoProvider, frameMetricsCollector));
+
+    // Check if the profiler was already instantiated in the startup.
+    // We use the Android profiler, that uses a global start/stop api, so we need to preserve the
+    // state of the profiler, and it's only possible retaining the instance.
+    final @Nullable ITransactionProfiler startupProfiler =
+        AppStartMetrics.getInstance().getStartupProfiler();
+    if (startupProfiler != null) {
```
romtsn, line 157: dunno if we need synchronization here, in theory `SentryAndroid.init` might be called from a background thread, but the `shutdown` from the contentprovider might be called in the meantime

romtsn (approved):
some comments, but overall looks great, amazing stuff! Let's slim down that SentryInitProvider time

sentry-android-core/src/main/java/io/sentry/android/core/SentryStartupProfilingProvider.java
stefanosiano: yeah, we'd have a new thread for the profiler only

sentry/src/main/java/io/sentry/Hub.java
stefanosiano: you're right, we don't. It's just late 😅

stefanosiano:
@romtsn Want to have a last look before merging? i should have updated it with all your feedback

sentry/src/main/java/io/sentry/Sentry.java
romtsn: @stefanosiano this is still a `RejectedExeuctionException` ;)

sentry-android-core/src/main/java/io/sentry/android/core/SentryPerformanceProvider.java
```diff
@@ -32,9 +51,26 @@ public final class SentryPerformanceProvider extends EmptySecureContentProvider
...
+    logger = new AndroidLogger();
+    buildInfoProvider = new BuildInfoProvider(logger);
+  }
+
   @Override
   public boolean onCreate() {
-    onAppLaunched(getContext());
+    final @NotNull AppStartMetrics appStartMetrics = AppStartMetrics.getInstance();
+    launchStartupProfiler(appStartMetrics);
+    onAppLaunched(getContext(), appStartMetrics);
```
romtsn, line 73: `h`: I think we should probably swap this with `launchStartupProfiler`, to make it how it was before merging the two content providers. `onAppLaunched` reports a high-level metric which will probably be surfaced to the user somewhere in the beginning of the Starfish workflow, and if we don't include what it takes to launch the profiler we're gonna be lying (especially because `launchStartupProfiler` does file i/o and json deser) about the app startup metric.

You probably wanted to include `onAppLaunched` into the profile as well, but there really shouldn't be anything that takes a lot of time.

stefanosiano: > You probably wanted to include onAppLaunched into the profile as well, but there really shouldn't be anything that takes a lot of time.
Yeah, that was the reason, and it was it was working with 2 different providers. But it's fine to swap them, as the profiler launch should be the slowest of the two

romtsn:
> @romtsn Want to have a last look before merging? i should have updated it with all your feedback

LGTM now, thanks for addressing!

PR #3146, opened 2024-01-17
Renamed startupProfiling to appStartProfiling

stefanosiano:
## :scroll: Description
Manifest option: 
io.sentry.traces.profiling.enable-startup -> io.sentry.traces.profiling.enable-app-start

Sentry option:
enableStartupProfiling -> enableAppStartProfiling

AppStartMetrics fields:
startupSamplingDecision -> appStartSamplingDecision
startupProfiler -> appStartProfiler

TransactionContext field:
isForNextStartup -> isForNextAppStart

TransactionOptions field:
isStartupTransaction -> isAppStartTransaction

Class
SentryStartupProfilingOptions -> SentryAppStartProfilingOptions

Config file name:
startup_profiling_config -> app_start_profiling_config

#skip-changelog

Let's align with appStartTracing naming. Also, cocoa is going to do the same.
We didn't release a version, so we don't need to support the old name.
This is mostly to make this feature more clear to the user 

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/SentryPerformanceProvider.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/SentryPerformanceProvider.java
@@ -1,6 +1,6 @@
 package io.sentry.android.core;
 
-import static io.sentry.Sentry.App_START_PROFILING_CONFIG_FILE_NAME;
+import static io.sentry.Sentry.APP_START_PROFILING_CONFIG_FILE_NAME;
 
 import android.annotation.SuppressLint;
 import android.app.Activity;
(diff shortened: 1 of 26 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/SentryPerformanceProvider.java
```diff
@@ -1,6 +1,6 @@
 package io.sentry.android.core;
 
-import static io.sentry.Sentry.STARTUP_PROFILING_CONFIG_FILE_NAME;
+import static io.sentry.Sentry.App_START_PROFILING_CONFIG_FILE_NAME;
```
romtsn, line 3:
```suggestion
import static io.sentry.Sentry.APP_START_PROFILING_CONFIG_FILE_NAME;
```
I guess find+replace got you here :P

romtsn (approved):

sentry-android-core/src/main/java/io/sentry/android/core/SentryPerformanceProvider.java
stefanosiano: yeah, i started using case sensitive replace after a while, too, but it was too late 😅

romtsn:
@stefanosiano you'd also have to change the const name in other places ;)

PR #3151, opened 2024-01-19
Ignore TestOrchestrator failures

romtsn:
_#skip-changelog_

Since we can't do much when test orchestrator itself fails, I think it's fair to ignore those failures.

Ignore TestOrchestrator failures by checking the logs for presence of "process crashed" entry. If there's a failing test, it will actually print a proper stacktrace (like [here](https://github.com/getsentry/sentry-java/actions/runs/7624407511/job/20766479920)), so we're narrowing this down only to test orchestration issues.

### DIFF
--- a/.github/workflows/integration-tests-ui.yml
+++ b/.github/workflows/integration-tests-ui.yml
@@ -57,7 +57,7 @@ jobs:
 
       - name: Verify Test Results
         run: |
-          processCrashed=$(cat test_logs.txt | grep "Process crashed." | wc -l)
+          processCrashed=$(cat test_logs.txt | grep "Instrumentation run failed due to 'Process crashed.'" | wc -l)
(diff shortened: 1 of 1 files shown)


markushi (approved):
LGTM!

stefanosiano:
i just fear we may hide something we don't want to hide (what if a test fails and in the logs there's "Process Crashed"?).
We may try and check how it goes

romtsn:
> i just fear we may hide something we don't want to hide (what if a test fails and in the logs there's "Process Crashed"?). We may try and check how it goes

ok, I've narrowed it down even more by grepping over the whole message. The only potential problem I see if there's one test suite fails due to test orchestrator and another one due to a different reason, then this check would succeed. But I think it's very unlikely. Let's merge it and see how it goes, I agree!

PR #3155, opened 2024-01-22
Collect slow and frozen frames as measurements

markushi:
## :scroll: Description

Similar to https://github.com/getsentry/sentry-java/blob/a45911fbaca266a0af3eee9a66bbcfecbfe0b40c/sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesTracker.java#L161-L172 we want to report slow and frozen frame count for transactions, just like we had before.

#skip-changelog

## :pencil: Checklist


romtsn (approved):
So spans do not support explicit measurements (not from the data bag) yet? It's gonna be available with only-spans, right?

markushi:
> So spans do not support explicit measurements (not from the data bag) yet? It's gonna be available with only-spans, right?

It looks like it's actually possible, but as far as I can see no SDK has it implemented yet => https://github.com/getsentry/team-mobile/issues/172

PR #3158, opened 2024-01-23
Interpolate total frame count to match span duration

markushi:
## :scroll: Description

Interpolate the total_frame count in order to match the span duration.

If there are no content changes on Android, also no frames are rendered, thus no frame metrics are provided by the system.
In order to match the span duration with the total frame count, we simply interpolate the total number of frames based on the span duration, based on the last known refresh rate.

This way the data is more sound and we also match the output of the cocoa SDK

## :pencil: Checklist


romtsn (approved):
nice!

philipphofmann (commented):
Fancy, hack 😃

Issue #3152, opened 2024-01-19
[Starfish] Measure process app-start timing


markushi:
### Description

We want to measure the process-init duration to give a better picture about app-start performance. 
Ideally we report this as a sub-span of the `app.start.cold` span, describing how long the Android OS takes to load the app until the point where the first application class is loaded by the classloader.

markushi:
Did some feasibility check: Ideally we'd use `AppComponentFactory` as it gets initialized before any `Application` or `ContentProvider` class, but this requires Android P (see https://github.com/getsentry/sentry-java/issues/3152 for more details).

Pre-P we could have a static class field within our `AppStartMetrics`, which will get automatically set whenever the first `ContentProvider.onCreate()` is called.

So the span timespan would last from `SystemClock.uptimeMillis()` to ~`AppStartMetrics.getInitUptimeMs()`.

PR #3159, opened 2024-01-23
Report process init time for app start

markushi:
## :scroll: Description

Example event: https://sentry-sdks.sentry.io/performance/sentry-android:5173ff17f6644be5b5fe016352349679
op: `process.load`, description: `"Process Initialization"`

image

To better understand the app start, we want to surface the "system" potion of app start as well. It's mainly unactionable, but showcases how long it takes the system to start the app-process, load some resources and finally execute app specific code. 

Fixes: https://github.com/getsentry/sentry-java/issues/3152

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/PerformanceAndroidEventProcessor.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/PerformanceAndroidEventProcessor.java
@@ -33,6 +33,7 @@ final class PerformanceAndroidEventProcessor implements EventProcessor {
   private static final String APP_METRICS_CONTENT_PROVIDER_OP = "contentprovider.load";
   private static final String APP_METRICS_ACTIVITIES_OP = "activity.load";
   private static final String APP_METRICS_APPLICATION_OP = "application.load";
+  private static final String APP_METRICS_PROCESS_INIT_OP = "process.init";
 
   private boolean sentStartMeasurement = false;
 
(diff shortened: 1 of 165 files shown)


markushi:
@narsaynorath that's how this system span on Android would look like. Please let me know if the op/description would be fine for you.

narsaynorath:
Perfect, the system span looks good to me!

afaik in the last design conversation we had around the app start UI, we're going to drop the system v. application widget, but this would likely be a span that we would surface in the span op table for users to see the change there

sentry-android-core/src/main/java/io/sentry/android/core/PerformanceAndroidEventProcessor.java
```diff
@@ -155,6 +156,23 @@ private void attachColdAppStartSpans(
...
+    // Process init
+    final long classInitUptimeMs = appStartMetrics.getClassLoadedUptimeMs();
+    final @NotNull TimeSpan appStartTimeSpan = appStartMetrics.getAppStartTimeSpan();
+    if (appStartTimeSpan.hasStarted()) {
+      final @NotNull TimeSpan processInitTimeSpan = new TimeSpan();
+      processInitTimeSpan.setStartedAt(appStartTimeSpan.getStartUptimeMs());
+      processInitTimeSpan.setStartUnixTimeMs(appStartTimeSpan.getStartTimestampMs());
+
+      processInitTimeSpan.setStoppedAt(classInitUptimeMs);
+      processInitTimeSpan.setDescription("Process Initialization");
```
stefanosiano, line 168: Should we add a guard in cases the app process is created a lot of time before the app is started (e.g. a notification or a service)?

markushi: That's a good point! iOS already discards app starts longer than a certain period - let me have a look into this!

stefanosiano: @markushi did you have the chance to look into this?

markushi: 👍 I've added an extra check here: https://github.com/getsentry/sentry-java/pull/3159/commits/4e16c88756164c7d71fce7cb07ec514561be7ed0

sentry-android-core/src/main/java/io/sentry/android/core/PerformanceAndroidEventProcessor.java
```diff
@@ -155,6 +157,25 @@ private void attachColdAppStartSpans(
       }
     }
 
+    // Process init
+    final long classInitUptimeMs = appStartMetrics.getClassLoadedUptimeMs();
+    final @NotNull TimeSpan appStartTimeSpan = appStartMetrics.getAppStartTimeSpan();
+    if (appStartTimeSpan.hasStarted()
+        && Math.abs(classInitUptimeMs - appStartTimeSpan.getStartUptimeMs())
+            <= MAX_PROCESS_INIT_APP_START_DIFF_MS) {
```
stefanosiano, line 165: looks good now, just add a comment for your future self 😄

stefanosiano (approved):

Issue #2541, opened 2023-02-14
Initialize SentryOptions members lazily


romtsn:
### Description

When [SentryOptions.empty()](https://github.com/getsentry/sentry-java/blob/4b32504a16c1ffd2325208184a90b34a8252fb02/sentry/src/main/java/io/sentry/SentryOptions.java#L2047-L2049) is used (e.g. in a `NoOpHub`), we still eagerly initialize some of the class members, for example:
* [JsonSerializer](https://github.com/getsentry/sentry-java/blob/4b32504a16c1ffd2325208184a90b34a8252fb02/sentry/src/main/java/io/sentry/SentryOptions.java#L105)
* [ClientReportRecorder](https://github.com/getsentry/sentry-java/blob/4b32504a16c1ffd2325208184a90b34a8252fb02/sentry/src/main/java/io/sentry/SentryOptions.java#L369)
* [EnvelopeReader](https://github.com/getsentry/sentry-java/blob/4b32504a16c1ffd2325208184a90b34a8252fb02/sentry/src/main/java/io/sentry/SentryOptions.java#L102)

I think, those should be fine to initialize lazily, so in case of the empty options they do not cause potential issues. There was a recent customer report of an ANR caused by with Sentry frames in it (although the ANR is most likely not caused by Sentry, because there's nothing involved that should cause an ANR, but still we'd like to eliminate such issues):
```
main (runnable):tid=1 systid=27171 
       at io.sentry.JsonSerializer.<init>(JsonSerializer.java:84)
       at io.sentry.SentryOptions.<init>(SentryOptions.java:99)
       at io.sentry.SentryOptions.empty(SentryOptions.java:1899)
       at io.sentry.NoOpHub.<init>(NoOpHub.java:14)
       at io.sentry.NoOpHub.<clinit>(NoOpHub.java:12)
       at io.sentry.Sentry.<clinit>(Sentry.java:29)
       at io.sentry.HubAdapter.getSpan(HubAdapter.java:246)
       at io.sentry.instrumentation.file.FileIOSpanManager.startSpan(FileIOSpanManager.java:26)
       at io.sentry.instrumentation.file.SentryFileInputStream.init(SentryFileInputStream.java:65)
       at io.sentry.instrumentation.file.SentryFileInputStream.access$000(SentryFileInputStream.java:22)
       at io.sentry.instrumentation.file.SentryFileInputStream$Factory.create(SentryFileInputStream.java:137)
[...]

PR #3166, opened 2024-01-25
Integrate Spotlight

markushi:
## :scroll: Description

Add `beforeSendEnvelope` callback similar to JS and utilize it to send envelopes to spotlight. 🔥 
image

Fixes https://github.com/getsentry/sentry-java/issues/3097

Better DX

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
@@ -212,7 +212,12 @@ final class AndroidOptionsInitializer {
           new AndroidCpuCollector(options.getLogger(), buildInfoProvider));
 
       if (options.isEnablePerformanceV2()) {
-        options.addPerformanceCollector(new SpanFrameMetricsCollector(options));
+        options.addPerformanceCollector(
+            new SpanFrameMetricsCollector(
+                options,
+                Objects.requireNonNull(
+                    options.getFrameMetricsCollector(),
+                    "options.getFrameMetricsCollector is required")));
       }
     }
     options.setTransactionPerformanceCollector(new DefaultTransactionPerformanceCollector(options));
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
@@ -27,6 +27,7 @@ import io.sentry.cache.PersistingOptionsObserver;
 import io.sentry.cache.PersistingScopeObserver;
 import io.sentry.compose.gestures.ComposeGestureTargetLocator;
 import io.sentry.compose.viewhierarchy.ComposeViewHierarchyExporter;
+import io.sentry.internal.SpotlightIntegration;
 import io.sentry.internal.gestures.GestureTargetLocator;
 import io.sentry.internal.viewhierarchy.ViewHierarchyExporter;
 import io.sentry.transport.NoOpEnvelopeCache;
@@ -321,6 +322,12 @@ final class AndroidOptionsInitializer {
       } catch (RuntimeException e) {
         options.getLogger().log(SentryLevel.ERROR, "Could not generate distinct Id.", e);
       }
+    }
+
+    if (options.getSpotlightConnectionUrl() == null) {
+      // developer machine should be the same across emulators
+      // see https://developer.android.com/studio/run/emulator-networking.html
+      options.setSpotlightConnectionUrl("http://10.0.2.2:8969/stream");
     }
   }
 
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
@@ -332,12 +332,6 @@ final class AndroidOptionsInitializer {
         options.getLogger().log(SentryLevel.ERROR, "Could not generate distinct Id.", e);
       }
     }
-
-    if (options.getSpotlightConnectionUrl() == null) {
-      // developer machine should be the same across emulators
-      // see https://developer.android.com/studio/run/emulator-networking.html
-      options.setSpotlightConnectionUrl("http://10.0.2.2:8969/stream");
-    }
   }
 
   /**
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
@@ -103,6 +103,9 @@ final class AndroidOptionsInitializer {
     // set a lower flush timeout on Android to avoid ANRs
     options.setFlushTimeoutMillis(DEFAULT_FLUSH_TIMEOUT_MS);
 
+    options.setFrameMetricsCollector(
+        new SentryFrameMetricsCollector(context, logger, buildInfoProvider));
+
     ManifestMetadataReader.applyMetadata(context, options, buildInfoProvider);
     options.setCacheDirPath(getCacheDir(context).getAbsolutePath());
 
@@ -159,14 +162,16 @@ final class AndroidOptionsInitializer {
         options.setTransactionProfiler(appStartProfiler);
         AppStartMetrics.getInstance().setAppStartProfiler(null);
       } else {
-        final SentryFrameMetricsCollector frameMetricsCollector =
-            new SentryFrameMetricsCollector(context, options, buildInfoProvider);
         options.setTransactionProfiler(
             new AndroidTransactionProfiler(
-                context, options, buildInfoProvider, frameMetricsCollector));
+                context,
+                options,
+                buildInfoProvider,
+                Objects.requireNonNull(
+                    options.getFrameMetricsCollector(),
+                    "options.getFrameMetricsCollector is required")));
       }
     }
-
     options.setModulesLoader(new AssetsModulesLoader(context, options.getLogger()));
     options.setDebugMetaLoader(new AssetsDebugMetaLoader(context, options.getLogger()));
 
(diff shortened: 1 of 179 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/SpotlightIntegration.java
```diff
@@ -0,0 +1,122 @@
...
+import java.net.HttpURLConnection;
+import java.net.URI;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.RejectedExecutionException;
+import java.util.zip.GZIPOutputStream;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class SpotlightIntegration
```
adinauer, line 23: Could we also build this in a way that works for backend, desktop, etc. as well?

markushi: :+1: I just moved it over to sentry core.

I couldn't get the spring-boot example running though, as it somehow didn't pick up my config. Maybe you can have a quick look? 

It's easiest to get started by using the pre-built spotlight electron app (https://spotlightjs.com/) +
setting `options.setSpotlightConnectionUrl("http://localhost:8969/stream");`

adinauer: Will give it a try.

sentry/src/main/java/io/sentry/internal/SpotlightIntegration.java
```diff
@@ -0,0 +1,119 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+@ApiStatus.Internal
+public final class SpotlightIntegration
+    implements Integration, SentryOptions.BeforeEnvelopeCallback, Closeable {
+
+  private @NotNull SentryOptions options = SentryOptions.empty();
+
+  private final @NotNull ExecutorService executorService = Executors.newSingleThreadExecutor();
```
adinauer, line 31: `h` can we use executor service from options? This one doesn't set `daemon` and thus will likely block shutdown for backend (https://github.com/getsentry/sentry-java/pull/2747)

adinauer (commented):
I'm able to run the `SpotlightIntegration` for Spring Boot 3 by moving it to `SentryOptions` as suggested in one of the comments but no errors / transactions / ... show up in spotlight. It only shows the SDK on the SDKs tab, nothing else.

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -442,6 +442,10 @@ public class SentryOptions {
   /** Whether to send modules containing information about versions. */
   private boolean sendModules = true;
 
+  private @Nullable BeforeEnvelopeCallback beforeEnvelopeCallback;
+
+  private @Nullable String spotlightConnectionUrl;
```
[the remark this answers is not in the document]
stefanosiano: I agree on the `setEnableSpotlight()` option.
That would just put `10.0.2.2` on Android and `localhost` on Java. Don't care about real devices for now, as that would require the device to be in the same network, and the host to set its local ip in the option. Better to just point it out in the option javadoc (and in the docs) directly

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -321,6 +323,12 @@ private static void readDefaultOptionValues(
         options.getLogger().log(SentryLevel.ERROR, "Could not generate distinct Id.", e);
       }
     }
+
+    if (options.getSpotlightConnectionUrl() == null) {
+      // developer machine should be the same across emulators
+      // see https://developer.android.com/studio/run/emulator-networking.html
+      options.setSpotlightConnectionUrl("http://10.0.2.2:8969/stream");
```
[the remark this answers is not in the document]
stefanosiano: we have to init it differently on Android (10.0.2.2) and Java (192.168.1.1)
Also, having a null url should disable it (guess this is the summary of your comment, right?)
Unless we add the `enableSpotlight` option, too

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/MainActivity.java
```diff
@@ -58,11 +58,12 @@ protected void onCreate(Bundle savedInstanceState) {
       Sentry.captureException(e);
     }
 
-    final Attachment image = new Attachment(imageFile.getAbsolutePath(), "sentry.png", "image/png");
-    Sentry.configureScope(
-        scope -> {
-          scope.addAttachment(image);
-        });
+    //    final Attachment image = new Attachment(imageFile.getAbsolutePath(), "sentry.png",
```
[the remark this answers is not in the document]
stefanosiano: I remember a problem with images (and screenshots) in spotlight, but that was going to be fixed anyway
Hopefully it gets fixed before releasing 😅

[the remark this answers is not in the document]
markushi: Yes, there's an open issue for image attachements: https://github.com/getsentry/spotlight/issues/274
And I opened up a PR for fixing sentry-java related issues as well: https://github.com/getsentry/spotlight/pull/357

sentry/src/main/java/io/sentry/internal/SpotlightIntegration.java
markushi: I've switched to using our `SentryExecutorService`, which is setup in daemon mode. I still wouldn't use the executor service from our options, as it could slow down any disk/cpu bound tasks when many envelopes are queuing up.

sentry-samples/sentry-samples-android/src/main/AndroidManifest.xml
```diff
@@ -143,7 +144,7 @@
         <!--    <meta-data android:name="io.sentry.breadcrumbs.app-components" android:value="false" />-->
 
         <!--    how to enable the attach screenshot feature-->
-        <meta-data android:name="io.sentry.attach-screenshot" android:value="true" />
+        <meta-data android:name="io.sentry.attach-screenshot" android:value="false" />
```
romtsn, line 147: should we enable it back?

```suggestion
        <meta-data android:name="io.sentry.attach-screenshot" android:value="true" />
```

sentry/src/main/java/io/sentry/SpotlightIntegration.java
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  @Override
+  @SuppressWarnings("FutureReturnValueIgnored")
+  public @NotNull SentryEnvelope execute(
+      final @NotNull SentryEnvelope envelope, final @Nullable Hint hint) {
+    try {
+      executorService.submit(() -> sendEnvelope(envelope));
+    } catch (RejectedExecutionException ex) {
+      // ignored
```
romtsn, line 51: `l`: maybe add a warning log?

sentry/src/main/java/io/sentry/SpotlightIntegration.java
```diff
@@ -0,0 +1,140 @@
...
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.jetbrains.annotations.TestOnly;
+
+@ApiStatus.Internal
+public final class SpotlightIntegration
+    implements Integration, SentryOptions.BeforeEnvelopeCallback, Closeable {
+
+  private @NotNull SentryOptions options = SentryOptions.empty();
```
romtsn, line 23: `m`: shall we make it nullable rather, to not hit https://github.com/getsentry/sentry-java/issues/2541?

sentry/src/main/java/io/sentry/SentryClient.java
```diff
@@ -582,14 +582,29 @@ public void captureSession(final @NotNull Session session, final @Nullable Hint
...
+    if (envelopeToSend != null) {
+      if (hint == null) {
+        transport.send(envelope);
+      } else {
+        transport.send(envelope, hint);
+      }
+      final @Nullable SentryId id = envelope.getHeader().getEventId();
+      return id != null ? id : SentryId.EMPTY_ID;
     } else {
       return SentryId.EMPTY_ID;
```
romtsn, line 609: in theory we'd have to record a client report on the lost envelope here with `DiscardReason.BEFORE_SEND`, but dunno if other platforms cared about that

markushi: Good point, I guess it's fine for now - as beforeEnvelope is internal only. We could also change beforeEnvelope to not allow any filtering. At least e.g. RN does not allow that: https://github.com/getsentry/sentry-react-native/blob/a5f6b2f078f0434fcd46545b6b4f789d58b30769/src/js/client.ts#L138-L149

romtsn: yeah I'm good with whatever other sdks do in this case 👍

markushi: I've switched to a `void` return type for now. less code, less complexity 😅

sentry/src/main/java/io/sentry/internal/SpotlightIntegration.java
markushi: @adinauer would the mentioned `SentryExecutorService` work fine for java backend use cases?

adinauer: Do you mean the one from options or a new one? Either should work. Is there a specific reason you're asking?
Can't we increase the number of threads `SentryExecutorService ` is allowed to use in case we run into problems?

There's `Executors.newScheduledThreadPool` which we could pass a pool size to.

markushi: Afaik we can't increase the pool size for `options.executorService`. At least not on Android, right @romtsn? IIRC on SDK init we rely on some background tasks to be executed in a certain order.

PR #3167, opened 2024-01-25
Apply OkHttp listener auto finish timestamp to all running spans

stefanosiano:
… to http root call span

## :scroll: Description
scheduled finish timestamp now applies to all running spans, not only to http root call span

When the response is not read, we finish the OkHttp span automatically.
We were finishing the root call span at the last http span timestamp, but we were not applying it to all other OkHttp running spans.
[An image is worth more than a thousand words](https://demo.sentry.io/performance/android:c2475d976c3942c296351f558a7dc936/?limit=100&transaction=EmpowerPlantActivity)

Unit test

## :pencil: Checklist

### DIFF
--- a/sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpEvent.kt
+++ b/sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpEvent.kt
@@ -7,7 +7,6 @@ import io.sentry.ISpan
 import io.sentry.SentryDate
 import io.sentry.SentryLevel
 import io.sentry.SpanDataConvention
-import io.sentry.SpanStatus
 import io.sentry.TypeCheckHint
 import io.sentry.okhttp.SentryOkHttpEventListener.Companion.CONNECTION_EVENT
 import io.sentry.okhttp.SentryOkHttpEventListener.Companion.CONNECT_EVENT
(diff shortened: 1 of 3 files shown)


sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpEvent.kt
```diff
@@ -158,10 +157,12 @@ internal class SentryOkHttpEvent(private val hub: IHub, private val request: Req
 
         // We forcefully finish all spans, even if they should already have been finished through finishSpan()
         eventSpans.values.filter { !it.isFinished }.forEach {
-            // If a status was set on the span, we use that, otherwise we set its status as error.
-            it.status = it.status ?: SpanStatus.INTERNAL_ERROR
             moveThrowableToRootSpan(it)
-            it.finish()
+            if (finishDate != null) {
+                it.finish(it.status, finishDate)
```
markushi, line 162: Shouldn't this be
```suggestion
                it.finish(it.status ?: SpanStatus.INTERNAL_ERROR, finishDate)
```

markushi (changes requested):
LGTM! I have one question I'd like to get answered before merging it. 😅

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpEvent.kt
stefanosiano: it was, but it actually didn't make sense, at least in the case linked in this PR.
If the response is not read, there will be some running spans, and they don't have any error.
So they didn't have to be reported as errored in the first place, i guess

stefanosiano: If there is another possible use case, we can think more about it

markushi: 👍 makes a lot of sense this way then!

markushi (approved):

PR #3175, opened 2024-01-29
Avoid multiple breadcrumbs from OkHttpEventListener

stefanosiano:
## :scroll: Description
increased SentryOkHttpEvent.scheduleFinish timeout 500ms -> 800ms
added a guard to avoid finishing an already finished event

If an OkHttp event was finished multiple times, multiple breadcrumbs would been sent.
This was possible when the response of a call was closed (or read) after more than 500ms since it was received, as our timeout automatically closes the call after 500ms.
I increased the timeout to 800ms to make it less likely to happen. And I added a flag to skip closing the event if it was already closed.
Likely relates to https://github.com/getsentry/sentry-android-gradle-plugin/issues/617

Unit tests

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 
 ### Fixes
 
+- Avoid multiple breadcrumbs from OkHttpEventListener ([#3175](https://github.com/getsentry/sentry-java/pull/3175))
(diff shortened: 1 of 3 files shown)


markushi (approved):
LGTM!

Issue #2444, opened 2022-12-28
SDKs should not send tags by default but rather index them in snuba if it needs to be searchable


markushi:
### Description

We should avoid setting tags when possible.

See meta issue for more details: https://github.com/getsentry/team-mobile/issues/60

I just checked Java/Android usage:

* Any tags set via SentryOptions.addContextTag() - but since they're custom values and empty by default I guess we can ignore those
* setTag("TimberTag", ...) -> [link](https://github.com/getsentry/sentry-java/blob/3f8d7b1cc869bb056c9db99b459e43f6c375784a/sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt#L266)
* setSideLoadedInfo(...) -> [link](https://github.com/getsentry/sentry-java/blob/81a3c3273f211de79706ad83e24929ef1ec58f0a/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java#L896)

kahest:
Need to talk to SnS to find out if using Tags makes sense in this case, and to make sure functionality is not lost if moved to Context.

marandaneto:
Per https://github.com/getsentry/team-mobile/issues/60#issuecomment-1468487797
This issue can also be closed unless there are something new after my discussion with S&S

smeubank:
closing issue, per Manoel's comment above. Yes, SDKs should not set Tags, users set tags and tags can be set on ingestion. If there are tags being set by the SDK, it should be resolved.

PR #3193, opened 2024-02-06
Add Split Apks info as extras

### LINKED ISSUE #3192: Add Split Apks information as tag
### Problem Statement

Now, Google Play forced to use App Bundle, and Android apps are installed as `Split Apks` from Google Play Store.
And Crashes of Android Apps may occur depending on whether parts of Split Apks are installed such as
* DPI Split Apks (ex: config.xhdpi)
  * If DPI Split apk are not installed, app can be crashed with `android.view.InflateException` by missing drawables.
* ABI Split Apks (ex: config.arm64-v8a)
  * If ABI Split apk are not installed, app can be crashed with `java.lang.UnsatisfiedLinkError` by missing so files.

These crashes occurs because all `split apks` are not installed from Google Play or the manufacturer's store, and are issues in the distribution process.
And if Sentry Android SDK provides these information, project owner can identify these issues are from distribution process or app's logical problem.

### Solution Brainstorm

* Whether installed apk is split apk
  * Need to check `com.android.vending.splits.required` from metadata of package info
  * https://github.com/google/bundletool/blob/master/src/main/java/com/android/tools/build/bundletool/model/AndroidManifest.java#L263C63-L263C98
* Which split apks are installed
  *  check `splitNames` from [PackageInfo](https://developer.android.com/reference/android/content/pm/PackageInfo?hl=en#splitNames)

ganadist:
## :scroll: Description

Fix #3192

## :pencil: Checklist

## :crystal_ball: Next steps
Need to check test and unittest cases.


romtsn:
hi @ganadist thanks for opening an issue for this and the PR! Before proceeding, I'd like to better understand your use-case - would you like to search for events with a specific split name?

In general, we don't set tags by default in the SDKs, but rather extract them on backend, if there's a good case for those and cardinality is not high.

romtsn:
@ganadist thanks, but can you please answer the question: would you like to search for events with a specific split name? If you don't need to search for it but just see it in the issue details, we could just send this info as part of App context then. 

Sending this as tag would require us to store it, and given that this field is high-cardinality (afaik apk splits are unique in their names and if multiply it by number of apks our customers might have that's huge), so this is unlikely that we'll take that route.

`setSideLoadedInfo` is doing it wrong and we have an [issue](https://github.com/getsentry/sentry-java/issues/2444) to fix that, so don't see it as a reference for implementation.

ganadist:
Added 42a229f to use `extras` instead of `tags`.

romtsn:
>  it will be helpful to know the crash rate that occurs due to these problems.

Unfortunately crash rate is based on Sessions dataset and it's not possible to filter session by custom tags/values yet. If search is not crucial for you, we should just add it as part of the [App](https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/protocol/App.java) context - you will be able to see this information on individual events/issues then

romtsn (approved):

romtsn:
hi @ganadist sorry for abandoning this for so long, finally got the time to move the info to the app context. We'll ship this improvement as part of version 8.2.0. Thanks for your contribution!

Google Chrome 2025-02-06 23 34 52

Issue #3183, opened 2024-02-01
Attach frame delay to spans


markushi:
See https://github.com/getsentry/sentry-java/issues/3099

PR #3197, opened 2024-02-07
Calculate frame delay on a span level

markushi:
## :scroll: Description
Fixes https://github.com/getsentry/sentry-java/issues/3183

https://github.com/getsentry/sentry-java/assets/1411808/9b78e1c8-110a-4b32-bc3d-a1fd023f22a6

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
@@ -41,13 +41,17 @@ public class SpanFrameMetricsCollector
 
   private volatile @Nullable String listenerId;
 
-  // a map of <span-id, span-start-time>
+  // a map of <span-id, span-start-time-nanos>
+  // as spans do not contain a nano start time, it's being tracked here
   private final @NotNull Map<String, NanoTimeStamp> spanStarts;
 
-  // all running spans, sorted by span start time
+  // all running spans, sorted by span start nano time
   private final @NotNull SortedSet<SpanStart> runningSpans;
 
   // all collected frames, sorted by frame end time
+  // this is a concurrent set, as the frames are added on the main thread,
+  // but span starts/finish may happen on any thread
+  // the list only holds Frames, but in order to query for a specific span NanoTimeStamp is used
   private final @NotNull ConcurrentSkipListSet<NanoTimeStamp> frames;
 
   // assume 60fps until we get a value reported by the system
@@ -71,7 +75,6 @@ public class SpanFrameMetricsCollector
     enabled = options.isEnablePerformanceV2() && options.isEnableFramesTracking();
 
     frames = new ConcurrentSkipListSet<>();
-
     spanStarts = new HashMap<>();
     runningSpans = new TreeSet<>();
   }
@@ -139,8 +142,6 @@ public class SpanFrameMetricsCollector
       }
       runningSpans.remove(new SpanStart(id, spanStartTimeStamp));
 
-      final long spanStartNanos = spanStartTimeStamp.timeNanos;
-
       // ignore spans with no finish date
       final @Nullable SentryDate spanFinishDate = span.getFinishDate();
       if (spanFinishDate == null) {
@@ -149,23 +150,23 @@ public class SpanFrameMetricsCollector
 
       final @NotNull SentryFrameMetrics frameMetrics = new SentryFrameMetrics();
 
-      // project the span end date into a frame time format
+      // project the span end date into our nano time format
       final @NotNull SentryDate now = options.getDateProvider().now();
       final long nowNanos = frameTimeProvider.now();
 
       final long diffNanos = now.diff(spanFinishDate);
       final long spanEndNanos = nowNanos - diffNanos;
 
+      final long spanStartNanos = spanStartTimeStamp.timeNanos;
       if (spanStartNanos >= spanEndNanos) {
         return;
       }
 
       final long spanDurationNanos = spanEndNanos - spanStartNanos;
       long frameDurationNanos = lastKnownFrameDurationNanos;
-      long lastFrameEndNanos = spanStartNanos;
 
       if (!frames.isEmpty()) {
-        // determine relevant start idx in frames list
+        // determine relevant start in frames list
         final Iterator<NanoTimeStamp> iterator =
             frames.tailSet(new NanoTimeStamp(spanStartNanos)).iterator();
         while (iterator.hasNext()) {
@@ -176,7 +177,7 @@ public class SpanFrameMetricsCollector
           }
 
           if (frame.startNanos >= spanStartNanos && frame.endNanos <= spanEndNanos) {
-            // if the span contains the frame as a whole, add it 1:1 to the span metrics
+            // if the frame is contained within the span, add it 1:1 to the span metrics
             frameMetrics.addFrame(
                 frame.durationNanos, frame.delayNanos, frame.isSlow, frame.isFrozen);
           } else if ((spanStartNanos > frame.startNanos && spanStartNanos < frame.endNanos)
@@ -200,51 +201,31 @@ public class SpanFrameMetricsCollector
           }
 
           frameDurationNanos = frame.expectedDurationNanos;
-          lastFrameEndNanos = frame.endNanos;
         }
       }
 
       int totalFrameCount = frameMetrics.getTotalFrameCount();
 
       final long nextScheduledFrameNanos = frameMetricsCollector.getLastKnownFrameStartTimeNanos();
-      final long pendingDurationNanos = Math.max(0, spanEndNanos - nextScheduledFrameNanos);
-      final boolean isSlow =
-          SentryFrameMetricsCollector.isSlow(pendingDurationNanos, frameDurationNanos);
-      if (isSlow) {
-        // add a single slow/frozen frame
-        final boolean isFrozen = SentryFrameMetricsCollector.isFrozen(pendingDurationNanos);
-        final long pendingDelayNanos = Math.max(0, pendingDurationNanos - frameDurationNanos);
-        frameMetrics.addFrame(pendingDurationNanos, pendingDelayNanos, true, isFrozen);
-        totalFrameCount++;
-      }
-
-      // if there are no content changes on Android, also no new frame metrics are provided by the
-      // system
-      // in order to match the span duration with the total frame count,
-      // we simply interpolate the total number of frames based on the span duration
-      // this way the data is more sound and we also match the output of the cocoa SDK
-      final long frameMetricsDurationNanos = frameMetrics.getTotalDurationNanos();
-      final long nonRenderedDuration = spanDurationNanos - frameMetricsDurationNanos;
-      if (nonRenderedDuration > 0) {
-        final int normalFrames = (int) (nonRenderedDuration / frameDurationNanos);
-        totalFrameCount += normalFrames;
-      }
+      totalFrameCount +=
+          addPendingFrameDelay(
+              frameMetrics, frameDurationNanos, spanEndNanos, nextScheduledFrameNanos);
+      totalFrameCount += interpolateFrameCount(frameMetrics, frameDurationNanos, spanDurationNanos);
 
       final long frameDelayNanos =
           frameMetrics.getSlowFrameDelayNanos() + frameMetrics.getFrozenFrameDelayNanos();
-
-      final double frameDelaySeconds = frameDelayNanos / 1e9d;
+      final double frameDelayInSeconds = frameDelayNanos / 1e9d;
 
       span.setData(SpanDataConvention.FRAMES_TOTAL, totalFrameCount);
       span.setData(SpanDataConvention.FRAMES_SLOW, frameMetrics.getSlowFrameCount());
       span.setData(SpanDataConvention.FRAMES_FROZEN, frameMetrics.getFrozenFrameCount());
-      span.setData(SpanDataConvention.FRAMES_DELAY, frameDelaySeconds);
+      span.setData(SpanDataConvention.FRAMES_DELAY, frameDelayInSeconds);
 
       if (span instanceof ITransaction) {
         span.setMeasurement(MeasurementValue.KEY_FRAMES_TOTAL, totalFrameCount);
         span.setMeasurement(MeasurementValue.KEY_FRAMES_SLOW, frameMetrics.getSlowFrameCount());
         span.setMeasurement(MeasurementValue.KEY_FRAMES_FROZEN, frameMetrics.getFrozenFrameCount());
-        span.setMeasurement(MeasurementValue.KEY_FRAMES_DELAY, frameDelaySeconds);
+        span.setMeasurement(MeasurementValue.KEY_FRAMES_DELAY, frameDelayInSeconds);
       }
     }
   }
@@ -272,7 +253,8 @@ public class SpanFrameMetricsCollector
       boolean isFrozen,
       float refreshRate) {
 
-    // buffer is full, skip adding new frames
+    // buffer is full, skip adding new frames for now
+    // once a span finishes, the buffer will trimmed
     if (frames.size() > MAX_FRAMES_COUNT) {
       return;
     }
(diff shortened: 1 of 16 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
```diff
@@ -1,43 +1,82 @@
...
 public class SpanFrameMetricsCollector
     implements IPerformanceContinuousCollector,
         SentryFrameMetricsCollector.FrameMetricsCollectorListener {
 
-  private @NotNull final Object lock = new Object();
-  private @Nullable final SentryFrameMetricsCollector frameMetricsCollector;
-  private @Nullable volatile String listenerId;
-  private @NotNull final Map<SpanId, SentryFrameMetrics> metricsAtSpanStart;
+  // 30s span duration at 120fps = 3600 frames
+  private static final int MAX_FRAMES_COUNT = 3600;
```
stefanosiano, line 32: Please, don't make any assumption on frame numbers. IOS can, but Android is a different thing. There are devices with 144, 165, and even [240](https://www.gsmarena.com/sharp_aquos_zero_2-10259.php) fps at the moment. And I'm pretty sure it will increase over time 😅 
Also, Android doesn't call the callback on any frame drawn, but only when frames change. 
We can avoid checking frames count imho

sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
```diff
@@ -1,43 +1,82 @@
...
+  @TestOnly
+  public SpanFrameMetricsCollector(
+      final @NotNull SentryAndroidOptions options,
+      final @NotNull FrameTimeProvider frameTimeProvider) {
+    this.options = options;
+    this.frameTimeProvider = frameTimeProvider;
+
+    frameMetricsCollector =
+        Objects.requireNonNull(
+            options.getFrameMetricsCollector(), "FrameMetricsCollector is required");
```
stefanosiano, line 74: since `options.getFrameMetricsCollector()` is nullable, what about adding a `SentryFrameMetricsCollector` in the params of the constructor?
We have access to it anyway in the `AndroidOptionsInitializer`

sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
```diff
@@ -68,63 +110,122 @@ public void onSpanFinished(final @NotNull ISpan span) {
...
-      final @Nullable SentryFrameMetrics metricsAtStart =
-          metricsAtSpanStart.remove(span.getSpanContext().getSpanId());
-      if (metricsAtStart != null) {
-        diff = currentFrameMetrics.diffTo(metricsAtStart);
+      if (runningSpans.isEmpty()) {
+        clear();
+      } else {
+        // otherwise only remove old/irrelevant frames
+        final SpanStart oldestSpan = runningSpans.first();
+        frames.tailSet(oldestSpan.timeNanos).clear();
```
stefanosiano, line 130: here we lose info in case of multiple spans running at the same time
frame delay should be added to all of them, but we're adding only to one in this case

sentry-android-core/src/test/java/io/sentry/android/core/SpanFrameMetricsCollectorTest.kt
```diff
@@ -139,42 +174,50 @@ class SpanFrameMetricsCollectorTest {
...
+        val span1 = createFakeSpan(820, 840)
         sut.onSpanStarted(span1)
 
         // and another slow frame is recorded
-        sut.onFrameMetricCollected(0, 20, 20, 4, true, false, 60.0f)
-
-        sut.onSpanFinished(span0)
+        fixture.timeNanos = 840
+        sut.onFrameMetricCollected(820, 840, 20, 4, true, false, 60.0f)
         sut.onSpanFinished(span1)
```
stefanosiano, line 198: adding `sut.onSpanFinished(createFakeSpan(820, 840))` right before `sut.onSpanFinished(span1)` make it fail, due to concurrent spans not working
create another test specifically for it, after making it work

sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
```diff
@@ -68,63 +110,122 @@ public void onSpanFinished(final @NotNull ISpan span) {
...
       return;
     }
+
     if (span instanceof NoOpSpan) {
       return;
     }
+
     if (span instanceof NoOpTransaction) {
       return;
     }
```
stefanosiano, line 121: guess we should also ignore a span that never started collecting frames?
something like `if (!spanStarts.contains(span.getSpanContext().getSpanId().toString()) { return; }`

stefanosiano (changes requested):
It looks good overall.
I guess we have to revisit the `SpanFrameMetricsCollector`, as it doesn't work correctly for multiple spans at the same time.
Maybe there is also a way to get rid of those intermediate classes (`SpanStart` and `NanoTimeStamp`), too?

sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
markushi: This is just an upper limit to ensure we don't run into OOMs when we have long running spans, as we collect every frame metric callback data.

sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
markushi: Here `runningSpans` will contain all  running spans (except the one which just finished above). Then we get the "oldest" span which is still running and delete all frame data prior to it.

sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
```diff
@@ -1,43 +1,82 @@
 package io.sentry.android.core;
 
+import androidx.annotation.NonNull;
```
romtsn, line 3: let's use `org.jetbrains.annotations.NotNull;` as everywhere else

```suggestion
import org.jetbrains.annotations.NotNull;
```

sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
markushi: Update: Found the issue and added tests, thanks!

markushi:
> It looks good overall. I guess we have to revisit the `SpanFrameMetricsCollector`, as it doesn't work correctly for multiple spans at the same time. Maybe there is also a way to get rid of those intermediate classes (`SpanStart` and `NanoTimeStamp`), too?

I've resolved the issue for multi-spans, mixed up head/tailSet APIs 🙈 
I'd love to get rid of `SpanStart` and `NanoTimeStamp`, I'll have a look and provide a follow up PR if something more elegant would work here.

sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
romtsn: should we use the upper upper limit then? (30 * 240)?

sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
```diff
@@ -7,37 +7,73 @@
...
-  private @Nullable final SentryFrameMetricsCollector frameMetricsCollector;
-  private @Nullable volatile String listenerId;
-  private @NotNull final Map<SpanId, SentryFrameMetrics> metricsAtSpanStart;
+  // 30s span duration at 120fps = 3600 frames
+  private static final int MAX_FRAMES_COUNT = 3600;
+  private static final long ONE_SECOND_NANOS = TimeUnit.SECONDS.toNanos(1);
 
-  private @NotNull final SentryFrameMetrics currentFrameMetrics;
+  private final @NotNull SentryAndroidOptions options;
+  private final @NotNull FrameTimeProvider frameTimeProvider;
```
romtsn, line 35: `l`: I wonder if we could reuse our `SentryNanotimeDateProvider`?

sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
```diff
@@ -133,34 +239,161 @@ public void onSpanFinished(final @NotNull ISpan span) {
...
+    return 0;
+  }
+
+  @ApiStatus.Internal
+  public interface FrameTimeProvider {
+    long now();
+  }
+
+  @ApiStatus.Internal
+  public static class NanoTimeStamp implements Comparable<NanoTimeStamp> {
```
romtsn, line 323: `l`: this also potentially could be replaced by the existing `SentryNanotimeDate`

stefanosiano: I agree, but we'd have to add a method to the `SentryDate` to get the current nano time, as the current implementation returns the date converted to nanoseconds (btw, is that intentional?)

sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
```diff
@@ -68,63 +107,130 @@ public void onSpanFinished(final @NotNull ISpan span) {
...
+
+      final long nextScheduledFrameNanos = frameMetricsCollector.getLastKnownFrameStartTimeNanos();
+      totalFrameCount +=
+          addPendingFrameDelay(
+              frameMetrics, frameDurationNanos, spanEndNanos, nextScheduledFrameNanos);
+      totalFrameCount += interpolateFrameCount(frameMetrics, frameDurationNanos, spanDurationNanos);
+
+      final long frameDelayNanos =
+          frameMetrics.getSlowFrameDelayNanos() + frameMetrics.getFrozenFrameDelayNanos();
+      final double frameDelayInSeconds = frameDelayNanos / 1e9d;
```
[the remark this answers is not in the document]
romtsn: ah I think it returns an `int` so nvm

sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
```diff
@@ -68,63 +107,130 @@ public void onSpanFinished(final @NotNull ISpan span) {
...
+
+      final long diffNanos = now.diff(spanFinishDate);
+      final long spanEndNanos = nowNanos - diffNanos;
+
+      final long spanStartNanos = spanStartTimeStamp.timeNanos;
+      if (spanStartNanos >= spanEndNanos) {
+        return;
+      }
+
+      final long spanDurationNanos = spanEndNanos - spanStartNanos;
```
romtsn, line 170: just trying to wrap my head around - why this wasn't working and why do we need a separate collection to track start timestamps?

```
 final long spanDurationNanos = spanFinishDate.diff(span.getStartDate());
```

romtsn (approved):
Great stuff overall, I just have a couple of questions to clarify. Great work with the new sample screen for testing this stuff, looks really helpful 🚀

sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
stefanosiano: we can keep this number, as we will only receive slow frames anyway (normal frames are not recorded), but let's change the comment to explicit this is just a constraint to avoid OOMs

sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
stefanosiano: in this case it should work fine

sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
stefanosiano: Actually, if we use dates only for differences (and we are here), we can use `SentryNanotimeDate` without problems

sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
markushi: Because our `ISpans` only have `SentryDate`, which itself has no guarantees around precision - it could be e.g. a time synced unix timestamp. `SentryDate` is also not `Comparable`, so it can't be saved in an ordered set.
I'd be happy to change this within the whole SDK though. Maybe an extra one, to not blow this PR up too much 😅

romtsn: but `SentryDate` is an abstract class, in the end it'll be either `SentryNanotimeDate` or `SentryInstantDate` which both have the nanoseconds precision. And it's actually `Comparable` ;) https://github.com/getsentry/sentry-java/blob/a537f8a45828b751272940034c4a64ee5f8fd278/sentry/src/main/java/io/sentry/SentryDate.java#L48-L51

sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
```diff
@@ -68,63 +107,130 @@ public void onSpanFinished(final @NotNull ISpan span) {
...
+
+    captureFrameMetrics(span);
+
+    synchronized (lock) {
+      if (runningSpans.isEmpty()) {
+        clear();
+      } else {
+        // otherwise only remove old/irrelevant frames
+        final SpanStart oldestSpan = runningSpans.first();
+        frames.headSet(oldestSpan.timeNanos).clear();
```
stefanosiano, line 135: i would remove the frames ended before the start of the first span, while this removes the frames ended before the end of the first span.
Otherwise concurrent spans could have some missing frames, correct?

markushi: > i would remove the frames ended before the start of the first span
Yes, we do exactly that: Only frames which end (`frame.end`) before the oldest span start(`min(span.start)`) are removed here.

sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
markushi: Alright, I've looked into it and got it working [by applying a little trick](https://github.com/getsentry/sentry-java/blob/eedcf5ac53387d44c99f17cdc472ce0bb18dfb31/sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java#L298-L307)

sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
markushi: I don't know why but `SentryNanotimeDate.nanoTimestamp()` only gives you millisecond precision, `diff` however works in nanos. Created a ticket here to follow up on this in the next sync: https://github.com/getsentry/sentry-java/issues/3210

sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
stefanosiano: ok, i've seen you updated it, and it's fine now

sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
stefanosiano: you can remove that trick if you use the `SentryNanotimeDate` in the `Frame` object, instead of nanoseconds

sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
```diff
@@ -133,34 +221,125 @@ public void onSpanFinished(final @NotNull ISpan span) {
...
+   *
+   * @param date the input date
+   * @return a timestamp in nano precision
+   */
+  private static long realNanos(final @NotNull SentryDate date) {
+    return date.diff(UNIX_START_DATE);
+  }
+
+  @ApiStatus.Internal
+  public static class Frame implements Comparable<Frame> {
```
stefanosiano, line 310: this class can be private

stefanosiano (approved):
just make `Frame`  class private and we're good to go.
Your call if you want to change the nanoseconds to `SentryNanotimeDate` (or evaluate to do it in another PR)
very nice!

PR #3200, opened 2024-02-08
Don't wait on main thread when SDK restarts

stefanosiano:
## :scroll: Description
added new methods to close instances with isRestarting flag:
- IHub.close(Boolean)
- ISentryClient.close(Boolean)
- ITransport.close(Boolean)

when the SDK restarts 
- the executor service is now closed in the background
- current network connections are dropped. This is not a problem, as data will be sent anyway after the SDK restarts

We still wait for any processing that is happening, as that's more complex, but the biggest slowdown when restarting the SDK is waiting for the network to send data.

When the SDK is initialized multiple times, it closes the old instance, waiting for any task to be completed.
This PR aims to reduce the wait time by dropping network connections and let any job scheduled to the executor to be finished in the background.
Relates to https://github.com/getsentry/sentry-java/issues/3162

Unit tests
Added Ui test to check it waits for network to timeout on Sentry.close and doesn't wait on multiple Sentry.init()

## :pencil: Checklist

### DIFF
--- a/sentry-android-integration-tests/sentry-uitest-android/src/androidTest/java/io/sentry/uitest/android/SdkInitTests.kt
+++ b/sentry-android-integration-tests/sentry-uitest-android/src/androidTest/java/io/sentry/uitest/android/SdkInitTests.kt
@@ -9,8 +9,6 @@ import io.sentry.android.core.AndroidLogger
 import io.sentry.android.core.SentryAndroidOptions
 import io.sentry.assertEnvelopeTransaction
 import io.sentry.protocol.SentryTransaction
-import okhttp3.mockwebserver.MockResponse
-import okhttp3.mockwebserver.SocketPolicy
 import org.junit.runner.RunWith
 import kotlin.test.Test
 import kotlin.test.assertEquals
@@ -81,9 +79,8 @@ class SdkInitTests : BaseUiTest() {
     @Test
     fun doubleInitDoesNotWait() {
         relayIdlingResource.increment()
-        relayIdlingResource.increment()
         // Let's make the first request timeout
-        relay.addResponse { MockResponse().setSocketPolicy(SocketPolicy.NO_RESPONSE) }
+        relay.addTimeoutResponse()
 
         initSentry(true) { options: SentryAndroidOptions ->
             options.tracesSampleRate = 1.0
@@ -92,14 +89,14 @@ class SdkInitTests : BaseUiTest() {
         Sentry.startTransaction("beforeRestart", "emptyTransaction").finish()
 
         // We want the SDK to start sending the event. If we don't wait, it's possible we don't send anything before the SDK is restarted
-        Thread.sleep(500)
+        waitUntilIdle()
+
+        relayIdlingResource.increment()
+        relayIdlingResource.increment()
 
         val beforeRestart = System.currentTimeMillis()
         // We restart the SDK. This shouldn't block the main thread, but new options (e.g. profiling) should work
-        initSentry(false) { options: SentryAndroidOptions ->
-            // Registering again the idling resource blocks for some time. But we already registered it before.
-            // So we just need to update the mock relay flag, as we are overwriting it in this "initSentry(false)".
-            relay.waitForRequests = true
+        initSentry(true) { options: SentryAndroidOptions ->
             options.tracesSampleRate = 1.0
             options.profilesSampleRate = 1.0
         }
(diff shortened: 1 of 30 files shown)


stefanosiano:
I added a small timeout to `AsyncHttpTransport` when restarting, so that we have time to eventually write any rejected envelope to disk.
I don't really like it, so I'm trying to find a way to improve it

stefanosiano:
> I added a small timeout to `AsyncHttpTransport` when restarting, so that we have time to eventually write any rejected envelope to disk. I don't really like it, so I'm trying to find a way to improve it

Ok, I removed that small timeout. Now, when `AsyncHttpTransport` closes and calls `executor.shutdownNow()`, all the current runnables go throught the rejectedExecutionHandler, that store the envelope to disk

sentry-android-integration-tests/sentry-uitest-android/src/androidTest/java/io/sentry/uitest/android/SdkInitTests.kt
```diff
@@ -74,4 +77,109 @@ class SdkInitTests : BaseUiTest() {
...
+        relay.addResponse { MockResponse().setSocketPolicy(SocketPolicy.NO_RESPONSE) }
+
+        initSentry(true) { options: SentryAndroidOptions ->
+            options.tracesSampleRate = 1.0
+        }
+
+        Sentry.startTransaction("beforeRestart", "emptyTransaction").finish()
+
+        // We want the SDK to start sending the event. If we don't wait, it's possible we don't send anything before the SDK is restarted
+        Thread.sleep(500)
```
markushi, line 95: would it make sense to have a `@TestOnly` hook inside our transport instead? 500ms sounds like a lot of waiting time at least.

sentry/src/main/java/io/sentry/transport/AsyncHttpTransport.java
```diff
@@ -163,16 +164,29 @@ public boolean isHealthy() {
...
             .getLogger()
             .log(
                 SentryLevel.WARNING,
-                "Failed to shutdown the async connection async sender within 1 minute. Trying to force it now.");
-        executor.shutdownNow();
+                "Failed to shutdown the async connection async sender  within "
+                    + timeout
+                    + " ms. Trying to force it now.");
+        final @NotNull List<Runnable> runnables = executor.shutdownNow();
+        for (final @NotNull Runnable r : runnables) {
```
markushi, line 186: Is this really required? I thought this is already happening here: https://github.com/getsentry/sentry-java/blob/a537f8a45828b751272940034c4a64ee5f8fd278/sentry/src/main/java/io/sentry/transport/AsyncHttpTransport.java#L134-L146

markushi (changes requested):
Looking good, I left 2 comments for clarification!

sentry-android-integration-tests/sentry-uitest-android/src/androidTest/java/io/sentry/uitest/android/SdkInitTests.kt
stefanosiano: Looks like an overkill honestly 😅 
we can reduce the sleep to 100ms, it should work fine.
Let me check if i can use some hook (maybe one in the test's mock relay), if not i'll reduce the sleep

stefanosiano: I made the idlingResource decrement when receiving a network call and responding with a timeout, so i'll just use that, and it will work

sentry/src/main/java/io/sentry/transport/AsyncHttpTransport.java
stefanosiano: for some reason, the `shutdownNow` method was not always returning the runnable currently executing, so I had to use a field to store the current runnable and use that to store the envelope in case it closes

sentry/src/test/java/io/sentry/transport/AsyncHttpTransportTest.kt
```diff
@@ -336,6 +338,39 @@ class AsyncHttpTransportTest {
...
+    fun `close with isRestarting true does not await termination`() {
+        fixture.sentryOptions.flushTimeoutMillis = 123
+        val sut = fixture.getSUT()
+        sut.close(true)
+
+        verify(fixture.executor).awaitTermination(eq(0), eq(TimeUnit.MILLISECONDS))
+    }
+
+    @Test
+    fun `close shuts down the executor and runs executing runnable through rejectedExecutionHandler`() {
```
markushi, line 360: nice :+1:

markushi (approved):
LGTM!

PR #3203, opened 2024-02-13
[SR] Add screenshot recorder

romtsn:
_#skip-changelog_

This is just a dump of the PoC phase more-or-less, so a lot will be refactored and made pretty, but you're welcome to review nonetheless. I've tried to left TODOs where I see changes forthcoming 

Relates to: 
* https://github.com/getsentry/sentry/issues/63255

### DIFF
--- a/sentry-android-replay/src/main/java/io/sentry/android/replay/ScreenshotRecorder.kt
+++ b/sentry-android-replay/src/main/java/io/sentry/android/replay/ScreenshotRecorder.kt
@@ -1,0 +1,150 @@
+package io.sentry.android.replay
+
+import android.graphics.Bitmap
+import android.graphics.Canvas
+import android.graphics.Color
+import android.graphics.Paint
+import android.graphics.RectF
+import android.os.Handler
+import android.os.HandlerThread
+import android.os.Looper
+import android.os.SystemClock
+import android.util.Log
+import android.view.PixelCopy
+import android.view.View
+import android.view.ViewGroup
+import android.view.ViewTreeObserver
+import io.sentry.android.replay.video.SimpleVideoEncoder
+import io.sentry.android.replay.viewhierarchy.ViewHierarchyNode
+import java.lang.ref.WeakReference
+import java.util.WeakHashMap
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit.MILLISECONDS
+import kotlin.system.measureTimeMillis
+
+class ScreenshotRecorder(
+  val rootView: WeakReference<View>,
+  val encoder: SimpleVideoEncoder
+) : ViewTreeObserver.OnDrawListener {
+
+  private val thread = HandlerThread("SentryReplay").also { it.start() }
+  private val handler = Handler(thread.looper)
+  private val bitmapToVH = WeakHashMap<Bitmap, ViewHierarchyNode>()
+
+  companion object {
+    const val TAG = "ScreenshotRecorder"
+  }
+
+  private var lastCapturedAtMs: Long? = null
+  override fun onDraw() {
+    // cheap debounce for testing
+    val now = SystemClock.uptimeMillis()
+    if (lastCapturedAtMs != null && (now - lastCapturedAtMs!!) < 1000L) {
+      return
+    }
+    lastCapturedAtMs = now
+
+    val root = rootView.get()
+    if (root == null || root.width <= 0 || root.height <= 0) {
+      return
+    }
+
+    val window = root.phoneWindow ?: return
+    val bitmap = Bitmap.createBitmap(
+      root.width,
+      root.height,
+      Bitmap.Config.ARGB_8888
+    )
+    Log.e("BITMAP CREATED", bitmap.toString())
+
+    val time = measureTimeMillis {
+      val rootNode = ViewHierarchyNode.fromView(root)
+      root.traverse(rootNode)
+      bitmapToVH[bitmap] = rootNode
+    }
+    Log.e("TIME", time.toString())
+
+//    val latch = CountDownLatch(1)
+
+    Handler(Looper.getMainLooper()).postAtFrontOfQueue {
+      PixelCopy.request(
+        window,
+        bitmap,
+        { copyResult: Int ->
+          Log.d(TAG, "PixelCopy result: $copyResult")
+          if (copyResult != PixelCopy.SUCCESS) {
+            Log.e(TAG, "Failed to capture screenshot")
+            return@request
+          }
+
+          Log.e("BITMAP CAPTURED", bitmap.toString())
+          val viewHierarchy = bitmapToVH[bitmap]
+
+          if (viewHierarchy != null) {
+            val canvas = Canvas(bitmap)
+            viewHierarchy.traverse {
+              if (it.shouldRedact && (it.width > 0 && it.height > 0)) {
+                it.visibleRect ?: return@traverse
+
+                val paint = Paint().apply {
+                  color = it.dominantColor ?: Color.BLACK
+                }
+                canvas.drawRoundRect(RectF(it.visibleRect), 10f, 10f, paint)
+              }
+            }
+          }
+
+          val scaledBitmap = Bitmap.createScaledBitmap(
+            bitmap,
+            encoder.muxerConfig.videoWidth,
+            encoder.muxerConfig.videoHeight,
+            true
+          )
+//        val baos = ByteArrayOutputStream()
+//        scaledBitmap.compress(Bitmap.CompressFormat.JPEG, 75, baos)
+//        val bmp = BitmapFactory.decodeByteArray(baos.toByteArray(), 0, baos.size())
+          encoder.encode(scaledBitmap)
+//        bmp.recycle()
+          scaledBitmap.recycle()
+          bitmap.recycle()
+          Log.i(TAG, "Captured a screenshot")
+//          latch.countDown()
+        },
+        handler
+      )
+    }
+
+//    val success = latch.await(200, MILLISECONDS)
+//    Log.i(TAG, "Captured a screenshot: $success")
+  }
+
+  private fun ViewHierarchyNode.traverse(callback: (ViewHierarchyNode) -> Unit) {
+    callback(this)
+    if (this.children != null) {
+      this.children!!.forEach {
+        it.traverse(callback)
+      }
+    }
+  }
+
+  private fun View.traverse(parentNode: ViewHierarchyNode) {
+    if (this !is ViewGroup) {
+      return
+    }
+
+    if (this.childCount == 0) {
+      return
+    }
+
+    val childNodes = ArrayList<ViewHierarchyNode>(this.childCount)
+    for (i in 0 until childCount) {
+      val child = getChildAt(i)
+      if (child != null) {
+        val childNode = ViewHierarchyNode.fromView(child)
+        childNodes.add(childNode)
+        child.traverse(childNode)
+      }
+    }
+    parentNode.children = childNodes
+  }
+}
--- a/sentry-android-replay/src/main/java/io/sentry/android/replay/ScreenshotRecorder.kt
+++ b/sentry-android-replay/src/main/java/io/sentry/android/replay/ScreenshotRecorder.kt
@@ -18,132 +18,130 @@ import io.sentry.android.replay.video.SimpleVideoEncoder
 import io.sentry.android.replay.viewhierarchy.ViewHierarchyNode
 import java.lang.ref.WeakReference
 import java.util.WeakHashMap
-import java.util.concurrent.CountDownLatch
-import java.util.concurrent.TimeUnit.MILLISECONDS
 import kotlin.system.measureTimeMillis
 
 internal class ScreenshotRecorder(
-  val rootView: WeakReference<View>,
-  val encoder: SimpleVideoEncoder
+    val rootView: WeakReference<View>,
+    val encoder: SimpleVideoEncoder
 ) : ViewTreeObserver.OnDrawListener {
 
-  private val thread = HandlerThread("SentryReplay").also { it.start() }
-  private val handler = Handler(thread.looper)
-  private val bitmapToVH = WeakHashMap<Bitmap, ViewHierarchyNode>()
+    private val thread = HandlerThread("SentryReplay").also { it.start() }
+    private val handler = Handler(thread.looper)
+    private val bitmapToVH = WeakHashMap<Bitmap, ViewHierarchyNode>()
 
-  companion object {
-    const val TAG = "ScreenshotRecorder"
-  }
-
-  private var lastCapturedAtMs: Long? = null
-  override fun onDraw() {
-    val now = SystemClock.uptimeMillis()
-    if (lastCapturedAtMs != null && (now - lastCapturedAtMs!!) < 1000L) {
-      return
-    }
-    lastCapturedAtMs = now
-
-    val root = rootView.get()
-    if (root == null || root.width <= 0 || root.height <= 0) {
-      return
+    companion object {
+        const val TAG = "ScreenshotRecorder"
     }
 
-    val window = root.phoneWindow ?: return
-    val bitmap = Bitmap.createBitmap(
-      root.width,
-      root.height,
-      Bitmap.Config.ARGB_8888
-    )
-    Log.e("BITMAP CREATED", bitmap.toString())
+    private var lastCapturedAtMs: Long? = null
+    override fun onDraw() {
+        val now = SystemClock.uptimeMillis()
+        if (lastCapturedAtMs != null && (now - lastCapturedAtMs!!) < 1000L) {
+            return
+        }
+        lastCapturedAtMs = now
 
-    val time = measureTimeMillis {
-      val rootNode = ViewHierarchyNode.fromView(root)
-      root.traverse(rootNode)
-      bitmapToVH[bitmap] = rootNode
-    }
-    Log.e("TIME", time.toString())
+        val root = rootView.get()
+        if (root == null || root.width <= 0 || root.height <= 0) {
+            return
+        }
+
+        val window = root.phoneWindow ?: return
+        val bitmap = Bitmap.createBitmap(
+            root.width,
+            root.height,
+            Bitmap.Config.ARGB_8888
+        )
+        Log.e("BITMAP CREATED", bitmap.toString())
+
+        val time = measureTimeMillis {
+            val rootNode = ViewHierarchyNode.fromView(root)
+            root.traverse(rootNode)
+            bitmapToVH[bitmap] = rootNode
+        }
+        Log.e("TIME", time.toString())
 
 //    val latch = CountDownLatch(1)
 
-    Handler(Looper.getMainLooper()).postAtFrontOfQueue {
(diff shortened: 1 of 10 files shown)


sentry-android-replay/src/main/java/io/sentry/android/replay/video/SimpleVideoEncoder.kt
```diff
@@ -0,0 +1,167 @@
...
+import android.graphics.Bitmap
+import android.media.MediaCodec
+import android.media.MediaCodecInfo
+import android.media.MediaFormat
+import android.os.Handler
+import android.os.Looper
+import android.view.Surface
+import java.io.File
+
+internal class SimpleVideoEncoder(
```
romtsn, line 41: @vaind I'd imagine you want this public right? Your idea is to take screenshots in flutter but then use the native video encoder to create a video?

vaind: Yes, although the exact API draw API that would have the least overhead is TBD. But the DrawBitmap looks like a good start, I'm just not sure I can make it work without an additional mem copy. I've noticed there's DrawPicture but I'm not sure that would be any better.

romtsn: Ok, so the idea how we wanna do it is:
* Take screenshots every second and write them to disk
* Have a file observer or something that tracks when there are 5 screenshots (=5 seconds), then it:
  * Encodes a 5sec video
  * Creates a replay envelope and sends it

I wonder if could piggy-back on that behavior in flutter? Probably envelope creation should still be on your side because you want to pass the event through processors, collect flutter breadcrumbs etc etc, but other parts can be certainly reused. 

basically it'll be `$options.cacheDir/replay` where you'd store screenshots and then when we create a video, and we could give you a callback  like `onVideoEncoded(segmentId, bytes)` that you could use to create envelopes on your side. wdyt?

vaind: > Take screenshots every second and write them to disk

Any reason to do this instead of encoding directly to file while the screenshot is already in memory? IO is slow and costly

romtsn: yeah, at the very least - native crashes and ANRs where we'd lose the last segment if we were to keep it in memory. @brustolin or @bruno-garcia might give more details, I think I was absent when they've implemented that

vaind: I have no clue how the video encoder works, but if you can flush it to disk incrementally, than you could do that after each frame. But maybe that's not an option because then it actually cannot encode efficiently? 🤷

romtsn: yeah, I don't think that's an option, the video would be corrupted if you don't call `encoder.stop` or something from what I recall. but I can try that out

bruno-garcia: if we have a safe way to write to disk on case of a crash, we could keep it in memory. We'd lose the frames in case the app was killed by the user though

romtsn: only for crashes, but native crashes and ANRs we don't have signals for to safely finish encoding, so I guess we'd have to continously store them to disk and encode a video on demand

sentry-android-replay/src/main/java/io/sentry/android/replay/viewhierarchy/ViewHierarchyNode.kt
```diff
@@ -0,0 +1,173 @@
...
+                    } else {
+                        AccessibilityNodeInfo.obtain()
+                    }
+                    view.onInitializeAccessibilityNodeInfo(nodeInfo)
+                    shouldRedact = nodeInfo.isVisibleToUser
+                    nodeInfo.recycle()
+                    if (shouldRedact) {
+                        val bounds = Rect()
+                        val text = view.text.toString()
+                        view.paint.getTextBounds(text, 0, text.length, bounds)
```
brustolin, line 79: This is nice. I need to check whether I can do this in iOS.

sentry-android-replay/src/main/java/io/sentry/android/replay/viewhierarchy/ViewHierarchyNode.kt
```diff
@@ -0,0 +1,173 @@
...
+                        }
+                        // TODO: support known 3rd-party widgets like MaterialButton with an icon
+                        // TODO: also calculate height properly based on text bounds
+                        rect.left += textStart + view.paddingStart
+                        rect.right = rect.left + (textEnd - textStart)
+                    }
+                }
+
+                is ImageView -> {
+                    shouldRedact = isVisible(view) && (view.drawable?.isRedactable() ?: false)
```
brustolin, line 111: m: Can we check whether the image belongs to the APK or it was loaded from outside?

sentry-android-replay/src/main/java/io/sentry/android/replay/viewhierarchy/ViewHierarchyNode.kt
```diff
@@ -0,0 +1,173 @@
...
+                shouldRedact,
+                dominantColor,
+                rect
+            )
+        }
+
+        private fun Drawable.isRedactable(): Boolean {
+            return when (this) {
+                is InsetDrawable, is ColorDrawable, is VectorDrawable, is GradientDrawable -> false
+                is BitmapDrawable -> !bitmap.isRecycled && bitmap.height > 0 && bitmap.width > 0
```
brustolin, line 133: m: I would not use 0 but something a little bigger (im using 10 for iOS), some backgrounds are composed by small tiled images.

sentry-android-replay/src/main/java/io/sentry/android/replay/viewhierarchy/ViewHierarchyNode.kt
```diff
@@ -0,0 +1,173 @@
...
+                else -> true
+            }
+        }
+
+        private fun Drawable.pickDominantColor(): Int {
```
brustolin, line 138: m: This sounds heavy. Creating a new bitmap to calculate the average color seems to be an overkill, can we just use black or white based on the theme?

sentry-android-replay/src/main/java/io/sentry/android/replay/viewhierarchy/ViewHierarchyNode.kt
romtsn: no, there's no really a way to know this for sure, but we can do heuristics like seen in the `isRedactable` method - if it's not a `BitmapDrawable` then most likely it's coming from the APK (like icons or backgrounds). But if it's a BItmapDrawable then there's a chance we'll redact something that is bundled with the APK, but I think it's ok. We could do something at build time to preserve this information via the gradle plugin, but sounds like an overkill

sentry-android-replay/src/main/java/io/sentry/android/replay/viewhierarchy/ViewHierarchyNode.kt
romtsn: yeah, good idea 👍

sentry-android-replay/src/main/java/io/sentry/android/replay/viewhierarchy/ViewHierarchyNode.kt
romtsn: yeah, it doesn't look as nice with black rectangles, but I also left a TODO down below sayin the same thing - maybe we can expose an option for that, and default to black/white rectangle. I'll also look into doing it off the main thread if possible

sentry-android-replay/src/main/java/io/sentry/android/replay/ScreenshotRecorder.kt
```diff
@@ -0,0 +1,148 @@
...
+    override fun onDraw() {
+        // TODO: replace with Debouncer from sentry-core
+        val now = SystemClock.uptimeMillis()
+        if (lastCapturedAtMs != null && (now - lastCapturedAtMs!!) < 1000L) {
+            return
+        }
+        lastCapturedAtMs = now
+
+        val root = rootView.get()
+        if (root == null || root.width <= 0 || root.height <= 0) {
```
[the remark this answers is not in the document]
romtsn: unlikely the root view will be opaque or invisible, but worth to check nonetheless 👍

sentry-android-replay/src/main/java/io/sentry/android/replay/viewhierarchy/ViewHierarchyNode.kt
brustolin: Im just worried avout redacting big portions based on a image background.

romtsn: yeah, makes sense 👍 I'll leave a TODO to see if we can do anything about that

vaind (commented):
Not much to review for me here, because, as you're saying, "This is just a dump of the PoC".

I'd be interested to see the refactored version that would be usable from outside sentry-java

sentry-android-replay/src/main/java/io/sentry/android/replay/video/SimpleFrameMuxer.kt
```diff
@@ -0,0 +1,47 @@
+/**
+ * Adapted from https://github.com/fzyzcjy/flutter_screen_recorder/blob/master/packages/fast_screen_recorder/android/src/main/kotlin/com/cjy/fast_screen_recorder/SimpleFrameMuxer.kt
```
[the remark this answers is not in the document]
romtsn: I think we have to put it to every file we borrowed according to our open-source guidelines

romtsn:
> Not much to review for me here, because, as you're saying, "This is just a dump of the PoC".
> 
> I'd be interested to see the refactored version that would be usable from outside sentry-java

You got to tell me what do you want to have exposed then :)

vaind:
> You got to tell me what do you want to have exposed then :)

I imagine the code that captures images is going to be separate from the code that produces the video, at least from your description how it's going to be done in two steps.
In order to use that from another SDK, we'd need an interface that allows constructing this envelope, i.e. creating the video (from bitmaps), adding replay events ans such. 
I assume sending the envelope is no different than any other envelope so that should be already done.

romtsn:
> i.e. creating the video (from bitmaps)

Yeah, this makes sense, I'll create a new method in SimpleVideoEncoder that creates a video out of a bitmap list. But this:

> we'd need an interface that allows constructing this envelope

I'm not sure. Don't we usually do that on the hybrid SDK level and then just pass the envelope bytes to `captureEnvelope`? https://github.com/getsentry/sentry-dart/blob/732a7b44ad1624e75967be2227665c066ef0154e/dart/lib/src/sentry_envelope.dart

romtsn:
> > > i.e. creating the video (from bitmaps)
> > 
> > 
> > Yeah, this makes sense, I'll create a new method in SimpleVideoEncoder that creates a video out of a bitmap list. But this:
> > > we'd need an interface that allows constructing this envelope
> > 
> > 
> > I'm not sure. Don't we usually do that on the hybrid SDK level and then just pass the envelope bytes to `captureEnvelope`? [getsentry/sentry-dart@`732a7b4`/dart/lib/src/sentry_envelope.dart](https://github.com/getsentry/sentry-dart/blob/732a7b44ad1624e75967be2227665c066ef0154e/dart/lib/src/sentry_envelope.dart)
> 
> In this case we won't be creating the envelope in dart due to the cost of transferring the data. Instead, sentry-dart has a native kotlin binding that would call native functions to create and send the envelope.

Alright, let me give it a stab for the Android SDK first and then we can discuss in the following PR what has to be changed and so on. I imagine this is all the glue code I'm gonna write to connect recording and envelopes that is the most valuable for you.

markushi (changes requested):
Looking good overall! Left a few comments. We probably can improve on caching / handling bitmap allocation as they seem to happen in multiple places.

sentry-android-replay/src/main/java/io/sentry/android/replay/ScreenshotRecorder.kt
```diff
@@ -0,0 +1,148 @@
...
+
+                                val paint = Paint().apply {
+                                    color = it.dominantColor ?: Color.BLACK
+                                }
+                                canvas.drawRoundRect(RectF(it.visibleRect), 10f, 10f, paint)
+                            }
+                        }
+                    }
+
+                    val scaledBitmap = Bitmap.createScaledBitmap(
```
markushi, line 95: You could also retain an instance of `scaledBitmap` and use `canvas.drawBitmap` to draw the original bitmap into the scaled down version

sentry-android-replay/src/main/java/io/sentry/android/replay/ScreenshotRecorder.kt
```diff
@@ -0,0 +1,148 @@
...
+                        return@request
+                    }
+
+                    Log.e("BITMAP CAPTURED", bitmap.toString())
+                    val viewHierarchy = bitmapToVH[bitmap]
+
+                    if (viewHierarchy != null) {
+                        val canvas = Canvas(bitmap)
+                        viewHierarchy.traverse {
+                            if (it.shouldRedact && (it.width > 0 && it.height > 0)) {
```
markushi, line 84: in theory all redaction could happen on the scaled down version
to  avoid manually scaling every `it.visibleRect`, simply set a matrix on the canvas based on the aspect ratio

sentry-android-replay/src/main/java/io/sentry/android/replay/ScreenshotRecorder.kt
```diff
@@ -0,0 +1,148 @@
...
+
+    companion object {
+        const val TAG = "ScreenshotRecorder"
+    }
+
+    private var lastCapturedAtMs: Long? = null
+    override fun onDraw() {
+        // TODO: replace with Debouncer from sentry-core
+        val now = SystemClock.uptimeMillis()
+        if (lastCapturedAtMs != null && (now - lastCapturedAtMs!!) < 1000L) {
```
markushi, line 40: - [ ] move 1s delay into a constant

sentry-android-replay/src/main/java/io/sentry/android/replay/video/SimpleVideoEncoder.kt
romtsn: created a task under SDK Common just not to forget https://github.com/getsentry/sentry/issues/63255

sentry-android-replay/src/main/java/io/sentry/android/replay/Windows.kt
```diff
@@ -0,0 +1,203 @@
+/**
+ * Adapted from https://github.com/square/curtains
```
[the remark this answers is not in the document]
romtsn: this is composed from different files of the repo, but I'll change to having a tag in the link at least

sentry-android-replay/src/main/java/io/sentry/android/replay/ScreenshotRecorder.kt
```diff
@@ -0,0 +1,148 @@
...
+            return
+        }
+
+        val window = root.phoneWindow ?: return
+        val bitmap = Bitmap.createBitmap(
+            root.width,
+            root.height,
+            Bitmap.Config.ARGB_8888
+        )
+        Log.e("BITMAP CREATED", bitmap.toString())
```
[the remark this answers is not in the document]
romtsn: will leave a TODO after hooking it with the sdk

sentry-android-replay/src/main/java/io/sentry/android/replay/viewhierarchy/ViewHierarchyNode.kt
romtsn: We've improved this significantly in https://github.com/getsentry/sentry-java/pull/3203/commits/1ce57cbd24afa08a583d28e0ab6e2365a0592203 - now we just reuse the same single-pixel bitmap where we draw portions of the screenshot that need redacting + also doing this off-the-main-thread

sentry-android-replay/src/main/java/io/sentry/android/replay/viewhierarchy/ViewHierarchyNode.kt
```diff
@@ -47,11 +47,11 @@ data class ViewHierarchyNode(
...
-        fun adjustAlpha(color: Int): Int {
+        fun Int.toOpaque(): Int {
             val alpha = 255
-            val red = Color.red(color)
-            val green = Color.green(color)
-            val blue = Color.blue(color)
+            val red = Color.red(this)
+            val green = Color.green(this)
+            val blue = Color.blue(this)
             return Color.argb(alpha, red, green, blue)
```
markushi, line 55: Future improvement in case you need low level optimizations: do some bit wise composition, like [it's done here](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/com/android/internal/graphics/ColorUtils.java;l=372-379;drc=dde365777a3364f5ec32975fcd5af7f0fd681c80).

sentry-android-replay/src/main/java/io/sentry/android/replay/ScreenshotRecorder.kt
```diff
@@ -78,26 +90,37 @@ internal class ScreenshotRecorder(
                     Log.e("BITMAP CAPTURED", bitmap.toString())
                     val viewHierarchy = bitmapToVH[bitmap]
 
-                    if (viewHierarchy != null) {
-                        val canvas = Canvas(bitmap)
+                    val scaledBitmap = Bitmap.createScaledBitmap(
```
markushi, line 93: `l`: You could have the `if (viewHierarchy == null) {` check before creating the `scaledBitmap`, to save some unnecessary memory allocation

markushi (approved):
Nice one, looks great!

sentry-android-replay/src/main/java/io/sentry/android/replay/viewhierarchy/ViewHierarchyNode.kt
```diff
@@ -47,11 +47,11 @@ data class ViewHierarchyNode(
             return actualPosition.intersects(screen.left, screen.top, screen.right, screen.bottom)
         }
 
-        fun adjustAlpha(color: Int): Int {
+        fun Int.toOpaque(): Int {
```
markushi, line 50: `l`: could be simplified to 
```kotlin
fun Int.toOpaque() = this or 0xFF000000.toInt()`
```

romtsn: Yeah I've tried that but for some reason it didn't work on RN somehow... I will add a todo to verify whether it works

Issue #3208, opened 2024-02-02
Kotlin Compiler Plugin does not tag @Composable functions


audkar:
### Gradle Version

8.5

### AGP Version

8.2.2

### Code Minifier/Optimizer

None

### Version

4.2.0

### Sentry SDK Version

7.3.0

### Steps to Reproduce

Using `androidx.compose:compose-bom:2023.10.01`

Basically following docs https://docs.sentry.io/platforms/android/enhance-errors/kotlin-compiler-plugin/

1. Add Sentry Kotlin Gradle Plugin
2. Create simple Composable (same as in Sentry Docs). Add to Activity View Hiearary with `binding.composeView.setContent { LoginScreen() }`
3. Build application `gradle assembleDebug`
4. ~~Inspect compiled output (Maybe I am missing something, but I can't find anything changed)~~ Found the modifications in bytecode.
5. Click an child Composable
6. Observe logs printed by Sentry SDK (with `isDebug=true`): `Unable to find click target. No breadcrumb captured.` 

Project is set up correctly. If I add `testTag` manually, I can see user interaction breadcrumbs logged in Sentry Web dashboard.

### Expected Result

User interaction breadcrumbs are logged for `Composable` that do not have explicitly set `testTag`

### Actual Result

Nothing logged

PR #3209, opened 2024-02-16
Fix Compose widgets are not being correctly identified

markushi:
## :scroll: Description

Using the latest version of Compose, the compiler plugin sometimes does not fold `SentryTraced` into the same UI Node element anymore. This causes our `SentryTag` semantic to be inside the parent node, but the `Clickable` semantic ends up in the child node.

Fixes https://github.com/getsentry/sentry-java/issues/3208

I was able to reproduce the issue locally, did some manual tests.

We definitely need to get better at automating this, right now it's super painful as newer versions of Compose require a newer versions of Kotlin. This in turn requires bumping some plugins (e.g. detekt), which then in turn emit new build time errors.

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
 - Add new threshold parameters to monitor config ([#3181](https://github.com/getsentry/sentry-java/pull/3181))
 - Report process init time as a span for app start performance ([#3159](https://github.com/getsentry/sentry-java/pull/3159))
 
+## Fixes
+
+- Fix Jetpack Compose widgets are not being correctly identified for user interaction tracing ([#3209](https://github.com/getsentry/sentry-java/pull/3209))
+
 ## 7.3.0
 
 ### Features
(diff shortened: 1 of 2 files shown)


markushi:
> Do you mean our compiler plugin or the compose compiler plugin?
Ah sorry, I meant the Jetpack Compose Compiler plugin.

romtsn:
> > Do you mean our compiler plugin or the compose compiler plugin?
> 
> Ah sorry, I meant the Jetpack Compose Compiler plugin.

yeah, I will think of strategies, I still have it on my to-do. Could you post which versions of Compose and Kotlin are specifically causing this issue?

markushi:
> > > Do you mean our compiler plugin or the compose compiler plugin?
> > 
> > 
> > Ah sorry, I meant the Jetpack Compose Compiler plugin.
> 
> yeah, I will think of strategies, I still have it on my to-do. Could you post which versions of Compose and Kotlin are specifically causing this issue?
```
val kotlinVersion = "1.9.22"
val composeVersion = "1.5.12"
val androidComposeCompilerVersion = "1.5.9"
```

PR #3214, opened 2024-02-20
[SR] Add session replay envelope and events

### LINKED ISSUE #3213: Implement new envelope items: Replay, ReplayRecording, ReplayVideo

romtsn:
_#skip-changelog_
## :scroll: Description

You can read up more about the new envelope spec here https://github.com/getsentry/rfcs/pull/129

It has a good chunk of work that you've done during the hackweek, but with several enhancements:
* RRWeb events are now typed
* Properly capture replay events with recordings in SentryClient and record client reports if e.g. an event is dropped by an event processor
* Also have a separate method for applying scope as replay events don't have everything as regular events (e.g. no breadcrumbs)
* I had to introduce a new `MapObjectReader` and extract an `ObjectReader` interface from `JsonObjectReader` but I made a separate PR for this (this one will be failing until the other one is merged)

Closes #3213 

With tests

## :pencil: Checklist

## :crystal_ball: Next steps
Glue recording and envelopes together


romtsn:
@markushi @stefanosiano I'll wait with merging #3215 otherwise it'll be hell to review this

sentry/src/main/java/io/sentry/SentryReplayEvent.java
```diff
@@ -0,0 +1,324 @@
...
+    writer
+        .name(JsonKeys.TIMESTAMP)
+        .value(logger, BigDecimal.valueOf(DateUtils.dateToSeconds(timestamp)));
+    if (replayId != null) {
+      writer.name(JsonKeys.REPLAY_ID).value(logger, replayId);
+    }
+    if (replayStartTimestamp != null) {
+      writer
+          .name(JsonKeys.REPLAY_START_TIMESTAMP)
+          .value(logger, BigDecimal.valueOf(DateUtils.dateToSeconds(replayStartTimestamp)));
```
markushi, line 204:
```suggestion
          .value(logger, DateUtils.dateToSeconds(replayStartTimestamp));
```

markushi (approved):
Left a few minor comments, and stopped at some point adding final and `@NotNull` annotations. But apart from that, LGTM!

sentry/src/main/java/io/sentry/ReplayRecording.java
```diff
@@ -0,0 +1,182 @@
...
+                    final String tag =
+                        (String) data.getOrDefault(RRWebEvent.JsonKeys.TAG, "default");
+                    switch (tag) {
+                      case RRWebVideoEvent.EVENT_TAG:
+                        final RRWebEvent videoEvent =
+                            new RRWebVideoEvent.Deserializer().deserialize(mapReader, logger);
+                        payload.add(videoEvent);
+                        break;
+                      default:
+                        logger.log(SentryLevel.DEBUG, "Unsupported rrweb event type %s", type);
```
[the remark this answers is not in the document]
romtsn: I think it's better to have an incomplete recording, e.g. if there's now a new type like breadcrumbs added on hybrid sdks or wherever, we can still have a replay but without breadcrumbs, which is still valuable

sentry/src/main/java/io/sentry/SentryReplayEvent.java
```diff
@@ -0,0 +1,324 @@
...
+  @SuppressWarnings("JdkObsolete")
+  public void serialize(final @NotNull ObjectWriter writer, final @NotNull ILogger logger)
+      throws IOException {
+    writer.beginObject();
+    writer.name(JsonKeys.TYPE).value(type);
+    writer.name(JsonKeys.REPLAY_TYPE).value(logger, replayType);
+    writer.name(JsonKeys.SEGMENT_ID).value(segmentId);
+    writer
+        .name(JsonKeys.TIMESTAMP)
+        .value(logger, BigDecimal.valueOf(DateUtils.dateToSeconds(timestamp)));
```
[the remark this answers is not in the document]
romtsn: Yeah the reason was that the milliseconds precision is lost (e.g. `12345.678` will transform to something like `12345.6E10`) when using plain double and converting it to BigDecimal helped. But I think it's not necessary actually, because we can just send ISO Date and relay should take care of converting it to whatever is needed for frontend, so I'll change that 👍

romtsn:
> and stopped at some point adding final and @NotNull annotations

thank you! I will revise that myself and add where necessary

sentry/src/main/java/io/sentry/Sentry.java
```diff
@@ -1025,4 +1025,9 @@ public interface OptionsConfiguration<T extends SentryOptions> {
   public static @NotNull SentryId captureCheckIn(final @NotNull CheckIn checkIn) {
     return getCurrentHub().captureCheckIn(checkIn);
   }
+
+  public static void captureReplay(
+      final @NotNull SentryReplayEvent replay, final @Nullable Hint hint) {
+    getCurrentHub().captureReplay(replay, hint);
+  }
```
brustolin, line 1032: m: Why is this public? In what scenario can the user manually create a replay event?

IMO we should not expose any replay related object in the first versions.

sentry/src/main/java/io/sentry/rrweb/RRWebVideoEvent.java
```diff
@@ -0,0 +1,433 @@
...
+    this.left = left;
+  }
+
+  public int getTop() {
+    return top;
+  }
+
+  public void setTop(final int top) {
+    this.top = top;
+  }
```
brustolin, line 156: l: Can you see this ever being any other value than zero?

brustolin (approved):
I reviewed only the Session Replay related changes (all the serialization change could have been a different PR), and it looks good.
I wrote just a few comments for discussion.

sentry/src/main/java/io/sentry/Sentry.java
romtsn: yeah, I wanted to remove it but overlooked 👍 thanks

sentry/src/main/java/io/sentry/rrweb/RRWebVideoEvent.java
romtsn: yeah, probalby when picture-in-picture mode or window is like a dialog window. And it is also relevant for flutter https://github.com/getsentry/rfcs/pull/129#discussion_r1483928742

PR #3215, opened 2024-02-20
[SR] Introduce MapObjectReader

romtsn:
_#skip-changelog_

Add new sentry-android-replay module## :scroll: Description

* Extract an interface `ObjectReader` from `JsonObjecteReader` and introduce a new `MapObjectReader` analogous to what @markushi has done for the hybrid SDKs in https://github.com/getsentry/sentry-java/pull/2814
  * This is necessary to transform object from Maps to typed object for RRWebEvents since we hold a list of generics and lose their type at runtime.
* I also think this probably will be useful in the future if hybrid sdks will pass the data to the native ones in form of dictionaries (probably gonna be the case for Flutter session replay)

There are still some rough edges I believe, but it works for the current replay usecase, and I think it should be fine. I'll add some tests but won't cover everything (will fix stuff along-the-way if I find something).

## Next Steps
Add some tests for the MapObjectReader

### DIFF
--- a/sentry/src/main/java/io/sentry/Breadcrumb.java
+++ b/sentry/src/main/java/io/sentry/Breadcrumb.java
@@ -86,8 +86,7 @@ public final class Breadcrumb implements JsonUnknown, JsonSerializable {
       switch (entry.getKey()) {
         case JsonKeys.TIMESTAMP:
           if (value instanceof String) {
-            Date deserializedDate =
-                JsonObjectReader.dateOrNull((String) value, options.getLogger());
+            Date deserializedDate = ObjectReader.dateOrNull((String) value, options.getLogger());
             if (deserializedDate != null) {
               timestamp = deserializedDate;
             }
@@ -700,8 +699,8 @@ public final class Breadcrumb implements JsonUnknown, JsonSerializable {
   public static final class Deserializer implements JsonDeserializer<Breadcrumb> {
     @SuppressWarnings("unchecked")
     @Override
-    public @NotNull Breadcrumb deserialize(
-        @NotNull JsonObjectReader reader, @NotNull ILogger logger) throws Exception {
+    public @NotNull Breadcrumb deserialize(@NotNull ObjectReader reader, @NotNull ILogger logger)
+        throws Exception {
       reader.beginObject();
       @NotNull Date timestamp = DateUtils.getCurrentDateTime();
       String message = null;
--- a/sentry/src/main/java/io/sentry/ObjectReader.java
+++ b/sentry/src/main/java/io/sentry/ObjectReader.java
@@ -11,7 +11,8 @@ import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 public interface ObjectReader extends Closeable {
-  static @Nullable Date dateOrNull(@Nullable String dateString, final @NotNull ILogger logger) {
+  static @Nullable Date dateOrNull(
+      final @Nullable String dateString, final @NotNull ILogger logger) {
     if (dateString == null) {
       return null;
     }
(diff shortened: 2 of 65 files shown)


sentry/src/main/java/io/sentry/Breadcrumb.java
```diff
@@ -86,8 +86,7 @@ public static Breadcrumb fromMap(
       switch (entry.getKey()) {
         case JsonKeys.TIMESTAMP:
           if (value instanceof String) {
-            Date deserializedDate =
-                JsonObjectReader.dateOrNull((String) value, options.getLogger());
+            Date deserializedDate = ObjectReader.dateOrNull((String) value, options.getLogger());
```
markushi, line 89:
```suggestion
            final Date deserializedDate = ObjectReader.dateOrNull((String) value, options.getLogger());
```

sentry/src/main/java/io/sentry/ObjectReader.java
```diff
@@ -0,0 +1,101 @@
...
+import java.io.IOException;
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
+import java.util.TimeZone;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public interface ObjectReader extends Closeable {
+  static @Nullable Date dateOrNull(@Nullable String dateString, final @NotNull ILogger logger) {
```
markushi, line 14:
```suggestion
  static @Nullable Date dateOrNull(final @Nullable String dateString, final @NotNull ILogger logger) {
```

sentry/src/main/java/io/sentry/util/MapObjectWriter.java
```diff
@@ -151,6 +156,12 @@ public MapObjectWriter value(final @Nullable String value) throws IOException {
     return this;
   }
 
+  @Override
+  public ObjectWriter jsonValue(@Nullable String value) throws IOException {
```
markushi, line 160: What's the reason for adding this?

sentry/src/main/java/io/sentry/util/MapObjectReader.java
```diff
@@ -0,0 +1,350 @@
...
+import java.util.Date;
+import java.util.Deque;
+import java.util.List;
+import java.util.Map;
+import java.util.TimeZone;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+@SuppressWarnings("unchecked")
+public final class MapObjectReader implements ObjectReader {
```
markushi, line 20: there are quite a few local variables to add the `final` keyword to

markushi, line 20: You could add a simple test where you go full circle: e.g. serialize a context to a map, and the deserialize it back and see if it matches

markushi (approved):
LGTM, nice!

sentry/src/main/java/io/sentry/util/MapObjectReader.java
romtsn: yeah I was wrapping it up yday https://github.com/getsentry/sentry-java/pull/3215/commits/1e76fc7499e8a1f4747c67c3aed4d1547c83759e

sentry/src/main/java/io/sentry/util/MapObjectWriter.java
romtsn: sorry this sneaked in from the target branch, I've updated it now to remove from this PR

sentry/src/main/java/io/sentry/util/MapObjectReader.java
romtsn: added in https://github.com/getsentry/sentry-java/pull/3215/commits/86baf7fc968a8ccc7dc7e8bdce22bd37e60bf8a6

PR #3216, opened 2024-02-21
Fix old profiles deletion on SDK init

stefanosiano:
## :scroll: Description
profiles deleted on init now need to be at least 5 minutes older than the class creation time to account for app start profiles

App start profiles could have a lower timestamp than the class creation, resulting in the SDK deleting the app start profile during init

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 ### Fixes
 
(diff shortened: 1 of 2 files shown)


markushi (approved):
LGTM!

Issue #3168, opened 2024-01-26
Reducing writes performed by the PersistingScopeObserver


sam-step:
### Problem Statement

Hello, I’m looking into CPU usage of our Android React Native app and noticed we’re spending an unexpectedly large amount of time converting breadcrumb records to JSON and persisting them.

From what I can tell, the [PersistingScopeObserver](https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/cache/PersistingScopeObserver.java) will delete and re-write the entire breadcrumb cache (by default up to 100 records) of the current Scope any time a new breadcrumb is added.

In some flows where we’re logging a fair amount (and therefore creating many breadcrumbs), these add up and noticeably affect performance.

My understanding is this writing of breadcrumb entries to disk is just for [ANR detection](https://github.com/getsentry/sentry-java/pull/2697) and, today, there is no way to disable this behavior without disabling all caching, which may affect other features.

Assuming my understanding is correct, is there an appetite to make this behavior configurable (supporting batching writes, or even disabling writes, etc)? For our usecase, this is an instance where we’re happy to lose a bit of observability to retain performance.

romtsn:
hi @sam-step that's a valid request, this logic definitely needs some improvement. One workaround you could for now is to just remove it from the scope observer list as follows:

```kotlin
SentryAndroid.init(context) { options ->
        options.scopeObservers.removeAll { it is PersistingScopeObserver }

        // if you still wanna collect other data except breadcrumbs for ANR events, you can add it back but 
        // void the `setBreadcrumbs` method
        val delegate = PersistingScopeObserver(options)
        options.addScopeObserver(object : IScopeObserver by delegate {
            override fun setBreadcrumbs(breadcrumbs: MutableCollection<Breadcrumb>) {
                TODO("drop breadcrumbs you are not interested in and call the line below, or just don't do anything to drop all breadcrumbs")
                // delegate.setBreadcrumbs(breadcrumbs)
            }
        })
}
```

markushi:
We see a few options on how we could improve the performance:
* Instead of writing to disk on every breadcrumb-add, only sync to disk every X seconds (potential loss of data)
* Think about a append-able file format, just like our envelopes, to only perform small writes (the newly added breadcrumb data)
* Use other signal to determine when to write to disk and when not (system broadcasts, app goes in background, events being sent, etc..)

@sam-step does any of the above sound reasonable to you? It would be great to get some feedback!

[...]

Issue #547, opened 2018-01-17
Retain breadcrumbs even if Android kill app process


piotrek-chmielowski:
I have a following scenario:
1. User uses my app
2. User clicks home button so the app goes foreground
3. Android kills app's process in order to free memory
4. User opens the app again, Android restores its state
5. Crash happens

In this scenario I have full stack trace and error info on Sentry, however I do not know what happened before app was killed and reopened because I have only breadcrumbs from the beginning of current app process. 

Is it possible to force breadcrumbs to be retained during the process death?

bretthoerner:
Good idea, I honestly never considered it. We should probably attempt to store all context information on Android.

romtsn:
should be pretty straightforward to implement, once ANRv2 lands

dharmita1990:
Any update on this?

markushi:
Ideally we can just restore the breadcrumbs on app-start

PR #3218, opened 2024-02-22
Add enableScopePersistence option

romtsn:
## :scroll: Description

For now PersistingScopeObserver is used only for ANRv2, but it's generic enough that we can use it for other features (e.g. to address #547), hence we're introducing a new flag to disable scope persistence separately. 

We should also think how to improve its performance see #3168 

Part of https://github.com/getsentry/sentry-unity/issues/1555

## :pencil: Checklist

### DIFF
--- a/sentry/src/main/java/io/sentry/SentryOptions.java
+++ b/sentry/src/main/java/io/sentry/SentryOptions.java
@@ -448,6 +448,9 @@ public class SentryOptions {
 
   private @Nullable String spotlightConnectionUrl;
 
+  /** Whether to enable scope persistence so the scope values are preserved if the process dies */
+  private boolean enableScopePersistence = true;
+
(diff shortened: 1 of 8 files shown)


sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -2313,6 +2316,14 @@ public void setEnableSpotlight(final boolean enableSpotlight) {
     this.enableSpotlight = enableSpotlight;
   }
 
+  public boolean isEnableScopePersistence() {
+    return enableScopePersistence;
+  }
+
+  public void setEnableScopePersistence(boolean enableScopePersistence) {
```
markushi, line 2323:
```suggestion
  public void setEnableScopePersistence(final boolean enableScopePersistence) {
```

PR #3221, opened 2024-02-22
Ensure performance measurement collection is not taken too frequently

stefanosiano:
## :scroll: Description
performance measurement collection now waits at least 10ms between collections
compiled regex in AndroidCpuCollector to parse the cpu file, to improve speed

Some UI tests are failing. Also, measurements taken too quickly (1ms) caused issues when the backend parsed the data. It was resolved in the backend, but it's something we should avoid in the first place

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixes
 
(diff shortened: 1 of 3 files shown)


stefanosiano:
Do we need a changelog entry for this change?

PR #3271, opened 2024-03-19
ref: Remove unused Metrics code

philipphofmann:
Remove the timestamp and SentryMetric because they aren't used.

@markushi, I think it's OK to remove that code. Please have a look.

#skip-changelog


markushi:
cc @bitsandfoxes I guess it's not [used anywhere in .NET](https://github.com/getsentry/sentry-dotnet/blob/277f865bafd63423a476a8672326164cf6792c8e/src/Sentry/Protocol/Metrics/Metric.cs#L47) as well?

markushi (approved):
LGTM!

bitsandfoxes:
> cc @bitsandfoxes I guess it's not [used anywhere in .NET](https://github.com/getsentry/sentry-dotnet/blob/277f865bafd63423a476a8672326164cf6792c8e/src/Sentry/Protocol/Metrics/Metric.cs#L47) as well?

Nope, it's fine. Thanks!

PR #3272, opened 2024-03-19
[SR] Add replay integration

romtsn:
_#skip-changelog_

## :scroll: Description

* Adds `ReplayIntegration` which takes care of persisting screenshots, making a video segment out of them and sending them as envelope
* Wires up sentry-android-replay module with sentry-android-core through compileOnly dependency and runtime checks
  * also hijacks the current session lifecycle logic to start/stop session replay
* Adds replayId to Scope, which later we should send in the envelope header to connect with errors and performance

## :pencil: Checklist

## :crystal_ball: Next steps
* add tests
* add replay options
* add buffer mode (to capture 30s before an error happened)

### DIFF
--- a/sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
+++ b/sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
@@ -1,0 +1,171 @@
+package io.sentry.android.replay
+
+import android.graphics.Bitmap
+import android.graphics.Bitmap.CompressFormat.JPEG
+import android.graphics.BitmapFactory
+import io.sentry.SentryLevel.DEBUG
+import io.sentry.SentryLevel.ERROR
+import io.sentry.SentryLevel.WARNING
+import io.sentry.SentryOptions
+import io.sentry.android.replay.video.MuxerConfig
+import io.sentry.android.replay.video.SimpleVideoEncoder
+import io.sentry.protocol.SentryId
+import io.sentry.util.FileUtils
+import java.io.Closeable
+import java.io.File
+
+internal class ReplayCache(
+    private val options: SentryOptions,
+    private val replayId: SentryId,
+    private val recorderConfig: ScreenshotRecorderConfig
+) : Closeable {
+
+    private val encoderLock = Any()
+    private var encoder: SimpleVideoEncoder? = null
+
+    private val replayCacheDir: File? by lazy {
+        if (options.cacheDirPath.isNullOrEmpty()) {
+            options.logger.log(
+                WARNING,
+                "SentryOptions.cacheDirPath is not set, session replay is no-op"
+            )
+            null
+        } else {
+            File(options.cacheDirPath!!, "replay_$replayId").also { it.mkdirs() }
+        }
+    }
+
+    // TODO: maybe account for multi-threaded access
+    private val frames = mutableListOf<ReplayFrame>()
+
+    fun addFrame(bitmap: Bitmap, frameTimestamp: Long) {
+        if (replayCacheDir == null) {
+            return
+        }
+
+        val screenshot = File(replayCacheDir, "$frameTimestamp.jpg").also {
+            it.createNewFile()
+        }
+        screenshot.outputStream().use {
+            bitmap.compress(JPEG, 80, it)
+            it.flush()
+        }
+
+        val frame = ReplayFrame(screenshot, frameTimestamp)
+        frames += frame
+    }
+
+    fun createVideoOf(duration: Long, from: Long, segmentId: Int): GeneratedVideo? {
+        if (frames.isEmpty()) {
+            options.logger.log(
+                DEBUG,
+                "No captured frames, skipping generating a video segment"
+            )
+            return null
+        }
+
+        // TODO: reuse instance of encoder and just change file path to create a different muxer
+        val videoFile = File(replayCacheDir, "$segmentId.mp4")
+        encoder = synchronized(encoderLock) {
+            SimpleVideoEncoder(
+                options,
+                MuxerConfig(
+                    file = videoFile,
+                    recorderConfig = recorderConfig,
+                    frameRate = recorderConfig.frameRate.toFloat(),
+                    bitrate = 20 * 1000
+                )
+            )
+        }.also { it.start() }
+
+        val step = 1000 / recorderConfig.frameRate.toLong()
+        var frameCount = 0
+        var lastFrame: ReplayFrame = frames.first()
+        for (timestamp in from until (from + (duration)) step step) {
+            val iter = frames.iterator()
+            val frameCountBefore = frameCount
+            while (iter.hasNext()) {
+                val frame = iter.next()
+                if (frame.timestamp in (timestamp..timestamp + step)) {
+                    frameCount++
+                    encode(frame)
+                    lastFrame = frame
+                    break // we only support 1 frame per given interval
+                }
+            }
+
+            // if the frame count hasn't changed we just replicate the last known frame to respect
+            // the video duration.
+            if (frameCountBefore == frameCount) {
+                frameCount++
+                encode(lastFrame)
+            }
+        }
+
+        if (frameCount == 0) {
+            options.logger.log(
+                DEBUG,
+                "Generated a video with no frames, not capturing a replay segment"
+            )
+            deleteFile(videoFile)
+            return null
+        }
+
+        var videoDuration: Long
+        synchronized(encoderLock) {
+            encoder?.release()
+            videoDuration = encoder?.duration ?: 0
+            encoder = null
+        }
+
+        frames.removeAll {
+            if (it.timestamp < (from + duration)) {
+                deleteFile(it.screenshot)
+                return@removeAll true
+            }
+            return@removeAll false
+        }
+
+        return GeneratedVideo(videoFile, frameCount, videoDuration)
+    }
+
+    private fun encode(frame: ReplayFrame) {
+        val bitmap = BitmapFactory.decodeFile(frame.screenshot.absolutePath)
+        synchronized(encoderLock) {
+            encoder?.encode(bitmap)
+        }
+        bitmap.recycle()
+    }
+
+    private fun deleteFile(file: File) {
+        try {
+            if (!file.delete()) {
+                options.logger.log(ERROR, "Failed to delete replay frame: %s", file.absolutePath)
+            }
+        } catch (e: Throwable) {
+            options.logger.log(ERROR, e, "Failed to delete replay frame: %s", file.absolutePath)
+        }
+    }
+
+    fun cleanup() {
+        FileUtils.deleteRecursively(replayCacheDir)
+    }
+
+    override fun close() {
+        synchronized(encoderLock) {
+            encoder?.release()
+            encoder = null
+        }
+    }
+}
+
+internal data class ReplayFrame(
+    val screenshot: File,
+    val timestamp: Long
+)
+
+internal data class GeneratedVideo(
+    val video: File,
+    val frameCount: Int,
+    val duration: Long
+)
--- a/sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
+++ b/sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
@@ -13,6 +13,19 @@ import io.sentry.protocol.SentryId
 import java.io.Closeable
 import java.io.File
 
+/**
+ * A basic in-memory and disk cache for Session Replay frames. Frames are stored in order under the
+ * [SentryOptions.cacheDirPath] + [replayId] folder. The class is also capable of creating an mp4
+ * video segment out of the stored frames, provided start time and duration using the available
+ * on-device [android.media.MediaCodec].
+ *
+ * This class is not thread-safe, meaning, [addFrame] cannot be called concurrently with
+ * [createVideoOf], and they should be invoked from the same thread.
+ *
+ * @param options SentryOptions instance, used for logging and cacheDir
+ * @param replayId the current replay id, used for giving a unique name to the replay folder
+ * @param recorderConfig ScreenshotRecorderConfig, used for video resolution and frame-rate
+ */
 public class ReplayCache internal constructor(
     private val options: SentryOptions,
     private val replayId: SentryId,
@@ -54,6 +67,16 @@ public class ReplayCache internal constructor(
     // TODO: maybe account for multi-threaded access
     internal val frames = mutableListOf<ReplayFrame>()
 
+    /**
+     * Stores the current frame screenshot to in-memory cache as well as disk with [frameTimestamp]
+     * as filename. Uses [Bitmap.CompressFormat.JPEG] format with quality 80. The frames are stored
+     * under [replayCacheDir].
+     *
+     * This method is not thread-safe.
+     *
+     * @param bitmap the frame screenshot
+     * @param frameTimestamp the timestamp when the frame screenshot was taken
+     */
     internal fun addFrame(bitmap: Bitmap, frameTimestamp: Long) {
         if (replayCacheDir == null) {
             return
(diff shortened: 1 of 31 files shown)


romtsn:
marking as ready for review - I'll add remaining tests before merging into main as a separate PR

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
```diff
@@ -0,0 +1,167 @@
...
+        }
+        screenshot.outputStream().use {
+            bitmap.compress(JPEG, 80, it)
+            it.flush()
+        }
+
+        val frame = ReplayFrame(screenshot, frameTimestamp)
+        frames += frame
+    }
+
```
vaind, line 67: In Flutter, I think I could use an `addFrame()` method that accepts a file path as an argument.

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
```diff
@@ -0,0 +1,167 @@
...
+import io.sentry.SentryLevel.ERROR
+import io.sentry.SentryLevel.WARNING
+import io.sentry.SentryOptions
+import io.sentry.android.replay.video.MuxerConfig
+import io.sentry.android.replay.video.SimpleVideoEncoder
+import io.sentry.protocol.SentryId
+import java.io.Closeable
+import java.io.File
+
+internal class ReplayCache(
```
vaind, line 16: This looks like a good thing to expose to Flutter. My idea is 
- crate a `ReplayCache` from a Flutter kotlin binding code
- create screenshots in Flutter
- write bitmaps to the cache directory directly in Flutter
- notify ReplayCache by calling `addFrame(filename)`
- call `createVideoOf()` to get the video file path
- send the replay envelope from Flutter

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
```diff
@@ -0,0 +1,317 @@
...
+import java.util.concurrent.Executors
+import java.util.concurrent.ThreadFactory
+import java.util.concurrent.TimeUnit.MILLISECONDS
+import java.util.concurrent.atomic.AtomicBoolean
+import java.util.concurrent.atomic.AtomicInteger
+import java.util.concurrent.atomic.AtomicReference
+import kotlin.LazyThreadSafetyMode.NONE
+import kotlin.math.roundToInt
+
+class ReplayIntegration(
```
vaind, line 38: Alternatively to exposing ReplayCache to Flutter, it may be even better to expose the functionality in this class, but with custom actual `recorder` implementation in Flutter

romtsn: I think I'd opt for exposing `ReplayCache` and some methods of this class (like `captureReplay`) through `InternalSentrySdk`, wdyt?

then you'd soemthing like:
```kotlin
val cache = ReplayCache(options, replayId, ScreenshotRecorderConfig(width, height, frameRate, bitRate))
cache.addFrame(screenshotFilename)

val video = cache.createVideoOf(duration, timestamp, segmentId)
// either full envelope details
val replayData = mapOf("replay_event" to mapOf(...), "replay_recording" to mapOf(...), "video" to videoFilePath)
InternalSentrySDK.captureReplay(replayData)

// or just the params we need for replay, and we build the event on our side
val replayData = mapOf(
        video: File,
        currentReplayId: SentryId,
        segmentTimestamp: Date,
        segmentId: Int,
        frameCount: Int,
        duration: Long,
        replayType: ReplayType
)
InternalSentrySDK.captureReplay(replayData)
```

vaind: looks good to me with either API

romtsn: probably the former is better cause you have full control of what to send then, and wouldn't need to update the native sdk everytime you want to change/add something

vaind: I understand. As for the beforeSend callback. Does the event already have the binary data loaded or is it still a file path in the event?

romtsn: still a file path, we load it later here https://github.com/getsentry/sentry-java/blob/79151e9a7103dd1f1f945648a22c05d1554e4fd5/sentry/src/main/java/io/sentry/SentryEnvelopeItem.java#L377-L385

vaind: that sounds great then!

romtsn: @vaind lemme know if this works: https://github.com/getsentry/sentry-java/pull/3272/commits/0f4e718fc20c2001efed3ac4ec1f1947f4366a46

romtsn:
integration tests are failing probably due to a running replay recorder, I guess I'll disable it in the following PR where we introduced sample rates

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
```diff
@@ -0,0 +1,181 @@
...
+        }
+
+        // TODO: reuse instance of encoder and just change file path to create a different muxer
+        encoder = synchronized(encoderLock) { encoderCreator(videoFile) }
+
+        val step = 1000 / recorderConfig.frameRate.toLong()
+        var frameCount = 0
+        var lastFrame: ReplayFrame = frames.first()
+        for (timestamp in from until (from + (duration)) step step) {
+            val iter = frames.iterator()
```
markushi, line 99: this could definitely cause `ConcurrentModificationException` issues in case a frame is added in the meantime

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
```diff
@@ -0,0 +1,181 @@
...
+                    break // we only support 1 frame per given interval
+                }
+            }
+
+            // if the frame count hasn't changed we just replicate the last known frame to respect
+            // the video duration.
+            if (frameCountBefore == frameCount) {
+                frameCount++
+                encode(lastFrame)
+            }
```
markushi, line 116: I guess this could be simplified a bit
1. iterate over frames and store matching frame as `lastFrame`
2. exit early (assuming frames are in order)
3. always encode `lastFrame` (either a fresh one, or the one from before) 

```suggestion
            while (iter.hasNext()) {
                val frame = iter.next()
                if (frame.timestamp in (timestamp..timestamp + step)) {
                    lastFrame = frame
                    break // we only support 1 frame per given interval
                }
                // assuming frames are in order, if out of bounds exit early
                if (frame.timestamp < timestamp + step) {
                    break
                }
            }
            frameCount++
            encode(lastFrame)
```

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
```diff
@@ -0,0 +1,181 @@
...
+                return@removeAll true
+            }
+            return@removeAll false
+        }
+
+        return GeneratedVideo(videoFile, frameCount, videoDuration)
+    }
+
+    private fun encode(frame: ReplayFrame) {
+        val bitmap = BitmapFactory.decodeFile(frame.screenshot.absolutePath)
```
markushi, line 147: since this is performing file IO, it probably makes sense to guard this via try/catch. e.g. for disk errors or file corruptions

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
```diff
@@ -0,0 +1,317 @@
...
+        } else {
+            val screenBounds = Point()
+            @Suppress("DEPRECATION")
+            wm.defaultDisplay.getRealSize(screenBounds)
+            Rect(0, 0, screenBounds.x, screenBounds.y)
+        }
+    }
+
+    private val aspectRatio by lazy(NONE) {
+        screenBounds.bottom.toFloat() / screenBounds.right.toFloat()
```
markushi, line 76: nit, would be semantically more correct:
```suggestion
        screenBounds.height.toFloat() / screenBounds.width.toFloat()
```

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
```diff
@@ -0,0 +1,317 @@
...
+
+        if (isRecording.getAndSet(true)) {
+            options.logger.log(
+                DEBUG,
+                "Session replay is already being recorded, not starting a new one"
+            )
+            return
+        }
+
+        currentSegment.set(0)
```
markushi, line 121: Should all of this only happen if hub != null?

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
```diff
@@ -0,0 +1,317 @@
...
+        // TODO: finalize old recording if there's some left on disk and send it using the replayId from persisted scope (e.g. for ANRs)
+    }
+
+    fun resume() {
+        segmentTimestamp.set(DateUtils.getCurrentDateTime())
+        recorder?.resume()
+    }
+
+    fun pause() {
+        val now = dateProvider.currentTimeMillis
```
markushi, line 137: I guess we need to align the date provider usage to have the same source of truth, e.g. `DateUtils.getCurrentDateTime()` vs . `DateUtils.getCurrentDateTime()`

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
```diff
@@ -0,0 +1,317 @@
...
+        // TODO: check for replaysSessionSampleRate and replaysOnErrorSampleRate
+
+        this.hub = hub
+        recorder = WindowRecorder(options, recorderConfig, this)
+        isEnabled.set(true)
+    }
+
+    fun isRecording() = isRecording.get()
+
+    fun start() {
```
markushi, line 104: We probably should have some overall lifecycle state to guard against random invocations. E.g. pause() is ignored unless the current state is `resumed`.

markushi (commented):
About halfway through, looking good, no blockers so far. Left a few comments which can be addressed as future improvements PRs as well.

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
vaind: looks good afaict without testing this out, which I'm not sure I'd know how to include in a flutter app without it being released.

Maybe it's worth adding a test case in this repo to check the workflow with image files? It would also prevent future regressions.

sentry/src/main/java/io/sentry/rrweb/RRWebVideoEvent.java
```diff
@@ -26,7 +26,7 @@ public final class RRWebVideoEvent extends RRWebEvent implements JsonUnknown, Js
   private @NotNull String tag;
   private int segmentId;
   private long size;
-  private int duration;
+  private long duration;
```
markushi, line 29: `nit`: What's the unit here? Maybe rename or add a comment.

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
romtsn: yeah, I'll add a test 👍 you can absolutely test it from a branch by using gradle included builds, that's how I test it now on a pocketcasts app

in `settings.gradle(.kts)`:
```
includeBuild("path/to/sentry-java") {
  dependencySubstitution {
    substitute(module("io.sentry:sentry")).using(project(":sentry"))
    substitute(module("io.sentry:sentry-android-core")).using(project(":sentry-android-core"))
    substitute(module("io.sentry:sentry-replay")).using(project(":sentry-android-replay"))
  }
}
```

you just have to add a fake dependency `io.sentry:sentry-replay:1.0.0` alongside `sentry-android:x.y.z` until it's shipped to maven

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
```diff
@@ -0,0 +1,181 @@
...
+        replayId: SentryId,
+        recorderConfig: ScreenshotRecorderConfig
+    ) : this(options, replayId, recorderConfig, encoderCreator = { videoFile ->
+        SimpleVideoEncoder(
+            options,
+            MuxerConfig(
+                file = videoFile,
+                recorderConfig = recorderConfig,
+                frameRate = recorderConfig.frameRate.toFloat(),
+                bitrate = 20 * 1000
```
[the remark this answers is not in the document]
romtsn: yep, I've changed that in the follow-up PR: https://github.com/getsentry/sentry-java/pull/3283/files#diff-14316f61844c51d22004ca969f6fc58f13c4292d055b190af0e7f5192b9a34c1R269

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
romtsn: good point!

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
romtsn: nice, I thought `Rect` doesn't have those 👍

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
romtsn: yeah, they are guarded by `if (isEnabled.get())` a few lines above, and that one we only set in `register` after the hub is initialized, so technically should not be the case

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
romtsn: > DateUtils.getCurrentDateTime() vs . DateUtils.getCurrentDateTime()

you mean `dateProvider.currentTimeMillis vs DateUtils.getCurrentDateTime()` right?

I'd love to align them, but the thing is that rrweb expects timestamp as `long` in milliseconds, and relay expects Event timestamp in ISO format, hence the divergence. I'll leave a TODO here though, to explore if we can align those 👍

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
romtsn: great idea! For now, we won't expose any public API to the customers, but once we do, I'll make sure to add the lifecycle state and checks, will leave a TODO as well

sentry/src/main/java/io/sentry/IScope.java
```diff
@@ -84,6 +85,23 @@ public interface IScope {
...
   void setScreen(final @Nullable String screen);
 
+  /**
+   * Returns the Scope's current replay_id, previously set by {@link IScope#setReplayId(SentryId)}
+   *
+   * @return the id of the current session replay
+   */
+  @ApiStatus.Internal
+  @Nullable
+  SentryId getReplayId();
```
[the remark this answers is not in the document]
romtsn: right, nice find, will change 👍

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
romtsn: for sure, I kind of didn't care about that because I use a single-thread executor for calling these methods.

Do you think it'd be enough to just leave comments and say this class is not thread-safe (because there's no way to enforce it really), or we still should make it thread-safe?

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
romtsn: I knew this logic looked like something to improve, thanks for this ❤️

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
romtsn: added test in https://github.com/getsentry/sentry-java/pull/3272/commits/18eb67ebf6a7ea188175fb819fe67c7a76d7a561

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
romtsn: yeah for sure I'll change that to be able to test it, will do that later when adding tests 👍

markushi (approved):

PR #3283, opened 2024-03-25
Add SessionReplayOptions

romtsn:
_#skip-changelog_

## :scroll: Description

* Introduce `SessionReplayOptions` which are nested under `sentryOptions._experimental` 
* Refactor ReplayIntegration to use the options

## :pencil: Checklist

### DIFF
--- a/sentry/src/main/java/io/sentry/ExperimentalOptions.java
+++ b/sentry/src/main/java/io/sentry/ExperimentalOptions.java
@@ -1,0 +1,16 @@
+package io.sentry;
+
+import org.jetbrains.annotations.NotNull;
+
+public final class ExperimentalOptions {
+  private @NotNull SessionReplayOptions sessionReplayOptions = new SessionReplayOptions();
+
+  @NotNull
+  public SessionReplayOptions getSessionReplayOptions() {
+    return sessionReplayOptions;
+  }
+
+  public void setSessionReplayOptions(final @NotNull SessionReplayOptions sessionReplayOptions) {
+    this.sessionReplayOptions = sessionReplayOptions;
+  }
+}
--- a/sentry/src/main/java/io/sentry/SessionReplayOptions.java
+++ b/sentry/src/main/java/io/sentry/SessionReplayOptions.java
@@ -1,0 +1,85 @@
+package io.sentry;
+
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.Nullable;
+
+public final class SessionReplayOptions {
+
+  /**
+   * Indicates the percentage in which the replay for the session will be created. Specifying 0
+   * means never, 1.0 means always. The value needs to be >= 0.0 and <= 1.0 The default is null
+   * (disabled).
+   */
+  private @Nullable Double sessionSampleRate;
+
+  /**
+   * Indicates the percentage in which a 30 seconds replay will be send with error events.
+   * Specifying 0 means never, 1.0 means always. The value needs to be >= 0.0 and <= 1.0. The
+   * default is null (disabled).
+   */
+  private @Nullable Double errorSampleRate;
+
+  /**
+   * Defines the quality of the session replay. Higher bit rates have better replay quality, but
+   * also affect the final payload size to transfer. The default value is 20kbps;
+   */
+  private int bitRate = 20_000;
+
+  /**
+   * Number of frames per second of the replay. The bigger the number, the more accurate the replay
+   * will be, but also more data to transfer and more CPU load.
+   */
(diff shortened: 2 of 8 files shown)


sentry/src/main/java/io/sentry/SessionReplayOptions.java
```diff
@@ -0,0 +1,85 @@
+package io.sentry;
+
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.Nullable;
+
+public final class SessionReplayOptions {
```
romtsn, line 6: @brustolin I've seen you used `SentryReplayOptions` instead of `SessionReplayOptions`, I thought of skipping the "Sentry" part, since it's anyway a nested object under `sentryOptions._experimental` already. But let me know if you want me to change that, I'm fine with either

brustolin: the class name for objc can collide with any framework in the app. That's why we need to have the prefix everywhere for the class name. And SentrySessionReplayOptions looks too big IMO.

romtsn: ah makes sense, I can change that, no problem 👍

sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
```diff
@@ -100,6 +100,10 @@ final class ManifestMetadataReader {
 
   static final String ENABLE_APP_START_PROFILING = "io.sentry.profiling.enable-app-start";
 
+  static final String REPLAYS_SESSION_SAMPLE_RATE = "io.sentry.replays.session-sample-rate";
```
markushi, line 103: Would it make sense to have this more consistent with other options, e.g. `"io.sentry.traces.profiling.sample-rate"`
```suggestion
  static final String REPLAYS_SESSION_SAMPLE_RATE = "io.sentry.session-replay.sample-rate";
```

sentry/src/main/java/io/sentry/ExperimentalOptions.java
```diff
@@ -0,0 +1,16 @@
+package io.sentry;
+
+import org.jetbrains.annotations.NotNull;
+
+public final class ExperimentalOptions {
```
markushi, line 5: To be honest, I wouldn't add this extra class. Once the feature is stable you'd need to move this over to `Sentry(Android)Options` causing a breaking change. Wouldn't `@Experimental` be enough? At least that's what I did for Metrics 😅

markushi (approved):
Looking good, just not super happy with your variable naming 😄

sentry/src/main/java/io/sentry/ExperimentalOptions.java
romtsn: yeah, the problem with the `@Experimental` annotation is that it doesn't really highlight it on the consumer's side (i.e. when you use it, the IDE does not highlight that this is experimental), and not everyone uses IDEs that highlight such things. So in order to know about it being experimental we'd have to actually look at the source code.

We wanted to stick to JS pattern which has `experimental` suboptions I think. Alternatively, we could just name the options `ExperimentalSentryReplayOptions` or something like that.

I think it's totally fine to make a breaking change for this, then the users will know that the feature was promoted to stable. We actually want to have `_experimental` only until it's GA'd I believe, and then promote it to SentryOptions, do you think that makes it more plausible?

sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
romtsn: hm, I actually tried to match `io.sentry.traces.sample-rate` where `traces = replays` and we have two types of sample rates for session replay. Does this make sense?

How would you name the other sampler rate in your case? `io.sentry.error-replay.sample-rate` doesn't look good tbh 😅

sentry/src/main/java/io/sentry/ExperimentalOptions.java
markushi: 👍 Fine for me, but please add a comment, so it's clear why this is different than usual.

sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
romtsn: yeah, js calls it e.g. `replaysOnErrorSampleRate`, but I'm good with your suggestion, and weill change it to session-replay 👍

PR #3327, opened 2024-04-05
[SR] Fix crashing MediaCodec and use density to determine recording resolution

romtsn:
_#skip-changelog_

## :scroll: Description

* Moves away from using a static height, but rather determine the recording size based on density (use 1x=mdpi for now, which means 2280x1080 will translate to something like 840x400)
* Fixes crashing MediaCodec on some devices, because we were not rounding height and width so they're divisible by 16 (pixel block size, which codecs work on)
* Increases default bitrate because 20k bps is of very low quality (we'll introduce different quality levels in the future, so it's easier for customers to chose from low|medium|high rather than picking arbitrary values)
* Fixes out-of-sync viewHierarchy in some cases by setting it to null immediately so it's not used for a later (maybe irrelevant) screenshot

## :pencil: Checklist

### DIFF
--- a/sentry-android-replay/src/main/java/io/sentry/android/replay/WindowRecorder.kt
+++ b/sentry-android-replay/src/main/java/io/sentry/android/replay/WindowRecorder.kt
@@ -55,11 +55,6 @@ internal class WindowRecorder(
             return
         }
 
-//    val (height, width) = (wm.currentWindowMetrics.bounds.bottom /
-//        context.resources.displayMetrics.density).roundToInt() to
-//        (wm.currentWindowMetrics.bounds.right /
-//            context.resources.displayMetrics.density).roundToInt()
-
         recorder = ScreenshotRecorder(recorderConfig, options, screenshotRecorderCallback)
         rootViewsSpy.listeners += onRootViewsChangedListener
         capturingTask = capturer.scheduleAtFixedRateSafely(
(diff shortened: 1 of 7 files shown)


sentry-android-replay/src/main/java/io/sentry/android/replay/WindowRecorder.kt
```diff
@@ -55,11 +55,6 @@ internal class WindowRecorder(
             return
         }
 
-//    val (height, width) = (wm.currentWindowMetrics.bounds.bottom /
```
markushi, line 58: I guess this can be safely removed

sentry/src/main/java/io/sentry/SentryReplayOptions.java
```diff
@@ -24,7 +24,7 @@ public final class SentryReplayOptions {
    * Defines the quality of the session replay. Higher bit rates have better replay quality, but
    * also affect the final payload size to transfer, defaults to 20kbps.
```
markushi, line 25:
```suggestion
   * also affect the final payload size to transfer, defaults to 100kbps.
```

markushi (approved):

sentry-android-replay/src/main/java/io/sentry/android/replay/WindowRecorder.kt
romtsn: yeah red lines means these have been removed, haha ;)

PR #3335, opened 2024-04-08
[SR] Handle orientation change

romtsn:
## :scroll: Description

* Observe device configuration changes to stop and restart the replay with new config
  * This is only support in full `session` mode for now, to support this in `buffer` mode we'd have to split the buffered video into multiple segments (as many as the number of orientation changes), but this will come later 
* Fix bug where we were not tagging events with `replayId` in `session` mode

### DIFF
--- a/sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
+++ b/sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
@@ -1,6 +1,8 @@
 package io.sentry.android.replay
 
+import android.content.ComponentCallbacks
 import android.content.Context
+import android.content.res.Configuration
 import android.graphics.Bitmap
 import android.os.Build
 import io.sentry.DateUtils
@@ -41,7 +43,7 @@ import kotlin.LazyThreadSafetyMode.NONE
 class ReplayIntegration(
     private val context: Context,
     private val dateProvider: ICurrentDateProvider
-) : Integration, Closeable, ScreenshotRecorderCallback, ReplayController {
+) : Integration, Closeable, ScreenshotRecorderCallback, ReplayController, ComponentCallbacks {
 
     internal companion object {
         private const val TAG = "ReplayIntegration"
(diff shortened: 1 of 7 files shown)


sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
```diff
@@ -388,10 +415,46 @@ class ReplayIntegration(
...
         stop()
         replayExecutor.gracefullyShutdown(options)
     }
 
+    override fun onConfigurationChanged(newConfig: Configuration) {
+        if (!isEnabled.get()) {
+            return
+        }
+
+        recorder?.stopRecording()
```
markushi, line 431: I'm wondering if `onConfigurationChanged` could be triggered without the width/height actually changing? Would it make sense to guard this the following way ~ `if (oldConfig == newConfig) { return }`

markushi (approved):

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
romtsn: yeah it can be triggered in the following cases 
image

we're mostly interested in the first two, but I guess it's fine to stop and restart the recording for the others as well, just to make sure we don't try to encode different configurations into the same video segment, wdyt? Otherwise, we'd probably have to check what actually changed in the configuration, and if it's not window/screen size, then we do nothing.

markushi: Yeah it makes sense to me to create a new segment, let's keep it simple and check later if too many config changes are causing an issue 👍  (e.g. keyboard show/hide, network provider changing, ...)

romtsn: sounds good! I guess we'd need some kind of internal tracking for that, but that's for later

PR #3337, opened 2024-04-09
[SR] More gates

romtsn:
## :scroll: Description

* Just adding more gates for the methods that shouldn't be called until we introduce replay lifecycle state
* Changed RootViewsSpy so that whenever we add a new listener to the list, we notify it immediately about any existing root views. This way we can ensure in the future if `start()` is called manually in the middle of app flow, we get the current root views to track


markushi (approved):
LGTM!

PR #3339, opened 2024-04-09
[SR] Session Replay

romtsn:
Session Replay for Android (beta)

Related to:
* https://github.com/getsentry/sentry/issues/70065


sentry-android-replay/src/main/java/io/sentry/android/replay/capture/BufferCaptureStrategy.kt
```diff
@@ -0,0 +1,174 @@
...
+            if (segment is ReplaySegment.Created) {
+                segment.capture(hub, hint ?: Hint())
+
+                // we only want to increment segment_id in the case of success, but currentSegment
+                // might be irrelevant since we changed strategies, so in the callback we increment
+                // it on the new strategy already
+                onSegmentSent()
+            }
+        }
+    }
```
vaind, line 95: Is there a plan to use eventId here? AFAICT, event-replay mapping is done via trace context rather than any field. The argument here is used just for logging

stefanosiano (approved):

PR #3357, opened 2024-04-15
[SR] Introduce capture strategies and handle orientation change for buffer

romtsn:
_#skip-changelog_

## :scroll: Description

* Introduce different capture strategies (Buffer and Session) to simplify their handling and not rely on if-else check everywhere. ReplayIntegration is very lean now used for glueing everything together
  * BaseCaptureStrategy contains common pieces for both
* Handle orientation change for  `buffer` mode - we capture segments whenever orientation change happens and keep them in memory until an error happens. We also rotate already captured segments when necessary
* Added one small thing for `session` mode - if a fatal crash happens, we try to capture the current segment to send as complete replay as possible


sentry-android-replay/src/main/java/io/sentry/android/replay/capture/SessionCaptureStrategy.kt
```diff
@@ -0,0 +1,144 @@
...
+                        options.experimental.sessionReplay.sessionSegmentDuration,
+                        currentSegmentTimestamp,
+                        replayId,
+                        segmentId,
+                        height,
+                        width
+                    )
+                if (segment is ReplaySegment.Created) {
+                    segment.capture(hub)
+                    currentSegment.getAndIncrement()
```
markushi, line 101: in case segment creation fails, would you need to increment `currentSegment.getAndIncrement()` as well?

markushi (approved):

sentry-android-replay/src/main/java/io/sentry/android/replay/capture/SessionCaptureStrategy.kt
romtsn: no, I think we'd want to try and capture the next time with the same segmentId to avoid gaps. That's a good  point for future though, as we could create an rrweb_event indicating the failure and reflect the gap somehow on the playback. I'll add this to ideas list 👍

PR #3380, opened 2024-04-22
Capture network requests

romtsn:
## :scroll: Description

* Adds RRWebSpanEvent and transforms our breadcrumbs from the scope to these, filtered by the replay segment timespan
* Had to add `start_timestamp`/`end_timestamp` to the http breadcrumbs data bag, because we expect "spans" with start and end time on the frontend side for network requests
* Small fix for MapObjectReader where `nextObjectOrNull` would not remove the last entry from stack
* Adds missing rrweb events to ReplayRecordingTest

image


sentry-android-replay/src/main/java/io/sentry/android/replay/capture/BaseCaptureStrategy.kt
```diff
@@ -180,7 +191,32 @@ internal abstract class BaseCaptureStrategy(
...
+                            if (!breadcrumb.isValidForRRWebSpan()) {
+                                return@forEach
+                            }
+                            recordingPayload += RRWebSpanEvent().apply {
+                                timestamp = breadcrumb.timestamp.time
+                                op = "resource.http"
+                                description = breadcrumb.data["url"] as String
+                                startTimestamp =
+                                    (breadcrumb.data["start_timestamp"] as Long) / 1000.0
+                                this.endTimestamp =
```
markushi, line 204: would it make sense to factor this out into a separate method to avoid `endTimestamp` vs `this.endTimestamp`?

sentry-android-replay/src/main/java/io/sentry/android/replay/capture/BaseCaptureStrategy.kt
```diff
@@ -280,4 +317,15 @@ internal abstract class BaseCaptureStrategy(
...
     }
+
+    private fun Breadcrumb.isValidForRRWebSpan(): Boolean {
+        return !(data["url"] as? String).isNullOrEmpty() &&
+            "start_timestamp" in data &&
+            "end_timestamp" in data
+    }
+
+    private fun String.snakeToCamelCase(): String {
+        val pattern = "_[a-z]".toRegex()
```
markushi, line 328: IIRC compiling a regex pattern is expensive, maybe worth moving into a static object const?

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpEvent.kt
```diff
@@ -150,6 +153,8 @@ internal class SentryOkHttpEvent(private val hub: IHub, private val request: Req
         hint.set(TypeCheckHint.OKHTTP_REQUEST, request)
         response?.let { hint.set(TypeCheckHint.OKHTTP_RESPONSE, it) }
 
+        // needs this as unix timestamp for rrweb
+        breadcrumb.setData("end_timestamp", CurrentDateProvider.getInstance().currentTimeMillis)
```
markushi, line 157: Maybe makes sense to have `"start_timestamp"` and `"end_timestamp"` as constants, similar to `SpanDataConvention`?

markushi (approved):
Nice, looking good! Left a few minor comments.

PR #3382, opened 2024-04-23
Fix Frame measurements in app start transactions

stefanosiano:
## :scroll: Description
performance collectors are called independently from profiling (now transactions get frames measurements)
`TimeSpan.getStartTimestamp` is now a `SentryNanotimeDate`, to fix `FrameMetricsCollector` on app start transactions

`FrameMetricsCollector` checks the start and end timestamp with a difference to a common date. The issue is that using a `SentryLongDate` and a `SentryNanotimeDate` gives different results. We always use `SentryNanotimeDate` on Android, except for app start timestamps, which are then propagated to app start transaction, TTID and TTFD.
Also, `SentryTracer` was not added to the `PerformaceCollectors` when profiling was disabled, due to previous constraints

Added UI test

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
@@ -560,7 +560,7 @@ public final class ActivityLifecycleIntegration
       if (ttfdSpan != null && ttfdSpan.isFinished()) {
         ttfdSpan.updateEndDate(endDate);
         // If the ttfd span was finished before the first frame we adjust the measurement, too
-        ttidSpan.setMeasurement(
+        ttfdSpan.setMeasurement(
             MeasurementValue.KEY_TIME_TO_FULL_DISPLAY, durationMillis, MILLISECOND);
       }
       finishSpan(ttidSpan, endDate);
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
@@ -560,7 +560,7 @@ public final class ActivityLifecycleIntegration
       if (ttfdSpan != null && ttfdSpan.isFinished()) {
         ttfdSpan.updateEndDate(endDate);
         // If the ttfd span was finished before the first frame we adjust the measurement, too
-        ttfdSpan.setMeasurement(
+        ttidSpan.setMeasurement(
             MeasurementValue.KEY_TIME_TO_FULL_DISPLAY, durationMillis, MILLISECOND);
       }
       finishSpan(ttidSpan, endDate);
--- a/sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
@@ -144,7 +144,7 @@ public class SpanFrameMetricsCollector
         return;
       }
       // Note: The comparison between two values obtained by realNanos() works only if both are the
-      // same kind of dates (both SentryNanotimeDate or both SentryLongDate)
+      // same kind of dates (both are SentryNanotimeDate or both SentryLongDate)
       final long spanEndNanos = realNanos(spanFinishDate);
 
       final @NotNull SentryFrameMetrics frameMetrics = new SentryFrameMetrics();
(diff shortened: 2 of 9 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
```diff
@@ -135,11 +136,15 @@ private void captureFrameMetrics(@NotNull final ISpan span) {
         return;
       }
 
-      // ignore spans with no finish date
-      final @Nullable SentryDate spanFinishDate = span.getFinishDate();
+      // Ignore spans with no finish date, but SentryTracer is not finished when executing this
+      // callback, yet, so in that case we use the current timestamp.
+      final @Nullable SentryDate spanFinishDate =
+          span instanceof SentryTracer ? new SentryNanotimeDate() : span.getFinishDate();
```
stefanosiano, line 142: My only doubt is here. In case of idle transactions, the end date could be trimmed, causing the transaction to have more frames than reality (just until we move to span-only). But this actually matches what we already have in perfV1.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -560,7 +560,7 @@ private void onFirstFrameDrawn(final @Nullable ISpan ttfdSpan, final @Nullable I
       if (ttfdSpan != null && ttfdSpan.isFinished()) {
         ttfdSpan.updateEndDate(endDate);
         // If the ttfd span was finished before the first frame we adjust the measurement, too
-        ttidSpan.setMeasurement(
+        ttfdSpan.setMeasurement(
```
markushi, line 563: well spotted!

sentry-android-core/src/main/java/io/sentry/android/core/performance/TimeSpan.java
```diff
@@ -90,7 +91,8 @@ public long getStartTimestampMs() {
    */
   public @Nullable SentryDate getStartTimestamp() {
     if (hasStarted()) {
-      return new SentryLongDate(DateUtils.millisToNanos(getStartTimestampMs()));
+      return new SentryNanotimeDate(
+          DateUtils.nanosToDate(DateUtils.millisToNanos(getStartTimestampMs())), System.nanoTime());
```
markushi, line 95: `m`: since this is just a getter, it will return a different `SentryDate` every time it gets invoked. Would it make more sense to just provide `0` for the nanos?
```suggestion
          DateUtils.nanosToDate(DateUtils.millisToNanos(getStartTimestampMs())), 0);
```
If we need nano precision, we could store the `System.nanoTime()` within the `start() { }` block and then re-use it here later.

sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
markushi: Yeah at least for spans, the finish date is [always set](https://github.com/getsentry/sentry-java/blob/61981dc1039b766300d6e38e82ccacb439387cb7/sentry/src/main/java/io/sentry/Span.java#L200) before invoking the [span finish callback](https://github.com/getsentry/sentry-java/blob/61981dc1039b766300d6e38e82ccacb439387cb7/sentry/src/main/java/io/sentry/Span.java#L236), which itself [invokes this callback](https://github.com/getsentry/sentry-java/blob/305baf5bd419920f4b80ff6f0075c13dd42c2b94/sentry/src/main/java/io/sentry/SentryTracer.java#L440). Well spotted, thanks for fixing this! Yeah I guess we have no mechanism right now to deal with trimming.

markushi (approved):
Looking good, that's some nice fixes! Left on comment which needs clarification around `TimeSpan.getStartTimestamp()`, once that is sorted out, we're good to go.

sentry-android-core/src/main/java/io/sentry/android/core/performance/TimeSpan.java
stefanosiano: we need nano precision, as in the `FrameMetricsCollector` we call `date.diff`, which uses that value and compares it to the value of other dates, which are all `System.nanotime()`.
I'm setting it during `start()`

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
stefanosiano: I had to revert it, as it's not a simple enough change (the ttfd span is already finished and we can't set measurements). [Created an issue for this](https://github.com/getsentry/sentry-java/issues/3384)

PR #3398, opened 2024-04-30
Fix UncaughtExceptionHandlerIntegration Memory Leak

### LINKED ISSUE #3266: UncaughtExceptionHandlerIntegration leaks memory
### Integration

sentry

### Java Version

17

### Version

7.6.0

### Steps to Reproduce

Cannot be very specific because sentry SDK is being used by a plugin within a framework, and within that environment we cannot reproduce the issue unless we are receiving production traffic

### Expected Result

no memory leak

### Actual Result

the UncaughtExceptionHandlerIntegration appears to occupy the majority of the heap, eventually resulting in memory exhaustion and a crash. 

On auditing the code it appears that 
https://github.com/getsentry/sentry-java/blob/28c9a83af55b119517dbd2fcf74975fb5822907b/sentry/src/main/java/io/sentry/UncaughtExceptionHandlerIntegration.java#L68-L81
causes a stack to form, which prevents any of the exeption handlers from being GC'd and results in OOM. No idea why that register method is being called over and over again

image

we are planning to simply turn the UncaughtExceptionHandler off in the options to resolve this issue, but it's weird that the Sentry SDK apparently behaves this way by default

lbloder:
Upon registration check if an `UncaughtExceptionHandlerIntegration` is already set as `defaultUncaughtExceptionHandler`, if so, set its `defaultExceptionHandler` as the default of the new `UncaughtExceptionHandlerIntegration`. 

Fixes #3266

## :pencil: Checklist

### DIFF
--- a/sentry/src/main/java/io/sentry/UncaughtExceptionHandlerIntegration.java
+++ b/sentry/src/main/java/io/sentry/UncaughtExceptionHandlerIntegration.java
@@ -75,7 +75,14 @@ public final class UncaughtExceptionHandlerIntegration
                 "default UncaughtExceptionHandler class='"
                     + currentHandler.getClass().getName()
                     + "'");
-        defaultExceptionHandler = currentHandler;
+
+        if(currentHandler instanceof UncaughtExceptionHandlerIntegration) {
+          final UncaughtExceptionHandlerIntegration currentHandlerIntegration = (UncaughtExceptionHandlerIntegration) currentHandler;
+          defaultExceptionHandler = currentHandlerIntegration.defaultExceptionHandler;
+        } else {
+          defaultExceptionHandler = currentHandler;
+        }
+
       }
 
       threadAdapter.setDefaultUncaughtExceptionHandler(this);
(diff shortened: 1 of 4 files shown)


sentry/src/main/java/io/sentry/UncaughtExceptionHandlerIntegration.java
```diff
@@ -75,7 +75,16 @@ public final void register(final @NotNull IHub hub, final @NotNull SentryOptions
...
+
+        if (currentHandler instanceof UncaughtExceptionHandlerIntegration) {
+          final UncaughtExceptionHandlerIntegration currentHandlerIntegration =
+              (UncaughtExceptionHandlerIntegration) currentHandler;
+          defaultExceptionHandler = currentHandlerIntegration.defaultExceptionHandler;
+        } else {
+          defaultExceptionHandler = currentHandler;
+        }
+
+        //        defaultExceptionHandler = currentHandler;
```
romtsn, line 87:
```suggestion
```

romtsn (approved):
nice!

adinauer (approved):
Thanks for taking care of this!

PR #3403, opened 2024-05-02
Metrics clean

stefanosiano:
## :scroll: Description
- Remove `timestampMs` from `LocalMetricsAggregator`
- put `toStatsdType` in enum directly
- beforeMetric wrapped in try catch
- flush check if there are metrics before scheduling again

Just a small cleanup of the metrics code

## :pencil: Checklist

### DIFF
--- a/sentry/src/main/java/io/sentry/metrics/MetricType.java
+++ b/sentry/src/main/java/io/sentry/metrics/MetricType.java
@@ -1,12 +1,19 @@
 package io.sentry.metrics;
 
 import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
 
 /** The metric instrument type */
 @ApiStatus.Internal
 public enum MetricType {
-  Counter,
-  Gauge,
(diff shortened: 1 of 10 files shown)


sentry/src/main/java/io/sentry/metrics/MetricType.java
```diff
@@ -1,12 +1,19 @@
...
-  Set
+  Counter("c"),
+  Gauge("g"),
+  Distribution("d"),
+  Set("s");
+
+  final @NotNull String statsdCode;
+
+  MetricType(final @NotNull String statsdCode) {
+    this.statsdCode = statsdCode;
```
markushi, line 17: `nit`: I separated `MetricType` from Statsd on purpose: If we ever decide to use a different protocol, we would only need to refactor some helpers and transport specific code. But I'm still fine with doing it this way, as that's what other SDKs are doing as well.

markushi (approved):
LGTM, very nice!

PR #3406, opened 2024-05-03
[SR] Align breadcrumbs with frontend

romtsn:
_#skip-changelog_

## :scroll: Description

* Change categories and data according to [this table](https://www.notion.so/sentry/Mobile-breadcrumb-UI-next-steps-8b6566484fd149259ffc5a05ecac7fe6?pvs=4#270369d36b1a4ad2b285fdd17a612e83) (what frontend expects)
* Had to introduce a new breadcrumb for battery status changes (with throttling of 1 minute, analogous to what iOS [does](https://github.com/getsentry/sentry-cocoa/blob/8aec30ebd6b60262f8b935b5475741c92da38834/Sources/Sentry/SentrySystemEventBreadcrumbs.m#L104))


sentry/src/main/java/io/sentry/SentryLevel.java
```diff
@@ -18,7 +18,7 @@ public void serialize(final @NotNull ObjectWriter writer, final @NotNull ILogger
     writer.value(name().toLowerCase(Locale.ROOT));
   }
 
-  static final class Deserializer implements JsonDeserializer<SentryLevel> {
+  public static final class Deserializer implements JsonDeserializer<SentryLevel> {
```
markushi, line 21: would it make sense to mark it as `@ApiStatus.Internal` then?

sentry-android-core/src/main/java/io/sentry/android/core/SystemEventsBreadcrumbsIntegration.java
```diff
@@ -204,45 +207,69 @@ public void close() throws IOException {
...
-        for (String item : extras.keySet()) {
-          try {
-            @SuppressWarnings("deprecation")
-            Object value = extras.get(item);
-            if (value != null) {
-              newExtras.put(item, value.toString());
+      if (isBatteryChanged) {
+        final Float batteryLevel = DeviceInfoUtil.getBatteryLevel(intent, options);
+        if (batteryLevel != null) {
+          breadcrumb.setData("level", batteryLevel);
```
markushi, line 243: float is fine here right? The notion doc is not 100% clear about this.

markushi (approved):
LGTM!

sentry-android-replay/src/main/java/io/sentry/android/replay/capture/BaseCaptureStrategy.kt
```diff
@@ -201,12 +202,12 @@ internal abstract class BaseCaptureStrategy(
...
+                if (breadcrumb.timestamp.time >= segmentTimestamp.time &&
+                    breadcrumb.timestamp.time < endTimestamp.time
                 ) {
-                    // TODO: rework this later when aligned with iOS and frontend
                     var breadcrumbMessage: String? = null
-                    val breadcrumbCategory: String?
+                    var breadcrumbCategory: String? = null
+                    var breadcrumbLevel: SentryLevel? = null
                     val breadcrumbData = mutableMapOf<String, Any?>()
                     when {
```
vaind, line 212: @romtsn what do you think about extracting this to a class so we can override it from RN and Flutter to have custom mapping?

sentry/src/main/java/io/sentry/SentryLevel.java
romtsn: hm, we don't really do that for other Deserializers that are public, so I guess it's fine to keep them aligned?

PR #3411, opened 2024-05-07
FEAT: Add version of withSentryObservableEffect that has better inter…

designedbyz:
## :scroll: Description

This PR adds a version of `NavHostController.withSentryObservableEffect` that accepts a `SentryNavigationListener` as a parameter to support better navigation tracing in apps that make use of both Fragments and Compose. 

A challenge with the existing Navigation and Compose Navigation integrations is that, in the testing I've done, they seem to clobber each other. My guess as to why is because each listener will try to register itself as the integration, meaning that you only get one. Additionally, even if the instances didn't clobber each other, navigation history will end up separated between instances, as the Compose version will only have the nav destinations for Compose, whereas the fragment version will only have the destinations for fragments. This makes it hard to put traces back together for apps that make use of both fragments and compose navigation and navigate between the two.

Thankfully, the fix is pretty simple and is implemented here. I added a new version of `withSentryObservableEffect` that accepts a `SentryNavigationListener` as a parameter, which allows consumers of the SDK to pass the same version of the `SentryNavigationListener to `withSentryObservableEffect` when using the compose integration as they do to `navController.addOnDestinationChangedListener` when adding the integration for fragment navigation tracing. 

There is one possible downside to this approach. Specifically, the trace origin will show up as `navigation` for compose navigation instead of `compose`, but I personally don't see this as an issue. 

Tested locally; and we'll likely ship an internal version of this to prod at Maven before this is merged

## :pencil: Checklist

## :crystal_ball: Next steps

needs a review from the Sentry team

### DIFF
--- a/sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryNavigationIntegration.kt
+++ b/sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryNavigationIntegration.kt
@@ -48,30 +48,23 @@ internal class SentryLifecycleObserver(
  * A [DisposableEffect] that captures a [Breadcrumb] and starts an [ITransaction] and sends
  * them to Sentry for every navigation event when being attached to the respective [NavHostController].
  *
- * @param enableNavigationBreadcrumbs Whether the integration should capture breadcrumbs for
- * navigation events.
- * @param enableNavigationTracing Whether the integration should start a new [ITransaction]
- * with [SentryOptions.idleTimeout] for navigation events.
+ * @param navListener An instance of a [SentryNavigationListener] that is shared with other sentry integrations, like
+ * the fragment navigation integration. 
  */
 @Composable
 @NonRestartableComposable
-public fun NavHostController.withSentryObservableEffect(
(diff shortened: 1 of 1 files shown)


designedbyz:
Question: how do I get a review for this?

romtsn (approved):
hi @designedbyz sorry for the delays, the team was overloaded with other tasks.

I've taken a look at the PR and it makes a lot of sense to have this functionality in the SDK, so I'm just approving it. You'd also have to add a changelog entry to make CI green, you can take a look at the other entries to comply with the changelog format.

Ideally we'd also update Sentry docs in this [section](https://docs.sentry.io/platforms/android/integrations/jetpack-compose/#jetpack-compose-navigation) and mention how to handover the navigation listener between compose and the old view system, but we could take care of that ourselves 👍 

Thanks for your contribution!

romtsn:
@designedbyz sorry for abandoning this, I've reopened #4143 as I was unable to push to this branch and update it for some reason. We'll merge it and ship in version 8.2.0. Thanks for your contribution!

PR #3422, opened 2024-05-14
Adjust publications to publish gradle module metadata

romtsn:
* Packages and renames `.module` files into the final dist.zip, so they are ready to be published by craft
* Makes `sentry-compose-helper` a `compileOnly` dependency, so it doesn't get added to the .pom and .module files
  * This has a side-effect that we have to specify it explicitly as `implementation` for the modules that are using `sentry-compose`. Our customers are still fine though, as we embed the classes from the `-helper` module and it's published to maven like that

Relevant PR in craft: https://github.com/getsentry/craft/pull/535
Closes https://github.com/getsentry/sentry-java/issues/2233

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -1,10 +1,10 @@
 minVersion: 0.29.3
 changelogPolicy: auto
 targets:
-  - name: symbol-collector
-    includeNames: /libsentry(-android)?\.so/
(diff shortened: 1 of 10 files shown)


markushi (approved):
Nice!

Issue #3269, opened 2024-03-18
SentryFragmentLifecycleCallbacks creates long running spans for ViewPager2


markushi:
### Integration

sentry-android

### Build System

Gradle

### AGP Version

unknown

### Proguard

Enabled

### Version

unknown

### Steps to Reproduce

1. Use [ViewPager2](https://developer.android.com/jetpack/androidx/releases/viewpager2)
2. Enable performance and our [FragmentLifecycleIntegration](https://github.com/getsentry/sentry-java/blob/main/sentry-android-fragment/src/main/java/io/sentry/android/fragment/FragmentLifecycleIntegration.kt) 

### Expected Result

No span is created for non-visible fragments.

### Actual Result

A span is created, but due to [this](https://developer.android.com/jetpack/androidx/releases/viewpager2#1.0.0-alpha06):
> non-primary-item Fragments are capped at STARTED

the spans end up running into an timeout, causing extra long transaction durations.

We probably should be using [`getUserVisibleHint()`](https://developer.android.com/reference/androidx/fragment/app/Fragment.html#getUserVisibleHint()) or [`getMinimumMaxLifecycleState()`](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:fragment/fragment/src/main/java/androidx/fragment/app/Fragment.java;l=435-440;drc=82d7c26098cb0328538c839aabfba0df3dfff7f9) to determine Fragments which won't be activated.

┆Issue is synchronized with this [Jira Improvement](https://getsentry.atlassian.net/browse/FEEDBACK-2128) by [Unito](https://www.unito.io)

realkosty:
@markushi any updates on this?
I added a report from an enterprise customer in the [linked Jira ticket](https://getsentry.atlassian.net/browse/FEEDBACK-2128). They believe the same issue is affecting their TTID.

kahest:
cc @stefanosiano

stefanosiano:
we are going to take a look at this in the next week

realkosty:
Hi @stefanosiano, were you guys able to make any progress on this one?

stefanosiano:
Hi @realkosty 
I was having some problem, but now I'm finally able to reproduce it and working on it 👍

stefanosiano:
Options considered:
* [getUserVisibleHint()](https://developer.android.com/reference/androidx/fragment/app/Fragment.html#getUserVisibleHint()): it's deprecated and returns true for all fragments
[...]

PR #3424, opened 2024-05-15
Move fragment auto span finish to onFragmentStarted

stefanosiano:
## :scroll: Description
moved fragment auto span finish to onFragmentStarted instead of onFragmentResumed to account for ViewPager2 behaviour

Other options considered, but failed:
* [getUserVisibleHint()](https://developer.android.com/reference/androidx/fragment/app/Fragment.html#getUserVisibleHint()): it's deprecated and returns true for all fragments
* [mMaxState](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:fragment/fragment/src/main/java/androidx/fragment/app/Fragment.java;l=297): only accessible through reflection and returns the current max state achieved (in onFragmentStarted it will be STARTED, while in onFragmentResumed it will be RESUMED
* [isVisible](https://developer.android.com/reference/androidx/fragment/app/Fragment.html#isVisible()): returns true for all fragments

Fixes https://github.com/getsentry/sentry-java/issues/3269

Unit tests

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Move fragment auto span finish to onFragmentStarted ([#3424](https://github.com/getsentry/sentry-java/pull/3424))
+
 ## 7.9.0
 
 ### Features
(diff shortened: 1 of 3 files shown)


markushi (approved):
🚀

PR #3442, opened 2024-05-24
Add timestamp when a profile starts

stefanosiano:
## :scroll: Description
added start timestamp in ProfileStartData when a profile starts
added the timestamp in ProfilingTraceData and send it in the json payload

Example of [wrong profile](https://demo.sentry.io/profiling/profile/android/625533174daa4b7b964d8ac08b7ab75b/flamegraph/?colorCoding=by+system+vs+application+frame&fov=0%2C5%2C10353642496%2C13&query=&sorting=call+order&tid=8734&view=top+down) (before)
Example of [correct profile](https://sentry-sdks.sentry.io/profiling/profile/sentry-android/dc544fbc9c4b496a88ec4b62cfe0a3e0/flamegraph/?colorCoding=by+system+vs+application+frame&fov=0%2C0%2C2867905792%2C19&query=&sorting=call+order&tid=8998&view=top+down) (after)

Profiles are a not aligned to spans, especially when we change the transaction start timestamp. In that case, it is assumed that the profile and the transaction starts at the same time, which is wrong. I added a `timestamp` field set when the profile starts.

Unit tests

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ### Fixes
 
 - Fix faulty `span.frame_delay` calculation for early app start spans ([#3427](https://github.com/getsentry/sentry-java/pull/3427))
+- Add timestamp when a profile starts ([#3442](https://github.com/getsentry/sentry-java/pull/3442))
 
 ## 7.9.0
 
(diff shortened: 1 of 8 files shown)


stefanosiano:
It looks like there is something wrong, as changes are not reflected in the flamechart in Sentry. I'll keep it as a draft until further investigation.

stefanosiano:
frontend was updated, and now everything works as expected

markushi (approved):
LGTM!

Issue #3228, opened 2024-02-26
Move the onFinishCallback before the span is finished


stefanosiano:
### Description

https://github.com/getsentry/sentry-java/pull/3219#discussion_r1500496194

markushi:
Let's also keep our `SpanFrameMetricsCollector` as it's `onSpanFinished` callback requires the finish date to be already set.

markushi:
Given that we have a mixed behavior for txn/spans, let's consider this broken and fix it.

PR #3459, opened 2024-06-05
Move onFinishCallback before span or transaction is finished

stefanosiano:
## :scroll: Description
moved span and transaction finish callback before they are actually finished

Closes https://github.com/getsentry/sentry-java/issues/3228
This fixes and aligns the behaviour among spans and transactions.

Unit tests

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
@@ -131,26 +131,12 @@ public class SpanFrameMetricsCollector
   private void captureFrameMetrics(@NotNull final ISpan span) {
     // TODO lock still required?
     synchronized (lock) {
-      String x = "";
-      for (ISpan runningSpan : runningSpans) {
-        x += runningSpan.getOperation() + " - " + runningSpan.getDescription() + ", ";
-      }
-      System.out.println(
-          "SpanFrameMetricsCollector.onSpanFinished - "
-              + span.getOperation()
-              + " - "
-              + span.getDescription()
-              + " ---> \ncurrent spans: "
-              + x);
-
       boolean removed = runningSpans.remove(span);
-      System.out.println(removed);
       if (!removed) {
         return;
       }
 
       final @Nullable SentryDate spanFinishDate = span.getFinishDate();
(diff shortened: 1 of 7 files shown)


sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -200,6 +200,24 @@ public void finish(
...
+              oldCallback.execute(span);
+            }
+
+            // Let's call the finishCallback here, when the root span has a finished date but it's
+            // not finished, yet
+            final @Nullable TransactionFinishedCallback finishedCallback =
+                transactionOptions.getTransactionFinishedCallback();
+            if (finishedCallback != null) {
+              finishedCallback.execute(this);
+            }
```
stefanosiano, line 218: I just moved the finishedCallback here, but we could move everything else, too, if needed, like the performanceCollectors

sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
```diff
@@ -131,19 +131,34 @@ public void onSpanFinished(final @NotNull ISpan span) {
   private void captureFrameMetrics(@NotNull final ISpan span) {
     // TODO lock still required?
     synchronized (lock) {
+      String x = "";
+      for (ISpan runningSpan : runningSpans) {
+        x += runningSpan.getOperation() + " - " + runningSpan.getDescription() + ", ";
+      }
+      System.out.println(
```
markushi, line 138: I guess you still want to remove those right? 😅

sentry/src/main/java/io/sentry/Span.java
```diff
@@ -37,7 +37,9 @@ public final class Span implements ISpan {
 
   private final @NotNull IHub hub;
 
-  private final @NotNull AtomicBoolean finished = new AtomicBoolean(false);
+  private boolean finished = false;
```
markushi, line 40: `nit` this probably should be volatile, but then on the other hand we also don't do this for other span properties ¯\_(ツ)_/¯

markushi (approved):
Nice one!

sentry/src/main/java/io/sentry/Span.java
stefanosiano: let's keep `AtomicBoolean` and change it later if ever needed, as i don't think it really changes anything

PR #3466, opened 2024-06-07
[SR] Redaction fixes part 1

romtsn:
_#skip-changelog_

## :scroll: Description

Multiple improvements to redaction logic:
* Split up ViewHierarchyNodes by type: TextView, ImageView and Generic, because they all have different properties and with the help of sealed classes we can make easier type checks and inference
* We now do way less stuff on the main thread when collecting the view hierarchy - just capture necessary things and pass it to a bg thread, and I also adopted only necessary code for determining if a view is visible or not, without all the Accessibility shenanigans around it
* (probably the most complex thing here): we had to determine if a view is obscured by another one or not, because Android does not tell us that. I've tried to write kdoc for the respective methods and the algorithm, but let me know if I should walk you through it.
* We also support multi-line test and draw rectangles with the respective width/height per-line

Some screenshots before/after (notice on the 1st sscreesnhot there are white rectangles drawn on top of the bottom navigation bar, now this is fixed):

Google Chrome 2024-06-07 21 04 45

image


sentry-android-replay/src/main/java/io/sentry/android/replay/ScreenshotRecorder.kt
```diff
@@ -122,21 +126,42 @@ internal class ScreenshotRecorder(
...
-                                    val color = if (it.dominantColor == null) {
-                                        singlePixelBitmapCanvas.drawBitmap(bitmap, it.visibleRect, Rect(0, 0, 1, 1), null)
-                                        singlePixelBitmap.getPixel(0, 0)
-                                    } else {
-                                        it.dominantColor
+                            viewHierarchy.traverse { node ->
+                                if (node.shouldRedact && (node.width > 0 && node.height > 0)) {
+                                    node.visibleRect ?: return@traverse false
+
+                                    if (viewHierarchy.isObscured(node)) {
```
markushi, line 133: since you need to traverse every node here, this is a n^2 operation. Not saying that this is bad, but we should probably profile this at some point.

sentry-android-replay/src/main/java/io/sentry/android/replay/viewhierarchy/ViewHierarchyNode.kt
```diff
@@ -2,133 +2,291 @@ package io.sentry.android.replay.viewhierarchy
...
+                return@traverse false
+            }
+
+            // if the other node is not visible, or not important for content capture (empty container)
+            // or doesn't contain the node's visible rect, we can skip it
+            if (!otherNode.isVisible ||
+                !otherNode.isImportantForContentCapture ||
+                !otherNode.visibleRect.contains(node.visibleRect)
+            ) {
+                return@traverse false
```
markushi, line 126: nice optimization!

markushi (approved):
LGTM!

sentry-android-replay/src/main/java/io/sentry/android/replay/ScreenshotRecorder.kt
romtsn: yep, for sure. I don't think there's any other way though, at least not from the top of my head. I've tried to optimize it with DFS and skipping traversal of irrelevant nodes/subtrees, so n^2 is really the worst case here, and probably will never happen, because the likelihood of all the nodes' visible rects intersecting is very small

PR #3473, opened 2024-06-11
[SR] Support multi-touch gestures

romtsn:
_#skip-changelog_

## :scroll: Description

* Support multi-touch gestures by keeping track of pointerIds via a map of `pointerId to array[position]` 

There are also 2 irrelevant to this PR changes, but I've jsut made them as I discovered to not forget about them later:
* Debounce connectivity breadcrumbs when the state hasn't changed (`DefaultReplayBreadcrumbConverter`)
* Fix crash in case segmentTimestamp is not available (discovered in one of alpha customer's issues). In case it's not available we simply set the duration to 0 meaning that we will not capture the segment (`SessionCaptureStrategy`)

Example replay: https://sentry-sdks.sentry.io/replays/9aedc6321c9949e09d649d7757cd5b32/?project=5428559&query=&referrer=%2Freplays%2F%3AreplaySlug%2F&statsPeriod=1h&yAxis=count%28%29&t=0


brustolin (approved):
LGTM

sentry-android-replay/src/main/java/io/sentry/android/replay/capture/BaseCaptureStrategy.kt
```diff
@@ -295,48 +295,110 @@ internal abstract class BaseCaptureStrategy(
...
+                val pIndex = event.findPointerIndex(pId)
+
+                if (pIndex == -1) {
+                    // no data for this pointer
+                    return null
                 }
+
+                // finger lift up - remove the pointer from tracking
+                currentPositions[pId]?.clear()
+                currentPositions.remove(pId)
```
markushi, line 377: wouldn't it be enough to just remove the `pid` entry?
```suggestion
                currentPositions.remove(pId)
```

markushi (approved):
Nice one, LGTM!

PR #3480, opened 2024-06-14
fix(session): When capturing unhandled hybrid exception session should be ended

krystofwoldrich:
## :scroll: Description

This PR fixes the session handling for hybrid envelopes with unhandled exceptions. 

If hybrid exception is unhandled but doesn't crash the application the native `captureEnvelope` should create a new session.

If the hybrid exception is unhandled and crashes the application the native `captureEnvelope` should not create a new session.

- This is aligned with the iOS behaviour implemented in https://github.com/getsentry/sentry-cocoa/pull/3091

- fixes https://github.com/getsentry/sentry-react-native/issues/3614
- related iOS fix https://github.com/getsentry/sentry-cocoa/pull/4073
- related RN fix https://github.com/getsentry/sentry-react-native/pull/3900

rn sample app, unit tests

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
@@ -192,9 +192,9 @@ public final class InternalSentrySdk {
         final SentryEnvelopeItem sessionItem = SentryEnvelopeItem.fromSession(serializer, session);
         envelopeItems.add(sessionItem);
         deleteCurrentSessionFile(
-          options,
-          // should be sync if going to crash or already not a main thread
-          !maybeStartNewSession || !hub.getOptions().getMainThreadChecker().isMainThread());
+            options,
+            // should be sync if going to crash or already not a main thread
+            !maybeStartNewSession || !hub.getOptions().getMainThreadChecker().isMainThread());
         if (maybeStartNewSession) {
           hub.startSession();
         }
@@ -266,14 +266,20 @@ public final class InternalSentrySdk {
     spans.add(spanMap);
   }
 
-  private static void deleteCurrentSessionFile(final @NotNull SentryOptions options, boolean isSync) {
+  private static void deleteCurrentSessionFile(
+      final @NotNull SentryOptions options, boolean isSync) {
     if (!isSync) {
       try {
-        options.getExecutorService().submit(() -> {
-          deleteCurrentSessionFile(options);
-        });
+        options
+            .getExecutorService()
+            .submit(
+                () -> {
+                  deleteCurrentSessionFile(options);
+                });
       } catch (Throwable e) {
-        options.getLogger().log(WARNING, "Submission of deletion of the current session file rejected.", e);
+        options
+            .getLogger()
+            .log(WARNING, "Submission of deletion of the current session file rejected.", e);
       }
     } else {
       deleteCurrentSessionFile(options);
--- a/sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
@@ -150,8 +150,7 @@ public final class InternalSentrySdk {
    */
   @Nullable
   public static SentryId captureEnvelope(
-    final @NotNull byte[] envelopeData,
-    final boolean maybeStartNewSession) {
+      final @NotNull byte[] envelopeData, final boolean maybeStartNewSession) {
     final @NotNull IHub hub = HubAdapter.getInstance();
     final @NotNull SentryOptions options = hub.getOptions();
 
--- a/sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
@@ -14,7 +14,6 @@ import io.sentry.ILogger;
 import io.sentry.IScope;
 import io.sentry.ISerializer;
 import io.sentry.ObjectWriter;
-import io.sentry.Sentry;
 import io.sentry.SentryEnvelope;
 import io.sentry.SentryEnvelopeItem;
 import io.sentry.SentryEvent;
@@ -194,7 +193,7 @@ public final class InternalSentrySdk {
         envelopeItems.add(sessionItem);
         deleteCurrentSessionFile(options);
         if (maybeStartNewSession) {
-          Sentry.startSession();
+          hub.startSession();
         }
       }
 
--- a/sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
@@ -273,8 +273,8 @@ public final class InternalSentrySdk {
 
     if (!options.isEnableAutoSessionTracking()) {
       options
-        .getLogger()
-        .log(DEBUG, "Session tracking is disabled, bailing from deleting current session file.");
+          .getLogger()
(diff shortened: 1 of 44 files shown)


krystofwoldrich:
This is also important for Flutter, as dart uncaught error do not crash the application.

@buenaflor

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
```diff
@@ -148,7 +155,8 @@ public static Map<String, Object> serializeScope(
    *     captured
    */
   @Nullable
-  public static SentryId captureEnvelope(final @NotNull byte[] envelopeData) {
+  public static SentryId captureEnvelope(
+      final @NotNull byte[] envelopeData, final boolean maybeStartNewSession) {
```
buenaflor, line 159: So if I get it right on the Flutter Android side I just need to check if the error is unhandled and set `maybeStartNewSession` to true in those cases

krystofwoldrich: Yes, exactly.

buenaflor:
Yeah makes sense for Flutter as well.

Just an idea, maybe we could surface it in the product a bit better that Flutter does not crash (only the native parts) so users understand what crash free sessions are made out of for Flutter specific

buenaflor:
@krystofwoldrich did you mean to close this pr?

krystofwoldrich:
@buenaflor Thanks, no that happened by accident.

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
```diff
@@ -184,6 +192,10 @@ public static SentryId captureEnvelope(final @NotNull byte[] envelopeData) {
       if (session != null) {
         final SentryEnvelopeItem sessionItem = SentryEnvelopeItem.fromSession(serializer, session);
         envelopeItems.add(sessionItem);
+        deleteCurrentSessionFile(options);
+        if (maybeStartNewSession) {
+          Sentry.startSession();
```
markushi, line 197: Shouldn't make a real difference, but it's better to use hub here.
```suggestion
          hub.startSession();
```

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
markushi: Does the envelope contain any extra information we could use instead of `maybeStartNewSession`? 
I see that we have `event.isCrashed()` maybe, having a `event.isUnhandled()` would make sense as well.

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
```diff
@@ -252,6 +264,26 @@ private static void addTimeSpanToSerializedSpans(TimeSpan span, List<Map<String,
     spans.add(spanMap);
   }
 
+  private static void deleteCurrentSessionFile(final @NotNull SentryOptions options) {
```
markushi, line 267: Hmm why do you need to delete the file manually here? I see we anyway delete the session file on session end [here](https://github.com/getsentry/sentry-java/blob/7ab32b6ab8bc72b973328a00840eb192b151b59f/sentry/src/main/java/io/sentry/cache/EnvelopeCache.java#L103).

markushi (commented):
Looks good overall! I left a few clarifying questions

krystofwoldrich:
>Does the envelope contain any extra information we could use instead of maybeStartNewSession?
I see that we have event.isCrashed() maybe, having a event.isUnhandled() would make sense as well.

Not at the moment, if the event is coming from the global error handler in RN, we have ensure no new session is created, as the application is going to crash.

If the user set the unhandled flag we need to create a new session, because the application is not going to crash.

In flutter all unhandled events are not crashing so we alway need to create a new session if unhandled.

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
krystofwoldrich: Hmm, when I don't explicitly delete it, it's not removed and it's read on the next app start. But I didn't know it should be deleted automatically.

krystofwoldrich: If I'm seeing this right, the envelope is only deleted when the session end is send as standalone envelope, but in the case of hybrids the session end will be send with the crash event in the same envelope.

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
```diff
@@ -252,6 +263,26 @@ private static void addTimeSpanToSerializedSpans(TimeSpan span, List<Map<String,
...
+
+    if (!options.isEnableAutoSessionTracking()) {
+      options
+          .getLogger()
+          .log(DEBUG, "Session tracking is disabled, bailing from deleting current session file.");
+      return;
+    }
+
+    final File sessionFile = EnvelopeCache.getCurrentSessionFile(cacheDirPath);
+    if (!sessionFile.delete()) {
```
stefanosiano, line 281: is this happening in the main thread?

stefanosiano (approved):
lgtm!

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
krystofwoldrich: depends on the hybrid platform that will be used, on RN not, on Flutter yes

buenaflor: For Flutter I guess because the invocation of the method channels happen on the main thread?

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
```diff
@@ -263,6 +266,20 @@ private static void addTimeSpanToSerializedSpans(TimeSpan span, List<Map<String,
...
   }
 
+  private static void deleteCurrentSessionFile(final @NotNull SentryOptions options, boolean isSync) {
+    if (!isSync) {
+      try {
+        options.getExecutorService().submit(() -> {
+          deleteCurrentSessionFile(options);
+        });
+      } catch (Throwable e) {
+        options.getLogger().log(WARNING, "Submission of deletion of the current session file rejected.", e);
```
stefanosiano: if there is an exception you can even decide to call `deleteCurrentSessionFile` here in the catch clause, if it is important

stefanosiano (approved):

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
krystofwoldrich: Just for this is the warning I get if not cleaning the session file
https://github.com/getsentry/sentry-java/blob/6de823e05405663927ff66a16cfc5ab645529cc1/sentry/src/main/java/io/sentry/PreviousSessionFinalizer.java#L74

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
krystofwoldrich: If there is an error, this should be ended and cleaned up on the next app start, but we this deletion should avoid https://github.com/getsentry/sentry-java/blob/6de823e05405663927ff66a16cfc5ab645529cc1/sentry/src/main/java/io/sentry/PreviousSessionFinalizer.java#L74

PR #3495, opened 2024-06-18
Use SecureRandom instead of Random for Metrics

markushi:
## :scroll: Description
In order to not get flagged by security scanners we should use `SecureRandom` in favor of `Random`. Whilst the use-case (random flush shift) doesn't really require a strong random, I guess it's maintenance-friendlier to use `SecureRandom`.

Fixes https://github.com/getsentry/sentry-react-native/issues/3897

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Move fragment auto span finish to onFragmentStarted ([#3424](https://github.com/getsentry/sentry-java/pull/3424))
 - Remove profiling timeout logic and disable profiling on API 21 ([#3478](https://github.com/getsentry/sentry-java/pull/3478))
 - Properly reset metric flush flag on metric emission ([#3493](https://github.com/getsentry/sentry-java/pull/3493))
+- Use SecureRandom in favor of Random for Metrics ([#3495](https://github.com/getsentry/sentry-java/pull/3495))
 
 ## 7.10.0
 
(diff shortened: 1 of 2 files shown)


krystofwoldrich (approved):
LGTM 🚀 

Thank you for the quick PR.

kahest:
@markushi is this on a hot path / do we know if/how much `SecureRandom` is slower than `Random`?

markushi:
> @markushi is this on a hot path / do we know if/how much `SecureRandom` is slower than `Random`?

It's only computed once during class init and we access SecureRandom [here](https://github.com/getsentry/sentry-java/blob/2e90ac7733c6c9b941e7c64175f3d097d1d15a37/sentry/src/main/java/io/sentry/SentryClient.java#L70) and [here](https://github.com/getsentry/sentry-java/blob/2e90ac7733c6c9b941e7c64175f3d097d1d15a37/sentry/src/main/java/io/sentry/TracesSampler.java#L16) already during app start.

I also tried to run it on a test device, takes <1ms to execute.
```Java
final long totalDurationNanos = 0;
final int n = 1000;
for (int i = 0; i < n; i++) {
	final long start = System.nanoTime();
	final double value = new SecureRandom().nextFloat();
	totalDurationNanos += System.nanoTime() - s;
}
Log.d("Test", "took " + TimeUnit.NANOSECONDS.toMillis(totalDurationNanos) + "ms to generate " + n + " items.");


// took 129ms to generate 1000 items.
```

kahest (approved):

PR #3505, opened 2024-06-21
[SR]  Capture screen names as urls 

romtsn:
_#skip-changelog_

## :scroll: Description

* Adds an option to disable screen tracking - this is necessary for RN/Flutter, so they can set their own screens to the native scope and we won't override them
* Adds screen tracking to the missing places (fragment and navigation integrations) to match what we capture as navigation breadcrumbs for replay
* Fills in `replay.urls` array based on navigation breadcrumbs
* For the buffer mode: we have to know what was the last screen before the buffer replay started to display the correct screen name. For that we:
  * Install ScopeObserver to observe whenever `scope.screen` changes and persist it with timestamp
  * When capturing the buffered replay we then search for the last screen before the buffered replay started and add it to the `replay.urls` array

example buffered replay: https://sentry-sdks.sentry.io/replays/0a7eb9067746443c90cfad13662349d3/?project=5428559&query=&referrer=%2Freplays%2F%3AreplaySlug%2F&statsPeriod=1h&yAxis=count%28%29&t=0

and its json payload:
https://us.sentry.io/api/0/projects/sentry-sdks/sentry-android/replays/0a7eb9067746443c90cfad13662349d3/

### DIFF
--- a/sentry/src/main/java/io/sentry/SentryOptions.java
+++ b/sentry/src/main/java/io/sentry/SentryOptions.java
@@ -487,6 +487,7 @@ public class SentryOptions {
    * Controls whether to enable screen tracking. When enabled, the SDK will automatically capture
    * screen transitions as context for events.
    */
+  @ApiStatus.Experimental
   private boolean enableScreenTracking = true;
 
   /**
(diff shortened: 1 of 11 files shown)


brustolin (approved):
LGTM

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -2403,6 +2409,14 @@ public void setReplayController(final @Nullable ReplayController replayControlle
         replayController != null ? replayController : NoOpReplayController.getInstance();
   }
 
+  public boolean isEnableScreenTracking() {
+    return enableScreenTracking;
+  }
+
+  public void setEnableScreenTracking(final boolean enableScreenTracking) {
```
markushi, line 2416: Should we mark it as experimental for now?

markushi (approved):
Nice one! LGTM!

PR #3542, opened 2024-07-02
[SR] Redaction fixes pt. 2

romtsn:
_#skip-changelog_

## :scroll: Description

* Fixes out-of-sync redaction (hopefully) for good:
  * We now listen for the screenshot result on the same (=main) thread, and only capture the view hierarchy after the fact, so in theory there should be no new `onDraw` call in between  
  * However, if there was a `onDraw` call in between, we say that the view hierarchy is invalid and drop the frame (which should never happen)
* Switched `postAtFrontQueue` to `post` as we don't need the front queue anymore
* Additionally made an perf improvement to capture the screenshot of the target (scaled) size, which should be faster
  * For that we also had to adjust how we calculate the dominant color for images, by applying the prescaled matrix to the image's visibleRect before drawing it to the single-pixel bitmap

Tested on this replay:
https://sentry-sdks.sentry.io/replays/daa173f2316c41948b3fa269af59ec4f/?project=5428559&query=&referrer=%2Freplays%2F%3AreplaySlug%2F&statsPeriod=1h&yAxis=count%28%29&t=0

### DIFF
--- a/sentry-android-replay/src/main/java/io/sentry/android/replay/ScreenshotRecorder.kt
+++ b/sentry-android-replay/src/main/java/io/sentry/android/replay/ScreenshotRecorder.kt
@@ -74,7 +74,7 @@ internal class ScreenshotRecorder(
             return
         }
 
-        if (!contentChanged.get() && lastScreenshot != null) {
+        if (!contentChanged.get() && lastScreenshot != null && !lastScreenshot!!.isRecycled) {
             options.logger.log(DEBUG, "Content hasn't changed, repeating last known frame")
 
             lastScreenshot?.let {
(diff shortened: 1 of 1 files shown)


sentry-android-replay/src/main/java/io/sentry/android/replay/ScreenshotRecorder.kt
```diff
@@ -163,16 +160,15 @@ internal class ScreenshotRecorder(
...
-                        contentChanged.set(false)
+                            val screenshot = bitmap.copy(ARGB_8888, false)
+                            screenshotRecorderCallback?.onScreenshotRecorded(screenshot)
+                            lastScreenshot?.recycle()
+                            lastScreenshot = screenshot
+                            contentChanged.set(false)
 
-                        scaledBitmap.recycle()
-                        bitmap.recycle()
+                            bitmap.recycle()
```
markushi, line 170: I guess at some point we could check if recycling the bitmap is really required.

markushi (approved):
LGTM, nice improvement!

sentry-android-replay/src/main/java/io/sentry/android/replay/ScreenshotRecorder.kt
romtsn: yep, was planning to profile memory and see how it behaves - but a bit later, after the beta probably

PR #3546, opened 2024-07-03
[SR] Add `redactClasses` option

krystofwoldrich:
## :scroll: Description

This adds option to redact any java class selected by users. 

This can be also used by React Native to redact SVGs.
 
- https://github.com/getsentry/sentry-react-native/pull/3930

rn sample app

Screenshot 2024-07-03 at 14 12 06

## :pencil: Checklist

#skip-changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Add `redactClasses` to Session Replay options ([#3546](https://github.com/getsentry/sentry-java/pull/3546))
+
 ## 7.11.0-alpha.2
 
 - Session Replay for Android ([#3339](https://github.com/getsentry/sentry-java/pull/3339))
--- a/sentry-android-replay/src/main/java/io/sentry/android/replay/viewhierarchy/ViewHierarchyNode.kt
+++ b/sentry-android-replay/src/main/java/io/sentry/android/replay/viewhierarchy/ViewHierarchyNode.kt
@@ -232,6 +232,10 @@ sealed class ViewHierarchyNode(
             }
         }
 
+        private fun shouldRedact(view: View, options: SentryOptions): Boolean {
+            return options.experimental.sessionReplay.redactClasses.contains(view.javaClass.canonicalName)
+        }
+
         fun fromView(view: View, parent: ViewHierarchyNode?, distance: Int, options: SentryOptions): ViewHierarchyNode {
             val (isVisible, visibleRect) = view.isVisibleToUser()
             when {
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Add `redactClasses` to Session Replay options ([#3546](https://github.com/getsentry/sentry-java/pull/3546))
+
```
romtsn, line 8:
```suggestion
```
we can omit the changelog entry since we still haven't merged to main, so it will all be just `Features: Session Replay` eventually

sentry/src/main/java/io/sentry/SentryReplayOptions.java
```diff
@@ -147,6 +157,14 @@ public void setRedactAllImages(final boolean redactAllImages) {
     this.redactAllImages = redactAllImages;
   }
 
+  public Set<String> getRedactClasses() {
+    return this.redactClasses;
+  }
+
+  public void setRedactClasses(final Set<String> redactClasses) {
```
romtsn, line 164: `m`: can we instead opt for maintaining the in-memory list ourselves and just expose the `addClassToRedact(String className)`? Similar to e.g. how it's done with `bundleIds` here: https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/SentryOptions.java#L1833-L1840

sentry-android-replay/src/main/java/io/sentry/android/replay/viewhierarchy/ViewHierarchyNode.kt
```diff
@@ -282,7 +282,7 @@ sealed class ViewHierarchyNode(
                 (parent?.elevation ?: 0f) + view.elevation,
                 distance = distance,
                 parent = parent,
-                shouldRedact = false,
+                shouldRedact = options.experimental.sessionReplay.redactClasses.contains(view.javaClass.canonicalName),
```
romtsn, line 285: `m`: I think for completeness we should also have this check for `TextViewHierarchyNode` and  `ImageViewHierarchyNode`, but you can skip this if you don't have time right now

romtsn (approved):
left some comments, but approving to unblock

sentry-android-replay/src/main/java/io/sentry/android/replay/viewhierarchy/ViewHierarchyNode.kt
krystofwoldrich: Thanks, yes, that make sense.

sentry/src/main/java/io/sentry/SentryReplayOptions.java
krystofwoldrich: Yeah, I can adjust it to what ever is the best practice in java.

sentry-android-replay/src/main/java/io/sentry/android/replay/viewhierarchy/ViewHierarchyNode.kt
romtsn: yeah but I meant we should change that, i.e. if you disable `redactAllText` but add `TextView` to `redactClasses` it should still redact textviews. But as I said you don't have to do it now, so it's fine to ignore this comment

romtsn:
LGTM after addressing the PR feedback - feel free to merge this in

Issue #3538, opened 2024-07-02
App startup measurements are incorrectly reported


markushi:
### Description

It looks like our app start measurements are incorrectly reported, leading to incorrect values.
 * `app.start.cold` is wrong
 * `ui.load.initial_display` duration is wrong
 * relevant measurements are wrong

[Example event](https://markus-hintersteiner.sentry.io/performance/trace/97570d78828b480786bfec5b42b8e5f1/?fov=318322.76631822775%2C896812.1155886777&node=span-51fe1075f1ce48e1&node=txn-702f487e23e04b7fa368a6a1a4b82b44)

<img width="1267" alt="image" src="https://github.com/getsentry/sentry-java/assets/1411808/5fa29334-b837-475a-9bb5-997c47e3c727">

Issue #3084, opened 2023-12-05
[Starfish] Use a better way to determine app foreground launch


markushi:
### Description

Instead of checking for process foreground like we did earlier 
https://github.com/getsentry/sentry-java/blob/d43311af4de9697012a932ff860c9f2b2ff3d3e0/sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java#L136

we could use a similar approach [as firebase](https://github.com/firebase/firebase-android-sdk/blob/28e3a6821910a0ae452fbcc5b173fafe919ddbaf/firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java#L552-L573).

In simplified terms this would looks like the following:
```java
final AtomicBoolean anyActivityCreated = new AtomicBoolean(false);
registerActivityLifecycleCallbacks(new ActivityLifecycleCallbacksAdapter() {
  @Override
  public void onActivityCreated(@NonNull Activity activity, @Nullable Bundle savedInstanceState) {
    super.onActivityCreated(activity, savedInstanceState);
    anyActivityCreated.set(true);
    unregisterActivityLifecycleCallbacks(this);
  }
});
new Handler(Looper.getMainLooper()).post(() -> {
  final boolean isForegroundAppLaunch = anyActivityCreated.get();
  Log.e(TAG, "onCreate: isForegroundAppLaunch: " + isForegroundAppLaunch);
});
```

stefanosiano:
Implemented by https://github.com/getsentry/sentry-java/pull/3550

Issue #3329, opened 2024-04-05
Android transactions lasting for days, TTID halfway in, all spans at the very end.


realkosty:
### Integration

sentry-android

### Build System

Gradle

### AGP Version

unknown

### Proguard

Enabled

### Version

7.2.0

### Steps to Reproduce

Customer is observing transactions lasting for multiple days where (👉 **see linked Jira issue** for examples)

About 60-70% of all transactions over 1 hour (about 1% of total transactions) exhibit this pattern:
* 2 days until TTID,
* (sometimes) then another day of no spans
* and finally normal spans at the end culminating in transaction being finished
<img width="928" alt="3day-transaction" src="https://github.com/getsentry/sentry-java/assets/490201/37cf8cf4-d3ed-4962-a921-35b86c2ec87d">

(this is the main pattern and the one that I we don't have explanation for, there’s also a second, less common, pattern among extra-long transactions where all spans happen early on, except for one long file.read that delays TTID - this is out of scope of this report, please ignore those).

Below is customer's configuration:

**Settings set during application oncreate via dagger/anvil DI**
```
override fun configure(options: SentryAndroidOptions) {
        dsnValidator.ensureDsnIsValid(sentryOptions.dsn)
        options.dsn = sentryOptions.dsn
        options.isAttachScreenshot = false
        options.isAnrEnabled = sentryFeatureGates.get().isAnrEnabled()
        options.tracesSampleRate = sentryFeatureGates.get().getTracesSampleRate()
        options.isAttachAnrThreadDump = true
        options.environment = buildInfo.getBuildChannel().name.lowercase()
        options.setBeforeSend { event, _ ->
            tagDecorators.forEach {
                event.setTag(it.key, it.value.get())
            }
            event
        }

        options.addInAppInclude("com.redactedcompanyname")

        // Remove the Timber integration so that we can use our own
        options.integrations.removeIf { it is SentryTimberIntegration }

        if (sentryFeatureGates.get().isFragmentsEnabled()) {
            options.addIntegration(
                FragmentLifecycleIntegration(
                    context.applicationContext as Application,
                    enableFragmentLifecycleBreadcrumbs = true,
                    enableAutoFragmentLifecycleTracing = true,
                ),
            )
            options.isEnableUserInteractionTracing = true
            options.isEnableUserInteractionBreadcrumbs = true
        }

// MANIFEST
    <application>
        <provider
            android:name="io.sentry.android.core.SentryInitProvider"
[...]

PR #3550, opened 2024-07-04
Check app start spans time and foreground state

stefanosiano:
## :scroll: Description
App start now takes AppStartMetrics.appLaunchedInForeground variable in consideration when adding spans to the transaction. For some reason we were setting it, but never used it.
App starts longer than 1 minute are dropped (approach taken from Firebase)
Updated foreground app launch check, based on Firebase 

Should fix https://github.com/getsentry/sentry-java/issues/3538 and https://github.com/getsentry/sentry-java/issues/3329, implementing https://github.com/getsentry/sentry-java/issues/3084

Unit tests

## :pencil: Checklist


sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
```diff
@@ -137,12 +146,27 @@ public long getClassLoadedUptimeMs() {
...
-    return getSdkInitTimeSpan();
+    return validateAppStartSpan(getSdkInitTimeSpan());
+  }
+
+  private @NotNull TimeSpan validateAppStartSpan(final @NotNull TimeSpan appStartSpan) {
+    long spanStartMillis = appStartSpan.getStartTimestampMs();
+    long spanEndMillis =
+        appStartSpan.hasStopped()
+            ? appStartSpan.getProjectedStopTimestampMs()
+            : SystemClock.uptimeMillis();
```
markushi, line 162: `h`: span start/stop `TimestampMs` are in unix time, so this should be unix time as well right?
```suggestion
            : System.currentTimeMillis();
```

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
```diff
@@ -102,6 +106,11 @@ public boolean isAppLaunchedInForeground() {
     return appLaunchedInForeground;
   }
 
+  @VisibleForTesting
+  public void setAppLaunchedInForeground(boolean appLaunchedInForeground) {
```
markushi, line 110:
```suggestion
  public void setAppLaunchedInForeground(final boolean appLaunchedInForeground) {
```

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
```diff
@@ -137,12 +146,27 @@ public long getClassLoadedUptimeMs() {
...
     }
 
     // fallback: use sdk init time span, as it will always have a start time set
-    return getSdkInitTimeSpan();
+    return validateAppStartSpan(getSdkInitTimeSpan());
+  }
+
+  private @NotNull TimeSpan validateAppStartSpan(final @NotNull TimeSpan appStartSpan) {
+    long spanStartMillis = appStartSpan.getStartTimestampMs();
+    long spanEndMillis =
```
markushi, line 159:
```suggestion
    final long spanStartMillis = appStartSpan.getStartTimestampMs();
    final long spanEndMillis =
```

markushi (approved):
Nice! If left one comment about uptime vs. unix time that should be addressed, but apart from that LGTM!

sentry-android-core/src/main/java/io/sentry/android/core/ContextUtils.java
```diff
@@ -161,22 +162,77 @@ static String getVersionName(final @NotNull PackageInfo packageInfo) {
...
+
+    // Do not call ProcessStats.getActivityManger, caching will break tests that indirectly depend
+    // on ProcessStats.
+    ActivityManager activityManager =
+        (ActivityManager) appContext.getSystemService(Context.ACTIVITY_SERVICE);
+    if (activityManager == null) {
+      return true;
+    }
+    List<ActivityManager.RunningAppProcessInfo> appProcesses =
+        activityManager.getRunningAppProcesses();
```
romtsn, line 201: fyi, we've already been using `getRunningAppProcesses` but it has been flagged by some stores as unsafe, that's why we changed it to `getMyMemoryState` here: https://github.com/getsentry/sentry-java/pull/3004/files. Can we still keep it?

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
```diff
@@ -195,7 +219,9 @@ public static void onApplicationCreate(final @NotNull Application application) {
     final @NotNull AppStartMetrics instance = getInstance();
     if (instance.applicationOnCreate.hasNotStarted()) {
       instance.applicationOnCreate.setStartedAt(now);
-      instance.appLaunchedInForeground = ContextUtils.isForegroundImportance();
+      instance.appLaunchedInForeground =
```
romtsn, line 222: `h`: I think the real fix that we actually have to implement is properly detecting background starts (which FOREGROUND_IMPORTANCE does not always cover), and we should do it like firebase does: https://github.com/firebase/firebase-android-sdk/blob/58540de24c9b1eb7780c9f642c2cf17478e65734/firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java#L552-L573.

After implementing it we can also close https://github.com/getsentry/sentry-java/issues/3084

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
stefanosiano: yep, nice catch 😄

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
```diff
@@ -23,7 +32,7 @@
  * transformed into SDK specific txn/span data structures.
  */
 @ApiStatus.Internal
-public class AppStartMetrics {
+public class AppStartMetrics implements Application.ActivityLifecycleCallbacks {
```
romtsn, line 35: could use `ActivityLifecycleCallbacksAdapter` I guess

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
```diff
@@ -195,10 +222,57 @@ public static void onApplicationCreate(final @NotNull Application application) {
     final @NotNull AppStartMetrics instance = getInstance();
     if (instance.applicationOnCreate.hasNotStarted()) {
       instance.applicationOnCreate.setStartedAt(now);
+      application.registerActivityLifecycleCallbacks(instance);
```
romtsn, line 225: this looks way better - thanks! However, I think this is only executed for the case where our Gradle plugin instrumentation is applied, but not for other cases (e.g. no gradle plugin, or SentryPerformanceProvider removed, or enablePerformanceV2 is disabled, etc.). I feel like we should register for callbacks in the AppStartMetrics constructor, or something like that to cover all the cases

stefanosiano: The problem is that I need an application instance to register the callbacks
I should add another API for that, and maybe that's an overkill
wdyt?

romtsn: adding another method that accepts Application instance is fine I think

stefanosiano: I added the foreground check in the constructor, though, so that should already improve those cases

romtsn: yeah, but it's still incorrect, foreground importance doesn't always catch background starts. I think with only this change we're not event fixing #3329 properly, because I don't think perfV2 is enabled there, looking at the screenshot. So every background start taking less than 1 minute will still be reported as if it was a proper app start

stefanosiano: actually, right now we are not even checking the foreground flag for some reason 😅 
this means adding a manual API, unless we call it automatically in our contentProvider.
That would leave out only the case where the user doesn't use the gradle plugin and removes the content provider, but there wouldn't be the need to call a manual API, and imho it should be good enough

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
```diff
@@ -195,7 +223,46 @@ public static void onApplicationCreate(final @NotNull Application application) {
...
+      instance.registerApplicationForegroundCheck(application);
+    }
+  }
+
+  /**
+   * Register a callback to check if an activity was started after the application was created
+   *
+   * @param application The application object to register the callback to
+   */
+  public void registerApplicationForegroundCheck(final @NotNull Application application) {
```
romtsn, line 237: nice ❤️ can we also call this method in SentryAndroid then, so it's also covered? We probably will have to also keep a boolean flag to check if the callbacks were already `isRegistered`, to prevent this class from being added multiple times to the lifecycle callbacks (it's just an [ArrayList](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/app/Application.java;l=58) from what I can see, so may have duplicates).

kahest:
@stefanosiano would be good to test if the scenario in https://github.com/getsentry/sentry-java/issues/3567 is also fixed:

> reproduce it if a push notification is received and the app is opened after some time(minutes or hours).

romtsn (approved):
LGTM!

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Check app start spans time and foreground state ([#3550](https://github.com/getsentry/sentry-java/pull/3550))
```
```suggestion
- Check app start spans time and ignore background app starts ([#3550](https://github.com/getsentry/sentry-java/pull/3550))
  - This should eliminate long-lasting App Start transactions
```

PR #3565, opened 2024-07-11
[SR] Capture Replays for ANRs and crashes

romtsn:
## :scroll: Description

The main changes are:
### ANRs part
* We persist `replay_id` from scope to disk
  * We then retrieve it in AnrV2EventProcessor and set it as context to read it out later, when constructing DSC Baggage, so it gets linked with the replay properly
  * Also made a fix that we fall back to `propagationContext` in case there's no active transaction on scope, to always have a trace to link to

### Replay part
*  We now observe values that are necessary for capturing a replay segment and store the to disk continuously 
* On next app launch we read out those values + construct a new cache out of the screenshots saved to disk in the previous run
  * For that we also made `BaseCaptureStrategy.createSegment` method fully parameterized, to supply things like breadcrumbs/rrweb events/segment values not only from the current app run, but also from the previously serialized values when needed

### TODO leftovers
There are some leftovers which (imo) is fine to address later, namely:
* Infer `traceContext` for replay events from the event values, when the event is Backfillable (=sent on next app launch)
* Infer options and tags for replay events from disk (we already have `.options_cache` for that), when the event is Backfillable (=sent on next app launch)
* Choose better serialization format for an ongoing segment, because it's currently awful

Manually + automated

## Next up

Do the same for crashes and JS crashes, and after that get to the buffer mode as it's a bit more work there

## Example ANR with a linked replay

https://sentry-sdks.sentry.io/issues/5587028220/events/635b45b8a6254aa29f8d4f1734480613


sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
```diff
@@ -237,9 +240,169 @@ public class ReplayCache internal constructor(
             encoder?.release()
             encoder = null
         }
+        isClosed.set(true)
+    }
+
+    // TODO: it's awful, choose a better serialization format
```
romtsn, line 246: it's pretty bad right now, cause we load the entire thing into memory, then modify, then serialize it back, and this is called quite often (especially when there's touch events going on). If you have better ideas here, please throw them in.

My one idea so far was to use something like `RandomAccessFile` and only read/write at the specified position directly from/to file

sentry-android-core/src/test/java/io/sentry/android/core/AnrV2EventProcessorTest.kt
```diff
@@ -118,6 +120,7 @@ class AnrV2EventProcessorTest {
                     REQUEST_FILENAME,
                     Request().apply { url = "google.com"; method = "GET" }
                 )
+                persistScope(REPLAY_FILENAME, SentryId("64cf554cc8d74c6eafa3e08b7c984f6d"))
```
brustolin, line 123: Why is this SentryId hardcoded in here?

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
stefanosiano: how big is the full map?
is it feasible to keep it in memory and save to disk on close()?

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
```diff
@@ -237,9 +240,184 @@ public class ReplayCache internal constructor(
...
+
+            cache.frames.sortBy { it.timestamp }
+
+            fun roundToNearestFrame(duration: Long, frameDuration: Int): Long {
+                val frames = duration.toDouble() / frameDuration.toDouble()
+                return ceil(frames).toLong() * frameDuration
+            }
+
+            // we need to round to the nearest frame to include breadcrumbs/events happened after the frame was captured
+            val duration = roundToNearestFrame(
```
stefanosiano, line 381:
```suggestion
            val durationMillis = roundToNearestFrame(
```

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
romtsn: it's actually small 5-10 entries. I'll measure the perf currently, maybe it's not that bad

> is it feasible to keep it in memory and save to disk on close()?

Unfortunately, no, because we don't know when the process is terminated (and close() is not called in that case), so we have to continuously persist the values to have the latest state on restart

sentry-android-replay/src/main/java/io/sentry/android/replay/capture/BaseCaptureStrategy.kt
```diff
@@ -416,4 +521,53 @@ internal abstract class BaseCaptureStrategy(
...
+        crossinline onChange: (propertyName: String?, oldValue: T?, newValue: T?) -> Unit = { _, _, newValue ->
+            propertyName ?: error("Can't persist value without a property name")
+
+            if (options.mainThreadChecker.isMainThread) {
+                persistingExecutor.submit {
+                    cache?.persistSegmentValues(propertyName, newValue.toString())
+                }
+            } else {
+                cache?.persistSegmentValues(propertyName, newValue.toString())
+            }
```
stefanosiano, line 569: this pattern is repeated several times
what about adding an extension to the executor `submitOrNow` or something like that (or a custom Executor class)?

romtsn: yeah, already done here in the other stacked PR: https://github.com/getsentry/sentry-java/pull/3609/files#diff-2f6abcfaf76542e5e58a9fbd4458b23ca658763d9bc0f989abfdeeee5c5898ddR334

sentry/src/main/java/io/sentry/ReplayRecording.java
```diff
@@ -81,7 +81,9 @@ public void serialize(final @NotNull ObjectWriter writer, final @NotNull ILogger
     // {"segment_id":0}\n{json-serialized-rrweb-protocol}
 
     writer.setLenient(true);
-    writer.jsonValue("\n");
+    if (segmentId != null) {
+      writer.jsonValue("\n");
```
markushi, line 85: should you also write the `segmentId`?

sentry/src/main/java/io/sentry/Scope.java
```diff
@@ -940,6 +942,10 @@ public void clearSession() {
   @Override
   public void setPropagationContext(final @NotNull PropagationContext propagationContext) {
     this.propagationContext = propagationContext;
+
+    for (final IScopeObserver observer : options.getScopeObservers()) {
+      observer.setTrace(propagationContext.toSpanContext(), this);
```
markushi, line 947: `propagationContext.toSpanContext()` could be moved out of the loop

markushi (approved):
Looking good, left a few comments 🚀

sentry/src/main/java/io/sentry/ReplayRecording.java
romtsn: it's written on the line 70 above

sentry-android-replay/src/main/java/io/sentry/android/replay/capture/BaseCaptureStrategy.kt
```diff
@@ -416,4 +521,53 @@ internal abstract class BaseCaptureStrategy(
             else -> null
         }
     }
+
+    private inline fun <T> persistableAtomicNullable(
+        initialValue: T? = null,
+        propertyName: String? = null,
```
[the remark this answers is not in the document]
romtsn: yeah there's one case with ScreenshotRecorderConfig where we persist multiple values for the single property, so we provide a custom `onChange` lambda and `propertyName` is not used at all. I've added another overload though and removed the nullables

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
romtsn: will address in the other PR since this line has changed there

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
romtsn: okie, I've made a small improvement for now where we:
* Don't read from the file, but keep the map in-memory (only read on the first access)
* Moved file instantiation out and made it lazy
* Write the in-memory map on change

I'm still going to rework it and use RandomAccessFile most likely (and probably flush periodically, let's see), but a bit later, for now I've measured the time and it's taking 0.5ms on average to write the entire map (also done on a background thread), so I think it's ok for the first iteration

PR #3571, opened 2024-07-12
[SR] Add stopgap for offline session recording

romtsn:
_#skip-changelog_

## :scroll: Description

* Fixes small bug when there were no nav breadcrumb, we were trying to access an empty list
* Add an if-check and avoid capturing frames when offline in session mode

### DIFF
--- a/sentry-android-replay/src/main/java/io/sentry/android/replay/capture/SessionCaptureStrategy.kt
+++ b/sentry-android-replay/src/main/java/io/sentry/android/replay/capture/SessionCaptureStrategy.kt
@@ -2,6 +2,7 @@ package io.sentry.android.replay.capture
 
 import io.sentry.DateUtils
 import io.sentry.Hint
+import io.sentry.IConnectionStatusProvider.ConnectionStatus.DISCONNECTED
 import io.sentry.IHub
 import io.sentry.SentryLevel.DEBUG
 import io.sentry.SentryLevel.INFO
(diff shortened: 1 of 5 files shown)


sentry-android-replay/src/main/java/io/sentry/android/replay/capture/SessionCaptureStrategy.kt
```diff
@@ -74,6 +75,10 @@ internal class SessionCaptureStrategy(
     }
 
     override fun onScreenshotRecorded(store: ReplayCache.(frameTimestamp: Long) -> Unit) {
+        if (options.connectionStatusProvider.connectionStatus == DISCONNECTED) {
+            options.logger.log(DEBUG, "Skipping screenshot recording, no internet connection")
+            return
+        }
```
krystofwoldrich, line 81: How will this behave when the device is offline?

THe screenshots will be taken, but not added to the video?

The Replay session will continue, but there will be a gap? (Assuming the device comes back online)

romtsn: Yep, screenshot will be taken and then dropped immediately without being persisted to disk. I know it's not ideal, ideally we'd not even take a screenshot, but it'd require quite some refactoring and I didn't wanna do that given that it's a temporary measure.

After getting back online, the session will continue and there will be a gap, that's right!

romtsn: but good point, it reminded me that we should also recycle the taken bitmap to not blow up memory, so I've added it in https://github.com/getsentry/sentry-java/pull/3571/commits/7a9658d21daa9c928d119e86298a38c3f3111ba8

krystofwoldrich: Got it, thank you for the details.

krystofwoldrich (approved):
Loog good! 🚀

PR #3601, opened 2024-07-23
ci: dependency review action

mdtro:
Add dependency review action. Any new dependencies added with known high or critical vulnerabilities will fail CI.

### DIFF
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -1,0 +1,19 @@
+name: 'Dependency Review'
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/dependency-review.yml
```diff
@@ -0,0 +1,19 @@
+name: 'Dependency Review'
+on:
+  pull_request:
+    branches: ['master']
```
markushi, line 4:
```suggestion
    branches: ['main']
```

mdtro:
Going to deploy this via global workflows.

PR #3609, opened 2024-07-29
[SR] Support replays for crashes in buffer and session modes

romtsn:
_#skip-changelog_

## :scroll: Description

* Make buffer mode work for ANRs
  * For that we have to store errorSampleRate to disk and then roll the dice on next launch, because it could change in the meantime
  * Also, since in buffer mode we don't actually write `replayId` to scope until an error happens (and in the case of an ANR we don't know until next restart), the persisted `replayId` to disk might be irrelevant. So we have to iterate through the leftover  replay folders and find the one that was last modified closest to the ANR and take it as the previous replay
  * Important part is: we set the relevant `replayId` back to the disk cache, so later `ReplayIntegration` can pick it up and finalize and send it. The order will be ensured by an integration test
* Make buffer mode work for Crashes
  * Renamed `sendReplayForEvent` to `captureReplay` which just accept one argument now to define whether the app process is terminating or not. If it is, then we don't capture the buffered replay in the current process, but will capture it on next launch, to avoid weird issues with half-captured videos
  * Refactored `createSegment` method and moved it to a static method under `CaptureStrategy` so we could call it from `ReplayIntegration` as well
* Small fix for network breadcrumbs where we could deserialize double values sometimes (didn't want to change the deserializer for that to not break something accidentally)
* Small fix for the video encoder, set IFRAME_INTERVAL to -1 so it always only encode partial updates without full frames to save the video size

manually + automated

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AnrV2EventProcessor.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AnrV2EventProcessor.java
@@ -49,7 +49,6 @@ import io.sentry.protocol.Mechanism;
 import io.sentry.protocol.OperatingSystem;
 import io.sentry.protocol.Request;
 import io.sentry.protocol.SdkVersion;
-import io.sentry.protocol.SentryId;
 import io.sentry.protocol.SentryStackTrace;
 import io.sentry.protocol.SentryThread;
 import io.sentry.protocol.SentryTransaction;
@@ -171,43 +170,45 @@ public final class AnrV2EventProcessor implements BackfillingEventProcessor {
     setReplayId(event);
   }
 
-  private boolean sample(final @NotNull Double sampleRate, final @NotNull SecureRandom random) {
-    return !(sampleRate < random.nextDouble());
-  }
-
-  private void setReplayId(final @NotNull SentryEvent event) {
+  private boolean sampleReplay(final @NotNull SentryEvent event) {
     final @Nullable String replayErrorSampleRate =
         PersistingOptionsObserver.read(options, REPLAY_ERROR_SAMPLE_RATE_FILENAME, String.class);
 
     if (replayErrorSampleRate == null) {
-      return;
+      return false;
     }
 
     try {
       // we have to sample here with the old sample rate, because it may change between app launches
       final SecureRandom random = this.random != null ? this.random : new SecureRandom();
-      final Double replayErrorSampleRateDouble = Double.parseDouble(replayErrorSampleRate);
-      if (!sample(replayErrorSampleRateDouble, random)) {
+      final double replayErrorSampleRateDouble = Double.parseDouble(replayErrorSampleRate);
+      if (replayErrorSampleRateDouble < random.nextDouble()) {
         options
             .getLogger()
             .log(
                 SentryLevel.DEBUG,
                 "Not capturing replay for ANR %s due to not being sampled.",
                 event.getEventId());
-        return;
+        return false;
       }
     } catch (Throwable e) {
       options.getLogger().log(SentryLevel.ERROR, "Error parsing replay sample rate.", e);
-      return;
+      return false;
     }
 
+    return true;
+  }
+
+  private void setReplayId(final @NotNull SentryEvent event) {
     String persistedReplayId = PersistingScopeObserver.read(options, REPLAY_FILENAME, String.class);
     final File replayFolder = new File(options.getCacheDirPath(), "replay_" + persistedReplayId);
-    if (persistedReplayId == null
-        || persistedReplayId.equals(SentryId.EMPTY_ID.toString())
-        || !replayFolder.exists()) {
+    if (!replayFolder.exists()) {
+      if (!sampleReplay(event)) {
+        return;
+      }
       // if the replay folder does not exist (e.g. running in buffer mode), we need to find the
       // latest replay folder that was modified before the ANR event.
+      persistedReplayId = null;
       long lastModified = Long.MIN_VALUE;
       final File[] dirs = new File(options.getCacheDirPath()).listFiles();
       if (dirs != null) {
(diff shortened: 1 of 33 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/AnrV2EventProcessor.java
```diff
@@ -156,14 +170,67 @@ private void backfillScope(final @NotNull SentryEvent event, final @NotNull Obje
...
+    final @Nullable String replayErrorSampleRate =
+        PersistingOptionsObserver.read(options, REPLAY_ERROR_SAMPLE_RATE_FILENAME, String.class);
+
+    if (replayErrorSampleRate == null) {
+      return false;
+    }
+
+    try {
+      // we have to sample here with the old sample rate, because it may change between app launches
+      final SecureRandom random = this.random != null ? this.random : new SecureRandom();
```
markushi, line 183:
```suggestion
      final @NotNull SecureRandom random = this.random != null ? this.random : new SecureRandom();
```

sentry-android-core/src/main/java/io/sentry/android/core/AnrV2EventProcessor.java
```diff
@@ -156,14 +170,67 @@ private void backfillScope(final @NotNull SentryEvent event, final @NotNull Obje
...
+
+    return true;
+  }
+
   private void setReplayId(final @NotNull SentryEvent event) {
-    final @Nullable String persistedReplayId =
-        PersistingScopeObserver.read(options, REPLAY_FILENAME, String.class);
+    @Nullable
+    String persistedReplayId = PersistingScopeObserver.read(options, REPLAY_FILENAME, String.class);
+    final File replayFolder = new File(options.getCacheDirPath(), "replay_" + persistedReplayId);
```
markushi, line 205:
```suggestion
    final @NotNull File replayFolder = new File(options.getCacheDirPath(), "replay_" + persistedReplayId);
```

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
```diff
@@ -259,4 +256,67 @@ public class ReplayIntegration(
...
+        // clean up old replays
+        options.cacheDirPath?.let { cacheDir ->
+            File(cacheDir).listFiles { dir, name ->
+                if (name.startsWith("replay_") &&
+                    !name.contains(replayId.toString()) &&
+                    !(unfinishedReplayId.isNotBlank() && name.contains(unfinishedReplayId))
+                ) {
+                    FileUtils.deleteRecursively(File(dir, name))
+                }
+                false
```
markushi, line 270: I don't think the `.listFiles(filter: FileNameFilter)` ([docs here](https://docs.oracle.com/javase/8/docs/api/java/io/File.html#listFiles--)) is meant to be used this way, as your filter code deletes folders while it is still iterating 😅 How about this?
```suggestion
            File(cacheDir).listFiles()?.forEach { file ->
                val name = file.name
                if (name.startsWith("replay_") &&
                    !name.contains(replayId.toString()) &&
                    !(unfinishedReplayId.isNotBlank() && name.contains(unfinishedReplayId))
                ) {
                    FileUtils.deleteRecursively(file)
                }
            }
```

sentry-android-replay/src/main/java/io/sentry/android/replay/capture/BufferCaptureStrategy.kt
```diff
@@ -228,7 +179,75 @@ internal class BufferCaptureStrategy(
...
+            options.logger.log(ERROR, e, "Failed to delete replay segment: %s", file.absolutePath)
+        }
+    }
+
+    private fun MutableList<ReplaySegment.Created>.capture() {
+        var bufferedSegment = removeFirstOrNull()
+        while (bufferedSegment != null) {
+            bufferedSegment.capture(hub)
+            bufferedSegment = removeFirstOrNull()
+            Thread.sleep(100L)
```
markushi, line 204: reserved for future performance optimization? 😄 In any case probably makes sense to add a comment why you're doing that.

sentry-android-replay/src/main/java/io/sentry/android/replay/capture/CaptureStrategy.kt
```diff
@@ -34,4 +51,190 @@ internal interface CaptureStrategy {
...
+            frameRate: Int,
+            videoDuration: Long,
+            replayType: ReplayType,
+            screenAtStart: String?,
+            breadcrumbs: List<Breadcrumb>,
+            events: LinkedList<RRWebEvent>
+        ): ReplaySegment {
+            val endTimestamp = DateUtils.getDateTime(segmentTimestamp.time + videoDuration)
+            val replay = SentryReplayEvent().apply {
+                eventId = currentReplayId
```
markushi, line 130: nit, for readability I'd use `this.` everywhere

markushi (approved):
Nice one, great to see this all coming together. Left a few minor comments

sentry-android-replay/src/main/java/io/sentry/android/replay/capture/BufferCaptureStrategy.kt
romtsn: it's rather not to burst the server with envelopes and active rate limiting, similar to what we do here: https://github.com/getsentry/sentry-java/blob/3a89243fa0ec48fc16193070c2224833d94f9156/sentry/src/main/java/io/sentry/DirectoryProcessor.java#L103-L106

but great point about adding the comment, will do 👍

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
romtsn: yeah, I've figured this in some other place but forgot to change here 😅 thanks for pointing out!

PR #3622, opened 2024-08-06
[SR] Buffer mode improvements

romtsn:
## :scroll: Description

Resolved some TODOs:
* Align next segment timestamp with the end of the buffered segment, when converting from buffer mode to session mode
* Persist `buffer` replay type for the entire replay when converting from buffer mode to session mode
* Rework how `screenAtStart` is stored - now we retrieve current screen name for each frame and store it in the replay cache, and then when we have to rotate the cached frames, we always get the correct screen at start of the buffered replay

part of https://github.com/getsentry/sentry/issues/74441

manually + automated

## :pencil: Checklist


stefanosiano (approved):
lgtm!

markushi (approved):

PR #3642, opened 2024-08-16
Add data from transaction root span to trace.context.data

### LINKED ISSUE #3626: Emit `transaction.data` inside `contexts.trace.data`
- find out if we already do this
- if we don't implement it like this :)

lbloder:
## :scroll: Description
Add root span data to `traceContext` when `SentryTransaction` is created

Fixes #3626 

Modified existing serialization tests

## :pencil: Checklist

### DIFF
--- a/sentry/src/main/java/io/sentry/protocol/SentryTransaction.java
+++ b/sentry/src/main/java/io/sentry/protocol/SentryTransaction.java
@@ -80,8 +80,9 @@ public final class SentryTransaction extends SentryBaseEvent
     contexts.putAll(sentryTracer.getContexts());
 
     final SpanContext tracerContext = sentryTracer.getSpanContext();
+    Map<String, Object> data = sentryTracer.getData();
     // tags must be placed on the root of the transaction instead of contexts.trace.tags
-    contexts.setTrace(
+    final SpanContext tracerContextToSend =
         new SpanContext(
             tracerContext.getTraceId(),
             tracerContext.getSpanId(),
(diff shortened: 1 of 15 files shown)


sentry/src/main/java/io/sentry/protocol/SentryTransaction.java
```diff
@@ -90,15 +91,18 @@ public SentryTransaction(final @NotNull SentryTracer sentryTracer) {
...
-            tracerContext.getOrigin()));
+            tracerContext.getOrigin());
+
+    contexts.setTrace(tracerContextToSend);
     for (final Map.Entry<String, String> tag : tracerContext.getTags().entrySet()) {
       this.setTag(tag.getKey(), tag.getValue());
     }
 
-    final Map<String, Object> data = sentryTracer.getData();
+    // TODO should we remove setExtra here, or do we need it for backwards compat?
```
stefanosiano, line 101: I'd keep it for now

stefanosiano (approved):
nice!

sentry/src/main/java/io/sentry/protocol/SentryTransaction.java
lbloder: @stefanosiano as per the latest discussion (here)[https://github.com/getsentry/sentry-java/issues/3626#issuecomment-2324432211] I'll remove it

lbloder: @stefanosiano still in some discussions about this :) as both not setting the data in `extras` as well as adding data to `trace.context.data` might result in breaking changes for users

lbloder: Closing this for now, created a PR that targets 8.x which removes this (as both keeping and removing this would constitute a breaking change).

lbloder:
Closed in favor of #3735

Issue #3654, opened 2024-08-23
ANRs caused by Sentry.init() while it's doing heavy operations on the main thread.


Chog0:
### Integration

sentry-android

### Build System

Gradle

### AGP Version

8.5.2

### Proguard

Enabled

### Version

7.11.0

### What happened
`SentryAndroid.init(application)` is causing ANRs because of heavy operations on the main thread.
Both issues are caused by  `io.sentry.Sentry.init(Sentry.java:166)`

stacktrace:

```
at java.lang.Runtime.nativeLoad(Native Method)
 at java.lang.Runtime.nativeLoad(Runtime.java:1121)
 at java.lang.Runtime.loadLibrary0(Runtime.java:1075)
 at java.lang.Runtime.loadLibrary0(Runtime.java:998)
 at java.lang.System.loadLibrary(System.java:1661)
 at io.sentry.android.ndk.SentryNdk.<clinit>(SentryNdk.java:19)
 at java.lang.Class.classForName(Native Method)
 at java.lang.Class.forName(Class.java:454)
 at java.lang.Class.forName(Class.java:379)
 at io.sentry.android.core.LoadClass.loadClass(LoadClass.java:21)
 at io.sentry.android.core.AndroidOptionsInitializer.installDefaultIntegrations(AndroidOptionsInitializer.java:257)
 at io.sentry.android.core.SentryAndroid.lambda$init$1(SentryAndroid.java:117)
 at io.sentry.Sentry.applyOptionsConfiguration(Sentry.java:196)
 at io.sentry.Sentry.init(Sentry.java:166)
 at io.sentry.android.core.SentryAndroid.init(SentryAndroid.java:90)
 at io.sentry.android.core.SentryAndroid.init(SentryAndroid.java:73)
```
The second ANR is caused by `io.sentry.android.core.AndroidOptionsInitializer.getCacheDir(AndroidOptionsInitializer.java:362)`
stacktrace:

```
at android.app.ContextImpl.getCacheDir(ContextImpl.java:872)
 at android.content.ContextWrapper.getCacheDir(ContextWrapper.java:332)
 at io.sentry.android.core.AndroidOptionsInitializer.getCacheDir(AndroidOptionsInitializer.java:362)
 at io.sentry.android.core.AndroidOptionsInitializer.loadDefaultAndMetadataOptions(AndroidOptionsInitializer.java:110)
 at io.sentry.android.core.SentryAndroid.lambda$init$1(SentryAndroid.java:111)
[...]

Issue #3585, opened 2024-07-18
io.sentry.android.ndk.SentryNdk.<clinit> causes ANR


seven415:
### Integration

sentry-android

### Build System

Gradle

### AGP Version

7.3.1

### Proguard

Enabled

### Version

7.9.0

### Steps to Reproduce

sdk init 

### Expected Result

ANR

### Actual Result

#00 pc 0x5e3e8 libc.so (syscall + 28)
#01 pc 0x131bb3 libart.so (art::ConditionVariable::WaitHoldingLocks(art::Thread*) + 82)
#02 pc 0x2f5bed libart.so (art::JNI<false>::NewWeakGlobalRef(_JNIEnv*, _jobject*) + 372)
#03 pc 0x29d11b libart.so (art::JavaVMExt::LoadNativeLibrary(_JNIEnv*, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, _jobject*, _jclass*, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >*) + 1930)
#04 pc 0x4103 libopenjdkjvm.so (JVM_NativeLoad + 246)
       at java.lang.Runtime.nativeLoad(Native method)
       at java.lang.Runtime.nativeLoad(Runtime.java:1131)
       at java.lang.Runtime.loadLibrary0(Runtime.java:1085)
       at java.lang.Runtime.loadLibrary0(Runtime.java:1008)
       at java.lang.System.loadLibrary(System.java:1664)
       at io.sentry.android.ndk.SentryNdk.<clinit>(SentryNdk.java:18)
       at java.lang.Class.classForName(Native method)
       at java.lang.Class.forName(Class.java:454)
       at java.lang.Class.forName(Class.java:379)
       at io.sentry.android.core.LoadClass.loadClass(LoadClass.java:21)
       at io.sentry.android.core.AndroidOptionsInitializer.installDefaultIntegrations(AndroidOptionsInitializer.java:257)
       at io.sentry.android.core.SentryAndroid.lambda$init$1(SentryAndroid.java:114)
       at io.sentry.Sentry.applyOptionsConfiguration(Sentry.java:196)
       at io.sentry.Sentry.init(Sentry.java:165)
       at io.sentry.android.core.SentryAndroid.init(SentryAndroid.java:87)
       at io.sentry.android.core.SentryAndroid.init(SentryAndroid.java:58)
[...]

PR #3670, opened 2024-08-30
[QA] Offload System.loadLibrary call to background thread

markushi:
## :scroll: Description
Perform `System.loadLibrary()`  on a background thread.

Calling `System.loadLibrary()` causes StrictMode violations, as it performs disk reads, but any JNI calls like `initSentryNative` needs the libraries to be loaded beforehand.

Initializing the whole NDK off the main thread would be an option, but it has two main disadvantages:
* there's no guarantee that the NDK is initialized after `SentryAndroid.init()`, thus we could miss early startup crashes
* all NDK communication (setUser, setTag) needs to be done on a background thread as well

So the trick is to perform `System.loadLibrary()` as early as possible on a background thread, but still perform NDK init synchronous on the main thread. In my testing this works well, as the SDK init in-between takes longer than the `.loadLibrary()` calls, making the `loadLibraryLatch.await(2000, TimeUnit.MILLISECONDS)` return immediatelly.

Just manually, haven't found a better way yet.

## :pencil: Checklist


markushi:
Comparing the performance metrics with other PRs, we can see a nice 35-45% improvement of the startup time 🚀

kahest:
@markushi sounds very promising 👍 any creative ways to test this end-to-end?

kahest:
for reference: this is related to/fixes (partially) https://github.com/getsentry/sentry-java/issues/3585 and https://github.com/getsentry/sentry-java/issues/3654

stefanosiano (approved):
Looks good!
Not sure how/if we can test it, but "logically" it works

romtsn (approved):
I think we wanted to make it part of the next major (8.x.x) - do we still wanna do that?

Aside, this looks good, and I think eventually we could offload the native init to the background thread as well, because we wanted to move `setTag`, `setBreadcrumbs` to background anyway.

PR #3680, opened 2024-09-05
fix min required version for replay

bruno-garcia:


bruno-garcia:
We actually require 26

PR #3689, opened 2024-09-13
[SR] Add custom redaction options

romtsn:
## :scroll: Description

* Added a possibility to redact/ignore view types by adding the fq class name to the redact/ignore list in replay options
  * This will also consider subclasses of the added view type (e.g. EditText will be redacted because it inherits from TextView)
* Added a possibility to redact/ignore certain views by setting a tag
  * Either a regular `setTag("sentry-redact|sentry-ignore")`
  * Or in case of multiple tags for a view `setTag(R.id.sentry_privacy, "redact|ignore")`
* Added convenient view extensions that set tags under the hood
* Removed the `redactAllText` and `redactAllImages` booleans in favour of just adding the android view classes to the `redactClasses` list
* Some minor fixes along-the-way in `Windows.kt` where we were not properly handling double-init of the SDK

Also note that the order in which we define whether to redact a view or not matters, and marking individual views has precedence over a class. You can read up more on it here: https://www.notion.so/sentry/Redaction-Ignore-Mechanisms-for-Views-45e1302e8d844bd0b83a274a5d759e3f

Manually + automated

## :pencil: Checklist

- [] No breaking change or entry added to the changelog.


sentry/src/main/java/io/sentry/SentryReplayOptions.java
```diff
@@ -49,30 +52,28 @@ public enum SentryReplayQuality {
...
-   * redacted. ColorDrawable, InsetDrawable, VectorDrawable are all considered non-PII, as they come
-   * from the apk.
+   * <p>If you're using an obfuscation tool, make sure to add the respective proguard rules to keep
+   * the class names.
    *
-   * <p>Default is enabled.
+   * <p>Default is empty.
    */
-  private boolean redactAllImages = true;
+  private Set<String> redactClasses = new CopyOnWriteArraySet<>();
```
romtsn, line 64: @brustolin @bruno-garcia I guess we should align on naming here. Do you prefer this or `redactViewTypes`? I'm fine with either

brustolin: I think we should align but just Classes is too generic, what about "redactViewClasses"?

sentry-android-replay/src/main/java/io/sentry/android/replay/SessionReplayOptions.kt
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Redact all text content. Draws a rectangle of text bounds with text color on top. By default
+ * only views extending TextView are redacted.
+ *
+ * <p>Default is enabled.
+ */
+var SentryReplayOptions.redactAllText: Boolean
+    @Deprecated("Getter is unsupported.", level = DeprecationLevel.ERROR)
+    get() = error("Getter not supported")
```
brustolin, line 17: l: Why is this deprecated?

brustolin (approved):
Looks good to me.

One question: We just need to redact TextView? All the input views are based on TextView?

sentry-android-replay/src/main/java/io/sentry/android/replay/SessionReplayOptions.kt
romtsn: I explained in a comment on top of the file, but basically we scream to the dev that this should not be used to get the value. This extension only exists to set the value conveniently in Kotlin (`redactAllText = true` instead of `setRedactAllText(true)`), so only the setter matters here

sentry/src/main/java/io/sentry/SentryReplayOptions.java
romtsn: yeah, sounds good. I can also change it to `redactViewTypes` so you don't have to :)

romtsn:
> Looks good to me.
> 
> One question: We just need to redact TextView? All the input views are based on TextView?

yep, `EditText`, `RadioButton` and others all inherit from `TextView` 
image

sentry-android-replay/src/main/java/io/sentry/android/replay/SessionReplayOptions.kt
brustolin: Thanks!! I didn't get the comment.

Can you just remove `get()` as in C# where you can have a property with just a setter?

romtsn: nope, unfortunately not possible in Kotlin. It's possible to only have a getter without a setter, but not the other way around

markushi (approved):
LGTM!

PR #3706, opened 2024-09-20
Replace Android thread id with kernel thread id in span data

stefanosiano:
## :scroll: Description
renamed IMainThreadChecker to IThreadChecker
added getCurrentThreadSystemId() to IThreadChecker
replaced thread id of SentryTracer and PerformanceAndroidEventProcessor

Adding to v8, since it's a behavioural change and users could already use it in some dashboard, as it's indexed. This is very unlikely, though.

Continuous profiler needs to bind the trace and span to the corresponding thread in the profile.
The Android profiler retrieves the kernel thread id, that we can get via `Process.myTid()`, so we have to set that instead of `Thread.getThreadId()`

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/internal/util/AndroidThreadChecker.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/internal/util/AndroidThreadChecker.java
@@ -1,5 +1,6 @@
 package io.sentry.android.core.internal.util;
 
+import android.os.Handler;
 import android.os.Looper;
 import android.os.Process;
 import io.sentry.protocol.SentryThread;
@@ -12,13 +13,16 @@ import org.jetbrains.annotations.NotNull;
 public final class AndroidThreadChecker implements IThreadChecker {
 
   private static final AndroidThreadChecker instance = new AndroidThreadChecker();
-  public static final long mainThreadId = Process.myTid();
+  public static long mainThreadId = Process.myTid();
 
   public static AndroidThreadChecker getInstance() {
     return instance;
   }
 
-  private AndroidThreadChecker() {}
+  private AndroidThreadChecker() {
+    // The first time this class is loaded, we make sure to set the correct mainThreadId
+    new Handler(Looper.getMainLooper()).post(() -> mainThreadId = Process.myTid());
+  }
 
   @Override
   public boolean isMainThread(final long threadId) {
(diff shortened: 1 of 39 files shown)


stefanosiano:
Most changes come from renaming IMainThreadChecker to IThreadChecker.
I can revert the renaming if you like it more

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -2070,12 +2069,30 @@ public void setViewHierarchyExporters(@NotNull final List<ViewHierarchyExporter>
...
   }
 
-  public @NotNull IMainThreadChecker getMainThreadChecker() {
-    return mainThreadChecker;
+  /**
+   * @deprecated Use {@link SentryOptions#getThreadChecker} instead.
+   */
+  @Deprecated
+  @SuppressWarnings("InlineMeSuggester")
+  public @NotNull IThreadChecker getMainThreadChecker() {
```
stefanosiano, line 2077: i added this method to reduce breaking changes, but since it's a major and the IThreadMainChecker (now IThreadChecker) is internal, we can remove it directly. 
Wdyt?

romtsn: yeah, I think it's safe to remove it

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/AndroidThreadChecker.java
```diff
@@ -1,22 +1,28 @@
...
 import org.jetbrains.annotations.NotNull;
 
 /** Class that checks if a given thread is the Android Main/UI thread */
 @ApiStatus.Internal
-public final class AndroidMainThreadChecker implements IMainThreadChecker {
+public final class AndroidThreadChecker implements IThreadChecker {
 
-  private static final AndroidMainThreadChecker instance = new AndroidMainThreadChecker();
+  private static final AndroidThreadChecker instance = new AndroidThreadChecker();
+  public static long mainThreadId = Process.myTid();
```
romtsn, line 16: I think it should be volatile?

romtsn: also, maybe we should name it better, like `mainThreadKernelId` or `mainThreadSystemId` to distinguish from the thread id

romtsn (approved):
LGTM, just a couple of comments around mainThreadId

PR #3710, opened 2024-09-23
Add Continuous Profiling support (v8)

stefanosiano:
## :scroll: Description
This is the first part for the continuous profiling support for Android. 
Internal doc [is here](https://www.notion.so/sentry/Continuous-Profiling-Android-SDK-df867ce925a3449aa3c760717726240d?pvs=4)

Part 1
Add the implementation of the continuous profiler itself

Part 2 -> https://github.com/getsentry/sentry-java/pull/3711
Create new envelope type and payload
Create and send profile_chunk envelope

Part 3 -> https://github.com/getsentry/sentry-java/pull/3725
Instantiate profiler
Set ProfileContext to transactions and spans
Add performance collectors to the profiler

Part 4 -> https://github.com/getsentry/sentry-java/pull/3730
Handle app start profiling

Part 5 -> https://github.com/getsentry/sentry-java/pull/3844
Add new user-facing APIs

Part 6 -> https://github.com/getsentry/sentry-java/pull/3926
Handle rate limiting

Part 7 -> https://github.com/getsentry/sentry-java/pull/4013
Add sample rate

Part 8
Update docs
User docs -> https://github.com/getsentry/sentry-docs/pull/11872

Wrap everything up -> https://github.com/getsentry/sentry-java/pull/4069

We are waiting to merge this into master as the profile is not showing yet on the product side.
In the meantime we are implementing the finalized changes (which would be breaking if we released it)

* Changing sample rate to session sample rate -> https://github.com/getsentry/sentry-java/pull/4180
* Rename APIs -> https://github.com/getsentry/sentry-java/pull/4182
* add profile_lifecycle option (manual or trace) -> https://github.com/getsentry/sentry-java/pull/4202
* add `start_profiler_on_app_start` option -> https://github.com/getsentry/sentry-java/pull/4226
* deprecate and remove v0 APIs and options
* [enh] use single chunk for multiple transactions with a small gap
* Add onboarding docs
* Add continuous profiling to wizard

## :green_heart: How did you test it?
Unit tests

## :pencil: Checklist

- [ ] I reviewed the submitted code.
- [ ] I added tests to verify the changes.
- [ ] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] Review from the native team if needed.
- [ ] No breaking change or entry added to the changelog.
- [ ] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

PR #3711, opened 2024-09-23
Change payload for Continuous Profiling v8 (p2)

stefanosiano:
## :scroll: Description
added profile_chunk envelope create
added IHub.captureProfileChunk and ISentryClient.captureProfileChunk
added profilerId and chunkId reset logic to AndroidContinuousProfiler
added absolute timestamps to ProfileMeasurementValue
added ProfileContext to Contexts

#skip-changelog

Implements payload changes for Continuous Profiling
Part 2 of https://github.com/getsentry/sentry-java/pull/3710

## :pencil: Checklist


sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
```diff
@@ -138,14 +155,29 @@ private synchronized void stop(final boolean restartProfiler) {
       return;
```
romtsn, line 155: I'm thinking we should also reset the `chunkId` and the `profilerId` (in case of manual stop) in this case right?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
```diff
@@ -157,6 +189,28 @@ public synchronized void close() {
     stop();
   }
 
+  private void sendChunks(final @NotNull IScopes scopes, final @NotNull SentryOptions options) {
+    try {
+      options
+          .getExecutorService()
+          .submit(
+              () -> {
+                final ArrayList<ProfileChunk> payloads = new ArrayList<>(payloadBuilders.size());
```
romtsn, line 198: any reason we don't store chunks (not builders) directly in [here](https://github.com/getsentry/sentry-java/pull/3711/files#diff-282328031af7b576dd167682cc966fe6852b8b0a6edaf27cc1c68427497d0ad2R162)? This seems like an unnecessary allocation of another list here

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
```diff
@@ -157,6 +189,28 @@ public synchronized void close() {
...
+                synchronized (payloadBuilders) {
+                  for (ProfileChunk.Builder builder : payloadBuilders) {
+                    payloads.add(builder.build(options));
+                  }
+                  payloadBuilders.clear();
+                }
+                for (ProfileChunk payload : payloads) {
+                  scopes.captureProfileChunk(payload);
```
romtsn, line 206: how many pending chunks we could have potentially? Just thinking if we burst the backend here and cause rate-limit to kick in? Maybe we could add a short 100ms delay between sending chunks if we expect a few.

sentry/src/main/java/io/sentry/CpuCollectionData.java
```diff
@@ -1,15 +1,25 @@
 package io.sentry;
 
 import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
 
 @ApiStatus.Internal
 public final class CpuCollectionData {
   final long timestampMillis;
   final double cpuUsagePercentage;
+  final @NotNull SentryDate timestamp;
```
romtsn, line 10: could we just have one (SentryDate) timestamp here and for memory/frames?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
stefanosiano: yes, we do it [later](https://github.com/getsentry/sentry-java/pull/3711/files#diff-282328031af7b576dd167682cc966fe6852b8b0a6edaf27cc1c68427497d0ad2R180) in this function
But we can add it here, too

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
stefanosiano: this way we can call `build(SentryOptions)` when we send them
I did it because when we'll implement app start profiling we may not have the SDK initialized and the options when the profiler is running, so we'll wait to send the chunks as soon as we have them
But usually it should be a list with a single element
I'm using a different list so that we can clear the previous one in the synchronized block, without waiting for `scopes.captureProfileChunk(payload);`

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
stefanosiano: i expect only one, honestly
The only case when we can have more is app start profiling, but i'd expect one there, too, most of the times

sentry/src/main/java/io/sentry/CpuCollectionData.java
stefanosiano: we can change the collectors so that when `collect()` is called we can pass a timestamp, to have only one for cpu and memory
But frames are different, as they are captured with different intervals

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
romtsn: yeah I guess we have to, because if `endAndCollect` returns null, we're returning early and not resetting the ids, so the next time we start we'll use the wrong ids

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
romtsn: got it, makes sense!

sentry/src/main/java/io/sentry/CpuCollectionData.java
romtsn: yeah I guess what I meant is `timestamp` already contains `timestampMillis` in it in a way, so we don't need to manage two values that are the same thing here

stefanosiano: done, thanks for the suggestion
another look would be great :D

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
```diff
@@ -143,17 +159,35 @@ private synchronized void stop(final boolean restartProfiler) {
 
     // check if profiler end successfully
     if (endData == null) {
+      // A problem occurred. Profile chunk is not captured. Let's reset ids.
+      chunkId = SentryId.EMPTY_ID;
+      profilerId = SentryId.EMPTY_ID;
```
romtsn, line 164:
```suggestion
      if (!restartProfiler) {
        profilerId = SentryId.EMPTY_ID;
      }
```
should it be like this?

romtsn (approved):

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
stefanosiano: not sure
in this case we don't restart profile chunks, so the profiler effectively stops
we could also restart the profiler in here, and we could keep the same profiler id, but that would create a 10 second gap in the profile... 
not sure honestly, but we are talking about an edge case that should never happen (profiler didn't write a file or it never started in the first place)

stefanosiano: ok, i will just continue the execution and log an error message without returning early
i realized the `isRunning` flag is also not reset, and unexpected behaviours may happen: if a user starts a profile he expects to have it run continuously until he stops it, even in case some error occurs, imho

romtsn: fine by me!

PR #3725, opened 2024-09-26
Instantiate continuous profiling v8 (p3)

stefanosiano:
## :scroll: Description
added profile context to SentryTracer
removed isProfilingEnabled from AndroidContinuousProfiler, as it's useless
added continuous profiler to SentryOptions
added DefaultTransactionPerformanceCollector to AndroidContinuousProfiler and updated it to work with string ids other than transactions
fixed ProfileChunk measurements being modifiable from other code
added thread id and name to SpanContext.data

#skip-changelog

Instantiate Continuous Profiler
Part 3 of https://github.com/getsentry/sentry-java/pull/3710

Unit tests

## :pencil: Checklist


sentry/src/main/java/io/sentry/TransactionPerformanceCollector.java
```diff
@@ -7,8 +7,12 @@
 
 public interface TransactionPerformanceCollector {
```
romtsn, line 8: I guess we could rename it to `PerformanceMetricsCollector` or something since it's not bound to transactions anymore? We'd probably need something similar for replays in the future, so I guess it's better to rename it now since it's part of v8?

sentry/src/main/java/io/sentry/SpanContext.java
```diff
@@ -41,6 +41,8 @@ public class SpanContext implements JsonUnknown, JsonSerializable {
   /** Describes the status of the Transaction. */
   protected @Nullable SpanStatus status;
 
+  protected final @NotNull Map<String, Object> data;
```
romtsn, line 44: we're not ser/deser this, is this coming in the follow-up PRs or just missed?

sentry/src/main/java/io/sentry/SpanContext.java
```diff
@@ -119,6 +126,14 @@ public SpanContext(final @NotNull SpanContext spanContext) {
     if (copiedUnknown != null) {
       this.unknown = copiedUnknown;
     }
+    this.instrumenter = spanContext.instrumenter;
```
romtsn, line 129: just wanna check with @adinauer if this was just missed, or intended to not copy those?

sentry/src/main/java/io/sentry/SpanContext.java
```diff
@@ -94,6 +96,11 @@ public SpanContext(
     this.description = description;
     this.status = status;
     this.origin = origin;
+    final long threadId =
+        ScopesAdapter.getInstance().getOptions().getThreadChecker().currentThreadSystemId();
+    this.data = new ConcurrentHashMap<>();
+    this.data.put(SpanDataConvention.THREAD_ID, String.valueOf(threadId));
+    this.data.put(SpanDataConvention.THREAD_NAME, Thread.currentThread().getName());
```
romtsn, line 103: We're using [this lines](https://github.com/getsentry/sentry-java/pull/3725/files#diff-8bb325ca6fc9666c24826415b357d8e76a34cb455edc5709dd962bdcd34c6f16R527-R531) to set the thread name, should we also do the same thing here?

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -86,6 +86,12 @@ public SentryTracer(
       this.baggage = new Baggage(scopes.getOptions().getLogger());
     }
 
+    final @NotNull SentryId continuousProfilerId =
```
romtsn, line 89: just a thought: I'd imagine this would be better leaving on the Scope(s), but since it's a global profiler it's probably fine.

sentry/src/main/java/io/sentry/SpanContext.java
stefanosiano: nope, i forgot it 😅

romtsn (commented):
LGTM, but one comment about where we send `profiler_id`:
-  I don't know about backend implications, but why don't we make `profiler_id` part of the baggage/trace-header? I'd imagine it'd make sense to have it there to make sampling decisions for spans/txs etc etc, or is it already supposed to work with spans-without-sampling? Just did the same thing for replays recently, that's why I'm asking, because there it differs from this approach a little.

romtsn (approved):

sentry/src/main/java/io/sentry/TransactionPerformanceCollector.java
romtsn: Or maybe `AggregatedPerformanceCollector`/`CompositePerformanceCollector` since it combines all other `IPerformanceCollectors`

stefanosiano: I renamed it to `CompositePerformanceCollector`

sentry/src/main/java/io/sentry/SpanContext.java
stefanosiano: Good call, I added `IThreadChecker.getCurrentThreadName()`, so we can reuse the logic from there

stefanosiano:
> LGTM, but one comment about where we send `profiler_id`:
> 
> * I don't know about backend implications, but why don't we make `profiler_id` part of the baggage/trace-header? I'd imagine it'd make sense to have it there to make sampling decisions for spans/txs etc etc, or is it already supposed to work with spans-without-sampling? Just did the same thing for replays recently, that's why I'm asking, because there it differs from this approach a little.

Chunks are not bound to traces at the moment, but I agree on your thoughts.
We should probably do it at some point, and then we'll revisit this, but not now sadly

sentry/src/main/java/io/sentry/SpanContext.java
stefanosiano: i talked with him and it was just missed, but it hasn't caused any issue.
Also, it will be removed in the future, so we can remove it from here

PR #3730, opened 2024-09-30
Handle App Start Continuous Profiling v8 (p4)

stefanosiano:
## :scroll: Description
create app start continuous profiler instead of transaction profiler, based on config
updated SentryAppStartProfilingOptions with isContinuousProfilingEnabled flag
updated SentryOptions with isContinuousProfilingEnabled() method

#skip-changelog

Handle App Start Continuous Profiler
Part 4 of https://github.com/getsentry/sentry-java/pull/3710

## :pencil: Checklist


sentry-android-core/src/main/java/io/sentry/android/core/SentryPerformanceProvider.java
```diff
@@ -98,6 +99,11 @@ public void shutdown() {
       if (appStartProfiler != null) {
         appStartProfiler.close();
       }
+      final @Nullable IContinuousProfiler appStartContinuousProfiler =
```
romtsn, line 102: as I understood there could be only one profiler at a time (cont. or tx-based), right? Just got an idea we could have a general `IProfiler` interface (that would have all the methods that `IContinuousProfiler` currently has), and then it'd have a sub-interface `ITransactionProfiler` with the tx-specific methods.

This way you could only manage one instance of `IProfiler` everywhere (like here) and it would be more readable I guess.

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -1731,6 +1731,17 @@ public boolean isProfilingEnabled() {
...
   }
 
+  /**
+   * Returns if continuous profiling is enabled. This means that no profile sample rate has been
+   * set.
+   *
+   * @return if continuous profiling is enabled.
+   */
+  @ApiStatus.Internal
+  public boolean isContinuousProfilingEnabled() {
```
romtsn, line 1741: hm, so this means setting `profilesSampleRate` to 0 would disable both the continuous and the tx-based profilers? I guess we should document this very well as this may be confusing

romtsn (approved):

sentry-android-core/src/main/java/io/sentry/android/core/SentryPerformanceProvider.java
stefanosiano: you're probably right on this, but in the long term we will deprecate and optionally remove the transaction profiler.
I think the extraction of the `AndroidProfiler` is enough for now.
But we can refactor it if you have a strong preference

sentry/src/main/java/io/sentry/SentryOptions.java
stefanosiano: [yep](https://www.notion.so/sentry/Continuous-Profiling-SDK-46f8f4ce691a4fedbcc39c7f1e1e740e?pvs=4#12f8b10e4b5d803e959bca2b9d533afc)
It's already pointed out in the [docs](https://docs.sentry.io/platforms/apple/profiling/#continuous-profiling) (Android docs will be mostly the same): `If you had previously set SentryOptions.profilesSampleRate or SentryOptions.profilesSampler to use transaction-based profiling, then remove those lines of code from your configuration.`
I also added a log message when the startProfile() is called and a `profilesSampleRate` is set, to avoid this confusion in [p5](https://github.com/getsentry/sentry-java/pull/3844)

PR #3736, opened 2024-09-30
Replace Calendar.getInstance with System.currentTimeMillis for breadcrumb ctor

romtsn:
## :scroll: Description

Since we use UTC timezone for Calendar for Breadcrumb default ctor, we can actually replace it with `System.currentTimeMillis` as it's way faster, and then convert to `Date` only when serialization happens (which is performed on a bg thread already).

We can clearly see the improvement here:

### Before
Android Studio 2024-10-01 00 24 54

### After
Android Studio 2024-10-01 00 18 20

### Link to event
https://sentry-sdks.sentry.io/issues/4320603017/events/c829546cbfc742b1b333539c9f39cc93/

Improve SDK performance

### DIFF
--- a/sentry-samples/sentry-samples-android/build.gradle.kts
+++ b/sentry-samples/sentry-samples-android/build.gradle.kts
@@ -133,5 +133,5 @@ dependencies {
     implementation(Config.Libs.composeNavigation)
     implementation(Config.Libs.composeMaterial)
 
-    debugImplementation(Config.Libs.leakCanary)
+//    debugImplementation(Config.Libs.leakCanary)
 }
--- a/sentry-samples/sentry-samples-android/build.gradle.kts
+++ b/sentry-samples/sentry-samples-android/build.gradle.kts
@@ -133,5 +133,5 @@ dependencies {
     implementation(Config.Libs.composeNavigation)
     implementation(Config.Libs.composeMaterial)
 
-//    debugImplementation(Config.Libs.leakCanary)
+    debugImplementation(Config.Libs.leakCanary)
 }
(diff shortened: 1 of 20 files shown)


sentry-samples/sentry-samples-android/build.gradle.kts
```diff
@@ -133,5 +133,5 @@ dependencies {
     implementation(Config.Libs.composeNavigation)
     implementation(Config.Libs.composeMaterial)
 
-    debugImplementation(Config.Libs.leakCanary)
```
markushi, line 136: Any specific reason you commented this out?

romtsn: nope, reverted

markushi (approved):

stefanosiano (approved):

sentry/src/main/java/io/sentry/Breadcrumb.java
```diff
@@ -522,22 +528,21 @@ public Breadcrumb(@Nullable String message) {
...
-
   /**
    * Returns the Breadcrumb's timestamp as java.util.Date
    *
    * @return the timestamp
    */
+  @SuppressWarnings("JavaUtilDate")
   public @NotNull Date getTimestamp() {
-    return DateUtils.getDateTime(timestamp);
+    if (timestamp != null) {
```
stefanosiano, line 538: is it still showing improvements?
This looks like moving the issue from breadcrumb creation to event capture (as we sort the breadcrumbs). 
Or am i missing something?

romtsn: Yeah we only sort them when they are not created by us (=someone set them to the event), but I doubt anyone does that, so this will cover most of the cases probably: https://github.com/getsentry/sentry-java/blob/7e57220ee4ffa7af590fff675708884c4a07370c/sentry/src/main/java/io/sentry/SentryClient.java#L1054-L1058

PR #3754, opened 2024-10-04
[QA] Move NDK scope sync to background thread

### LINKED ISSUE #3046: Offload NDK Scope sync to background thread
### Problem Statement

As of now any context data (user, breadcrumb, tags, ...) is synced to `sentry-native` on the calling thread. The calling thread might be the main thread, which is not optimal.

### Solution Brainstorm

We should offload the scope sync to a background thread, ensuring we're not spending too much time on the main thread.

markushi:
## :scroll: Description
Utilizes our ExecutorService to perform NDK scope sync (and it's JSON serialization) on a background thread.
This will cause an async scope sync, meaning if e.g. `.setUser()` returns, the value is not visible on the NDK side right away. The upside is faster execution.

If we see there's demand for an in-sync solution, we can introduce a flag later.

## Motivation
Closes #3046

Extended tests 

## :pencil: Checklist


romtsn (approved):
LGTM! Just thinking if we should introduce a ndk-specific executor here? Our default executor does too many things already 😅

stefanosiano (approved):

PR #3775, opened 2024-10-08
[SR] Mask web and video views

romtsn:
## :scroll: Description

* Mask more things by default: WebViews, VideoViews, and androidx PlayerViews
* Reverted key_frame interval to 10s, because I realized for buffered replays (which are 30sec), the quality becomes garbage without having key_frames at all. 10s seems like a good middle-ground for low size but decent quality videos.

Manually on the pocketcasts app

### DIFF
--- a/sentry/src/main/java/io/sentry/SentryReplayOptions.java
+++ b/sentry/src/main/java/io/sentry/SentryReplayOptions.java
@@ -11,6 +11,9 @@ public final class SentryReplayOptions {
 
   public static final String TEXT_VIEW_CLASS_NAME = "android.widget.TextView";
   public static final String IMAGE_VIEW_CLASS_NAME = "android.widget.ImageView";
+  public static final String WEB_VIEW_CLASS_NAME = "android.webkit.WebView";
+  public static final String VIDEO_VIEW_CLASS_NAME = "android.widget.VideoView";
+  public static final String EXOPLAYER_VIEW_CLASS_NAME = "androidx.media3.ui.PlayerView";
 
   public enum SentryReplayQuality {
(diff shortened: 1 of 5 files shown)


sentry/src/main/java/io/sentry/SentryReplayOptions.java
```diff
@@ -11,6 +11,9 @@ public final class SentryReplayOptions {
 
   public static final String TEXT_VIEW_CLASS_NAME = "android.widget.TextView";
   public static final String IMAGE_VIEW_CLASS_NAME = "android.widget.ImageView";
+  public static final String WEB_VIEW_CLASS_NAME = "android.webkit.WebView";
+  public static final String VIDEO_VIEW_CLASS_NAME = "android.widget.VideoView";
+  public static final String ANDROIDX_MEDIA_VIEW_CLASS_NAME = "androidx.media3.ui.PlayerView";
```
markushi, line 17: How about adding the Exoplayer View too? It's `com.google.android.exoplayer2.ui.PlayerView` and `com.google.android.exoplayer2.ui.StyledPlayerView` https://github.com/google/ExoPlayer/blob/release-v2/library/ui/src/main/java/com/google/android/exoplayer2/ui/PlayerView.java

markushi: Oh I see, media3 is an actual replacement for exoplayer, but I'd still add it.

markushi (approved):
Nice!

sentry/src/main/java/io/sentry/SentryReplayOptions.java
romtsn: added both 👍

Issue #2779, opened 2023-06-07
Validate sessions don't create breadcrumbs


stefanosiano:
For more info: https://github.com/getsentry/team-mobile/issues/124

Also, remember to update the docs

PR #3778, opened 2024-10-08
[QA] Fix potential ANRs due to default integrations

markushi:
## :scroll: Description
Analyzed all Android integrations which execute code during app runtime.

Integration | Notes | Needs work?
-- | -- | --
ActivityBreadcrumbsIntegration | creates breadcrumbs, but adds hints with ref to Activity | generates breadcrumbs on main thread
ActivityLifecycleIntegration | creates breadcrumbs, txn, etc | yes, guarded TTFD listeners for now
AnrIntegration | obsolete on older devices | no
AnrV2Integration | already uses executor service | no
AppComponentsBreadcrumbsIntegration | creates breadcrumbs | yes, moved breadcrumb generation to background thread
AppLifecycleIntegration | breadcrumbs, session management | yes, but likely will cause a change in behavior if moved to background
CurrentActivityIntegration | updates CurrentActivityHolder | no
FragmentLifecycleIntegration | TBD | needs more investigation
NetworkBreadcrumbsIntegration | calls registerDefaultNetworkCallback, but this defaults to a background handler anyway, so should be good (see 1, 2) | no
PhoneStateBreadcrumbsIntegration | only relevant when READ_PHONE_STATE permission is granted | maybe, skipping for now
SentryTimberIntegration | shouldn’t be too relevant for release builds | maybe
SystemEventsBreadcrumbsIntegration | creates breadcrumbs | yes, moved processing to bg
UserInteractionIntegration | adds breadcrumbs, does tracing if enabled | needs more investigation

Reduce overhead on main thread.

Also fixes https://github.com/getsentry/sentry-java/issues/2779

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
@@ -382,7 +382,7 @@ public final class ActivityLifecycleIntegration
 
     firstActivityCreated = true;
 
-    if (fullyDisplayedReporter != null) {
+    if (performanceEnabled && ttfdSpan != null && fullyDisplayedReporter != null) {
       fullyDisplayedReporter.registerFullyDrawnListener(() -> onFullFrameDrawn(ttfdSpan));
     }
   }
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AppComponentsBreadcrumbsIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AppComponentsBreadcrumbsIntegration.java
@@ -85,41 +85,20 @@ public final class AppComponentsBreadcrumbsIntegration
   @SuppressWarnings("deprecation")
   @Override
   public void onConfigurationChanged(@NotNull Configuration newConfig) {
-    if (hub != null) {
-      final Device.DeviceOrientation deviceOrientation =
-          DeviceOrientations.getOrientation(context.getResources().getConfiguration().orientation);
-
-      String orientation;
-      if (deviceOrientation != null) {
-        orientation = deviceOrientation.name().toLowerCase(Locale.ROOT);
-      } else {
-        orientation = "undefined";
-      }
-
-      final Breadcrumb breadcrumb = new Breadcrumb();
-      breadcrumb.setType("navigation");
-      breadcrumb.setCategory("device.orientation");
-      breadcrumb.setData("position", orientation);
-      breadcrumb.setLevel(SentryLevel.INFO);
-
-      final Hint hint = new Hint();
-      hint.set(ANDROID_CONFIGURATION, newConfig);
-
-      hub.addBreadcrumb(breadcrumb, hint);
-    }
+    executeInBackground(() -> captureConfigurationChangedBreadcrumb(newConfig));
   }
 
   @Override
   public void onLowMemory() {
-    createLowMemoryBreadcrumb(null);
+    executeInBackground(() -> captureLowMemoryBreadcrumb(null));
   }
 
   @Override
   public void onTrimMemory(final int level) {
-    createLowMemoryBreadcrumb(level);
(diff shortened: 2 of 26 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -382,7 +382,7 @@ public synchronized void onActivityCreated(
 
     firstActivityCreated = true;
 
-    if (fullyDisplayedReporter != null) {
+    if (performanceEnabled && ttfdSpan != null && fullyDisplayedReporter != null) {
```
markushi, line 385: I've added this, as otherwise we'd add a new listener inside FullyDisplayedReporter without ever clearing that list.

sentry-android-core/src/main/java/io/sentry/android/core/NetworkBreadcrumbsIntegration.java
```diff
@@ -63,40 +63,71 @@ public void register(final @NotNull IHub hub, final @NotNull SentryOptions optio
...
 
+    this.options = options;
+
     if (androidOptions.isEnableNetworkEventBreadcrumbs()) {
 
       // The specific error is logged in the ConnectivityChecker method
-      if (buildInfoProvider.getSdkInfoVersion() < Build.VERSION_CODES.LOLLIPOP) {
-        networkCallback = null;
-        logger.log(SentryLevel.DEBUG, "NetworkBreadcrumbsIntegration requires Android 5+");
+      if (buildInfoProvider.getSdkInfoVersion() < Build.VERSION_CODES.N) {
```
markushi, line 71: I've changed this to `N`, as our `AndroidConnectionStatusProvider.registerNetworkCallback` itself only supports >= `N`

sentry-android-core/src/main/java/io/sentry/android/core/NetworkBreadcrumbsIntegration.java
```diff
@@ -63,40 +63,71 @@ public void register(final @NotNull IHub hub, final @NotNull SentryOptions optio
...
+      try {
+        options
+            .getExecutorService()
+            .submit(
+                new Runnable() {
+                  @Override
+                  public void run() {
+                    // in case integration is closed before the task is executed, simply return
+                    final @Nullable NetworkBreadcrumbsNetworkCallback callback = networkCallback;
+                    if (isClosed || callback == null) {
```
romtsn, line 88: was thinking if we should synchronize here, but since we use the same thread (executorService), I guess it should be fine, right?

sentry-android-core/src/main/java/io/sentry/android/core/AppComponentsBreadcrumbsIntegration.java
```diff
@@ -85,41 +85,20 @@ public void close() throws IOException {
...
-      breadcrumb.setCategory("device.orientation");
-      breadcrumb.setData("position", orientation);
-      breadcrumb.setLevel(SentryLevel.INFO);
-
-      final Hint hint = new Hint();
-      hint.set(ANDROID_CONFIGURATION, newConfig);
-
-      hub.addBreadcrumb(breadcrumb, hint);
-    }
+    executeInBackground(() -> captureConfigurationChangedBreadcrumb(newConfig));
```
romtsn, line 88: i'm a little bit worried how this would work with session replay, because we're listening to orientation changes and change the replay orientation, but if the queue is busy here, the breadcrumb will be out-of-sync. Can we maybe keep this breadcrumb on the main thread still?

romtsn (approved):
noice!

sentry-android-core/src/main/java/io/sentry/android/core/AppComponentsBreadcrumbsIntegration.java
markushi: Let me switch to the new breadcrumb ctor with timestamp instead, then it should be fine.

PR #3782, opened 2024-10-10
[QA] Avoid collecting normal frames

stefanosiano:
## :scroll: Description
avoid keeping normal frames in memory when collecting slow/frozen frames

We were keeping a lot of useless frames in memory, without using them. This means that when a span finishes, we were iterating and then cleaning a lot more items than necessary.

Unit tests

## :pencil: Checklist


sentry-android-core/src/main/java/io/sentry/android/core/SentryFrameMetrics.java
```diff
@@ -73,16 +62,14 @@ public long getFrozenFrameDelayNanos() {
   }
 
   public int getTotalFrameCount() {
-    return normalFrameCount + slowFrameCount + frozenFrameCount;
+    return slowFrameCount + frozenFrameCount;
```
romtsn, line 65: but the total frame count would be incorrect now, why are we removing it?

stefanosiano: we are still calculating the normal frames [based on the span duration](https://github.com/getsentry/sentry-java/pull/3782/files#diff-8fb51c60f3cd9eba198cd0702cf5b06a1b4e0ff676dccaa371840a44a5a5bd8eL268).
Actually, previously we were interpolating the number of frames by removing the fast frames duration, too, which could lead to having more total frames than expected. Now it should be more accurate.
E.g. a frame take 2ms to render in a span of 10ms.
Previously we were adding the frames that we were interpolating based on the 8ms (10-2) which was logically wrong. 

Does it make sense?

stefanosiano: > If the list doesn't contain the normal frames anymore, the duration may be considered as a delayed frame. E.g. if a span takes 1000ms and only normal frames happen, the list will be empty and it will be considered a frame delay

That part is skipped if the list is empty, and the whole span is considered without delay

> the method is called getTotalFrameCount but it only returns the sum of slow+frozen

Ok 😅  I'll rename it to `getSlowFrozenFrameCount`

sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
```diff
@@ -277,7 +279,7 @@ private static int interpolateFrameCount(
     final long frameMetricsDurationNanos = frameMetrics.getTotalDurationNanos();
     final long nonRenderedDuration = spanDurationNanos - frameMetricsDurationNanos;
     if (nonRenderedDuration > 0) {
-      return (int) (nonRenderedDuration / frameDurationNanos);
+      return (int) Math.ceil((double) nonRenderedDuration / frameDurationNanos);
```
romtsn, line 282: not sure if `ceil` here now clashes a little with [addPendingFrameDelay](https://github.com/getsentry/sentry-java/blob/fd1151b5c7696d17711c40c2facee553a43d07c3/sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java#L285), I assume it has the same goal (to round the remaining time in a finished span to include one more frame)?

romtsn (approved):
lgtm, assuming Max is fine with that, I have no idea about this whole thing :D

sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java
stefanosiano: It shouldn't clash, because `addPendingFrameDelay` only adds a frame if it's slow.
Also, it considers the last frame duration, which is subtracted in this function (`frameMetrics.getTotalDurationNanos`)

romtsn: mm, but if `addPendingFrameDelay` adds a slow frame, we will also add it here don't we?
Like let's say the span is 50ms, and we have one normal frame and 34ms left which were all consumed by a slow frame. `addPendingFrameDelay` would add 1 frame, but then we come here and add 3 more? Or am I missing something obvious?

stefanosiano: yep

stefanosiano:
@markushi wanna look again?

PR #3801, opened 2024-10-17
[QA] Replace setOf with HashSet.add

stefanosiano:
## :scroll: Description
replaced kotlin setOf() with HashSet<>() + add() calls

setOf() was taking ~10ms to run, while instantiating an HashSet and then calling add multiple times is almost instantaneous - at least in debug builds

Unit test

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Fixes
 
+- Replace setOf with HashSet.add ([#3801](https://github.com/getsentry/sentry-java/pull/3801))
 - Cache parsed Dsn ([#3796](https://github.com/getsentry/sentry-java/pull/3796))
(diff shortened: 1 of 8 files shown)


adinauer (approved):

bendb-instacart:
I, a totally random bystander, am intrigued by the performance difference.  I see that `someEnum.values().toSet()` is replaced with, essentially, `HashSet().apply { add(ONE); add(TWO); ...}`.

Is `toSet()` the big difference, or maybe is it the call to `values()`?  IIRC the latter involves some (one-time) reflection.  As far as I can see, `toSet()` boils down to `Iterable<T>.toCollection(LinkedHashSet<T>())`, which in turn is just `for (elem in this) { destination.add(elem) }`, more or less equivalent to the manual HashSet stuff here.

Not that it really matters of course, since you've found a solution that works for sentry-java, but I'm curious all the same.  Feel free to ignore.

PR #3844, opened 2024-11-04
Add new APIs for Continuous Profiling v8 (p5)

stefanosiano:
## :scroll: Description
added Sentry.startProfiler and Sentry.stopProfiler APIs
AndroidContinuousProfiler now checks the scopes (and options) on profile start

#skip-changelog

Implements user APIs for Continuous Profiling
Part 5 of https://github.com/getsentry/sentry-java/pull/3710

## :pencil: Checklist


sentry/src/main/java/io/sentry/Sentry.java
```diff
@@ -1022,6 +1022,16 @@ public static void endSession() {
     return getCurrentScopes().startTransaction(transactionContext, transactionOptions);
   }
 
+  /** Starts the continuous profiler, if enabled. */
+  public static void startProfiler() {
```
romtsn, line 1026: wdyt of introducing a subclass (like a namespace) `Sentry.profiling().start()` akin to `Sentry.metrics()`? We're also gonna do that for replay soon

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
```diff
@@ -88,12 +90,14 @@ private void init() {
...
 
-  public synchronized void setScopes(final @NotNull IScopes scopes) {
-    this.scopes = scopes;
-    this.performanceCollector = scopes.getOptions().getTransactionPerformanceCollector();
-  }
-
   public synchronized void start() {
+    if ((scopes == null || scopes != NoOpScopes.getInstance())
+        && Sentry.getCurrentScopes() != NoOpScopes.getInstance()) {
+      this.scopes = Sentry.getCurrentScopes();
```
romtsn, line 96: in theory you could pass `Scopes` as an argument here, because this method is called from within `Scopes`

romtsn (approved):

sentry/src/main/java/io/sentry/Sentry.java
stefanosiano: We want to keep the SDKs aligned. Other SDKs, like cocoa, is doing `SentrySDK.startProfiler`

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
stefanosiano: it is also called in the `SentryPerformanceProvider`, when it's used as app start profiler. That's why i'm retrieving the scopes whenever I can

PR #3854, opened 2024-11-06
[SR] Disable replay in session mode when rate limit is active

romtsn:
## :scroll: Description

* Add mappings for DataCategory <-> ItemType to make replays respect rate-limiting
* Check if rate-limiting is active for Replays or All before recording a screenshot in session mode to not consume disk space. In session mode we capture segments every 5s so if a rate limit is active for e.g. 10 minutes we'd quickly overflow the envelope cache with replay segments. In buffer mode, however, it's still fine to capture the replay because it'll be only one envelope that we store if an error occurs.
* Had to observer rate-limits and connections status to properly resume/pause the replay and update timestamps
	* Had to introduce `IRateLimitObserver` that notifies whenever any of the rate limits is applied/lifted

Manually + automated

## :pencil: Checklist


sentry-android-replay/src/main/java/io/sentry/android/replay/capture/SessionCaptureStrategy.kt
```diff
@@ -78,6 +80,12 @@ internal class SessionCaptureStrategy(
             bitmap?.recycle()
             return
         }
+        val rateLimiter = hub?.rateLimiter
+        if (rateLimiter?.isActiveForCategory(All) == true || rateLimiter?.isActiveForCategory(Replay) == true) {
+            options.logger.log(DEBUG, "Skipping screenshot recording, rate limit is active")
+            bitmap?.recycle()
+            return
+        }
```
philipphofmann, line 88: @brustolin and @romtsn I think this is a different behaviour than on iOS https://github.com/getsentry/sentry-cocoa/pull/4496. On iOS, when a limit for SR gets active, we drop everything until the next session starts. Please align.

brustolin: Yes, in our last replay meeting, we agreed to start with the basics and stop session replay altogether for the remainder of the app session.

philipphofmann: >i'd prefer to do our best and start capturing from when the rate-limit is lifted?

That's also my opinion. This would also simplify the logic.

brustolin: But your current logic is not taking into account that the envelope with segment 0 could be the one that was lost. It does not matter anymore so send to following segments, they all will be ignored in the server.

brustolin: Btw, this approach is just throwing away the already-taken screenshot. Maybe Android is different; for iOS, making the screenshot is the most expensive part of the process.

brustolin: Another point to take into account, someone can correct me on this one, is that the most common form of rate limiting is quota exceeded. And this will only be lifted at the end of the month.

romtsn: > But your current logic is not taking into account that the envelope with segment 0 could be the one that was lost. It does not matter anymore to send to following segments, they all will be ignored in the server.

Why would it be lost? We're not capturing envelopes if rate limit is active right? Or do you mean if we get rate-limited upon sending the segment 0 event? In this case on Android it'd be stored in cache until rate-limit is lifted. The only case it gets lost if the cache gets rotated, but it's an edge-case enough to not handle it for now I guess.

> Btw, this approach is just throwing away the already-taken screenshot. Maybe Android is different; for iOS, making the screenshot is the most expensive part of the process.

That's a good point, I should probably move this logic to where we capture the screenshot 👍 

> Another point to take into account, someone can correct me on this one, is that the most common form of rate limiting is quota exceeded. And this will only be lifted at the end of the month.

Probably yes, but not sure about the end of the month (could be on-demand events etc etc).  I think this is also mostly relevant for big customers who have spike protection and whatnot in place, so we shouldn't think in "months" here probably :)

philipphofmann: > And this will only be lifted at the end of the month.

Nope, spike protection works on an hour basis: https://docs.sentry.io/pricing/quotas/spike-protection/#how-sentry-detects-spikes

philipphofmann: I don't have enough insight into SR. I let you two decide this.

sentry/src/main/java/io/sentry/transport/RateLimiter.java
```diff
@@ -301,4 +328,39 @@ private long parseRetryAfterOrDefault(final @Nullable String retryAfterHeader) {
...
+
+  public interface IRateLimitObserver {
+    /**
+     * Invoked whenever the rate limit changed. You should use {@link
+     * RateLimiter#isActiveForCategory(DataCategory)} to check whether the category you're
+     * interested in has changed.
+     *
+     * @param applied true if the rate limit was applied, false if it was lifted
+     */
+    void onRateLimitChanged(boolean applied);
```
stefanosiano, line 364: you suggest to use `RateLimiter.isActiveForCategory(DataCategory)` to check the limit
What's the reason of `applied` param? you don't seem to use it. 
Instead, we can pass the `RateLimiter` directly as parameter, so that you don't have to call the hub and handle nullable values on the call site. Wdyt?

sentry/src/main/java/io/sentry/transport/RateLimiter.java
```diff
@@ -301,4 +328,39 @@ private long parseRetryAfterOrDefault(final @Nullable String retryAfterHeader) {
...
+  }
+
+  @Override
+  public void close() throws IOException {
+    synchronized (timerLock) {
+      if (timer != null) {
+        timer.cancel();
+        timer = null;
+      }
+    }
```
markushi, line 353: Probably makes sense to clear the list of observers, right?
```suggestion
        rateLimitObservers.clear();
    }
```

markushi (approved):

PR #3855, opened 2024-11-06
[SR] Add beforeSendReplay callback

romtsn:
## :scroll: Description

* Add `beforeSendReplay` callback to allow attaching tags and modifying/dropping events analogous to `beforeSend`

Automated

## :pencil: Checklist

- [] I updated the docs if needed.

## :crystal_ball: Next steps
Docs


sentry/src/main/java/io/sentry/SentryClient.java
```diff
@@ -285,6 +285,17 @@ private void finalizeTransaction(final @NotNull IScope scope, final @NotNull Hin
 
     event = processReplayEvent(event, hint, options.getEventProcessors());
 
+    if (event != null) {
```
stefanosiano, line 288: I'd move this block after the "dropped by event processor" log (L302) to be more readable imho
Also, as it is now, if the replay is dropped by beforeSendReplay, we log `Event was dropped by beforeSendReplay` and right after that `Replay was dropped by Event processors.` (it's missing `return SentryId.EMPTY_ID;`)

sentry/src/main/java/io/sentry/SentryClient.java
```diff
@@ -285,6 +285,17 @@ private void finalizeTransaction(final @NotNull IScope scope, final @NotNull Hin
...
+      if (event == null) {
+        options.getLogger().log(SentryLevel.DEBUG, "Event was dropped by beforeSendReplay");
+        options
+            .getClientReportRecorder()
+            .recordLostEvent(DiscardReason.BEFORE_SEND, DataCategory.Replay);
+      }
+    }
+
     if (event == null) {
       options.getLogger().log(SentryLevel.DEBUG, "Replay was dropped by Event processors.");
```
stefanosiano, line 300: should we add a client report here, with a EventProcessor reason?

romtsn: we already do that in [processReplayEvent](https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/SentryClient.java#L530)

sentry/src/main/java/io/sentry/SentryClient.java
romtsn: Ok, I just removed the log line for now, because we're already logging it in `processReplayEvent` anyway, so this log was excessive. I'll keep the order as-is, as it reflects what we have in `captureEvent` as well

stefanosiano (approved):

PR #3926, opened 2024-11-22
Add rate limit for Continuous Profiling v8 (p6)

stefanosiano:
## :scroll: Description
continuous profiler now doesn't start if offline or rate limited
continuous profiler stops when rate limited
continuous profiler prevents sending chunks after being closed
added profile_chunk rate limit

#skip-changelog

Implements rate limit for Continuous Profiling
Part 6 of https://github.com/getsentry/sentry-java/pull/3710

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
@@ -25,6 +25,8 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.Future;
 import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.atomic.AtomicBoolean;
+
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
@@ -50,7 +52,7 @@ public class AndroidContinuousProfiler
   private final @NotNull List<ProfileChunk.Builder> payloadBuilders = new ArrayList<>();
   private @NotNull SentryId profilerId = SentryId.EMPTY_ID;
   private @NotNull SentryId chunkId = SentryId.EMPTY_ID;
-  private boolean isClosed = false;
+  private @NotNull AtomicBoolean isClosed = new AtomicBoolean(false);
 
   public AndroidContinuousProfiler(
       final @NotNull BuildInfoProvider buildInfoProvider,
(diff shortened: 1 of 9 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
```diff
@@ -45,6 +50,7 @@ public class AndroidContinuousProfiler implements IContinuousProfiler {
   private final @NotNull List<ProfileChunk.Builder> payloadBuilders = new ArrayList<>();
   private @NotNull SentryId profilerId = SentryId.EMPTY_ID;
   private @NotNull SentryId chunkId = SentryId.EMPTY_ID;
+  private boolean isClosed = false;
```
romtsn, line 53: should we make it `AtomicBoolean` ?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
```diff
@@ -27,7 +31,8 @@
 import org.jetbrains.annotations.VisibleForTesting;
 
 @ApiStatus.Internal
-public class AndroidContinuousProfiler implements IContinuousProfiler {
+public class AndroidContinuousProfiler
+    implements IContinuousProfiler, RateLimiter.IRateLimitObserver {
```
romtsn, line 35: do you also want to observe network connectivity actually? In case the device gets offline in the middle of a profile running, we'd still overflow the cache I guess?

romtsn (approved):
A couple of comments, otherwise LGTM!

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
stefanosiano: not sure, as there's only close() setting it, and it's already synchronized
but i can add it

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
stefanosiano: I do the check when each profile chunk starts. 
I just want to avoid the case where the device goes offline for 1 second and then reconnects. In this case we would drop the whole profile
And the overhead is low enough not to "care" about stopping early when device is offline

romtsn: ah yeah, I forgot that we call start() continuously, that makes sense then!

Issue #3899, opened 2024-11-18
Not all warm start cases covered by SDK?


zhukic:
### Problem Statement

Hi!

Sentry SDK version - `7.14.0`

Currently, SDK reports the warm start only in case when an application was launched from background state and activity is created with non-null bundle([how to reproduce it](https://github.com/getsentry/sentry-java/issues/1950#issuecomment-1067587818)).

But, according to the [documentation](https://developer.android.com/topic/performance/vitals/launch-time#warm) there are another cases like `The user backs out of your app but then re-launches it. The process might continue to run, but the app must recreate the activity from scratch using a call to onCreate().`

![Image](https://github.com/user-attachments/assets/3f46a5a4-9a9c-46d2-b42b-a5f3571980e8)

For example, when an application receives a push notification, the process and application are initialized; if we launch the app after that, only activity is created, and it looks like a case for the warm start, but the SDK doesn't report it.

Another case(but I'm not sure if it's a really warm start) - the app goes to the background, the system kills all activities, and then we launch the app. In this case, the process still exists, and only activity is created(can be reproduced with "Don't keep activities" enabled).

We see that the number of warm starts reported by SDK is much less than the number of starts we see in Google Play Console, so covering all such possible cases should reduce the difference.

What do you think?

markushi:
@zhukic Thanks for reporting, it definitely looks like we're under-reporting warm app starts. We'll take a look into this.

markushi:
Instead of dropping warm app starts with older process inits, we should create a new app start transaction once the warm start happens and properly date it back to the second app launch time.

The papa library provides some good reference for the different warm app start types: https://github.com/square/papa/blob/be9dd6d466fbc2de8722aa83d80bfcf6b17317c5/papa/src/main/java/papa/PreLaunchState.kt#L39-L58

Issue #3896, opened 2024-11-18
Initial display time is reported incorrectly


zhukic:
### Integration

sentry-android

### Build System

Gradle

### AGP Version

8.7.2

### Proguard

Enabled

### Version

7.14.0

### Steps to Reproduce

1. Enable "Don't keep activities" in developer options.
2. Open app
3. Leave the app
4. Open the app again in 20 seconds 

### Expected Result

Correct `ui.load` span

### Actual Result

`ui.load` span duration is more than 20 seconds

![Image](https://github.com/user-attachments/assets/f8797a72-50ae-4f5e-b537-e497dbb2a7ac)

As I understand, the root cause is [here](https://github.com/getsentry/sentry-java/blob/ce09ad4523b10ddaa0df90c12c17fb0b7a89bf78/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java#L222C25-L222C31) - when an activity gets destroyed in the background its paused time is used as the start time for the next `ui.load` transaction

markushi:
@zhukic Thanks for reporting, we'll have a closer look at this.

PR #3937, opened 2024-11-28
Warm start detection

stefanosiano:
## :scroll: Description
* ActivityLifecycleIntegration:
  - creates `onCreate` and `onStart` TimeSpans
  - set app start type to warm in AppStartMetrics when needed
* AppStartMetrics has now a method to restart appStartSpan and reset its uptime_ms
* PerformanceAndroidEventProcessor now attaches activity start spans to warm starts, too
* SentryPerformanceProvider doesn't create spans anymore

Confirm warm starts are registered for background starts (BroadcastReceiver)

Cold start:
Cold start
Open second Activity:
open second activity
Reopen first Activity without closing app:
reopen first activity
Open first Activity after close app with `Don't keep activity` developer option enabled:
open after close and no keep activity

Fixes https://github.com/getsentry/sentry-java/issues/3896
Relates and possibly fixes https://github.com/getsentry/sentry-java/issues/3899

## :pencil: Checklist

## :crystal_ball: Next steps

In a [followup PR](https://github.com/getsentry/sentry-java/pull/3954):
- Remove the activityCallback completely from the `SentryPerformanceProvider`
- Create and attach spans directly in `ActivityLifecycleIntegration` instead of adding them to `AppStartMetrics` and let the `PerformanceAndroidEventProcessor` add them to the transaction
- Avoid creating process init span using `setStartUnixTimeMs` as it is a testOnly method. This means care, as the method is exposed to hybrid sdks

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
@@ -429,13 +429,12 @@ public final class ActivityLifecycleIntegration
 
   @Override
   public void onActivityPreStarted(final @NotNull Activity activity) {
-    final long now = SystemClock.uptimeMillis();
     if (appStartSpan == null) {
       return;
     }
     final @Nullable ActivityLifecycleTimeSpan timeSpan = activityLifecycleMap.get(activity);
     if (timeSpan != null) {
-      timeSpan.getOnStart().setStartedAt(now);
+      timeSpan.getOnStart().setStartedAt(SystemClock.uptimeMillis());
     }
   }
 
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
@@ -672,6 +672,16 @@ public final class ActivityLifecycleIntegration
   }
 
   @TestOnly
+  @NotNull WeakHashMap<Activity, ActivityLifecycleTimeSpan> getActivityLifecycleMap() {
+    return activityLifecycleMap;
+  }
+
+  @TestOnly
+  void setFirstActivityCreated(boolean firstActivityCreated) {
+    this.firstActivityCreated = firstActivityCreated;
+  }
+
+  @TestOnly
   @NotNull
   ActivityFramesTracker getActivityFramesTracker() {
     return activityFramesTracker;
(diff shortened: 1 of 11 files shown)


romtsn (approved):
LGTM, but I'm a bit rusty in this part of the codebase already, would love @markushi to take a look

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -387,8 +412,39 @@ public synchronized void onActivityCreated(
...
+  @Override
+  public void onActivityPreStarted(final @NotNull Activity activity) {
+    final long now = SystemClock.uptimeMillis();
+    if (appStartSpan == null) {
+      return;
+    }
+    final @Nullable ActivityLifecycleTimeSpan timeSpan = activityLifecycleMap.get(activity);
+    if (timeSpan != null) {
+      timeSpan.getOnStart().setStartedAt(now);
+    }
```
markushi, line 439: `nit`: `SystemClock.uptimeMillis();` could be moved inside the if-check
```suggestion
    if (appStartSpan == null) {
      return;
    }
    final @Nullable ActivityLifecycleTimeSpan timeSpan = activityLifecycleMap.get(activity);
    if (timeSpan != null) {
      timeSpan.getOnStart().setStartedAt(SystemClock.uptimeMillis());
    }
```

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -400,74 +456,75 @@ public synchronized void onActivityStarted(final @NotNull Activity activity) {
...
+    if (!isAllActivityCallbacksAvailable) {
+      onActivityPostStarted(activity);
+    }
     if (performanceEnabled) {
-
       final @Nullable ISpan ttidSpan = ttidSpanMap.get(activity);
       final @Nullable ISpan ttfdSpan = ttfdSpanMap.get(activity);
-      final View rootView = activity.findViewById(android.R.id.content);
-      if (rootView != null) {
+      if (activity.getWindow() != null) {
```
markushi, line 480: Do we still need this check? Since we have a reference to the activity, we could always call `irstDrawDoneListener.registerForNextDraw`

markushi (approved):
Looking good! Nice job! Since `SentryPerformanceProvider` doesn't create any spans anymore it's worth mentioning that the app start spans will only be created if the SDK gets initialized before the first activity is launched. Maybe it's worth adding this to our changelog?

stefanosiano:
> Looking good! Nice job! Since `SentryPerformanceProvider` doesn't create any spans anymore it's worth mentioning that the app start spans will only be created if the SDK gets initialized before the first activity is launched. Maybe it's worth adding this to our changelog?

i think this was the case before, too, as the cold start span was created by the `ActivityLifecycleIntegration` anyway, and the `SentryPerformanceProvider` spans were only added to it.
We guard the app start span creation with the `firstActivityCreated` flag, so there should be no app start span if the SDK is initialized after the first activity.
Or am i missing a case here?

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
stefanosiano: It's a safety guard: if the window is null (i'm not sure it ever happens), the `FirstDrawDoneListener.registerForNextDraw` returns early, and the ttid is never stopped.
In this case i just post the ttid finish on the main thread, as it should be good enough

PR #4001, opened 2024-12-18
[SR] Fix ANRs and speed up

### LINKED ISSUE #3693: [SR] ANRs in Session Replay
### Description

There are a few ANRs reported in the google play console, namely:
- `ScreenshotRecorder` ANR on Samsung devices with Android < 10 https://play.google.com/sdk-console/u/0/accounts/3062283563667932501/sdks/5488485728742614670/crashes/cff428d7/
- `BaseCaptureStrategy.onTouchEvent` ANR where we hold a lock over `currentEvents` (should probably be moved to a bg thread this whole thing) https://play.google.com/sdk-console/u/0/accounts/3062283563667932501/sdks/5488485728742614670/crashes/e62fe147/

romtsn:
## :scroll: Description

* Switch touch events from `LinkedList` with a lock to `ConcurrentLinkedDeque` which suits this case better. Iterators of this deque are [weakly consistent](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ConcurrentLinkedQueue.html), which means it's possible another thread will add new touch events while we're iterating but this is fine for us, because we filter them out based on timestamp upon sending, so the newly-added touch events won't make it to the current segment, but to the next one.
* Get rid of storing touch events to disk (for now), because it was overloading cpu a lot, and we only need them for the last segment in case of an ANR or Crash. iOS is currently not storing them either, so I think it's fine. But we should do it eventually, after we have proper disk serialization support (like serialize a single touch event, instead of the entire collection everytime)
* Pre-allocate some arraylists for GestureRecorder to avoid growing them on the main thread
* Get rid of one (out of 4) background threads in replay to avoid thread context switching. We reuse the same executor for masking screenshots that we use for creating segments/etc. My plan is to eventually have only 1 (or maybe 2) threads for the alll the replay stuff, but that's for followup PRs.

Fixes #3693 (partially, but not sure how to fix the first ANR there, so let's close it for now until we get actual reports about that)

Manually + automated

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Change TTFD timeout to 25 seconds ([#3984](https://github.com/getsentry/sentry-java/pull/3984))
 - Session Replay: Fix memory leak when masking Compose screens ([#3985](https://github.com/getsentry/sentry-java/pull/3985))
+- Session Replay: Fix potential ANRs in `GestureRecorder` ([#4001](https://github.com/getsentry/sentry-java/pull/4001))
 
 ## 7.19.0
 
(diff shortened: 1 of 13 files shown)


stefanosiano (approved):
looks good!

PR #4016, opened 2025-01-02
[SR] Reduce memory and disk consumption

romtsn:
## :scroll: Description

Squeezed various improvements here to not open multiple PRs, but they are mostly around memory/allocations/size:

- Reduce JPEG quality that we store to disk, it's also now part of the `quality` option, default is 30. This reduced the payload size even further and saved disk space up to ~40%. It's still of a good quality though for the end user, you can check [this replay](https://sentry-sdks.sentry.io/replays/42b01b634d2c49b6995bbedcf2e785f4/?project=4507596011798528&query=&referrer=%2Freplays%2F%3AreplaySlug%2F&statsPeriod=14d&yAxis=count%28%29&t=0).
- Change bitmap config from ARGB_8888 to RGB_565 as we don't really care about the alpha channel here.
- Delete a frame screenshot if we were not able to decode it right away instead of trying to decode the file again and again.
- Use only one bitmap as opposed to two + don't copy that one when passing it to replay cache - it's not necessary as we just flush its bytes to a file, but never recycle (only when the recording is stopped). It also works fine with multiple screenshots to the same bitmap, as PixelCopy would just overwrite it. Here's the resulting allocations (reduced the number of allocations and also the size by ~30%):

### Before
Android Studio 2025-01-02 10 47 15

### After
Android Studio 2025-01-02 10 21 01

Manually + automated

## :pencil: Checklist

### DIFF
--- a/.github/workflows/agp-matrix.yml
+++ b/.github/workflows/agp-matrix.yml
@@ -38,7 +38,7 @@ jobs:
           java-version: '17'
 
       - name: Setup Gradle
-        uses: gradle/actions/setup-gradle@4ba34e96c5f6493e99d0696180a9a8d431577ba9 # pin@v3
+        uses: gradle/actions/setup-gradle@3839b20885c2c3507be5f0521853826f4b37038a # pin@v3
         with:
           gradle-home-cache-cleanup: true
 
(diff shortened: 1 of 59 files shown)


stefanosiano (approved):
nice!

PR #4025, opened 2025-01-07
Create onCreate and onStart spans for all Activities

stefanosiano:
## :scroll: Description
ActivityLifecycleIntegration creates activity spans for all Activities, not only for appStart ones.

We are currently adding `onCreate` and `onStart` spans only in the app start transaction. This PR will add these spans to all automatic transactions, too

Unit test

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - The Kotlin Language version is now set to 1.6 ([#3936](https://github.com/getsentry/sentry-java/pull/3936))
 
+### Features
+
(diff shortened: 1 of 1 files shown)


stefanosiano:
Just a note:
In the app start transaction the spans are children of app.start.cold or app.start.warm span
For other Activities the spans are direct children of the parent transaction.

Also, this works for automatic transactions only. If we want to instrument custom transactions, we can easily use Sentry.getSpan instead of the transaction reference.
Any though on this?

romtsn (approved):

romtsn:
removed "closes" because that issue was about adding onResume spans which this PR does not

PR #3823, opened 2024-10-25
Accept manifest integer values when requiring floating values

stefanosiano:
## :scroll: Description
ManifestMetadataReader now accepts integers other than floats

## :bulb: Motivation and Context
Fixes https://github.com/getsentry/sentry-java/issues/3603

## :green_heart: How did you test it?
Unit test

## :pencil: Checklist

- [ ] I reviewed the submitted code.
- [ ] I added tests to verify the changes.
- [ ] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] Review from the native team if needed.
- [ ] No breaking change or entry added to the changelog.
- [ ] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

PR #4031, opened 2025-01-08
Avoid logging an error when a float is passed in the manifest

stefanosiano:
## :scroll: Description
Avoid logging an error when a float is passed in the manifest

https://github.com/getsentry/sentry-java/pull/3823 added support for integer values in the manifest.
But that's used as default value when parsing floats. This means that all float values are evaluated as integers and floats, causing one of the checks to fail.
Everything still works correctly, but when the type is wrong, the error is logged in the logcat.
This PR avoids the log if the value passed in the manifest is a float

Closes https://github.com/getsentry/sentry-android-gradle-plugin/issues/803

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Enable `ThreadLocalAccessor` for Spring Boot 3 WebFlux by default ([#4023](https://github.com/getsentry/sentry-java/pull/4023))
 
+### Fixes
+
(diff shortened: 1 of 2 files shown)


romtsn (approved):

romtsn:
@stefanosiano @adinauer wanted to hold off with merging PRs to v8, so you could wait until next week (or convince Alex :)) to merge the PR.

PR #4039, opened 2025-01-10
Various fixes to instrumentations running on the main thread

romtsn:
## :scroll: Description

Here are just some preventive measure that will make less code/instructions to run on the main thread:
- Safety check for the `Window.Callback` to not be a `SentryWindowCallback` already to prevent multiple callbacks running simultaneously
- Get rid of `requireNonNull` calls as they are redundant (we do `instanceof` checks later on anyway)
- Check if tracing is enabled before instrumenting File I/O streams, because otherwise they go into the void anyway
- Speed up ViewUtils.findTarget method by avoiding ViewGroup children traversal if the touch was not within the ViewGroup bounds

### Before

Screenshot 2025-01-13 at 23 36 38

### After
 
Screenshot 2025-01-14 at 00 09 42

Internal customer discussions

automated

## :pencil: Checklist


sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
```diff
@@ -127,28 +128,42 @@ public static final class Factory {
...
         final @Nullable File file,
         final @NotNull IScopes scopes)
         throws FileNotFoundException {
-      return new SentryFileInputStream(init(file, delegate, scopes));
+      return isTracingEnabled(scopes)
+          ? new SentryFileInputStream(init(file, delegate, scopes))
+          : delegate;
+    }
+
+    private static boolean isTracingEnabled(final @NotNull IScopes scopes) {
```
stefanosiano, line 164: you can just use `ScopesAdapter.getInstance()` here, without passing it as parameter, as it is always the same.
Same for `SentryFileOutputStream`

sentry/src/test/java/io/sentry/instrumentation/file/SentryFileInputStreamTest.kt
```diff
@@ -274,6 +276,15 @@ class SentryFileInputStreamTest {
         assertEquals(false, fileIOSpan.data[SpanDataConvention.BLOCKED_MAIN_THREAD_KEY])
         assertNull(fileIOSpan.data[SpanDataConvention.CALL_STACK_KEY])
     }
+
+    @Test
+    fun `when tracing is disabled does not instrument the stream`() {
```
stefanosiano, line 281: should we add the same test in `SentryFileOutputStreamTest`?

stefanosiano (approved):
just small details, but nice

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
romtsn: nope, we gotta use the parameter one because we pass a mocked one in tests.

stefanosiano: i don't get it
You pass `ScopesAdapter.getInstance()` when you call this function, and there's no need to pass a Scopes in the tests.
Or are you going to add a new test?

romtsn: we pass it here right? https://github.com/getsentry/sentry-java/pull/4039/files#diff-aca0bdac62da1139f1a2196522ff6b0d33f9e700797b88b6c8de04f6bd14b134L72

sentry/src/test/java/io/sentry/instrumentation/file/SentryFileInputStreamTest.kt
romtsn: yeah I can add one!

sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
stefanosiano: i missed that one 😅

stefanosiano (approved):

romtsn:
Superseded by #4051

PR #4052, opened 2025-01-15
Fix BroadcastReceivers

### LINKED ISSUE #1774: deprecate `android.telephony.PhoneStateListener` and use `TelephonyCallback`
https://developer.android.com/reference/android/telephony/TelephonyCallback replaces `PhoneStateListener`

ideally `PhoneStateBreadcrumbsIntegration` conditionally use one or the other depending on OS version

### LINKED ISSUE #2395: Consider removing TempSensorBreadcrumbsIntegration from default integrations
### Description

We had a few reports about some Chinese stores flagging this integration as violating their policies, and therefore not allowing the apps to the stores.

* https://github.com/getsentry/sentry-java/issues/2355
* https://github.com/getsentry/sentry-java/discussions/2393

We could consider removing it from default integrations if the impact is high, or we could introduce a flag which specifically disables this integration (at the moment it sits behind `enableSystemBreadcrumbs` flag, which either disabled or enables _all_ integrations related to system events)

romtsn:
## :scroll: Description

* Removes `TempSensor` and `PhoneState` breadcrumbs integrations, as those don't seem to be very useful and potentially register for too many broadcasts that sometimes cannot be delivered. We plan to introduce a new integration that would listen to the ThermalState changes instead.
* Reduces the number of actions that we subscribe for in `SystemEvents` breadcrumbs integration. Some of those seem to be not helpful, but there are some that are sketchy, like [ACTION_REBOOT](https://developer.android.com/reference/android/content/Intent#ACTION_REBOOT) (only for use of system apps) or [ACTION_POWER_CONNECTED](https://developer.android.com/reference/android/content/Intent#ACTION_POWER_CONNECTED) (which will wake the app up)
* Changed receiver flag to be `RECEIVER_NOT_EXPORTED`. If you check the [ContextCompat docs](https://developer.android.com/reference/androidx/core/content/ContextCompat#registerReceiver(android.content.Context,android.content.BroadcastReceiver,android.content.IntentFilter,java.lang.String,android.os.Handler,int)), it says RECEIVER_NOT_EXPORTED is enough to receive broadcasts from the system. I tested it locally with a script that runs over the actions we're subscribed for and send them via `adb shell` on API 33, and it worked fine with NOT_EXPORTED.

The event with all of the broadcast breadcrumbs that we listen to now can be found here: https://sentry-sdks.sentry.io/issues/5990712373/events/dc048d7441d7496f99f09b1c38c56f7e/

Closes #2395 
Closes #1774 

Internal customer reports. Looks like when we're registered for too many broadcasts, the OS will kill the app process silently if it can't deliver a broadcast, so we're trying to reduce the likelihood of that by reducing the number of broadcasts we're listening to, just keeping the essential ones.

See here: https://cs.android.com/android/platform/superproject/+/android14-qpr3-release:frameworks/base/services/core/java/com/android/server/am/BroadcastQueueModernImpl.java;l=1166-1167?q=broadcastqueuemod

Manually (on API 33+) and automated

## :pencil: Checklist

## :crystal_ball: Next steps
We may need to add docs around SystemEventsIntegration to show off how to add more actions. For now I'm mentioning it in the changelog


CHANGELOG.md
```diff
@@ -1,13 +1,31 @@
...
-## Unreleased
+### Unreleased
 
 ### Fixes
 
 - Do not instrument File I/O operations if tracing is disabled ([#4051](https://github.com/getsentry/sentry-java/pull/4051))
 - Do not instrument User Interaction multiple times ([#4051](https://github.com/getsentry/sentry-java/pull/4051))
 - Speed up view traversal to find touched target in `UserInteractionIntegration` ([#4051](https://github.com/getsentry/sentry-java/pull/4051))
 
+### Changes
```
stefanosiano, line 11: I just don't like too generic "changes"
```suggestion
### Behavioural Changes
```

stefanosiano (approved):
Nice!

PR #4058, opened 2025-01-16
Reduce the number of IPC calls

romtsn:
## :scroll: Description

* Most of the changes I've done were mostly around caching values that are not going to change during the runtime of a process (like device name or app version code).
* Introduced AndroidLazyEvaluator that is like LazyEvaluator but gets Context as an argument

### Before (Binder calls on the main thread)
Screenshot 2025-01-15 at 22 50 58

### After (Binder calls on the main thread)
Code 2025-01-16 23 37 58

Finding Binder calls with [trace-ipc](https://developer.android.com/topic/performance/vitals/render#scheduling-delays) and trying to reduce them to prevent main thread blockage potentially and other problems.

Manually + automated

## :pencil: Checklist

## Next steps

There's still one standing out binder call left around checking network connectivity but this will do in a follow-up PR


stefanosiano (approved):
looks good
In case it will ever be needed again, we could change the LazyEvaluator to have a generic parameter in the `getValue()`. 
That way we could even pass a `Pair<Context, BuildInfoProvider>` so that we don't need multiple evaluators for different API levels.
Not sure if it's worth it now, but let's keep it in mind in the future

PR #4082, opened 2025-01-21
Fix swallow NDK loadLibrary errors

markushi:
## :scroll: Description

Restores the same behaviour we had in 7.x.x: 
https://github.com/getsentry/sentry-java/blob/1e796d816d8218e33084e02800c7485dcdd5b722/sentry-android-ndk/src/main/java/io/sentry/android/ndk/SentryNdk.java#L29-L32

## :pencil: Checklist


adinauer (approved):
LGTM! If you add this to the changelog, please make sure to duplicate the entry into the section at the bottom where we list all changes that happened since `rc.4`.

Issue #3925, opened 2024-11-21
[Android] Add internal hybrid sdk api for fetching debug images that were referenced in the stacktrace


buenaflor:
### Problem Statement

In Flutter when loading debug images on Android we fetch all (up to 82) but most of the time only `libapp.so` and `libflutter.so` is needed so it makes sense to only fetch images that are referenced in the stacktrace

### Solution Brainstorm

Similar to cocoa: https://github.com/getsentry/team-mobile/issues/158

Implementation how we access it in Cocoa on Flutter: https://github.com/getsentry/sentry-dart/blob/8549081071fd2bf85331d9988710b968ced8b2f1/flutter/ios/Classes/SentryFlutterPluginApple.swift#L280C1-L307C1

buenaflor:
@markushi @romtsn I'd like to assign this issue to me if you're ok with it. anything specific I have to consider?

romtsn:
@buenaflor sure, take it all! I don't think there's anything specific (I don't know the native part that well though), from what I know the images list will be loaded lazily and cached in the native layer, and then you can add some methods to filter out only the necessary ones for you.

PR #4089, opened 2025-01-21
feat(android-ndk): add api for getting debug images by addresses

buenaflor:
question: can I push this change onto a v7 branch? We cannot bump to Android v8 at this moment because we also depend on min android api 19 which has been bumped to 21 here. if not possible we'll mark this as a Flutter v9 feature

## :scroll: Description

Adds debug images loader API for getting images by addresses

On Flutter we are currently loading all available debug images which can sometimes range up to 300 images but in most cases only 1 is actually used so this is much more efficient and the envelope is therefore smaller

Closes https://github.com/getsentry/sentry-java/issues/3925

Unit test, e2e test with Flutter app

## :pencil: Checklist


sentry-android-ndk/src/main/java/io/sentry/android/ndk/DebugImagesLoader.java
```diff
@@ -75,7 +77,74 @@ public DebugImagesLoader(
...
+    }
+  }
+
+  /**
+   * Finds a debug image containing the given address using binary search. Requires the images to be
+   * sorted.
+   *
+   * @return The matching debug image or null if not found
+   */
+  private @Nullable DebugImage findImageByAddress(long address, List<DebugImage> images) {
```
markushi, line 121: Is there a specific reason you're assuming that the list of debug images is sorted? I [see we have quite a few different implementations](https://github.com/getsentry/sentry-native/tree/master/src/modulefinder), but I don't think there's any guarantee here.

buenaflor: pretty much from testing manually it was always sorted on my end

but tbh the performance gains of using binary search are marginal at best since the number of images are not that big, I wouldn't mind switching the impl to just search for it normally to make it more stable if we're not confident it's always sorted

markushi: Alright I just talked to Mischan, on Android/Linux based flavors the list should always be sorted. I still made some adaptions to account for a nullable `image.getImageSize()`

buenaflor:
tested it on Flutter and looks good 👍

markushi (approved):

romtsn (approved):
LGTM, but can we please add a changelog entry

stefanosiano (approved):

PR #4137, opened 2025-02-04
feat: Propagate trace to NDK

bitsandfoxes:


sentry-android-ndk/src/main/java/io/sentry/android/ndk/NdkScopeObserver.java
```diff
@@ -125,4 +129,17 @@ public void removeExtra(final @NotNull String key) {
           .log(SentryLevel.ERROR, e, "Scope sync removeExtra(%s) has an error.", key);
     }
   }
+
+  @Override
+  public void setTrace(@Nullable SpanContext spanContext, @NotNull IScope scope) {
+    if(spanContext == null) {
+      return;
+    }
```
bitsandfoxes, line 137: This part is questionable. I think we should "clear" the trace here too.
My question is: Since I'm "reusing" the `setTrace` this receives `null` when the transaction get cleared from the scope. But I think in a hybrid SDK context, there should always be a trace active.

sentry/src/main/java/io/sentry/JsonSerializer.java
```diff
@@ -200,8 +200,8 @@ public <T> void serialize(@NotNull T entity, @NotNull Writer writer) throws IOEx
     Objects.requireNonNull(writer, "The Writer object is required.");
 
     if (options.getLogger().isEnabled(SentryLevel.DEBUG)) {
-      String serialized = serializeToString(entity, options.isEnablePrettySerializationOutput());
-      options.getLogger().log(SentryLevel.DEBUG, "Serializing object: %s", serialized);
+//      String serialized = serializeToString(entity, options.isEnablePrettySerializationOutput());
+//      options.getLogger().log(SentryLevel.DEBUG, "Serializing object: %s", serialized);
```
bitsandfoxes, line 204:
```suggestion
      String serialized = serializeToString(entity, options.isEnablePrettySerializationOutput());
      options.getLogger().log(SentryLevel.DEBUG, "Serializing object: %s", serialized);
```
I've had this in to make the logs readable. Please disregard.

sentry-android-ndk/src/main/java/io/sentry/android/ndk/NdkScopeObserver.java
bruno-garcia: Right, with TWP (Trace without performance) we always have a trace in the scope. Regardless of spans or transactions.
That's so we can propagate this to downstream services, as well as connecting events (errors) happning on the same process, that are on the same trace. Now how long we keep the same trace is something to be defined depednant on what's possible. We want something that's less than the whole lifetime of the process (it's OK on a CLI for example, but not on a game). And more than just a few milliseconds (if arbitrary). Ideally it's something that's reasonable. On mobile for example, could be a trace id for each screen. But that would reset if the app goes on to the background for 30 seconds or longer (e.g: Session). This way we'd have some related telemetry tied to each other via that trace id

sentry/src/main/java/io/sentry/Sentry.java
```diff
@@ -1147,6 +1147,18 @@ public interface OptionsConfiguration<T extends SentryOptions> {
...
+   * Continue a trace based on the trace ID and span ID provided
+   *
+   * @param traceId the trace ID
+   * @param spanId the span ID
+   */
+  // return TransactionContext (if performance enabled) or null (if performance disabled)
+  public static void setTrace(
+      final @NotNull String traceId, final @NotNull String spanId) {
+    getCurrentScopes().setTrace(traceId, spanId);
+  }
```
krystofwoldrich, line 1160: This looks good and could be used in RN as is.

Rn will propagate the `traceId` from the JS scope.

https://github.com/getsentry/sentry-javascript/blob/fc48f8397d1ed647a4e2d95dba33f022df03daa3/packages/core/src/scope.ts#L120

sentry-android-ndk/src/main/java/io/sentry/android/ndk/NdkScopeObserver.java
romtsn: yeah we're falling back to `scope.getPropagationContext().getSpanContext()` here: https://github.com/getsentry/sentry-java/blob/6cd406da2503830b4afdcbbda46b8e1abfa6bffd/sentry/src/main/java/io/sentry/cache/PersistingScopeObserver.java#L114-L120

maybe you could do the same?

krystofwoldrich: @bitsandfoxes When do you generate new traceId and spanId in unity? 

RN for example generated new traceId and spanId on every new screen, browser on every reload.

RN never clears the id. When does unity clear the traceId?

sentry-android-ndk/src/main/java/io/sentry/android/ndk/NdkScopeObserver.java
```diff
@@ -125,4 +129,17 @@ public void removeExtra(final @NotNull String key) {
           .log(SentryLevel.ERROR, e, "Scope sync removeExtra(%s) has an error.", key);
     }
   }
+
+  @Override
+  public void setTrace(@Nullable SpanContext spanContext, @NotNull IScope scope) {
```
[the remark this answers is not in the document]
krystofwoldrich: Good point, I assumed that the `sentry-android` doesn't generate a new `traceId` since we are not using performance/tracing on the java layer.

We set the following.

```java
    options.setTracesSampleRate(null);
    options.setTracesSampler(null);
    options.setEnableTracing(false);
```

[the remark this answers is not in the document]
krystofwoldrich: Looking at https://github.com/getsentry/sentry-java/blob/f4162ef9982e0e4bd860c1ecd415fc61fa885b05/sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java#L164 we will need an option to disable generating of the `traceId` in java.

At the moment the trace would be broken because RN would set `traceId` when navigation starts, but `sentry-android` would generate new one once the navigation was processed in JS and thus native created a new `activity`.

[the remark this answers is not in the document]
romtsn: yeah that's the issue, since we _always_ need a trace running, we're still spinning up a new one even if perf/tracing is disabled 😅 I guess we need a different mechanism of letting us know we shouldn't create a new trace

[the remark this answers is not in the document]
krystofwoldrich: Do we also generate a new trace on a new `fragment` the most popular RN library creates `fragments` as `"screens"`.

sentry/src/main/java/io/sentry/Sentry.java
```diff
@@ -1147,6 +1147,18 @@ public interface OptionsConfiguration<T extends SentryOptions> {
...
 
+  /**
+   * Continue a trace based on the trace ID and span ID provided
+   *
+   * @param traceId the trace ID
+   * @param spanId the span ID
+   */
+  // return TransactionContext (if performance enabled) or null (if performance disabled)
+  public static void setTrace(
+      final @NotNull String traceId, final @NotNull String spanId) {
```
krystofwoldrich, line 1158: Based on the `sample_rand` spec. https://github.com/getsentry/sentry-docs/pull/12592

The continuing of a trace should require `sample_rand`.

sentry-android-ndk/src/main/java/io/sentry/android/ndk/NdkScopeObserver.java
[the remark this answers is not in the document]
bitsandfoxes: Disabling generating a new trace ID in https://github.com/getsentry/sentry-java/pull/4188

sentry-android-ndk/src/main/java/io/sentry/android/ndk/NdkScopeObserver.java
bitsandfoxes: In Unity the only fixed points of time I can think of right now is changing active scenes. But a scene could be active for a very long time.

sentry/src/main/java/io/sentry/Sentry.java
bitsandfoxes: I've moved `setTrace` to `InternalSentrySdk` and it additionally accepts a `sample rate` and `sample rand`.

sentry-android-ndk/src/main/java/io/sentry/android/ndk/NdkScopeObserver.java
bitsandfoxes: I think regenerating the trace is the hybrid SDK's responsibility and outside of the the scope of this PR.

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
```diff
@@ -329,4 +330,20 @@ private static Session updateSession(
...
+   * @param spanId the trace origin's span ID
+   * @param sampleRate the sample rate used by the origin of the trace
+   * @param sampleRand the random value used to sample with by the origin of the trace
+   */
+  public static void setTrace(
+      final @NotNull String traceId,
+      final @NotNull String spanId,
+      final @Nullable Double sampleRate,
+      final @Nullable Double sampleRand) {
+    getCurrentScopes().setTrace(traceId, spanId, sampleRate, sampleRand);
```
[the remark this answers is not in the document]
adinauer:
```suggestion
    TracingUtils.setTrace(getCurrentScopes(), PropagationContext.fromExistingTrace(traceId, spanId, sampleRate, sampleRand));
```

sentry/src/main/java/io/sentry/PropagationContext.java
```diff
@@ -51,6 +51,19 @@ public static PropagationContext fromHeaders(
         sentryTraceHeader.isSampled());
   }
 
+  public static @NotNull PropagationContext fromExistingTrace(
+      final @NotNull String traceId,
+      final @NotNull String spanId,
+      final @Nullable Double decisionSampleRate,
+      final @Nullable Double decisionSampleRand) {
```
adinauer, line 58: Can we add an optional `Baggage` param here as well?

adinauer (commented):
Left some thoughts after giving this a quick look

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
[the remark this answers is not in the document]
bitsandfoxes: Gotcha! Nice point.

sentry/src/main/java/io/sentry/PropagationContext.java
bitsandfoxes: I don't see the need or benefit of doing that from the Unity SDK's perspective. @krystofwoldrich would this help RN?

adinauer (approved):
LGTM

markushi (approved):
LGTM!

sentry/src/main/java/io/sentry/PropagationContext.java
bruno-garcia: I believe we pass replayId via baggage to downstream services
Might need that on React Native (which has Replay)

krystofwoldrich: We don't need that because the ReplayID is generated on native and synced up to RN Replay Integration which adds it to DSC. 

https://github.com/getsentry/sentry-react-native/blob/5dff5ee95fb4c3d8f2589b80d4d355fecde17778/packages/core/src/js/replay/mobilereplay.ts#L107

The traceID is the opposite, hybrid (RN) generates it and we need to sync it down to native.

Issue #4053, opened 2025-01-15
Reduce IPC calls performed when checking connectivity


romtsn:
### Description

We could be smarter here perhaps and cache the last known state which we receive in https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/internal/util/AndroidConnectionStatusProvider.java#L69-L90

Most of the calls are in the replay integration now but there are definitely more, this is just a subset of them:

```
Count: 48
Trace:
  Trace: java.lang.Throwable
  at android.os.BinderProxy.transact(BinderProxy.java:509)
  at android.net.IConnectivityManager$Stub$Proxy.getActiveNetworkInfo(IConnectivityManager.java:2140)
  at android.net.ConnectivityManager.getActiveNetworkInfo(ConnectivityManager.java:998)
  at io.sentry.android.core.internal.util.AndroidConnectionStatusProvider.getConnectionStatus(AndroidConnectionStatusProvider.java:124)
  at io.sentry.android.core.internal.util.AndroidConnectionStatusProvider.getConnectionStatus(AndroidConnectionStatusProvider.java:53)
  at io.sentry.android.replay.ReplayIntegration.checkCanRecord(ReplayIntegration.kt:334)
  at io.sentry.android.replay.ReplayIntegration.access$checkCanRecord(ReplayIntegration.kt:54)
  at io.sentry.android.replay.ReplayIntegration$onScreenshotRecorded$2.invoke(ReplayIntegration.kt:249)
  at io.sentry.android.replay.ReplayIntegration$onScreenshotRecorded$2.invoke(ReplayIntegration.kt:247)
  at io.sentry.android.replay.capture.SessionCaptureStrategy.onScreenshotRecorded$lambda$3(SessionCaptureStrategy.kt:81)
  at io.sentry.android.replay.capture.SessionCaptureStrategy.$r8$lambda$I7g5B7f5XkbkRoflY1TgbtY-Js0(Unknown Source:0)
  at io.sentry.android.replay.capture.SessionCaptureStrategy$$ExternalSyntheticLambda2.run(Unknown Source:10)
  at io.sentry.android.replay.util.ExecutorsKt.submitSafely$lambda$2(Executors.kt:61)
  at io.sentry.android.replay.util.ExecutorsKt.$r8$lambda$OI4or47ljFo6-uWPc4iTyEjSAc8(Unknown Source:0)
[...]

PR #4164, opened 2025-02-13
Move to a single NetworkCallback listener to reduce number of IPC calls

markushi:
## :scroll: Description

We don't seem to be able to get rid of calling `getConnectionStatus()` on connectivity change, as the NetworkCallback seems to fire for any network, e.g. reporting a disconnect of WIFI, although Cellular connection is still available.

But I noticed every internal `IConnectionStatusObserver` was tied to it's own `NetworkCallback`, causing us to call `getConnectionStatus()` per change per observer - which quickly blows our number of IPC calls up.

So instead let's have a single `NetworkCallback`, call `getConnectionStatus()` once per network change and then broadcast the state to all observers. This should already reduce the number of IPC calls by 3.

Less IPC calls
Fixes https://github.com/getsentry/sentry-java/issues/4053

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/internal/util/AndroidConnectionStatusProvider.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/internal/util/AndroidConnectionStatusProvider.java
@@ -4,17 +4,19 @@ import android.Manifest;
 import android.annotation.SuppressLint;
 import android.content.Context;
 import android.net.ConnectivityManager;
+import android.net.ConnectivityManager.NetworkCallback;
 import android.net.Network;
 import android.net.NetworkCapabilities;
 import android.os.Build;
-import androidx.annotation.NonNull;
 import io.sentry.IConnectionStatusProvider;
 import io.sentry.ILogger;
+import io.sentry.ISentryLifecycleToken;
 import io.sentry.SentryLevel;
 import io.sentry.android.core.BuildInfoProvider;
 import io.sentry.android.core.ContextUtils;
-import java.util.HashMap;
-import java.util.Map;
+import io.sentry.util.AutoClosableReentrantLock;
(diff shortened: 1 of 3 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/internal/util/AndroidConnectionStatusProvider.java
```diff
@@ -41,7 +44,7 @@ public AndroidConnectionStatusProvider(
     this.context = ContextUtils.getApplicationContext(context);
     this.logger = logger;
     this.buildInfoProvider = buildInfoProvider;
-    this.registeredCallbacks = new HashMap<>();
+    this.connectionStatusObservers = new ArrayList<>();
```
romtsn, line 47: just thinking if it's better to use CopyOnWrite here to avoid potential lock contention?

romtsn (approved):
LGTM, very simple change with a very big impact 🚀

romtsn:
Oh btw, I also thought of maybe persisting the last known connection status in `AndroidConnectionStatusProvider` so we don't have to do IPC (`getActiveNetworkInfo`) when we want to know the current status at a certain moment in time? In that case we'd probably have to registerNetworkCallback in the `AndroidConnectionStatusProvider` ctor so we always have access to the last known state after the SDK is initialized.

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/AndroidConnectionStatusProvider.java
markushi: I think it's fine, I'm using a lock mechanism for safe access and there aren't many writes to this list: In my testing there were 3 listeners attached.

PR #4180, opened 2025-02-18
Change continuous profiling to session sample rate

stefanosiano:
## :scroll: Description
replaced continuousProfilesSampleRate with profileSessionSampleRate (Default 0.0)
sample rate is now evaluated inside AndroidContinuousProfiler and every time a session finishes

#skip-changelog

We are now introducing the profileSessionSampleRate, as per [docs](https://www.notion.so/sentry/Continuous-UI-Profiling-SDK-API-Spec-17e8b10e4b5d80c59a40c6e114470934)
Since we didn't release Android Continuous Profiling, we can safely delete all the code pointing to the removed continuousProfilesSampleRate

Unit tests

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
@@ -16,7 +16,9 @@ import io.sentry.NoOpScopes;
 import io.sentry.PerformanceCollectionData;
 import io.sentry.ProfileChunk;
 import io.sentry.Sentry;
+import io.sentry.SentryDate;
 import io.sentry.SentryLevel;
+import io.sentry.SentryNanotimeDate;
 import io.sentry.SentryOptions;
 import io.sentry.android.core.internal.util.SentryFrameMetricsCollector;
 import io.sentry.protocol.SentryId;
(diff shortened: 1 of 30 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
```diff
@@ -100,7 +103,24 @@ private void init() {
             logger);
   }
 
-  public synchronized void start() {
+  public synchronized void start(final @NotNull TracesSampler tracesSampler) {
```
romtsn, line 106: kind of unrelated to this PR but we now have AuotCloseableReentrantLock that we use everywhere instead of synchronized, perhaps you'd want to switch over to it

sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
```diff
@@ -318,11 +318,11 @@ static void applyMetadata(
           }
         }
 
-        if (options.getContinuousProfilesSampleRate() == 1.0) {
-          final double continuousProfilesSampleRate =
-              readDouble(metadata, logger, CONTINUOUS_PROFILES_SAMPLE_RATE);
-          if (continuousProfilesSampleRate != -1) {
-            options.getExperimental().setContinuousProfilesSampleRate(continuousProfilesSampleRate);
+        if (options.getProfileSessionSampleRate() == 0.0) {
```
romtsn, line 321: hm, shouldn't we use a nullable Double here to have the "unset" state rather?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
stefanosiano: I'll do it in another PR, thanks for raising it!

PR #4188, opened 2025-02-20
feat: Added `enableTraceIdGeneration` option

bitsandfoxes:
## :scroll: Description
This PR adds a `enableTraceIdGeneration` option to the `AndroidOptions`. This option is then used in the `ActiviytLifeCycleIntegration` and the `SentryGestureListener` to control whether the SDK generates a new trace.

The idea is that Hybrid SDK should control trace generation. For that they need a way to disable the automatic trace creation in the Java SDK. This is so we can ensure that a single trace ID manages the entire lifecycle of a game/app, linking errors and events top-down.

## :pencil: Checklist


sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
```diff
@@ -107,6 +107,8 @@ final class ManifestMetadataReader {
 
   static final String IGNORED_ERRORS = "io.sentry.ignored-errors";
 
+  static final String ENABLE_AUTO_TRACE_ID_GENERATION = "io.sentry.enable-auto-trace-id-generation";
```
romtsn, line 110:
```suggestion
  static final String ENABLE_AUTO_TRACE_ID_GENERATION = "io.sentry.traces.enable-auto-id-generation";
```
to better align with the `io.sentry.<feature>` pattern we have here already

romtsn (approved):
LGTM, I only have one concern about hybrid-mixed apps (e.g. UaaL or a RN bundle for some screens within a regular android app): this would disable it for the entire run of a process, so it's either-or and not that you can have mixed traces, but I'm not sure if we should think/support this already now.

bitsandfoxes:
> I only have one concern about hybrid-mixed apps

The main use-case is Hybrid SDKs controlling the generation. And it's `true` by default.

PR #4202, opened 2025-02-24
Add continuous profiling ProfileLifecycle

stefanosiano:
## :scroll: Description
Added ProfileLifecycle
Sentry.startProfileSession() will work only in MANUAL mode
Tracer start will start profiler only in TRACE mode
Tracer and spans now attach profilerId only when sampled

#skip-changelog

We are adding profile lifecycle (`manual` | `trace`), as per [docs](https://www.notion.so/sentry/Continuous-UI-Profiling-SDK-API-Spec-17e8b10e4b5d80c59a40c6e114470934)

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
@@ -16,7 +16,9 @@ import io.sentry.NoOpScopes;
 import io.sentry.PerformanceCollectionData;
 import io.sentry.ProfileChunk;
 import io.sentry.Sentry;
+import io.sentry.SentryDate;
 import io.sentry.SentryLevel;
+import io.sentry.SentryNanotimeDate;
 import io.sentry.SentryOptions;
 import io.sentry.android.core.internal.util.SentryFrameMetricsCollector;
 import io.sentry.protocol.SentryId;
@@ -52,6 +54,7 @@ public class AndroidContinuousProfiler
   private @NotNull SentryId profilerId = SentryId.EMPTY_ID;
   private @NotNull SentryId chunkId = SentryId.EMPTY_ID;
   private final @NotNull AtomicBoolean isClosed = new AtomicBoolean(false);
+  private @NotNull SentryDate startProfileChunkTimestamp = new SentryNanotimeDate();
 
   public AndroidContinuousProfiler(
       final @NotNull BuildInfoProvider buildInfoProvider,
@@ -138,8 +141,10 @@ public class AndroidContinuousProfiler
         stop();
         return;
       }
+      startProfileChunkTimestamp = scopes.getOptions().getDateProvider().now();
+    } else {
+      startProfileChunkTimestamp = new SentryNanotimeDate();
     }
-
     final AndroidProfiler.ProfileStartData startData = profiler.start();
     // check if profiling started
     if (startData == null) {
@@ -213,7 +218,11 @@ public class AndroidContinuousProfiler
       synchronized (payloadBuilders) {
         payloadBuilders.add(
             new ProfileChunk.Builder(
-                profilerId, chunkId, endData.measurementsMap, endData.traceFile));
+                profilerId,
+                chunkId,
+                endData.measurementsMap,
+                endData.traceFile,
+                startProfileChunkTimestamp));
       }
     }
 
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
@@ -36,7 +36,7 @@ import org.jetbrains.annotations.VisibleForTesting;
 @ApiStatus.Internal
 public class AndroidContinuousProfiler
     implements IContinuousProfiler, RateLimiter.IRateLimitObserver {
-  private static final long MAX_CHUNK_DURATION_MILLIS = 10000;
+  private static final long MAX_CHUNK_DURATION_MILLIS = 60000;
 
   private final @NotNull ILogger logger;
   private final @Nullable String profilingTracesDirPath;
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
@@ -20,6 +20,7 @@ import io.sentry.SentryDate;
 import io.sentry.SentryLevel;
 import io.sentry.SentryNanotimeDate;
 import io.sentry.SentryOptions;
+import io.sentry.TracesSampler;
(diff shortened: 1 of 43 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
```diff
@@ -103,7 +104,10 @@ private void init() {
             logger);
   }
 
-  public synchronized void start(final @NotNull TracesSampler tracesSampler) {
+  @Override
+  public synchronized void startProfileSession(
```
markushi, line 108: AFAIK we should avoid using synchronized, and use lock objects in favor instead (e.g. see https://github.com/getsentry/sentry-java/blob/f1b71160573bb0a81119da99a4b995b5a03d47b4/sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesTracker.java#L141)

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
```diff
@@ -112,15 +116,31 @@ public synchronized void start(final @NotNull TracesSampler tracesSampler) {
...
       return;
     }
-    if (isRunning()) {
-      logger.log(SentryLevel.DEBUG, "Profiler is already running.");
-      return;
+    switch (profileLifecycle) {
+      case TRACE:
+        // rootSpanCounter should never be negative, unless the user changed profile lifecycle while
+        // the profiler is running or close() is called. This is just a safety check.
+        if (rootSpanCounter < 0) {
```
markushi, line 123: Why do you need keep track of counting here? Wouldn't it be easier to simply check if there's a running transaction or not? E.g. via `Sentry.getCurrentScopes().getTransaction()`

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
markushi, line 108: Would it make sense to return a boolean, indicating if profiling actually started or not?

sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
```diff
@@ -326,6 +329,19 @@ static void applyMetadata(
           }
         }
 
+        final String profileLifecycle =
```
markushi, line 332:
```suggestion
        final @Nullable String profileLifecycle =
```

sentry/src/main/java/io/sentry/SentryAppStartProfilingOptions.java
```diff
@@ -235,6 +248,18 @@ public static final class Deserializer
               options.isContinuousProfilingEnabled = isContinuousProfilingEnabled;
             }
             break;
+          case JsonKeys.PROFILE_LIFECYCLE:
+            String profileLifecycle = reader.nextStringOrNull();
```
markushi, line 252:
```suggestion
            @Nullable String profileLifecycle = reader.nextStringOrNull();
```

markushi (changes requested):
Left a few comments, mainly for getting some clarification. Otherwise looks pretty good to me! 🔥

sentry/src/main/java/io/sentry/SentryAppStartProfilingOptions.java
stefanosiano: updated in the other PR

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
stefanosiano: the counter increases for any transaction, while `Sentry.getCurrentScopes().getTransaction()` returns only the transaction bound to the scope, right?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
stefanosiano: maybe yes, but at the moment it wouldn't be used anywhere. We can always add it later imho

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
```diff
@@ -58,6 +58,7 @@ public class AndroidContinuousProfiler
   private @NotNull SentryDate startProfileChunkTimestamp = new SentryNanotimeDate();
   private boolean shouldSample = true;
   private boolean isSampled = false;
+  private int rootSpanCounter = 0;
```
markushi, line 61:
```suggestion
  private volatile int rootSpanCounter = 0;
```

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
```diff
@@ -58,6 +58,7 @@ public class AndroidContinuousProfiler
   private @NotNull SentryDate startProfileChunkTimestamp = new SentryNanotimeDate();
   private boolean shouldSample = true;
```
markushi, line 59:
```suggestion
  private volatile boolean shouldSample = true;
```

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
```diff
@@ -58,6 +58,7 @@ public class AndroidContinuousProfiler
...
   private boolean isSampled = false;
```
markushi, line 60: I only see this being used within `startProfileSession()`, can't we make this a local variable then?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
markushi: Ah ok, my assumption was that this is being used only when a transaction is bound to scope. All good then :+1:

markushi (approved):
Looks good to me, left some minor nits around volatile field, it's probably worth checking other fields within `AndroidContinuousProfiler.java` as well.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
stefanosiano: this is only set when `shouldSample` is true: it's evaluated once per session, then the value is simply reused, so it cannot be a local variable

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
stefanosiano: don't think we need it, as we access it only in synchronized blocks

Issue #3610, opened 2024-07-30
Changes in url from interceptor chain not represented inside span/breadcrumb


JohannesPtaszyk:
### Integration

sentry

### Java Version

17

### Version

7.12.1

### Steps to Reproduce

1. Create OkHttp instance
2. Create API with route with parts, which should be replaced e.g. fancy.api/REPLACABLE_VERSION/health (REPLACABLE_VERSION would be replaced by our OkHttp interceptor chain with meta information from a custom annotation
3. Look at sentry after calling the API

Additional information:
This seems to happen, because the sauce of truth for the path is not comming from the Sentry interceptor, but from the event listener. 
It would be nice to have an option to tell the Sentry interceptor to override that value whith the one from the interceptor chain or something like that, because right now we are missing important information in our http breadcrumbs.

### Expected Result

placeholder parts are replaced and fancy.api/v1/health is logged instead of fancy.api/REPLACABLE_VERSION/health

### Actual Result

fancy.api/REPLACABLE_VERSION/health is logged

markushi:
Hey @JohannesPtaszyk, thanks for reaching out. Just to better understand your use-case: How are you replacing the placeholders in your urls. Are you using a custom interceptor for this?

JohannesPtaszyk:
Yes, we are using a custom interceptor, which checks for annotations on the API method of our retrofit interface

PR #4238, opened 2025-03-06
Propagate modifications of OkHttp requests to the affected spans / breadcrumbs

markushi:
## :scroll: Description
Makes span/breadcrumb data mutable, so we can update it whenever the underlying http request changes.

Unfortunately the okhttp `EventListener` always gets the _original_ request in most of it's callbacks, even though interceptors have changed the request already. Only when a response is received (e.g. `responseHeadersStart()`) the `response.request.call` would contain the actual request, which probably isn't always called due to caching / errors, etc..

Thus we need to wire the update logic into our interceptor instead - and assume it's running last in the chain.

Fixes https://github.com/getsentry/sentry-java/issues/3610

## :pencil: Checklist

### DIFF
--- a/sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
+++ b/sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
@@ -82,9 +82,6 @@ public open class SentryOkHttpInterceptor(
             span = parentSpan?.startChild("http.client", "$method $url")
         }
 
-        // interceptors may change the request details, so let's update it here
-        // this only works correctly if SentryOkHttpInterceptor is the last one in the chain
-        okHttpEvent?.setRequest(request)
         val startTimestamp = CurrentDateProvider.getInstance().currentTimeMillis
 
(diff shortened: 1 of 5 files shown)


sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
```diff
@@ -81,6 +81,10 @@ public open class SentryOkHttpInterceptor(
             val parentSpan = if (Platform.isAndroid()) scopes.transaction else scopes.span
             span = parentSpan?.startChild("http.client", "$method $url")
         }
+
+        // interceptors may change the request details, so let's update it here
+        // this only works correctly if SentryOkHttpInterceptor is the last one in the chain
+        okHttpEvent?.setRequest(request)
```
stefanosiano, line 87: probably better to have it in the `finally` block, just before the `finishSpan` call, for consistency.
In the meantime we change the request (we add our headers) and at the end we call the `beforeSpan` on the updated request, so it's better to set the updated request into the event?
Not a strong opinion, though

stefanosiano (approved):

Issue #4273, opened 2025-03-20
View Hierarchy for Compose testTag stopped working


tipapro:
### Integration

sentry-android

### Build System

Gradle

### AGP Version

8.6.0

### Proguard

Enabled

### Version

7.18.1

### Steps to Reproduce

Use View Hierarchy with Compose 1.7

### Expected Result

The view hierarchy will contain test tags as composable names

### Actual Result

The view hierarchy not contain test tags as composable names.

Looks like it needs the same fix with TestTagElement in else branch as #3878

https://github.com/getsentry/sentry-java/blob/b5b9f8be9ea477f9a9cab811b5dfd2734c7debcc/sentry-compose-helper/src/jvmMain/java/io/sentry/compose/viewhierarchy/ComposeViewHierarchyExporter.java#L89

tipapro:
https://github.com/getsentry/sentry-android-gradle-plugin/issues/784

PR #4275, opened 2025-03-21
Fix tags missing for compose view hierarchies

markushi:
## :scroll: Description
Only on newer versions of Jetpack Compose

Fixes https://github.com/getsentry/sentry-java/issues/4273

## :pencil: Checklist


sentry-compose-helper/src/jvmMain/java/io/sentry/compose/gestures/ComposeGestureTargetLocator.java
```diff
@@ -75,6 +75,7 @@ public ComposeGestureTargetLocator(final @NotNull ILogger logger) {
         boolean isClickable = false;
         boolean isScrollable = false;
 
+        // needs to be in-sync with ComposeGestureTargetLocator
```
romtsn, line 78: can we commonize this in a separate class easily?

romtsn: also, I guess you meant "with ComposeViewHierarchyExporter" right?

sentry-compose-helper/src/jvmMain/java/io/sentry/compose/gestures/ComposeGestureTargetLocator.java
```diff
@@ -103,7 +104,8 @@ public ComposeGestureTargetLocator(final @NotNull ILogger logger) {
               isClickable = true;
             } else if ("androidx.compose.foundation.ScrollingLayoutElement".equals(type)) {
               isScrollable = true;
-            } else if ("androidx.compose.ui.platform.TestTagElement".equals(type)) {
+            } else if ("androidx.compose.ui.platform.TestTagElement".equals(type)
+                || "io.sentry.compose.SentryModifier.SentryTagModifierNodeElement".equals(type)) {
```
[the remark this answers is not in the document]
romtsn: ah just noticed you already added it, nvm!

sentry-compose-helper/src/jvmMain/java/io/sentry/compose/viewhierarchy/ComposeViewHierarchyExporter.java
```diff
@@ -99,6 +105,21 @@ private static void setTag(
             }
           }
         }
+      } else if ("androidx.compose.ui.platform.TestTagElement".equals(type)
```
romtsn, line 108: is it possible to test it with robolectric/instrumented tests to avoid regressions in future?

romtsn (approved):
would be nice to have tests for this (and extract that logic into a separate class if possible), but LGTM otherwise!

sentry-compose-helper/src/jvmMain/java/io/sentry/compose/SentryComposeHelper.java
```diff
@@ -15,6 +20,40 @@ public class SentryComposeHelper {
...
+  @Nullable
+  public static String extractTag(final @NotNull Modifier modifier) {
+    final @Nullable String type = modifier.getClass().getCanonicalName();
+    // Newer Jetpack Compose uses TestTagElement as node elements
+    // See
+    // https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/ui/ui/src/commonMain/kotlin/androidx/compose/ui/platform/TestTag.kt;l=34;drc=dcaa116fbfda77e64a319e1668056ce3b032469f
+    if ("androidx.compose.ui.platform.TestTagElement".equals(type)
+        || "io.sentry.compose.SentryModifier.SentryTagModifierNodeElement".equals(type)) {
+      try {
+        final Field tagField = modifier.getClass().getDeclaredField("tag");
```
tipapro, line 32: It might be worth caching the class field for `TestTagElement` and `SentryTagModifierNodeElement` in a static variable. Local benchmarks showed about a 5x - 20x improvement.

```
10 objects: 874,2 ns vs 366,1 ns, 2,4 times faster
100 objects: 3927,4 ns vs 146,1 ns, 26,9 times faster
1000 objects: 16092,8 ns vs 1568,3 ns, 10,3 times faster
10000 objects: 111213,5 ns vs 17257,7 ns, 6,4 times faster
100000 objects: 1107549,3 ns vs 256179,3 ns, 4,3 times faster
```

https://gist.github.com/tipapro/ba54c52ea6110583e6d97e5f3ec6f08e#file-benchmark-java

romtsn: great point, we already do that for `layoutDelegateField`, so should do it here as well

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeHelper.kt
```diff
@@ -0,0 +1,91 @@
...
+                val value = testTagElementField.get(modifier)
+                return value as String?
+            } else if ("io.sentry.compose.SentryModifier.SentryTagModifierNodeElement" == type &&
+                sentryTagElementField != null
+            ) {
+                val value = sentryTagElementField.get(modifier)
+                return value as String?
+            }
+        } catch (e: Throwable) {
+            // ignored
```
romtsn, line 42: maybe worth to add a debug log here so we can troubleshoot easier?

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeHelper.kt
```diff
@@ -0,0 +1,91 @@
...
+                }
+            }
+        }
+        return null
+    }
+
+    fun getLayoutNodeBoundsInWindow(node: LayoutNode): Rect? {
+        if (layoutDelegateField != null) {
+            try {
+                val delegate =
```
romtsn, line 64: hmm, I'm thinking you could directly access coordinates in LayoutNode `node.coordinates.boundsInWindow()` without reflection. At least that's what I'm doing in here: https://github.com/getsentry/sentry-java/blob/801e6773af3b3543214ddd2453264fc5a9c85f00/sentry-android-replay/src/main/java/io/sentry/android/replay/viewhierarchy/ComposeViewHierarchyNode.kt#L85

however I think it returns an [innerCoordinator](https://github.com/androidx/androidx/blob/ad593cc5efcc0703fd177d5a806c0a8898ba914a/compose/ui/ui/src/commonMain/kotlin/androidx/compose/ui/node/LayoutNode.kt#L984-L985) as opposed to outerCoordinator you use - do you know the difference between those? 😅

romtsn: Also, I think you could make use of this vendored `boundsInWindow` method too right? https://github.com/getsentry/sentry-java/blob/801e6773af3b3543214ddd2453264fc5a9c85f00/sentry-android-replay/src/main/java/io/sentry/android/replay/util/Nodes.kt#L161-L206

You also have access to the root node (at least in the view hierarchy exporter), I think it would speed up things by quite a bit. Ideally we'd share that method via a separate module, but for simplicity I'd suggest just copying it over in this package for now to not deal with another module.

sentry-compose/src/androidUnitTest/kotlin/io/sentry/compose/viewhierarchy/ComposeViewHierarchyExporterTest.kt
```diff
@@ -0,0 +1,108 @@
...
+            isPlaced: Boolean,
+            tag: String?,
+            width: Int,
+            height: Int,
+            children: List<LayoutNode> = emptyList()
+        ): LayoutNode {
+            val nodeA = Mockito.mock(
+                LayoutNode::class.java
+            )
+            `when`(nodeA.isPlaced).thenReturn(isPlaced)
```
romtsn, line 71: I guess we could use `whenever` from mockito-kotlin here

sentry-compose/src/androidMain/kotlin/io/sentry/compose/viewhierarchy/ComposeViewHierarchyExporter.kt
```diff
@@ -0,0 +1,114 @@
...
+        private fun setBounds(
+            composeHelper: SentryComposeHelper,
+            node: LayoutNode,
+            parentNode: LayoutNode?,
+            vhNode: ViewHierarchyNode
+        ) {
+            val nodeHeight = node.height
+            val nodeWidth = node.width
+
+            vhNode.height = nodeHeight.toDouble()
```
romtsn, line 94: wondering if we could use `bounds.bottom` and `bounds.right` here? should be the same value (in theory 😅 )

romtsn (approved):
LGTM again, but these two comments I guess would be nice to address: https://github.com/getsentry/sentry-java/pull/4275/files#r2014326540

Great tests btw, and very happy to get rid of that workaround with the java-only module

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeHelper.kt
markushi: Since the ViewHierarchy uses relative coordinates, I switched to `boundsInParent()` which is looking good: https://sentry-sdks.sentry.io/issues/5893003489/events/bd2a88e3d9a9436eb3f93004753d9715?project=5428559

Issue #3614, opened 2024-07-31
Implement new user feedback API


markushi:
### Description

Add new API, as well as implement new Protocol definition

PR #4286, opened 2025-03-25
Add new User Feedback API

stefanosiano:
## :scroll: Description
Added new User Feedback APIs (Sentry.captureFeedback)

Still need to add tests

Implements https://github.com/getsentry/sentry-java/issues/3614

## :pencil: Checklist


sentry/src/main/java/io/sentry/protocol/Feedback.java
```diff
@@ -90,6 +92,25 @@ public void setMessage(final @NotNull String message) {
     this.message = message;
   }
 
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) return true;
+    if (!(o instanceof Feedback)) return false;
+    Feedback feedback = (Feedback) o;
+    return Objects.equals(message, feedback.message)
```
markushi, line 100: We probably should use `io.sentry.util.Objects.equals()` instead

sentry/src/main/java/io/sentry/protocol/Feedback.java
```diff
@@ -0,0 +1,233 @@
...
+import io.sentry.util.CollectionUtils;
+import io.sentry.vendor.gson.stream.JsonToken;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+// Specs can be found at https://develop.sentry.dev/sdk/data-model/envelope-items/#user-feedback
```
markushi, line 19: nice!

markushi (approved):
LGTM!

sentry/src/main/java/io/sentry/Scopes.java
```diff
@@ -230,6 +231,37 @@ private IScope buildLocalScope(
...
+        final IScope localScope = buildLocalScope(getCombinedScopeView(), scopeCallback);
+
+        sentryId = getClient().captureFeedback(feedback, hint, localScope);
+      } catch (Throwable e) {
+        getOptions()
+            .getLogger()
+            .log(SentryLevel.ERROR, "Error while capturing feedback: " + feedback.getMessage(), e);
+      }
+    }
+    updateLastEventId(sentryId);
```
romtsn, line 261: Hm, do we do this on other platforms too? Not sure if feedback event_id is useful as lastEventId, wdyt?

sentry/src/main/java/io/sentry/SentryClient.java
```diff
@@ -160,7 +160,8 @@ private boolean shouldApplyScopeData(final @NotNull CheckIn event, final @NotNul
         scope != null ? scope.withSession((@Nullable Session session) -> {}) : null;
     @Nullable Session session = null;
 
-    if (event != null) {
+    // Feedbacks shouldn't be sampled, and they don't affect sessions
```
romtsn, line 163: Do we need to exclude it from beforeSend/EventProcessors and shall we apply the scope data? In web feedbacks I only see tags being applied, so I imagine everything else is redundant?

sentry/src/main/java/io/sentry/protocol/Feedback.java
```diff
@@ -0,0 +1,233 @@
...
+
+public final class Feedback implements JsonUnknown, JsonSerializable {
+  public static final String TYPE = "feedback";
+
+  private @NotNull String message;
+  private @Nullable String contactEmail;
+  private @Nullable String name;
+  private @Nullable SentryId associatedEventId;
+  private @Nullable SentryId replayId;
+  private @Nullable String url;
```
romtsn, line 29: I guess we have to set these two from the scope (url can be scope.screen), but I guess you'll take care of this later in SentryClient

sentry/src/main/java/io/sentry/Scopes.java
stefanosiano: Yeah, we don't
Technically, the last event is the feedback, but i think we introduced this API mainly to attach the last event id to the feedbacks, so yeah, I can remove it

sentry/src/main/java/io/sentry/protocol/Feedback.java
stefanosiano: I forgot to add it 😄 
I added it directly in `ISentryClient`

sentry/src/main/java/io/sentry/SentryClient.java
romtsn: After reading the internal thread, I'm thinking we should rather introduce a new method in SentryClient (`captureFeedback`) that only does the necessary things for the feedback envelope item type, wdyt? Just to make it easier to distinguish

sentry/src/main/java/io/sentry/ISentryClient.java
```diff
@@ -78,6 +79,22 @@ public interface ISentryClient {
...
+   * @param hint An optional hint to be applied to the event.
+   * @param scope An optional scope to be applied to the event.
+   * @return The Id (SentryId object) of the event
+   */
+  default @NotNull SentryId captureFeedback(
+      @NotNull Feedback feedback, @Nullable Hint hint, @Nullable IScope scope) {
+    SentryEvent event = new SentryEvent();
+    event.getContexts().setFeedback(feedback);
+
+    return captureEvent(event, scope, hint);
```
romtsn, line 105: I've also read in the iOS [docs](https://docs.sentry.io/platforms/apple/guides/ios/user-feedback/#session-replay) that when feedback is capture, we also capture an ongoing replay. So I think you'd have to call `options.getReplayController().captureReplay(false);` in the new `captureFeedback` method.

sentry/src/main/java/io/sentry/SentryClient.java
stefanosiano: agree, but i'll wait a little until [docs are updated](https://github.com/getsentry/sentry/issues/88232#issue-2956830633)

sentry/src/main/java/io/sentry/ISentryClient.java
stefanosiano: that one is talking about the UI widget, to capture the replay the makes the user open send the feedback.
I don't think we should add it to the API itself, but only when using the widget (which will come later)

romtsn: hm, I'd expect it to work even if I had my own widget I guess... but that's another point to clarify and align on with other sdks maybe?

stefanosiano: [Added it](https://github.com/getsentry/sentry/issues/88232#issuecomment-2765477810) to the questions 👍

stefanosiano: added it 👍

sentry/src/main/java/io/sentry/protocol/Feedback.java
```diff
@@ -0,0 +1,233 @@
...
+    this.url = url;
+  }
+
+  public @NotNull String getMessage() {
+    return message;
+  }
+
+  public void setMessage(final @NotNull String message) {
+    this.message = message;
+  }
```
buenaflor, line 93: according to develop there's a limit of `4096` characters for the message, is this something we wanna let relay handle?

buenaflor: imo relay should just handle it so it's fine

iirc we're generally now relying on relay to truncate things vs doing it client side

buenaflor: maybe worth checking

stefanosiano: nice catch!
Just checked and confirm the SDK should truncate it, as it's dropped otherwise

sentry/src/main/java/io/sentry/SentryClient.java
```diff
@@ -995,6 +990,126 @@ public void captureSession(final @NotNull Session session, final @Nullable Hint
...
+    if (event == null) {
+      return SentryId.EMPTY_ID;
+    }
+
+    SentryId sentryId = SentryId.EMPTY_ID;
+    if (event.getEventId() != null) {
+      sentryId = event.getEventId();
+    }
+
+    final boolean isBackfillable = HintUtils.hasType(hint, Backfillable.class);
```
romtsn, line 1051: i think we don't need this check because the FeedbackEvent will never be `Backfillable` (it's only for ANRv2 events basically, which go through `SentryClient.captureEvent`)

romtsn (approved):
Looks great, thanks for all the work you've done wrt documenting and aligning this between SDKs!

PR #4287, opened 2025-03-25
added `platform` to SentryEnvelopeItemHeader

stefanosiano:
## :scroll: Description
Added `platform` to SentryEnvelopeItemHeader
Added platform "android" in ProfileChunk envelope items

Implements client side of https://github.com/getsentry/team-ingest/issues/679
We want to rate limit UI and backend profile chunks differently.
In order to avoid having to maintain a list of sdk names, we are required to send the platform in the item header.

Unit tests

## :pencil: Checklist

### DIFF
--- a/sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
+++ b/sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
@@ -302,7 +302,9 @@ public final class SentryEnvelopeItem {
             SentryItemType.ProfileChunk,
             () -> cachedItem.getBytes().length,
             "application-json",
-            traceFile.getName());
+            traceFile.getName(),
+            null,
+            "android");
 
     // avoid method refs on Android due to some issues with older AGP setups
     // noinspection Convert2MethodRef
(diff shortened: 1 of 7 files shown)


sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
```diff
@@ -302,7 +302,9 @@ private static void ensureAttachmentSizeLimit(
             SentryItemType.ProfileChunk,
             () -> cachedItem.getBytes().length,
             "application-json",
-            traceFile.getName());
+            traceFile.getName(),
+            null,
+            "android");
```
markushi, line 307: I'm not sure if/how hybrid SDKs are using this, but should this always be "android"?

markushi (approved):
LGTM, left a minor comment!

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
viglia: The platform in the chunk item-type header should always match the platform in the chunk item-type payload. 

So maybe instead of hard coding `android`, using `profileChunk.getPlatform()` could be better ?

stefanosiano: @markushi i did it this way so the other item headers are not changed. Also, the same code creates and sends the envelopes of the jvm world, so it cannot be set always to "android"
Regarding hybrid, we will see, but they shouldn't use our `captureProfileChunk` method, so it should be fine
@viglia 👍

PR #4293, opened 2025-03-27
Continuous Profiling - Add delayed stop

stefanosiano:
## :scroll: Description
Replaced synchronized blocks with AutoClosableReentrantLock in AndroidContinuousProfiler - unrelated to delayed stop, but small enough to be added in this PR
Added "delayed" stop of profiler, which stops the profiler after the current chunk is finished
Added default span data (profiler id, thread name and thread id) to transaction root span - unrelated to delayed stop, but this was causing the `ui.load` span to not have the thread id and thread name attached, breaking the link to the continuous profile. A fix is already deployed on the frontend, which fallbacks to the main thread in such cases, but we should still fix it anyway in the SDK

The goal of continuous profiling is to send profile chunks 1 minute long, as it's more efficient to store few large chunks than multiple small chunks. To avoid the case when a profiler is started and stopped multiple times in the minute time range, causing multiple small chunks to be captured, we now wait until the chunk is finished to stop the profiler.
Note: closing the SDK through `Sentry.close()` doesn't wait anything

Unit tests

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
@@ -11,6 +11,7 @@ import io.sentry.IContinuousProfiler;
 import io.sentry.ILogger;
 import io.sentry.IScopes;
 import io.sentry.ISentryExecutorService;
+import io.sentry.ISentryLifecycleToken;
 import io.sentry.NoOpScopes;
 import io.sentry.PerformanceCollectionData;
 import io.sentry.ProfileChunk;
@@ -24,6 +25,7 @@ import io.sentry.TracesSampler;
 import io.sentry.android.core.internal.util.SentryFrameMetricsCollector;
 import io.sentry.protocol.SentryId;
 import io.sentry.transport.RateLimiter;
+import io.sentry.util.AutoClosableReentrantLock;
 import io.sentry.util.SentryRandom;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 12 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
```diff
@@ -213,103 +223,112 @@ private synchronized void start() {
...
+        // When the profiler is stopped manually, we have to reset its id
+        profilerId = SentryId.EMPTY_ID;
+        logger.log(SentryLevel.DEBUG, "Profile chunk finished.");
+      }
     }
   }
 
-  public synchronized void reevaluateSampling() {
+  public void reevaluateSampling() {
     shouldSample = true;
```
markushi, line 322: any thread could cause a write here, right? If so, we should mark the field as `volatile`.

markushi (approved):
LGTM!

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
stefanosiano: yeah, but it's a value that is evaluated after a while, and not even a problem if it's skipped the first time.
This way we can skip the overhead of volatile
wdyt?

markushi: I don't think there's a huge overhead of marking the field as volatile, as it's value not read in a tight loop, thus I would prefer correctness over overhead.

Issue #3647, opened 2024-08-20
ANR in Android from Screenshot


JRomeroCar:
### Integration

sentry-android

### Build System

Gradle

### AGP Version

8.1.1

### Proguard

Enabled

### Version

4.5.1

### Steps to Reproduce

Direct operation

### Expected Result

No ANR

### Actual Result

        

![Image](https://github.com/user-attachments/assets/5e65e488-c784-4e42-8866-d5acb6768ba2)

```
  main (native):tid=1 systid=18806 
#00 pc 0x89730 libc.so (syscall + 32) (BuildId: 68a68ca7f3b5f0855cb4842d8ccd258e)
#01 pc 0x232670 libart.so (art::ConditionVariable::WaitHoldingLocks + 140) (BuildId: 2452917c4ff69cbb6e75e5512260946b)
#02 pc 0x6be854 libart.so (art::JNI<false>::SetByteArrayRegion + 1692) (BuildId: 2452917c4ff69cbb6e75e5512260946b)
#03 pc 0x2167e8 libhwui.so (SkJavaOutputStream::write + 116) (BuildId: a747ddbc7c084af82b716a41a91c3357)
#04 pc 0x3ffaa4 libhwui.so (sk_write_fn + 52) (BuildId: a747ddbc7c084af82b716a41a91c3357)
#05 pc 0x31d0c libpng.so (png_compress_IDAT + 560) (BuildId: 1f5d2db6586c5a1002e07d4da15b2d2c)
#06 pc 0x35ac0 libpng.so (png_write_find_filter + 1472) (BuildId: 1f5d2db6586c5a1002e07d4da15b2d2c)
#07 pc 0x2e8c4 libpng.so (png_write_row + 828) (BuildId: 1f5d2db6586c5a1002e07d4da15b2d2c)
#08 pc 0x2e568 libpng.so (png_write_rows + 48) (BuildId: 1f5d2db6586c5a1002e07d4da15b2d2c)
#09 pc 0x400780 libhwui.so (SkPngEncoder::onEncodeRows + 224) (BuildId: a747ddbc7c084af82b716a41a91c3357)
#10 pc 0x3fea8c libhwui.so (SkEncoder::encodeRows + 72) (BuildId: a747ddbc7c084af82b716a41a91c3357)
#11 pc 0x400a1c libhwui.so (SkPngEncoder::Encode + 524) (BuildId: a747ddbc7c084af82b716a41a91c3357)
[...]

PR #4295, opened 2025-03-28
Compress Screenshots on a background thread

markushi:
## :scroll: Description
Fixes https://github.com/getsentry/sentry-java/issues/3647

## :pencil: Checklist


sentry-android-core/src/main/java/io/sentry/android/core/internal/util/ScreenshotUtils.java
```diff
@@ -132,7 +157,19 @@ public class ScreenshotUtils {
...
 
+  public static @Nullable byte[] compressBitmapToPng(
+      final @Nullable Bitmap bitmap, final @NotNull ILogger logger) {
+    if (bitmap == null || bitmap.isRecycled()) {
+      return null;
+    }
+    try (final ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream()) {
       // 0 meaning compress for small size, 100 meaning compress for max quality.
       // Some formats, like PNG which is lossless, will ignore the quality setting.
       bitmap.compress(Bitmap.CompressFormat.PNG, 0, byteArrayOutputStream);
```
markushi, line 175: It's worth mentioning that we never call `bitmap.recycle()` - and we also didn't do before. But I guess it makes sense to do so!

romtsn: yeah, since we also create a new bitmap for every new screenshot we should probably recycle 😅

romtsn (approved):
LGTM!

PR #4311, opened 2025-04-01
Continuous Profiling - stop when app goes in background

stefanosiano:
## :scroll: Description
app going in the background now stops the continuous profiler

Going to the background (after the session timer) should stop the profiler, as decided in the [doc](https://www.notion.so/sentry/Continuous-UI-Profiling-SDK-API-Spec-17e8b10e4b5d80c59a40c6e114470934?pvs=4). Note that it shouldn't restart when it goes to the foreground again

Unit tests

## :pencil: Checklist


sentry-android-core/src/main/java/io/sentry/android/core/LifecycleWatcher.java
```diff
@@ -122,6 +122,7 @@ public void run() {
                   scopes.endSession();
                 }
                 scopes.getOptions().getReplayController().stop();
+                scopes.getOptions().getContinuousProfiler().stopAllProfiles();
```
markushi, line 125: This means the profiler will continue to run for 30s in the background before stopping, is this what we want?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
```diff
@@ -322,6 +322,15 @@ public void reevaluateSampling() {
     shouldSample = true;
   }
 
+  @Override
+  public void stopAllProfiles() {
```
markushi, line 326: can't we use `.stop()` instead?

markushi (changes requested):
Looking good, I left two comments I'd like to clarified before merging.

sentry-android-core/src/main/java/io/sentry/android/core/LifecycleWatcher.java
stefanosiano: yes, and actually stops after 30s and the last chunk is finished

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
stefanosiano: I added another method to explicitly have a difference with the current `stop(ProfileLifecycle)` method, as it stops everything regardless of the profileLifecycle used before

sentry-android-core/src/main/java/io/sentry/android/core/LifecycleWatcher.java
markushi: Alright, to be honest I still wouldn't do this. Once the app is in background, any executed code can easily produce a background ANR.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
markushi: `.stop()` has no `ProfileLifecycle` argument:

`private synchronized void stop(final boolean restartProfiler) {`

So I would either call `.stop()`
```
@Override
 public void stopAllProfiles() {
    stop(false);
 }
 ```
 
 Or actually use `.close()` instead of `.stopAllProfiles()`.
 
 My point is that it's a bit confusing, because we would have 3 ways of stopping: `stopProfiler(final @NotNull ProfileLifecycle profileLifecycle)`, `close()` and `stopAllProfiles()`

markushi (approved):
LGTM, I left a few comments, but nothing major which should stop us from shipping this.

sentry-android-core/src/main/java/io/sentry/android/core/LifecycleWatcher.java
stefanosiano: true, but if we don't, the profiler could continue to run indefinitely, right?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
stefanosiano: Yeah, but the thing is:
* `stopProfiler(final @NotNull ProfileLifecycle profileLifecycle)` is used to stop the profiler in specific points of the SDK, based on the lifecycle
* `close()` is meant to be used when the SDK closes. After calling `close()`, the profiler will not start again, nor send profile chunks anymore
* `stopAllProfiles()` is the least confusing option i found. If we allow to call `stop(final boolean restartProfiler)`, it would be confusing on why we have to call `stopProfiler(final @NotNull ProfileLifecycle profileLifecycle)` in the SDK, instead. 

I thought about it more, and I could add a boolean parameter `isTerminating` to `close()`.
So i can remove the `stopAllProfiles()` method, as it's very similar to `close()` - with the difference it just doesn't stop immediately and doesn't set the internal flag.
Wdyt @markushi?

markushi: Sounds good! Thanks for explaining the different use-cases!

PR #4316, opened 2025-04-04
Fix `SentryUserFilter` in `spring` and `spring-jakarta` samples

lcian:
## :scroll: Description

Previously this was 
```
  @Bean
  SentryUserFilter sentryUserFilter(
      final IScopes scopes, final List<SentryUserProvider> sentryUserProviders) {
    return new SentryUserFilter(scopes, sentryUserProviders);
  }
```
This didn't register the bean because "Could not autowire. No beans of 'IScopes' type found. ".

This seems to work, not sure if it's the correct approach.

#skip-changelog

### DIFF
--- a/sentry-samples/sentry-samples-spring-jakarta/src/main/java/io/sentry/samples/spring/jakarta/AppConfig.java
+++ b/sentry-samples/sentry-samples-spring-jakarta/src/main/java/io/sentry/samples/spring/jakarta/AppConfig.java
@@ -1,6 +1,6 @@
 package io.sentry.samples.spring.jakarta;
 
-import io.sentry.IScopes;
+import io.sentry.ScopesAdapter;
 import io.sentry.spring.jakarta.SentryUserFilter;
 import io.sentry.spring.jakarta.SentryUserProvider;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


adinauer (commented):
While this should work, it sounds like a bug if the bean can't be found. We should investigate that.

lcian:
> While this should work, it sounds like a bug if the bean can't be found. We should investigate that.

I've changed this to use `applicationContext.getBean(IScopes.class)` to retrieve the bean.
As far as I understand this is the correct way to do it because we register the bean dynamically in `SentryHubRegistrar`:
```java
  private void registerSentryHubBean(final @NotNull BeanDefinitionRegistry registry) {
    final BeanDefinitionBuilder builder =
        BeanDefinitionBuilder.genericBeanDefinition(ScopesAdapter.class);
    builder.setInitMethodName("getInstance");

    registry.registerBeanDefinition("sentryHub", builder.getBeanDefinition());
  }
```
So using something like `@Autowired private IScopes scopes;` would not work.

lcian:
This actually works fine, my IDE produces an error because it doesn't know that the bean will be there at runtime, obviously.

Issue #620, opened 2018-10-04
Support for multiple DSNs in Sentry.init()


StreethawkNick:
The documentation for the Android configuration doesn't seem to outline how to use multiple DSNs.

        // Use the Sentry DSN (client key) from the Project Settings page on Sentry
        String sentryDsn = "https://publicKey:secretKey@host:port/1?options";
        Sentry.init(sentryDsn, new AndroidSentryClientFactory(ctx));

Diving deeper into the code it shows that storedClient is a static volatile which makes me believe it can't support it. 

     public static void setStoredClient(SentryClient client) {
        if (storedClient != null) {
            logger.warn("Overwriting statically stored SentryClient instance {} with {}.", storedClient, client);
        }

        storedClient = client;
    }

The reasoning behind wanting multiple DNS is because I'm developing a third party library, so I need Sentry to send events to me, while in the App they also have Sentry to send events to their own. 

Is this possible?

bretthoerner:
You aren't required to use the static `Sentry` API (which stores and uses the global reference), that should probably be reserved for application owners.

What you can do is create your own `SentryClient` with `SentryClientFactory.sentryClient(dsn)` and store the reference somewhere yourself, then use `sendEvent`, `sendMessage`, or `sendException` yourself.

As an aside, I would make it very clear that you are doing this, and offer a way to disable it. You could send all kinds of user/application state to Sentry from your library and users of the library should know that information could be leaking to you/us.

StreethawkNick:
Wonderful, thanks a million, I will use that approach. 

> As an aside, I would make it very clear that you are doing this, and offer a way to disable it. You could send all kinds of user/application state to Sentry from your library and users of the library should know that information could be leaking to you/us.

Thanks again for the advice, we will offer a way to disable it. In addition, our exception handling ignores App level exceptions since it detects which package name the exception came from.

rrauenza:
I posted an answer of how to do this in the Python SDK [here](https://stackoverflow.com/questions/58718196/how-to-have-multiple-global-dsn-in-sentry-sdk-in-python)

Basically, I create my own transport that wraps around multiple transports.  In my case, I wanted to have the global handler send to both DSNs.

carloswabreu1:
Hey there! `SentryClientFactory` seems to no longer exist in recent SDK versions, are multiple DSNs still supported?

marandaneto:
[...]

PR #4323, opened 2025-04-08
Add multiplexed transport

milanboers:
## :scroll: Description
Adds a `MultiplexedTransportFactory` (as experimental API) that sends all envelopes to multiple Sentry DSNs (backends or projects).

Usage:
```java
Sentry.init(options -> {
  options.setDsn(dsn1);
  options.setTransportFactory(new MultiplexedTransportFactory(
    new AsyncHttpTransportFactory(),
    List.of(dsn1, dsn2)
  ));
});
```

This is more or less a port of makeMultiplexedTransport [in the Javascript SDK](https://github.com/getsentry/sentry-javascript/blob/30591d4f536a65418577bef8255f0ea959b497c5/packages/core/src/transports/multiplexed.ts#L90) (relevant https://github.com/getsentry/sentry-javascript/pull/7926).

(The Javascript SDK also allows routing events based on the event content. This seems more useful for the microfrontends use case, but it can be added)

Opened a similar PR for Python: https://github.com/getsentry/sentry-python/pull/4256

It's a common ask from customers, and there seem to be several issues about it on github/online. Would you consider adding it to the Java SDK, like it is in the Javascript SDK?
For Java in particular, it's harder to make your own Transport to do this (because the authentication is added in `RequestDetailsResolver`).

https://github.com/getsentry/sentry-python/discussions/1484
https://github.com/getsentry/sentry-java/issues/620
https://forum.sentry.io/t/java-sdk-multiple-dsns/12715
https://stackoverflow.com/questions/58718196/how-to-have-multiple-global-dsn-in-sentry-sdk-in-python

Unit tests + tried with a local project

## :pencil: Checklist

### DIFF
--- a/sentry/api/sentry.api
+++ b/sentry/api/sentry.api
@@ -1422,6 +1422,11 @@ public final class io/sentry/MonitorScheduleUnit : java/lang/Enum {
 	public static fun values ()[Lio/sentry/MonitorScheduleUnit;
 }
 
+public final class io/sentry/MultiplexedTransportFactory : io/sentry/ITransportFactory {
+	public fun <init> (Lio/sentry/ITransportFactory;Ljava/util/List;)V
+	public fun create (Lio/sentry/SentryOptions;Lio/sentry/RequestDetails;)Lio/sentry/transport/ITransport;
+}
+
 public final class io/sentry/NoOpCompositePerformanceCollector : io/sentry/CompositePerformanceCollector {
 	public fun close ()V
 	public static fun getInstance ()Lio/sentry/NoOpCompositePerformanceCollector;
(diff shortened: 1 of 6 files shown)


lcian:
Hi @milanboers, thanks for the PR!
As my colleague mentioned in https://github.com/getsentry/sentry-python/pull/4256#issuecomment-2789640224, we want to make it possible for users to achieve this usecase through custom transports instead of incorporating the feature directly in the SDK.

As you point out, in Java this is harder due to `RequestDetailsResolver`. 
I think we could merge your changes to it that facilitate this use.
Then we could add your `MultiplexedTransportFactory` to the docs so that other people trying to achieve this can use your approach.

Feel free to create a new PR or update this one by keeping just the changes to `RequestDetailsResolver`.

milanboers:
@lcian thanks! I'll open a new one with the `RequestDetailsResolver` changes :)

Issue #2904, opened 2023-08-24
Crash for composable preview due to lack of DSN declaration in manifest 


hangduykhiem:
### Integration

sentry-android

### Build System

Gradle

### AGP Version

7.4.0

### Proguard

Disabled

### Version

6.28.0

### Steps to Reproduce

- Install Sentry with Gradle
- Add DNS into Application Manifest
- Write a simple compose preview, for example

```kotlin
@Composable
fun Greeting(name: String, modifier: Modifier = Modifier) {
    Text(
        text = "Hello $name!",
        modifier = modifier
    )
}

@Preview(showBackground = true)
@Composable
fun GreetingPreview() {
    MyApplicationTheme {
        Greeting("Android")
    }
}
```

- Run this compose preview **on the device**, and **not** in the preview panel on Android Studio.

### Expected Result

The preview runs normally and we can test our composables on a real device.

### Actual Result

The PreviewActivity crashes with the following log

```
08-24 11:17:29.226  2992  2992 E AndroidRuntime: java.lang.RuntimeException: Unable to get provider io.sentry.android.core.SentryInitProvider: java.lang.IllegalArgumentException: DSN is required. Use empty string to disable SDK.
08-24 11:17:29.226  2992  2992 E AndroidRuntime: 	at android.app.ActivityThread.installProvider(ActivityThread.java:8328)
08-24 11:17:29.226  2992  2992 E AndroidRuntime: 	at android.app.ActivityThread.installContentProviders(ActivityThread.java:7828)
08-24 11:17:29.226  2992  2992 E AndroidRuntime: 	at android.app.ActivityThread.handleBindApplication(ActivityThread.java:7577)
08-24 11:17:29.226  2992  2992 E AndroidRuntime: 	at android.app.ActivityThread.-$$Nest$mhandleBindApplication(Unknown Source:0)
08-24 11:17:29.226  2992  2992 E AndroidRuntime: 	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2400)
08-24 11:17:29.226  2992  2992 E AndroidRuntime: 	at android.os.Handler.dispatchMessage(Handler.java:106)
08-24 11:17:29.226  2992  2992 E AndroidRuntime: 	at android.os.Looper.loopOnce(Looper.java:226)
08-24 11:17:29.226  2992  2992 E AndroidRuntime: 	at android.os.Looper.loop(Looper.java:313)
[...]

PR #4324, opened 2025-04-08
Fix do not initialize SDK for Jetpack Compose Preview builds

markushi:
## :scroll: Description
If you launch a Jetpack Compose Preview for a `@Composable` within a library, the library will be packaged as an app (`androidTest` flavor) and the Compose Tooling will add a [preview activity](https://developer.android.com/reference/kotlin/androidx/compose/ui/tooling/PreviewActivity).

If the library has a dependency to Sentry, this will cause the preview-app to crash, as typically no DSN is set within the library and our SDK will try to auto-init.

There's no easy way around this, but we can detect at runtime if the app is a preview app.

Fixes https://github.com/getsentry/sentry-java/issues/2904

Added unit tests.

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ContextUtils.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ContextUtils.java
@@ -7,6 +7,7 @@ import static android.content.pm.PackageInfo.REQUESTED_PERMISSION_GRANTED;
 import android.annotation.SuppressLint;
 import android.app.ActivityManager;
 import android.content.BroadcastReceiver;
+import android.content.ComponentName;
 import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
@@ -27,6 +28,7 @@ import java.io.FileReader;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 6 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/ContextUtils.java
```diff
@@ -285,6 +287,36 @@ public static boolean isForegroundImportance() {
...
+   *
+   * @param context the context
+   * @return true, if the app is actually a library running as an app for Compose Preview Mode
+   */
+  @ApiStatus.Internal
+  public static boolean appIsLibraryForComposePreview(final @NotNull Context context) {
+    // Jetpack Compose Preview (aka "Run Preview on Device")
+    // uses the androidTest flavor for android library modules,
+    // so let's fail-fast by checking this first
+    if (context.getPackageName().endsWith(".test")) {
```
romtsn, line 301: I think this will cover the majority of customers already, great find 💯

romtsn (approved):
:shipit:

sentry-android-core/src/main/java/io/sentry/android/core/SentryInitProvider.java
```diff
@@ -21,7 +21,9 @@ public boolean onCreate() {
       logger.log(SentryLevel.FATAL, "App. Context from ContentProvider is null");
       return false;
     }
-    if (ManifestMetadataReader.isAutoInit(context, logger)) {
+
+    if (ManifestMetadataReader.isAutoInit(context, logger)
+        && !ContextUtils.appIsLibraryForComposePreview(context)) {
```
romtsn, line 26: I'm wondering if we should do this check inside `SentryAndroid.init` actually? For the case when someone manually initializes the SDK in the library module?

markushi: Yeah I thought about this too, my take: If you manually call init within a library module (or a dependency of that library) you very likely have a DSN set inside that lib or within code. Otherwise any `androidTest` builds would fail for that library as well. 
So I'd keep it as is for now, and if someone still has issue, we can move this block elsewhere too.

PR #4325, opened 2025-04-09
fix(breadcrumbs): Improve low memory breadcrumb capturing

romtsn:
## :scroll: Description

- `onLowMemory` is a deprecated API and should not be used in conjunction with `onTrimMemory`
image

- Move the `level` check out of the runnable lambda, so we don't even enqueue a task if the level is not suitable for reporting
- Debounce low memory breadcrumbs by minute

## :pencil: Checklist


markushi (approved):
LGTM!

PR #4326, opened 2025-04-10
Make RequestDetailsResolver public

milanboers:
## :scroll: Description
Follow up to https://github.com/getsentry/sentry-java/pull/4323

This makes `RequestDetailsResolver` public and adds a constructor which just takes the a DSN and client name.

`ITransportFactory.create` currently [receives](https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/ITransportFactory.java#L17) a `RequestDetails`. This means that, when creating a custom transport, you can't change the request details (dsn, auth header, ..) without resorting to ugly rewrites on the url / headers.

Having a public `RequestDetailsResolver` makes it easy to make a new instance with the desired DSN / client name.

Use case (from https://github.com/getsentry/sentry-java/pull/4323) is to make a transport that can send to multiple different DSNs. (Somewhat simplified) example usage for that:
```java
record MultiplexedTransportFactory(List<String> dsns) implements ITransportFactory {
  @Override
  public ITransport create(final SentryOptions options, final RequestDetails requestDetails) {
    final var transports = dsns.stream()
        .map(dsn -> new RequestDetailsResolver(dsn, options.getSentryClientName()).resolve())
        .map(newRequestDetails -> new AsyncHttpTransportFactory().create(options, newRequestDetails))
        .toList();
    return new ITransport() {
      @Override
      public void send(SentryEnvelope envelope, Hint hint) throws IOException {
        for (ITransport transport : transports) {
          transport.send(envelope, hint);
        }
      }

      @Override
      public boolean isHealthy() {
        return transports.stream().allMatch(ITransport::isHealthy);
      }

      @Override
      public void flush(long timeoutMillis) {
        transports.forEach(transport -> transport.flush(timeoutMillis));
      }

      @Override
      public RateLimiter getRateLimiter() {
        return null;
      }

      @Override
      public void close(boolean isRestarting) throws IOException {
        for (ITransport transport : transports) {
          transport.close(isRestarting);
        }
      }

      @Override
        for (ITransport transport : transports) {
          transport.close();
        }
      }
    };
  }
}
```

## :pencil: Checklist

### DIFF
--- a/sentry/src/main/java/io/sentry/RequestDetailsResolver.java
+++ b/sentry/src/main/java/io/sentry/RequestDetailsResolver.java
@@ -4,24 +4,39 @@ import io.sentry.util.Objects;
 import java.net.URI;
 import java.util.HashMap;
 import java.util.Map;
+import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 /** Resolves {@link RequestDetails}. */
-final class RequestDetailsResolver {
+@ApiStatus.Experimental
+public final class RequestDetailsResolver {
   /** HTTP Header for the user agent. */
   private static final String USER_AGENT = "User-Agent";
   /** HTTP Header for the authentication to Sentry. */
   private static final String SENTRY_AUTH = "X-Sentry-Auth";
 
-  private final @NotNull SentryOptions options;
+  private final @NotNull Dsn dsn;
+  private final @Nullable String sentryClientName;
 
+  public RequestDetailsResolver(
+      final @NotNull String dsn, final @Nullable String sentryClientName) {
+    Objects.requireNonNull(dsn, "dsn is required");
+
+    this.dsn = new Dsn(dsn);
+    this.sentryClientName = sentryClientName;
+  }
+
+  @ApiStatus.Internal
   public RequestDetailsResolver(final @NotNull SentryOptions options) {
-    this.options = Objects.requireNonNull(options, "options is required");
+    Objects.requireNonNull(options, "options is required");
+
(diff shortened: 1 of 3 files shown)


sentry/src/main/java/io/sentry/RequestDetailsResolver.java
```diff
@@ -4,24 +4,39 @@
...
   private static final String USER_AGENT = "User-Agent";
   /** HTTP Header for the authentication to Sentry. */
   private static final String SENTRY_AUTH = "X-Sentry-Auth";
 
-  private final @NotNull SentryOptions options;
+  private final @NotNull Dsn dsn;
+  private final @Nullable String sentryClientName;
 
+  public RequestDetailsResolver(
+      final @NotNull String dsn, final @Nullable String sentryClientName) {
```
milanboers, line 23: Could technically do without this constructor, make the `SentryOptions` one public, and pass fake `SentryOptions` with only dsn and client name from the caller side. Disadvantage is it would break if any other options are used in the future. (but disadvantage of this constructor is that adding a parameter would be breaking API change)

lcian:
Thanks @milanboers. 
Yeah in theory you could do this by using a fake SentryOptions but I think it's better to have a new constructor. 
You could also circumvent the fact that the class is not public by defining your custom transport class in the `io.sentry` package.
I'll check with the team if we want to make this a public API or not.

lcian:
The call to `RequestDetailsResolver.resolve()` will now happen just a line below the original call in `SentryClient`, so there is nothing to worry about regarding behavioral changes.

adinauer (approved):
LGTM! Thanks for taking care of this.

PR #4330, opened 2025-04-11
perf(breadcrumbs): Make SystemEventsBreadcrumbsIntegration faster

romtsn:
## :scroll: Description

The PR attempts to make `SystemEventsBreadcrumbsIntegration` faster by:
- Using a fixed-size array for storing default actions we're subscribing for
  - This will avoid growing (previously unbounded) array list
  - But also will avoid allocating a new iterator in the for-loop
- Improve `getStringAfterDot` by using a pre-allocated buffer and iterating from the end of the string until the first dot to create a String after that. This yielded ~2x faster results on Android (10k iterations x 3 times):

Google Chrome 2025-04-10 23 50 17

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/SystemEventsBreadcrumbsIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/SystemEventsBreadcrumbsIntegration.java
@@ -65,13 +65,13 @@ public final class SystemEventsBreadcrumbsIntegration implements Integration, Cl
 
   private SystemEventsBreadcrumbsIntegration(
       final @NotNull Context context, final @NotNull String[] actions) {
-    this.context = context;
+    this.context = ContextUtils.getApplicationContext(context);
     this.actions = actions;
   }
 
   public SystemEventsBreadcrumbsIntegration(
       final @NotNull Context context, final @NotNull List<String> actions) {
-    this.context = context;
+    this.context = ContextUtils.getApplicationContext(context);
     this.actions = new String[actions.size()];
     actions.toArray(this.actions);
   }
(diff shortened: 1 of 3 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/SystemEventsBreadcrumbsIntegration.java
```diff
@@ -206,10 +216,43 @@ public void onReceive(final Context context, final @NotNull Intent intent) {
...
-            .getLogger()
-            .log(SentryLevel.ERROR, t, "Failed to submit system event breadcrumb action.");
+        // ignored
+      }
+    }
+
+    // in theory this should be ThreadLocal, but we won't have more than 1 thread accessing it,
+    // so we save some memory here and CPU cycles. 64 is because all intent actions we subscribe for
+    // are less than 64 chars. We also don't care about encoding as those are always UTF.
+    // TODO: _MULTI_THREADED_EXECUTOR_
```
romtsn, line 226: decided to mark things that depend on the single-threaded executor from now on so we won't shoot ourselves in the foot if we ever gonna switch to a multi-threaded one 😅

sentry-android-core/src/main/java/io/sentry/android/core/SystemEventsBreadcrumbsIntegration.java
```diff
@@ -206,10 +216,43 @@ public void onReceive(final Context context, final @NotNull Intent intent) {
...
+    private final char[] buf = new char[64];
+
+    @TestOnly
```
stefanosiano, line 229: is this `@TestOnly` wanted?
The code is used normally, so i would either use VisibleForTesting or Internal

stefanosiano (approved):
nice!

sentry-android-core/src/main/java/io/sentry/android/core/SystemEventsBreadcrumbsIntegration.java
romtsn: It's package-private, so I added TestOnly so it gets warned if it's used in a different class. Otherwise could've been private but yeah we need it for tests

PR #4338, opened 2025-04-15
fix(breadcrumbs): Unregister SystemEventsBroadcastReceiver when entering background

### LINKED ISSUE #4294: SystemEventsBroadcastReceiver may produce ANR when receiving android.intent.action.SCREEN_OFF
### What React Native libraries do you use?

RN New Architecture

### Are you using sentry.io or on-premise?

sentry.io (SaS)

### @sentry/react-native SDK Version

@sentry/react-native : 5.35.0

### How does your development environment look like?

react-native version : 0.73.7
hermesEnabled: true
newArchEnabled: true

### Sentry.init()

    Sentry.init({
        dsn,
        sampleRate: 1.0,
        tracesSampleRate: 0.01,
        enableTracing: true,
        enableAppHangTracking: false,
        ignoreErrors,
        environment: Constants.useBetaApi || DebugService.isDebugEnable() ? "development" : "production",
        beforeSend: (event, hint) => {
            if (isEmulatorSync()) {
                console.log("SentryUtils", event, hint);
                return null;
            }
            try {
                if (event.exception && event.exception.values) {
                    const error = event.exception.values[0];
                    if (error.type === "CustomAPIError") {
                        return null;
                    }
                }
            } catch (e) {
                console.log("SentryUtils", e);
            }
            return event;
        },
        integrations: [
            new Sentry.ReactNativeTracing({
                routingInstrumentation,
            }),
        ],
        _experiments: {
            // The sampling rate for profiling is relative to TracesSampleRate.
            // In this case, we'll capture profiles for 100% of transactions.
            profilesSampleRate: 0.01,
        },
    });

### Steps to Reproduce

We cannot provide certain steps to reproduce it for sure, but it seems to happen when the APP is in background and the device is in dreaming state for about several minutes, then unlock it. 
We have also noticed that SystemEventsBroadcastReceiver receives and handles the android.intent.action.SCREEN_OFF broadcast after turning on the device, along with several battery change events to be handled in a short time.

### Expected Result

not to create ANRs

### Actual Result

We received such issues on the Google Play Console with such similar title:
Subject: Broadcast of Intent { act=android.intent.action.SCREEN_OFF flg=0x58200010 cmp=******/io.sentry.android.core.SystemEventsBreadcrumbsIntegration$SystemEventsBroadcastReceiver }

romtsn:
### Some background
Android 14 introduced new behavior for broadcast receivers, and our SystemEvents receiver (used to add breadcrumbs like screen on/off, airplane mode change, etc.) is falling under this. What's happening is the process gets cached for some period of time (can be hours or days) and after getting uncached it receives all those pending broadcasts in a quick succession which most likely causes ANRs. This also correlates with the numbers we see in SDKCD where the biggest number of ANRs is coming from Android 14+. 

I did some tests locally and it is indeed happening as described on the screenshot. 

---

- `SystemEventsBreadcrumbsIntegration` is now lifecycle-aware:
  -  When an app enters background we unregister the receiver
  -  When it enters foreground back, we enqueue a task to register the receiver again
  - This way we avoid being registered in the OS for all those broadcasts and do not receive all the pending ones after the app gets uncached 

Closes #4294 
Internal customer case

Manually + automated

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AppLifecycleIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AppLifecycleIntegration.java
@@ -69,9 +69,8 @@ public final class AppLifecycleIntegration implements Integration, Closeable {
         options
             .getLogger()
             .log(
-                SentryLevel.INFO,
-                "androidx.lifecycle is not available, AppLifecycleIntegration won't be installed",
-                e);
+                SentryLevel.WARNING,
+                "androidx.lifecycle is not available, AppLifecycleIntegration won't be installed");
       } catch (IllegalStateException e) {
         options
             .getLogger()
--- a/sentry-android-core/src/main/java/io/sentry/android/core/SystemEventsBreadcrumbsIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/SystemEventsBreadcrumbsIntegration.java
@@ -67,6 +67,7 @@ public final class SystemEventsBreadcrumbsIntegration implements Integration, Cl
   private @Nullable IScopes scopes;
 
   private final @NotNull String[] actions;
+  private volatile boolean isClosed = false;
   private volatile boolean isStopped = false;
   private volatile IntentFilter filter = null;
   private final @NotNull AutoClosableReentrantLock receiverLock = new AutoClosableReentrantLock();
@@ -129,7 +130,7 @@ public final class SystemEventsBreadcrumbsIntegration implements Integration, Cl
     }
 
     try (final @NotNull ISentryLifecycleToken ignored = receiverLock.acquire()) {
-      if (isStopped || receiver != null) {
+      if (isClosed || isStopped || receiver != null) {
         return;
       }
     }
@@ -140,7 +141,7 @@ public final class SystemEventsBreadcrumbsIntegration implements Integration, Cl
           .submit(
               () -> {
                 try (final @NotNull ISentryLifecycleToken ignored = receiverLock.acquire()) {
-                  if (isStopped || receiver != null) {
+                  if (isClosed || isStopped || receiver != null) {
                     return;
                   }
 
@@ -265,6 +266,7 @@ public final class SystemEventsBreadcrumbsIntegration implements Integration, Cl
   @Override
   public void close() throws IOException {
     try (final @NotNull ISentryLifecycleToken ignored = receiverLock.acquire()) {
+      isClosed = true;
       filter = null;
     }
 
--- a/sentry-android-core/src/main/java/io/sentry/android/core/SystemEventsBreadcrumbsIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/SystemEventsBreadcrumbsIntegration.java
@@ -65,13 +65,13 @@ public final class SystemEventsBreadcrumbsIntegration implements Integration, Cl
 
   private SystemEventsBreadcrumbsIntegration(
       final @NotNull Context context, final @NotNull String[] actions) {
-    this.context = context;
+    this.context = ContextUtils.getApplicationContext(context);
     this.actions = actions;
   }
 
   public SystemEventsBreadcrumbsIntegration(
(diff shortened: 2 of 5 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/AppLifecycleIntegration.java
```diff
@@ -69,9 +69,8 @@ public void register(final @NotNull IScopes scopes, final @NotNull SentryOptions
         options
             .getLogger()
             .log(
-                SentryLevel.INFO,
-                "androidx.lifecycle is not available, AppLifecycleIntegration won't be installed",
-                e);
+                SentryLevel.WARNING,
```
romtsn, line 72: Unrelated to this PR, but I just noticed while testing no androidx.lifecycle case that it's noisy and decided to change that

sentry-android-core/src/main/java/io/sentry/android/core/SystemEventsBreadcrumbsIntegration.java
```diff
@@ -92,46 +115,168 @@ public void register(final @NotNull IScopes scopes, final @NotNull SentryOptions
...
-            .getExecutorService()
-            .submit(
-                () -> {
-                  try (final @NotNull ISentryLifecycleToken ignored = startLock.acquire()) {
-                    if (!isClosed) {
-                      startSystemEventsReceiver(scopes, (SentryAndroidOptions) options);
+  private void registerReceiver(
+      final @NotNull IScopes scopes,
+      final @NotNull SentryAndroidOptions options,
+      final boolean newIntegration) {
```
markushi, line 126: nit, to make it a bit more clear what this flag is about
```suggestion
      final boolean reportAsNewIntegration) {
```

sentry-android-core/src/main/java/io/sentry/android/core/SystemEventsBreadcrumbsIntegration.java
```diff
@@ -92,46 +115,168 @@ public void register(final @NotNull IScopes scopes, final @NotNull SentryOptions
...
-      filter.addAction(item);
+  private void unregisterReceiver() {
+    try (final @NotNull ISentryLifecycleToken ignored = receiverLock.acquire()) {
+      isStopped = true;
     }
+
+    if (receiver != null) {
+      context.unregisterReceiver(receiver);
+      receiver = null;
+    }
```
markushi, line 193: I'd also put writing receiver inside the lock block, so it aligns with `registerReceiver()`
```suggestion
    try (final @NotNull ISentryLifecycleToken ignored = receiverLock.acquire()) {
      isStopped = true;
      if (receiver != null) {
        context.unregisterReceiver(receiver);
        receiver = null;
      }
    }
```

markushi (approved):
Awesome, I left two small comments!

sentry-android-core/src/main/java/io/sentry/android/core/SystemEventsBreadcrumbsIntegration.java
romtsn: I guess the idea here was to only lock over the boolean flag and don't block for the heavy `unregisterReceiver` call. Since we check the for the boolean flag in `registerReceiver` (inside the same lock), it should be fine? Or do you see a case where this would still be racy?

markushi: Oh I see, yeah I would still keep the write to receiver within the block. How about the following?

```java
final @Nullable SystemEventsBroadcastReceiver receiverRef = null;

try (final @NotNull ISentryLifecycleToken ignored = receiverLock.acquire()) {
  isStopped = true;
  receiverRef = receiver;
  receiver = null;
}

if (receiverRef != null) {
  context.unregisterReceiver(receiverRef);
}
```

romtsn: that makes sense, thanks!

Issue #909, opened 2020-07-07
Initializing Sentry triggers a StrictMode violation


mshafrir-stripe:
_Platform:_
- [x]  Android -> API 23 & 29 emulator,`compileSdkVersion`/`targetSdkVersion` are set to `30`
- [ ]  Java -> If yes, which Java (and sourceCompatibility/targetCompatibility) version?
- [ ]  Kotlin -> If yes, which Kotlin (and jvmTarget) version?
- [ ]  NDK -> If yes, which NDK/CMake version?
- [ ]  React-Native -> If yes, which version?

_IDE:_
- [x]  Android Studio -> `4.0`
- [ ]  IntelliJ -> If yes, which version?
- [ ]  Other -> If yes, which one?

_Build system:_
- [x]  Gradle -> `6.5.1`
- [ ]  Buck -> If yes, which version?
- [ ]  Bazel -> If yes, which version?
- [ ]  Other -> If yes, which one?

_Android Gradle Plugin:_
- [x]  Yes -> `4.0.0`
- [ ]  No

_Sentry Android Gradle Plugin:_
- [x]  Yes -> `1.7.35`
- [ ]  No

_Proguard/R8:_
- [x]  Enabled
- [ ]  Disabled

_sentry-android installed with:_
- [x] JCenter
- [ ] Bintray
- [ ] Maven Central
- [ ] Manually

The version of sentry-android:
**2.2.0**

---
I have the following issue:

Initializing Sentry triggers a [StrictMode](https://developer.android.com/reference/android/os/StrictMode) violation

**Steps to reproduce:**
- Enable `StrictMode` logging in `Application` subclass
   ```
    StrictMode.setThreadPolicy(
        StrictMode.ThreadPolicy.Builder()
            .detectAll()
            .penaltyLog()
            .build()
    )
    StrictMode.setVmPolicy(
        StrictMode.VmPolicy.Builder()
            .detectAll()
            .penaltyLog()
            .build()
    )
   ```
- Initialize Sentry automatically
- Run `adb logcat -s StrictMode`
- Launch application

**Actual result:**

```
07-07 11:23:32.694 16309 16317 E StrictMode: A resource was acquired at attached stack trace but never released. See java.io.Closeable for information on avoiding resource leaks.
07-07 11:23:32.694 16309 16317 E StrictMode: java.lang.Throwable: Explicit termination method 'end' not called
[...]

PR #4340, opened 2025-04-15
fix(transport): Tag sockets traffic originating from Sentry's HttpConnection

romtsn:
## :scroll: Description

To get rid of UntaggedSocket strict mode violation

Addresses #909

With tests and verified no strict mode violation when running the sample app

## :pencil: Checklist


sentry-android-core/src/main/java/io/sentry/android/core/AndroidSocketTagger.java
```diff
@@ -0,0 +1,27 @@
...
+
+  private AndroidSocketTagger() {}
+
+  public static AndroidSocketTagger getInstance() {
+    return instance;
+  }
+
+  @Override
+  public void tagSockets() {
+    TrafficStats.setThreadStatsTag(1);
```
stefanosiano, line 20: Would it be better to have the current thread id as tag?
I see the value in not showing the violation, but if we want to tackle it, we can show the current thread id or a number hardly used by the user, to ensure we don't mess with any report of the app.
Also, we could use `TrafficStats.getAndSetThreadStatsTag()`, so that we can restore previous value, and if it's -1 clear it later. Not sure how much worth it is, as we are the only one using this thread, i think, but why not

stefanosiano (approved):
Just a small comment on the number used to tag, but no strong opinion on it

sentry-android-core/src/main/java/io/sentry/android/core/AndroidSocketTagger.java
romtsn: not sure if it'd change much because it's a threadlocal and we anyway use only our own single thread for our requests. But I changed it from 1 to a random number, I think it should be fine like that.

PR #4347, opened 2025-04-22
perf(user-interactions): Make user interaction tracing faster and allocate less

romtsn:
## :scroll: Description

- Use `class.getName` instead of `class.getCanonicalName` as it's almost ~2x faster, and there's almost no difference in what it returns except for inner classes (tested 10k iterations over 10 times)

Google Chrome 2025-04-21 10 46 11

- Fix `SentryTag` extraction for compose, because `SentryTagModifierNodeElement` is an inner class and we have to use a dollar-sign to properly get its class name.

- Use `fori` instead of `foreach` loop to avoid allocating an Iterator in hot paths (e.g. when traversing the view hierarchy which can happen 100s of times for complex UIs). Reduced allocations by almost 2x doing the same user flow (launch -> scroll -> button click)

### Before
Android Studio 2025-04-21 23 12 32

### After

Android Studio 2025-04-21 23 15 46

Improve perf of user interaction tracing

existing tests

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@
 - Fix unregister `SystemEventsBroadcastReceiver` when entering background ([#4338](https://github.com/getsentry/sentry-java/pull/4338))
   - This should reduce ANRs seen with this class in the stack trace for Android 14 and above
 
+### Improvements
+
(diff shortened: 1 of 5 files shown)


markushi (approved):
🙏 👏

Issue #3574, opened 2024-07-15
[SR] Dialog windows are stretched


romtsn:
Replay: https://sentry-sdks.sentry.io/replays/816c1cd2294a45ee99cc12a968a4ab7b

When taking a screenshot of a Dialog window it gets stretched (probably because we take a SS at a different size than the screen resolution)

![Google Chrome 2024-07-15 11 56 30](https://github.com/user-attachments/assets/c45078ff-8cf0-42d0-8f72-47a8150fda2a)

PR #4354, opened 2025-04-24
Determine recording size based on active window

markushi:
## :scroll: Description
Instead of relying on the window metrics, we now determine the size based on the top level Window View, once determined the `RecordingConfig` is created and actual recording is started.

I couldn't re-use the `onConfigurationChanged` hook, as it removes all `rootViews`. But in case of Dialogs, the root views are stacked and as soon as the dialog is popped, no more root view would be available.

I tried to apply my changes with minimal impact to the existing structure, happy to restructure it in a better way too, I simply wasn't sure how hybrids consume the existing APIs.

Fixes https://github.com/getsentry/sentry-java/issues/3574

Manual testing, see e.g. https://sentry-sdks.sentry.io/replays/f278df82391b4b30a7b1cb8b99b15578/?project=5428559&query=&referrer=%2Freplays%2F&statsPeriod=10m&yAxis=count%28%29

## :pencil: Checklist

### DIFF
--- a/sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
+++ b/sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
@@ -69,7 +69,8 @@ public class ReplayIntegration(
     ReplayController,
     ComponentCallbacks,
     IConnectionStatusObserver,
-    IRateLimitObserver {
+    IRateLimitObserver,
+    WindowCallback {
 
     private companion object {
         init {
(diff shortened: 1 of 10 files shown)


sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
```diff
@@ -322,17 +320,16 @@ public class ReplayIntegration(
...
-
-        // refresh config based on new device configuration
-        val recorderConfig = recorderConfigProvider?.invoke(true) ?: ScreenshotRecorderConfig.from(context, options.sessionReplay)
-        captureStrategy?.onConfigurationChanged(recorderConfig)
-
-        recorder?.start(recorderConfig)
-        // we have to restart recorder with a new config and pause immediately if the replay is paused
-        if (lifecycle.currentState == PAUSED) {
-            recorder?.pause()
+        captureStrategy?.stop()
```
romtsn, line 323: hm, I'm thinking this is too destructive for this case, or am I missing something? Because `stop` here will reset everything and start a new replay pretty much

PR #4355, opened 2025-04-24
fix(init): Try to get Application from Context to register App Lifecycle Integrations

krystofwoldrich:
## :scroll: Description

React Native Native Modules receive ReactContext which is not instance of Application class.

This fixes missing integrations in RN applications and other cases when supplied context is not directly Application class.

RN Sample app, I would like to add unit/integration tests after confirming that we want to do this change and that it makes sense for all Android apps.

## :pencil: Checklist


stefanosiano:
this needs some more discussion, as there are things that may not make sense for react.
We currently add spans and breadcrumbs for Activities, Fragments and User Interactions. And i don't think they are needed in react. For example, spans may be created by native android, but not passed to react, or they could be passed and get duplicated (like app start or TTID) or there could be some weird behaviour, like TTFD started in native but only stopped in react.
@krystofwoldrich What's the reason behind this change?
Also, i think this would affect flutter as well, we should verify it.

markushi:
Let's have a quick call (also with @buenaflor) before getting this merged, to clarify what kind of integrations are useful / or not. This probably also affects Unity/Unreal.

romtsn:
This is likely a behavioural change, and we also haven't heard complaints about missing integrations so far, closing until further notice.

krystofwoldrich:
Just for context [this commit [a2f777c]](https://github.com/getsentry/sentry-react-native/commit/a2f777c74b2a3e2b71b04177b2f6b57086410375) in Sentry RN SDK ensures Application is passed to the Sentry Android SDK (when available, so no integrations should be missing on the RN side).

PR #4357, opened 2025-04-25
Add masking debug overlay

markushi:
## :scroll: Description
Masks, revealed
For debugging purposes

https://github.com/user-attachments/assets/21a5e85d-7d11-4dcf-8324-f0b9143c7d2a

## :pencil: Checklist


sentry-android-replay/src/main/java/io/sentry/android/replay/util/DebugOverlayDrawable.kt
```diff
@@ -0,0 +1,66 @@
...
+
+        paint.strokeWidth = 4f
+
+        for (mask in masks) {
+            paint.setColor(Color.BLACK)
+            paint.style = Paint.Style.STROKE
+            canvas.drawRect(mask, paint)
+
+            paint.style = Paint.Style.FILL
+            val label = "${mask.left} ${mask.top}"
```
romtsn, line 30: ideally I'd want to have bottom and right too, shall we maybe use a smaller textSize but in include all the bounds? e.g. in this format `0, 0 - 1, 1`

sentry-android-replay/src/main/java/io/sentry/android/replay/util/DebugOverlayDrawable.kt
```diff
@@ -0,0 +1,66 @@
...
+    private var masks: List<Rect> = emptyList()
+
+    override fun draw(canvas: Canvas) {
+        paint.textSize = 32f
+        paint.setColor(Color.BLACK)
+
+        paint.strokeWidth = 4f
+
+        for (mask in masks) {
+            paint.setColor(Color.BLACK)
```
romtsn, line 25: wondering if we should use Color.RED (or some other which is more prominent in case of a black/white background for example)

sentry-android-replay/src/main/java/io/sentry/android/replay/util/DebugOverlayDrawable.kt
```diff
@@ -0,0 +1,66 @@
...
+            paint.setColor(Color.WHITE)
+            canvas.drawRect(
+                mask.left.toFloat() + paint.strokeWidth / 2,
+                mask.top.toFloat() + paint.strokeWidth / 2,
+                mask.left.toFloat() + tmpRect.width() + padding + padding,
+                mask.top.toFloat() + tmpRect.height() + padding + padding,
+                paint
+            )
+            paint.setColor(Color.BLACK)
+            canvas.drawText(
```
romtsn, line 42: could we actually draw this label outside the mask rect? I think it's nicer visually and wouldn't potentially overlap the content inside the mask (which may be useful to see in some cases)

romtsn:
> Do we want to expose this via options?
How can we ensure the flag to turn it on isn't accidentally committed

Yeah so I talked to @krystofwoldrich yesterday, and turns out iOS also has a similar option - we were thinking of introducing a top-level api like `Sentry.toggleMaskingDebugOverlay()` so it can be used with some kind of a debug/floating button and easily controlled at runtime instead of an init option. Wdyt?

krystofwoldrich:
Cocoa has `SentrySDK.replay.showMaskPreview(0.5)`, the number is opacity of the overlay.

Related cocoa PRs:
- https://github.com/getsentry/sentry-cocoa/pull/4737
- https://github.com/getsentry/sentry-cocoa/pull/4761

romtsn:
> Cocoa has `SentrySDK.replay.showMaskPreview(0.5)`, the number is opacity of the overlay.
> 
> Related cocoa PRs:
> 
> * [feat: Session replay masking preview for SwiftUI sentry-cocoa#4737](https://github.com/getsentry/sentry-cocoa/pull/4737)
> * [feat:  Add `showMaskPreview` to `SentrySDK.replay` api to debug replay masking sentry-cocoa#4761](https://github.com/getsentry/sentry-cocoa/pull/4761)

alright, couldn't we draw stroke rectangles instead of solid? then we wouldn't need the opacity parameter. Just trying to align I guess

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/MainActivity.java
```diff
@@ -273,6 +274,11 @@ public void run() {
           CoroutinesUtil.INSTANCE.throwInCoroutine();
         });
 
+    binding.enableReplayDebugMode.setOnClickListener(
+        view -> {
+          ReplayIntegration.enableDebugMasking();
```
romtsn, line 279: Wondering if we should expose it via `Sentry.replay.enable/disableDebugMaskingOverlay()`? We could introduce `IReplayApi` similar to `ILoggerApi` with this single method. And then we already have [ReplayController](https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/ReplayController.java) which is accessible via `SentryOptions` and we could also access it this way from within `ScreenshotRecorder` e.g. `options.replayController.isDebugMaskingEnabled`

sentry-android-replay/src/main/java/io/sentry/android/replay/util/DebugOverlayDrawable.kt
```diff
@@ -0,0 +1,103 @@
...
+                bottomRightLabel,
+                mask.right.toFloat() - tmpRect.width(),
+                mask.bottom.toFloat() + tmpRect.height()
+            )
+        }
+    }
+
+    private fun drawTextWithOutline(
+        canvas: Canvas,
+        bottomRightLabel: String,
```
romtsn, line 65:
```suggestion
        label: String,
```

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/MainActivity.java
markushi: Yes, let's do that!

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Add debug mode for Session Replay masking ([#4357](https://github.com/getsentry/sentry-java/pull/4357))
```
```suggestion
- Add debug mode for Session Replay masking ([#4357](https://github.com/getsentry/sentry-java/pull/4357))
```

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
...
+  Use `ReplayIntegration.enableDebugMasking()` to overlay the screen with the Session Replay masks
```
romtsn, line 8:
```suggestion
  Use `ReplayIntegration.enableDebugMasking()` to overlay the screen with the Session Replay masks. The masks will be invalidated at most once per `frameRate` (default 1 fps).
```

sentry-android-core/api/sentry-android-core.api
```diff
@@ -280,6 +280,7 @@ public final class io/sentry/android/core/SentryAndroid {
 	public static fun init (Landroid/content/Context;Lio/sentry/ILogger;)V
 	public static fun init (Landroid/content/Context;Lio/sentry/ILogger;Lio/sentry/Sentry$OptionsConfiguration;)V
 	public static fun init (Landroid/content/Context;Lio/sentry/Sentry$OptionsConfiguration;)V
+	public static fun replay ()Lio/sentry/IReplayApi;
```
romtsn, line 283: Can we move this under `Sentry`? We don't use `SentryAndroid` anywhere except for init, so I think it'd be a bit awkward to start throwing `SentryAndroid` around as public API.

romtsn (approved):

PR #4361, opened 2025-04-25
fix(replay): Use global visible rect when text layout is not laid out in Compose

romtsn:
## :scroll: Description

When a `Text` component does not specify any font parameters (like fontSize, etc.), the underlying `TextLayout` returns bogus values. So we're checking now if the fontSize != Unspecified (though there may be some other indicators we'd need to use in the future), and if it is we fallback to just masking by global visible bounds.

### Before
Google Chrome 2025-04-24 23 27 04

### After
Google Chrome 2025-04-24 23 25 44

Probably part of #4177 

manually + automated

## :pencil: Checklist

### DIFF
--- a/scripts/toggle-codec-logs.sh
+++ b/scripts/toggle-codec-logs.sh
@@ -45,6 +45,10 @@ case "$ACTION" in
         adb shell setprop log.tag.BufferQueueProducer D
         adb shell setprop log.tag.ReflectedParamUpdater D
         adb shell setprop log.tag.hw-BpHwBinder D
+        adb shell setprop log.tag.ACodec D
+        adb shell setprop log.tag.VideoCapabilities D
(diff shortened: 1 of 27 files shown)


scripts/toggle-codec-logs.sh
```diff
@@ -45,6 +45,10 @@ case "$ACTION" in
         adb shell setprop log.tag.BufferQueueProducer D
         adb shell setprop log.tag.ReflectedParamUpdater D
         adb shell setprop log.tag.hw-BpHwBinder D
+        adb shell setprop log.tag.ACodec D
```
romtsn, line 48: I found this when running on a pixel device, so just added them here, not related to this PR

markushi (approved):

PR #4384, opened 2025-05-09
New User Feedback form

stefanosiano:
## :scroll: Description
Continuation of https://github.com/getsentry/sentry-java/pull/4286
Add the new User Feedback form (an AlertDialog)
Add all relevant options for customizations and callbacks

[Here](https://sentry-sdks.sentry.io/issues/feedback/?feedbackSlug=sentry-android%3A6617547311) is the result.

How the form is shown in the example app with light theme and no customization
Screenshot 2025-05-14 at 18 00 02

How the form is shown in the example app with dark theme and no customization
Screenshot 2025-05-14 at 17 59 44

How the form is shown in the example app with (pretty bad) customization
Screenshot 2025-05-14 at 18 02 16

Implements first part of https://github.com/getsentry/sentry-java/issues/3613

ui test (`UserFeedbackUiTest`)

## :pencil: Checklist

## :crystal_ball: Next steps
Next PR will have the widget (the button that actually show the form)
After that, we'll have screenshots and whatever


sentry-android-core/src/main/AndroidManifest.xml
```diff
@@ -2,7 +2,8 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
   <uses-permission android:name="android.permission.INTERNET"/>
 
-  <application>
+  <application
+    android:supportsRtl="true">
```
markushi, line 6: do we need to set `supportsRtl`? I'm wondering if this could cause manifest merging issues.

sentry-android-core/src/main/java/io/sentry/android/core/SentryUserFeedbackDialog.java
```diff
@@ -0,0 +1,225 @@
...
+
+  private boolean isCancelable = false;
+  private @Nullable SentryId currentReplayId;
+  private @Nullable OnDismissListener delegate;
+
+  public SentryUserFeedbackDialog(final @NotNull Context context) {
+    super(context);
+  }
+
+  public SentryUserFeedbackDialog(
```
markushi, line 33: Can we have a ctor or a Dialog builder which takes `SentryFeedbackOptions` as input? IMHO I would even do that in favor of Manifest options, as it allows you to e.g. use localized labels

sentry-android-core/src/main/res/layout/sentry_dialog_user_feedback.xml
```diff
@@ -0,0 +1,112 @@
...
+        android:lines="6"
+        android:inputType="textMultiLine"
+        android:gravity="top|start"
+        android:hint="What's the bug? What did you expect?"
+        android:background="@drawable/edit_text_border"
+        android:paddingHorizontal="8dp"
+        android:layout_below="@id/sentry_dialog_user_feedback_txt_description" />
+
+    <Button
+        android:id="@+id/sentry_dialog_user_feedback_btn_send"
```
markushi, line 94: Could we have the buttons horizontally, side by side? At least that would be more Android-ish.
See https://developer.android.com/develop/ui/views/components/dialogs
image

markushi (commented):
Looks good to me, I left a few comments around the API, let's sort those out before merging. On top of that: Do you know if other SDKs provide default i18n for the widget? Or is it really up the the user to provide their own?

sentry-android-core/src/main/res/layout/sentry_dialog_user_feedback.xml
```diff
@@ -0,0 +1,112 @@
...
+    tools:ignore="HardcodedText"
+    android:theme="?android:attr/dialogTheme"
+    android:padding="24dp">
+
+
+    <TextView
+        android:id="@+id/sentry_dialog_user_feedback_title"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:text="Report a Bug"
```
markushi, line 16: You could use `"tools:text"` instead for some nice preview, as it all will be overwritten anyway right?

stefanosiano:
> Do you know if other SDKs provide default i18n for the widget? Or is it really up the the user to provide their own?

[nope](https://docs.sentry.io/platforms/javascript/user-feedback/configuration/#text-customization), all SDKs support english only by default.
Actually, i'm not sure if supporting multiple languages would actually bloat the apk size?

markushi:
> > Do you know if other SDKs provide default i18n for the widget? Or is it really up the the user to provide their own?
> 
> [nope](https://docs.sentry.io/platforms/javascript/user-feedback/configuration/#text-customization), [all](https://docs.sentry.io/platforms/apple/user-feedback/configuration/#form-configuration) SDKs support english only by default. Actually, i'm not sure if supporting multiple languages would actually bloat the apk size?

You're right, it would bloat the APK, especially if we support languages the consuming app doesn't. Let's stick to EN then.

sentry-android-core/src/main/res/layout/sentry_dialog_user_feedback.xml
stefanosiano: I tried to make it more similar to the [other SDKs](https://docs.sentry.io/mdx-images/user-feedback-widget-customization-52PBCJUY.png)
While it's an `AlertDialog` under the hood, having the layour of a normal dialog would be strange to me

sentry-android-core/src/main/res/drawable/edit_text_border.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
```
markushi: nit: can we prefix the filename with `sentry_`?

sentry/src/main/java/io/sentry/SentryFeedbackOptions.java
```diff
@@ -0,0 +1,543 @@
...
+  public @NotNull String getSubmitButtonLabel() {
+    return submitButtonLabel;
+  }
+
+  /**
+   * Sets the label of the submit button. Defaults to "Send Bug Report".
+   *
+   * @param submitButtonLabel the label of the submit button
+   */
+  public void setSubmitButtonLabel(final @NotNull String submitButtonLabel) {
```
romtsn, line 246: I'm thinking it'd be a nice DevX to have overloads that accept `@StringRes int resId` like [this](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/widget/TextView.java;l=7613-761).

Also this one (and other Strings) should probably be changed to a `CharSequence` to make it more versatile

sentry-android-core/src/main/res/drawable-hdpi/sentry_logo_dark.webp
romtsn, line 1: any reason all of those are just not a single [vector drawable](https://developer.android.com/studio/write/vector-asset-studio)?

sentry/src/main/java/io/sentry/SentryFeedbackOptions.java
romtsn: ah I see it's in core sentry... still thinking we could expose those just without the annotation and have something like `SentryOptions.ResourceProvider` which on android would return `context.getResources()`

stefanosiano: That's the reason we added the `SentryUserFeedback.Builder`, which has the `OptionsConfiguration` with `void configure(final @NotNull Context context, final @NotNull SentryFeedbackOptions options);`
I agree on the `int` overload, though, and if you think it makes sense, i can add it in a followup PR

stefanosiano: i will change the options to `CharSequence`

sentry-android-core/src/main/res/drawable-hdpi/sentry_logo_dark.webp
stefanosiano: nope, just an oversight. Will do in a followup pr

PR #4385, opened 2025-05-09
Run CodeQL only post merge

runningcode:
## :scroll: Description
This runs CodeQL with build caching disabled post-merge. 

#skip-changelog

CodeQL is slow and requires the compilation tasks to re-run. By disabling build caching we can ensure that CodeQL is run correctly. However since this is one of the longest checks, it doesn’t make sense to run it during PRs.

## :pencil: Checklist


.github/workflows/codeql-analysis.yml
```diff
@@ -3,9 +3,6 @@ name: 'CodeQL'
 on:
   push:
     branches: [main]
-  pull_request:
```
markushi, line 6: To be honest I would still run CodeQL for PRs against main, as it will block the PR from being merged if there are CodeQL issues.

romtsn: I think it's probably fine to run it on main only, because I don't remember a single time where it failed for a valid reason 😅 We'll still get the notification if it's failed on main and see the results here: https://github.com/getsentry/sentry-java/security/code-scanning

The reason we don't want to run it on PRs because without build-cache it'd be the longest one to run. Alternatively we could skip the run altogether, if there were no code changes to prevent the job from failing, but I think it's a good middle-ground for now.

markushi (approved):

romtsn (approved):
LGTM, let's see how it goes and we add a skip-check later, if we see it necessary 👍

PR #4423, opened 2025-05-20
Send UI Profiling app start chunk when it finishes

stefanosiano:
## :scroll: Description
init the scopes at profile finish, too, so that it sends the chunk the first time

When the startOnAppStart option is used, the first chunk is started when the SDK is not enabled, so it's stored in memory.
The SDK reevaluates the scopes when the next chunk is started, and in the meantime it keeps the chunks in memory until the second chunk finishes.
This is not an issue normally, as it affects only the very first chunk when using the startProfilerOnAppStart option.
But we do it in the UI tests of our demo app, which doesn't send any chunk unless another transaction is started one minute after the app start.

## :pencil: Checklist


sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
```diff
@@ -162,6 +162,10 @@ private void start() {
         rateLimiter.addRateLimitObserver(this);
```
romtsn, line 162: can this be potentially added multiple times now?

PR #4428, opened 2025-05-21
More version catalog adoption

runningcode:
## :scroll: Description
Version catalog adoption

#skip-changelog

## :pencil: Checklist


gradle/libs.versions.toml
```diff
@@ -6,6 +6,10 @@ composeVersion = "1.6.11"
 okhttp = "4.9.2"
 springTwo = "2.7.18"
 springThree = "3.4.2"
+targetSdk = "34"
```
runningcode, line 9: There’s no way to have aliases or have `targetSdk` and `compileSdk` values reference each other.

sentry-samples/sentry-samples-spring-jakarta/build.gradle.kts
```diff
@@ -9,7 +8,7 @@ plugins {
     kotlin("jvm")
     alias(libs.plugins.kotlin.spring)
     id("war")
-    id(Config.BuildPlugins.gretty) version Config.BuildPlugins.grettyVersion
+    alias(libs.plugins.gretty)
```
romtsn, line 11: do we need to specify it in the root project with the version then?

romtsn (approved):

sentry-samples/sentry-samples-spring-jakarta/build.gradle.kts
runningcode: that’s a good point, i’ll do this.

it isn’t required but it helps maintain a single classpath for all projects. this build is a bit tricky since different subprojects use different versions of the same plugin (like spring) so the buildscript classpath cannot by synchronized 100% for now.

runningcode: Actually it looks like gretty depends on spring boot and since we don’t have spring boot in the root buildscript classpath, this causes conflicts.

PR #4454, opened 2025-05-27
[SR] Remove configuration from start() method

markushi:
## :scroll: Description

As we anyway have `onConfigurationChanged` now. This removes the need for odd constructs like this one

```Kotlin
captureStrategy?.let { capture ->
    if (capture.currentReplayId == SentryId.EMPTY_ID) {
        capture.start(recorderConfig)
    } else {
        capture.onConfigurationChanged(recorderConfig)
    }
}
```

#skip-changelog

## :pencil: Checklist

### DIFF
--- a/sentry-android-replay/src/main/java/io/sentry/android/replay/WindowRecorder.kt
+++ b/sentry-android-replay/src/main/java/io/sentry/android/replay/WindowRecorder.kt
@@ -96,13 +96,13 @@ internal class WindowRecorder(
         isRecording.getAndSet(true)
     }
 
-    override fun onConfigurationChanged(recorderConfig: ScreenshotRecorderConfig) {
+    override fun onConfigurationChanged(config: ScreenshotRecorderConfig) {
         if (!isRecording.get()) {
             return
         }
 
         recorder = ScreenshotRecorder(
-            recorderConfig,
+            config,
             options,
             mainLooperHandler,
             replayExecutor,
@@ -119,7 +119,7 @@ internal class WindowRecorder(
             options,
             "$TAG.capture",
             100L, // delay the first run by a bit, to allow root view listener to register
-            1000L / recorderConfig.frameRate,
+            1000L / config.frameRate,
             MILLISECONDS
         ) {
             recorder?.capture()
--- a/sentry-android-replay/src/main/java/io/sentry/android/replay/WindowRecorder.kt
+++ b/sentry-android-replay/src/main/java/io/sentry/android/replay/WindowRecorder.kt
@@ -92,8 +92,12 @@ internal class WindowRecorder(
         }
     }
 
-    override fun start(recorderConfig: ScreenshotRecorderConfig) {
-        if (isRecording.getAndSet(true)) {
+    override fun start() {
+        isRecording.getAndSet(true)
+    }
+
+    override fun onConfigurationChanged(recorderConfig: ScreenshotRecorderConfig) {
+        if (!isRecording.get()) {
             return
         }
 
--- a/sentry-android-replay/src/main/java/io/sentry/android/replay/capture/BaseCaptureStrategy.kt
+++ b/sentry-android-replay/src/main/java/io/sentry/android/replay/capture/BaseCaptureStrategy.kt
@@ -61,7 +61,7 @@ internal abstract class BaseCaptureStrategy(
 
     protected val isTerminating = AtomicBoolean(false)
     protected var cache: ReplayCache? = null
-    protected var recorderConfig: ScreenshotRecorderConfig by persistableAtomic { _, _, newValue ->
+    protected var recorderConfig: ScreenshotRecorderConfig? by persistableAtomic { _, _, newValue ->
         if (newValue == null) {
             // recorderConfig is only nullable on init, but never after
             return@persistableAtomic
@@ -85,7 +85,6 @@ internal abstract class BaseCaptureStrategy(
     protected val currentEvents: Deque<RRWebEvent> = ConcurrentLinkedDeque()
 
     override fun start(
-        recorderConfig: ScreenshotRecorderConfig,
         segmentId: Int,
         replayId: SentryId,
         replayType: ReplayType?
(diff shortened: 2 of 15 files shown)


sentry-android-replay/src/main/java/io/sentry/android/replay/WindowRecorder.kt
```diff
@@ -92,8 +92,12 @@ internal class WindowRecorder(
...
     }
 
-    override fun start(recorderConfig: ScreenshotRecorderConfig) {
-        if (isRecording.getAndSet(true)) {
+    override fun start() {
+        isRecording.getAndSet(true)
+    }
+
+    override fun onConfigurationChanged(recorderConfig: ScreenshotRecorderConfig) {
+        if (!isRecording.get()) {
```
romtsn, line 100: I don't think this is going to work for orientation change, because we call `recorder.stop()` which sets this to `false`, so we'll not restart the recording.

sentry-android-replay/src/main/java/io/sentry/android/replay/capture/SessionCaptureStrategy.kt
```diff
@@ -140,17 +141,26 @@ internal class SessionCaptureStrategy(
...
+                        currentSegmentTimestamp,
+                        replayId,
+                        segmentId,
+                        config.recordingHeight,
+                        config.recordingWidth,
+                        config.frameRate,
+                        config.bitRate
+                    )
+                onSegmentCreated(segment)
+            }
```
romtsn: can we add a log maybe saying that recorderConfig isn't available that's why we're not capturing? Would help us investigate issues.

sentry-android-replay/src/main/java/io/sentry/android/replay/capture/BufferCaptureStrategy.kt
```diff
@@ -189,24 +189,33 @@ internal class BufferCaptureStrategy(
...
+                        currentSegmentTimestamp,
+                        replayId,
+                        segmentId,
+                        config.recordingHeight,
+                        config.recordingWidth,
+                        config.frameRate,
+                        config.bitRate
+                    )
+                onSegmentCreated(segment)
+            }
```
romtsn: same here with the log

sentry-android-replay/src/main/java/io/sentry/android/replay/capture/SessionCaptureStrategy.kt
```diff
@@ -98,20 +95,24 @@ internal class SessionCaptureStrategy(
...
-                        currentSegment,
-                        height,
-                        width
-                    )
-                if (segment is ReplaySegment.Created) {
-                    segment.capture(scopes)
-                    currentSegment++
-                    // set next segment timestamp as close to the previous one as possible to avoid gaps
-                    segmentTimestamp = segment.replay.timestamp
+                recorderConfig?.let { config ->
```
romtsn: can we capture values of the recorderConfig on the main thread (basically before replayExecutor.submit) and then pass them to the runnable? 

I'm thinking of a case where recorderConfig changes (e.g. orientation change), and we'd be trying to capture a segment with the new values already, however the screenshots would be still from before orientation change

sentry/src/main/java/io/sentry/SentryReplayOptions.java
```diff
@@ -119,10 +119,13 @@ public enum SentryReplayQuality {
...
   /**
-   * Whether to track orientation changes in session replay. Used in Flutter as it has its own
-   * callbacks to determine the orientation change.
+   * Whether to track the screen configuration (e.g. window size, orientation, etc.) automatically.
+   * A valid configuration is required to capture a session replay. If set to false,
+   * ReplayIntegration.onConfigurationChanged() must be called manually to update the configuration.
+   *
+   * <p>Defaults to true.
    */
-  private boolean trackOrientationChange = true;
```
markushi, line 125: Happy to keep the old option as well, not sure how eager we are to break this 😅

sentry-android-replay/src/main/java/io/sentry/android/replay/capture/BaseCaptureStrategy.kt
```diff
@@ -61,7 +61,7 @@ internal abstract class BaseCaptureStrategy(
 
     protected val isTerminating = AtomicBoolean(false)
     protected var cache: ReplayCache? = null
-    protected var recorderConfig: ScreenshotRecorderConfig by persistableAtomic { _, _, newValue ->
+    protected var recorderConfig: ScreenshotRecorderConfig? by persistableAtomic { _, _, newValue ->
```
romtsn:
```suggestion
    protected var recorderConfig by persistableAtomicNullable<ScreenshotRecorderConfig> { _, _, newValue ->
```
I guess it should be like this

sentry-android-replay/src/test/java/io/sentry/android/replay/ReplayIntegrationWithRecorderTest.kt
```diff
@@ -123,12 +125,22 @@ class ReplayIntegrationWithRecorderTest {
             }
         }
 
-        replay = fixture.getSut(context, recorder, recorderConfig, dateProvider)
+        replay = fixture.getSut(context, recorder, dateProvider)
         replay.register(fixture.scopes, fixture.options)
 
         assertEquals(INITALIZED, recorder.state)
 
+        // have to access 'replayCacheDir' after calling replay.start(), BUT can already be accessed
```
romtsn, line 135: hm, so this comment is not relevant anymore?

romtsn: I think replay dir will still be `replay_0000` at this point because no replayId was generated yet?

sentry/src/main/java/io/sentry/SentryReplayOptions.java
romtsn: I think it's fine to change it since you already made the changes to Flutter (and it was the only user of this flag)

sentry-android-replay/src/test/java/io/sentry/android/replay/ReplayIntegrationWithRecorderTest.kt
markushi: Ah true, let me change it back to the original order then

romtsn (approved):
LGTM!

PR #4445, opened 2025-05-26
fix(docs) Temporary disable gh-pages javadoc generation

markushi:
## :scroll: Description

As https://github.com/getsentry/sentry-java/blob/8295b2f1edf2dd9a99ae837e1289dd843410e329/build.gradle.kts#L267 mangles the classpath.

As far as I can see this is only used for the gh-pages branch docs: https://getsentry.github.io/sentry-java/

Note: This is a temporary solution, we should apply a proper fix in the long run.

## :bulb: Motivation and Context

Working CI

#skip-changelog

## :pencil: Checklist

- [ ] I added tests to verify the changes.
- [ ] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [ ] Review from the native team if needed.
- [ ] No breaking change or entry added to the changelog.
- [ ] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

PR #4457, opened 2025-05-27
Aggregate javadoc using build-logic

runningcode:
## :scroll: Description
This PR has a few parts to it:
1. Adds a new `build-logic` module to organize our build logic.
2. Adds two convention plugins.
   - `sentry.javadoc` publishes the javadoc
   to a consumable configuration.
   - `sentry.javadoc.aggregate` consumes the published javadoc and
     aggregates them in the root build folder.
3. Deletes the `stylesheet.css`. This was causing the javadoc to look
   unusable. Deleting it fixes this.
4. Each subproject publishes the javadoc to its own subfolder. The
   previous behavior would overwrite the javadoc with each new project
   in the same directory which meant some parts of the javadoc were
   unreachable.

The producer/consumer configuration is based on this blog post: https://www.liutikas.net/2024/12/11/Together-In-Isolation.html
It should be project isolation compatible.

Here is what the javadoc looks like right now: https://getsentry.github.io/sentry-java/
This is what the javadoc will look like without the stylesheet:
image

The previous mechanism for publishing javadoc was broken because it combines the classpath of all subproject dependencies leading to some version conflicts. This is why the publication was disabled here: https://github.com/getsentry/sentry-java/pull/4445

I checked the build folder that the correct javadoc was published but wasn't able to test the github action.

## :pencil: Checklist


build.gradle.kts
```diff
@@ -246,31 +247,6 @@ spotless {
     }
 }
 
-tasks.register("aggregateJavadocs", Javadoc::class.java) {
-    setDestinationDir(project.layout.buildDirectory.file("docs/javadoc").get().asFile)
```
runningcode, line 250: This would previously put all the javadocs from all subprojects in the root which meant a lot of the files were overwritten like `index.html`.

build.gradle.kts
```diff
@@ -237,7 +238,7 @@ spotless {
     kotlin {
         target("**/*.kt")
         ktlint()
-        targetExclude("**/sentry-native/**")
+        targetExclude("**/sentry-native/**", "**/build/**")
```
runningcode, line 241: Had to add this exclusion otherwise spotless tries to format the generated code in the `build` folder. This was causing an issue since the generated accessors have an underscore (`_`) in the path name and this is was failing the check.

build-logic/src/main/kotlin/io/sentry/gradle/AggregateJavadoc.kt
```diff
@@ -0,0 +1,36 @@
...
+    @get:OutputDirectory
+    abstract val outputDir: DirectoryProperty
+
+    @TaskAction
+    fun aggregate() {
+        javadocFiles.get().forEach { file ->
+            fs.copy {
+                // Get the third to last part (project name) to use as the directory name for the output
+                val parts = file.path.split('/')
+                val projectName = parts[parts.size - 4]
```
rbro112, line 29: Should there be some safety around this in case the `parts.size` is < 4?

rbro112 (approved):

build-logic/src/main/kotlin/io/sentry/gradle/AggregateJavadoc.kt
runningcode: This is a good point. To be honest, this logic is a bit hacky. The `ArtifactView` has this information as the component idenitifer, but I can’t have the ArtifactView as an input to the task since it isn’t serializeable. This was my hack around this.

Since these are fully qualified paths it is nearly impossible for them to be less than 4. If they are then we should fail and investigate. The bigger issue is if these are not the project names but I’m not sure how to check for that.

markushi: How about getting the root project folder (`<path>/sentry-java`) and treating that as a prefix which can be removed from `file.path` (`<path>/sentry-java/module/build/docs/javadoc`)?
Something along those lines:
```
file.path.replace(rootPath, "").split('/')[0]
```

build.gradle.kts
```diff
@@ -27,6 +27,7 @@ plugins {
     alias(libs.plugins.errorprone) apply false
     alias(libs.plugins.gradle.versions) apply false
     alias(libs.plugins.spring.dependency.management) apply false
+    id("sentry.javadoc.aggregate")
```
markushi: It's probably best to use the full `io.sentry.**` namespace instead of `sentry.**` for the plugin ids.

settings.gradle.kts
```diff
@@ -17,7 +17,7 @@ dependencyResolutionManagement {
 
 rootProject.name = "sentry-root"
 rootProject.buildFileName = "build.gradle.kts"
-
+includeBuild("build-logic")
```
markushi, line 20: Is it common convention to use dashes here? I know we have `buildSrc`(camel case) and I'm wondering if we should align the naming here.

markushi (approved):
Looks great, thank you 🙏

build-logic/src/main/kotlin/io/sentry/gradle/AggregateJavadoc.kt
romtsn: I think there's also `relativize` or something in Kotlin which will return the path relative to root in this case

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
```
romtsn, line 5:
```suggestion
### Fixes

```

build-logic/src/main/kotlin/sentry.javadoc.gradle.kts
```diff
@@ -0,0 +1,27 @@
+val javadocConfig: Configuration by configurations.creating {
```
romtsn:
```suggestion
val javadocPublisher: Configuration by configurations.creating {
```
maybe call it a `publisher` so it better aligns with `consumer`?

romtsn (approved):
LGTM, great findings and fixes 👍

settings.gradle.kts
runningcode: Good point. I think the general convention in the gradle community is dashes. All our other projects in this `sentry-java` are also dashes.
Gradle’s [sharing build logic sample](https://docs.gradle.org/current/samples/sample_convention_plugins.html) also uses kebab case (aka dashes).

In addition [name abbreviation](https://docs.gradle.org/current/userguide/command_line_interface.html#sec:name_abbreviation) works for both camel case and kebab case names.

build.gradle.kts
runningcode: It’s a good point, i wanted to distinguish this internal plugin from externally published plugins. I’ll do the switch but maybe its worth discussing how to distinguish internal vs external plugins.

build-logic/src/main/kotlin/io/sentry/gradle/AggregateJavadoc.kt
runningcode: This is a good point. I can add the root directory to the task in order to compute the relative path.
Then we also need to remove the `build/doc/javadoc` from the end of the path. I will do this.

CHANGELOG.md
runningcode: Added back the space, ty

PR #4441, opened 2025-05-22
Ryan/version catalog 10

rbro112:
## :scroll: Description
Version catalog adoption

#skip-changelog

## :pencil: Checklist

- [ ] I added tests to verify the changes.
- [ ] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [ ] Review from the native team if needed.
- [ ] No breaking change or entry added to the changelog.
- [ ] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

PR #4463, opened 2025-05-29
Move compileOnly dependencies to use version catalog

rbro112:
## :scroll: Description
Version catalog adoption cont'd

#skip-changelog

## :pencil: Checklist

### DIFF
--- a/buildSrc/src/main/java/Config.kt
+++ b/buildSrc/src/main/java/Config.kt
@@ -5,8 +5,6 @@ object Config {
     val AGP = System.getenv("VERSION_AGP") ?: "8.6.0"
     val kotlinStdLib = "stdlib-jdk8"
 
-    val springBootVersion = "2.7.18"
-    val springBoot3Version = "3.5.0"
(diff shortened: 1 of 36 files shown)


buildSrc/src/main/java/Config.kt
```diff
@@ -5,8 +5,6 @@ object Config {
     val AGP = System.getenv("VERSION_AGP") ?: "8.6.0"
     val kotlinStdLib = "stdlib-jdk8"
 
-    val springBootVersion = "2.7.18"
```
rbro112, line 8: Were unused as of https://github.com/getsentry/sentry-java/pull/4441

gradle/libs.versions.toml
```diff
@@ -1,10 +1,11 @@
 [versions]
 androidxNavigation = "2.4.2"
 androidxTestCore = "1.6.1"
-androidxComposeVersion = "1.6.3"
```
rbro112, line 4: Just naming change to remove "Version" as it's redundant

PR #4469, opened 2025-06-04
fix(okhttp): Finalize OkHttpEvent even if no active span

romtsn:
## :scroll: Description

- When SentryOkHttpInterceptor is used in conjunction with SentryOkHttpEventListener and there's no active span we were not properly finalizing OkHttpEvent. And edge-case could happen when OkHttp is used directly and the response body is not closed, resulting in OkHttpEvent never being finalized.

Found in our sales-demo app

Automated

## :pencil: Checklist

### DIFF
--- a/sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
+++ b/sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
@@ -192,6 +192,7 @@ public open class SentryOkHttpInterceptor(
 
     private fun finishSpan(span: ISpan?, request: Request, response: Response?, isFromEventListener: Boolean, okHttpEvent: SentryOkHttpEvent?) {
         if (span == null) {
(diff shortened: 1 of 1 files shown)


sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
```diff
@@ -192,6 +192,7 @@ public open class SentryOkHttpInterceptor(
 
     private fun finishSpan(span: ISpan?, request: Request, response: Response?, isFromEventListener: Boolean, okHttpEvent: SentryOkHttpEvent?) {
         if (span == null) {
+            okHttpEvent?.finish()
```
stefanosiano, line 196: Can you add a comment here describing the reason, please?

romtsn, line 196:
```suggestion
            // tracing can be disabled, or there can be no active span, but we still want to finalize the OkHttpEvent when both SentryOkHttpInterceptor and SentryOkHttpEventListener are used
            okHttpEvent?.finish()
```

PR #4476, opened 2025-06-06
Remove clean before test

runningcode:
## :scroll: Description
This removes the redundant dependency on `cleanTest` from each test task. This still allows the test to come from the cache so it doesn't really make sense.
Also add build caching for spotless format execution
#skip-changelog

## :pencil: Checklist


runningcode:
Ah yeah, the one about having a hermetic testing environment? Still hasn’t become a thing unfortunately.

PR #4477, opened 2025-06-06
fix(replay): Fix crash on devices with the Unisoc/Spreadtrum T606 chipset

### LINKED ISSUE #4411: [Session Replay] Crashes on `Spreadtrum T606` chipset
### Description

We have reports that devices using the `Spreadtrum T606` chipset experience crashes when Session Replay is active:
- [internal report](https://linear.app/getsentry/issue/MOBILE-920)
- reported on https://github.com/getsentry/sentry-android-gradle-plugin/issues/861#issuecomment-2867021256

Since only a specific chipset is affected, this is potentially a bug in the chipset. We need to investigate if there is an effective fix, or if we need to deactivate SR on these devices.

romtsn:
## :scroll: Description

- Check chipset (only available on API 31 and above) for the faulty one and use software canvas instead of hardware canvas.

See replay from one of the devices: https://sentry-sdks.sentry.io/replays/17e3e42695c945378af6689e2aa35b7b/

Fixes #4411 

Manually on one of the faulty devices

## :pencil: Checklist

### DIFF
--- a/sentry-android-replay/src/main/java/io/sentry/android/replay/util/SystemProperties.kt
+++ b/sentry-android-replay/src/main/java/io/sentry/android/replay/util/SystemProperties.kt
@@ -1,0 +1,34 @@
+package io.sentry.android.replay.util
+
+import android.annotation.SuppressLint
+import android.os.Build
+import java.lang.reflect.Method
+
+internal object SystemProperties {
+    // from https://cs.android.com/android/platform/superproject/main/+/main:out/soong/.intermediates/system/libsysprop/srcs/PlatformProperties/android_common/xref/srcjars.xref/android/sysprop/SocProperties.java;l=163-171
+    // these props are not available on API < 31 via Build, so we use reflection to access them
+    const val SOC_MODEL = "ro.soc.model"
+    const val SOC_MANUFACTURER = "ro.soc.manufacturer"
+
+    @delegate:SuppressLint("PrivateApi")
+    private val getProperty: Method by lazy {
+        val clazz = Class.forName("android.os.SystemProperties")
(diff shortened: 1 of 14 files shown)


sentry-android-replay/src/main/java/io/sentry/android/replay/util/SystemProperties.kt
```diff
@@ -0,0 +1,34 @@
...
+    @delegate:SuppressLint("PrivateApi")
+    private val getProperty: Method by lazy {
+        val clazz = Class.forName("android.os.SystemProperties")
+        clazz.getMethod("get", String::class.java)
+    }
+
+    fun get(key: String, defaultValue: String = ""): String {
+        return if (Build.VERSION.SDK_INT < Build.VERSION_CODES.S) {
+            try {
+                getProperty.invoke(null, key) as? String ?: defaultValue
```
romtsn: I actually have a feeling this will always return an empty string on API < 31, see (from Jan'21) https://android.googlesource.com/platform/system/libsysprop/+/ba383e1deb52077aa4ae446a983bbe9bf5497a34%5E1..ba383e1deb52077aa4ae446a983bbe9bf5497a34/, so we might as well just drop it.

According to the web, the T606 processor was introduced later that year (Sep'21), so there's a good chance it's not used on devices with API < 31.

image

markushi (approved):
LGTM!

sentry-android-replay/src/main/java/io/sentry/android/replay/util/SystemProperties.kt
markushi: Yeah I'd drop it then

Issue #4479, opened 2025-06-09
SentryStackFrame not serializing/deserializing preContext & postContext


limbonaut:
### Integration

sentry-android

### Build System

Gradle

### AGP Version

8.9.2

### Proguard

Enabled

### Version

8.11.1

### Steps to Reproduce

1. Create `SentryStackFrame` object.
2. Set `filename`, `function`, `lineno`, `contextLine`, `preContext` and `postContext`. 
3. Add it to an event stack frames and capture it.
4. Check JSON output.

### Expected Result

Having `pre_context` and `post_context` in the JSON `stacktrace->frames`.

### Actual Result

Both `preContext` and `postContext` are not serialized in `serialize()`. Take a look at [SentryStackFrame code](https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/protocol/SentryStackFrame.java): `preContext` and `postContext` are unused variables.

<img width="760" alt="Image" src="https://github.com/user-attachments/assets/696139f8-9506-42f1-bd15-fbe2c9a92155" />

limbonaut:
I discovered this issue while working on the Android support in Sentry for Godot SDK: https://github.com/getsentry/sentry-godot/pull/169

PR #4482, opened 2025-06-10
Serialize `preContext` and `postContext` in `SentryStackFrame`

lcian:
## :scroll: Description

Serializes and deserializes the `preContext` and `postContext` fields in `SentryStackFrame`.
They don't seem to be used anywhere in the SDK, so perhaps that's why they were ignored before.

Closes https://github.com/getsentry/sentry-java/issues/4479

Updated existing test.

## :pencil: Checklist


sentry/src/main/java/io/sentry/protocol/SentryStackFrame.java
```diff
@@ -411,6 +413,12 @@ public void serialize(final @NotNull ObjectWriter writer, final @NotNull ILogger
     if (lock != null) {
       writer.name(JsonKeys.LOCK).value(logger, lock);
     }
+    if (preContext != null && !preContext.isEmpty()) {
+      writer.name(JsonKeys.PRE_CONTEXT).value(logger, preContext);
```
romtsn, line 417:
```suggestion
      writer.name(JsonKeys.PRE_CONTEXT).value(preContext);
```
since we know it's a string can we use the normal `value` overload here to avoid going through the reflection-based serializer?

sentry/src/main/java/io/sentry/protocol/SentryStackFrame.java
```diff
@@ -411,6 +413,12 @@ public void serialize(final @NotNull ObjectWriter writer, final @NotNull ILogger
...
+    }
+    if (postContext != null && !postContext.isEmpty()) {
+      writer.name(JsonKeys.POST_CONTEXT).value(logger, postContext);
```
romtsn, line 420:
```suggestion
      writer.name(JsonKeys.POST_CONTEXT).value(postContext);
```

sentry/src/main/java/io/sentry/protocol/SentryStackFrame.java
romtsn: ah sorry, i noticed it's a list, disregard!

romtsn (approved):
lgtm, but can we please add a changelog entry? Got bitten a few times where we didn't mention things and I was looking for them

lcian:
> lgtm, but can we please add a changelog entry? Got bitten a few times where we didn't mention things and I was looking for them

Thanks, sure, adding it!

Issue #4467, opened 2025-05-30
Text compostables not masked in session replay


jvanderwee:
### Integration

sentry-android

### Build System

Gradle

### AGP Version

8.10.0

### Proguard

Enabled

### Version

8.13.2

### Steps to Reproduce

1. Enable session replay like so:

```
        <meta-data
            android:name="io.sentry.session-replay.on-error-sample-rate"
            android:value="1.0" />
        <meta-data
            android:name="io.sentry.session-replay.session-sample-rate"
            android:value="0.1" />
```

2. Display Material 3 Text composable e.g.:

```
@Composable
private fun HelloText() {
        androidx.compose.material3.Text("Hello")
}
```

### Expected Result

Text composable is masked

### Actual Result

Text composable is not masked

romtsn:
hi @jvanderwee, thanks for reporting! Which Compose version do you use?

romtsn:
thanks, I believe I know what the problem is: Compose 1.8+ changed how they handle semantics and we're not able to retrieve them. There should be a warning printed out in logcat somewhere. We'll prioritize this!

PR #4485, opened 2025-06-11
Fix broken view hierarchy retrieval for Jetpack Compose 1.8+

markushi:
## :scroll: Description

Fixes missing masking for Jetpack Compose 1.8+ as `LayoutNode.collapsedSemantics` was removed in favor of `LayoutNode.getSemanticsConfiguration()`

We used to swallow errors when retrieving the Compose ViewHierarchy, leading to un-masked replays. After some discussion we're changing this to a more defensive approach: If parts of the View Hierarchy can't be retrieved we simply mask the relevant parts of the UI.

Fixes https://github.com/getsentry/sentry-java/issues/4467

Added tests + manual testing.

Before:

After:

## :pencil: Checklist


sentry-android-replay/src/main/java/io/sentry/android/replay/viewhierarchy/ComposeViewHierarchyNode.kt
```diff
@@ -83,16 +109,53 @@ internal object ComposeViewHierarchyNode {
             _rootCoordinates = WeakReference(node.coordinates.findRootCoordinates())
         }
 
-        val semantics = node.collapsedSemantics
         val visibleRect = node.coordinates.boundsInWindow(_rootCoordinates?.get())
+        val semantics: SemanticsConfiguration?
+
+        try {
+            semantics = retrieveSemanticsConfiguration(node)
+        } catch (t: Throwable) {
```
stefanosiano, line 117: 👍

sentry-android-replay/src/main/java/io/sentry/android/replay/viewhierarchy/ComposeViewHierarchyNode.kt
```diff
@@ -83,16 +109,53 @@ internal object ComposeViewHierarchyNode {
...
+                x = visibleRect.left.toFloat(),
+                y = visibleRect.top.toFloat(),
+                width = node.width,
+                height = node.height,
+                elevation = (parent?.elevation ?: 0f),
+                distance = distance,
+                parent = parent,
+                shouldMask = true,
+                isImportantForContentCapture = false, /* will be set by children */
+                isVisible = !node.outerCoordinator.isTransparent(),
```
stefanosiano: do you also need `&& visibleRect.height() > 0 && visibleRect.width() > 0`?

stefanosiano (approved):
just a little doubt, but looks good!

sentry-android-replay/src/main/java/io/sentry/android/replay/viewhierarchy/ComposeViewHierarchyNode.kt
```diff
@@ -83,16 +109,53 @@ internal object ComposeViewHierarchyNode {
...
+                Error retrieving semantics information from Compose tree. Most likely you're using
+                an unsupported version of androidx.compose.ui:ui. The supported
+                version range is 1.5.0 - 1.8.0.
+                If you're using a newer version, please open a github issue with the version
+                you're using, so we can add support for it.
+                    """.trimIndent()
+                )
+            }
+
+            // If we're unable to retrieve the semantics configuration
```
romtsn, line 133: I'm thinking if we should check if any masking is actually enabled? Thinking of a case where nothing should be masked, but we still would weirdly mask something when failed to retrieve the semantics.

romtsn: (maskAllText and maskAllImages, that is)

romtsn (approved):
one comment, but otherwise LGTM, great workaround and tests!

sentry-android-replay/src/main/java/io/sentry/android/replay/viewhierarchy/ComposeViewHierarchyNode.kt
markushi: I'm wondering if we should even process the view hierarchy if no masking is active? I'll get this merged now, as by default masking is active anyway. We can always improve on top.

Issue #4484, opened 2025-06-11
Sentry <> Timber integration: Missing `TimberTag` tag


gent-ahmeti:
### Integration

sentry-android

### Build System

Gradle

### AGP Version

8.6.0

### Proguard

Enabled

### Sentry Version

7.4.0

### Timber Version

5.0.1

### Steps to Reproduce

1. Report issue like so `Timber.tag("MyTag").w(throwable, "Something went wrong")` (It's possible that it might be working for `Timber.e()`)
2. Open Sentry dashboard and locate the aforementioned issue

### Expected Result

In the list of tags we should see `TimberTag` [tag](https://docs.sentry.io/platforms/android/enriching-events/tags/) with the value`MyTag`

As mentioned by [the changelog](https://docs.sentry.io/platforms/android/migration/#migrating-to-iosentrysentry-android-timber-562), `TimberTag` is supported from version `5.7.2` onwards
> Support for Timber.tag has been brought back in version 5.7.2.

### Actual Result

When checking tags you can notice that `TimberTag` tag is missing completely. See attached screenshot with what I mean by `tag`
![Image](https://github.com/user-attachments/assets/81af8789-aa27-4996-800c-23e33fd9a6c7)

markushi:
@gent-ahmeti thanks for opening this up! We'll have a look into this. Could also share the version of Timber you're using?

gent-ahmeti:
Hi @markushi , we are using Timber version 5.0.1. I'll update the description to have this version too

markushi:
@gent-ahmeti thank you! Could you also share some more details on how you adapted it for the usage with Coroutines? Timber itself utilizes `ThreadLocal` for tags which could cause troubles with Coroutines:

https://github.com/JakeWharton/timber/blob/beb8051248164a74b264d30427f633aaf4bda841/timber/src/main/java/timber/log/Timber.kt#L19-L23

gent-ahmeti:
Looks like we are catching the "tag" before we launch the coroutine. See abbreviated code:
```kotlin
private val pendingTag = ThreadLocal<String?>()

private fun retrieveTag(): String? {
    val tag = pendingTag.get()
    if (tag != null) {
        this.pendingTag.remove()
    }
    return tag
}

private fun logWithSentry(
        priority: Int,
        throwable: Throwable?,
        message: String?,
        vararg args: Any?
    ) {
        val tag = retrieveTag()
        // ...
[...]

PR #4489, opened 2025-06-13
Fix tag extraction logic, bump to Timber 5.0.1

markushi:
## :scroll: Description
Timber 5.+ switched to Kotlin, allowing us to directly access the internal `tag` value, instead of the existing workaround.

According to the release Timber 5.x is binary compatible: https://github.com/JakeWharton/timber/releases/tag/5.0.0 so we should be good with just bumping the version here.

Fixes https://github.com/getsentry/sentry-java/issues/4484

Added extra unit tests.

We should still manually test compatibility with Timber 4.x, as [the internal tag value](https://github.com/JakeWharton/timber/blob/84b157605552e7b30f0d20e105d383b48ce8a030/timber/src/main/java/timber/log/Timber.java#L377) was declared the following:
```java
final ThreadLocal<String> explicitTag = new ThreadLocal<>();
```

And [with 5.x](https://github.com/JakeWharton/timber/blob/beb8051248164a74b264d30427f633aaf4bda841/timber/src/main/java/timber/log/Timber.kt#L21-L22) it's
```kotlin
@get:JvmSynthetic // Hide from public API.
internal val explicitTag = ThreadLocal<String>()
```

## :pencil: Checklist


sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
```diff
@@ -17,23 +19,15 @@ public class SentryTimberTree(
...
-        }
-        return tag
-    }
 
     /** Log a verbose message with optional format args. */
     override fun v(
         message: String?,
         vararg args: Any?
     ) {
+        val tag = explicitTag.get()
```
romtsn, line 28: hm, but am I missing something or this ain't gonna work for Timber < 5? Because we only define it as compileOnly so if the hosting app is using a lower version it will crash?

markushi:
Closing this PR as it only works for Timber 5.x and does not seem to fix the underlying root issue of ThreadLocal usages within the Timber library.

PR #3814, opened 2024-10-23
ci(build): Lower gradle workers to max 2

krystofwoldrich:
## :scroll: Description

Lower is number of gradle workers to 2. With this the build job passes without timeout.

The timeout started after https://github.com/getsentry/sentry-java/pull/3802

## :bulb: Motivation and Context

- https://github.com/getsentry/sentry-java/pull/3813

## :green_heart: How did you test it?
- ci https://github.com/getsentry/sentry-java/actions/runs/11477696719

## :pencil: Checklist

- [x] I reviewed the submitted code.
- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [x] I updated the docs if needed.
- [x] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

## :crystal_ball: Next steps
#skip-changelog

PR #4496, opened 2025-06-18
ALL THE CORES

runningcode:
## :scroll: Description
Use all of them!

#skip-changelog

## :pencil: Checklist

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -7,9 +7,6 @@ org.gradle.configuration-cache=true
 
 org.jetbrains.dokka.experimental.gradle.pluginMode=V2Enabled
 
-# Daemons workers
-org.gradle.workers.max=2
-
 # AndroidX required by AGP >= 3.6.x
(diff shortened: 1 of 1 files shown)


stefanosiano (approved):
If i remember correctly we did this when trying to fix CI freezing randomly.
Our suspect was out of memory because of too many things running at the same time.
Let's do this (things improved in the meantime), and keep it in mind if we have other problems with CI again

romtsn:
> If i remember correctly we did this when trying to fix CI freezing randomly. Our suspect was out of memory because of too many things running at the same time. Let's do this (things improved in the meantime), and keep it in mind if we have other problems with CI again

yup it was this PR https://github.com/getsentry/sentry-java/pull/3814. Unsure if it got better, but we can try again!

runningcode:
Good to know the context. If this is a CI only thing then we should set the workers to 2 only on CI. I’ll do that if this happens again.

PR #4499, opened 2025-06-18
More accurate, faster spotless config

runningcode:
## :scroll: Description

This PR ensures that spotless is actually running over all our files. It
wasn't before and you can see this generated a massive diff.
There are still a number of exceptions that were suppressed in order to
save time going through the hundreds of them.

In other news, this moves the spotless logic to its own convention
plugin and runs over each source set of each project separately. This
means that most of the tasks should be cached independently.

#skip-changelog

## :pencil: Checklist


build-logic/src/main/kotlin/io.sentry.spotless.gradle.kts
```diff
@@ -0,0 +1,37 @@
+import com.diffplug.spotless.LineEnding
+
+plugins {
+    id("com.diffplug.spotless")
+}
+
+spotless {
```
runningcode, line 7: This is the new logic.

build-logic/src/main/kotlin/io.sentry.spotless.gradle.kts
```diff
@@ -0,0 +1,37 @@
...
+        googleJavaFormat()
+        targetExclude("src/**/java/io/sentry/vendor/**")
+    }
+    kotlin {
+        target("**/*.kt")
+        ktlint()
+        targetExclude("src/test/java/io/sentry/apollo4/generated/**", "src/test/java/io/sentry/apollo3/adapter/**")
+        suppressLintsFor {
+            step = "ktlint"
+            shortCode = "standard:max-line-length"
```
runningcode: `max-line-length`, `property-naming` and `function-naming` had hundreds of violations that aren’t auto-fixable and would have made this PR even larger so i just supressed them for now.

lcian (approved):
I suggest adding this to `.git-blame-ignore-revs`

build-logic/src/main/kotlin/io.sentry.spotless.gradle.kts
```diff
@@ -0,0 +1,37 @@
...
+    lineEndings = LineEnding.UNIX
+    java {
+        target("**/*.java")
+        removeUnusedImports()
+        googleJavaFormat()
+        targetExclude("src/**/java/io/sentry/vendor/**")
+    }
+    kotlin {
+        target("**/*.kt")
+        ktlint()
```
romtsn: since we're introducing a huge diff, I'm wondering if we should use this as a chance to switch over to ktfmt? I think it's less troublesome to maintain as it pretty-prints the entire file as opposed to modifying in-place like ktlint. I guess we could also update `.editorconfig` to reflect that one of the Sentry Gradle Plugin repo.

runningcode: Happy to do that!

romtsn (approved):

Issue #4334, opened 2025-04-14
Setting maxBreadcrumbs 0 and beforeBreadcrumb callback crash


mathisfouques:
### Integration

sentry-android

### Build System

Gradle

### AGP Version

7.4.2

### Proguard

Enabled

### Version

7.22.4

### Steps to Reproduce

Initially had the problem with my app that integrates flutter with sentry and therefore the sentry android SDK (with version 7.22.4, which is the one used by sentry sdk for flutter 8.14.1). 

I found this PR #3836 which allowed maxBreadcrumbs to 0 , because of an initial react native sdk problem ([here](https://github.com/getsentry/sentry-react-native/issues/3634))

I cloned this repo https://github.com/sentry-demos/android to test the example with just android sdk, and in fact : 

On main on this repo : 

- Go to MyBaseActivity 
- Add to configureScope callback : 
```
scope.getOptions().setMaxBreadcrumbs(0);
scope.getOptions().setBeforeBreadcrumb( (breadcrumb, hint) -> {
    return null;
        }
);

```

### Expected Result

Surely not a crash of the app, but it does crash 

### Actual Result

Crash, stack trace on my side : 

```
ANR in com.example.vu.android (com.example.vu.android/.empowerplant.EmpowerPlantActivity)
PID: 31047
Reason: Input dispatching timed out (29bd6f9 com.example.vu.android/com.example.vu.android.empowerplant.EmpowerPlantActivity is not responding. Waited 5005ms for MotionEvent).
Parent: com.example.vu.android/.empowerplant.EmpowerPlantActivity
```

We solved it by removing the maxBreadcrumbs to 0 on our side, but it shouldnt be available to do both if it makes the app crash

stefanosiano:
hi @mathisfouques 
thanks for the feedback!
It looks like you're experiencing an ANR, which then causes the crash. Anyway, the SDK should work fine in this case.
We'll try to reproduce it and we'll let you know!

PR #4504, opened 2025-06-20
Optimize when maxBreadcrumb is 0

stefanosiano:
## :scroll: Description
breadcumbs queue is now a simple DisabledQueue when maxBreadcrumb is 0
if breadcumbs is a DisabledQueue, no more processing is done (beforeBreadcrumb or scopeObservers)

Fixes https://github.com/getsentry/sentry-java/issues/4334
Previously we were calling beforeBreadcrumb and scopeObservers even when maxBreadcrumbs was 0, which was useless

## :pencil: Checklist


sentry/src/main/java/io/sentry/Scope.java
```diff
@@ -862,7 +862,7 @@ public void clearAttachments() {
   static @NotNull Queue<Breadcrumb> createBreadcrumbsList(final int maxBreadcrumb) {
     return maxBreadcrumb > 0
         ? SynchronizedQueue.synchronizedQueue(new CircularFifoQueue<>(maxBreadcrumb))
-        : SynchronizedQueue.synchronizedQueue(new DisabledQueue<>());
+        : new DisabledQueue<>();
```
stefanosiano, line 865: I tried calling queue.add from multiple threads concurrently, and it didn't crash, so i think we it's fine to remove SynchronizedQueue

romtsn (approved):

PR #4524, opened 2025-06-26
chore(docs): fix javadoc link

JoshuaMoelans:
like Roman mentioned below, fixing gh-pages is a better solution 😅 didn't notice we had something being deployed on there, but that it's just the modules overview not existing

---

Just happened to spot a dead link in the Readme pointing to https://getsentry.github.io/sentry-java/ , and saw on [this issue](https://github.com/getsentry/sentry-java/issues/4522) that [javadoc.io](https://javadoc.io/doc/io.sentry/sentry/latest/index.html) has javadocs for the latest release. 

Not sure if we want to point at javadoc.io instead of our own github page, but having a dead link is not much better either :p 

#skip-changelog

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ This repo uses the following ways to release SDK updates:
 # Useful links and docs
 
 * A deep dive into how we built [Session Replay for Android](https://www.droidcon.com/2024/11/22/rewind-and-resolve-a-deep-dive-into-building-session-replay-for-android/) at Droidcon London 2024.
-* Current Javadocs [generated from source code](https://getsentry.github.io/sentry-java/).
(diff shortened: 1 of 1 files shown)


romtsn:
I'd rather us fix our gh-pages deployments because javadoc.io takes a while to sync the latest version and also does not include all modules

JoshuaMoelans:
@romtsn I just looked at the [gh-pages branch](https://github.com/getsentry/sentry-java/tree/gh-pages) and saw that we can point at specific modules (e.g. https://getsentry.github.io/sentry-java/sentry/) , so yeah since it is not fully empty, think using the existing pages is better than changing it to the javadocs site indeed ^

JoshuaMoelans:
Superseded by https://github.com/getsentry/sentry-java/issues/4525

PR #4542, opened 2025-07-07
ref(replay): Use main thread to schedule capture

romtsn:
## :scroll: Description

- This reworks the logic to schedule screenshot + VH capture on the main thread using Handler.postDelayed
  - We were anyway posting a message from the capturer thread to the main thread before, so this gets rid of thread context switch all the time, plus ensures we run view checks (like root.width/height) on the main thread right before we capture

A couple of replay examples with the new capture logic:

https://sentry-sdks.sentry.io/explore/replays/699aa0aff2e2496db716274d637f6a33
https://sentry-sdks.sentry.io/explore/replays/d043fb059023427b97fe54f4abf39832

Manually

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Improvements
+
+- Session Replay: Use main thread looper to schedule replay capture ([#4542](https://github.com/getsentry/sentry-java/pull/4542))
+
 ## 8.18.0
 
 ### Features
@@ -20,10 +26,6 @@
         }
       }
     ```
-
-### Improvements
-
-- Session Replay: Use main thread looper to schedule replay capture ([#4542](https://github.com/getsentry/sentry-java/pull/4542))
 
 ### Fixes
 


sentry-android-replay/src/main/java/io/sentry/android/replay/util/MainLooperHandler.kt
```diff
@@ -6,7 +6,15 @@ import android.os.Looper
 internal class MainLooperHandler(looper: Looper = Looper.getMainLooper()) {
```
stefanosiano, line 6: I'd name it (and its references) `LooperHandlerWrapper`, as it's more clear that it's not just to run on the main thread, even if i think you did it just for testing

stefanosiano (approved):
looks good
just a small renaming

sentry-android-replay/src/main/java/io/sentry/android/replay/util/MainLooperHandler.kt
romtsn: I think I'd keep it like that, otherwise would have to change it also in sentry-android-core. Not sure if LooperHandlerWrapper will be any more descriptive tbh 😅

PR #4546, opened 2025-07-08
Rename minLogsLevel option in SentryTimberIntegration

stefanosiano:
## :scroll: Description

renamed `minLogsLevel` to `minLogLevel` in `SentryTimberIntegration`

#skip-changelog

## :pencil: Checklist


romtsn:
Since this is a breaking change now we should probably do it as part of a new major. Besides, it's probably fine as it is.

PR #4554, opened 2025-07-14
Improve setup for backend e2e tests

lcian:
## :scroll: Description

Stacked on: https://github.com/getsentry/sentry-java/pull/4552
- Add an envrc so that the virtualenv is created automatically (if you have direnv and allow the dir) -- it is not strictly required to use direnv though
- Add requirements.txt and use it in the GH workflow and makefile
- Add test outputs to gitignore
- Use a dataclass for the test run configs and improve the output a bit

My editor added some whitespace changes due to autoformatting.

#skip-changelog

### DIFF
--- a/test/system-test-runner.py
+++ b/test/system-test-runner.py
@@ -494,14 +494,14 @@ class SystemTestRunner:
 
         failed_tests = []
 
-        for module_config in test_configs:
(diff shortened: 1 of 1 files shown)


adinauer (approved):
Nice, Thanks!

PR #4560, opened 2025-07-16
perf(connectivity): Cache network capabilities and status to reduce IPC calls

### LINKED ISSUE #4200: Cache current network connection state
### Description

As a follow-up of https://github.com/getsentry/sentry-java/pull/4164 we want to cache the last known connection state, so calling `AndroidConnectionStatusProvider.getConnectionStatus()` returns a cached value.

romtsn:
## :scroll: Description

- Register network callback on a bg thread upon AndroidConnectionStatusProvider instantiation
- Cache latest available NetworkCapabilities for 2 minutes
- Use the capabilities to determine the current network status and type
  - If it's not available (cache expired), we'd fallback to retrieving it proactively and caching it
  - pre-API-23 we'd be always querying it, but it's a very small number of devices
- `onLost`/`onUnavailable` would immediately send DISCONNECTED now as opposed to retrieving `connectionStatus` again
- Implement a `isSignificantChange` check, because `onNetworkCapabilitiesChanged` can be called many times due to bandwidth change/etc. and we're only interested in transport/connectivity changes
- Unregister the network callback when the SDK is closing
- Cache ConnectivityManager instance too (it should be safe to cache once per process as it's not dependant on any particular Context)

Ran a test with perfetto to count the number of binder calls, and these are the results:
- Send app to background
- Freeze it
- Turn wifi on and off 5 times,
- Unfreeze it back

### Before (83 binder calls from SentryExecService)
image

### After (5 binder calls from SentryExecService)
image

Fixes #4200

Manually + automated

## :pencil: Checklist

## :crystal_ball: Next steps
- Refactor NetworkBreadcrumbsIntegration so we register only one callback in AndroidConnectionStatusProvider and multiplex it
- Register/unregister connectivity callback in foreground/background similar to SystemEvents integration


sentry-android-core/src/main/java/io/sentry/android/core/internal/util/AndroidConnectionStatusProvider.java
```diff
@@ -31,104 +34,402 @@
...
+                    : null;
           } else {
-            return false;
+            cachedNetworkCapabilities =
+                null; // Clear cached capabilities if connectivity manager is null
           }
         }
+        lastCacheUpdateTime = timeProvider.getCurrentTimeMillis();
+
+        options
```
romtsn, line 335: could drop this log call, but I thought it might be very useful to investigate potential issues. Under the hood getConnectionStatus and getConnectionType are very lightweight for a cache NetworkCapabilities instance, they just do bitmask operations to get the transport and capability

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/AndroidConnectionStatusProvider.java
```diff
@@ -31,104 +34,402 @@
...
-    // underlying network
 
-    // when min. API 24, use:
-    // connectivityManager.registerDefaultNetworkCallback(...)
+    if (!hasInternetAndValidated) {
+      return false;
+    }
+
+    // Additionally, ensure it's a recognized transport type for general internet access
+    return networkCapabilities.hasTransport(NetworkCapabilities.TRANSPORT_WIFI)
```
markushi: We could iterate the transports here instead
```suggestion
 for (final int transport in transports) {
   if (networkCapabilities.hasTransport(transport)) {
     return true;
   }
 }
```

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/AndroidConnectionStatusProvider.java
```diff
@@ -31,104 +34,402 @@
...
+  private final @NotNull ICurrentDateProvider timeProvider;
   private final @NotNull List<IConnectionStatusObserver> connectionStatusObservers;
   private final @NotNull AutoClosableReentrantLock lock = new AutoClosableReentrantLock();
   private volatile @Nullable NetworkCallback networkCallback;
 
+  private static final @NotNull AutoClosableReentrantLock connectivityManagerLock =
+      new AutoClosableReentrantLock();
+  private static volatile @Nullable ConnectivityManager connectivityManager;
+
+  private static final int[] transports = {
```
markushi, line 48: I'd add `TRANSPORT_BLUETOOTH` here, e.g. for wearables without cellular connection.

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/AndroidConnectionStatusProvider.java
```diff
@@ -31,104 +34,406 @@
...
-    this.logger = logger;
+    this.options = options;
     this.buildInfoProvider = buildInfoProvider;
+    this.timeProvider = timeProvider;
     this.connectionStatusObservers = new ArrayList<>();
+
+    capabilities[0] = NetworkCapabilities.NET_CAPABILITY_INTERNET;
+    if (buildInfoProvider.getSdkInfoVersion() >= Build.VERSION_CODES.M) {
+      capabilities[1] = NetworkCapabilities.NET_CAPABILITY_VALIDATED;
+    }
```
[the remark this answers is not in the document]
romtsn: While there might be multiple SDK instances running at a time, it's not possible for them to run on different API levels at the same time, because the device can only target one API level at a time.

PR #4561, opened 2025-07-16
fix(breadcrumbs): Deduplicate battery breadcrumbs

romtsn:
## :scroll: Description

- Remember last state of the battery breadcrumbs 
  - Do not capture if the level or charging state is the same (or similar)

Closes https://github.com/getsentry/sentry-react-native/issues/4668

Manually + automated

## :pencil: Checklist


sentry-android-core/src/main/java/io/sentry/android/core/SystemEventsBreadcrumbsIntegration.java
```diff
@@ -339,6 +339,43 @@ static final class SystemEventsBroadcastReceiver extends BroadcastReceiver {
...
+      }
+
+      @Override
+      public int hashCode() {
+        // Use rounded level for hash consistency
+        Float roundedLevel = level != null ? Math.round(level * 100f) / 100f : null;
+        return Objects.hash(roundedLevel, charging);
+      }
+
+      private boolean isSimilarLevel(final @Nullable Float level1, final @Nullable Float level2) {
```
markushi: IMHO I'd rather store the battery level as an integer value directly, saves us some extra logic. I guess it's fine to have the breadcrumb report an integer value too.

romtsn: yeah sounds good, I also was hesitating with this logic tbh. Will adjust 👍

PR #4562, opened 2025-07-18
perf(connectivity): Have only one NetworkCallback active at a time

romtsn:
## :scroll: Description

- Have only single NetworkCallback registered at a time
  - Multi-cast it to the internal child NetworkCallbacks (NetworkBreadcrumbsIntegration)
- Remove executor in NetworkBreadcrumbsIntegration since we just add a callback to the list now instead of doing any IPC
- While testing it I noticed replay wasn't always catching network_available breadcrumb due to timing issues, so I special-cased it in this PR too

Event example: https://sentry-sdks.sentry.io/issues/6560149804/events/960094050aa341cca6ac9d0c175f699e/
Replay example: https://sentry-sdks.sentry.io/explore/replays/fff819393dd44cfa9a06b54255664529

Less IPC

Manually + automated

## :pencil: Checklist


sentry-android-core/src/main/java/io/sentry/android/core/NetworkBreadcrumbsIntegration.java
```diff
@@ -59,87 +53,61 @@ public void register(final @NotNull IScopes scopes, final @NotNull SentryOptions
...
-                });
-      } catch (Throwable t) {
-        logger.log(SentryLevel.ERROR, "Error submitting NetworkBreadcrumbsIntegration task.", t);
+      try (final @NotNull ISentryLifecycleToken ignored = lock.acquire()) {
+        networkCallback =
+            new NetworkBreadcrumbsNetworkCallback(
+                scopes, buildInfoProvider, options.getDateProvider());
+
+        final boolean registered =
+            AndroidConnectionStatusProvider.addNetworkCallback(
```
romtsn, line 79: one caveat: This now will **not** emit AVAILABLE and CAPABILITIES_CHANGED callbacks immediately after registration as opposed to calling `ConnectivityManager.registerDefaultNetworkCallback`. 

But I think it's fine and how it should be - previously any event would always start with NETWORK_AVAILABLE breadcrumb, which could have been misleading, now we're actually adding breadcrumbs whenever there's a change.

sentry-android-core/src/main/java/io/sentry/android/core/SendCachedEnvelopeIntegration.java
```diff
@@ -75,7 +75,9 @@ public void close() throws IOException {
   @Override
   public void onConnectionStatusChanged(
       final @NotNull IConnectionStatusProvider.ConnectionStatus status) {
-    if (scopes != null && options != null) {
+    if (scopes != null
+        && options != null
+        && status != IConnectionStatusProvider.ConnectionStatus.DISCONNECTED) {
```
romtsn, line 80: kind of unrelated to this PR, but I noticed that we also submit a task even if the status was DISCONNECTED, which was unnecessary since we'd anyway bail out in the submitted task itself.

markushi (approved):

PR #4579, opened 2025-07-30
perf(integrations): Do not register for SystemEvents and NetworkCallbacks when launched with background importance

romtsn:
## :scroll: Description

- Do not registerDefaultNetworkCallback when importance is not foreground
  - Will be registered on next foreground
  - Also unregister the network callback on background to avoid being spammed by Android when the app is being unfreezed. This will also invalidate the cached capabilities, but everything else should still work because it would query the network status directly from the OS. It's a small compromise between doing a bit more IPC calls while in background (which should be a small number of things to do anyway) and ensuring the correct connectivity state.
- Do not register BroadcastReceiver in SystemEvents integration when importance is not foreground
  - Will be registered on next foreground
- Do not cache `isForegroundImportance` anymore, because it can actually change during a single process lifetime
- Added a `sdkperf/` folder with explanations and small scripts used to reproduce all kinds of non-standard behavior that the changes from above have been tested against.

### Before
This is a network connectivity showcase. App is frozen, disable/enable network 10 times and unfreeze again. You can see that our network status observer is fired many times in a quick succession. This was resulting in submitting many tasks to our background queue in [SendCachedEnvelopeIntegration](https://github.com/getsentry/sentry-java/blob/dea75aafa78b0917ead5c026fe0cc02136ddab19/sentry-android-core/src/main/java/io/sentry/android/core/SendCachedEnvelopeIntegration.java#L76-L81):

Android Studio 2025-07-30 00 49 04

### After 
Same scenario, but now no network observer is notified, because we're still in background:

Android Studio 2025-07-30 00 49 49

Manually + automated

## :pencil: Checklist


sentry-samples/sentry-samples-android/sdkperf/README.md
```diff
@@ -0,0 +1,78 @@
...
+1. Launch the sample app and send it to background
+2. Freeze it with `adb shell am freeze --sticky io.sentry.samples.android`
+3. Run the `./wifi_flap` script which looses and obtains network connectivity 10 times.
+4. Unfreeze the app with `adb shell am unfreeze io.sentry.samples.android`
+
+You can either watch Logcat or better start a Perfetto trace beforehand and then open it and observe the number of binder calls our SDK is doing to the Connectivity service.
+
+### Solution
+
+We have addressed the issue in [link PR]() by unsubscribing from network connectivity updates when app goes to background and re-subscribing again on foreground.
```
romtsn:
```suggestion
We have addressed the issue in [#4579](https://github.com/getsentry/sentry-java/pull/4579) by unsubscribing from network connectivity updates when app goes to background and re-subscribing again on foreground.
```

sentry-samples/sentry-samples-android/sdkperf/README.md
```diff
@@ -0,0 +1,78 @@
...
+1. Launch the sample app
+2. Kill it with `adb shell am force-stop io.sentry.samples.android`
+3. Now launch a dummy service with `adb shell am start-foreground-service -n io.sentry.samples.android/io.sentry.samples.android.DummyService`. This ensures the app process is run with non-foreground importance.
+4. Follow any of the steps described in the sections above.
+
+Observe (Logcat or Perfetto) that the faulty behaviours are still reproducible.
+
+### Solution
+
+We have addressed the issue in [link PR]() by not registering any of the offending integrations when the hosting app process is launched with non-foreground `importance`. We still keep observing the App Lifecycle to ensure we register the integrations when the App has been brought to foreground.
```
romtsn:
```suggestion
We have addressed the issue in [#4579](https://github.com/getsentry/sentry-java/pull/4579) by not registering any of the offending integrations when the hosting app process is launched with non-foreground `importance`. We still keep observing the App Lifecycle to ensure we register the integrations when the App has been brought to foreground.
```

sentry-samples/sentry-samples-android/sdkperf/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+And watch Logcat for a bunch of `SCREEN_OFF`/`SCREEN_ON` breadcrumbs created microseconds apart.
+
+### Solution
+
+We have addressed the issue in [#4338](https://github.com/getsentry/sentry-java/pull/4338) by unregistering the `BroadcastReceiver` when app goes to background and registering it again on foreground.
+
+## App Launch with Background Importance 
+
+While the above two issues can be fixed by observing the App lifecycle, they still may become a problem if the app process has been launched with non-foreground importance (e.g. received a push notification). In this case our SDK would be initialized too
```
romtsn:
```suggestion
While the above two issues can be fixed by observing the App lifecycle, they still may become a problem if the app process has been launched with non-foreground importance (e.g. received a push notification). In this case our SDK would be initialized too and would still subscribe for SystemEvents and NetworkCallbacks while in background.
```

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/AndroidConnectionStatusProvider.java
```diff
@@ -730,15 +792,31 @@ public NetworkCallback getNetworkCallback() {
...
+    try {
+      options.getExecutorService().submit(r);
+    } catch (Throwable e) {
+      options
+          .getLogger()
+          .log(SentryLevel.ERROR, "AndroidConnectionStatusProvider submit failed", e);
+    }
+  }
+
+  private static class ConnectionStatusProviderThreadyFactory implements ThreadFactory {
```
markushi:
```suggestion
  private static class ConnectionStatusProviderThreadFactory implements ThreadFactory {
```

sentry-samples/sentry-samples-android/sdkperf/README.md
markushi, line 1: Awesome!

sentry-samples/sentry-samples-android/sdkperf/screen_flap.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+# Loop 4 times to toggle Wi-Fi
```
markushi:
```suggestion
# Loop 4 times to toggle Screen
```

sentry-samples/sentry-samples-android/sdkperf/wifi_flap.sh
```diff
@@ -0,0 +1,19 @@
...
+do
+  echo "[$i] Disabling Wi-Fi..."
+  adb shell svc wifi disable
+  sleep 2
+
+  echo "[$i] Enabling Wi-Fi..."
+  adb shell svc wifi enable
+  sleep 6
+done
+
```
markushi:
```suggestion
# Turn mobile data back on
adb shell svc data enable
```

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/AndroidConnectionStatusProvider.java
```diff
@@ -413,42 +428,90 @@ public void removeConnectionStatusObserver(final @NotNull IConnectionStatusObser
...
+        () -> {
+          // proactively update cache and notify observers on foreground to ensure connectivity
+          // state is not stale
+          updateCache(null);
+
+          final @NotNull ConnectionStatus status = getConnectionStatusFromCache();
+          if (status == ConnectionStatus.DISCONNECTED) {
+            // onLost is not called retroactively when we registerNetworkCallback (as opposed to
+            // onAvailable), so we have to do it manually for the DISCONNECTED case
+            isConnected.set(false);
```
markushi, line 486: Shouldn't `isConnected` be set to true if the status is connected?

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/AndroidConnectionStatusProvider.java
```diff
@@ -413,42 +428,90 @@ public void removeConnectionStatusObserver(final @NotNull IConnectionStatusObser
...
+            try (final @NotNull ISentryLifecycleToken ignored = childCallbacksLock.acquire()) {
+              for (final @NotNull NetworkCallback cb : childCallbacks) {
+                //noinspection DataFlowIssue
+                cb.onLost(null);
```
markushi, line 490: Wouldn't something like this make more sense here:
```suggestion
                if (connected) { cb.onConnected(null); } else { cb.onLost(null); }
```

markushi (commented):
Left a few comments for clarification, but already looks really good - nice work!

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/AndroidConnectionStatusProvider.java
romtsn: yeah so that's the stupid part of NetworkCallbacks - when you call `connectivityManager.registerNetworkCallback` it will automatically fire `onAvailable` + `onCapabilitiesChanged` for that callback, BUT only if you're currently connected. For the disconnected case it won't fire `onLost`/`onUnavailable`, so we have to manually retrofit it here.

Same applies to the comment below. Basically the `onConnected` case is handled for us by the system and all of the respective flags/callbacks will be set/notified in `onAvailable`/`onCapabilitiesChanged`

markushi (approved):

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/AndroidConnectionStatusProvider.java
romtsn: Forgot to remove it - it's actually not needed anymore

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/AndroidConnectionStatusProvider.java
```diff
@@ -413,42 +428,90 @@ public void removeConnectionStatusObserver(final @NotNull IConnectionStatusObser
...
+      if (callbackRef != null) {
+        unregisterNetworkCallback(context, options.getLogger(), callbackRef);
+      }
+      // Clear cached state
+      cachedNetworkCapabilities = null;
+      currentNetwork = null;
+      lastCacheUpdateTime = 0;
+    }
+    options.getLogger().log(SentryLevel.DEBUG, "Network callback unregistered");
+  }
```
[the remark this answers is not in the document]
romtsn: that's intentional, because registerNetworkCallback will automatically fire onAvailable and onCapabilitiesChanged and we don't want to store multiple network_available breadcrumbs multiple times if it actually hasn't changed

Issue #4577, opened 2025-07-30
[Crash] java.lang.ArrayIndexOutOfBoundsException at io.sentry.android.core.internal.util.SentryFrameMetricsCollector$WindowFrameMetricsManager.addOnFrameMetricsAvailableListener


kibotu:
### Integration

sentry-android

### Build System

Gradle

### AGP Version

8.11.0

### Proguard

Enabled

### Version

8.16.0

### Steps to Reproduce

io.sentry:sentry-android-gradle-plugin:5.8.0

[Samsung Galaxy A53 5G UW]
[Android 15 (SDK 35)]

```
Exception java.lang.ArrayIndexOutOfBoundsException: length=10; index=-1
  at java.util.ArrayList.add (ArrayList.java:491)
  at java.util.ArrayList.add (ArrayList.java:503)
  at android.view.View.addFrameMetricsListener (View.java:8333)
  at android.view.Window.addOnFrameMetricsAvailableListener (Window.java:1007)
  at io.sentry.android.core.internal.util.SentryFrameMetricsCollector$WindowFrameMetricsManager.addOnFrameMetricsAvailableListener (SentryFrameMetricsCollector.java:376)
  at io.sentry.android.core.internal.util.SentryFrameMetricsCollector.trackCurrentWindow (SentryFrameMetricsCollector.java:317)
  at io.sentry.android.core.internal.util.SentryFrameMetricsCollector.startCollection (SentryFrameMetricsCollector.java:266)
  at io.sentry.android.core.SpanFrameMetricsCollector.onSpanStarted (SpanFrameMetricsCollector.java:96)
  at io.sentry.DefaultCompositePerformanceCollector.start (DefaultCompositePerformanceCollector.java:65)
  at io.sentry.SentryTracer.<init> (SentryTracer.java:95)
  at io.sentry.DefaultSpanFactory.createTransaction (DefaultSpanFactory.java:15)
  at io.sentry.Scopes.createTransaction (Scopes.java:942)
  at io.sentry.Scopes.startTransaction (Scopes.java:886)
  at io.sentry.IScopes.startTransaction (IScopes.java:616)
  at io.sentry.IScopes.startTransaction (IScopes.java:599)
  at io.sentry.Sentry.startTransaction (Sentry.java:1113)
  at de.check24.profis.partner.core.services.network.interceptor.SentryTransactionInterceptor$intercept$1.invokeSuspend (SentryTransactionInterceptor.kt:21)
[...]

PR #4582, opened 2025-07-31
fix(android): Ensure frame metrics listeners are registered/unregistered on the main thread

markushi:
## :scroll: Description

Fixes https://github.com/getsentry/sentry-java/issues/4577

The Android View system expects add/remove calls to happen on the main thread.

Added unit tests

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Deduplicate battery breadcrumbs ([#4561](https://github.com/getsentry/sentry-java/pull/4561))
 - Remove unused method in ManifestMetadataReader ([#4585](https://github.com/getsentry/sentry-java/pull/4585))
 - Have single `NetworkCallback` registered at a time to reduce IPC calls ([#4562](https://github.com/getsentry/sentry-java/pull/4562))
+- Do not register for SystemEvents and NetworkCallbacks immediately when launched with non-foreground importance ([#4579](https://github.com/getsentry/sentry-java/pull/4579))
 - Limit ProGuard keep rules for native methods within `sentry-android-ndk` to the `io.sentry.**` namespace. ([#4427](https://github.com/getsentry/sentry-java/pull/4427))
   - If you relied on the Sentry SDK to keep native method names for JNI compatibility within your namespace, please review your ProGuard rules and ensure the configuration still works. Especially when you're not consuming any of the default Android proguard rules (`proguard-android.txt` or `proguard-android-optimize.txt`) the following config should be present:
   ```
--- a/sentry-android-core/api/sentry-android-core.api
+++ b/sentry-android-core/api/sentry-android-core.api
@@ -167,14 +167,27 @@ public final class io/sentry/android/core/AppLifecycleIntegration : io/sentry/In
 }
 
 public final class io/sentry/android/core/AppState : java/io/Closeable {
+	public fun addAppStateListener (Lio/sentry/android/core/AppState$AppStateListener;)V
 	public fun close ()V
 	public static fun getInstance ()Lio/sentry/android/core/AppState;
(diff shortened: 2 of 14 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
```diff
@@ -282,17 +286,24 @@ public void stopCollection(final @Nullable String listenerId) {
...
-    }
+    new Handler(Looper.getMainLooper())
+        .post(
+            () -> {
+              try {
+                // Re-check if we should still remove the listener for this window
+                // in case trackCurrentWindow was called in the meantime
+                final boolean shouldRemove;
+                try (final @NotNull ISentryLifecycleToken ignored = trackedWindowsLock.acquire()) {
+                  shouldRemove = trackedWindows.contains(window) && trackedWindows.remove(window);
```
romtsn: I think should be enough to just call `remove` because it will return false if the set did not contain the element?
image
```suggestion
                  shouldRemove = trackedWindows.remove(window);
```

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
```diff
@@ -373,6 +401,9 @@ default void addOnFrameMetricsAvailableListener(
         final @NotNull Window window,
         final @Nullable Window.OnFrameMetricsAvailableListener frameMetricsAvailableListener,
         final @Nullable Handler handler) {
+      if (frameMetricsAvailableListener == null) {
```
romtsn, line 391: probably also worth to add this check to `removeOnFrameMetricsAvailableListener`?

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
```diff
@@ -305,18 +316,35 @@ private void setCurrentWindow(final @NotNull Window window) {
...
+    if (handler != null) {
+      // Ensure the addOnFrameMetricsAvailableListener is called on the main thread
+      new Handler(Looper.getMainLooper())
+          .post(
+              () -> {
+                // Re-check if we should still track this window
+                // in case stopTrackingWindow was called for the same Window in the meantime
+                final boolean shouldAdd;
+                try (final @NotNull ISentryLifecycleToken ignored = trackedWindowsLock.acquire()) {
+                  shouldAdd = !trackedWindows.contains(window) && trackedWindows.add(window);
```
romtsn: same here I guess, could omit `contains` since it's a set?

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
```diff
@@ -282,17 +286,24 @@ public void stopCollection(final @Nullable String listenerId) {
...
-      trackedWindows.remove(window);
-    }
+    new Handler(Looper.getMainLooper())
+        .post(
+            () -> {
+              try {
+                // Re-check if we should still remove the listener for this window
+                // in case trackCurrentWindow was called in the meantime
+                final boolean shouldRemove;
+                try (final @NotNull ISentryLifecycleToken ignored = trackedWindowsLock.acquire()) {
```
romtsn: do we actually need this lock considering we only add/remove windows on the main thread now?

markushi: Oh true, that's some leftover, `trackedWindows` is only accessed within the main thread - thanks!

PR #4604, opened 2025-08-04
Improve SentryTraceHeader constructor parameter validation

alexander-alderman-webb:
## :scroll: Description

Use similar, regex-based, validation logic to the PHP SDK for the sentry-trace HTTP header: https://github.com/getsentry/sentry-php/blob/master/src/Tracing/TransactionContext.php#L9

#skip-changelog

- Alert the user early with an exception when a malformed header is received.
- https://github.com/getsentry/sentry-java/issues/2799

Added test cases with malformed arguments, which would have caused errors later on previously. These include tests with IDs that are too short or too long, and IDs with invalid characters.

## :pencil: Checklist

### DIFF
--- a/sentry/src/main/java/io/sentry/SentryTraceHeader.java
+++ b/sentry/src/main/java/io/sentry/SentryTraceHeader.java
@@ -2,6 +2,8 @@ package io.sentry;
 
 import io.sentry.exception.InvalidSentryTraceHeaderException;
 import io.sentry.protocol.SentryId;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
@@ -13,6 +15,11 @@ public final class SentryTraceHeader {
   private final @NotNull SpanId spanId;
   private final @Nullable Boolean sampled;
 
+  final Pattern SENTRY_TRACEPARENT_HEADER_REGEX =
+      Pattern.compile(
+          "^[ \\t]*(?<traceId>[0-9a-f]{32})-(?<spanId>[0-9a-f]{16})-?(?<sampled>[01])?[ \\t]*$",
+          Pattern.CASE_INSENSITIVE);
+
   public SentryTraceHeader(
       final @NotNull SentryId traceId,
       final @NotNull SpanId spanId,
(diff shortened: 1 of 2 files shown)


sentry/src/main/java/io/sentry/SentryTraceHeader.java
```diff
@@ -13,6 +15,11 @@ public final class SentryTraceHeader {
   private final @NotNull SpanId spanId;
   private final @Nullable Boolean sampled;
 
+  final Pattern SENTRY_TRACEPARENT_HEADER_REGEX =
+      Pattern.compile(
+          "^[ \\t]*(?<traceId>[0-9a-f]{32})-(?<spanId>[0-9a-f]{16})(?<sampled>-[01])?[ \\t]*$",
+          Pattern.CASE_INSENSITIVE);
```
lcian:
```suggestion
  private static final Pattern SENTRY_TRACEPARENT_HEADER_REGEX =
      Pattern.compile(
          "^[ \\t]*(?<traceId>[0-9a-f]{32})-(?<spanId>[0-9a-f]{16})(?<sampled>-[01])?[ \\t]*$",
          Pattern.CASE_INSENSITIVE);
```

We can make it `static` so we're only going to compile it once, as compiling could be expensive. Otherwise, without `static` we would potentially compile the regex for each incoming request.

sentry/src/main/java/io/sentry/SentryTraceHeader.java
```diff
@@ -23,20 +30,17 @@ public SentryTraceHeader(
...
-      this.traceId = new SentryId(parts[0]);
-      this.spanId = new SpanId(parts[1]);
-    } catch (Throwable e) {
-      throw new InvalidSentryTraceHeaderException(value, e);
     }
+
+    this.traceId = new SentryId(matcher.group("traceId"));
+    this.spanId = new SpanId(matcher.group("spanId"));
+    this.sampled =
+        matcher.group("sampled") == null ? null : "1".equals(matcher.group("sampled").substring(1));
```
lcian, line 43:
```suggestion
	if (matcher.group("sampled") != null) {
		this.sampled = "1".equals(matcher.group("sampled").substring(1));
	}
```

I think this would be slightly more readable

alexander-alderman-webb: The field must be initialized because the field is final. I believe I tried the simple if condition initially, but to keep the field final we must explicitly assign null.

sentry/src/main/java/io/sentry/SentryTraceHeader.java
alexander-alderman-webb: Yes great point. Addressed this.

PR #4606, opened 2025-08-05
perf(executor): Prewarm SentryExecutorService

romtsn:
## :scroll: Description

- Prewarm SentryExecutorService so we don't spend time for growing the internal task queue when it's full
- Make SentryExecutorService bounded to 271 task (x8 times growth from the initial capacity), drop everything that's on top
- Use LazyEvaluator for a couple of things I noticed were slowing down the init (e.g. instantiating a timer thread in LifecycleWatcher, when we actually only need it when finalizing a session)
- Use HandlerThread for getting receiving broadcasts off-the-main-thread

Manually + automated

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
   }
   ```
 - Fix abstract method error in `SentrySupportSQLiteDatabase` ([#4597](https://github.com/getsentry/sentry-java/pull/4597))
+- Ensure frame metrics listeners are registered/unregistered on the main thread ([#4582](https://github.com/getsentry/sentry-java/pull/4582))
 
 ## 8.18.0
 
--- a/sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
@@ -38,6 +38,7 @@ public final class SentryFrameMetricsCollector implements Application.ActivityLi
 
   private final @NotNull BuildInfoProvider buildInfoProvider;
   private final @NotNull Set<Window> trackedWindows = new CopyOnWriteArraySet<>();
+
   private final @NotNull ILogger logger;
   private @Nullable Handler handler;
   private @Nullable WeakReference<Window> currentWindow;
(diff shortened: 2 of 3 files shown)


sentry/src/main/java/io/sentry/SentryExecutorService.java
```diff
@@ -65,6 +118,21 @@ public boolean isClosed() {
...
+        () -> {
+          // schedule a bunch of dummy runnables in the future that will never execute to trigger
+          // queue
+          // growth and then clear the queue up
+          for (int i = 0; i < INITIAL_QUEUE_SIZE; i++) {
+            executorService.schedule(dummyRunnable, Long.MAX_VALUE, TimeUnit.DAYS);
+          }
+          executorService.getQueue().clear();
+        });
+  }
```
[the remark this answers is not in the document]
romtsn: could be the case but we must call prewarm directly after instantiating the executor service. I will add a comment though 👍

sentry/src/main/java/io/sentry/SentryExecutorService.java
```diff
@@ -2,43 +2,96 @@
...
+    this(new ScheduledThreadPoolExecutor(1, new SentryExecutorServiceThreadFactory()), null);
   }
 
   @Override
   public @NotNull Future<?> submit(final @NotNull Runnable runnable) {
-    return executorService.submit(runnable);
+    if (executorService.getQueue().size() < MAX_QUEUE_SIZE) {
+      return executorService.submit(runnable);
+    }
+    // TODO: maybe RejectedExecutionException?
```
markushi, line 61: Yes, I'd say so, when we did the SDK restart work we should have accounted for `RejectedExecutionException` everywhere.

sentry/src/main/java/io/sentry/SentryExecutorService.java
```diff
@@ -65,6 +118,24 @@ public boolean isClosed() {
...
+  @SuppressWarnings({"FutureReturnValueIgnored"})
+  @Override
+  public void prewarm() {
+    executorService.submit(
+        () -> {
+          // schedule a bunch of dummy runnables in the future that will never execute to trigger
+          // queue
+          // growth and then clear the queue up
+          for (int i = 0; i < INITIAL_QUEUE_SIZE; i++) {
+            executorService.schedule(dummyRunnable, Long.MAX_VALUE, TimeUnit.DAYS);
```
markushi: Would something like this work? Scheduling and then canceling the future?
```suggestion
            try {
                final ScheduledFuture future = executorService.schedule(dummyRunnable, Long.MAX_VALUE, TimeUnit.DAYS);
                future.cancel(true);
            } catch (RejectedExecutionException ignore) {
            }
```

sentry/src/main/java/io/sentry/SentryExecutorService.java
```diff
@@ -2,43 +2,96 @@
...
   private final @NotNull AutoClosableReentrantLock lock = new AutoClosableReentrantLock();
 
+  @SuppressWarnings("UnnecessaryLambda")
+  private final @NotNull Runnable dummyRunnable = () -> {};
+
+  private final @Nullable SentryOptions options;
+
   @TestOnly
-  SentryExecutorService(final @NotNull ScheduledExecutorService executorService) {
+  SentryExecutorService(
```
markushi, line 41: The core SDK, logging, spotlight and profiling all use their own instance of  `SentryExecutorService`. I guess it makes sense to pass the queue size as a constructor parameter instead and only prewarm the core SDK one?

markushi: Oh I see, you only call pre-warm on the core one anyway.

markushi (approved):
Looking good, nice work! I left a few small-ish comments.

sentry/src/main/java/io/sentry/SentryExecutorService.java
romtsn: yeah, I think other usages are fine and shouldn't have many tasks scheduled

sentry/src/main/java/io/sentry/SentryExecutorService.java
romtsn: I've seen some instances where we don't handle it (e.g. [here](https://github.com/getsentry/sentry-java/blob/bc1e990dfb6cf7e9baa300f95d7d50a2d360d78f/sentry/src/main/java/io/sentry/Scopes.java#L452)), so I'd rather do that in a separate PR later with fixing RejectedExecutionException along the way

PR #4624, opened 2025-08-08
fix(sessions): Move and flush unfinished previous session on init

romtsn:
## :scroll: Description

- Previously, we were waiting on previous session flush latch on our background queue when processing ANRs or finalizing previous session for 15 seconds to give it a good chance to be flushed. However, the trigger for the flush was actually a new session start:
  - We start a new session on foreground
  - We see that there's an existing unfinished session
  - We move it to its own file
  - We flush the latch
However this was not working well for background app starts. Since there's no foreground event, a new session would not be started and we were waiting for 15 seconds for nothing (on a background thread, but still). So this PR addresses it by moving any unfinished session to its own file right away on `init` (but before any of the integrations are registered, so those that rely on the previous session latch (like ANRs), do not need to wait for it).

- We also just rename `session.json` to `previous_session.json` now, as opposed to ser/deserializing the session multiple times from/into different files

### Before (perfetto trace of a background app start)
Google Chrome 2025-08-08 16 14 48

### After (perfetto trace of a background app start)

Screenshot 2025-08-08 at 11 14 24

Speed up things on our background queue

Manually + automated

## :pencil: Checklist

### DIFF
--- a/sentry/src/main/java/io/sentry/cache/EnvelopeCache.java
+++ b/sentry/src/main/java/io/sentry/cache/EnvelopeCache.java
@@ -73,6 +73,7 @@ public class EnvelopeCache extends CacheStrategy implements IEnvelopeCache {
 
   private final @NotNull Map<SentryEnvelope, String> fileNameMap = new WeakHashMap<>();
   protected final @NotNull AutoClosableReentrantLock cacheLock = new AutoClosableReentrantLock();
+  protected final @NotNull AutoClosableReentrantLock sessionLock = new AutoClosableReentrantLock();
 
   public static @NotNull IEnvelopeCache create(final @NotNull SentryOptions options) {
     final String cacheDirPath = options.getCacheDirPath();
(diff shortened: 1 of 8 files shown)


romtsn:
@markushi I'm unsure about how this affects the Hybrid SDKs, but form a quick look into InternalSentrySdk I don't think anything needs to change. But let me know if there's something I missed.

sentry/src/main/java/io/sentry/cache/EnvelopeCache.java
```diff
@@ -454,4 +437,29 @@ public boolean waitPreviousSessionFlush() {
   public void flushPreviousSession() {
     previousSessionLatch.countDown();
   }
+
+  public void movePreviousSession(
+      final @NotNull File currentSessionFile, final @NotNull File previousSessionFile) {
+    try (final @NotNull ISentryLifecycleToken ignored = sessionLock.acquire()) {
+      if (previousSessionFile.exists()) {
```
markushi, line 444: I'm wondering if we should simply overwrite the file instead. If the file still exists, chances are that we've already sent that session information. On the other hand `currentSessionFile` was never flushed, so renaming it to `previousSessionFile` and later flushing that session feels more correct.

markushi (approved):
LGTM, left a comment about edge case handling.

sentry/src/main/java/io/sentry/cache/EnvelopeCache.java
romtsn: yeah I've been torn between this approach and what you suggest. Basically, if the `previous_session` file still exists it means the previous process exited unexpectedly before we managed to finish the previous-previous session and flush it (due to a crash or any other process death). In that case that session still hasn't been properly finalized so we're anyway losing out on it. So the question is which session do we want to lose 😅 

I guess to properly fix it we need some better storage of sessions (like a ring buffer, not just previous + current), but that's probably for later. I will change it to your suggestion though 👍

sentry/src/main/java/io/sentry/cache/EnvelopeCache.java
```diff
@@ -454,4 +424,29 @@ public boolean waitPreviousSessionFlush() {
...
+        options.getLogger().log(DEBUG, "Previous session file already exists.");
+        return;
+      }
+
+      if (currentSessionFile.exists()) {
+        options.getLogger().log(INFO, "Moving current session to previous session.");
+
+        try {
+          final boolean renamed = currentSessionFile.renameTo(previousSessionFile);
+          if (!renamed) {
```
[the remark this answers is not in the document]
romtsn: Valid concern, but we're not going to retry (only on next SDK init), so we'd rather unblock right away

PR #4666, opened 2025-08-26
Disable Gradle configuration cache for enforce-license-compliance workflow

lcian:
## :pencil: Checklist


lcian:
> nice job! so can we use this action to detect licence violations in source files?

Apparently this works only for Gradle dependencies.
For vendored dependencies, you would have to create a `.fossa.yml` to point the tool to them.
Even then (I've tried), it seems those are not detected reliably.
We can probably use https://github.com/getsentry/sentry-java/pull/4680 once it's ready, it's a simple grep that should work for vendored deps.

PR #4684, opened 2025-08-29
Flush logs on crash

stefanosiano:
## :scroll: Description
When a crash occurs, logs are flushed

We ensure to store logs to disk (or even send them) instead of losing them due to being only in memory when a crash occurs

## :pencil: Checklist

### DIFF
--- a/sentry/src/main/java/io/sentry/SentryClient.java
+++ b/sentry/src/main/java/io/sentry/SentryClient.java
@@ -245,6 +245,12 @@ public final class SentryClient implements ISentryClient {
       finalizeTransaction(scope, hint);
     }
 
+    // if event is backfillable or cached we don't need to flush the logs, because it's an event
+    // from the past. Otherwise we need to flush the logs to ensure they are sent on crash
+    if (event != null && !isBackfillable && !isCached && event.isCrashed()) {
+      loggerBatchProcessor.flush(options.getFlushTimeoutMillis());
+    }
+
     return sentryId;
   }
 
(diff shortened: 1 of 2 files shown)


sentry/src/main/java/io/sentry/SentryClient.java
```diff
@@ -245,6 +245,12 @@ private boolean shouldApplyScopeData(final @NotNull CheckIn event, final @NotNul
       finalizeTransaction(scope, hint);
     }
 
+    // if event is backfillable or cached we don't need to flush the logs, because it's an event
+    // from the past. Otherwise we need to flush the logs to ensure they are sent on crash
+    if (event != null && !isBackfillable && !isCached && event.isCrashed()) {
+      loggerBatchProcessor.flush(options.getFlushTimeoutMillis());
```
stefanosiano: I'm not sure on this. 
In theory we could wait for `flushTimeoutMillis` twice, once here and once to flush the crash.
In practice I expect this method to be instantaneous in Android, as it doesn't wait for them to be sent, and there shouldn't be many logs on Android anyway

markushi: IMHO crashes have the highest priority, so I would avoid any extra work/wait which could cause them to be not sent. => Just flush the logs to disk with a minimal timeout.

stefanosiano: technically, the event is captured first, so it will be sent before the logs anyway
but yeah, let's put a minimal timeout

stefanosiano: I moved the capture of replay and logs after the crash event one.
But now we are waiting only for the crash to be flushed to disk, with the risk of losing replay and logs (except for those 500ms on logs)
Is it fine this way?
@romtsn wdyt?

romtsn:
Closing as this will be tackled as part of a larger initiative properly https://github.com/getsentry/rfcs/pull/148

PR #4686, opened 2025-09-02
feat: Frame vars

limbonaut:
## :scroll: Description

This PR introduces the `vars` attribute to the `SentryFrame` class. While it existed previously, it wasn’t serialized and had a String->String mapping type.

These changes are necessary for the [Godot SDK](https://github.com/getsentry/sentry-godot) to capture script variables on Android. For GDScript errors, we construct a stack trace and attach variables for each frame. There is a similar setup with Sentry Native and Cocoa backend SDKs.

I've added serialization tests.

## :pencil: Checklist

### DIFF
--- a/sentry/api/sentry.api
+++ b/sentry/api/sentry.api
@@ -5933,6 +5933,7 @@ public final class io/sentry/protocol/SentryStackFrame$JsonKeys {
 	public static final field RAW_FUNCTION Ljava/lang/String;
 	public static final field SYMBOL Ljava/lang/String;
 	public static final field SYMBOL_ADDR Ljava/lang/String;
+	public static final field VARS Ljava/lang/String;
 	public fun <init> ()V
 }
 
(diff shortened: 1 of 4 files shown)


limbonaut:
Technically, it would be a breaking change if it were in use and working as expected. Seeing that it's actually not serialized, I strongly suspect that it's not being used anywhere, or otherwise you'd have it at least reported. Not sure it's worth the trouble engineering a workaround for API contract that didn't work in the first place.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
   - NOTE: Our `sentry-opentelemetry-agentless-spring` is not working yet for Spring Boot 4. Please use `sentry-opentelemetry-agent` until OpenTelemetry has support for Spring Boot 4.
 - Replace `UUIDGenerator` implementation with Apache licensed code ([#4662](https://github.com/getsentry/sentry-java/pull/4662))
 - Replace `Random` implementation with MIT licensed code ([#4664](https://github.com/getsentry/sentry-java/pull/4664))
+- Add support for `vars` attribute in `SentryStackFrame` ([#4686](https://github.com/getsentry/sentry-java/pull/4686))
```
romtsn, line 15:
```suggestion
- Add support for `vars` attribute in `SentryStackFrame` ([#4686](https://github.com/getsentry/sentry-java/pull/4686))
  - **Breaking change**: The type of the `vars` attribute has been changed from `Map<String, String>` to `Map<String, Object>`.
```

romtsn (approved):
LGTM, let's just add a note to the changelog, but the impact should be minimal-to-none, so we should be good here.

PR #4699, opened 2025-09-05
Add build distribution module.

runningcode:
## :scroll: Description
This is the first step towards the build distribution library.

I named it `sentry-build-distribution` not
`sentry-android-build-distribution` because i thought the name was too
long. This will only ever be an `android` based sdk so feel free to push
back on the naming.

#skip-changelog We're not releasing anything yet

## :pencil: Checklist


settings.gradle.kts
```diff
@@ -21,6 +21,7 @@ includeBuild("build-logic")
 include(
     "sentry",
     "sentry-kotlin-extensions",
+    "sentry-build-distribution",
```
romtsn: I would love to name it `sentry-android-distribution` can we do that? Historically, we add `-android` to everything android-related. And imho it's fine to omit the `-build` part here as to me `distribution` already implies builds/artifacts/apks/etc.

runningcode: Done!

romtsn:
oh btw, we may need to exclude it from publications in the root project, wdyt? https://github.com/getsentry/sentry-java/blob/5af7d2337209e7219776e012b1cce40b8e8c9efb/build.gradle.kts#L164

runningcode:
We will eventually need to publish it, but do you mean that we should exclude it in case we have another version published before it is ready?

romtsn:
yes, I mainly meant to avoid publishing an empty module for now (we'll probably ship a new version next week already)

Issue #4651, opened 2025-08-22
8.19+ lost compatibility with API older than 26?


mauriciospinardi-cloudwalk:
### Integration

sentry-android

### Build System

Gradle

### AGP Version

8.12.1

### Proguard

Enabled

### Version

8.19.1

### Steps to Reproduce

Just noticed a crash on a device running Android 5.1 (API 22). Upon further analysis, it seems we have an addition to `SentryTraceHeader` that could be the cause. The crash does not seem to happen in older versions:

Possible culprit:
```java
private static final Pattern SENTRY_TRACEPARENT_HEADER_REGEX =
      Pattern.compile(
          "^[ \\t]*(?<traceId>[0-9a-f]{32})-(?<spanId>[0-9a-f]{16})(?<sampled>-[01])?[ \\t]*$",
          Pattern.CASE_INSENSITIVE);
```

(AI)  `The regex pattern uses named capturing groups ((?<traceId>, (?<spanId>, (?<sampled>), which were only added to Android's Pattern class in API 26 (Android 8.0)`.

Was that intentional?

### Expected Result

Release 8.19.1 should work just fine on devices with API 22.

### Actual Result

Release 8.19.1 crashes the application when running on devices with API 22.

```log
Process: io.cloudwalk.appclientcw, PID: 4128
java.lang.ExceptionInInitializerError
	at kotlin.random.RandomKt.traceIfAllowed(r8-map-id-af09a800d99bce4296167a985f9f674ea32b4cb61321121f585812368ad730f8:83)
	at io.sentry.okhttp.SentryOkHttpInterceptor.intercept(r8-map-id-af09a800d99bce4296167a985f9f674ea32b4cb61321121f585812368ad730f8:175)
	at okhttp3.internal.http.RealInterceptorChain.proceed(r8-map-id-af09a800d99bce4296167a985f9f674ea32b4cb61321121f585812368ad730f8:128)
	at io.cloudwalk.core.client.communication.retrofit.JsonInterceptor.intercept(r8-map-id-af09a800d99bce4296167a985f9f674ea32b4cb61321121f585812368ad730f8:45)
	at okhttp3.internal.http.RealInterceptorChain.proceed(r8-map-id-af09a800d99bce4296167a985f9f674ea32b4cb61321121f585812368ad730f8:128)
[...]

PR #4705, opened 2025-09-08
build: Add animalsniffer to check that source is compatible with Android API

alexander-alderman-webb:
## :scroll: Description

#skip-changelog

Add `animalsniffer` to check the source files in the `main` directories of plain Java repositories. The signature in `coreLib2` allows classes and methods and JDK APIs available through desugaring. Using a less permissive signature that does not allow desugaring causes failures, so I opted for `coreLib2` for now.

Prevent issues like https://github.com/getsentry/sentry-java/issues/4651 reaching users.

The `animalsnifferMain` task would have failed the responsible PR: https://github.com/getsentry/sentry-java/pull/4604.

Run `./gradlew animalsnifferMain`.

## :pencil: Checklist

### DIFF
--- a/sentry-spring-boot/build.gradle.kts
+++ b/sentry-spring-boot/build.gradle.kts
@@ -11,6 +11,7 @@ plugins {
   alias(libs.plugins.gradle.versions)
   alias(libs.plugins.buildconfig)
   alias(libs.plugins.springboot2) apply false
+  alias(libs.plugins.animalsniffer)
 }
 
 tasks.withType<KotlinCompile>().configureEach {
@@ -72,6 +73,9 @@ dependencies {
   testImplementation(projects.sentryOpentelemetry.sentryOpentelemetryAgent)
   testImplementation(projects.sentryOpentelemetry.sentryOpentelemetryAgentcustomization)
   testImplementation(projects.sentryOpentelemetry.sentryOpentelemetryBootstrap)
+
+  val gummyBearsModule = libs.gummy.bears.api21.get().module
+  signature("${gummyBearsModule}:${libs.versions.gummyBears.get()}:coreLib2@signature")
 }
 
 configure<SourceSetContainer> { test { java.srcDir("src/test/java") } }
@@ -92,6 +96,7 @@ tasks {
   check {
     dependsOn(jacocoTestCoverageVerification)
     dependsOn(jacocoTestReport)
+    dependsOn(animalsnifferMain)
   }
 }
 
--- a/sentry/build.gradle.kts
+++ b/sentry/build.gradle.kts
@@ -9,6 +9,7 @@ plugins {
   alias(libs.plugins.errorprone)
   alias(libs.plugins.gradle.versions)
   alias(libs.plugins.buildconfig)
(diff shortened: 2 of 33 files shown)


sentry/build.gradle.kts
```diff
@@ -31,6 +32,9 @@ dependencies {
   testImplementation(libs.msgpack)
   testImplementation(libs.okio)
   testImplementation(projects.sentryTestSupport)
+
+  val gummyBearsModule = libs.gummy.bears.api21.get().module
+  signature("${gummyBearsModule}:${libs.versions.gummyBears.get()}:coreLib2@signature")
```
romtsn: unfortunately, we cannot guarantee that core-library desugaring is enabled by our customers, so we have to be conservative here and test against the plain Android APIs assuming there's no desugaring and our SDK should still work.

romtsn: what are the failures you're experiencing? If it fails, we should actually fix those

sentry-spring-boot/build.gradle.kts
romtsn, line 1: imo it's perfectly fine to use newer Java features in non-Android-related modules, so I'd only apply this plugin to what can be used on Android:
- sentry-apollo-3
- sentry-apollo-4
- sentry-apollo
- sentry-kotlin-extensions
- sentry-ktor-client
- sentry-okhttp
- sentry

sentry/build.gradle.kts
alexander-alderman-webb: Only one task fails when I use the list of modules you gave below. Do these need to be fixed?

```
> Task :sentry:animalsnifferMain FAILED

[Undefined reference] io.sentry.(SentryInstantDate.java:1) #date
  >> java.time.Instant

[Undefined reference] io.sentry.(SentryInstantDate.java:16)
  >> java.time.Instant java.time.Instant.now()

[Undefined reference] io.sentry.(SentryInstantDate.java:26)
  >> long java.time.Instant.getEpochSecond()

[Undefined reference] io.sentry.(SentryInstantDate.java:26)
  >> int java.time.Instant.getNano()

[Undefined reference] io.sentry.(SentryWrapper.java:54)
  >> java.util.function.Supplier

[Undefined reference] io.sentry.(SentryWrapper.java:56)
  >> Object java.util.function.Supplier.get()
```

sentry-spring-boot/build.gradle.kts
alexander-alderman-webb: Yes, thanks for the suggestion

sentry/build.gradle.kts
romtsn: is there a way to suppress these? For `Instant` we actually have a check on Android if it's available (API 26+) and then use it, for `SentryWrapper` - it's not used unless the customers use it, which won't be the case for Android (Virtual Threads is not available I believe).

alexander-alderman-webb: Thanks for the context!

I added config to ignore all `java.time.Instant` warnings and exclude `SentryWrapper.java` from the check.

romtsn (approved):
LGTM, very impactful change, thanks!

PR #4712, opened 2025-09-10
feat(android-distribution): Add module foundation with compilation stubs

runningcode:
## Summary

Adds Sentry Android Distribution module with automatic integration and `Sentry.distribution()` API.

## Key Changes
- **Automatic integration**: Module auto-enables when included (like replay/timber)
- **Top-level API**: Added `Sentry.distribution()` method  
- **Build-time safety**: Compiles whether module included or not
- **Clean architecture**: No circular dependencies, ProGuard ready

## Usage
```kotlin
// Automatic when module included
val distribution = Sentry.distribution() as? Distribution
distribution?.checkForUpdate(context) { result -> /* handle */ }
```

#skip-changelog

### DIFF
--- a/sentry-android-distribution/api/sentry-android-distribution.api
+++ b/sentry-android-distribution/api/sentry-android-distribution.api
@@ -1,40 +1,25 @@
 public final class io/sentry/android/distribution/Distribution {
 	public static final field INSTANCE Lio/sentry/android/distribution/Distribution;
-	public final fun checkForUpdate (Landroid/content/Context;)Lio/sentry/android/distribution/UpdateStatus;
-	public final fun checkForUpdateCompletableFuture (Landroid/content/Context;)Ljava/util/concurrent/CompletableFuture;
+	public final fun checkForUpdate (Landroid/content/Context;Lkotlin/jvm/functions/Function1;)V
+	public final fun checkForUpdateBlocking (Landroid/content/Context;)Lio/sentry/android/distribution/UpdateStatus;
 	public final fun downloadUpdate (Landroid/content/Context;Lio/sentry/android/distribution/UpdateInfo;)V
-	public final fun init (Landroid/content/Context;Lio/sentry/android/distribution/DistributionOptions;)V
+	public final fun init (Landroid/content/Context;)V
+	public final fun init (Landroid/content/Context;Lkotlin/jvm/functions/Function1;)V
 	public final fun isEnabled ()Z
 }
 
-public final class io/sentry/android/distribution/DistributionContentProvider : android/content/ContentProvider {
-	public fun <init> ()V
-	public fun delete (Landroid/net/Uri;Ljava/lang/String;[Ljava/lang/String;)I
-	public fun getType (Landroid/net/Uri;)Ljava/lang/String;
-	public fun insert (Landroid/net/Uri;Landroid/content/ContentValues;)Landroid/net/Uri;
-	public fun onCreate ()Z
-	public fun query (Landroid/net/Uri;[Ljava/lang/String;Ljava/lang/String;[Ljava/lang/String;Ljava/lang/String;)Landroid/database/Cursor;
-	public fun update (Landroid/net/Uri;Landroid/content/ContentValues;Ljava/lang/String;[Ljava/lang/String;)I
-}
-
 public final class io/sentry/android/distribution/DistributionOptions {
-	public fun <init> (Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;)V
-	public synthetic fun <init> (Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
-	public final fun component1 ()Ljava/lang/String;
-	public final fun component2 ()Ljava/lang/String;
-	public final fun component3 ()Ljava/lang/String;
-	public final fun component4 ()Ljava/lang/String;
-	public final fun component5 ()Ljava/lang/String;
-	public final fun copy (Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;)Lio/sentry/android/distribution/DistributionOptions;
-	public static synthetic fun copy$default (Lio/sentry/android/distribution/DistributionOptions;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;ILjava/lang/Object;)Lio/sentry/android/distribution/DistributionOptions;
-	public fun equals (Ljava/lang/Object;)Z
+	public fun <init> ()V
 	public final fun getBuildConfiguration ()Ljava/lang/String;
 	public final fun getOrgAuthToken ()Ljava/lang/String;
 	public final fun getOrganizationSlug ()Ljava/lang/String;
 	public final fun getProjectSlug ()Ljava/lang/String;
 	public final fun getSentryBaseUrl ()Ljava/lang/String;
-	public fun hashCode ()I
-	public fun toString ()Ljava/lang/String;
+	public final fun setBuildConfiguration (Ljava/lang/String;)V
+	public final fun setOrgAuthToken (Ljava/lang/String;)V
+	public final fun setOrganizationSlug (Ljava/lang/String;)V
+	public final fun setProjectSlug (Ljava/lang/String;)V
+	public final fun setSentryBaseUrl (Ljava/lang/String;)V
 }
 
 public final class io/sentry/android/distribution/UpdateInfo {
@@ -85,5 +70,10 @@ public final class io/sentry/android/distribution/UpdateStatus$NewRelease : io/s
 
 public final class io/sentry/android/distribution/UpdateStatus$UpToDate : io/sentry/android/distribution/UpdateStatus {
 	public static final field INSTANCE Lio/sentry/android/distribution/UpdateStatus$UpToDate;
+}
+
+public final class io/sentry/android/distribution/internal/DistributionIntegration : io/sentry/Integration {
+	public fun <init> ()V
+	public fun register (Lio/sentry/IScopes;Lio/sentry/SentryOptions;)V
 }
 


sentry-android-distribution/src/main/java/io/sentry/android/distribution/Distribution.kt
```diff
@@ -1,3 +1,70 @@
...
+import io.sentry.android.distribution.internal.DistributionInternal
+import java.util.concurrent.CompletableFuture
+
+/**
+ * The public Android SDK for Sentry Build Distribution.
+ *
+ * Provides functionality to check for app updates and download new versions from Sentry's preprod
+ * artifacts system.
+ */
+public object Distribution {
```
runningcode, line 14: This is almost exactly the same API as in the Emerge Build Distribution.

sentry-android-distribution/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+  <uses-permission android:name="android.permission.INTERNET" />
+
+  <application>
+    <provider
+      android:name="io.sentry.android.distribution.DistributionContentProvider"
+      android:authorities="${applicationId}.sentry-distribution-provider"
```
chromy: Looks like we can make this a bit shorter: https://github.com/getsentry/sentry-java/blob/c8125f3c9015a841a3830b4544ea9f86ae7b4f76/sentry-android-core/src/main/AndroidManifest.xml#L13

- Just `.SentryDistributionProvider` for name
- `${applicationId}.SentryDistributionProvider` for android:authorities to match prtg

sentry-android-distribution/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,12 @@
...
+      android:exported="false"
+      android:initOrder="100" />
```
chromy: Might be good to add a note (e.g. we want to be lower than perf so perf is enabled first)

sentry-android-distribution/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+  <uses-permission android:name="android.permission.INTERNET" />
```
chromy: I think it's fine to keep this either way but I was wondering - the requirement is you have to have the main Sentry SDK right? You won't be able to have just distro.

sentry-android-distribution/src/main/java/io/sentry/android/distribution/Distribution.kt
```diff
@@ -1,3 +1,70 @@
...
+  /**
+   * Initialize build distribution with the provided options. This should be called once per
+   * process, typically in Application.onCreate().
+   *
+   * @param context Android context
+   * @param options Configuration options for build distribution
+   */
+  public fun init(context: Context, options: DistributionOptions) {
```
chromy: I don't know how well the API surface matches the rest of the Sentry SDK but if it looks good to one of the Sentry folks then fine by me.

chromy (approved):
Boilerplate lgtm! I think it would be good to get a thumbs from SDK team just on the API surface before landing this.

sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionContentProvider.kt
```diff
@@ -0,0 +1,42 @@
...
+import android.net.Uri
+
+/**
+ * ContentProvider that automatically initializes the Sentry Distribution SDK.
+ *
+ * This provider is automatically instantiated by the Android system when the app starts, ensuring
+ * the Distribution SDK is available without requiring manual initialization in
+ * Application.onCreate().
+ */
+public class DistributionContentProvider : ContentProvider() {
```
markushi: It might be worth having a look at `EmptySecureContentProvider`, I remember we had some security issue reports around default content providers.

sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionOptions.kt
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Configuration options for Sentry Build Distribution.
+ *
+ * @param orgAuthToken Organization authentication token for API access
+ * @param organizationSlug Sentry organization slug
+ * @param projectSlug Sentry project slug
+ * @param sentryBaseUrl Base URL for Sentry API (defaults to https://sentry.io)
+ * @param buildConfiguration Optional build configuration name for filtering
+ */
+public data class DistributionOptions(
```
markushi: It's kinda discouraged (e.g. [see this article](https://android.googlesource.com/platform/developers/docs/+/refs/heads/main/api-guidelines/classes.md#data-class)) to use data classes for public APIs, I'm not sure how often this changes - but maybe makes sense to not use it here too.

sentry-android-distribution/src/main/java/io/sentry/android/distribution/Distribution.kt
markushi: For sentry android we also use the following pattern:
```
SentryAndroid.init(context) { config ->
    // config.xyz = 
}
```
In pure Java uses cases this is a bit more readable and easier to maintain when options change a lot, at least when compared against a plain `new DistributionOptions()` ctor, as you need to name the properties. But with named args in Kotlin, we can use this approach too.

sentry-android-distribution/src/main/java/io/sentry/android/distribution/Distribution.kt
```diff
@@ -1,3 +1,70 @@
...
+
+  /**
+   * Check for available updates synchronously (blocking call). This method will block the calling
+   * thread while making the network request. Consider using checkForUpdateCompletableFuture for
+   * non-blocking behavior.
+   *
+   * @param context Android context
+   * @return UpdateStatus indicating if an update is available, up to date, or error
+   */
+  public fun checkForUpdate(context: Context): UpdateStatus {
```
markushi: nit: I'd make this super clear that this is a blocking call
```suggestion
  public fun checkForUpdateBlocking(context: Context): UpdateStatus {
```

sentry-android-distribution/src/main/java/io/sentry/android/distribution/Distribution.kt
```diff
@@ -1,3 +1,70 @@
...
+    return DistributionInternal.checkForUpdate(context)
+  }
+
+  /**
+   * Check for available updates using CompletableFuture for Java compatibility.
+   *
+   * @param context Android context
+   * @return CompletableFuture with UpdateStatus result
+   */
+  public fun checkForUpdateCompletableFuture(context: Context): CompletableFuture<UpdateStatus> {
```
markushi: nit: As a beginner I would prefer a simple callback mechanism instead. 
```suggestion
  public fun checkForUpdate(context: Context, onUpdateStatusLoaded: (UpdateStatus) -> Unit) {
```

sentry-android-distribution/src/main/java/io/sentry/android/distribution/Distribution.kt
markushi: Right now triggering the update check and getting the latest update status is one operation. I'm wondering what the ideal scenario would look like, and if it could make sense to split those two.
- If the ContentProvider would trigger an update check automatically during init, it could make sense to have a `registerUpdateListener()` API, which emits the latest status (if any) + and future triggered checks 
- If you expect people to add "check for update" buttons to their app the existing APIs would better suit this

sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionOptions.kt
```diff
@@ -0,0 +1,18 @@
...
+  val orgAuthToken: String,
+  val organizationSlug: String,
+  val projectSlug: String,
+  val sentryBaseUrl: String = "https://sentry.io",
+  val buildConfiguration: String? = null,
```
markushi: What could be an example value for this?

sentry-android-distribution/src/main/java/io/sentry/android/distribution/Distribution.kt
markushi, line 14: For new-ish features like logging and replay, we're exposing top level entry points under the Sentry class. E.g. `Sentry.logger()` and `Sentry.replay()`. It could make sense to do the same for distribution as well.

sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionContentProvider.kt
runningcode: Good idea!

sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionOptions.kt
runningcode: Good point!

sentry-android-distribution/src/main/AndroidManifest.xml
runningcode: noted

sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionOptions.kt
runningcode: buildConfiguration is like `debug` or `release`. You can only check for updates with the same build configuration. You can’t be on a `debug` build and pull a `release` update.

sentry-android-distribution/src/main/java/io/sentry/android/distribution/Distribution.kt
runningcode: This is a good point, I looked in to it and it seems like a bigger change, I will add this in a future PR.

sentry/src/main/java/io/sentry/Sentry.java
```diff
@@ -1301,6 +1301,23 @@ public static IReplayApi replay() {
     return getCurrentScopes().getScope().getOptions().getReplayController();
   }
 
+  /**
+   * Returns the distribution API. This feature is only available when the
+   * sentry-android-distribution module is included in the build.
+   *
+   * @return The distribution API object that provides update checking functionality
+   */
+  public static @Nullable Object distribution() {
```
runningcode, line 1310: This method uses reflection whereas the `replay()` and `logger()` methods work by calling no-op APIs when the classes aren’t available. To save the extra boilerplate of creating the interfaces and no-op classes, I just used reflection. Thoughts?

sentry-android-distribution/src/main/java/io/sentry/android/distribution/Distribution.kt
runningcode: This is a good point. We expect that most people will not use the API directly, it will just work automatically in the background. When an update is available, then it will be displayed to the user directly.
The API is for scenarios where users want more control over how the update check works.

sentry-android-distribution/src/main/java/io/sentry/android/distribution/Distribution.kt
runningcode: done

sentry-android-distribution/src/main/java/io/sentry/android/distribution/Distribution.kt
runningcode: Updated with this

sentry-android-distribution/src/main/AndroidManifest.xml
runningcode: Good point. I removed this

sentry-android-distribution/src/main/AndroidManifest.xml
romtsn, line 1: I think this is not needed anymore if we define a `namespace` in the gradle config

sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionOptions.kt
```diff
@@ -0,0 +1,19 @@
+package io.sentry.android.distribution
+
+/** Configuration options for Sentry Build Distribution. */
+public class DistributionOptions {
+  /** Organization authentication token for API access */
+  public var orgAuthToken: String = ""
+
+  /** Sentry organization slug */
+  public var organizationSlug: String = ""
```
romtsn, line 9: just a nit, but would be nice to be consistent with naming - i.e. either call this `orgSlug` or change the above to `organizationAuthToken`

sentry-android-distribution/src/main/java/io/sentry/android/distribution/UpdateInfo.kt
romtsn, line 1: since it is part of the public API, should we also use regular `class` here?

sentry-android-distribution/src/main/java/io/sentry/android/distribution/UpdateStatus.kt
```diff
@@ -0,0 +1,13 @@
+package io.sentry.android.distribution
+
+/** Represents the result of checking for app updates. */
+public sealed class UpdateStatus {
+  /** Current app version is up to date, no update available. */
+  public object UpToDate : UpdateStatus()
+
+  /** A new release is available for download. */
+  public data class NewRelease(val info: UpdateInfo) : UpdateStatus()
```
romtsn, line 9: same here about using regular `classes`

sentry-android-distribution/src/main/java/io/sentry/android/distribution/internal/DistributionIntegration.kt
```diff
@@ -0,0 +1,18 @@
+package io.sentry.android.distribution.internal
```
romtsn, line 1: I'm thinking if we should move this out of the `internal` package? Because this will be the entry point for this submodule/integration and be accessed publicly from within `sentry-android-core`

sentry/src/main/java/io/sentry/Sentry.java
romtsn: Cannot say I'm a fan especially of the calling side 😅 
```
val distribution = Sentry.distribution() as? Distribution
```
If it's not a big lift to add the boilerplate, would you mind doing that (or asking claude lol)? I think it'd be great to keep it consistent with the other sub-APIs. You could retrieve the DistriubtionIntegration from `getOptions.getIntegrations()` or have an explicit `distributionController` like I did for replay.

romtsn (commented):
Left some more comments, but otherwise looks great!

sentry/src/main/java/io/sentry/Sentry.java
runningcode: Sounds good, I can do that in a follow up

sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionOptions.kt
runningcode: Sounds good. I’ll do this in a follow up

sentry-android-distribution/src/main/java/io/sentry/android/distribution/UpdateInfo.kt
runningcode: sounds good, will do in a follow up

sentry-android-distribution/src/main/java/io/sentry/android/distribution/internal/DistributionIntegration.kt
runningcode: oh, hmm i messed this up then, I think I should have made the `Distribution` class the entry point. thanks for catching this!

Issue #4261, opened 2025-03-15
Ten DiskReadViolation for


rynkowsg:
### Integration

sentry-android

### Build System

Gradle

### AGP Version

8.9.0

### Proguard

Disabled

### Version

8.3.0, 8.4.0

### Steps to Reproduce

This problem is related to using Strict Mode.
Here is a sample code enabling StrictMode:
```
StrictMode.setThreadPolicy(
            StrictMode.ThreadPolicy.Builder()
                .detectDiskReads()
                .detectDiskWrites()
                .detectCustomSlowCalls()
                .detectAll()
                .penaltyLog()
                .build(),
        )
        StrictMode.setVmPolicy(
            StrictMode.VmPolicy.Builder()
                .detectLeakedSqlLiteObjects()
                .detectLeakedClosableObjects()
                .penaltyLog()
                .build(),
        )
```
and sample function used to init Sentry:
```
internal fun initSentry(
    app: Application,
    sentryDsn: String,
    buildInfo: SentryBuildInfo,
) {
    val ctx = app.applicationContext
    SentryAndroid.init(ctx) { opts ->
        opts.dsn = sentryDsn
        opts.environment = buildInfo.buildEnv
        opts.maxBreadcrumbs = 400
        opts.isAttachViewHierarchy = true
        opts.isAttachScreenshot = true
        opts.isAnrEnabled = true
        opts.isAttachAnrThreadDump = true
        opts.isReportHistoricalAnrs = true
        opts.anrTimeoutIntervalMillis = 2000
        opts.addInAppInclude("my.package.etc")
        opts.release = "${ctx.packageName}@${buildInfo.versionName}"
        opts.addIntegration(
            SentryTimberIntegration(
                minEventLevel = SentryLevel.ERROR,
                minBreadcrumbLevel = SentryLevel.DEBUG,
                // TODO: convert all logs in the project to INFO and here set INFO,
                //  and then gradually change from logs from INFO do DEBUG if consider them redundant.
            ),
        )
        opts.addIntegration(
            FragmentLifecycleIntegration(
                application = app,
                enableFragmentLifecycleBreadcrumbs = true, // enabled by default
                enableAutoFragmentLifecycleTracing = true, // disabled by default
            ),
        )
        opts.isEnableUserInteractionTracing = true
        opts.isEnableUserInteractionBreadcrumbs = true
        opts.isDebug = buildInfo.isDebug
        opts.setDiagnosticLevel(SentryLevel.ERROR)
[...]

PR #4724, opened 2025-09-12
Avoid StrictMode warnings

stefanosiano:
## :scroll: Description
SentryAndroid.init now sets strictMode to Lax and reset it after it finishes
Removed few useless IO calls

Closes https://github.com/getsentry/sentry-java/issues/4261

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/internal/util/AndroidRuntimeManager.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/internal/util/AndroidRuntimeManager.java
@@ -13,9 +13,11 @@ public final class AndroidRuntimeManager implements IRuntimeManager {
     final @NotNull StrictMode.VmPolicy oldVmPolicy = StrictMode.getVmPolicy();
     StrictMode.setThreadPolicy(StrictMode.ThreadPolicy.LAX);
     StrictMode.setVmPolicy(StrictMode.VmPolicy.LAX);
-    final @NotNull T t = toRun.run();
-    StrictMode.setThreadPolicy(oldPolicy);
-    StrictMode.setVmPolicy(oldVmPolicy);
-    return t;
+    try {
+      return toRun.run();
+    } finally {
+      StrictMode.setThreadPolicy(oldPolicy);
+      StrictMode.setVmPolicy(oldVmPolicy);
+    }
   }
 }
--- a/sentry-android-core/src/test/java/io/sentry/android/core/internal/util/AndroidRuntimeManagerTest.kt
+++ b/sentry-android-core/src/test/java/io/sentry/android/core/internal/util/AndroidRuntimeManagerTest.kt
@@ -53,4 +53,35 @@ class AndroidRuntimeManagerTest {
     // Ensure the code ran
     assertTrue(called)
   }
+
+  @Test
+  fun `runWithRelaxedPolicy changes policy and restores it afterwards even if the code throws`() {
+    var called = false
+    val threadPolicy = StrictMode.ThreadPolicy.Builder().detectAll().penaltyDeath().build()
+    val vmPolicy = StrictMode.VmPolicy.Builder().detectAll().penaltyDeath().build()
+
+    // Set and assert the StrictMode policies
+    StrictMode.setThreadPolicy(threadPolicy)
+    StrictMode.setVmPolicy(vmPolicy)
+
+    // Run the function and assert LAX policies
+    try {
+      sut.runWithRelaxedPolicy {
+        assertEquals(
+          StrictMode.ThreadPolicy.LAX.toString(),
+          StrictMode.getThreadPolicy().toString(),
+        )
+        assertEquals(StrictMode.VmPolicy.LAX.toString(), StrictMode.getVmPolicy().toString())
+        called = true
+        throw Exception("Test exception")
+      }
+    } catch (_: Exception) {}
+
+    // Policies should be reverted back
+    assertEquals(threadPolicy.toString(), StrictMode.getThreadPolicy().toString())
+    assertEquals(vmPolicy.toString(), StrictMode.getVmPolicy().toString())
+
+    // Ensure the code ran
+    assertTrue(called)
+  }
 }


sentry/src/main/java/io/sentry/DirectoryProcessor.java
```diff
@@ -40,34 +40,22 @@ public void processDirectory(final @NotNull File directory) {
...
 
       logger.log(
           SentryLevel.DEBUG,
           "Processing %d items from cache dir %s",
-          filteredListFiles != null ? filteredListFiles.length : 0,
+          filteredListFiles.length,
           directory.getAbsolutePath());
 
-      for (File file : listFiles) {
+      for (File file : filteredListFiles) {
```
stefanosiano, line 58: I found we were filtering the files in the directory, but then looped over all the files anyway
Was it by design or by accident?
And is there any risk in changing this?

markushi: I'm pretty sure that was an oversight. IMHO the risk is low as all non-envelope files will fail to be parsed / sent anyway.

markushi (approved):

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
```diff
@@ -93,6 +94,10 @@ public static void init(
       @NotNull final Context context,
       @NotNull ILogger logger,
       @NotNull Sentry.OptionsConfiguration<SentryAndroidOptions> configuration) {
+    final @NotNull StrictMode.ThreadPolicy oldPolicy = StrictMode.getThreadPolicy();
+    final @NotNull StrictMode.VmPolicy oldVmPolicy = StrictMode.getVmPolicy();
+    StrictMode.setThreadPolicy(StrictMode.ThreadPolicy.LAX);
```
romtsn: So I'm thinking this is too relaxing - e.g. the user init config (line 142) will also be under the LAX policy now, which is not what we want probably (e.g. if the user does I/O inside the `init { }` block this would not be flagged anymore). Can we be more specific and only wrap the actual offending calls into LAX?

romtsn (changes requested):
blocking so you don't accidentally merge since there's one approval already

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
stefanosiano: i was also thinking about resetting the policies before the config init (user callback) and then set it back to lax until the end of init()
Wdyt?

romtsn: that probably could do, but we're risking suppressing any new issues that might will have been introduced in the future outside the current ones. I'd really prefer to only target specific issues and set LAX as tight as possible, if it's not a big lift.

sentry/src/main/java/io/sentry/util/runtime/IRuntimeManager.java
```diff
@@ -0,0 +1,13 @@
+package io.sentry.util.runtime;
+
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+@ApiStatus.Internal
+public interface IRuntimeManager {
```
romtsn, line 7: nit: maybe call it `IStrictModeManager` to be more specific? Runtime seems to be too broad for the use-case it solves

sentry/src/main/java/io/sentry/DirectoryProcessor.java
```diff
@@ -40,34 +40,22 @@ public void processDirectory(final @NotNull File directory) {
...
       logger.log(SentryLevel.DEBUG, "Processing dir. %s", directory.getAbsolutePath());
 
-      if (!directory.exists()) {
+      final File[] filteredListFiles = directory.listFiles((d, name) -> isRelevantFileName(name));
+      if (filteredListFiles == null) {
         logger.log(
-            SentryLevel.WARNING,
-            "Directory '%s' doesn't exist. No cached events to send.",
+            SentryLevel.ERROR,
+            "Cache dir %s is null or is not a directory.",
```
romtsn, line 47: so does `listFiles` have the `isDirectory` check under the hood?

sentry/src/main/java/io/sentry/util/runtime/IRuntimeManager.java
stefanosiano: yeah, i wanted to make it more "broad" on purpose, in case we want to do something else with the system.
But i can change it if you think `IStrictModeManager` is more appropriate

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/AndroidRuntimeManager.java
```diff
@@ -0,0 +1,21 @@
...
+  public <T> T runWithRelaxedPolicy(final @NotNull IRuntimeManagerCallback<T> toRun) {
+    final @NotNull StrictMode.ThreadPolicy oldPolicy = StrictMode.getThreadPolicy();
+    final @NotNull StrictMode.VmPolicy oldVmPolicy = StrictMode.getVmPolicy();
+    StrictMode.setThreadPolicy(StrictMode.ThreadPolicy.LAX);
+    StrictMode.setVmPolicy(StrictMode.VmPolicy.LAX);
+    final @NotNull T t = toRun.run();
+    StrictMode.setThreadPolicy(oldPolicy);
+    StrictMode.setVmPolicy(oldVmPolicy);
+    return t;
+  }
```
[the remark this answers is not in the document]
romtsn: sounds legit

sentry/src/main/java/io/sentry/util/runtime/IRuntimeManager.java
romtsn: alright, let's keep it then, I'm fine with that too!

sentry/src/main/java/io/sentry/DirectoryProcessor.java
stefanosiano: yep. From the javadoc
`If this abstract pathname does not denote a directory, then this method returns null`

romtsn:
@stefanosiano I realized there's a couple of more cases which we didn't cover from the issue:
```
Installation.id()
AndroidEnvelopeCache.hasStartupCrashMarker
```
Shall we also wrap them?

stefanosiano:
> @stefanosiano I realized there's a couple of more cases which we didn't cover from the issue:
> 
> ```
> Installation.id()
> AndroidEnvelopeCache.hasStartupCrashMarker
> ```
> 
> Shall we also wrap them?

Covered in https://github.com/getsentry/sentry-java/pull/4809

PR #4730, opened 2025-09-15
feat(android-distribution): Use ProGuard UUID for build identification

runningcode:
Use existing ProGuard UUID system for build identification instead of custom APK parsing.

Uses: `Sentry.getCurrentScopes().options.proguardUuid`

#skip-changelog


runningcode:
We'll need a separate UUID.

PR #4731, opened 2025-09-15
refactor(distribution): Use interface pattern for distribution API

runningcode:
Addresses architectural feedback from PR #4712 by refactoring the distribution API to follow established Sentry patterns.

## Changes
- Created `IDistributionApi` with `checkForUpdateBlocking()`, `checkForUpdate()`, `downloadUpdate()` methods
- Moved`UpdateStatus`, `UpdateInfo`, and `UpdateCallback` types to core sentry module because they are part of the API
- Added `distributionController` to `SentryOptions` following replay API pattern
- Updated `Sentry.distribution()` to use `getCurrentScopes().getScope().getOptions().getDistributionController()` instead of reflection
- Updated `DistributionIntegration` to implement `IDistributionApi` directly
- Follows exact same pattern as `replay()` API

#skip-changelog

Distribution integrations register real implementations during setup, with graceful NoOp fallback when the module is unavailable.


sentry/src/main/java/io/sentry/UpdateStatus.java
```diff
@@ -0,0 +1,44 @@
+package io.sentry;
+
+import org.jetbrains.annotations.NotNull;
+
+/** Represents the result of checking for app updates. */
+public abstract class UpdateStatus {
```
romtsn, line 6: wondering if this can be used as a sealed class in Kotlin (i.e. doing a `when` over this)? I guess not, but this API is fine for me - just brainstorming if we should change it an enum to support `when`-statements (but then it wouldn't be easily typed)

romtsn (approved):
LGTM, let's try to init automatically in `register` as discussed internally

sentry/src/main/java/io/sentry/UpdateStatus.java
runningcode: Yes, it was originally a sealed class but now that it is in the Sentry module I changed it to java. are we allowed to use Kotlin in the Sentry module??

PR #4733, opened 2025-09-17
feat(distribution): Mark build distribution APIs as experimental

runningcode:
## Summary
- Mark all build distribution APIs with `@ApiStatus.Experimental`

There was a dependency conflict because the kotlin-stdlib bring in 13.0 of the annotations. 
Adjusted `compileOnly(libs.jetbrains.annotations)` to `implementation(libs.jetbrains.annotations)` in the sentry-android-distribution module. 

Closes EME-281

#skip-changelog since it isn't released

### DIFF
--- a/sentry-android-distribution/build.gradle.kts
+++ b/sentry-android-distribution/build.gradle.kts
@@ -25,5 +25,8 @@ androidComponents.beforeVariants {
 
 dependencies {
   implementation(projects.sentry)
+  implementation(
+    libs.jetbrains.annotations
+  ) // Use implementation instead of compileOnly to override kotlin stdlib's version
   implementation(kotlin(Config.kotlinStdLib, Config.kotlinStdLibVersionAndroid))
 }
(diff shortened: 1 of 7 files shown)


sentry-android-distribution/build.gradle.kts
```diff
@@ -25,5 +25,8 @@ androidComponents.beforeVariants {
 
 dependencies {
   implementation(projects.sentry)
+  implementation(
```
runningcode, line 28: Other modules force the version (like `sentry-android-core`) which could cause unexpected behavior in consumers.

I thought using `implementation` is simpler to understand than a version constraint.

romtsn: hm I guess this is a bit weird that we leak the annotations to the runtime classpath (given that we only need them for compilation), but if there's no easy way around it, it's fine I guess.

romtsn (approved):

sentry-android-distribution/build.gradle.kts
runningcode: Annotations are quite leanient in the jvm world IIRC. If their dependencies aren’t present at compile time, it will still work so clients of the library can use an older version and it should be fine.

PR #4734, opened 2025-09-17
feat(android-distribution): add httpclient for checking for build distribution updates

runningcode:
## Summary

Implements the `checkForUpdateBlocking` method in `DistributionIntegration` to check for app updates via Sentry's distribution API.

## Why not reuse existing HttpConnection?

The existing `HttpConnection` class is designed specifically for Sentry event transport and is not suitable for distribution API calls:
- Hardcoded for POST requests (we need GET)
- Expects Sentry envelopes with gzip encoding (we need simple JSON)
- Only considers status 200 successful (REST APIs use 200-299 range)
- Includes Sentry-specific rate limiting logic

## Changes

- **DistributionHttpClient**: New HTTP client for distribution API requests
  - Supports GET requests with query parameters (main_binary_identifier, app_id, platform, version)
  - Uses SentryOptions.DistributionOptions for configuration (orgSlug, projectSlug, orgAuthToken)

- **UpdateResponseParser**: JSON response parser for API responses

#skip-changelog

---

> [!NOTE]
> Adds an HTTP client and JSON parser for distribution update checks, implements checkForUpdateBlocking with network/error handling, updates API types, and adds tests.
> 
> - **Android Distribution**:
>   - **Update Check Flow**: Implements `DistributionIntegration.checkForUpdateBlocking` using new `DistributionHttpClient` and `UpdateResponseParser`; builds params from app package info and opens download via `downloadUpdate`.
>   - **Networking & Errors**: Adds detailed error handling (e.g., `UnknownHostException`, `SocketTimeoutException`) and logs; introduces `UpdateStatus.NoNetwork` result.
>   - **API Changes**: `UpdateInfo.getCreatedDate()` becomes nullable; new `UpdateStatus.NoNetwork` class.
> - **Testing**:
>   - Adds unit tests `DistributionHttpClientTest` (ignored real API test) and `UpdateResponseParserTest` (covers success, errors, edge cases).
> - **Build**:
>   - Enables unit test Android resources/default values and adds test dependencies in `build.gradle.kts`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ff0acf5ed2473f50445a52baa24a81f66e3dc9ad. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionHttpClient.kt
+++ b/sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionHttpClient.kt
@@ -24,7 +24,7 @@ internal class DistributionHttpClient(private val options: SentryOptions) {
     val mainBinaryIdentifier: String,
     val appId: String,
     val platform: String = "android",
-    val versionCode: Int,
+    val versionCode: Long,
     val versionName: String,
   )
 
--- a/sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionHttpClient.kt
+++ b/sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionHttpClient.kt
@@ -1,0 +1,124 @@
+package io.sentry.android.distribution
+
+import io.sentry.SentryLevel
+import io.sentry.SentryOptions
+import java.io.BufferedReader
+import java.io.IOException
+import java.io.InputStreamReader
+import java.net.HttpURLConnection
+import java.net.URL
+import java.net.URLEncoder
+import javax.net.ssl.HttpsURLConnection
+
+/** HTTP client for making requests to Sentry's distribution API. */
+internal class DistributionHttpClient(private val options: SentryOptions) {
+
+  /** Represents the result of an HTTP request. */
+  data class HttpResponse(
+    val statusCode: Int,
+    val body: String,
+    val isSuccessful: Boolean = statusCode in 200..299,
+  )
+
+  /** Parameters for checking updates. */
+  data class UpdateCheckParams(
+    val mainBinaryIdentifier: String,
+    val appId: String,
+    val platform: String = "android",
+    val version: String,
+  )
+
+  /**
+   * Makes a GET request to the distribution API to check for updates.
+   *
+   * @param params Update check parameters
+   * @return HttpResponse containing the response details
+   */
+  fun checkForUpdates(params: UpdateCheckParams): HttpResponse {
+    val distributionOptions = options.distribution
+    val orgSlug = distributionOptions.orgSlug
+    val projectSlug = distributionOptions.projectSlug
+    val authToken = distributionOptions.orgAuthToken
+    val baseUrl = distributionOptions.sentryBaseUrl
+
+    if (orgSlug.isEmpty() || projectSlug.isEmpty() || authToken.isEmpty()) {
+      throw IllegalStateException(
+        "Missing required distribution configuration: orgSlug, projectSlug, or orgAuthToken"
+      )
+    }
+
+    val queryParams = buildQueryParams(params)
+    val url =
+      URL(
+        "$baseUrl/api/0/projects/$orgSlug/$projectSlug/preprodartifacts/check-for-updates/?$queryParams"
+      )
+
+    return try {
+      makeRequest(url, authToken)
+    } catch (e: IOException) {
+      options.logger.log(SentryLevel.ERROR, e, "Network error while checking for updates")
+      throw e
+    }
+  }
+
+  private fun makeRequest(url: URL, authToken: String): HttpResponse {
+    val connection = url.openConnection() as HttpURLConnection
+
+    try {
+      // Configure connection
+      connection.requestMethod = "GET"
+      connection.setRequestProperty("Authorization", "Bearer $authToken")
+      connection.setRequestProperty("Accept", "application/json")
+      connection.setRequestProperty(
+        "User-Agent",
+        options.sentryClientName ?: "sentry-android-distribution",
+      )
+      connection.connectTimeout = options.connectionTimeoutMillis
+      connection.readTimeout = options.readTimeoutMillis
+
+      // Set SSL socket factory if available
+      if (connection is HttpsURLConnection && options.sslSocketFactory != null) {
+        connection.sslSocketFactory = options.sslSocketFactory
+      }
+
+      // Get response
+      val responseCode = connection.responseCode
+      val responseBody = readResponse(connection)
+
+      options.logger.log(
+        SentryLevel.DEBUG,
+        "Distribution API request completed with status: $responseCode",
+      )
+
+      return HttpResponse(responseCode, responseBody)
+    } finally {
+      connection.disconnect()
+    }
+  }
+
+  private fun readResponse(connection: HttpURLConnection): String {
+    val inputStream =
+      if (connection.responseCode in 200..299) {
+        connection.inputStream
+      } else {
+        connection.errorStream ?: connection.inputStream
+      }
+
+    return inputStream?.use { stream ->
+      BufferedReader(InputStreamReader(stream, "UTF-8")).use { reader -> reader.readText() }
+    } ?: ""
+  }
+
+  private fun buildQueryParams(params: UpdateCheckParams): String {
+    val queryParams = mutableListOf<String>()
+
+    queryParams.add(
+      "main_binary_identifier=${URLEncoder.encode(params.mainBinaryIdentifier, "UTF-8")}"
+    )
+    queryParams.add("app_id=${URLEncoder.encode(params.appId, "UTF-8")}")
+    queryParams.add("platform=${URLEncoder.encode(params.platform, "UTF-8")}")
+    queryParams.add("version=${URLEncoder.encode(params.version, "UTF-8")}")
+
+    return queryParams.joinToString("&")
+  }
+}
--- a/sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionHttpClient.kt
+++ b/sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionHttpClient.kt
@@ -7,6 +7,7 @@ import java.io.IOException
 import java.io.InputStreamReader
 import java.net.HttpURLConnection
 import java.net.URL
+import java.net.URLEncoder
 import javax.net.ssl.HttpsURLConnection
 
 /** HTTP client for making requests to Sentry's distribution API. */
@@ -41,7 +42,7 @@ internal class DistributionHttpClient(private val options: SentryOptions) {
     val authToken = distributionOptions.orgAuthToken
     val baseUrl = distributionOptions.sentryBaseUrl
 
-    if (orgSlug.isEmpty() || projectSlug.isEmpty() || authToken.isEmpty()) {
+    if (orgSlug.isNullOrEmpty() || projectSlug.isNullOrEmpty() || authToken.isNullOrEmpty()) {
       throw IllegalStateException(
         "Missing required distribution configuration: orgSlug, projectSlug, or orgAuthToken"
       )
@@ -49,12 +50,14 @@ internal class DistributionHttpClient(private val options: SentryOptions) {
 
     val urlString = buildString {
       append(baseUrl.trimEnd('/'))
-      append("/api/0/projects/$orgSlug/$projectSlug/preprodartifacts/check-for-updates/")
-      append("?main_binary_identifier=${params.mainBinaryIdentifier}")
-      append("&app_id=${params.appId}")
-      append("&platform=${params.platform}")
-      append("&build_number=${params.versionCode}")
-      append("&build_version=${params.versionName}")
+      append(
+        "/api/0/projects/${URLEncoder.encode(orgSlug, "UTF-8")}/${URLEncoder.encode(projectSlug, "UTF-8")}/preprodartifacts/check-for-updates/"
+      )
+      append("?main_binary_identifier=${URLEncoder.encode(params.mainBinaryIdentifier, "UTF-8")}")
+      append("&app_id=${URLEncoder.encode(params.appId, "UTF-8")}")
+      append("&platform=${URLEncoder.encode(params.platform, "UTF-8")}")
+      append("&build_number=${URLEncoder.encode(params.versionCode.toString(), "UTF-8")}")
+      append("&build_version=${URLEncoder.encode(params.versionName, "UTF-8")}")
     }
     val url = URL(urlString)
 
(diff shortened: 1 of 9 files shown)


sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionHttpClient.kt
```diff
@@ -0,0 +1,124 @@
...
+import java.io.BufferedReader
+import java.io.IOException
+import java.io.InputStreamReader
+import java.net.HttpURLConnection
+import java.net.URL
+import java.net.URLEncoder
+import javax.net.ssl.HttpsURLConnection
+
+/** HTTP client for making requests to Sentry's distribution API. */
+internal class DistributionHttpClient(private val options: SentryOptions) {
```
runningcode, line 14: I had claude generate an HttpClient for me. It works. I'm just not sure if it is a good idea to develop another client.

The other alternative is to expand the existing `HttpConnection` to support our use case OR shadow oktthp in to our library. Open to thoughts!

sentry-android-distribution/src/test/java/io/sentry/android/distribution/DistributionHttpClientTest.kt
```diff
@@ -0,0 +1,54 @@
+package io.sentry.android.distribution
+
+import io.sentry.SentryOptions
+import org.junit.Assert.*
+import org.junit.Before
+import org.junit.Ignore
+import org.junit.Test
+
+class DistributionHttpClientTest {
```
runningcode, line 9: This is just a test class I used to test the integration. In a future, SAGP will inject the authToken and other things so we won't need to specify this manually. I'll also integrate distribution in to the sample app.

sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionHttpClient.kt
```diff
@@ -0,0 +1,124 @@
...
+      options.logger.log(SentryLevel.ERROR, e, "Network error while checking for updates")
+      throw e
+    }
+  }
+
+  private fun makeRequest(url: URL, authToken: String): HttpResponse {
+    val connection = url.openConnection() as HttpURLConnection
+
+    try {
+      // Configure connection
```
chromy: nit: the classic claude comments I don't find valuable normally

sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionHttpClient.kt
```diff
@@ -0,0 +1,124 @@
...
+      connection.setRequestProperty("Authorization", "Bearer $authToken")
+      connection.setRequestProperty("Accept", "application/json")
+      connection.setRequestProperty(
+        "User-Agent",
+        options.sentryClientName ?: "sentry-android-distribution",
+      )
+      connection.connectTimeout = options.connectionTimeoutMillis
+      connection.readTimeout = options.readTimeoutMillis
+
+      // Set SSL socket factory if available
```
chromy: same here and below

sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionHttpClient.kt
```diff
@@ -0,0 +1,124 @@
...
+    if (orgSlug.isEmpty() || projectSlug.isEmpty() || authToken.isEmpty()) {
+      throw IllegalStateException(
+        "Missing required distribution configuration: orgSlug, projectSlug, or orgAuthToken"
+      )
+    }
+
+    val queryParams = buildQueryParams(params)
+    val url =
+      URL(
+        "$baseUrl/api/0/projects/$orgSlug/$projectSlug/preprodartifacts/check-for-updates/?$queryParams"
```
chromy: Could use https://developer.android.com/reference/android/net/Uri.Builder.html here

sentry-android-distribution/src/main/java/io/sentry/android/distribution/UpdateResponseParser.kt
```diff
@@ -0,0 +1,68 @@
...
+
+import io.sentry.SentryLevel
+import io.sentry.SentryOptions
+import io.sentry.UpdateInfo
+import io.sentry.UpdateStatus
+import org.json.JSONException
+import org.json.JSONObject
+
+/** Parser for distribution API responses. */
+internal class UpdateResponseParser(private val options: SentryOptions) {
```
chromy, line 11: It might be worth having some tests for this also since it is nicely separated

sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionIntegration.kt
```diff
@@ -44,7 +54,19 @@ public class DistributionIntegration(context: Context) : Integration, IDistribut
...
+
+      val params = createUpdateCheckParams()
+      val response = httpClient.checkForUpdates(params)
+      responseParser.parseResponse(response.statusCode, response.body)
+    } catch (e: IllegalStateException) {
+      sentryOptions.logger.log(SentryLevel.WARNING, e.message ?: "Configuration error")
+      UpdateStatus.UpdateError(e.message ?: "Configuration error")
+    } catch (e: Exception) {
+      sentryOptions.logger.log(SentryLevel.ERROR, e, "Failed to check for updates")
+      UpdateStatus.UpdateError("Network error: ${e.message}")
```
chromy: It might be nice to be able to distinguish between 'not connected' and 'network error' in the return type if that is possible and easy.

sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionIntegration.kt
```diff
@@ -75,4 +97,30 @@ public class DistributionIntegration(context: Context) : Integration, IDistribut
...
+          packageManager.getPackageInfo(packageName, PackageManager.PackageInfoFlags.of(0))
+        } else {
+          @Suppress("DEPRECATION") packageManager.getPackageInfo(packageName, 0)
+        }
+
+      val versionName = packageInfo.versionName ?: "unknown"
+      val appId = context.applicationInfo.packageName
+
+      DistributionHttpClient.UpdateCheckParams(
+        mainBinaryIdentifier = appId, // Using package name as binary identifier
```
chromy: Do we want this?

sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionIntegration.kt
```diff
@@ -75,4 +97,30 @@ public class DistributionIntegration(context: Context) : Integration, IDistribut
...
+        appId = appId,
+        platform = "android",
+        version = versionName,
```
chromy: I think we're missing version code iirc (Android has both a human facing version and a machine facing version for apps, one is versionCode and the other versionNumber)

sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionHttpClient.kt
```diff
@@ -0,0 +1,124 @@
...
+    val connection = url.openConnection() as HttpURLConnection
+
+    try {
+      // Configure connection
+      connection.requestMethod = "GET"
+      connection.setRequestProperty("Authorization", "Bearer $authToken")
+      connection.setRequestProperty("Accept", "application/json")
+      connection.setRequestProperty(
+        "User-Agent",
+        options.sentryClientName ?: "sentry-android-distribution",
```
chromy: We maybe should default to the same as here:
https://github.com/getsentry/sentry-java/blob/d21770841d84eddba0636e2082c029db2a158457/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java#L231

(or, since it seems like it is always set despite being nullable we could throw if it is unset)

chromy (commented):
Thanks!

sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionIntegration.kt
runningcode: nope, will remove

sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionIntegration.kt
runningcode: Yes good point. I didn’t see versionCode as a parameter to the check for updates APi.
https://github.com/getsentry/sentry/blob/master/src/sentry/preprod/api/endpoints/project_preprod_check_for_updates.py#L62

Should I pass this in as versionName + versionCode or should we add an extra parameter to the API?

sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionHttpClient.kt
runningcode: sounds good, i added a throw if unset

sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionIntegration.kt
runningcode: `UnknownHostException` would mean there is possibly no network and `SocketTimeoutException` could mean a bad network. I added those to distinguish here.

sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionIntegration.kt
chromy: We should add an extra parameter, we called it 'build' before: https://github.com/EmergeTools/emerge/blob/97c2adfe7e133e871c37beae2ff1b618002b87bd/backend/services/emerge/src/distribution/api/checkForUpdates.ts#L25C9-L25C22.

In Sentry backend it's called `build_number` https://github.com/getsentry/sentry/blob/46a9e4fd4fedb3f195b77db0b137c7daf167e8e2/src/sentry/preprod/models.py#L110C5-L110C17

We can call the parameter `build_number` maybe? That seems more clear then `build` for me.

- `versionName` should be `version`
- `versionCode` should be the other one

sentry-android-distribution/src/main/java/io/sentry/android/distribution/UpdateResponseParser.kt
chromy: thanks!

sentry-android-distribution/src/main/java/io/sentry/android/distribution/UpdateResponseParser.kt
```diff
@@ -0,0 +1,75 @@
...
+
+  private fun parseUpdateInfo(json: JSONObject): UpdateInfo {
+    val id = json.optString("id", "")
+    val buildVersion = json.optString("buildVersion", "")
+    val buildNumber = json.optInt("buildNumber", 0)
+    val downloadUrl = json.optString("downloadUrl", "")
+    val appName = json.optString("appName", "")
+    val createdDate = json.optString("createdDate", "")
+
+    // Validate required fields (optString returns "null" for null values)
```
chromy, line 61: What an odd API :/

sentry-android-distribution/src/main/java/io/sentry/android/distribution/UpdateResponseParser.kt
```diff
@@ -0,0 +1,75 @@
...
+    }
+  }
+
+  private fun parseUpdateInfo(json: JSONObject): UpdateInfo {
+    val id = json.optString("id", "")
+    val buildVersion = json.optString("buildVersion", "")
+    val buildNumber = json.optInt("buildNumber", 0)
+    val downloadUrl = json.optString("downloadUrl", "")
+    val appName = json.optString("appName", "")
+    val createdDate = json.optString("createdDate", "")
```
chromy, line 59: If it's fine for this to be unset then maybe it should be nullable in the `UpdateInfo` type

sentry-android-distribution/src/main/java/io/sentry/android/distribution/UpdateResponseParser.kt
```diff
@@ -0,0 +1,75 @@
...
+    // Validate required fields (optString returns "null" for null values)
+    if (
+      id.isEmpty() ||
+        id == "null" ||
+        buildVersion.isEmpty() ||
+        buildVersion == "null" ||
+        downloadUrl.isEmpty() ||
+        downloadUrl == "null"
+    ) {
+      throw IllegalArgumentException("Missing required update information in API response")
```
chromy: Lets try and give a better error message for future us, either could do this more like:

```
if (is.isEmpty()) {
  throw IllegalArgumentException("id must be set")
}
```

or could just do:

```
if (...) {
  throw IllegalArgumentException(Missing required update information in API response '" + body + "'")
}
```
or something

sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionIntegration.kt
```diff
@@ -44,7 +56,31 @@ public class DistributionIntegration(context: Context) : Integration, IDistribut
...
+      sentryOptions.logger.log(SentryLevel.WARNING, e.message ?: "Configuration error")
+      UpdateStatus.UpdateError(e.message ?: "Configuration error")
+    } catch (e: UnknownHostException) {
+      // UnknownHostException typically indicates no internet connection available
+      sentryOptions.logger.log(
+        SentryLevel.ERROR,
+        e,
+        "DNS lookup failed - check internet connection",
+      )
+      UpdateStatus.UpdateError("No internet connection or invalid server URL")
```
chromy: We could add a state for this and below. `UpdateStatus.NoNetwork` or `UpdateStatus.Unknown`

sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionIntegration.kt
runningcode: Ah i see, but it looks like the backend API doesn’t support this yet. I’ll look in to adding this. https://github.com/getsentry/sentry/blob/master/src/sentry/preprod/api/endpoints/project_preprod_check_for_updates.py#L65

sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionIntegration.kt
runningcode: added!

sentry-android-distribution/src/main/java/io/sentry/android/distribution/UpdateResponseParser.kt
runningcode: added much clearer errors!

sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionIntegration.kt
runningcode: added

PR #4740, opened 2025-09-19
Add session replay id to Sentry Logs

stefanosiano:
## :scroll: Description
added session replay id to logs, if a replay is running

Resolves https://linear.app/getsentry/issue/ANDROID-210/add-sentryreplay-id-to-android-logs

## :pencil: Checklist

### DIFF
--- a/sentry/src/main/java/io/sentry/logger/LoggerApi.java
+++ b/sentry/src/main/java/io/sentry/logger/LoggerApi.java
@@ -211,6 +211,14 @@ public final class LoggerApi implements ILoggerApi {
           "sentry.environment",
           new SentryLogEventAttributeValue(SentryAttributeType.STRING, environment));
     }
+
+    final @Nullable SentryId replayId = scopes.getScope().getReplayId();
+    if (!replayId.equals(SentryId.EMPTY_ID)) {
+      attributes.put(
+          "sentry.replay_id",
+          new SentryLogEventAttributeValue(SentryAttributeType.STRING, replayId.toString()));
+    }
+
     final @Nullable String release = scopes.getOptions().getRelease();
     if (release != null) {
       attributes.put(
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
@@ -208,11 +208,11 @@ public class AndroidContinuousProfiler
 
     isRunning = true;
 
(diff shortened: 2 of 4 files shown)


stefanosiano:
I've seen that if the replay is in buffer mode, no replay id is set to the scope until it's captured, and then it is converted to session mode
So I think we should never see the `sentry._internal.replay_is_buffering` attribute set, right?
This means that we are actually never sending the replay id (because it's not set to the scope) when the replay is in buffer mode. Should we change that?

sentry/src/main/java/io/sentry/logger/LoggerApi.java
```diff
@@ -211,6 +212,19 @@ private void captureLog(
           "sentry.environment",
           new SentryLogEventAttributeValue(SentryAttributeType.STRING, environment));
     }
+
+    final @Nullable SentryId replayId = scopes.getCombinedScopeView().getReplayId();
```
romtsn: so I think we can simplify this by quite a bit:
- First get the replay_id from `scopes`
  - If it's not empty that means all good and we're in session mode (or about to enter it) - hence we don't have to send `sentry._internal.replay_is_buffering` at all
  - If it's empty - we try to get the replay_id from `scopes.getOptions().getReplayController().getReplayId()`
    - If it's not empty that means we're in buffer mode - hence we have to send  `sentry._internal.replay_is_buffering: true`
    - If this one is still empty - that means there's no replay being recorded right now and we just don't send any of that
    
 This way we wouldn't have to keep `replayType` in the scope either. How that sound?

stefanosiano: @romtsn do you like it now? 😄

sentry/src/main/java/io/sentry/logger/LoggerApi.java
```diff
@@ -211,6 +211,22 @@ private void captureLog(
...
+          new SentryLogEventAttributeValue(SentryAttributeType.STRING, scopeReplayId.toString()));
+    } else {
+      final @Nullable SentryId controllerReplayId =
+          scopes.getOptions().getReplayController().getReplayId();
+      if (!controllerReplayId.equals(SentryId.EMPTY_ID)) {
+        attributes.put(
+            "sentry._internal.replay_is_buffering",
+            new SentryLogEventAttributeValue(SentryAttributeType.BOOLEAN, true));
+      }
+    }
```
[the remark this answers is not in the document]
romtsn: i think the nullability point is also legit here, would be nice to address

PR #4748, opened 2025-09-23
Mark `SentryEnvelope` as not internal

lcian:
## :scroll: Description

To implement a custom transport, a user needs to define `ITransport::send` which takes a `SentryEnvelope`, so we should mark that one as not internal.

Other than that, this particular user also mentioned `SentryEnvelopeHeader, SentryEnvelopeItem, SentryItemType`, but I would avoid making those not internal unless they confirm they really need it for their usecase.

Close https://github.com/getsentry/sentry-java/issues/4722
Close JAVA-175

### DIFF
--- a/sentry/src/main/java/io/sentry/SentryItemType.java
+++ b/sentry/src/main/java/io/sentry/SentryItemType.java
@@ -4,7 +4,6 @@ import io.sentry.clientreport.ClientReport;
 import io.sentry.protocol.SentryTransaction;
 import java.io.IOException;
 import java.util.Locale;
-import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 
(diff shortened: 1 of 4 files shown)


sentry/src/main/java/io/sentry/SentryItemType.java
```diff
@@ -4,10 +4,8 @@
 import io.sentry.protocol.SentryTransaction;
 import java.io.IOException;
 import java.util.Locale;
-import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 
-@ApiStatus.Internal
 public enum SentryItemType implements JsonSerializable {
```
lcian, line 11: I wonder if adding a field to this enum counts as a breaking change in the Java/Kotlin world?
If yes, I'll avoid removing the annotation here.

PR #4752, opened 2025-09-24
Start performance collection on AppStart continuous profiling

stefanosiano:
## :scroll: Description
app start continuous profiler will now start performance collection when the SDK inits

app start continuous profiles were missing performance data on first chunk (cpu and ram)

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
@@ -5,6 +5,7 @@ import static io.sentry.android.core.NdkIntegration.SENTRY_NDK_CLASS_NAME;
 import android.app.Application;
 import android.content.Context;
 import android.content.pm.PackageInfo;
+import io.sentry.CompositePerformanceCollector;
 import io.sentry.DeduplicateMultithreadedEventProcessor;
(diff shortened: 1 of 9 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -262,6 +244,27 @@ static void initializeIntegrationsAndProcessors(
     if (options.getCompositePerformanceCollector() instanceof NoOpCompositePerformanceCollector) {
       options.setCompositePerformanceCollector(new DefaultCompositePerformanceCollector(options));
     }
+
+    // Check if the profiler was already instantiated in the app start.
+    // We use the Android profiler, that uses a global start/stop api, so we need to preserve the
+    // state of the profiler, and it's only possible retaining the instance.
+    final @NotNull AppStartMetrics appStartMetrics = AppStartMetrics.getInstance();
```
romtsn: I'm a little bit worried about moving this down here - there's some stuff happening in the `AppStartMetrics` constructor which may affect some other parts. Can we keep this line as it was before and only move the profiler instantiating down?

romtsn (approved):
one nit, but lgtm otherwise

PR #4773, opened 2025-10-01
Ref(android-distribution): Remove main_binary_identifier parameter (EME-391)

runningcode:
## Summary

Removes the redundant `main_binary_identifier` parameter from the Build Distribution feature. This parameter was always set to the same value as `app_id`, making it redundant.

## Changes

- Removed `mainBinaryIdentifier` field from `UpdateCheckParams` data class
- Removed `main_binary_identifier` query parameter from API URL construction
- Updated `DistributionIntegration` to no longer pass the redundant parameter
- Updated test to reflect the changes

### DIFF
--- a/sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionHttpClient.kt
+++ b/sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionHttpClient.kt
@@ -22,7 +22,6 @@ internal class DistributionHttpClient(private val options: SentryOptions) {
 
   /** Parameters for checking updates. */
   data class UpdateCheckParams(
-    val mainBinaryIdentifier: String,
     val appId: String,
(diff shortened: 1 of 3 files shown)


sentry-android-distribution/src/main/java/io/sentry/android/distribution/DistributionHttpClient.kt
```diff
@@ -22,7 +22,6 @@ internal class DistributionHttpClient(private val options: SentryOptions) {
 
   /** Parameters for checking updates. */
   data class UpdateCheckParams(
-    val mainBinaryIdentifier: String,
```
runningcode, line 25: I decided to just completely remove this field assuming we will never use it on android

markushi (approved):
LGTM!

PR #4777, opened 2025-10-02
Add Canvas Capture Strategy

markushi:
## :scroll: Description
Adds an experimental option to enable an alternate way to capture screenshots, using a fake Canvas which draws rectangles instead of text, producing a safer way to mask all sensitive content.

No masking issues.

Manually

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
@@ -481,7 +481,7 @@ final class ManifestMetadataReader {
         final @Nullable String screenshotStrategyRaw =
             readString(metadata, logger, REPLAYS_SCREENSHOT_STRATEGY, null);
         if (screenshotStrategyRaw != null) {
-          if ("canvas".equals(screenshotStrategyRaw)) {
+          if ("canvas".equals(screenshotStrategyRaw.toLowerCase(Locale.ROOT))) {
             options.getSessionReplay().setScreenshotStrategy(ScreenshotStrategyType.CANVAS);
           } else {
             // always default to PIXEL_COPY


sentry-android-replay/src/main/java/io/sentry/android/replay/WindowRecorder.kt
romtsn, line 1: are the changes in this class necessary? I don't see `rootView` or `currentCaptureDelay` being used anywhere

sentry-android-replay/src/main/java/io/sentry/android/replay/screenshot/CanvasStrategy.kt
```diff
@@ -0,0 +1,890 @@
...
+        PictureReaderHolder(config.recordingWidth, config.recordingHeight),
+      )
+    )
+  private val unprocessedPictures: MutableList<PictureReaderHolder> = LinkedList()
+
+  private val processingThread = HandlerThread("screenshot_recorder.canvas")
+  private val processingHandler: Handler
+
+  init {
+    processingThread.start()
```
romtsn: looking at this, we re-initialize `ScreenshoRecorder` every time there's a window/configuration change (which subsequently initializes this strategy), therefore we call `thread.start` and init the handler potentially multiple times during a single replay. Would it make sense to do that only once?

sentry-android-replay/src/main/java/io/sentry/android/replay/screenshot/CanvasStrategy.kt
```diff
@@ -0,0 +1,890 @@
...
+        when {
+          unprocessedPictures.isNotEmpty() -> unprocessedPictures.removeAt(0)
+          else -> null
+        }
+      }
+    if (holder == null) {
+      return@Runnable
+    }
+    try {
+      holder.reader.setOnImageAvailableListener(
```
romtsn: Is it possible to set this callback only once, or do we have to do it every time this Runnable runs?

sentry-android-replay/src/main/java/io/sentry/android/replay/screenshot/CanvasStrategy.kt
```diff
@@ -0,0 +1,890 @@
...
+        {
+          val hwImage = it?.acquireLatestImage()
+          if (hwImage != null) {
+            val hwScreenshot = toBitmap(hwImage)
+            screenshot = hwScreenshot
```
romtsn: I guess we also have to synchronize access to `screenshot` (or use atomicRef) because it's being accessed from different threads potentially?

sentry-android-replay/src/main/java/io/sentry/android/replay/screenshot/CanvasStrategy.kt
```diff
@@ -0,0 +1,890 @@
...
+            options.logger.log(SentryLevel.DEBUG, "Canvas Strategy: hardware image is null")
+          }
+          synchronized(freePictures) { freePictures.add(holder) }
+        },
+        processingHandler,
+      )
+
+      val surface = holder.reader.surface
+      val canvas = surface.lockHardwareCanvas()
+      canvas.drawColor(Color.BLACK, PorterDuff.Mode.CLEAR)
```
romtsn: could we use `TRANSPARENT` here actually? Wondering if there will be some weird artifacts due to us erasing with  the black color

sentry-android-replay/src/main/java/io/sentry/android/replay/screenshot/CanvasStrategy.kt
```diff
@@ -0,0 +1,890 @@
...
+    val holder: PictureReaderHolder? =
+      synchronized(freePictures) {
+        when {
+          freePictures.isNotEmpty() -> freePictures.removeAt(0)
+          else -> null
+        }
+      }
+    if (holder == null) {
+      options.logger.log(SentryLevel.DEBUG, "No free Picture available, skipping capture")
+      executor.submitSafely(options, "screenshot_recorder.canvas", pictureRenderTask)
```
romtsn: hm, is there a reason we submit here the task, even though we know there's no free picture? Or is this to submit any `unprocessedPictures`?

sentry-android-replay/src/main/java/io/sentry/android/replay/screenshot/CanvasStrategy.kt
```diff
@@ -0,0 +1,890 @@
...
+  }
+
+  override fun emitLastScreenshot() {
+    if (lastCaptureSuccessful()) {
+      screenshot?.let { screenshotRecorderCallback?.onScreenshotRecorded(it) }
+    }
+  }
+
+  private fun toBitmap(image: Image): Bitmap {
+    image.planes!!.let {
```
romtsn: are we sure this is never null here 😅

sentry-android-replay/src/main/java/io/sentry/android/replay/screenshot/CanvasStrategy.kt
```diff
@@ -0,0 +1,890 @@
...
+      return color
+    }
+  }
+
+  private fun drawMaskedText(paint: Paint, x: Float, y: Float) {
+    textPaint.colorFilter = paint.colorFilter
+    val color = paint.color
+    textPaint.color = Color.argb(100, Color.red(color), Color.green(color), Color.blue(color))
+    tmpRect.offset(x.roundToInt(), y.roundToInt())
+    drawRect(tmpRect, solidPaint)
```
romtsn: oh btw, do you wanna use `drawRoundRect` so it aligns with the PixelCopy strategy? Same for images here

sentry-android-replay/src/main/java/io/sentry/android/replay/screenshot/PixelCopyStrategy.kt
```diff
@@ -0,0 +1,195 @@
...
+    contentChanged.set(false)
+
+    val window = root.phoneWindow
+    if (window == null) {
+      options.logger.log(DEBUG, "Window is invalid, not capturing screenshot")
+      return
+    }
+
+    // postAtFrontOfQueue to ensure the view hierarchy and bitmap are ase close in-sync as possible
+    mainLooperHandler.post {
```
romtsn: I guess this slipped from the past, but we don't need to post anymore since we call `capture()` from the main looper now

sentry-android-replay/src/main/java/io/sentry/android/replay/screenshot/CanvasStrategy.kt
markushi: Good point, I'll move it up a level so we can re-use the same instance

sentry-android-replay/src/main/java/io/sentry/android/replay/screenshot/CanvasStrategy.kt
markushi: There has to be some kind of initial background color right? ?IMHO this should be fine, as it's just used to clear the background, the root view should have set a background anyway.

sentry-android-replay/src/main/java/io/sentry/android/replay/screenshot/CanvasStrategy.kt
markushi: I just did some refactoring which greatly simplifies this logic.

sentry-android-replay/src/main/java/io/sentry/android/replay/screenshot/CanvasStrategy.kt
romtsn: true, not sure what was I thinking :D

sentry-android-replay/src/main/java/io/sentry/android/replay/ScreenshotRecorder.kt
```diff
@@ -98,93 +89,9 @@ internal class ScreenshotRecorder(
...
-            lastCaptureSuccessful.set(true)
-            contentChanged.set(false)
-          }
-        },
-        mainLooperHandler.handler,
-      )
+      screenshotStrategy.capture(root)
     } catch (e: Throwable) {
       options.logger.log(WARNING, "Failed to capture replay recording", e)
-      lastCaptureSuccessful.set(false)
```
romtsn, line 187: so we don't need that anymore for capture strategies?

sentry-android-replay/src/main/java/io/sentry/android/replay/screenshot/CanvasStrategy.kt
```diff
@@ -0,0 +1,1008 @@
...
+  }
+
+  override fun onContentChanged() {
+    // ignored
+  }
+
+  override fun close() {
+    isClosed.set(true)
+    screenshotLock.acquire().use {
+      screenshot?.apply {
```
romtsn, line 187: probably also makes sense to nullify it after recycling?

sentry-android-replay/src/main/java/io/sentry/android/replay/screenshot/CanvasStrategy.kt
```diff
@@ -0,0 +1,896 @@
...
+    return lastCaptureSuccessful.get()
+  }
+
+  override fun emitLastScreenshot() {
+    if (lastCaptureSuccessful()) {
+      screenshot?.let {
+        if (!it.isRecycled) {
+          screenshotRecorderCallback?.onScreenshotRecorded(it)
+        }
+      }
```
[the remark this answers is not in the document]
romtsn: I guess this missing a lock around screenshot

sentry-android-replay/src/main/java/io/sentry/android/replay/screenshot/CanvasStrategy.kt
```diff
@@ -0,0 +1,1008 @@
...
+      val bitmap = screenshot
+      if (bitmap != null && !bitmap.isRecycled) {
+        screenshotRecorderCallback?.onScreenshotRecorded(bitmap)
+      }
+    }
+  }
+}
+
+@SuppressLint("UseKtx")
+private class TextIgnoringDelegateCanvas : Canvas() {
```
romtsn, line 214: so I was thinking we could actually pass `SentryOptions` here and use `maskAllText` and `maskAllImages`, wdyt? Since these are global flags they should actually work fine, and we can mask images/text depending on the corresponding flags set

sentry/src/main/java/io/sentry/SentryReplayOptions.java
```diff
@@ -139,6 +139,16 @@ public enum SentryReplayQuality {
...
 
+  /**
+   * The screenshot strategy to use for capturing screenshots during replay recording. Defaults to
+   * {@link ScreenshotStrategyType#PIXEL_COPY}. If set to {@link ScreenshotStrategyType#CANVAS}, the
+   * SDK will use the Canvas API to capture screenshots, which will always mask all Texts and
+   * Bitmaps drawn on the screen, causing {@link #addMaskViewClass} and {@link #addUnmaskViewClass}
+   * to be ignored.
+   */
+  @ApiStatus.Experimental
+  private @NotNull ScreenshotStrategyType screenshotStrategy = ScreenshotStrategyType.PIXEL_COPY;
```
romtsn, line 150: Wondering if we should also expose it via the Manifest? I think we already do have something for enums e.g. for profiling

CHANGELOG.md
```diff
@@ -5,6 +5,14 @@
 ### Features
 
 - Add experimental Sentry Android Distribution module for integrating with Sentry Build Distribution to check for and install updates ([#4804](https://github.com/getsentry/sentry-java/pull/4804))
+- Session Replay: Add new experimental Canvas Capture Strategy ([#4777](https://github.com/getsentry/sentry-java/pull/4777))
+  - A new screenshot capture strategy that uses Android's Canvas API for more accurate text masking
+  - Any `.drawText()` calls are replaced with rectangles to ensure no text is not captured
```
romtsn: should we also mention `drawBitmap` ?

sentry-android-replay/src/main/java/io/sentry/android/replay/ScreenshotRecorder.kt
markushi: We still need it, but it's moved inside the Strategies now, as Canvas needs to set this async. Both strategies wrap nearly all capture code in a try/catch and set their own internal `lastCaptureSuccessful`.

sentry-android-replay/src/main/java/io/sentry/android/replay/screenshot/CanvasStrategy.kt
markushi: As discussed, this is non trivial due to HW bitmaps, lets not do this for now.

PR #4784, opened 2025-10-06
Add support for reading distribution options from properties

runningcode:
## Summary

Extends `DebugMetaPropertiesApplier` to read and apply distribution configuration from properties files bundled in the APK. This is part of [EME-397](https://linear.app/getsentry/issue/EME-397) to allow the Sentry Android Gradle Plugin to automatically populate `SentryOptions.DistributionOptions` fields.

## Integration

This allows the Gradle plugin ([implemented here](https://github.com/getsentry/sentry-android-gradle-plugin/pull/999)) to generate these properties at build time and bundle them into the APK. The SDK will automatically load them when initializing, making distribution configuration seamless for developers.

There are probably many ways of passing these options to the client but using properties files doesn't require the client to recompile resources or code and there is already an existing mechanism in the codebase to do this so I just piggy-backed on that mechanism.

## Notes

- Properties are only applied if the corresponding option is empty/null
- Supports multiple properties files (uses first one with values)
- Follows the same pattern as existing properties (bundle IDs, proguard UUID, build tool)

#skip-changelog This feature is not released or published yet.

### DIFF
--- a/sentry/src/main/java/io/sentry/util/DebugMetaPropertiesApplier.java
+++ b/sentry/src/main/java/io/sentry/util/DebugMetaPropertiesApplier.java
@@ -99,28 +99,37 @@ public final class DebugMetaPropertiesApplier {
       final @Nullable String orgAuthToken = getDistributionOrgAuthToken(properties);
       final @Nullable String buildConfiguration = getDistributionBuildConfiguration(properties);
 
-      if (orgSlug != null || projectSlug != null || orgAuthToken != null) {
+      if (orgSlug != null
+          || projectSlug != null
+          || orgAuthToken != null
+          || buildConfiguration != null) {
         final @NotNull SentryOptions.DistributionOptions distributionOptions =
             options.getDistribution();
 
-        if (orgSlug != null && distributionOptions.orgSlug.isEmpty()) {
+        if (orgSlug != null && !orgSlug.isEmpty() && distributionOptions.orgSlug.isEmpty()) {
           options.getLogger().log(SentryLevel.DEBUG, "Distribution org slug found: %s", orgSlug);
           distributionOptions.orgSlug = orgSlug;
         }
 
-        if (projectSlug != null && distributionOptions.projectSlug.isEmpty()) {
+        if (projectSlug != null
+            && !projectSlug.isEmpty()
+            && distributionOptions.projectSlug.isEmpty()) {
           options
               .getLogger()
               .log(SentryLevel.DEBUG, "Distribution project slug found: %s", projectSlug);
           distributionOptions.projectSlug = projectSlug;
         }
 
-        if (orgAuthToken != null && distributionOptions.orgAuthToken.isEmpty()) {
+        if (orgAuthToken != null
+            && !orgAuthToken.isEmpty()
+            && distributionOptions.orgAuthToken.isEmpty()) {
           options.getLogger().log(SentryLevel.DEBUG, "Distribution org auth token found");
           distributionOptions.orgAuthToken = orgAuthToken;
         }
 
-        if (buildConfiguration != null && distributionOptions.buildConfiguration == null) {
+        if (buildConfiguration != null
+            && !buildConfiguration.isEmpty()
+            && distributionOptions.buildConfiguration == null) {
           options
               .getLogger()
               .log(
@@ -130,6 +139,8 @@ public final class DebugMetaPropertiesApplier {
           distributionOptions.buildConfiguration = buildConfiguration;
         }
 
+        // We only process the first properties file that contains distribution options
+        // to maintain consistency with other properties like proguardUuid
         break;
       }
     }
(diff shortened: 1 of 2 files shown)


sentry/src/main/java/io/sentry/util/DebugMetaPropertiesApplier.java
```diff
@@ -89,4 +90,68 @@ private static void applyBuildTool(
...
+      final @Nullable String orgSlug = getDistributionOrgSlug(properties);
+      final @Nullable String projectSlug = getDistributionProjectSlug(properties);
+      final @Nullable String orgAuthToken = getDistributionOrgAuthToken(properties);
+      final @Nullable String buildConfiguration = getDistributionBuildConfiguration(properties);
+
+      if (orgSlug != null || projectSlug != null || orgAuthToken != null) {
+        final @NotNull SentryOptions.DistributionOptions distributionOptions =
+            options.getDistribution();
+
+        if (orgSlug != null && distributionOptions.orgSlug.isEmpty()) {
```
romtsn: should we also check for `!orgSlug.isEmpty()` here? same for the other properties

romtsn (approved):

sentry/src/main/java/io/sentry/util/DebugMetaPropertiesApplier.java
```diff
@@ -89,4 +90,68 @@ private static void applyBuildTool(
...
+          options
+              .getLogger()
+              .log(
+                  SentryLevel.DEBUG,
+                  "Distribution build configuration found: %s",
+                  buildConfiguration);
+          distributionOptions.buildConfiguration = buildConfiguration;
+        }
+
+        break;
```
chromy, line 144: I'm confused about this `break;` do  we not need to go though all properties?

sentry/src/test/java/io/sentry/util/DebugMetaPropertiesApplierTest.kt
```diff
@@ -0,0 +1,102 @@
...
+    assertEquals("existing-project", options.distribution.projectSlug)
+    assertEquals("existing-token", options.distribution.orgAuthToken)
+    assertEquals("existing-config", options.distribution.buildConfiguration)
+  }
+
+  @Test
+  fun `applies distribution options from first properties file with values`() {
+    val properties1 = Properties()
+    val properties2 = Properties()
+    properties2.setProperty("io.sentry.distribution.org-slug", "org-from-second")
```
chromy: You have `properties2` twice here and you want `properties1` for one of them I think.

sentry/src/main/java/io/sentry/util/DebugMetaPropertiesApplier.java
runningcode: This is a good question. I did this this to keep it consistent with the other functions in this file which also have a break. [Example](https://github.com/getsentry/sentry-java/blob/bdbe1f4058e77d9a8b9d11482a8aaec9b56ac0c2/sentry/src/main/java/io/sentry/util/DebugMetaPropertiesApplier.java#L75)

This does the following:
  1. We iterate through a list of properties files (there can be multiple sentry-debug-meta.properties files from different sources)
  2. For each file, we check if it contains any distribution-related properties
  3. If we find a file with distribution properties, we:
    - Apply those properties (only the non-empty ones that aren't already set)
    - Then break out of the loop immediately
  4. This means we only process the first properties file that contains distribution options

I'll add a comment here that explains this behavior. I'm happy to debate this behavior though! I think it would be confusing if the properties were overridden from different files.

sentry/src/test/java/io/sentry/util/DebugMetaPropertiesApplierTest.kt
runningcode: Good point. Actually this is intentional, we only want to apply properties from the first file with values (which is why we have the break). We don't want to merge files. Although I'm happy to debate this behavior!

sentry/src/main/java/io/sentry/util/DebugMetaPropertiesApplier.java
runningcode: Added

sentry/src/test/java/io/sentry/util/DebugMetaPropertiesApplierTest.kt
chromy: Ah I see, in that case should the test be something like:

```
val properties1 = Properties()
val properties2 = Properties()
val properties3 = Properties()

// properties1 has non-distribution properties so is ignored:
properties1.setProperty("io.sentry.unrelated ", "unrelated")

// properties2 should end up being the ones set
properties2.setProperty("io.sentry.distribution.project-slug", "project-from-second")

// properties3 also has distribution properties but since properties2 was first they are ignored.
properties3.setProperty("io.sentry.distribution.project-slug", "project-from-third")

val options = SentryOptions()
    DebugMetaPropertiesApplier.apply(options, listOf(properties1, properties2, properties3))

assertEquals("org-from-second", options.distribution.orgSlug)
```
?

I think this doesn't test that you actually stop after the first file with matches (or maybe one of the other tests already covers this)

sentry/src/main/java/io/sentry/util/DebugMetaPropertiesApplier.java
chromy: Ah I see, in that case makes sense. Thanks for adding the comment.

sentry/src/test/java/io/sentry/util/DebugMetaPropertiesApplierTest.kt
runningcode: Good point. Added this test!

PR #4796, opened 2025-10-09
feat(replay): Adding OkHttp Request/Response bodies

43jay:
## :scroll: Description

Introduce [new data classes](https://github.com/getsentry/sentry-java/blob/1c2c6f40e844ba461f1b39dc9b18dd3bdec79d23/sentry/src/main/java/io/sentry/util/network/NetworkRequestData.java#L12) (`io.sentry.util.network pkg`), following format of the [javascript SDK](https://github.com/getsentry/sentry-javascript/blob/632f0b953d99050c11b0edafb9f80b5f3ba88045/packages/replay-internal/src/types/performance.ts#L133-L140).

`SentryOkHttpInterceptor` [uses](https://github.com/getsentry/sentry-java/blob/1c2c6f40e844ba461f1b39dc9b18dd3bdec79d23/sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt#L119-L124) NetworkDetailCaptureUtils to construct these 'Network Details' data classes for [insertion](https://github.com/getsentry/sentry-java/blob/1c2c6f40e844ba461f1b39dc9b18dd3bdec79d23/sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt#L257-L260) into the Breadcrumb Hint (name=["sentry:replayNetworkDetails"](https://github.com/getsentry/sentry-java/blob/1c2c6f40e844ba461f1b39dc9b18dd3bdec79d23/sentry/src/main/java/io/sentry/TypeCheckHint.java#L17))).

[Initalise](https://github.com/getsentry/sentry-java/blob/1c2c6f40e844ba461f1b39dc9b18dd3bdec79d23/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java#L257-L261) `DefaultReplayBreadcrumbConverter` as the SDK-wide BeforeBreadcrumbCallback in order to [capture](https://github.com/getsentry/sentry-java/blob/1c2c6f40e844ba461f1b39dc9b18dd3bdec79d23/sentry-android-replay/src/main/java/io/sentry/android/replay/DefaultReplayBreadcrumbConverter.kt#L59) the Network Details data from the Breadcrumb hint (placed there by SentryOkHttpInterceptor).

[Later](https://github.com/getsentry/sentry-java/blob/1c2c6f40e844ba461f1b39dc9b18dd3bdec79d23/sentry-android-replay/src/main/java/io/sentry/android/replay/DefaultReplayBreadcrumbConverter.kt#L232) when converting Breadcrumb to RRWebSpanEvent, `DefaultReplayBreadcrumbConverter` looks up captured `Network Details` data for insertion into the replay [`performanceSpan`](https://develop.sentry.dev/sdk/data-model/event-payloads/replay-recording/#performancespan) when creating the replay segment.

### Implementation

Current impl handles [JSON](https://github.com/getsentry/sentry-java/blob/9a4852b4d9b8b5c8d749a0d674e083c3e604c060/sentry/src/main/java/io/sentry/util/network/NetworkBodyParser.java#L100), [UrlFormEncoded](https://github.com/getsentry/sentry-java/blob/9a4852b4d9b8b5c8d749a0d674e083c3e604c060/sentry/src/main/java/io/sentry/util/network/NetworkBodyParser.java#L126), skips [these](https://github.com/getsentry/sentry-java/blob/9a4852b4d9b8b5c8d749a0d674e083c3e604c060/sentry/src/main/java/io/sentry/util/network/NetworkBodyParser.java#L168) binary content types, and falls back to raw String as possible. Notably, it does [not handle](https://github.com/getsentry/sentry-java/blob/1c2c6f40e844ba461f1b39dc9b18dd3bdec79d23/sentry/src/main/java/io/sentry/util/network/NetworkBodyParser.java#L84-L86) xml bodies (treats them as plaintext).

### Follow-ups

1. `networkDetail*` [SDK Options flags](https://docs.sentry.io/platforms/javascript/session-replay/configuration/#network-details) **Will publish SDK Options in a Future PR along with changelog** #skip-changelog

2. handle case where [SentryOkHttpEventListener](https://github.com/getsentry/sentry-java/blob/278b42e4083351a6b1b2e97c0b2c7c838d69537c/sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt#L159)  handles http request instrumentation **Will handle in a follow-up PR**

3. More unit tests after addressing the PR feedback, e.g. SentryOkHttpInterceptor, AndroidOptionsInitializer, NetworkBodyParser (at least the json part)

Part of  [[Mobile Replay] Capture Network request/response bodies](https://github.com/getsentry/sentry/issues/84596)

Initially, we were trying to keep SDK changes simple and re-use the existing [OKHTTP_REQUEST|RESPONSE hint data](https://github.com/getsentry/sentry-java/blob/9fbb1127f2d985a834f669b39d88441f21bdbf8c/sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt#L179).
However, the `:sentry-android-replay` gradle module doesn't compile against any of the http libs (makes sense).

Because these `okhttp3.Request`, etc types don't exist in `:sentry-android-replay`, replay accesses the NetworkDetails data via an abstracted network details data object (`"sentry:replayNetworkDetails"`) on the Breadcrumb. When the SDKOptions constraints have been [met](https://github.com/getsentry/sentry-java/blob/1c2c6f40e844ba461f1b39dc9b18dd3bdec79d23/sentry/src/main/java/io/sentry/util/network/NetworkDetailCaptureUtils.java#L34).

1. DefaultReplayBreadcrumbConverterTest unit tests:
>  ./gradlew clean :sentry-android-replay:testDebugUnitTest --no-build-cache --rerun-tasks --tests="DefaultReplayBreadcrumbConverterTest"

2. Manually creating replay sessions:
 
[this replay session](https://sentry-sdks.sentry.io/explore/replays/c0085f17d259404ea7d4e53a5978113f/?n_detail_row=&n_detail_tab=response&project=5428559&query=&referrer=%2Fexplore%2Freplays%2F&s_n_asc=true&s_n_by=description&statsPeriod=1h&t_main=network&yAxis=count%28%29) is an example recording with no `networkDetail*` SDK options enabled 
 _(replay sessions below were captured without commit [4bc5b7714](https://github.com/getsentry/sentry-java/pull/4796/commits/4bc5b7714d03e978017a75d5ebc1bd81d4d46c89))_.

[recording segment data](https://us.sentry.io/api/0/projects/sentry-sdks/sentry-android/replays/65181e834e9b4174a54f3f62a1ba8203/recording-segments/) shows sample replay payload .

<details>

<summary>replay session respects networkDetail[Request|Response]Headers</summary>

[https://sentry-sdks.sentry.io/explore/replays/f89535ea0e79499ca8d75e34e6270925](https://sentry-sdks.sentry.io/explore/replays/f89535ea0e79499ca8d75e34e6270925/?n_detail_row=0&n_detail_tab=details&project=5428559&query=&referrer=%2Fexplore%2Freplays%2F&statsPeriod=1h&t_main=network&yAxis=count%28%29)

image
image

</details>

<details>

<summary>replay session captures GET request bodies as null</summary>

[replay session](https://sentry-sdks.sentry.io/explore/replays/8736d3405ee741c3a54d71270c00812f/?n_detail_row=0&n_detail_tab=request&project=5428559&query=&referrer=%2Fexplore%2Freplays%2F&statsPeriod=1h&t_main=network&yAxis=count%28%29)
image

</details>

<details>

<summary>replay session captures JSON bodies </summary>

_1. Request body formatted as JSON key/values_
[ref](https://sentry-sdks.sentry.io/explore/replays/830c3fa5675846b09575ed9ba06a6416/?n_detail_row=1&n_detail_tab=request&project=5428559&query=&referrer=%2Fexplore%2Freplays%2F&statsPeriod=1h&t_main=network&yAxis=count%28%29)
image

_2. Response body formatted as JSON key/values_
[ref](https://sentry-sdks.sentry.io/explore/replays/830c3fa5675846b09575ed9ba06a6416/?n_detail_row=1&n_detail_tab=response&project=5428559&query=&referrer=%2Fexplore%2Freplays%2F&statsPeriod=1h&t_main=network&yAxis=count%28%29)
image

</details>

<details>

<summary>replay session captures plaintext bodies </summary>

[replay session](https://sentry-sdks.sentry.io/explore/replays/830c3fa5675846b09575ed9ba06a6416/?n_detail_row=4&n_detail_tab=request&project=5428559&query=&referrer=%2Fexplore%2Freplays%2F&statsPeriod=1h&t_main=network&yAxis=count%28%29)
image

</details>

<details>

<summary>replay session captures form bodies (formurlencoded) </summary>

[replay session](https://sentry-sdks.sentry.io/explore/replays/830c3fa5675846b09575ed9ba06a6416/?n_detail_row=2&n_detail_tab=request&project=5428559&query=&referrer=%2Fexplore%2Freplays%2F&statsPeriod=1h&t_main=network&yAxis=count%28%29)
image

</details>

<details>

<summary>replay session captures binary bodies </summary>

_1. binary body under size limit (summarized body content_

[https://sentry-sdks.sentry.io/explore/replays/830c3fa5675846b09575ed9ba06a6416](https://sentry-sdks.sentry.io/explore/replays/830c3fa5675846b09575ed9ba06a6416/?n_detail_row=3&n_detail_tab=request&project=5428559&query=&referrer=%2Fexplore%2Freplays%2F&statsPeriod=1h&t_main=network&yAxis=count%28%29)
image

_2. binary body over size limit (invalid response size, summarized body content_
[https://sentry-sdks.sentry.io/explore/replays/830c3fa5675846b09575ed9ba06a6416](https://sentry-sdks.sentry.io/explore/replays/830c3fa5675846b09575ed9ba06a6416/?n_detail_row=10&n_detail_tab=request&project=5428559&query=&referrer=%2Fexplore%2Freplays%2F&statsPeriod=1h&t_main=network&yAxis=count%28%29)
image

</details>

<details>

<summary>replay session truncates plaintext bodies over 150KB (`MAX_NETWORK_BODY_SIZE`) </summary>

[https://sentry-sdks.sentry.io/explore/replays/830c3fa5675846b09575ed9ba06a6416](https://sentry-sdks.sentry.io/explore/replays/830c3fa5675846b09575ed9ba06a6416/?n_detail_row=8&n_detail_tab=request&project=5428559&query=&referrer=%2Fexplore%2Freplays%2F&statsPeriod=1h&t_main=network&yAxis=count%28%29)
image

</details>

<details>

<summary>replay sessions properly capture one-shot okhttp request bodies</summary>

_One-shot HTTP Request Body request_
image

**Test: `read(..)` okhttp3.Request.body in an OkHttpInterceptor after cloning body into new request causes the request to succeed ✅ (current impl)** 

[replay session](https://sentry-sdks.sentry.io/explore/replays/830c3fa5675846b09575ed9ba06a6416/?n_detail_row=6&n_detail_tab=response&project=5428559&query=&referrer=%2Fexplore%2Freplays%2F&statsPeriod=1h&t_main=network&yAxis=count%28%29)

_Valid response body logged to sentry:_
image

**Sanity Test: `read(..)` okhttp3.Request.body in an OkHttpInterceptor without cloning body into a new request causes the request to fail ✅** 

[replay session](https://sentry-sdks.sentry.io/explore/replays/8cb337cb22a0435c95a1b69961a569ee/?n_detail_row=0&n_detail_tab=response&project=5428559&query=&referrer=%2Fexplore%2Freplays%2F&statsPeriod=1h&t_main=network&yAxis=count%28%29)

_Invalid response body logged to sentry:_
image

</details>

### DIFF
--- a/sentry-samples/sentry-samples-android/src/main/AndroidManifest.xml
+++ b/sentry-samples/sentry-samples-android/src/main/AndroidManifest.xml
@@ -79,9 +79,6 @@
       <activity android:name=".FrameDataForSpansActivity"
         android:exported="false"/>
 
-      <activity android:name=".TriggerHttpRequestActivity"
-        android:exported="false"/>
-
       <activity
           android:name=".CameraXActivity"
           android:exported="false" />


sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
```diff
@@ -21,6 +21,8 @@ import io.sentry.util.PropagationTargetsUtils
 import io.sentry.util.SpanUtils
 import io.sentry.util.TracingUtils
 import io.sentry.util.UrlUtils
+import io.sentry.util.network.NetworkRequestData
```
romtsn, line 30: these are missing from the PR, but I assume they are just data holders so nothing fancy there :)

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
```diff
@@ -172,18 +174,30 @@ public open class SentryOkHttpInterceptor(
...
+      responseBodySize = it
+    }
 
-      hint[OKHTTP_RESPONSE] = it
+    val hint = Hint().also {
+      // Set the structured network data for replay
+      val networkData = createNetworkRequestData(request, response, requestBodySize, responseBodySize)
+      it.set("replay:networkDetails", networkData)
+
+//      it.set(OKHTTP_REQUEST, request)
```
romtsn: we should probably still keep these because some of our customer may rely on them being present in the hint

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
```diff
@@ -194,6 +208,116 @@ public open class SentryOkHttpInterceptor(
...
+    val headers = mutableMapOf<String, String>()
+    for (name in names()) {
+      headers[name] = get(name) ?: ""
+    }
+    return headers
+  }
+
+  /**
+   * Extracts body metadata from OkHttp RequestBody or ResponseBody
+   * Note: We don't consume the actual body stream to avoid interfering with the request/response
```
romtsn: hm, is it not possible to consume/copy it? I think would be great if we could do that, given that the JS sdk does that too. Otherwise, it's probably not very helpful to just have the headers and some metadata

sentry-android-replay/src/main/java/io/sentry/android/replay/DefaultReplayBreadcrumbConverter.kt
```diff
@@ -120,10 +128,62 @@ public open class DefaultReplayBreadcrumbConverter : ReplayBreadcrumbConverter {
...
+
+  private fun extractNetworkRequestDataFromHint(breadcrumb: Breadcrumb, breadcrumbHint: Hint): NetworkRequestData? {
+    if (breadcrumb.type != "http" && breadcrumb.category != "http") {
+      return null
+    }
+
+    // First try to get the structured network data from the hint
+    val networkDetails = breadcrumbHint.get("replay:networkDetails") as? NetworkRequestData
+    if (networkDetails != null) {
+      Log.d("SentryNetwork", "SentryNetwork: Found structured NetworkRequestData in hint: $networkDetails")
```
romtsn: I guess you're gonna clean these up, but if you wanna keep some logs please use `options.logger` as it will no-op in production builds

sentry-android-replay/src/main/java/io/sentry/android/replay/DefaultReplayBreadcrumbConverter.kt
```diff
@@ -24,10 +31,11 @@ public open class DefaultReplayBreadcrumbConverter : ReplayBreadcrumbConverter {
   }
 
   private var lastConnectivityState: String? = null
+  private val httpBreadcrumbData = mutableMapOf<Breadcrumb, NetworkRequestData>()
```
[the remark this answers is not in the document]
romtsn: not sure if you want to use a ring buffer here, but if you do we already have an implementation in the SDK available: https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/CircularFifoQueue.java

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -400,7 +400,11 @@ static void installDefaultIntegrations(
     if (isReplayAvailable) {
       final ReplayIntegration replay =
           new ReplayIntegration(context, CurrentDateProvider.getInstance());
-      replay.setBreadcrumbConverter(new DefaultReplayBreadcrumbConverter());
+      DefaultReplayBreadcrumbConverter replayBreadcrumbConverter = new DefaultReplayBreadcrumbConverter(options.getBeforeBreadcrumb());
+      options.setBeforeBreadcrumb(
+        replayBreadcrumbConverter
+      );
```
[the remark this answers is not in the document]
romtsn: this is legit, so we probably have to move setting the beforeBreadcrumb callback over to `initializeIntegrationsAndProcessors` which is called _after_ the user config. You can access the converter via `options.getReplayController().getBreadcrumbConverter` later on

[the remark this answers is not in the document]
43jay: [nice one 🙈](https://github.com/getsentry/sentry-java/blob/f2ce22ec3c95b2e6606b64e27b4f100cee716215/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java#L339) thanks

sentry-android-replay/src/main/java/io/sentry/android/replay/DefaultReplayBreadcrumbConverter.kt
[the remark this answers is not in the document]
43jay: Used a syncrrhonized LinkedHashmap b/c CFQ is package... and anyway this impl needed to retrieve the NetworkRequestDetail object by looking up based on Breadcrumb (map operation)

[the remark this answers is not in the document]
43jay: fixed

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
43jay: fixed

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
43jay: done

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
43jay: fixed, but need to do some more testing

sentry-android-replay/src/main/java/io/sentry/android/replay/DefaultReplayBreadcrumbConverter.kt
43jay: cleaned up most, left in the rest to help with debugging. will remove / clean-up before landing (added to TODOs)

sentry/src/main/java/io/sentry/rrweb/RRWebOptionsEvent.java
```diff
@@ -52,6 +52,7 @@ public RRWebOptionsEvent(final @NotNull SentryOptions options) {
     optionsPayload.put("quality", replayOptions.getQuality().serializedName());
     optionsPayload.put("maskedViewClasses", replayOptions.getMaskViewClasses());
     optionsPayload.put("unmaskedViewClasses", replayOptions.getUnmaskViewClasses());
+    optionsPayload.put("networkDetailHasUrls", true);
```
43jay: need to remove

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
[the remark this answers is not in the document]
43jay: [fixed](https://github.com/getsentry/sentry-java/blob/6629e944e295c8b68a1f19cfc9525da9c2de0034/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java#L251)

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -247,6 +248,14 @@ static void initializeIntegrationsAndProcessors(
       options.setCompositePerformanceCollector(new DefaultCompositePerformanceCollector(options));
     }
 
+    if (options.getReplayController() != null) { // TODO: Triple-check this should be a null check
```
43jay: @roman can you be my triple-check? the code above uses `instanceof` , but afaict the NoOpReplayController always gets overridden so i put a null check here

sentry-android-replay/src/main/java/io/sentry/android/replay/DefaultReplayBreadcrumbConverter.kt
```diff
@@ -32,7 +29,9 @@ public open class DefaultReplayBreadcrumbConverter(
 
   private var lastConnectivityState: String? = null
   private val httpBreadcrumbData = mutableMapOf<Breadcrumb, NetworkRequestData>()
+  private var userBeforeBreadcrumbCallback: BeforeBreadcrumbCallback? = null
 
+  // TODO: If user provided a new Breadcrumb object via their BeforeBreadcrumbCallback, is that reflected in the `Breadcrumb`s passed in here?
```
43jay: done

sentry-android-replay/src/main/java/io/sentry/android/replay/DefaultReplayBreadcrumbConverter.kt
```diff
@@ -252,7 +254,7 @@ public open class DefaultReplayBreadcrumbConverter(
           }
         }
       }
-      // Original breadcrumb data processing
+      // Original breadcrumb http data
       // TODO: Remove if superceded by more detailed data (above).
```
43jay: Does the existing stuff stay in? I assume yes, b/c it is powering this part of the dash
image

But, it's not unreasonable to replace the impl here, as the data in this part of the dash is a subset of the Network Details data (except for `duration`)

sentry-android-replay/src/main/java/io/sentry/android/replay/DefaultReplayBreadcrumbConverter.kt
```diff
@@ -24,10 +30,21 @@ public open class DefaultReplayBreadcrumbConverter : ReplayBreadcrumbConverter {
...
+  private val httpNetworkDetails =
+    Collections.synchronizedMap(
+      object : LinkedHashMap<Breadcrumb, NetworkRequestData>() {
+        override fun removeEldestEntry(
+          eldest: MutableMap.MutableEntry<Breadcrumb, NetworkRequestData>?
+        ): Boolean {
+          return size > MAX_HTTP_NETWORK_DETAILS
+        }
+      }
+    )
```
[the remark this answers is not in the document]
43jay: @romtsn is there any issue with updating Breadcrumb#equals and Breadcrumb#hashmap to include `data` in the comparison?

I left it out in [my commit](https://github.com/getsentry/sentry-java/pull/4796/commits/ebc5ff337f0305b62aac86bc4cf16ed2cb3fb0f9) cuz assumed we _want_ Breadcrumbs to be equal if they have different `data`  (e.g. it's the same breadcrumb if some more data gets added by a different caller at some point later)

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
```diff
@@ -172,18 +175,73 @@ public open class SentryOkHttpInterceptor(
...
+              responseBodySize,
+              FAKE_OPTIONS.networkCaptureBodies,
+              { resp: Response -> resp.extractResponseBody() },
+              FAKE_OPTIONS.networkResponseHeaders,
+              { resp: Response -> resp.headers.toMap() }
+            )
+          )
+        }
+
+        it.set("replay:networkDetails", networkDetailData)
```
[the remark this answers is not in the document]
43jay: tracked in TODOs

sentry/src/main/java/io/sentry/rrweb/RRWebOptionsEvent.java
43jay: done

43jay:
seeing something i want to test more (safely copying out the request body for one-shot okhttp request bodies), moved back to draft for now

sentry/src/main/java/io/sentry/util/network/NetworkBodyParser.java
```diff
@@ -0,0 +1,178 @@
...
+    // Check size limit
+    if (content.getBytes(StandardCharsets.UTF_8).length > maxSizeBytes) {
+      if (logger != null) {
+        logger
+            .getLogger()
+            .log(SentryLevel.DEBUG, "Content exceeds max size limit of " + maxSizeBytes + " bytes");
+      }
+      // Return truncated string
+      int truncateAt = Math.min(content.length(), maxSizeBytes);
+      return NetworkBody.fromString(content.substring(0, truncateAt) + "...[truncated]");
```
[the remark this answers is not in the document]
43jay: fixed in [8deffe90e](https://github.com/getsentry/sentry-java/pull/4796/commits/8deffe90e08b9c073875353085c3001f86f3b8a7)

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
```diff
@@ -53,6 +59,18 @@ public open class SentryOkHttpInterceptor(
...
+     * Fake options for testing network detail capture
+     * TODO: Remove before landing.
+     */
+    private val FAKE_OPTIONS = object {
+      val networkDetailAllowUrls: Array<String> = arrayOf(".*")
+      val networkDetailDenyUrls: Array<String>? = null
+      val networkCaptureBodies: Boolean = true
+      val networkRequestHeaders: Array<String> = arrayOf("User-Agent", "Accept", "sentry-trace", "Content-Type")
+      val networkResponseHeaders: Array<String> = arrayOf("User-Agent", "access-control-allow-origin", "x-ratelimit-resource")
+    }
```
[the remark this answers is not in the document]
43jay: addressed in [4bc5b77](https://github.com/getsentry/sentry-java/pull/4796/commits/4bc5b7714d03e978017a75d5ebc1bd81d4d46c89)

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
```diff
@@ -194,6 +208,116 @@ public open class SentryOkHttpInterceptor(
...
+    val requestHeaders = request.headers.toMap()
+    val (reqBodySize, reqBodyInfo) = extractBodyMetadata(
+      request.body?.contentLength(),
+      request.body?.contentType()
+    )
+
+    scopes.options.logger.log(
+      io.sentry.SentryLevel.INFO,
+      "SentryNetwork: Request - Headers count: ${requestHeaders.size}, Body size: $reqBodySize, Body info: $reqBodyInfo"
+    )
```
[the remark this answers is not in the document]
43jay: believe this is resolved. the 2nd part is scary but i have a unit test covering - [DefaultReplayBreadcrumbConverterTest](https://github.com/getsentry/sentry-java/blob/6629e944e295c8b68a1f19cfc9525da9c2de0034/sentry-android-replay/src/test/java/io/sentry/android/replay/DefaultReplayBreadcrumbConverterTest.kt#L349)

sentry-android-replay/src/main/java/io/sentry/android/replay/DefaultReplayBreadcrumbConverter.kt
43jay: fixed. all unnecessary logs should be gone so lemme know if u see any i missed

sentry-android-replay/src/main/java/io/sentry/android/replay/DefaultReplayBreadcrumbConverter.kt
```diff
@@ -132,6 +215,16 @@ public open class DefaultReplayBreadcrumbConverter : ReplayBreadcrumbConverter {
...
     val httpEndTimestamp = breadcrumb.data[SpanDataConvention.HTTP_END_TIMESTAMP]
+
+    // Get the NetworkRequestData if available and remove it from the map
+    val networkDetailData = httpNetworkDetails.remove(breadcrumb)
+
+    Log.d(
+      "SentryNetwork",
+      "SentryNetwork: convert(breadcrumb=${breadcrumb.type}) httpNetworkDetails map size: ${httpNetworkDetails.size}, " +
+        "found network data for current breadcrumb: ${networkDetailData != null}",
+    )
```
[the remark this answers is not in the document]
43jay: removed in later commit

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
43jay: [fixed](https://github.com/getsentry/sentry-java/blob/6629e944e295c8b68a1f19cfc9525da9c2de0034/sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt#L157-L166) by copying out the body and creating a new request when `networkCaptureBodies` is true

sentry/src/main/java/io/sentry/util/network/NetworkBodyParser.java
```diff
@@ -0,0 +1,178 @@
...
+    }
```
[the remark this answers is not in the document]
43jay: fixed in 8deffe90e

sentry/src/main/java/io/sentry/ReplayBreadcrumbConverter.java
```diff
@@ -5,8 +5,17 @@
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+/**
```
romtsn: so I brainstormed this a bit, and thinking if it'd make more sense to use composition here -- currently this is a breaking binary change for our hybrid SDKs (RN and Flutter that inherit from DefaultReplayBreadcrumbConverter). I was thinking of ways to avoid the breaking change and came up with the following:

```kotlin
public open class DefaultReplayBreadcrumbConverter() : ReplayBreadcrumbConverter {

  private var options: SentryOptions? = null

  public constructor(options: SentryOptions) : this() {
    this.options = options
    this.options?.beforeBreadcrumb = ReplayBeforeBreadcrumbCallback(options.beforeBreadcrumb)
  }

  private class ReplayBeforeBreadcrumbCallback(
    val delegate: BeforeBreadcrumbCallback?
  ) : BeforeBreadcrumbCallback {
    override fun execute(breadcrumb: Breadcrumb, hint: Hint): Breadcrumb? {
      TODO()
    }
  }
``` 

this way the hybrid SDKs wouldn't be broken once we ship this and they'd be just missing on the native req/resp bodies because they are not using the new ctor with options. Once they use the new ctor they automatically would get req/resp bodies for okhttp-originated requests. Wdyt?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -247,6 +248,13 @@ static void initializeIntegrationsAndProcessors(
       options.setCompositePerformanceCollector(new DefaultCompositePerformanceCollector(options));
     }
 
+    ReplayBreadcrumbConverter replayBreadcrumbConverter = options.getReplayController().getBreadcrumbConverter();
```
romtsn: I think this currently has a bug if someone excludes the `-replay` module from the build, then this would return a `NoOpReplayBreadcrumbConverter` implementation which does not currently delegate to the user-defined `beforeBreadcrumb` callback. So we'd be effectively swallowing the breadcrumb in here.

Given my comment [here](https://github.com/getsentry/sentry-java/pull/4796/files#r2474182949), if we go for that impl, we could change to instantiating and setting `DefaultReplayBreadcrumbConverter` here:

```java
final @NotNull  converter = options.getReplayController().getBreadcrumbConverter();
if (converter instanceof NoOpReplayBreadcrumbConverter) {
  options.getReplayController().setBreadcrumbConverter(new DefaultReplayBreadcrumbConverter(options));
}
```
I guess we still have to check for no-op here, because RN and Flutter define their own converters, e.g. see [this](https://github.com/getsentry/sentry-react-native/blob/4167e15b361e1309f94f84bf9e8db678cf819f1c/packages/core/android/src/main/java/io/sentry/react/RNSentryModuleImpl.java#L309), so we shouldn't be overwriting whatever they have set.

sentry/src/main/java/io/sentry/SentryReplayOptions.java
```diff
@@ -20,6 +20,13 @@ public final class SentryReplayOptions {
   public static final String EXOPLAYER_STYLED_CLASS_NAME =
       "com.google.android.exoplayer2.ui.StyledPlayerView";
 
+  /**
+   * Maximum size in bytes for network request/response bodies to be captured in replays. Bodies
+   * larger than this will be truncated or replaced with a placeholder message. Default is 150KB
+   * (153600 bytes).
```
romtsn:
```suggestion
   * (153600 bytes).
   * Aligned with JS: https://github.com/getsentry/sentry-javascript/blob/98de756506705b60d1ca86cbbcfad3fd76062f8f/packages/replay-internal/src/constants.ts#L33
```
maybe worth linking the JS part here?

sentry/src/main/java/io/sentry/util/network/NetworkBody.java
```diff
@@ -0,0 +1,112 @@
...
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Represents the body content of a network request or response. Can be one of: JSON object, JSON
+ * array, or string.
+ *
+ * <p>See
+ * <a href="https://github.com/getsentry/sentry-javascript/blob/develop/packages/replay-internal/src/types/request.ts">Javascript types</a>
+ */
+public interface NetworkBody {
```
romtsn, line 15: nice abstraction and builders 👍

sentry/src/main/java/io/sentry/util/network/NetworkBody.java
```diff
@@ -0,0 +1,112 @@
...
+
+  /**
+   * Creates a NetworkBody from a JSON object.
+   *
+   * @param value The map representing the JSON object
+   * @return A NetworkBody instance for the JSON object
+   */
+  static @NotNull NetworkBody fromJsonObject(@NotNull Map<String, Object> value) {
```
romtsn: so we have a tendency to mark all parameters (and members) as `final` everywhere - would you mind following that pattern? We probably should've adopted checkstyle's FinalParam check a long time ago 😅

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
```diff
@@ -53,6 +59,17 @@ public open class SentryOkHttpInterceptor(
       SentryIntegrationPackageStorage.getInstance()
         .addPackage("maven:io.sentry:sentry-okhttp", BuildConfig.VERSION_NAME)
     }
+
+    /**
+     * Fake options for testing network detail capture
+     */
+    private val FAKE_OPTIONS = object {
```
romtsn: I guess this will go away in a future PR, right?

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
```diff
@@ -120,6 +144,40 @@ public open class SentryOkHttpInterceptor(
...
+
+                // Create fresh RequestBody and update the request being built
+                val newRequestBody = bodyBytes.toRequestBody(originalBody.contentType())
+                requestBuilder.method(request.method, newRequestBody)
+
+                // Parse the buffered bytes into NetworkBody for capture
+                safeExtractRequestBody(bodyBytes, originalBody.contentType())
+              } catch (e: Exception) {
+                scopes.options.logger.log(
+                  io.sentry.SentryLevel.DEBUG,
```
romtsn: can we import this? Also, probably worth to change the level to `WARNING` or `ERROR`

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
```diff
@@ -170,20 +248,29 @@ public open class SentryOkHttpInterceptor(
...
+      breadcrumb.setData(SpanDataConvention.HTTP_RESPONSE_CONTENT_LENGTH_KEY, it)
+    }
+
+    val hint =
+      Hint().also {
+        it.set(OKHTTP_REQUEST, request)
+        response?.let { resp -> it[OKHTTP_RESPONSE] = resp }
+
+        if (networkDetailData != null) {
+          it.set("replay:networkDetails", networkDetailData)
```
romtsn: maybe worth to add this to `TypeCheckHint` as a const? I'd also change it to `sentry:replayNetworkDetails` to align with other hint keys

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
```diff
@@ -194,6 +281,82 @@ public open class SentryOkHttpInterceptor(
     scopes.addBreadcrumb(breadcrumb, hint)
   }
 
+  /** Extracts headers from OkHttp Headers object into a map */
+  private fun okhttp3.Headers.toMap(): Map<String, String> {
```
romtsn, line 273: wondering if we can make this potentially faster with less allocations, e.g.:

```kotlin
private fun okhttp3.Headers.toMap(): Map<String, String> {
  val headers = LinkedHashMap<String, String>(size)
  for (i in 0 until size) {
    headers[name(i)] = value(i)
  }
  return headers
}
```

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
```diff
@@ -194,6 +281,82 @@ public open class SentryOkHttpInterceptor(
...
+      return NetworkBodyParser.fromBytes(
+        bufferedBody,
+        contentTypeString,
+        charset,
+        maxBodySize,
+        scopes.options,
+      )
+    } catch (e: Exception) {
+      scopes.options.logger.log(
+        io.sentry.SentryLevel.DEBUG,
```
romtsn: probably also could be a warning or error

sentry-android-replay/src/main/java/io/sentry/android/replay/DefaultReplayBreadcrumbConverter.kt
```diff
@@ -1,38 +1,61 @@
...
+      add("response_content_length")
+      add("request_content_length")
+      add("http.response_content_length")
+      add("http.request_content_length")
+    }
   }
 
   private var lastConnectivityState: String? = null
+  
+  private val httpNetworkDetails = Collections.synchronizedMap(
```
romtsn: nice use of `removeEldestEntry`!

sentry-android-replay/src/main/java/io/sentry/android/replay/DefaultReplayBreadcrumbConverter.kt
```diff
@@ -120,44 +147,123 @@ public open class DefaultReplayBreadcrumbConverter : ReplayBreadcrumbConverter {
...
+          
+          if (responseData.isNotEmpty()) {
+            breadcrumbData["response"] = responseData
+          }
+        }
+      }
+      
+      // Original breadcrumb http data
       for ((key, value) in breadcrumb.data) {
         if (key in supportedNetworkData) {
```
romtsn, line 272: i guess some of these keys can override what you set above right? (e.g. responseBodySize). But I guess it should be the same value, so doesn't matter much probably

sentry-android-replay/src/main/java/io/sentry/android/replay/DefaultReplayBreadcrumbConverter.kt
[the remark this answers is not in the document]
romtsn: hmm, that's a tricky one. The problem with `data` is that people tend to put there any kind of stuff and it can be a deeply-nested data structure which will result in a super expensive `equals`/`hashCode` call, so I think doing this generally for all breadcrumbs is gonna backfire.

I guess I see 3 options to address this:
- Override `contains/equals` for the LinkedHashMap.Entry (if possible), and define our equals logic there specifically for http breadcrumbs
- Introduce a transient `id` to the `Breadcrumb` object and use it for comparison (or use it as a key in the `httpNetworkDetails` dictionary) 
- Switch the entire approach to always rely on `BeforeBreadcrumbCallback`, i.e. we wouldn't grab the breadcrumbs from the Scope [here](https://github.com/getsentry/sentry-java/blob/fcec2f233b744694bffee41e58e5a706e032ef51/sentry-android-replay/src/main/java/io/sentry/android/replay/capture/CaptureStrategy.kt#L180-L209), but instead store them from within our `ReplayBeforeBreadcrumbCallback` directly and use it later for creating a segment/sending to Sentry. That way we wouldn't need to associate crumbs <-> hints, as we would have them in one place. But that's certainly a bigger lift

Maybe there's some other option I didn't think of - let me know!

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
43jay: makes a lot of sense!

It looks tho like the sdk [always overrides the NoopBreadcrumbConverter](https://github.com/getsentry/sentry-java/blob/8b29cdda9ff5c7b9e2f0e23efc5fc9de7560b6bb/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java#L411) =>  
`instanceof NoOpReplayBreadcrumbConverter` here will always returns false.
_(^? this is what i was uncertain of... but reading the code and adding a log line here confirms at least for my test app)_

One option: unwind that behaviour by moving the `DefaultReplayBreadcrumbConverter` initialization out of `installDefaultIntegrations` so that the `instanceof NoOpReplayBreadcrumbConverter` can evaluate to true here + DefaultReplayBreadcrumbConverter can be initialized here.

Alternatively, leave that code alone (which effectively means the NoOpReplayBreadcrumbConverter is dead code? unless there are tests, or other configurations that don't come through `installDefaultIntegrations`🤔): instead refactor the DefaultReplayBreadcrumbConverter to have post-initialization logic something like
```
final @NotNull  beforeBreadcrumb = options.getReplayController().getBreadcrumbConverter().getBeforeBreadcrumb();
if (beforeBreadcrumb == null) {
  options.getReplayController().getBreadcrumbConverter()
    .setBeforeBreadcrumb(new ReplayBeforeBreadcrumbCallback(options));
}
```

43jay: Moving the initialization of DefaultReplayBreadcrumbConverter out of `installDefaultIntegrations` entirely and depending on `instanceof NoOpReplayBreadcrumbConverter` seems cleaner (first option).

Assuming the initialization was put there b/c previously `DefaultReplayBreadcrumbConverter` was independent of SDKOptions... but "now it is" => move the entire thing out, instead of try to keep some initialization in `installDefaultIntegrations` and then do some more initialization later (here in `initializeIntegrationsAndProcessors`).

I've started working on this (and the refactor to use composition to not change the ReplayBreadcrumbConverter API as well)... any idea if there were other motivations for putting `DefaultReplayBreadcrumbConverter` initialization inside `installDefaultIntegrations`, that moving it down to `initializeIntegrationsAndProcessors` could break?

sentry-android-replay/src/main/java/io/sentry/android/replay/DefaultReplayBreadcrumbConverter.kt
43jay: that was my take as well - and left it in so that the old impl would 'win' in case there is some issue in the new one

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
43jay: yes my plan was to remove it in the next PR when I would introduce the actual SentryOptions API. 

I can take it out and in-line `emptyArray`, etc everywhere if you think that's cleaner though?

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
43jay: done

sentry/src/main/java/io/sentry/util/network/NetworkBody.java
43jay: ty

sentry/src/main/java/io/sentry/util/network/NetworkBody.java
43jay: done

sentry/src/main/java/io/sentry/ReplayBreadcrumbConverter.java
43jay: done

sentry/src/main/java/io/sentry/SentryReplayOptions.java
43jay: done

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
43jay: done!

sentry-android-replay/src/main/java/io/sentry/android/replay/DefaultReplayBreadcrumbConverter.kt
43jay: \0/

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
43jay: done

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
43jay: done

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -247,6 +248,14 @@ static void initializeIntegrationsAndProcessors(
       options.setCompositePerformanceCollector(new DefaultCompositePerformanceCollector(options));
     }
 
+    if (options.getReplayController() != null) { // TODO: Triple-check this should be a null check
+      ReplayBreadcrumbConverter replayBreadcrumbConverter = options.getReplayController().getBreadcrumbConverter();
+      replayBreadcrumbConverter.setUserBeforeBreadcrumbCallback(options.getBeforeBreadcrumb());
+      options.setBeforeBreadcrumb(
+        replayBreadcrumbConverter
+      );
+    }
```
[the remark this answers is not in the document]
43jay: getREplayController annotated @NotNull (and also looks like there's always a NoOpReplayController)
https://github.com/getsentry/sentry-java/blob/0e2c4782f886bfcc9aeffeba8a3eac382a61fc99/sentry/src/main/java/io/sentry/SentryOptions.java#L2857

sentry-android-replay/src/main/java/io/sentry/android/replay/DefaultReplayBreadcrumbConverter.kt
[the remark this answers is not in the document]
43jay: Okay, yeah - i see there will be an issue if multiple HTTP requests go out in the same millisecond, we will only record one of them.

Given it's an impl detail of `Breadcrumb`, which [already has helpers to create http breadcrumbs](https://github.com/getsentry/sentry-java/blob/ebc5ff337f0305b62aac86bc4cf16ed2cb3fb0f9/sentry/src/main/java/io/sentry/Breadcrumb.java#L185-L219) - wyt about adding custom equals/hashcode logic for http breadcrumb i.e. something like
```java
@Override
  public boolean equals(Object o) {
    if (this == o) return true;
    if (o == null || getClass() != o.getClass()) return false;
    Breadcrumb that = (Breadcrumb) o;
     
    switch (type) {
      case  'http':
        return httpBreadcrumbEquals(this, that);
      default:
         return breadcrumbEquals(this, that);
    }
  }

private static boolean breadcrumbEquals(Breadcrumb a, Breadcrumb b) {
return a.getTimestamp().getTime() == that.getTimestamp().getTime()
        && Objects.equals(message, that.message)
        && Objects.equals(type, that.type)
        && Objects.equals(category, that.category)
        && Objects.equals(origin, that.origin)
        && level == that.level;
}

private static boolean httpBreadcrumbEquals(Breadcrumb a, Breadcrumb b) {
    return breadcrumbEquals(a, b) 
        && a.getData("status_code") == b.getData("status_code")
        && a.getData("url") == b.getData("url")
        && a.getData("method") == b.getData("method")
        && a.getData("http.fragment") == b.getData("http.fragment")
        && a.getData("http.query") == b.getData("http.query")
}
```

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
43jay: still pending

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
romtsn: it's fine like that too :)

sentry-android-replay/src/main/java/io/sentry/android/replay/DefaultReplayBreadcrumbConverter.kt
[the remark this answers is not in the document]
romtsn: yeah, that sounds good!

sentry/src/main/java/io/sentry/util/network/NetworkDetailCaptureUtils.java
```diff
@@ -0,0 +1,155 @@
...
+        if (pattern != null && url.matches(pattern)) {
+          return false;
+        }
+      }
+    }
+
+    // Only capture developer-provided urls.
+    if (networkDetailAllowUrls == null) {
+      return false;
+    }
```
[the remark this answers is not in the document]
43jay: [done](https://github.com/getsentry/sentry-java/pull/4796/commits/74c8c6b417c5692f6e221758f6f174b6342bf17d)

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
43jay: done

sentry-android-replay/src/main/java/io/sentry/android/replay/DefaultReplayBreadcrumbConverter.kt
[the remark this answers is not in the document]
43jay: [done](https://github.com/getsentry/sentry-java/pull/4796/commits/85cda10805f7b8cf3b38501023e7ab2f2a96b696)

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -247,6 +248,11 @@ static void initializeIntegrationsAndProcessors(
       options.setCompositePerformanceCollector(new DefaultCompositePerformanceCollector(options));
     }
 
+    if (options.getReplayController().getBreadcrumbConverter() instanceof NoOpReplayBreadcrumbConverter) {
+      options.getReplayController().setBreadcrumbConverter(new DefaultReplayBreadcrumbConverter(options));
+
+    }
```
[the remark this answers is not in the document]
43jay: Believe this is inaccurate... bot believes that `installDefaultIntegrations` happens AFTER `initializeIntegrationsAndProcessors` ?

> ... However, this happens BEFORE the ReplayIntegration is registered in installDefaultIntegrations

but it doesn't seem like [that is the case](https://github.com/getsentry/sentry-java/blob/a9bfbb9f90d26de41ee2586a0539b6288fa73aa1/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java#L131-L173)

sentry/src/main/java/io/sentry/util/network/NetworkBodyParser.java
```diff
@@ -0,0 +1,193 @@
...
+
+    // Convert to string and parse
+    try {
+      String effectiveCharset = charset != null ? charset : "UTF-8";
+      String content = new String(bytes, effectiveCharset);
+      return parse(content, contentType, logger);
+    } catch (UnsupportedEncodingException e) {
+      if (logger != null) {
+        logger.log(SentryLevel.WARNING, "Failed to decode bytes: " + e.getMessage());
```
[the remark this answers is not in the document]
43jay: [fixed](https://github.com/getsentry/sentry-java/pull/4796/commits/0784c5180815b643c6c9e76082f72cb6ab672acd) - forgot to re-run ./gradlew apiDump

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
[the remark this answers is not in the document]
romtsn: yeah it got confused by the order of those methods I guess 🤣

sentry-samples/sentry-samples-android/src/main/AndroidManifest.xml
```diff
@@ -79,10 +79,16 @@
...
 
+      <activity android:name=".TriggerHttpRequestActivity"
+        android:exported="false"/>
+
       <activity
           android:name=".CameraXActivity"
           android:exported="false" />
 
+      <activity android:name=".TriggerHttpRequestActivity"
+        android:exported="false"/>
```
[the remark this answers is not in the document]
43jay: fixed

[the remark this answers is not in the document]
43jay: fixed

sentry/src/main/java/io/sentry/Breadcrumb.java
```diff
@@ -706,23 +706,88 @@ public void setLevel(@Nullable SentryLevel level) {
...
+   * when they share the same timestamp.
+   */
+  private static boolean httpBreadcrumbEquals(
+      final @NotNull Breadcrumb a, final @NotNull Breadcrumb b) {
+    return breadcrumbEquals(a, b)
+        && Objects.equals(a.getData("status_code"), b.getData("status_code"))
+        && Objects.equals(a.getData("url"), b.getData("url"))
+        && Objects.equals(a.getData("method"), b.getData("method"))
+        && Objects.equals(a.getData("http.fragment"), b.getData("http.fragment"))
+        && Objects.equals(a.getData("http.query"), b.getData("http.query"));
```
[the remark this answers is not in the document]
43jay: Including content_length for req/resp doesn't seem like it will add much as if everything else (url, timestamp, method) is equal likely req/resp bodies will be equal.

It could be worth adding SpanDataConvention.HTTP_START_TIMESTAMP, SpanDataConvention.HTTP_END_TIMESTAMP

but based on testing the current #equals imp is able to differentiate between 15 concurrent http requests adequately (i.e. millisecond granularity is sufficient)
https://sentry-sdks.sentry.io/explore/replays/b028d0401d424842bbc0546a217b9e61/?n_detail_row=13&n_detail_tab=response&project=5428559&query=&referrer=%2Fexplore%2Freplays%2F&statsPeriod=1h&t_main=network&yAxis=count%28%29

PR #4809, opened 2025-10-14
Avoid StrictMode warnings (followup)

stefanosiano:
## :scroll: Description
added AndroidRuntimeManager.runWithRelaxedPolicy to
* AndroidEnvelopeCache
* Installation.id calls
* DeviceInfoUtil updated tests

#skip-changelog

Follwup of https://github.com/getsentry/sentry-java/pull/4724

## :pencil: Checklist


sentry/src/test/java/io/sentry/util/runtime/NeutralRuntimeManagerTest.kt
markushi, line 1: Could you add a test to ensure if an exception is thrown it's properly propagated and not silently swallowed?

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 ### Fixes
 
+- Avoid StrictMode warnings (followup) ([#4809](https://github.com/getsentry/sentry-java/pull/4809))
```
markushi: Not sure if it's worth mentioning that as a separate item, maybe combine it with the line below? Not sure if the danger check likes that though 😅

sentry-android-core/src/test/java/io/sentry/android/core/internal/util/AndroidRuntimeManagerTest.kt
```diff
@@ -64,6 +64,37 @@ class AndroidRuntimeManagerTest {
...
+      sut.runWithRelaxedPolicy<Unit> {
+        assertEquals(
+          StrictMode.ThreadPolicy.LAX.toString(),
+          StrictMode.getThreadPolicy().toString(),
+        )
+        assertEquals(StrictMode.VmPolicy.LAX.toString(), StrictMode.getVmPolicy().toString())
+        called = true
+        throw Exception("Test exception")
+      }
+    } catch (_: Exception) {}
```
markushi: Maybe extend the test to assert that the exception is thrown, something like this:
```suggestion
    } catch (_: Exception) {
      exceptionThrown = true
    }
```

sentry-android-core/src/main/java/io/sentry/android/core/DeviceInfoUtil.java
```diff
@@ -227,18 +227,24 @@ private void setDeviceIO(final @NotNull Device device, final boolean includeDyna
...
-      device.setStorageSize(getTotalInternalStorage(internalStorageStat));
-      device.setFreeStorage(getUnusedInternalStorage(internalStorageStat));
-    }
-
-    final @Nullable StatFs externalStorageStat = getExternalStorageStat(internalStorageFile);
-    if (externalStorageStat != null) {
-      device.setExternalStorageSize(getTotalExternalStorage(externalStorageStat));
-      device.setExternalFreeStorage(getUnusedExternalStorage(externalStorageStat));
-    }
+    options
```
markushi, line 230: We do this to have a base of 0 strictmode warnings, right? I'm just wondering if we should keep this as-is, as the collection of these attributes can be turned off via settings anyway.

markushi (approved):
Looks good to me! I'd add tests to verify exceptions are thrown correctly.

sentry-android-core/src/main/java/io/sentry/android/core/DeviceInfoUtil.java
stefanosiano: yeah, but the default is enabled, so i guess it makes more sense for the user to have no strictMode warnings "by default"

Issue #4792, opened 2025-10-08
io.sentry.android.core.ApplicationNotResponding: Background ANR


Jeff11:
### Integration

sentry-android

### Build System

Gradle

### AGP Version

8.14.3

### Proguard

Disabled

### Version

8.19.1

### Steps to Reproduce

ANR happened 8s after the user logged out and app is in background

### Expected Result

no ANR

### Actual Result

```
io.sentry.android.core.ApplicationNotResponding: Background ANR
    at jdk.internal.misc.Unsafe.park
    at java.util.concurrent.locks.LockSupport.park(LockSupport.java:221)
    at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(AbstractQueuedSynchronizer.java:754)
    at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(AbstractQueuedSynchronizer.java:990)
    at java.util.concurrent.locks.ReentrantLock$Sync.lock(ReentrantLock.java:153)
    at java.util.concurrent.locks.ReentrantLock.lock(ReentrantLock.java:322)
    at io.sentry.util.AutoClosableReentrantLock.acquire(AutoClosableReentrantLock.java:12)
    at io.sentry.android.core.AndroidTransactionProfiler.start(AndroidTransactionProfiler.java:127)
    at io.sentry.Scopes.createTransaction(Scopes.java:957)
    at io.sentry.Scopes.startTransaction(Scopes.java:887)
    at io.sentry.Sentry.startTransaction(Sentry.java:1185)
    at io.sentry.ScopesAdapter.startTransaction(ScopesAdapter.java:298)
    at io.sentry.android.core.internal.gestures.SentryGestureListener.startTracing(SentryGestureListener.java:265)
    at io.sentry.android.core.internal.gestures.SentryGestureListener.onSingleTapUp(SentryGestureListener.java:118)
    at android.view.GestureDetector.onTouchEvent(GestureDetector.java:748)
    at androidx.core.view.GestureDetectorCompat.onTouchEvent(GestureDetectorCompat.java:98)
    at io.sentry.android.core.internal.gestures.SentryWindowCallback.handleTouchEvent(SentryWindowCallback.java:70)
    at io.sentry.android.core.internal.gestures.SentryWindowCallback.dispatchTouchEvent(SentryWindowCallback.java:57)
    at com.android.internal.policy.DecorView.dispatchTouchEvent(DecorView.java:456)
[...]

PR #4817, opened 2025-10-20
[ANR] Reduced AndroidTransactionProfiler lock

stefanosiano:
## :scroll: Description
Removed lock from AndroidTransactionProfiler

Trying to reduce ANRs. There are reports of ANRs happening when getting the lock on transaction start/end

resolves https://github.com/getsentry/sentry-java/issues/4792

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidProfiler.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidProfiler.java
@@ -94,7 +94,7 @@ public class AndroidProfiler {
   private final @NotNull Map<String, ProfileMeasurement> measurementsMap = new HashMap<>();
   private final @Nullable ISentryExecutorService timeoutExecutorService;
   private final @NotNull ILogger logger;
-  private boolean isRunning = false;
+  private volatile boolean isRunning = false;
   protected final @NotNull AutoClosableReentrantLock lock = new AutoClosableReentrantLock();
 
   public AndroidProfiler(


sentry-android-core/src/main/java/io/sentry/android/core/AndroidProfiler.java
```diff
@@ -354,4 +354,8 @@ private void putPerformanceCollectionDataInMeasurements(
       }
     }
   }
+
+  boolean isRunning() {
```
romtsn, line 358: haven't looked at the entire PR yet, but feels like this should be AtomicBoolean?

stefanosiano: don't think it's necessary
it's accessed by AndroidTransactionProfiler only in the block guarded by its own AtomicBoolean `isRunning`

romtsn: ah I see we're modifying it only inside the lock in `AndoridProfiler`, makes sense

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
```diff
@@ -219,118 +230,91 @@ public void bindTransaction(final @NotNull ITransaction transaction) {
...
-          currentProfilingTransactionData.getId(),
-          currentProfilingTransactionData.getTraceId(),
+          txData.getName(),
+          txData.getId(),
+          txData.getTraceId(),
           true,
           null,
           ScopesAdapter.getInstance().getOptions());
-    } else if (transactionsCounter != 0) {
+    } else if (isRunning.get()) {
```
romtsn: should this be `getAndSet(false)`?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
```diff
@@ -39,10 +37,10 @@ final class AndroidTransactionProfiler implements ITransactionProfiler {
   private final @NotNull ISentryExecutorService executorService;
   private final @NotNull BuildInfoProvider buildInfoProvider;
   private boolean isInitialized = false;
-  private int transactionsCounter = 0;
+  private final @NotNull AtomicBoolean isRunning = new AtomicBoolean(false);
   private final @NotNull SentryFrameMetricsCollector frameMetricsCollector;
   private @Nullable ProfilingTransactionData currentProfilingTransactionData;
```
romtsn: should this be volatile?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
```diff
@@ -124,22 +123,30 @@ private void init() {
...
-      // causes crashes on api 21 -> https://github.com/getsentry/sentry-java/issues/3392
-      if (buildInfoProvider.getSdkInfoVersion() < Build.VERSION_CODES.LOLLIPOP_MR1) return;
+    // Debug.startMethodTracingSampling() is only available since Lollipop, but Android Profiler
+    // causes crashes on api 21 -> https://github.com/getsentry/sentry-java/issues/3392
+    if (buildInfoProvider.getSdkInfoVersion() < Build.VERSION_CODES.LOLLIPOP_MR1) return;
 
+    // When the first transaction is starting, we can start profiling
+    if (!isRunning.getAndSet(true)) {
       // Let's initialize trace folder and profiling interval
       init();
```
romtsn, line 139: I guess we still need to wrap profiler instantiation into a synchronized lock? And also where we're checking `if (profiler == null)` ?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
stefanosiano: we access it only inside the lock, except in `close()` where we only read it to stop the profiler copying it into another variable
i guess it's fine to keep it as-is? wdyt?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
stefanosiano: profiler instantiation is guarded by `if (!isRunning.getAndSet(true)) {`. Do you think it's not enough?
all other occurrences just read the variable, so `volatile` should be enough imho
mostly because we never nullify it, not even in `close()`

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
romtsn: i think volatile is still needed for immediate visibility between threads? so we either need to wrap it in to a lock in `close()` or make it `volatile` I think

sentry-android-core/src/main/java/io/sentry/android/core/AndroidTransactionProfiler.java
romtsn: ah right, it should be enough!

PR #4825, opened 2025-10-21
[ANR] Defer some class availability checks

stefanosiano:
## :scroll: Description
added lazy class available check to LoadClass 
lazy checking available classes in:
* ActivityFramesTracker
* AndroidViewGestureTargetLocator
* UserInteractionIntegration

Checking if a class exists uses reflection, which can be slow especially during app startup. Deferring some calls, we can improve SDK init time

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/UserInteractionIntegration.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/UserInteractionIntegration.java
@@ -15,7 +15,6 @@ import io.sentry.SentryOptions;
 import io.sentry.android.core.internal.gestures.NoOpWindowCallback;
 import io.sentry.android.core.internal.gestures.SentryGestureListener;
 import io.sentry.android.core.internal.gestures.SentryWindowCallback;
-import io.sentry.util.LazyEvaluator;
 import io.sentry.util.Objects;
 import java.io.Closeable;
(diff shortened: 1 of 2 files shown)


stefanosiano:
Performance gains in release builds for `init()`: 
High end device: negligible - around ~0.2 ms
Low end device: low - around ~3 ms

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Miscellaneous
```
```suggestion
### Improvements
```
i think we use this for perf kind of improvements usually

sentry-android-core/src/main/java/io/sentry/android/core/UserInteractionIntegration.java
romtsn, line 1: not sure changes in this class have any positive effect, because we just moved the reflection checks down to `register` which is also a part of the init codepath.

romtsn (approved):
LGTM, with one thing in UserInteractionIntegration

sentry-android-core/src/main/java/io/sentry/android/core/UserInteractionIntegration.java
stefanosiano: you're right, i reverted the change in that class

PR #4827, opened 2025-10-22
meta(craft): Remove sentry-android-distribution from registry targets

runningcode:
## Summary
Remove sentry-android-distribution from the .craft.yml registry targets list.

We do want to release this module but apparently we need to do this before the release and then add it back afterwards.


runningcode:
Thanks, I’ll just do this so I don’t have to re-run the checks 😓

PR #4841, opened 2025-10-28
Avoid ExecutorService for DefaultCompositePerformanceCollector timeout

stefanosiano:
## :scroll: Description
DefaultCompositePerformanceCollector now stops the performance data collection for timeout on data collection rather than a scheduled background job

Every time a transaction run, the SDK starts collecting performance data (cpu and ram usage) and schedules a timeout job after 30 seconds to stop this data collection.

Since v `8.19.0` we have a limit on the queue of our background executor of 271 scheduled jobs. Any other scheduled job is ignored. 

This means that ***if a transaction runs forever and the background thread queue is full***, its performance data will be collected forever, leading eventually to an OOM.

More [in this doc](https://www.notion.so/sentry/Lyft-Android-memory-leak-investigation-2998b10e4b5d80468139cb206c557390?source=copy_link)

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - [ANR] Removed AndroidTransactionProfiler lock ([#4817](https://github.com/getsentry/sentry-java/pull/4817))
 - Avoid ExecutorService for DefaultCompositePerformanceCollector timeout ([#4841](https://github.com/getsentry/sentry-java/pull/4841))
   - This avoids infinite data collection for never stopped transactions, leading to OOMs
+- Fix wrong .super() call in SentryTimberTree ([#4844](https://github.com/getsentry/sentry-java/pull/4844))
 
 ### Improvements
 
--- a/sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
+++ b/sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
@@ -66,7 +66,7 @@ public class SentryTimberTree(
 
   /** Log an info message with optional format args. */
   override fun i(message: String?, vararg args: Any?) {
-    super.d(message, *args)
+    super.i(message, *args)
     logWithSentry(Log.INFO, null, message, *args)
   }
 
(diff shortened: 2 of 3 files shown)


sentry/src/main/java/io/sentry/DefaultCompositePerformanceCollector.java
```diff
@@ -131,14 +120,17 @@ public void run() {
...
                 for (IPerformanceSnapshotCollector collector : snapshotCollectors) {
                   collector.collect(tempData);
                 }
 
-                for (List<PerformanceCollectionData> data : performanceDataMap.values()) {
-                  data.add(tempData);
+                for (CompositeData data : compositeDataMap.values()) {
+                  if (data.addDataAndCheckTimeout(tempData)) {
+                    // timed out
+                    stop(data.id);
```
romtsn: so there's one side effect to this that we don't call `collector.onSpanFinished` anymore in this case (we used to call `stop(transaction)` previously). I don't know if this can potentially backfire or not

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- Avoid ExecutorService for DefaultCompositePerformanceCollector timeout ([#4841](https://github.com/getsentry/sentry-java/pull/4841))
```
romtsn, line 8: I think we need to be more user-facing in this changelog, e.g. `Fix memory leak for long-running transactions` or something like that?

sentry/src/main/java/io/sentry/DefaultCompositePerformanceCollector.java
stefanosiano: nice catch!

CHANGELOG.md
stefanosiano: wdyt?
```
- Avoid ExecutorService for DefaultCompositePerformanceCollector timeout ([#4841](https://github.com/getsentry/sentry-java/pull/4841))
  - This avoids infinite data collection for never stopped transactions, leading to OOMs
```

romtsn: looks good!

stefanosiano:
@romtsn last check and approval? 😄

PR #4843, opened 2025-10-28
Ensure tests cover non-error timber tags

markushi:
## :scroll: Description

Add another test to verify that https://github.com/getsentry/sentry-java/issues/4484 should be working as expected. Also ran it locally against the Timber 5.0.1 and still can not reproduce the issue.

#skip-changelog


markushi:
Closing in favor of https://github.com/getsentry/sentry-java/pull/4844

PR #4845, opened 2025-10-28
feat(logging): Log Timber tag as log attribute

markushi:
## :scroll: Description

Tags supplied via Timber
```
Timber.tag("tag").i("info message")
```
now get reported as a `timber.tag` attribute.

Better logging

Added unit tests

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Improvements
 
+- Fallback to distinct-id as user.id logging attribute when user is not set ([#4847](https://github.com/getsentry/sentry-java/pull/4847))
 - Report Timber.tag() as `timber.tag` log attribute ([#4845](https://github.com/getsentry/sentry-java/pull/4845))
 
 ## 8.25.0
--- a/sentry/src/main/java/io/sentry/logger/LoggerApi.java
+++ b/sentry/src/main/java/io/sentry/logger/LoggerApi.java
@@ -270,7 +270,14 @@ public final class LoggerApi implements ILoggerApi {
 
   private void setUser(final @NotNull HashMap<String, SentryLogEventAttributeValue> attributes) {
     final @Nullable User user = scopes.getCombinedScopeView().getUser();
-    if (user != null) {
+    if (user == null) {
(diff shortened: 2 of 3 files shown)


markushi:
@sentry review

markushi:
@sentry review

sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
```diff
@@ -240,12 +242,15 @@ public class SentryTimberTree(
...
     msg: String?,
+    tag: String?,
     throwable: Throwable?,
     vararg args: Any?,
   ) {
     // checks the log level
     if (isLoggable(sentryLogLevel, minLogLevel)) {
-      val params = SentryLogParameters()
+      val attributes =
+        tag?.let { SentryAttributes.of(SentryAttribute.stringAttribute("timber.tag", tag)) }
```
romtsn, line 252: so we have a convention for [logger.name](https://getsentry.github.io/sentry-conventions/generated/attributes/logger.html) - I'm wondering if we should treat the timber tag as logger in this case? Maybe worth to check #proj-logs for what's the best way to add this attribute

romtsn (approved):
LGTM with one Q

sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
markushi: Hmm, good point, but I wouldn't as we already set the log origin to `"auto.log.timber"`.

lcian: If we use `timber.tag` we should ideally add it to https://github.com/getsentry/sentry-conventions as well.

markushi:
@sentry review

sentry-android-timber/src/main/java/io/sentry/android/timber/SentryTimberTree.kt
markushi: https://github.com/getsentry/sentry-conventions/pull/183

PR #4846, opened 2025-10-29
Removed SentryExecutorService limit

stefanosiano:
## :scroll: Description
removed `SentryExecutorService` max queue limit

Some critical SDK internal operation depend on our `SentryExecutorService`, so it's better to avoid losing any scheduled job, or we risk things like https://github.com/getsentry/sentry-java/pull/4841/

## :pencil: Checklist


sentry/src/main/java/io/sentry/SentryExecutorService.java
```diff
@@ -23,12 +22,6 @@ public final class SentryExecutorService implements ISentryExecutorService {
    */
   private static final int INITIAL_QUEUE_SIZE = 40;
 
-  /**
-   * By default, the work queue is unbounded so it can grow as much as the memory allows. We want to
-   * limit it by 271 which would be x8 times growth from the default initial capacity.
-   */
-  private static final int MAX_QUEUE_SIZE = 271;
```
romtsn, line 30: sorry I guess I wasn't clear enough - my idea was we only remove the limit from all the `schedule` calls (because we use it for time-critical things), but keep it for `submit`.

sentry/src/main/java/io/sentry/SentryExecutorService.java
```diff
@@ -53,10 +53,19 @@ public SentryExecutorService() {
     this(new ScheduledThreadPoolExecutor(1, new SentryExecutorServiceThreadFactory()), null);
   }
 
+  private boolean isQueueAvailable() {
+    // If limit is reached, purge cancelled tasks from the queue
+    if (executorService.getQueue().size() >= MAX_QUEUE_SIZE) {
+      executorService.purge();
```
romtsn, line 59: hm, but this means now we're dropping existing tasks as opposed to dropping the new incoming ones. What's the intention behind?

stefanosiano: purge just removes cancelled futures from the queue, otherwise they would still be included in the queue and its size
We're not dropping anything, other than tasks that would not be run anyway

romtsn: ah right, it removes the cancelled tasks! makes sense

PR #4847, opened 2025-10-29
fix(logger): fallback to distinct-id when user is not set

markushi:
When no user is set in the scope, the logger now falls back to using the distinct ID (installation ID on Android) as the user ID in log attributes. This ensures logs always have a user identifier when available - and it also matches the `DefaultAndroidEventProcessor` behavior.


markushi:
@sentry review

PR #4858, opened 2025-11-03
feat(size-analysis): Integrate Size Analysis to measure SDK size overhead

romtsn:
_#skip-changelog_

Added some keep rules since we want to measure the worst case scenario (every integration is used)

Example: https://sentry-sdks.sentry.io/preprod/sentry-android/2379

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@
 - Report Timber.tag() as `timber.tag` log attribute ([#4845](https://github.com/getsentry/sentry-java/pull/4845))
 - Session Replay: Add screenshot strategy serialization to RRWeb events ([#4851](https://github.com/getsentry/sentry-java/pull/4851))
 
+### Dependencies
+
+- Bump Native SDK from v0.11.3 to v0.12.1 ([#4859](https://github.com/getsentry/sentry-java/pull/4859))
+  - [changelog](https://github.com/getsentry/sentry-native/blob/master/CHANGELOG.md#0121)
+  - [diff](https://github.com/getsentry/sentry-native/compare/0.11.3...0.12.1)
+
 ## 8.25.0
 
 ### Fixes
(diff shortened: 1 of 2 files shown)


sentry-android-integration-tests/test-app-size/build.gradle.kts
```diff
@@ -0,0 +1,60 @@
...
+
+sentry {
+  org.set("sentry-sdks")
+  projectName.set("sentry-android")
+  authToken.set(System.getenv("SENTRY_AUTH_TOKEN"))
+  includeProguardMapping.set(false)
+  tracingInstrumentation.enabled.set(false)
+  includeDependenciesReport.set(false)
+  telemetry.set(false)
+  sizeAnalysis.enabled.set(true)
```
runningcode: Which set of integration instructions were you following? i ask because we don't recommend enabling this as `true` because it will cause a long upload on every local debug build so something needs to be updated.

romtsn: yeah I know, but we're almost never gonna build it locally (only for the purpose of verifying size analysis works), because it's a dummy app created specifically for size analysis so there's no really other use for it.

romtsn: I followed the [draft docs](https://sentry-docs-git-mtopo27-size-analysis-docs.sentry.dev/platforms/android/size-analysis/#uploading-with-gradle), but set it to `true` purposefully so nothing needs to be updated :)

runningcode: Ah that makes sense. Well, it isn't just about uploading every time, someone without a sentry-auth-token will experience a failure to build.

romtsn: true that! I'll make it depend on the auth token then

rbro112:
Thanks a ton for dogfooding!

PR #4873, opened 2025-11-06
feat(android): Add log flushing on app backgrounding

markushi:
## Summary

Adds log flushing functionality that triggers when the Android app transitions to the background.

### DIFF
--- a/sentry-android-core/api/sentry-android-core.api
+++ b/sentry-android-core/api/sentry-android-core.api
@@ -82,6 +82,18 @@ public final class io/sentry/android/core/AndroidLogger : io/sentry/ILogger {
 	public fun log (Lio/sentry/SentryLevel;Ljava/lang/Throwable;Ljava/lang/String;[Ljava/lang/Object;)V
 }
 
+public class io/sentry/android/core/AndroidLoggerApi : io/sentry/logger/LoggerApi, io/sentry/android/core/AppState$AppStateListener, java/io/Closeable {
+	public fun <init> (Lio/sentry/Scopes;)V
+	public fun close ()V
+	public fun onBackground ()V
+	public fun onForeground ()V
+}
+
+public final class io/sentry/android/core/AndroidLoggerApiFactory : io/sentry/logger/ILoggerApiFactory {
+	public fun <init> ()V
+	public fun create (Lio/sentry/Scopes;)Lio/sentry/logger/LoggerApi;
+}
+
 public class io/sentry/android/core/AndroidMemoryCollector : io/sentry/IPerformanceSnapshotCollector {
 	public fun <init> ()V
 	public fun collect (Lio/sentry/PerformanceCollectionData;)V
--- a/sentry/api/sentry.api
+++ b/sentry/api/sentry.api
@@ -3409,6 +3409,7 @@ public class io/sentry/SentryOptions {
 	public fun getIntegrations ()Ljava/util/List;
 	public fun getInternalTracesSampler ()Lio/sentry/TracesSampler;
 	public fun getLogger ()Lio/sentry/ILogger;
+	public fun getLoggerApiFactory ()Lio/sentry/logger/ILoggerApiFactory;
 	public fun getLogs ()Lio/sentry/SentryOptions$Logs;
 	public fun getMaxAttachmentSize ()J
 	public fun getMaxBreadcrumbs ()I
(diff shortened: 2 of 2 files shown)


sentry/src/main/java/io/sentry/ISentryClient.java
```diff
@@ -47,6 +47,8 @@ public interface ISentryClient {
    */
   void flush(long timeoutMillis);
 
+  void flushLogs(long timeoutMillis);
```
markushi, line 50: Not the biggest fan of adding a new API here, an alternative would be to move this directly into `ILoggerApi`

stefanosiano:
Isn't log processor flushing every 5 seconds anyway?
Is there a new requirement for this?
Or am i misunderstanding something?

markushi:
> Isn't log processor flushing every 5 seconds anyway? Is there a new requirement for this? Or am i misunderstanding something?

The [spec](https://develop.sentry.dev/sdk/telemetry/telemetry-buffer/#specification) mentions this:
> 4. When the application moves to the background, the BatchProcessor SHOULD send all data in memory and stop the timer. This is mostly relevant for mobile SDKs.

CHANGELOG.md
```diff
@@ -49,6 +49,7 @@ SentryAndroid.init(
 ```
 
 </details>
+- Android: Flush logs when app enters background ([#4873](https://github.com/getsentry/sentry-java/pull/4873))
```
romtsn, line 52: maybe this should go above the previous entry, because it doesn't look right when rendered:
image

sentry-android-core/src/main/java/io/sentry/android/core/LifecycleWatcher.java
```diff
@@ -101,6 +108,29 @@ public void onBackground() {
     scheduleEndSession();
 
     addAppBreadcrumb("background");
+
+    if (enableLogFlushing) {
```
romtsn: hmm, wondering if it's a good idea to bind everything to LifecycleWatcher.. Since we now can addAppStateListener from anywhere, wdyt of introducing an `AndroidLoggerApi` which delegates everything to LoggerApi, but in addition subscribes for onBackground/onForeground to trigger logs flush? And then we override the Java LoggerApi via Options as always

sentry/src/main/java/io/sentry/ISentryClient.java
romtsn: if we're not going for https://github.com/getsentry/sentry-java/pull/4873/files#r2581483713, then I think this makes sense to make it cleaner and bake it into ILoggerApi instead

sentry-android-core/src/main/java/io/sentry/android/core/LifecycleWatcher.java
```diff
@@ -101,6 +108,29 @@ public void onBackground() {
...
+      try {
+        scopes
+            .getOptions()
+            .getExecutorService()
+            .submit(
+                new Runnable() {
+                  @Override
+                  public void run() {
+                    scopes
+                        .getGlobalScope()
```
adinauer: `m` in theory there could be a different `SentryClient` instance on each scope type but that's probably more of an edge case. Romans suggestion above should get rid of that problem tho.

As an alternative, on `Scopes` there's already an implementation:
```
  @ApiStatus.Internal
  @NotNull
  public ISentryClient getClient() {
    return getCombinedScopeView().getClient();
  }
```

We'd just have to expose that on `IScopes`.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidLoggerApi.java
```diff
@@ -0,0 +1,49 @@
...
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+@ApiStatus.Internal
+public class AndroidLoggerApi extends LoggerApi implements Closeable, AppState.AppStateListener {
+
+  public AndroidLoggerApi(final @NotNull Scopes scopes) {
+    super(scopes);
+    AppState.getInstance().addAppStateListener(this);
+  }
```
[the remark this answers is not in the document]
markushi: Yeah that's a valid point, although not a concern in globalhub mode. @romtsn I'm wondering if we should completely decouple the flush-triggering logic from logger(api), as the changes blew up quite a bit. We could move it to `SentryAndroid.init()`

markushi:
Closing in favor of https://github.com/getsentry/sentry-java/pull/4951

PR #4899, opened 2025-11-12
feat(anr): Profile main thread when ANR and report ANR profiles to Sentry

markushi:
## :scroll: Description
Adds ANR (Application Not Responding) profiling integration that profiles the main thread when an ANR is detected and reports the captured profiles to Sentry.

### Key Changes:
- New `AnrProfilingIntegration` to capture profiles during ANR events
- `AnrV2Integration` now takes care of matching and capturing the profile on the next start
- If the captured ANR event only contains system frames, a static fingerprint will get set, effectively changing the grouping behavior to group all noisy ANRs into a single issue

This feature enables better ANR diagnostics by capturing profiling data at the time of ANR detection, allowing developers to identify performance bottlenecks and problematic code paths causing application hangs.

Example event: https://sentry-sdks.sentry.io/issues/7229210096/events/4598ff6fcc0f402d8ecca615005e7f64/

- Added tests

## :pencil: Checklist
- [ ] I added GH Issue ID & Linear ID

## :crystal_ball: Next steps

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
@@ -734,8 +734,8 @@ public final class ApplicationExitInfoEventProcessor implements BackfillingEvent
         return;
       }
 
-      if (options.isAnrProfilingEnabled() && hasOnlySystemFrames(event)) {
-        // If profiling did not identify any app frames, we want to statically group these events
+      if (options.isEnableAnrFingerprinting() && hasOnlySystemFrames(event)) {
+        // If the stacktrace only contains system frames, we want to statically group these events
         // to avoid ANR noise due to {{ default }} stacktrace grouping
         event.setFingerprints(
             Arrays.asList(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,9 +36,11 @@
 - Add new experimental option to capture profiles for ANRs ([#4899](https://github.com/getsentry/sentry-java/pull/4899))
   - This feature will capture a stack profile of the main thread when it gets unresponsive
   - The profile gets attached to the ANR event on the next app start, providing a flamegraph of the ANR issue on the sentry issue details page
-  - Breaking change: if the ANR stacktrace contains only system frames (e.g. `java.lang` or `android.os`), a static fingerprint is set on the ANR event, causing all ANR events to be grouped into a single issue, reducing the overall ANR issue noise
   - Enable via `options.setAnrProfilingSampleRate(<sample-rate>)` or AndroidManifest.xml: `<meta-data android:name="io.sentry.anr.profiling.sample-rate" android:value="[0.0-1.0]" />`
   - The sample rate controls the probability of collecting a profile for each detected foreground ANR (0.0 to 1.0, null to disable)
+- Add `enableAnrFingerprinting` option to reduce ANR noise by assigning static fingerprints to ANR events with system-only stacktraces
+  - When enabled, ANRs whose stacktraces contain only system frames (e.g. `java.lang` or `android.os`) are grouped into a single issue instead of creating many separate issues
+  - Enable via `options.setEnableAnrFingerprinting(true)` or AndroidManifest.xml: `<meta-data android:name="io.sentry.anr.enable-fingerprinting" android:value="true" />`
 
 ### Fixes
 
--- a/sentry-android-core/api/sentry-android-core.api
+++ b/sentry-android-core/api/sentry-android-core.api
@@ -367,6 +367,7 @@ public final class io/sentry/android/core/SentryAndroidOptions : io/sentry/Sentr
 	public fun isCollectExternalStorageContext ()Z
 	public fun isEnableActivityLifecycleBreadcrumbs ()Z
 	public fun isEnableActivityLifecycleTracingAutoFinish ()Z
+	public fun isEnableAnrFingerprinting ()Z
 	public fun isEnableAppComponentBreadcrumbs ()Z
 	public fun isEnableAppLifecycleBreadcrumbs ()Z
 	public fun isEnableAutoActivityLifecycleTracing ()Z
@@ -396,6 +397,7 @@ public final class io/sentry/android/core/SentryAndroidOptions : io/sentry/Sentr
 	public fun setDebugImagesLoader (Lio/sentry/android/core/IDebugImagesLoader;)V
 	public fun setEnableActivityLifecycleBreadcrumbs (Z)V
 	public fun setEnableActivityLifecycleTracingAutoFinish (Z)V
+	public fun setEnableAnrFingerprinting (Z)V
 	public fun setEnableAppComponentBreadcrumbs (Z)V
 	public fun setEnableAppLifecycleBreadcrumbs (Z)V
 	public fun setEnableAutoActivityLifecycleTracing (Z)V
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
@@ -177,6 +177,8 @@ final class ManifestMetadataReader {
 
   static final String ANR_PROFILING_SAMPLE_RATE = "io.sentry.anr.profiling.sample-rate";
 
+  static final String ENABLE_ANR_FINGERPRINTING = "io.sentry.anr.enable-fingerprinting";
+
   /** ManifestMetadataReader ctor */
   private ManifestMetadataReader() {}
 
@@ -684,6 +686,10 @@ final class ManifestMetadataReader {
             options.setAnrProfilingSampleRate(anrProfilingSampleRate);
           }
         }
+
+        options.setEnableAnrFingerprinting(
+            readBool(
+                metadata, logger, ENABLE_ANR_FINGERPRINTING, options.isEnableAnrFingerprinting()));
       }
       options
           .getLogger()
--- a/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
@@ -255,6 +255,8 @@ public final class SentryAndroidOptions extends SentryOptions {
 
   private @Nullable Double anrProfilingSampleRate;
 
+  private boolean enableAnrFingerprinting = true;
+
   public SentryAndroidOptions() {
     setSentryClientName(BuildConfig.SENTRY_ANDROID_SDK_NAME + "/" + BuildConfig.VERSION_NAME);
     setSdkVersion(createSdkVersion());
@@ -714,6 +716,29 @@ public final class SentryAndroidOptions extends SentryOptions {
 
   public boolean isAnrProfilingEnabled() {
     return anrProfilingSampleRate != null && anrProfilingSampleRate > 0;
+  }
+
+  /**
+   * Returns whether ANR fingerprinting is enabled. When enabled, the SDK assigns static
+   * fingerprints to ANR events that would otherwise produce noisy grouping. Currently, this applies
+   * a static fingerprint to ANRs whose stacktraces contain only system frames and no application
+   * frames.
+   *
+   * @return true if ANR fingerprinting is enabled
+   */
+  public boolean isEnableAnrFingerprinting() {
+    return enableAnrFingerprinting;
+  }
+
+  /**
+   * Sets whether ANR fingerprinting is enabled. When enabled, the SDK assigns static fingerprints
+   * to ANR events that would otherwise produce noisy grouping. Currently, this applies a static
+   * fingerprint to ANRs whose stacktraces contain only system frames and no application frames.
+   *
+   * @param enableAnrFingerprinting true to enable ANR fingerprinting
+   */
+  public void setEnableAnrFingerprinting(final boolean enableAnrFingerprinting) {
+    this.enableAnrFingerprinting = enableAnrFingerprinting;
   }
 
   static class AndroidUserFeedbackIDialogHandler implements SentryFeedbackOptions.IDialogHandler {
--- a/sentry-android-core/src/test/java/io/sentry/android/core/ApplicationExitInfoEventProcessorTest.kt
+++ b/sentry-android-core/src/test/java/io/sentry/android/core/ApplicationExitInfoEventProcessorTest.kt
@@ -668,8 +668,8 @@ class ApplicationExitInfoEventProcessorTest {
   }
 
   @Test
-  fun `sets system-frames-only fingerprint when ANR profiling enabled and no app frames`() {
-    fixture.options.anrProfilingSampleRate = 1.0
+  fun `sets system-frames-only fingerprint when ANR fingerprinting enabled and no app frames`() {
+    fixture.options.isEnableAnrFingerprinting = true
     val hint = HintUtils.createWithTypeCheckHint(AbnormalExitHint(mechanism = "anr_foreground"))
 
     val processed =
@@ -697,8 +697,8 @@ class ApplicationExitInfoEventProcessorTest {
   }
 
   @Test
-  fun `does not set system-frames-only fingerprint when ANR profiling is disabled but no app frames are present`() {
-    fixture.options.anrProfilingSampleRate = null
+  fun `does not set system-frames-only fingerprint when ANR fingerprinting is disabled and no app frames are present`() {
+    fixture.options.isEnableAnrFingerprinting = false
     val hint = HintUtils.createWithTypeCheckHint(AbnormalExitHint(mechanism = "anr_foreground"))
 
     val processed =
@@ -726,8 +726,8 @@ class ApplicationExitInfoEventProcessorTest {
   }
 
   @Test
-  fun `sets default fingerprint when ANR profiling enabled and app frames are present`() {
-    fixture.options.anrProfilingSampleRate = 1.0
+  fun `sets default fingerprint when ANR fingerprinting enabled and app frames are present`() {
+    fixture.options.isEnableAnrFingerprinting = true
     val hint = HintUtils.createWithTypeCheckHint(AbnormalExitHint(mechanism = "anr_foreground"))
 
     val processed =
--- a/sentry-android-core/src/test/java/io/sentry/android/core/ManifestMetadataReaderTest.kt
+++ b/sentry-android-core/src/test/java/io/sentry/android/core/ManifestMetadataReaderTest.kt
@@ -2022,6 +2022,31 @@ class ManifestMetadataReaderTest {
     assertNull(fixture.options.anrProfilingSampleRate)
   }
 
+  @Test
+  fun `applyMetadata reads enableAnrFingerprinting to options`() {
+    // Arrange
+    val bundle = bundleOf(ManifestMetadataReader.ENABLE_ANR_FINGERPRINTING to true)
+    val context = fixture.getContext(metaData = bundle)
+
+    // Act
+    ManifestMetadataReader.applyMetadata(context, fixture.options, fixture.buildInfoProvider)
+
+    // Assert
+    assertTrue(fixture.options.isEnableAnrFingerprinting)
+  }
+
+  @Test
+  fun `applyMetadata keeps enableAnrFingerprinting default when not set in manifest`() {
+    // Arrange
+    val context = fixture.getContext()
+
+    // Act
+    ManifestMetadataReader.applyMetadata(context, fixture.options, fixture.buildInfoProvider)
+
+    // Assert
+    assertTrue(fixture.options.isEnableAnrFingerprinting)
+  }
+
   // Network Detail Configuration Tests
 
   @Test
--- a/sentry-samples/sentry-samples-android/src/main/AndroidManifest.xml
+++ b/sentry-samples/sentry-samples-android/src/main/AndroidManifest.xml
@@ -268,5 +268,8 @@
     <meta-data
       android:name="io.sentry.anr.profiling.sample-rate"
       android:value="1.0" />
+    <meta-data
+      android:name="io.sentry.anr.enable-fingerprinting"
+      android:value="true" />
   </application>
 </manifest>


sentry-android-core/src/main/java/io/sentry/android/core/AnrV2Integration.java
```diff
@@ -299,6 +315,78 @@ private void reportAsSentryEvent(
...
+          }
+        } else {
+          options.getLogger().log(SentryLevel.DEBUG, "ANR profile found, but doesn't match");
+        }
+      }
+    }
+
     private @NotNull ParseResult parseThreadDump(
         final @NotNull ApplicationExitInfo exitInfo, final boolean isBackground) {
       final byte[] dump;
```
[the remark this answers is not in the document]
markushi: That's fine, it should not call rotate. This should only happen once after app start.

markushi:
@sentry review

sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
```diff
@@ -700,8 +714,15 @@ public void applyPreEnrichment(
     public void applyPostEnrichment(
         @NotNull SentryEvent event, @NotNull Backfillable hint, @NotNull Object rawHint) {
       final boolean isBackgroundAnr = isBackgroundAnr(rawHint);
-      setAppForeground(event, !isBackgroundAnr);
+
+      if (options.isEnableAnrProfiling()) {
+        applyAnrProfile(event, hint, isBackgroundAnr);
+      }
+
```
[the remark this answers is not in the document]
markushi: No, this was already done before as well: https://github.com/getsentry/sentry-java/blob/ff8eea48bb74ce9298a85623cf2b7d64e6869bc2/sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java#L707-L716

sentry-android-core/src/test/java/io/sentry/android/core/anr/AnrProfilingIntegrationTest.kt
```diff
@@ -0,0 +1,255 @@
...
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFalse
+import kotlin.test.assertNotNull
+import kotlin.test.assertTrue
+import org.junit.runner.RunWith
+import org.mockito.kotlin.mock
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
```
romtsn: I think `AndroidJUnit4` should be used instead of robolectric here?

sentry-android-core/src/test/java/io/sentry/android/core/anr/AnrProfilingIntegrationTest.kt
```diff
@@ -0,0 +1,255 @@
...
+        cacheDirPath = tempDir.absolutePath
+        setLogger(mockLogger)
+        isEnableAnrProfiling = true
+      }
+    AppState.getInstance().resetInstance()
+  }
+
+  @AfterTest
+  fun cleanup() {
+    if (::tempDir.isInitialized && tempDir.exists()) {
```
romtsn: We could also use `@get:Rule val tmpDir = TemporaryFolder()` and not deal with the cleanup here. I think we should switch everywhere to use the temp folder rule but that's for later

sentry-android-core/src/test/java/io/sentry/android/core/anr/AnrProfileRotationHelperTest.kt
```diff
@@ -0,0 +1,121 @@
...
+import kotlin.test.assertEquals
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+class AnrProfileRotationHelperTest {
+  private lateinit var tempDir: File
+
+  @BeforeTest
+  fun setup() {
+    tempDir = Files.createTempDirectory("anr_profile_rotation_test").toFile()
```
romtsn: same here - could be replaced by the rule

sentry-android-core/src/test/java/io/sentry/android/core/anr/AnrProfileManagerTest.kt
```diff
@@ -0,0 +1,145 @@
...
+import java.nio.file.Files
+import kotlin.test.AfterTest
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNotNull
+import kotlin.test.assertTrue
+import org.mockito.kotlin.mock
+
+class AnrProfileManagerTest {
+  private lateinit var tempDir: File
```
romtsn: same here - could be a rule

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -140,6 +142,8 @@ static void loadDefaultAndMetadataOptions(
             .getRuntimeManager()
             .runWithRelaxedPolicy(() -> getCacheDir(finalContext).getAbsolutePath()));
 
+    AnrProfileRotationHelper.rotate();
```
romtsn, line 140: curious, why this would be needed? because the shouldRotate boolean is true by default. Or is this for the multiple init+close case?

sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
```diff
@@ -700,16 +714,33 @@ public void applyPreEnrichment(
...
         final @NotNull SentryEvent event, final boolean isBackgroundAnr) {
       // sentry does not yet have a capability to provide default server-side fingerprint rules,
       // so we're doing this on the SDK side to group background and foreground ANRs separately
       // even if they have similar stacktraces.
-      if (event.getFingerprints() == null) {
+      if (event.getFingerprints() != null) {
+        return;
+      }
+
+      if (options.isEnableAnrProfiling() && hasOnlySystemFrames(event)) {
```
romtsn: wondering whether we should guard this behind `isEnableAnrProfiling` or actually just do it for all ANRs going forward? I think AEI also gives quite a lot of noise with just system frames, right?

sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
```diff
@@ -777,5 +808,142 @@ private void setAnrExceptions(
...
+
+      // Capture profile chunk
+      final SentryId profilerId = captureAnrProfile(anrTimestamp, anrProfile);
+
+      final StackTraceElement[] stack = culprit.getStack();
+      if (stack.length > 0) {
+        final StackTraceElement stackTraceElement = stack[0];
+        final String message =
+            stackTraceElement.getClassName() + "." + stackTraceElement.getMethodName();
+        final AnrException exception = new AnrException(message);
```
romtsn: is the new AnrException type intentional here? Because it's a new type it will create new groups (even if the profile-derived stacktrace matches the one from AEI). Not sure if we should keep it as ApplicationNotResponding?

sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
```diff
@@ -777,5 +808,142 @@ private void setAnrExceptions(
...
+        exception.setStackTrace(stack);
+
+        final @NotNull List<SentryException> sentryException =
+            sentryExceptionFactory.getSentryExceptions(exception);
+        for (final @NotNull SentryException e : sentryException) {
+          final Mechanism mechanism = new Mechanism();
+          mechanism.setType("ANR");
```
romtsn: I guess this is also intentional to override `AppExitInfo`, right? Wondering if a new mechanism type would make more sense here (like `ProfiledAppExitInfo`, idk)

sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
```diff
@@ -777,5 +808,142 @@ private void setAnrExceptions(
...
+              new SentryId(),
+              new SentryId(),
+              null,
+              new HashMap<>(0),
+              anrTimestampMs / 1000.0d,
+              ProfileChunk.PLATFORM_JAVA,
+              options);
+      chunk.setSentryProfile(profile);
+
+      final SentryId profilerId = Sentry.getCurrentScopes().captureProfileChunk(chunk);
```
romtsn, line 918: just curious, but is there a way to send both the ANR event and the profile  chunk in the same envelope?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
markushi: Yes, in case of re-init, we also want to have a "fresh" ANR profile file for recording.

sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
markushi: Yes, exactly - we should do this everywhere in the long run. I still would guard this right now - otherwise we'll have a breaking change in default behavior. I can create a follow up ticket for the next major.

sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
markushi: Yes, I think I looked into this early on, but `ApplicationNotResponding` requires a non-null Thread object, which seemed to be widely used and hard to refactor. But that's a good point, let me check that again.

sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
markushi: Yeah that's because it would default to "chained" right now (our exception factory does that), when the exception is extracted from the ANR profile. I'd keep `ANR` here, as it's used in a few places within our product, and the `event->context->profiler_id` anyway would indicate that there's a profile attached.

sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
markushi: claude says yes, this should be valid for the backend😅 
> Yes, this works. The client can send both in the same envelope.
>  During split_envelope() at relay-server/src/services/processor.rs:357-365, Relay splits them apart before any event-type routing:
>  1. ProfileChunk items are extracted into their own ProcessingGroup::ProfileChunk envelope
>  2. The remaining error event goes into ProcessingGroup::Error
>  3. Both are processed independently through their respective pipelines
  
  
We'd need to attach the profile directly to the event (or via hint) and then do something similar as we do for e.g. attachments (https://github.com/getsentry/sentry-java/blob/abf451a1c165a9f3cc283004c6d8a319645d65ec/sentry/src/main/java/io/sentry/SentryClient.java#L422-L432)

romtsn: yeah, or replay_recording, too. 

I guess this will be better in terms of not sending one without the other, but I don't have a strong preference right now, we can probably create a followup issue if you don't feel like doing it now.

sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
romtsn: makes sense!

sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
romtsn: I think we've done this quite a lot in the past (breaking behaviour change that affects grouping) so I'm not opposed to doing this now and not wait for the next major 😅 If it improves things I think I'd rather do it sooner.  But your call here (we could also wait to get some adoption and see how it performs before doing this for everyone)

sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
```diff
@@ -777,5 +808,142 @@ private void setAnrExceptions(
...
+              final String module = frame.getModule();
+              if (module != null && !AnrCulpritIdentifier.isSystemFrame(module)) {
+                return false;
+              }
+            }
+          }
+        }
+      }
+      return true;
+    }
```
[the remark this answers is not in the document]
markushi: this case is quite unlikely and you could still argue that the lack of frames should set a static fingerprint, "system-frames-only-anr" in this case.

sentry-android-core/src/main/java/io/sentry/android/core/anr/StackTraceConverter.java
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    profile.setFrames(frames);
+    profile.setStacks(stacks);
+
+    final @NotNull SentryThreadMetadata threadMetadata = new SentryThreadMetadata();
+    threadMetadata.setName(MAIN_THREAD_NAME);
+    threadMetadata.setPriority(Thread.NORM_PRIORITY);
+
+    final @NotNull Map<String, SentryThreadMetadata> threadMetadataMap = new HashMap<>();
```
romtsn: I guess this could've been `Collections.singletonMap()`

sentry-android-core/src/main/java/io/sentry/android/core/anr/AnrProfileRotationHelper.java
```diff
@@ -0,0 +1,69 @@
...
+    performRotationIfNeeded(cacheDir);
+    return new File(cacheDir, OLD_FILE_NAME);
+  }
+
+  public static boolean deleteLastFile(final @NotNull File cacheDir) {
+    final File oldFile = new File(cacheDir, OLD_FILE_NAME);
+    if (!oldFile.exists()) {
+      return true;
+    }
+    return oldFile.delete();
```
romtsn, line 71: I guess we need to try-catch this (because of security or IO exceptions), and if we try-catch, we could drop the `exists()` check to save some I/O ?

sentry-android-core/src/main/java/io/sentry/android/core/anr/AnrProfileRotationHelper.java
```diff
@@ -0,0 +1,69 @@
...
+      final File currentFile = new File(cacheDir, RECORDING_FILE_NAME);
+      final File oldFile = new File(cacheDir, OLD_FILE_NAME);
+
+      if (oldFile.exists()) {
+        oldFile.delete();
+      }
+
+      if (currentFile.exists()) {
+        currentFile.renameTo(oldFile);
+      }
```
romtsn: I guess same here with try-catching these

sentry-android-core/src/main/java/io/sentry/android/core/anr/AnrProfile.java
```diff
@@ -0,0 +1,34 @@
...
+import org.jetbrains.annotations.ApiStatus;
+
+@ApiStatus.Internal
+public class AnrProfile {
+  public final List<AnrStackTrace> stacks;
+
+  public final long startTimeMs;
+  public final long endTimeMs;
+
+  public AnrProfile(List<AnrStackTrace> stacks) {
```
romtsn:
```suggestion
  public AnrProfile(final @NotNull List<AnrStackTrace> stacks) {
```

sentry-android-core/src/main/java/io/sentry/android/core/anr/AnrProfileManager.java
```diff
@@ -0,0 +1,101 @@
...
+    if (queueFile == null) {
+      queue = ObjectQueue.createEmpty();
+    } else {
+      queue =
+          ObjectQueue.create(
+              queueFile,
+              new ObjectQueue.Converter<AnrStackTrace>() {
+                @Override
+                public AnrStackTrace from(final byte[] source) throws IOException {
+                  final @NotNull ByteArrayInputStream bis = new ByteArrayInputStream(source);
```
romtsn, line 67: do we need to try-with-resources this (and in `toStream` too) to ensure resource cleanup?

sentry-android-core/src/main/java/io/sentry/android/core/anr/AnrProfileManager.java
```diff
@@ -0,0 +1,101 @@
...
+import java.io.DataOutputStream;
+import java.io.File;
+import java.io.IOException;
+import java.io.OutputStream;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+@ApiStatus.Internal
+public class AnrProfileManager implements Closeable {
```
romtsn: I think this needs to be [AutoCloseable](https://docs.oracle.com/javase/8/docs/api/java/lang/AutoCloseable.html) to be used in try-with-resouces?

sentry-android-core/src/main/java/io/sentry/android/core/anr/AnrProfileManager.java
```diff
@@ -0,0 +1,101 @@
...
+
+  public AnrProfileManager(final @NotNull SentryOptions options) {
+    this(
+        options,
+        new File(
+            Objects.requireNonNull(options.getCacheDirPath(), "cacheDirPath is required"),
+            "anr_profile"));
+  }
+
+  public AnrProfileManager(final @NotNull SentryOptions options, final @NotNull File file) {
```
romtsn, line 38: Is the ctor only being used from a bg thread? otherwise I guess we'd have to make the queueFile instantiation lazy

sentry-android-core/src/main/java/io/sentry/android/core/anr/AnrProfileManager.java
markushi: Quoting the docs, this shouldn't be required at all. I'll add a comment
```
/**
* Closing a {@code ByteArrayInputStream} has no effect. The methods in
* this class can be called after the stream has been closed without
* generating an {@code IOException}.
*/
```

sentry-android-core/src/main/java/io/sentry/android/core/anr/AnrProfileManager.java
markushi: Good catch, thanks!

sentry-android-core/src/main/java/io/sentry/android/core/anr/AnrProfileManager.java
markushi: Yes, it's from BG threads only.

sentry-android-core/src/main/java/io/sentry/android/core/anr/AnrCulpritIdentifier.java
```diff
@@ -0,0 +1,159 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+@ApiStatus.Internal
+public class AnrCulpritIdentifier {
+
+  // common Java and Android packages who are less relevant for being the actual culprit
+  private static final List<String> systemAndFrameworkPackages = new ArrayList<>(9);
+
+  static {
```
romtsn, line 18: not sure if it's worth adding some kotlin frames in here as well?

sentry-android-core/src/main/java/io/sentry/android/core/anr/AnrProfilingIntegration.java
```diff
@@ -0,0 +1,244 @@
...
+
+  @Override
+  public void register(final @NotNull IScopes scopes, final @NotNull SentryOptions options) {
+    this.options =
+        Objects.requireNonNull(
+            (options instanceof SentryAndroidOptions) ? (SentryAndroidOptions) options : null,
+            "SentryAndroidOptions is required");
+    this.logger = options.getLogger();
+
+    if (((SentryAndroidOptions) options).isEnableAnrProfiling()) {
```
romtsn:
```suggestion
    if (this.options.isEnableAnrProfiling()) {
```
since we already cast it above

sentry-android-core/src/main/java/io/sentry/android/core/anr/AnrProfilingIntegration.java
```diff
@@ -0,0 +1,244 @@
...
+  @Override
+  public void close() throws IOException {
+    onBackground();
+    enabled.set(false);
+    AppState.getInstance().removeAppStateListener(this);
+
+    try (final @NotNull ISentryLifecycleToken ignored = profileManagerLock.acquire()) {
+      final @Nullable AnrProfileManager p = profileManager;
+      if (p != null) {
+        p.close();
```
romtsn: hm, I think this one does I/O under the hood right (raf.close())? We should probably move it out of the lock here (since we capture it in the `p` variable) and also offload to a bg thread perhaps?

sentry-android-core/src/main/java/io/sentry/android/core/anr/AnrProfilingIntegration.java
```diff
@@ -0,0 +1,244 @@
...
+      }
+      inForeground = true;
+      updater.run();
+
+      final @Nullable Thread oldThread = thread;
+      if (oldThread != null) {
+        oldThread.interrupt();
+      }
+
+      final @NotNull Thread profilingThread = new Thread(this, "AnrProfilingIntegration");
```
romtsn: can we actually reuse the thread as opposed to creating a new one on every foreground? like checking the inForeground flag inside the `run` method in addition to isInterrupted?

also not sure if it makes any difference on android as to whether to make it `isDaemon` or not

sentry-android-core/src/main/java/io/sentry/android/core/anr/AnrProfilingIntegration.java
```diff
@@ -0,0 +1,244 @@
...
+      }
+    }
+  }
+
+  @Override
+  public void run() {
+    // get main thread Handler so we can post messages
+    final Looper mainLooper = Looper.getMainLooper();
+    final Thread mainThread = mainLooper.getThread();
+    final Handler mainHandler = new Handler(mainLooper);
```
romtsn: I guess we could memoize this and capture it in `register` if anr profiling is enabled? Would avoid allocating a new handler on every run

sentry-android-core/src/main/java/io/sentry/android/core/anr/AnrProfilingIntegration.java
```diff
@@ -0,0 +1,244 @@
...
+public class AnrProfilingIntegration
+    implements Integration, Closeable, AppState.AppStateListener, Runnable {
+
+  public static final long POLLING_INTERVAL_MS = 66;
+  private static final long THRESHOLD_SUSPICION_MS = 1000;
+  public static final long THRESHOLD_ANR_MS = 4000;
+  static final int MAX_NUM_STACKS = (int) (10_000 / POLLING_INTERVAL_MS);
+
+  private final AtomicBoolean enabled = new AtomicBoolean(true);
+  private final Runnable updater = () -> lastMainThreadExecutionTime = SystemClock.uptimeMillis();
```
romtsn, line 39: I realized we use `AndroidCourrentDateProvider` instead of SystemClock.uptimeMillis directly, but I'm fine with using systemclock directly since you can override it in tests easily

sentry-android-core/src/main/java/io/sentry/android/core/anr/AnrProfilingIntegration.java
```diff
@@ -0,0 +1,244 @@
...
+
+    // if we are suspicious, we need to collect stack traces
+    if (mainThreadState == MainThreadState.SUSPICIOUS
+        || mainThreadState == MainThreadState.ANR_DETECTED) {
+      if (numCollectedStacks.get() < MAX_NUM_STACKS) {
+        final long start = SystemClock.uptimeMillis();
+        final @NotNull AnrStackTrace trace =
+            new AnrStackTrace(System.currentTimeMillis(), mainThread.getStackTrace());
+        final long duration = SystemClock.uptimeMillis() - start;
+        if (logger.isEnabled(SentryLevel.DEBUG)) {
```
romtsn, line 241:
```suggestion
        if (logger.isEnabled(SentryLevel.DEBUG)) {
          final long duration = SystemClock.uptimeMillis() - start;
```

romtsn: btw, does `mainThread.getStackTrace()` affects the main thread? or is this log just for debugging purposes?

sentry-android-core/src/main/java/io/sentry/android/core/anr/AnrProfilingIntegration.java
```diff
@@ -0,0 +1,244 @@
...
+  protected AnrProfileManager getProfileManager() {
+    try (final @NotNull ISentryLifecycleToken ignored = profileManagerLock.acquire()) {
+      if (profileManager == null) {
+        final @NotNull SentryOptions opts =
+            Objects.requireNonNull(options, "Options can't be null");
+        final @Nullable String cacheDirPath = opts.getCacheDirPath();
+        if (cacheDirPath == null) {
+          throw new IllegalStateException("cacheDirPath is required for ANR profiling");
+        }
+        final @NotNull File currentFile =
```
romtsn, line 281: alright, I think I found a way this will cause ANRs (similar to replay) - if `close()` is called from the main thread, we'll be holding the lock, awaiting on the operations here to finish, which can take a while due to I/O in `getFileForRecording` and `AnrProfileManager.<init>`. 

I guess to make it non-blocking we'd still have to make QueueFile instantiation lazy and move `getFileForRecording` inside the lazy block, wdyt?

sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
```diff
@@ -777,5 +808,145 @@ private void setAnrExceptions(
...
+      if (stack.length > 0) {
+        final StackTraceElement stackTraceElement = stack[0];
+        final String message =
+            stackTraceElement.getClassName() + "." + stackTraceElement.getMethodName();
+        final ApplicationNotResponding exception = new ApplicationNotResponding(message);
+        exception.setStackTrace(stack);
+
+        final Mechanism mechanism = new Mechanism();
+        mechanism.setType("ANR");
+        final ExceptionMechanismException error =
```
romtsn, line 886: I think this one turns to be unused, shall we pass it to `getSentryExceptions`?

sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
markushi: Done ✅

sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
markushi: will be tracked here: https://github.com/getsentry/sentry-java/issues/5123

sentry-android-core/src/main/java/io/sentry/android/core/anr/AnrProfilingIntegration.java
markushi: Yeah it does impact the main thread (it needs to reach a safepoint, in order to collect the stacktrace) , but the impact from my research should be `<1ms`, which at the time of a main thread block / hiccup should be negligible.

sentry-android-core/src/main/java/io/sentry/android/core/anr/AnrProfilingIntegration.java
markushi: Ah yeah, good catch it's fixed now.

sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
```diff
@@ -777,5 +808,145 @@ private void setAnrExceptions(
...
+              final String module = frame.getModule();
+              if (module != null && !AnrCulpritIdentifier.isSystemFrame(module)) {
+                return false;
+              }
+            }
+          }
+        }
+      }
+      return true;
+    }
```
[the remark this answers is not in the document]
markushi: this is fine

sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
```diff
@@ -777,5 +808,145 @@ private void setAnrExceptions(
...
+      final AggregatedStackTrace culprit = AnrCulpritIdentifier.identify(anrProfile.stacks);
+      if (culprit == null) {
+        return;
+      }
+
+      // Capture profile chunk
+      final @Nullable SentryId profilerId = captureAnrProfile(anrTimestamp, anrProfile);
+      final @NotNull StackTraceElement[] stack = culprit.getStack();
+
+      if (stack.length > 0) {
```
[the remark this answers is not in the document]
markushi: this is fine, context `profiler_id` expects the actual chunk

sentry-android-core/src/main/java/io/sentry/android/core/anr/AnrProfileManager.java
```diff
@@ -0,0 +1,103 @@
...
+                @Override
+                public void toStream(
+                    final @NotNull AnrStackTrace value, final @NotNull OutputStream sink)
+                    throws IOException {
+                  try (final @NotNull DataOutputStream dos = new DataOutputStream(sink)) {
+                    value.serialize(dos);
+                    dos.flush();
+                    sink.flush();
+                  }
+                }
```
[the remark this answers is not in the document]
markushi: This is fine, as tape is using a ByteArrayStream internally which no-ops for close calls. See https://github.com/getsentry/sentry-java/blob/8b8369f06cbc5a9738de7810b1df5863b3ac6bcb/sentry/src/main/java/io/sentry/cache/tape/FileObjectQueue.java#L56-L60

PR #4930, opened 2025-11-24
fix(examples): raise proper SIGSEV instead of undefined behavior

markushi:
## Description
Fix native sample to raise a proper SIGSEV signal instead of causing undefined behavior.

## Motivation and Context
The current implementation causes undefined behavior in release builds, causing an endless loop of "Sending message." instead of actually producing a SIGSEV.

## How did you test it?
Verified the native sample raises SIGSEV correctly.

## Checklist

#skip-changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+### Improvements
+
+- Avoid forking `rootScopes` for Reactor if current thread has `NoOpScopes` ([#4793](https://github.com/getsentry/sentry-java/pull/4793))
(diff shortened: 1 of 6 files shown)


markushi:
Thank you @supervacuus for root-causing this!

supervacuus:
Just to be clear: my point wasn't necessarily to replace the "intentional UB" with an explicit user-space `raise`, but especially for a sample that only wants to trigger a native "crash" to demo its handling, that can be a solid alternative 👍

romtsn (approved):

PR #4933, opened 2025-11-25
feat: minimal tombstone integration

supervacuus:
## :scroll: Description

As discussed last week with @markushi, the process will be to make minimal atomic changes in each PR and merge directly to `main` rather than accumulating on an uber feature branch. This allows for easier review/feedback/corrections, and we can already test a subset of the entire feature "in the field".

The first minimal change is a basic tombstone integration:

* exposes _internal_ option to enable/disable (disabled by default)
* the integration will only ever be enabled if the runtime system is at least Android 12
* the current setup suggests an operation where users will have to disable NDK support (or only depend on `sentry-android-core`) or get two reports for the same crash
* ~only retrieves the most current tombstone (the current implementation is not entirely correct: we should either remove any remaining `ApplicationExitInfo` entries with `REASON_CRASH_NATIVE` or report them too, including the option for the latter; I left this out for minimal interface exposure in the first step, but either variant is easy to add to this PR or later)~
* adds a basic tombstone parser/decoder and accompanying snapshot test
* introduces the following dependencies:
  * runtime: `protobuf-javalite` (the entire features adds ca. 75KiB to the Android sample _release_ APK)
  * build: the `protobuf` plugin and the `protoc` compiler to automate protocol updates

_Open Issues:_
* ~While the `protobuf` runtime is relatively small, there is still the possibility of conflicting with client-side `protobuf` versions (major versions often introduce quite severe breakage, but I haven't tested this yet, only reviewed change logs).~
* ~finding clarity in how to proceed with older tombstones (ignore or handling similarly to `ANRv2`)~
* ~decision if this minimal setup already makes sense to release as an internal API~
* ~add `ManifestMetadataReader` to configure conveniently? (or not since the corresponding options are only internal?)~ deferred as long as the integration flags in the options are "internal"
* No changes yet to the UI (there are quite a few aspects that would make these stack-traces more readable, from my pov this is currently out of scope, but i am writing stuff down)

First sensible release step for https://github.com/getsentry/sentry-java/issues/3295

Part of https://linear.app/getsentry/project/tombstone-support-android-0024cb6e3ffd/

* Added a basic parser snapshot test for a serialized tombstone protobuf.
* Adapted and abstracted tests `ApplicationExitInfo` integrations and `EventProcessor`.
* Extended `EnvelopeCacheTest` with handling of previous sessions wrt `ApplicationExitInfo.REASON_CRASH_NATIVE`.
* Manual testing.

## :pencil: Checklist

## :crystal_ball: Next steps

* Decide if we want to go forward with client-side tombstone processing. _This decision should happen with this PR_.
  * ~If no, consider sending a tombstone as an attachment, similar to how we attach minidumps and let ingestion/processing deal with decoding (I can prototype this in parallel, since I think we can do most of it in relay, as a first step)~
  * If yes, 
    * ~decide on adding the `protobuf` runtime library as a dependency to `sentry-android-core` (or shade/relocate, or implement our own decoder, given that this is a stable format which only requires a subset of `protobuf`).~
    * ~Depending on feedback, adaptation to this minimal change (primarily handling older tombstones).~ 
    * Otherwise, integrate an `EventProcessor` that merges crash events from `sentry-native` with tombstones.
    
#skip-changelog (for now)

### DIFF
--- a/sentry/src/test/java/io/sentry/protocol/SentryStackTraceSerializationTest.kt
+++ b/sentry/src/test/java/io/sentry/protocol/SentryStackTraceSerializationTest.kt
@@ -8,6 +8,7 @@ import io.sentry.JsonSerializable
 import java.io.StringReader
 import java.io.StringWriter
 import kotlin.test.assertEquals
+import kotlin.test.assertTrue
 import org.junit.Test
 import org.mockito.kotlin.mock
 
@@ -58,6 +59,30 @@ class SentryStackTraceSerializationTest {
     val actual = deserialize(expectedJson)
     val actualJson = serialize(actual)
     assertEquals(expectedJson, actualJson)
+  }
+
+  @Test
+  fun `instructionAddressAdjustment serializes and deserializes correctly`() {
+    val testCases =
+      listOf(
+        SentryStackTrace.InstructionAddressAdjustment.AUTO to "auto",
+        SentryStackTrace.InstructionAddressAdjustment.ALL to "all",
+        SentryStackTrace.InstructionAddressAdjustment.ALL_BUT_FIRST to "all_but_first",
+        SentryStackTrace.InstructionAddressAdjustment.NONE to "none",
+      )
+
+    for ((enumValue, expectedJson) in testCases) {
+      val stackTrace = SentryStackTrace().apply { instructionAddressAdjustment = enumValue }
+      val json = serialize(stackTrace)
+
+      assertTrue(
+        json.contains("\"instruction_addr_adjustment\":\"$expectedJson\""),
+        "Expected $enumValue to serialize as \"$expectedJson\"",
+      )
+
+      val deserialized = deserialize(json)
+      assertEquals(enumValue, deserialized.instructionAddressAdjustment)
+    }
   }
 
   // Helper


sentry-android-core/src/main/java/io/sentry/android/core/TombstoneIntegration.java
```diff
@@ -0,0 +1,197 @@
...
+      // search for the latest Tombstone to report it separately as we're gonna enrich it. The
+      // latest
+      // Tombstone will be first in the list, as it's filled last-to-first in order of appearance
+      ApplicationExitInfo latestTombstone = null;
+      for (ApplicationExitInfo applicationExitInfo : exitInfos) {
+        if (applicationExitInfo.getReason() == ApplicationExitInfo.REASON_CRASH_NATIVE) {
+          latestTombstone = applicationExitInfo;
+          // remove it, so it's not reported twice
+          // TODO: if we fail after this, we effectively lost the ApplicationExitInfo (maybe only remove after we reported it)
+          exitInfos.remove(applicationExitInfo);
```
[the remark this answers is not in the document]
supervacuus: This is intentionally marked as `TODO`, so we can discuss handling history in this PR.

sentry-android-core/src/main/java/io/sentry/android/core/TombstoneIntegration.java
```diff
@@ -0,0 +1,197 @@
...
+      }
+
+      reportAsSentryEvent(latestTombstone);
+    }
+
+    @RequiresApi(api = Build.VERSION_CODES.R)
+    private void reportAsSentryEvent(ApplicationExitInfo exitInfo) {
+      SentryEvent event;
+      try {
+        TombstoneParser parser = new TombstoneParser(exitInfo.getTraceInputStream());
```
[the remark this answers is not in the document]
supervacuus: Agreed, but also left it open, since I don't know yet how the general propagation of errors is handled in integration `run`s (top-level handler? silent/logging discard? ...?).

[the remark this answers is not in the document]
supervacuus: See: https://github.com/getsentry/sentry-java/pull/4933#discussion_r2559336570

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -372,6 +373,10 @@ static void installDefaultIntegrations(
     final Class<?> sentryNdkClass = loadClass.loadClass(SENTRY_NDK_CLASS_NAME, options.getLogger());
     options.addIntegration(new NdkIntegration(sentryNdkClass));
 
+    if (buildInfoProvider.getSdkInfoVersion() >= Build.VERSION_CODES.R) {
```
supervacuus: Not entirely sure about this one:

* not sure if we want to add the integration conditionally here, or instead only run it conditionally (with a log that explains that the integration isn't really running)
* independent of where we check this: while it is true that `REASON_CRASH_NATIVE` is available with `R`, [the tombstone will not be available before `S`](https://developer.android.com/reference/android/app/ApplicationExitInfo#getTraceInputStream()).
* We can ignore the latter, since the API _is_ available with `R` and we have to handle a `null` `InputStream` anyway.

supervacuus:
@markushi, I just realized I cannot omit having a separate "tombstone" marker, even if I report all events, without repeating them. I mean, this was clear to me, but, in addition to it being a must for this PR already, unlike the ANR marker, I must also align it with `crashedLastRun`. 

I wonder if it would make the most sense to do it similarly to `ANR` by using a `TombstoneHint` to ensure writing the marker timestamp at the correct life-cycle stage of the event, but introduce a new interface analog to `AbnormalExit` called `CrashExit`, so that the hint can take a different route (since it isn't truly an abnormal exit and should affect similar paths to the Native SDK crash marker).

The biggest issue with that approach is that the `crashedLastRun` is handled entirely in `EnvelopeCache` rather than `AndroidEnvelopeCache`, whereas `ApplicationExitInfo` hint/marker handling happens in the latter.

The PR still makes sense for a first review from you (since, if the general direction makes sense and you have todos not in my list, I can also add a test for the integration itself). Still, I would appreciate a short sync on how to align these execution paths (maybe I don't need to align tombstones with `crashedLastRun` and can abuse `AbnormalExit` for the same outcome, though it feels like a significant shortcut, even for minimal internal exposure, which is not at all what I aimed for).

I can convert the PR back to a draft if you prefer.

sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/TombstoneParser.java
```diff
@@ -0,0 +1,195 @@
...
+import io.sentry.protocol.SentryThread;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+
+public class TombstoneParser {
```
markushi: If we don't want classes to be consumed outside our SDK we usually mark them as Internal 
```suggestion
@ApiStatus.Internal
public class TombstoneParser {
```

sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/TombstoneParser.java
```diff
@@ -0,0 +1,195 @@
...
+
+  private final InputStream tombstoneStream;
+  private final Map<String, String> excTypeValueMap = new HashMap<>();
+
+  public TombstoneParser(InputStream tombstoneStream) {
```
markushi: We usually are very keen on marking as much as possible as `final` as well as annotating any non-primitive method parameters as `@Nullable` or `@NotNull`
```suggestion
  public TombstoneParser(final @NotNull InputStream tombstoneStream) {
```

sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/TombstoneParser.java
markushi: I guess there are different ways to deal with closing the input stream at some point, a pattern we've used in the past would be to implement the `Closable` interface and then initiate the class using a try-with-resources wrapping call.
```suggestion
public class TombstoneParser implements Closable {
```

sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/TombstoneParser.java
```diff
@@ -0,0 +1,195 @@
...
+
+      SentryThread thread = new SentryThread();
+      thread.setId(Long.valueOf(threadEntry.getKey()));
+      thread.setName(threadEntry.getValue().getName());
+
+      SentryStackTrace stacktrace = createStackTrace(threadEntry);
+      thread.setStacktrace(stacktrace);
+      if (tombstone.getTid() == threadEntry.getValue().getId()) {
+        thread.setCrashed(true);
+        // even though we refer to the thread_id from the exception,
```
markushi, line 83: I'd say that's the perfect level of code comments I'd like to see, nice!

sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/TombstoneParser.java
```diff
@@ -0,0 +1,195 @@
...
+
+    SentryStackTrace stacktrace = new SentryStackTrace();
+    stacktrace.setFrames(frames);
+
+    Map<String, Object> unknown = new HashMap<>();
+    // `libunwindstack` used for tombstones already applies instruction address adjustment:
+    // https://android.googlesource.com/platform/system/unwinding/+/refs/heads/main/libunwindstack/Regs.cpp#175
+    // prevent "processing" from doing it again.
+    unknown.put("instruction_addr_adjustment", "none");
+    stacktrace.setUnknown(unknown);
```
markushi: `unknown` is mainly used for hybrid SDKs or when the Java SDK is used as a sending mechanism to ensure no data is lost due to parsing. In this case I'd favor extending the `SentryStackTrace` class with a `private @Nullable String instructionAddressAdjustment` field instead.

sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/TombstoneParser.java
```diff
@@ -0,0 +1,195 @@
...
+  private List<SentryException> createException(TombstoneProtos.Tombstone tombstone) {
+    TombstoneProtos.Signal signalInfo = tombstone.getSignalInfo();
+
+    SentryException exception = new SentryException();
+    exception.setType(signalInfo.getName());
+    exception.setValue(excTypeValueMap.get(signalInfo.getName()));
+    exception.setMechanism(createMechanismFromSignalInfo(signalInfo));
+    exception.setThreadId((long) tombstone.getTid());
+
+    List<SentryException> exceptions = new ArrayList<>();
```
markushi: nit: since we know the list will always contain one item we could supply the initial size here
```suggestion
    List<SentryException> exceptions = new ArrayList<>(1);
```

sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/TombstoneParser.java
```diff
@@ -0,0 +1,195 @@
...
+
+  @NonNull
+  private Message constructMessage(TombstoneProtos.Tombstone tombstone) {
+    Message message = new Message();
+    TombstoneProtos.Signal signalInfo = tombstone.getSignalInfo();
+
+    // reproduce the message `debuggerd` would use to dump the stack trace in logcat
+    message.setFormatted(
+        String.format(
+            Locale.getDefault(),
```
markushi: IMHO we should use `Locale.ROOT` here instead.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
markushi: As without tombstones the reported events won't be any useful I suggest to check against `S` instead,  `>= Build.VERSION_CODES.S`

sentry-android-core/src/main/java/io/sentry/android/core/TombstoneIntegration.java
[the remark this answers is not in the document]
markushi: Yeah good point, for ANRs we write a timestamp to disk ([impl](https://github.com/getsentry/sentry-java/blob/3258517af4c60fb19d8a93cdb148305c735ebf63/sentry-android-core/src/main/java/io/sentry/android/core/cache/AndroidEnvelopeCache.java#L155-L176)) to ensure we don't double report on the next app launch. @romtsn what do you think about this? I guess what's to discuss here:
- How many days back should we consider crashes?
the 90 days threshold sounds good to me

- Should we report them all?
In my opinion yes. TBD: How to deal with persisted scope data for enrichment?

- How to deal with unprocessable tombstones?
In my opinion we should skip them, otherwise the processing will get stuck until the crash is out of the 90 day window. TBD: should we report those as dropped events?

sentry-android-core/src/main/java/io/sentry/android/core/TombstoneIntegration.java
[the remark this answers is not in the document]
markushi: We usually use a wide try/catch and silently swallow those errors in combination with logging the error using `options.getLogger()`

sentry-android-core/src/main/proto/io/sentry/android/core/internal/tombstone/tombstone.proto
```diff
@@ -0,0 +1,204 @@
+// Added and adapted from: https://android.googlesource.com/platform/system/core/+/refs/heads/main/debuggerd/proto/tombstone.proto
```
markushi, line 1: We should also add some license attribution here, [e.g. like we did here](https://github.com/getsentry/sentry-java/blob/86f00968b4ba59f756bb4fe01e4fb1cc61bd4f1d/sentry/src/main/java/io/sentry/vendor/Base64.java#L1-L17).

gradle/libs.versions.toml
```diff
@@ -138,6 +140,8 @@ otel-javaagent-extension-api = { module = "io.opentelemetry.javaagent:openteleme
 otel-semconv = { module = "io.opentelemetry.semconv:opentelemetry-semconv", version.ref = "otelSemanticConventions" }
 otel-semconv-incubating = { module = "io.opentelemetry.semconv:opentelemetry-semconv-incubating", version.ref = "otelSemanticConventionsAlpha" }
 p6spy = { module = "p6spy:p6spy", version = "3.9.1" }
+protobuf-javalite = { module = "com.google.protobuf:protobuf-javalite", version.ref = "protobuf"}
```
markushi, line 144: @romtsn Not sure if you followed the conversations, but as you can see here, protobuf requires a runtime dependency.  It will have an impact of around 10kb. IMHO fine for now, we should still check how stable this library is to avoid and consumer version mismatch issues.

sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/TombstoneParser.java
supervacuus: Yeah, I already have a change in progress that annotates this more strictly.

supervacuus: Yeah, this makes sense. 

I haven't been closing anything in the integration yet because I wasn't sure about the semantics of the trace `InputStream`. The API docs mention that the system ring buffer evicts tombstones, which made me a bit wary RE: statefulness; however, after reading the code, it turns out that only means the returned `InputStream` could be `null` (which it can be for many reasons actually; curious choice to highlight the tombstone rotation). 

* Each returned trace stream is wrapped as an `AutoCloseInputStream` [on a fresh parcel fd](https://android.googlesource.com/platform/frameworks/base/%2B/HEAD/core/java/android/app/ApplicationExitInfo.java#839)
* The tombstones the app searches through were [retrieved by `system_server`](https://android.googlesource.com/platform/frameworks/base.git/%2B/master/services/core/java/com/android/server/os/NativeTombstoneManager.java#284) via a file watcher on `/data/tombstones` 
* and the rotation mentioned in the API docs only happens in the [`CrashQueue` of `tombstoned` via `unlinkat` and `linkat`](https://android.googlesource.com/platform//system/core/%2B/92ca277fb7c7c9d504ad68e415d5397d71388579/debuggerd/tombstoned/tombstoned.cpp#410).

All of the above operate on the same set of files (shared between system daemons and _all_ apps, where `binder` delivery per app/user is decided in the `system_server` module linked above), however, we still operate in classic UNIX territory, aka `unlink` removes the path name but not the inode if there are open fds.

sentry-android-core/src/main/java/io/sentry/android/core/AnrV2EventProcessor.java
```diff
@@ -564,12 +576,16 @@ private void setExceptions(final @NotNull SentryEvent event, final @NotNull Obje
...
+      // TODO: this currently overrides the signalhandler mechanism we set in the TombstoneParser
+      //       with a new type (can be the right choice down the road, but currently is
+      //       unintentional, and it might be better to leave it close to the Native SDK)
       // we only enrich the latest ANR in the list, so this is historical
       mechanism.setType("HistoricalAppExitInfo");
     } else {
       mechanism.setType("AppExitInfo");
     }
 
+    // TODO: this currently overrides the tombstone exceptions
```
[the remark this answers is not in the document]
supervacuus: What is the value of reporting my own code-level remarks back to me? Incredibly unhelpful and noisy.

sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
```diff
@@ -715,6 +663,106 @@ private void mergeOS(final @NotNull SentryBaseEvent event) {
...
+    boolean supports(@NotNull Object hint);
+
+    void applyPreEnrichment(
+        @NotNull SentryEvent event, @NotNull Backfillable hint, @NotNull Object rawHint);
+
+    void applyPostEnrichment(
+        @NotNull SentryEvent event, @NotNull Backfillable hint, @NotNull Object rawHint);
+  }
+
+  private final class AnrHintEnricher implements HintEnricher {
```
markushi, line 672: That sounds like a pretty good solution to me, nice!

sentry-android-core/src/main/java/io/sentry/android/core/TombstoneEventProcessor.java
```diff
@@ -0,0 +1,26 @@
...
+import android.content.Context;
+import androidx.annotation.WorkerThread;
+import io.sentry.BackfillingEventProcessor;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+/** originating from the Native SDK. */
+@ApiStatus.Internal
+@WorkerThread
+public class TombstoneEventProcessor implements BackfillingEventProcessor {
```
markushi: I guess we can get rid of this then in favor of `ApplicationExitInfoEventProcessor` and the hint based enrichers in there.

supervacuus: Yes, this is an obsolete leftover of my initial attempt. However, once we connect the Native SDK with tombstones, a separate `EventProcessor` could make sense again.

sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoHistoryDispatcher.java
```diff
@@ -0,0 +1,245 @@
...
+  private @Nullable ApplicationExitInfo removeLatest(
+      final @NotNull List<ApplicationExitInfo> exitInfos) {
+    for (ApplicationExitInfo applicationExitInfo : exitInfos) {
+      if (applicationExitInfo.getReason() == policy.getTargetReason()) {
+        exitInfos.remove(applicationExitInfo);
+        return applicationExitInfo;
+      }
+    }
+    return null;
+  }
```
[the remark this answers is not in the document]
supervacuus: Not if we exit the loop right after the removal. This is verbatim from the initial code. If we want to remove a potential foot gun RE: future changes, I can make the iterator explicit so that the removal updates both the iterator and the data structure.

sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
```diff
@@ -699,6 +665,118 @@ private void mergeOS(final @NotNull SentryBaseEvent event) {
...
+        // exception is properly created as well, but without stacktrace
+        mainThread = new SentryThread();
+        mainThread.setStacktrace(new SentryStackTrace());
+      }
+      event.setExceptions(
+          sentryExceptionFactory.getSentryExceptionsFromThread(mainThread, mechanism, anr));
+    }
   }
-  // endregion
 }
```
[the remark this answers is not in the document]
supervacuus: That is 100% intentional from my side, at least for now.

supervacuus:
@markushi and @romtsn, I think I also found an okay solution to reduce duplication between the tests for the two AEI integrations. That means the first three "milestones" required for a release (without Native SDK event enrichment/merging) are finished. Please let me know if something is missing from your POV or not yet up to par, and how you would like to proceed.

Btw, multiple `TODO` entries in the PR don't necessarily highlight a vital change still open, but do signal a decision we probably should have made in this PR. Please understand them as review guidance for open questions/decisions I still have.

sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/TombstoneParser.java
```diff
@@ -0,0 +1,222 @@
...
+        tombstone.getThreadsMap().entrySet()) {
+      final TombstoneProtos.Thread threadEntryValue = threadEntry.getValue();
+
+      final SentryThread thread = new SentryThread();
+      thread.setId(Long.valueOf(threadEntry.getKey()));
+      thread.setName(threadEntryValue.getName());
+
+      final SentryStackTrace stacktrace = createStackTrace(threadEntryValue);
+      thread.setStacktrace(stacktrace);
+      if (tombstone.getTid() == threadEntryValue.getId()) {
```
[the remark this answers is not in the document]
supervacuus: That is also intentional, because the key and the ID should be equivalent.

sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/TombstoneParser.java
```diff
@@ -0,0 +1,222 @@
...
+    final Mechanism mechanism = new Mechanism();
+    // this follows the current processing triggers strictly, changing any of these
+    // alters grouping and name (long-term we might want to have a tombstone mechanism)
+    // TODO: if we align this with ANRv2 this would be overwritten in a BackfillingEventProcessor as
+    //       `ApplicationExitInfo` not sure what the right call is. `ApplicationExitInfo` is
+    //       certainly correct. But `signalhandler` isn't wrong either, since all native crashes
+    //       retrieved via `REASON_CRASH_NATIVE` will be signals. I am not sure what the side-effect
+    //       in ingestion/processing will be if we change the mechanism, but initially i wanted to
+    //       stay close to the Native SDK.
+    mechanism.setType("signalhandler");
```
supervacuus: This `TODO` asks whether to use the mechanism type `signalhandler` (as we do in the Native SDK) or to change it. If we decide on the latter, should we do it in this PR now? What is the trade-off (what do we gain from `ApplicationExitInfo` as a mechanism type vs what do we lose when dropping `signalhandler`?

sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
```diff
@@ -699,6 +665,118 @@ private void mergeOS(final @NotNull SentryBaseEvent event) {
...
+  }
+
+  private final class AnrHintEnricher implements HintEnricher {
+
+    @Override
+    public boolean supports(@NotNull Object hint) {
+      // TODO: not sure about this. all tests are written with AbnormalExit, but enrichment changes
+      //  are ANR-specific. I called it AnrHintEnricher because that is what it does, but it
+      //  actually triggers on AbnormalExit. Let me know what makes most sense.
+      return hint instanceof AbnormalExit;
```
supervacuus: This is primarily about naming things: the class is called `AnrHintEnricher` (i.e., specific for ANR), but we discriminate based on `AbnormalExit` (which is arguably broader). 

But it is also about "correctness": right now, the tests related to the event processor all use `AbnormalExit` when constructing test events.

sentry/src/main/java/io/sentry/cache/EnvelopeCache.java
```diff
@@ -122,13 +122,17 @@ private boolean storeInternal(final @NotNull SentryEnvelope envelope, final @Not
...
 
     if (HintUtils.hasType(hint, SessionStart.class)) {
       movePreviousSession(currentSessionFile, previousSessionFile);
       updateCurrentSession(currentSessionFile, envelope);
 
       boolean crashedLastRun = false;
       final File crashMarkerFile = new File(options.getCacheDirPath(), NATIVE_CRASH_MARKER_FILE);
+      // TODO: this should probably check whether the Native SDK integration is currently enabled or
+      // remove the marker file if it isn't. Otherwise, application that disable the Native SDK,
+      // will report a crash for the last run forever.
```
supervacuus: This is unrelated to this PR, but it can serve as a trigger to fix it: if the native SDK was once enabled and then disabled, do we have a cleanup process elsewhere that would remove the old native marker? If the marker's lifecycle is managed only by the Native SDK, we would report `crashedLastRun == true` forever.

sentry/src/main/java/io/sentry/cache/EnvelopeCache.java
```diff
@@ -122,13 +122,17 @@ private boolean storeInternal(final @NotNull SentryEnvelope envelope, final @Not
     if (HintUtils.hasType(hint, AbnormalExit.class)) {
       tryEndPreviousSession(hint);
     }
+    // TODO: adapt       tryEndPreviousSession(hint); to CrashExit.class
```
supervacuus: likely relevant for this PR

sentry/src/main/java/io/sentry/cache/EnvelopeCache.java
```diff
@@ -153,6 +157,7 @@ private boolean storeInternal(final @NotNull SentryEnvelope envelope, final @Not
         }
       }
 
+      // TODO: maybe set crashLastRun for tombstone too?
```
supervacuus: likely relevant for this PR

sentry/src/main/java/io/sentry/protocol/SentryStackTrace.java
```diff
@@ -66,6 +66,16 @@ public final class SentryStackTrace implements JsonUnknown, JsonSerializable {
    */
   private @Nullable Boolean snapshot;
 
+  /**
+   * This value indicates if, and how, `instruction_addr` values in the stack frames need to be
+   * adjusted before they are symbolicated. TODO: should we make this an enum or is a string value
+   * fine?
```
supervacuus: Since this is part of the public API, it is also a sensible decision to make now.

sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/TombstoneParser.java
markushi: I think `signalhandler` is the best choice here, as it also makes a potential upgrade path easier (e.g. existing sentry.io dashboard queries will continue to work when switching from sentry-native to ApplicationExitInfo)

sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
markushi: I'd keep it as-is for compatibility reasons, maybe some javadoc on `AbnormalExit` could help to make it more clear.

sentry/src/main/java/io/sentry/cache/EnvelopeCache.java
markushi: We do some cleanup [within sentry core here](https://github.com/getsentry/sentry-java/blob/ca5593ebb79159c3c9db33439bc5974f41a99b9a/sentry/src/main/java/io/sentry/PreviousSessionFinalizer.java#L99), regardless if the native SDK is initialized or not.

sentry/src/main/java/io/sentry/protocol/SentryStackTrace.java
markushi: Yes, we should introduce an enum with the values outlined in https://github.com/timfish/sentry-rfcs/blob/main/text/0043-instruction-addr-adjustment.md#proposed-protocol-extension

gradle/libs.versions.toml
romtsn: as discussed, it should be fine! worst case we can vendor/shadow it later if we see problems with the dependency (size is not a concern here imo)

sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/TombstoneParser.java
supervacuus: Yeah, that is why I stayed with `signalhandler` for now. I don't see any apparent reason to change it at this point either.

sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
supervacuus: Added a note to `AbnormalExit` javadoc. Keeping the `AbnormalExit` discriminator for the enricher.

sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/TombstoneParser.java
```diff
@@ -0,0 +1,216 @@
...
+  @NonNull
+  private List<SentryThread> createThreads(
+      @NonNull final TombstoneProtos.Tombstone tombstone, @NonNull final SentryException exc) {
+    final List<SentryThread> threads = new ArrayList<>();
+    for (Map.Entry<Integer, TombstoneProtos.Thread> threadEntry :
+        tombstone.getThreadsMap().entrySet()) {
+      final TombstoneProtos.Thread threadEntryValue = threadEntry.getValue();
+
+      final SentryThread thread = new SentryThread();
+      thread.setId(Long.valueOf(threadEntry.getKey()));
```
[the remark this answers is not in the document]
supervacuus: @markushi and @romtsn: what do you think? I already responded here: https://github.com/getsentry/sentry-java/pull/4933#discussion_r2603048212.

I personally favor cross-checking for identity when the data structure supports it (i.e., not only checking against self-assigned values but also against values that should be the same, such as when the key of a map equals an identity property of the value). But I have no strong feelings, and following this advice could prevent future en masse false-positive review suggestions.

sentry/src/main/java/io/sentry/cache/EnvelopeCache.java
supervacuus: Yeah, but this code only executes if a valid `previous_session.json` exists, which in my case doesn't necessarily overlap. I am not sure whether this is a significant enough edge case, but this is repeatable for:

* `io.sentry.ndk.enable = true` **and** `io.sentry.tombstone.enable = false` (the configuration we expect for every user before they enable tombstone for the first time)
* run application and native crash (i.e., sentry-native registers the crash and writes a crash marker)
* `io.sentry.ndk.enable = false` **and** `io.sentry.tombstone.enable = true` (the configuration we expect when users release their apps with tombstone support instead of sentry-native based native crash reporting)
* On the next start (and every start after that), the `EnvelopeCache` will see the native crash marker, because nothing lifecycles the crash marker anymore

I am not saying we should necessarily lifecycle the native crash marker from the `EnvelopeCache`, but _somewhere_ since after the first report, it is clearly a false positive. Of course, it is also fair to say: this only happens when a native crash occurs, and there is no restart before the app is updated with tombstone enabled and the Native SDK disabled (so, a relatively rare sequence). But it is a persistent error.

supervacuus: Just to be clear: when the `EnvelopeCache` reads the crash marker to set `crashedLastRun`, the `previous_session.json` still exists. But by the time marker cleanup happens in the `PreviousSessionFinalizer`, it is no longer there. So, that code never gets executed for this particular path.

sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/TombstoneParser.java
```diff
@@ -0,0 +1,216 @@
...
+
+  @NonNull
+  private static SentryStackTrace createStackTrace(@NonNull final TombstoneProtos.Thread thread) {
+    final List<SentryStackFrame> frames = new ArrayList<>();
+
+    for (TombstoneProtos.BacktraceFrame frame : thread.getCurrentBacktraceList()) {
+      final SentryStackFrame stackFrame = new SentryStackFrame();
+      stackFrame.setPackage(frame.getFileName());
+      stackFrame.setFunction(frame.getFunctionName());
+      stackFrame.setInstructionAddr(String.format("0x%x", frame.getPc()));
```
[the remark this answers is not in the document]
supervacuus: The reason is straightforward: `%x` format specifiers are locale-invariant. Whereas `%d` is locale-dependent.

supervacuus:
Btw, I have no clue why the tests suddenly fail. I cannot repro locally. There are timeouts, report trimming/unavailability, but also mock failures. For the latter, I would expect to see them locally.

sentry/src/main/java/io/sentry/cache/EnvelopeCache.java
supervacuus: I have now adapted the `EnvelopeCache` to end the previous session on  `NativeCrashExit` and added some basic tests (https://github.com/getsentry/sentry-java/pull/4933/commits/37e43bb7bd2533ca79304d12726605b6bca36efe). However, the life-cycle of `previous_session.json` is still not entirely clear to me:

* there is a `MovePreviousSession` thread, which starts before any integrations, that creates a `previous_session.json` if there was a `session.json`
* this thread also deletes any `previous_session.json` before that via `EnvelopeCache.movePreviousSession()`
* however, I have yet to see a `previous_session.json` to exist long enough to trigger the clean up in the `PreviousSessionFinalizer` or the change addressed here in `EnvelopeCache.tryEndPreviousSession()` (outside of tests)

This effectively touches on 
* where to update `crashLastRun` when we have a tombstone as evidence for a crashed last run (for instance, can I do that in `tryEndPreviousSession()`, but that would be overwritten with the next `SessionStart`).
* what to do with leftover native crash markers from the native SDK ([see](https://github.com/getsentry/sentry-java/pull/4933#discussion_r2605617510))

sentry/src/main/java/io/sentry/cache/EnvelopeCache.java
romtsn: > But by the time marker cleanup happens in the PreviousSessionFinalizer, it is no longer there. So, that code never gets executed for this particular path.

sorry, would you mind explaining the case when the previous session file is no longer there? I guess I'm missing a bit of context

sentry/src/main/java/io/sentry/cache/EnvelopeCache.java
romtsn: yeah, I think this makes sense, but we can do it in a separate PR as well, if that's easier.

We probably should put that logic into `PreviousSessionFinalizer` though, because that's where the current logic for native crashes lives?

sentry-android-core/src/test/java/io/sentry/android/core/ApplicationExitIntegrationTestBase.kt
```diff
@@ -0,0 +1,435 @@
...
+import org.mockito.kotlin.inOrder
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.never
+import org.mockito.kotlin.times
+import org.mockito.kotlin.verify
+import org.mockito.kotlin.whenever
+import org.robolectric.shadow.api.Shadow
+import org.robolectric.shadows.ShadowActivityManager
+
+abstract class ApplicationExitIntegrationTestBase<THint : Any> {
```
romtsn, line 42: nice abstraction!

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
```diff
@@ -570,6 +604,16 @@ public void setReportHistoricalAnrs(final boolean reportHistoricalAnrs) {
     this.reportHistoricalAnrs = reportHistoricalAnrs;
   }
 
+  @ApiStatus.Internal
```
romtsn, line 618: any specific reason this is internal?

supervacuus: AFAICR, @markushi suggested that we release the first iteration of the feature as an internal API, at least until we integrate Native SDK reports with the tombstone-generated reports.

sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
```diff
@@ -530,26 +539,13 @@ private void setStaticValues(final @NotNull SentryEvent event) {
...
-        if (name != null && name.equals("main")) {
-          return thread;
-        }
-      }
-    }
-    return null;
-  }
-
   // by default we assume that the ANR is foreground, unless abnormalMechanism is "anr_background"
   private boolean isBackgroundAnr(final @NotNull Object hint) {
```
romtsn: I guess we could also move this method down to `AnrHintEnricher?

sentry/src/main/java/io/sentry/cache/EnvelopeCache.java
supervacuus: Besides, in tests, I have not yet observed (or been able to provoke) a situation in which the `PreviousSessionFinalizer` actually cleans up the native crash marker. And the reason for that is that a `previous_session.json` no longer exists (if it ever existed) when `PreviousSessionFinalizer` is invoked.

I might be holding it wrong when using the Android sample.

But the root context for why I am raising this topic is finding the correct location(s) to ensure that any sessions are correctly ended and crashes are correctly reported from the SDK (i.e. `crashedLastRun`). I tried to explain this here: https://github.com/getsentry/sentry-java/pull/4933#discussion_r2623602238

For me, this is the last topic in this PR, where I am not entirely sure what expected behavior is because we have:
* `Envelope.tryEndPreviousSession()` as a mechanism for the ApplicationExitInfo-triggered crashes and abnormal exits to end "previous sessions".
* the special case in `EnvelopeCache.storeInternal()` for `SessionStart` hints that reads the native crash marker or Java crash marker to set `crashedLastRun` and also triggers a "previous session" flush.
* the `PreviousSessionFinalizer` acts as another runnable that finalizes "previous sessions"

And what i currently observe is the following sequence:
1. the `MovePreviousSession` runnable runs first and creates a "previous session" via `movePreviousSession()` from an existing `session.json`.
2. `EnvelopeCache.storeInternal()` runs and receives an envelope with a `SessionStart` hint and also executes `movePreviousSession()`, which now removes the`previous_session.json` and then also doesn't create a new one because the previous invocation of `movePreviousSession()` moved `session.json` to `previous_session.json`
3. any component depending on a `previous_session.json` after that will not execute anything (which affects  `EnvelopeCache.tryEndPreviousSession()` and `PreviousSessionFinalizer.run()`)

sentry/src/main/java/io/sentry/cache/EnvelopeCache.java
supervacuus: That might be the right location, but as explained here: https://github.com/getsentry/sentry-java/pull/4933#discussion_r2630714065 from what i can currently observe `EnvelopeCache.storeInternal()` when called with a `SessionStart` hint, (accidentally?) removes the `previous_session.json`, so any followup finalizers (`EnvelopeCache.tryEndPreviousSession()` and `PreviousSessionFinalizer.run()`) don't really act on the "previous session".

sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
supervacuus: fixed in https://github.com/getsentry/sentry-java/pull/4933/commits/f5de2020397072bb769d1f4b4e69d7377ca1058b

sentry/src/main/java/io/sentry/cache/EnvelopeCache.java
supervacuus: We could address this by adding a third parameter `EnvelopeCache.movePreviousSession()`, for instance, `boolean clean`, where only the call from `MovePreviousSession` cleans any leftovers. The one in `EnvelopeCache.storeInternal()` doesn't remove the obsolete `previous_session.json` file (since we know it is the current one). 

However, in that case, I wonder why we'd call `movePreviousSession()` from `EnvelopeCache.storeInternal()` at all, since the `MovePreviousSession` runnable is triggered by `Sentry.init()`, meaning there should never be a case that requires starting the "previous session" lifecycle from `EnvelopeCache.storeInternal()`.

So, I would love a sanity check: am I missing some context, is this a bug, or did something in this PR specifically change the behavior?

Happy to hop on a quick sync, or I can put together the fix if that’s helpful. I mainly want to confirm what the intended behavior is here.

sentry-android-core/src/main/java/io/sentry/android/core/cache/AndroidEnvelopeCache.java
```diff
@@ -132,9 +126,9 @@ public static boolean hasStartupCrashMarker(final @NotNull SentryOptions options
     final File crashMarkerFile = new File(outboxPath, STARTUP_CRASH_MARKER_FILE);
     try {
       final boolean exists =
-          options.getRuntimeManager().runWithRelaxedPolicy(() -> crashMarkerFile.exists());
+          options.getRuntimeManager().runWithRelaxedPolicy(crashMarkerFile::exists);
```
romtsn: We usually don't use method refs to maintain compatibility with AGP < 3.0 (although, frankly, we should've dropped it long time ago), so I'd suggest to revert this, and we'll convert all of them at once in a new major version of the SDK with a proper changelog entry.

supervacuus: fixed in https://github.com/getsentry/sentry-java/pull/4933/commits/c602ec950be31ae88260b6dded1691e953b0512a

sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/TombstoneParser.java
```diff
@@ -0,0 +1,220 @@
...
+
+    // must use the "native" platform because otherwise the stack-trace wouldn't be correctly parsed
+    event.setPlatform("native");
+
+    event.setMessage(constructMessage(tombstone));
+    event.setDebugMeta(createDebugMeta(tombstone));
+    event.setExceptions(createException(tombstone));
+    if (event.getExceptions() == null || event.getExceptions().isEmpty()) {
+      throw new RuntimeException("Failed to decode exception information from tombstone");
+    }
```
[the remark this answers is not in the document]
supervacuus: Yes, I wrote this check with a previous version of `createException()` in mind. It should be removed.

However, I wouldn't throw here if `hasSignalInfo()` returned `false`. The exception interrupts any further report creation and propagates execution toward the handler in the dispatcher. Instead, if `hasSignalInfo()` returned `false`, we still want to see the crash and context.

sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/TombstoneParser.java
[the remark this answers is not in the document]
romtsn: yeah sounds like just a slop, we can keep it your way!

sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/TombstoneParser.java
```diff
@@ -0,0 +1,219 @@
...
+      // exclude anonymous and non-executable maps
+      if (module.getBuildId().isEmpty()
+          || module.getMappingName().isEmpty()
+          || !module.getExecute()) {
+        continue;
+      }
+      final DebugImage image = new DebugImage();
+      image.setCodeId(module.getBuildId());
+      image.setCodeFile(module.getMappingName());
+      image.setDebugId(module.getBuildId());
```
romtsn: I'm reading this [section](https://develop.sentry.dev/sdk/data-model/event-payloads/debugmeta/#elf-images:~:text=stack%20traces.-,debug_id,-Required.%20Debug%20identifier) and it says:

> the debug identifier is the little-endian UUID
representation of the first 16-bytes of that identifier

should we compute it here or symbolicator actually does that for us? And if it does it for us, should we leave it empty?

sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/TombstoneParser.java
```diff
@@ -0,0 +1,219 @@
...
+
+    return exceptions;
+  }
+
+  @NonNull
+  private static Mechanism createMechanismFromSignalInfo(
+      @NonNull final TombstoneProtos.Signal signalInfo) {
+
+    final Mechanism mechanism = new Mechanism();
+    // this follows the current processing triggers strictly, changing any of these
```
romtsn, line 145: I guess grouping is going to be different anyway because of properly symbolicated stacktraces compared to sentry-native? That said, I would be fine with changing the mechanism to "tombstone" now, even if it breaks existing dashboards/filters. We should just be explicit in the changelog perhaps that this requires new filters etc.? But not strong opinion, signalhandler is also fine (+ we can filter by sdk.name if we wanna whether this comes from sentry-native or the tombstone integration)

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
romtsn: sgtm!

sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/TombstoneParser.java
supervacuus: Leaving it empty seems to discard the event at some point. Writing it straight like I did here seems to work fine. However, I think we should compute it exactly as specified from the SDK (which the Native SDK also does).

sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/TombstoneParser.java
supervacuus: > I guess grouping is going to be different anyway because of properly symbolicated stacktraces compared to sentry-native? 

I didn't mean relative to the native SDK, but in how a native developer would expect stack contents to be considered for grouping and naming. If you change any of those top-level mechanism properties, you will get bizarre exception/stack-trace interpretations.

> That said, I would be fine with changing the mechanism to "tombstone" now, even if it breaks existing dashboards/filters.

I am open to changing the `type`, but if we do, we should clarify what we get from it and what we pay for. I am no longer sure what happened when changing the `type`, if anything. Do users actually filter on the mechanism?

_The primary question for me is: what do we lose if we keep `type` as `"signalhandler"` for now and defer that decision until we integrate reports from the Native SDK?_

romtsn: yeah the only benefit I see is to be able to distinguish where the event is coming from (sentry-native or AppExitInfo), but as I mentioned we could use sdk.name as a proxy for that (it will be `sentry.java.android` for tombstones) so I have no rush to make a decision here and we can keep it at `signalhandler` 👍

sentry/src/main/java/io/sentry/cache/EnvelopeCache.java
romtsn: this is indeed a bug 🙈 I have introduced it here while adressing the PR review: https://github.com/getsentry/sentry-java/pull/4624/commits/fb599673b7d2b03df54ea3debfc1ab9ec04f22c8 but haven't checked back on how it actually works e2e in the sample app.

> However, in that case, I wonder why we'd call movePreviousSession() from EnvelopeCache.storeInternal() at all, since the MovePreviousSession runnable is triggered by Sentry.init(), meaning there should never be a case that requires starting the "previous session" lifecycle from EnvelopeCache.storeInternal().

I think the problematic case is when the `MovePreviousSession` runnable hasn't been triggered yet (due to its async nature), but the new session is being started already (from a different thread). In that case if we don't call `movePreviousSession()` from `EnvelopeCache.storeInternal()` we risk losing the previous session altogether, because we'll overwrite the current `session.json` file with the new one.

I'm wondering if we should actually revert the commit mentioned above to not always overwrite previous_session but bail out if it already exists. This would probably solve it?

romtsn (approved):
This looks incredibly well done, hats off 🎩 I'm stamping this, there's still a couple of more things to resolve (debugId and crashedLastRun), but those can be done in a different PR. And I guess we need the changelog entry, if we're decide to finish everything up in this PR.

sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/TombstoneParser.java
supervacuus: fixed in https://github.com/getsentry/sentry-java/pull/4933/commits/bcf87cfa237532e3b35a559eb012fd051486b8bd

sentry/src/main/java/io/sentry/cache/EnvelopeCache.java
supervacuus: A very cheap (and I think durable) fix would be to exit early if no `session.json` file is found. This immediately makes `movePreviousSession()` idempotent wrt `previous_session.json`, without sacrificing prioritizing recent session closure over older ones. I attempted this here: https://github.com/getsentry/sentry-java/pull/4933/commits/5e9872efbff65efb336b4b9e944614665b1cf26b.

romtsn: sounds like a neat solution, looks good 👍

sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/TombstoneParser.java
romtsn: I think we're doing something similar here for ANRs https://github.com/getsentry/sentry-java/blob/d15471fde010743bf7ac30aa2c5a7ed61ebea807/sentry-android-core/src/main/java/io/sentry/android/core/internal/threaddump/ThreadDumpParser.java#L134-L150, do you think we can reuse the same solution between both?

supervacuus: I could have had a look at the code base first. Of course, we can use an existing solution. I like the `BigInteger` approach.

romtsn: sorry 😅 I only realized it now that we have this elsewhere

sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/TombstoneParser.java
```diff
@@ -0,0 +1,219 @@
...
+      // exclude anonymous and non-executable maps
+      if (module.getBuildId().isEmpty()
+          || module.getMappingName().isEmpty()
+          || !module.getExecute()) {
+        continue;
+      }
+      final DebugImage image = new DebugImage();
+      image.setCodeId(module.getBuildId());
+      image.setCodeFile(module.getMappingName());
+      image.setDebugId(OleGuidFormatter.convert(module.getBuildId()));
```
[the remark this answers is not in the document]
supervacuus: While highly unlikely, that is still good input. However, the code changes anyway because I now reuse the existing conversion in the code base.

sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/TombstoneParser.java
supervacuus: np, fixed in https://github.com/getsentry/sentry-java/pull/4933/commits/2f3558600e50640da58c8dd8cdf143532e8f2d06

sentry/src/main/java/io/sentry/cache/EnvelopeCache.java
supervacuus: The remaining topics here are resolved via https://github.com/getsentry/sentry-java/pull/4933/commits/5e9872efbff65efb336b4b9e944614665b1cf26b and https://github.com/getsentry/sentry-java/pull/4933/commits/b4973811890be52647b1f8f4b2c3b7384bfd69fb.

sentry/src/main/java/io/sentry/cache/EnvelopeCache.java
supervacuus: > We probably should put that logic into PreviousSessionFinalizer though, because that's where the current logic for native crashes lives?

addressed here: https://github.com/getsentry/sentry-java/pull/4933/commits/b4973811890be52647b1f8f4b2c3b7384bfd69fb (after fixing "previous session" life-cycle here: https://github.com/getsentry/sentry-java/pull/4933/commits/5e9872efbff65efb336b4b9e944614665b1cf26b)

supervacuus:
> there's still a couple of more things to resolve (debugId and crashedLastRun), but those can be done in a different PR

@romtsn, these should now both be addressed (`crashedLastRun` after https://github.com/getsentry/sentry-java/pull/4933/commits/5e9872efbff65efb336b4b9e944614665b1cf26b and https://github.com/getsentry/sentry-java/pull/4933/commits/b4973811890be52647b1f8f4b2c3b7384bfd69fb and `debugId` after https://github.com/getsentry/sentry-java/pull/4933/commits/2f3558600e50640da58c8dd8cdf143532e8f2d06).

Open topics, I am aware of:

- [x] changelog
- [x] ~docs~
- [x] 2 failing checks ("SDK size analysis" is an authentication issue; not sure what the failure cause is in the `TombstoneIntegrationTest`, can't repro locally)
- [x] ~For development convenience, I added a manifest entry to enable the tombstone integration. However, I am not sure whether this makes sense to release this if we keep that flag "internal" in the options. Let me know if I should add it anyway.~

This sounds like a manageable 2026 spillover.

romtsn:
> > there's still a couple of more things to resolve (debugId and crashedLastRun), but those can be done in a different PR
> 
> @romtsn, these should now both be addressed (`crashedLastRun` after [5e9872e](https://github.com/getsentry/sentry-java/commit/5e9872efbff65efb336b4b9e944614665b1cf26b) and [b497381](https://github.com/getsentry/sentry-java/commit/b4973811890be52647b1f8f4b2c3b7384bfd69fb) and `debugId` after [2f35586](https://github.com/getsentry/sentry-java/commit/2f3558600e50640da58c8dd8cdf143532e8f2d06)).
> 
> Open topics, I am aware of:
> 
> * [ ]  changelog
> * [ ]  docs
> * [ ]  2 failing checks ("SDK size analysis" is an authentication issue; not sure what the failure cause is in the `TombstoneIntegrationTest`, can't repro locally)
> * [ ]  For development convenience, I added a manifest entry to enable the tombstone integration. However, I am not sure whether this makes sense to release this if we keep that flag "internal" in the options. Let me know if I should add it anyway.
> 
> This sounds like a manageable 2026 spillover.

 We can skip the manifest option for the first iteration 👍 And re. SDK Size Analysis I will open a PR to not run it for forks soonish. The rest sounds good!

sentry/src/main/java/io/sentry/PreviousSessionFinalizer.java
```diff
@@ -86,16 +86,30 @@ public void run() {
...
-          Date timestamp = null;
           final File crashMarkerFile =
               new File(options.getCacheDirPath(), NATIVE_CRASH_MARKER_FILE);
-          if (crashMarkerFile.exists()) {
+
+          if (session.getStatus() == Session.State.Crashed) {
+            // this means the session was already updated from tombstone data and there is no need
+            // to consider the crash-marker file from the Native SDK. However, we can use this to
+            // set crashedLastRun.
+            SentryCrashLastRunState.getInstance().setCrashedLastRun(true);
```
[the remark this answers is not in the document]
supervacuus: good call, fixed in https://github.com/getsentry/sentry-java/pull/4933/commits/1f490e485a3a38ad358da165aa8ca398ed65d171

sentry/src/main/java/io/sentry/PreviousSessionFinalizer.java
```diff
@@ -86,16 +86,30 @@ public void run() {
...
+            // update the session state and timestamp.
             options
                 .getLogger()
                 .log(INFO, "Crash marker file exists, last Session is gonna be Crashed.");
 
-            timestamp = getTimestampFromCrashMarkerFile(crashMarkerFile);
+            @Nullable Date timestamp = getTimestampFromCrashMarkerFile(crashMarkerFile);
+            session.update(Session.State.Crashed, null, true);
+            session.end(timestamp);
+          }
```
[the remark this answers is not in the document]
supervacuus: Since the code in `main` only retrieves the timestamp from the native crash marker and sets it to `null` otherwise, I thought I could ignore that case. However, when `Session.end()` receives a `null` value, it sets the session's timestamp to the current `DateTime`.

The bigger problem is that there is a test that should cover this case, but actually doesn't correctly check it:

`if previous session exists, sends session update with current end time`

Fixed the test and updated the code in https://github.com/getsentry/sentry-java/pull/4933/commits/38b75dfc49bce4a96e2ee191294b42dd8f5416aa

PR #4958, opened 2025-12-09
fix: improve network body parsing and truncation handling

markushi:
## :scroll: Description
This PR refactors the `NetworkBody` class and improves how network request/response bodies are parsed and handled, particularly for large or truncated content.

Key changes:
- Simplified `NetworkBody` from an interface with multiple implementations to a single class with a `body` field and optional `warnings` list
- Added `NetworkBodyWarning` enum to communicate parsing issues (JSON_TRUNCATED, TEXT_TRUNCATED, INVALID_JSON, BODY_PARSE_ERROR)
- Improved truncation handling in `NetworkBodyParser` to properly detect and mark truncated content
- Switched from `JsonObjectReader` to vendor `JsonReader` for more robust JSON parsing
- Updated `SentryOkHttpInterceptor` to peek at the correct amount of bytes to detect truncation

Previously, the network body parsing had issues with:
1. Inconsistent handling of truncated content
2. No way to communicate parse warnings to consumers
3. Less robust JSON parsing

This change provides a cleaner API and better error/warning reporting.

- Existing tests should pass
- Manual testing with large response bodies

## :pencil: Checklist

## :crystal_ball: Next steps

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,4 +2,4 @@ certifi==2025.7.14
 charset-normalizer==3.4.2
 idna==3.10
 requests==2.32.4
-urllib3==2.5.0
+urllib3==2.6.0


sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
```diff
@@ -317,14 +317,10 @@ public open class SentryOkHttpInterceptor(
...
         val maxBodySize = SentryReplayOptions.MAX_NETWORK_BODY_SIZE
 
-        val contentLength = responseBody.contentLength()
-        if (contentLength > maxBodySize * 2) {
-          return NetworkBody.fromString("[Response body too large: $contentLength bytes]")
-        }
-
         // Peek at the body (doesn't consume it)
-        val peekBody = peekBody(maxBodySize.toLong())
+        val peekBody = peekBody(maxBodySize.toLong() + 1)
```
43jay, line 323: is this +1 (`maxBodySize +1`) needed here?

-> it's upping the MAX_NETWORK_BODY_SIZE by 1 (unnecessarily?)

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
```diff
@@ -317,14 +317,10 @@ public open class SentryOkHttpInterceptor(
         val contentTypeString = contentType?.toString()
         val maxBodySize = SentryReplayOptions.MAX_NETWORK_BODY_SIZE
 
-        val contentLength = responseBody.contentLength()
-        if (contentLength > maxBodySize * 2) {
```
43jay, line 321: fwiw my thinking was to properly handle the truncation inside NetworkBodyParser#fromBytes.

then this code checking `contentLength` was an "early abort" to avoid allocating a bunch of memory in `peekBody` only to immediately return after checking its length in `NetworkBodyParser#fromBytes`

43jay: but given contentLength can be -1 i think the right call is to remove this

sentry/src/main/java/io/sentry/util/network/NetworkBodyParser.java
```diff
@@ -42,31 +47,29 @@ private NetworkBodyParser() {}
...
           "[Binary data, " + bytes.length + " bytes, type: " + contentType + "]");
     }
 
-    // Check size limit and truncate if necessary
-    if (bytes.length > maxSizeBytes) {
-      logger.log(
-          SentryLevel.WARNING, "Content exceeds max size limit of " + maxSizeBytes + " bytes");
-      return createTruncatedNetworkBody(bytes, maxSizeBytes, charset);
-    }
+    final boolean isPartial = bytes.length >= maxSizeBytes;
```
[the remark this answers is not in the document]
43jay: this seems legit.

however looking at the LHS of this diff, i'm now a bit confused how `bytes.length > maxSizeBytes` could ever evaluate to `true` (to ever run `createTruncatedBody(..`)

peekBody [here](https://github.com/getsentry/sentry-java/blob/dba088c465439af6142917dc884a1919ff042722/sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt#L326) was using maxBodySize, so the largest `bytes[]` passed into NetworkBodyParser#fromBytes could only ever be `maxBodySize`🤔. 
^If so, I'd expect the fix to be to update the condition on ln 50 to ~`bytes.length > maxSizeBytes`~ `bytes.length >= maxSizeBytes` 

think ima need to build and do some debugging myself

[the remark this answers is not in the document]
43jay: Rest of the diff looks great - 
- adding support for `warnings` similar to how the JS SDK does it, and 
- using a better version of JSONReader with new and improved `readJSONSafely` 

^ The JS SDK does this, i.e it does attempt to read out the JSON up to the max size, but my initial android impl short-cut this by returning a string when max size is exceeded (didn't bother trying partially parse the JSON)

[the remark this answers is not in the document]
markushi: I've adapted the logic, we still need to +1 in one occurrence, in order to distinguish between the body being the same as max size or the body exceeding max size - I've left a comment to clarify this.

sentry/src/main/java/io/sentry/util/network/NetworkBodyParser.java
```diff
@@ -75,46 +78,45 @@ private NetworkBodyParser() {}
...
+    } catch (Exception e) {
+      if (logger != null) {
+        logger.log(SentryLevel.WARNING, "Failed to parse JSON: " + e.getMessage());
       }
     }
-
-    // Default to string representation
-    return NetworkBody.fromString(content);
+    return new NetworkBody(
+        null, Collections.singletonList(NetworkBody.NetworkBodyWarning.INVALID_JSON));
```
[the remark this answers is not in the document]
markushi: This is true, but for partial parsing support there's no easy way around this, so I'll keep as-is for now. Having a json `null` blob also sounds something which occurs rarely.

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
markushi: Yes, see my other comments. The `.parse()` method will trim according to max size.

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpInterceptor.kt
markushi: 👍 I've moved the truncation logic to `NetworkBodyParser#fromBytes.`

sentry/src/main/java/io/sentry/util/network/NetworkBodyParser.java
```diff
@@ -75,46 +78,47 @@ private NetworkBodyParser() {}
...
   }
```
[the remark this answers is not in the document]
markushi: As mentioned there's no easy way to check this right now, so I'd keep it as-is.

sentry-android-replay/src/main/java/io/sentry/android/replay/DefaultReplayBreadcrumbConverter.kt
```diff
@@ -241,7 +241,10 @@ public open class DefaultReplayBreadcrumbConverter() : ReplayBreadcrumbConverter
         networkData.request?.let { request ->
           val requestData = mutableMapOf<String, Any?>()
           request.size?.let { requestData["size"] = it }
-          request.body?.let { requestData["body"] = it.value }
+          request.body?.let {
+            requestData["body"] = it.body
+            requestData["warnings"] = it.warnings?.map { w -> w.value }
+          }
```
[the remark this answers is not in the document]
markushi: Good point!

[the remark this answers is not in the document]
romtsn: so @markushi do you want to convert this into `it.warnings?.let { }`? or there's no point in that

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- Improve network body parsing and truncation handling ([#4958](https://github.com/getsentry/sentry-java/pull/4958))
```
```suggestion
- Session Replay: Improve network body parsing and truncation handling ([#4958](https://github.com/getsentry/sentry-java/pull/4958))
```

sentry/src/main/java/io/sentry/util/network/NetworkBody.java
```diff
@@ -1,113 +1,61 @@
...
+  public @Nullable List<NetworkBodyWarning> getWarnings() {
+    return warnings;
   }
 
-  /** Implementation for string bodies */
-  final class StringBodyImpl implements NetworkBody {
-    private final @NotNull String value;
+  // Based on
+  // https://github.com/getsentry/sentry/blob/ccb61aa9b0f33e1333830093a5ce3bd5db88ef33/static/app/utils/replays/replay.tsx#L5-L12
+  public enum NetworkBodyWarning {
```
romtsn, line 40: so we can't exactly match all of the reasons on Android?

sentry/src/main/java/io/sentry/util/network/NetworkBodyParser.java
```diff
@@ -75,46 +78,47 @@ private NetworkBodyParser() {}
...
+      if (data != null) {
+        final @Nullable List<NetworkBody.NetworkBodyWarning> warnings;
+        if (isPartial) {
+          warnings = Collections.singletonList(NetworkBody.NetworkBodyWarning.JSON_TRUNCATED);
+        } else {
+          warnings = null;
         }
+        return new NetworkBody(data, warnings);
+      }
+    } catch (Exception e) {
```
romtsn, line 115: so we still wanna catch Exception and not Throwable?

sentry/src/main/java/io/sentry/util/network/NetworkBody.java
markushi: Some of them simply don't make sense on mobile, like `BODY_PARSE_TIMEOUT`. So I only added those which we actually use

sentry/src/main/java/io/sentry/util/network/NetworkBodyParser.java
markushi: I've read up on this and it seems like it's not safe it simply catch a StackOverflow as the JVM could be in an inconsistent state. I've added a `max_depth` const for parsing instead and combined with the truncation I think we have enough safe guards in place now.

romtsn: wondering if we this is already a problem elsewhere where we catch Throwables 😅

sentry/src/main/java/io/sentry/util/network/NetworkBodyParser.java
```diff
@@ -176,61 +182,88 @@ private static boolean isBinaryContentType(@NotNull final String contentType) {
...
+            try {
+              reader.beginObject();
+              while (reader.hasNext() && !result.errored) {
+                final String name = reader.nextName();
+                map.put(name, parse(reader, currentDepth + 1));
               }
+              reader.endObject();
+            } catch (Exception e) {
+              result.errored = true;
+              return map;
```
43jay, line 228: nit: unnecessary return

43jay (approved):
looks great!, and confirmed that 

1) works as expected for `SecondActivity` on my device (when string value is truncated/malformed [SR](https://sentry-sdks.sentry.io/explore/replays/6e266e647ee84534ba692ccfc40d3b64/?n_detail_row=0&n_detail_tab=response&playlistEnd=2025-12-11T19%3A43%3A43&playlistStart=2025-11-27T19%3A43%3A43&project=5428559&query=&referrer=replayList&t_main=network))
2) WAE when key name is truncated/malformed (malformed keys) [SR](https://sentry-sdks.sentry.io/explore/replays/5f68ba6a7c5548668f0f6ba57acd8273/?n_detail_row=0&n_detail_tab=response&playlistEnd=2025-12-11T19%3A43%3A43&playlistStart=2025-11-27T19%3A43%3A43&project=5428559&query=&referrer=replayList&t_main=network)

PR #5037, opened 2026-01-19
feat: merge tombstone and native sdk events

supervacuus:
## :scroll: Description
This implements the second aspect of the tombstone integration: merging tombstone events with crash events coming from sentry-native. It does this in the following way:

* No changes to the `OutboxSender`, but instead let `TombstoneIntegration` own the outbox folder before the `OutboxSender` operates on it (i.e., we rely on integration instantiation order and the fact that all async integration work is done on a single-threaded executor service). 
* `NativeEventCollector` streams and skips through all envelopes (without full hydration) to collect only timestamps and file references for matching with tombstone timestamps. Once it has match it loads the full envelopes and events into memory to merge and add native attachments. This reduces memory and CPU load since a large part of the outbox will be irrelevant to Tombstone matching.
* Correlation/matching between native events and tombstones only happens via timestamps.
* `TombstonePolicy` searches, matches, and merges events
* merging is essentially: take the native SDK event and replace `debugMeta`, `exception`, and `threads`
* This PR also fixes the module loader: in the initial PR, modules were created from executable VMAs, but this clashes with the offset assumption of the backend. `TombstoneParser` must coalesce the memory mappings for an inode into a single module, whose start address aligns with the file-offset. This reflects the implementation in the Native SDK. See also: https://github.com/getsentry/sentry-java/issues/3295#issuecomment-3833603140

This is the last essential step before releasing the tombstone integration to a wider audience. It maps directly to 
[Implement merge mechanism with the Native SDK envelope](https://linear.app/getsentry/project/tombstone-support-android-0024cb6e3ffd/overview#milestone-e4a6aaac-3d0c-4e7f-a537-2ea63a3b47a7)

* only manually up to now
* Once the direction is agreed on, will add unit+integration tests

## :pencil: Checklist

## :crystal_ball: Next steps

* ~decide on this approach or extending `OutboxSender` and surrounding infra more generically (we could introduce a callback for the core of the `OutboxSender` decision and set up integrations that use it accordingly)~
* ~decide merging strategy~
* ~decide on enrichment (maybe we should use a more default `EventProcessor` for merge events?)~
* ~Related to the above:~
  * ~Should correlation only be attempted for the latest tombstone?~
  * ~How should historical tombstones be matched? (Since we have all the actual data from the moment of the event in the native envelope, they will likely be more accurate for that moment in time vs backfilling, but we don't track the same amount of context information)~
* ~Add correlation id to the Native SDK~

### DIFF
--- a/sentry-android-core/src/test/resources/envelopes/native-event.txt
+++ b/sentry-android-core/src/test/resources/envelopes/native-event.txt
@@ -1,3 +1,3 @@
 {"event_id":"9ec79c33ec9942ab8353589fcb2e04dc"}
-{"type":"event","length":123,"content_type":"application/json"}
+{"type":"event","length":122,"content_type":"application/json"}
 {"event_id":"9ec79c33ec9942ab8353589fcb2e04dc","timestamp":"2023-07-15T10:30:00.000Z","platform":"native","level":"fatal"}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,6 @@
 
 - Add `installGroupsOverride` parameter and `installGroups` property to Build Distribution SDK ([#5062](https://github.com/getsentry/sentry-java/pull/5062))
 - Update Android targetSdk to API 36 (Android 16) ([#5016](https://github.com/getsentry/sentry-java/pull/5016))
-- Merge Tombstone and Native SDK events into single crash event. ([#5037](https://github.com/getsentry/sentry-java/pull/5037))
 - Add AndroidManifest support for Spotlight configuration via `io.sentry.spotlight.enable` and `io.sentry.spotlight.url` ([#5064](https://github.com/getsentry/sentry-java/pull/5064))
 - Collect database transaction spans (`BEGIN`, `COMMIT`, `ROLLBACK`) ([#5072](https://github.com/getsentry/sentry-java/pull/5072))
   - To enable creation of these spans, set `options.enableDatabaseTransactionTracing` to `true`
@@ -15,6 +14,20 @@
     ```yaml
     sentry:
       enable-database-transaction-tracing: true
+    ```
+- Add support for collecting native crashes using Tombstones ([#4933](https://github.com/getsentry/sentry-java/pull/4933), [#5037](https://github.com/getsentry/sentry-java/pull/5037))
+  - Added Tombstone integration that detects native crashes using `ApplicationExitInfo.REASON_CRASH_NATIVE` on Android 12+
+  - Crashes enriched with Tombstones contain more crash details and detailed thread info
+  - Tombstone and NDK integrations are now automatically merged into a single crash event, eliminating duplicate reports
+  - To enable it, add the integration in your Sentry initialization:
+    ```kotlin
+    SentryAndroid.init(context, options -> {
+        options.isTombstoneEnabled = true
+    })
+    ```
+    or in the `AndroidManifest.xml` using:
+    ```xml
+    <meta-data android:name="io.sentry.tombstone.enable" android:value="true" />
     ```
 
 ### Fixes
--- a/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
@@ -321,7 +321,6 @@ public final class SentryAndroidOptions extends SentryOptions {
    *
    * @param enableTombstone true for enabled and false for disabled
    */
-  @ApiStatus.Internal
   public void setTombstoneEnabled(boolean enableTombstone) {
     this.enableTombstone = enableTombstone;
   }
@@ -332,7 +331,6 @@ public final class SentryAndroidOptions extends SentryOptions {
    *
    * @return true if enabled or false otherwise
    */
-  @ApiStatus.Internal
   public boolean isTombstoneEnabled() {
     return enableTombstone;
   }
@@ -615,12 +613,10 @@ public final class SentryAndroidOptions extends SentryOptions {
     this.reportHistoricalAnrs = reportHistoricalAnrs;
   }
 
-  @ApiStatus.Internal
   public boolean isReportHistoricalTombstones() {
     return reportHistoricalTombstones;
   }
 
-  @ApiStatus.Internal
   public void setReportHistoricalTombstones(final boolean reportHistoricalTombstones) {
     this.reportHistoricalTombstones = reportHistoricalTombstones;
   }
--- a/sentry-android-core/src/test/resources/envelopes/native-with-attachment.txt
+++ b/sentry-android-core/src/test/resources/envelopes/native-with-attachment.txt
@@ -1,5 +1,5 @@
 {"event_id":"9ec79c33ec9942ab8353589fcb2e04dc"}
 {"type":"attachment","length":20,"filename":"log.txt","content_type":"text/plain"}
 some attachment data
-{"type":"event","length":123,"content_type":"application/json"}
+{"type":"event","length":122,"content_type":"application/json"}
 {"event_id":"9ec79c33ec9942ab8353589fcb2e04dc","timestamp":"2023-07-15T11:45:30.500Z","platform":"native","level":"fatal"}


supervacuus:
@markushi and @romtsn, can you have a first look so we can discuss trade-offs? Thx!

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -497,4 +506,33 @@ private static void readDefaultOptionValues(
...
+   * Sets a native crash correlation ID that can be used to associate native crash events (from
+   * sentry-native) with tombstone events (from ApplicationExitInfo). The ID is stored via
+   * ActivityManager.setProcessStateSummary() and passed to the native SDK.
+   *
+   * @param context the Application context
+   * @param options the SentryAndroidOptions
+   */
+  private static void setNativeCrashCorrelationId(
+      final @NotNull Context context, final @NotNull SentryAndroidOptions options) {
+    final String correlationId = UUID.randomUUID().toString();
```
markushi: As anyone could call this API to store information, it might be a good idea to ensure it's our data we're reading back later. I guess a simple prefix could work e.g. `sentry-app-run-<uuid>`

sentry-android-core/src/main/java/io/sentry/android/core/TombstoneIntegration.java
```diff
@@ -164,13 +168,69 @@ public boolean shouldReportHistorical() {
...
+
       final TombstoneHint tombstoneHint =
           new TombstoneHint(
               options.getFlushTimeoutMillis(), options.getLogger(), tombstoneTimestamp, enrich);
       final Hint hint = HintUtils.createWithTypeCheckHint(tombstoneHint);
 
       return new ApplicationExitInfoHistoryDispatcher.Report(event, hint, tombstoneHint);
     }
+
+    private SentryEvent mergeNaiveCrashes(
```
markushi:
```suggestion
    private SentryEvent mergeNativeCrashes(
```

sentry-android-core/src/main/java/io/sentry/android/core/NativeEventCollector.java
```diff
@@ -0,0 +1,276 @@
...
+    if (files == null || files.length == 0) {
+      options.getLogger().log(SentryLevel.DEBUG, "No envelope files found in outbox.");
+      return;
+    }
+
+    options
+        .getLogger()
+        .log(SentryLevel.DEBUG, "Scanning %d files in outbox for native events.", files.length);
+
+    for (final File file : files) {
```
markushi, line 140: I'm a bit worried of ending up with too many envelopes in memory here and potentially running into OOMs.

sentry-android-core/src/main/java/io/sentry/android/core/TombstoneIntegration.java
```diff
@@ -164,13 +168,69 @@ public boolean shouldReportHistorical() {
...
+            .log(
+                SentryLevel.DEBUG,
+                "Found matching native event for tombstone, removing from outbox: %s",
+                matchingNativeEvent.getFile().getName());
+
+        // Delete from outbox so OutboxSender doesn't send it
+        boolean deletionSuccess = nativeEventCollector.deleteNativeEventFile(matchingNativeEvent);
+
+        if (deletionSuccess) {
+          event = mergeNaiveCrashes(matchingNativeEvent.getEvent(), event);
```
markushi: Is there any specific information a native crash event could hold which is not available through the tombstone? I guess this is mainly about user-added context like e.g. breadcrumbs, right?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
supervacuus: True, and this is only meant as a proposal that survived from 2022:

* `processStateSummary` is global and generally owned by the app
* If we write into this, we should first check if there is already something in there
* Even if nothing was in there, any later code can overwrite it without us knowing anything about it
* The maximum length is 128 bytes, and writing into it is free-form, so there is no way for binary stability except if we communicate that Sentry owns this if you use the SDK (not sure if that is a good idea).
* If we prefix it with something generic like `sentry-app-run-`, it shows that this might be used beyond native crash correlation (it will be accessible from any `ApplicationExitInfo`).

I implemented this because we considered it a stable correlation mechanism back then, but I think timestamp correlation is stable enough and doesn't depend on app-owned storage.

It's likely a good idea to drop this altogether, and if we introduce it, maybe together with other `ApplicationExitInfo` handlers?

sentry-android-core/src/main/java/io/sentry/android/core/TombstoneIntegration.java
supervacuus: Thanks!

sentry-android-core/src/main/java/io/sentry/android/core/TombstoneIntegration.java
supervacuus: Yes, we're only interested in context data from the Native SDK. I extended the comments with merge reasoning.

sentry-android-core/src/main/java/io/sentry/android/core/NativeEventCollector.java
supervacuus: Can you tell me which code in particular is likely to cause references to spill? I'm not entirely sure. My assumption was that there would typically be very few native events (on average, 1 after each native crash), since they are usually sent on the next restart (and we only collect native events and their envelopes). So I would guess that, while iterating over and deserializing all envelopes, each iteration should leave the discarded ones ready for GC, right?

My bigger worry was the attachments that we pass around with the envelopes (yes, they are currently not attached to the tombstone events). I am fine with not passing around the envelope for the event, but we must ensure it is loaded at some point. Was this your concern? We could store only the envelope path in `NativeEventData` and only load it once we have a matching tombstone. Or did I completely misunderstand?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
markushi: 👍 fully agree, I think the timestamp correlation is already good enough!

sentry-android-core/src/main/java/io/sentry/android/core/NativeEventCollector.java
markushi: As far as I can see we `extractNativeEventFromFile()` parses any envelope, so there could be transactions, or multiple crashes, e.g. if the device was offline. Or e.g. the app continuously crashes (in native code) during startup, a fix get released and now all unset envelopes pile up.

So I would just be a bit more defensive here, e.g. drop transactions, and have a sane maximum for `nativeEvents.size()`

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
supervacuus: fixed in https://github.com/getsentry/sentry-java/pull/5037/commits/b832e7c89f7b1e310475cfc7cc5682b2b6cf015d

sentry-android-core/src/main/java/io/sentry/android/core/NativeEventCollector.java
supervacuus: > So I would just be a bit more defensive here, e.g. drop transactions, and have a sane maximum for `nativeEvents.size()`

I would love to be defensive; I would like to discard envelopes based solely on the filename (or other metadata), rather than having to read them. But even with reading them i am not sure about the concrete consequences of your proposal:

* In `extractNativeEventFromFile()`, when iterating over the envelope items, we drop everything that isn't an event; isn't that what you mean by "drop transactions"?
  https://github.com/getsentry/sentry-java/blob/b832e7c89f7b1e310475cfc7cc5682b2b6cf015d/sentry-android-core/src/main/java/io/sentry/android/core/NativeEventCollector.java#L208-L211
* RE: sane maximum: 
  * If we limit the native events that we collect for matching, should the other events be discarded (since we assume sending a pure tombstone in their place) or allow a duplicate?
  * How do we know which events are the most significant/recent? Should we rely on disk timestamps?
  * We would, in any case, read more envelopes at least until decoding and reaching the maximum collection size, than we actually collect. So if heap pressure during parsing/decoding until those pieces are GCed is the worry, then the collection threshold is still happening too late.
  * If otoh, passing around the event + envelope is the worry, then we could reduce the `NativeEventData` to just timestamp and File and re-read the data only once there is an actual match.

The true cost is, of course, that we have to read, decode all envelopes, and deserialize every event, to verify that it is a native event and read its timestamp.

supervacuus: > The true cost is, of course, that we have to read, decode all envelopes, and deserialize every event, to verify that it is a native event and read its timestamp.

To be clear: We can make this part of the `NativeEventCollector` stream the envelopes (we only need item headers to skip anything but events) and the event (using a `JsonObjectReader` on an `InputStream` that is bound by the payload size), avoiding loading the entire envelope and hydrating the full object graph of each event until we actually have a match with a tombstone (at which point we would just use the current implementation `extractNativeEventFromFile()`). 

I can't work on this today, but tomorrow you can have a basic implementation. That would eliminate the worry about loading a bunch of items into memory that we will never use.

supervacuus: https://github.com/getsentry/sentry-java/pull/5065

sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/TombstoneParser.java
```diff
@@ -91,14 +103,38 @@ private List<SentryThread> createThreads(
...
+      if (frame.getFileName().endsWith("libart.so")) {
+        // We ignore all ART frames for time being because they aren't actionable for app developers
+        continue;
+      }
       final SentryStackFrame stackFrame = new SentryStackFrame();
       stackFrame.setPackage(frame.getFileName());
       stackFrame.setFunction(frame.getFunctionName());
       stackFrame.setInstructionAddr(formatHex(frame.getPc()));
+
+      // TODO: is this the right order?
```
romtsn: I think it is, at least we're doing the same here: https://github.com/getsentry/sentry-java/blob/0eaac1e5224107ed013e25afa352fcc33f18f737/sentry/src/main/java/io/sentry/SentryStackTraceFactory.java#L81-L101

with one caveat though - we bail out early if inAppIncludes contains it, whereas here we proceed to checking inAppExcludes. So if there's an entry that happened to be in both lists, inAppExcludes wins in this codepath. Not sure if we want to align with SentryStackTraceFactory, but I'm fine with the current logic as well.

supervacuus: I think aligning it would make sense. Ideally, we would extract the innards into a static helper and implement both usages of the filter configuration with the same implementation.

sentry-android-core/src/main/java/io/sentry/android/core/NativeEventCollector.java
```diff
@@ -0,0 +1,528 @@
...
+    final @Nullable String outboxPath = options.getOutboxPath();
+    if (outboxPath == null) {
+      options
+          .getLogger()
+          .log(SentryLevel.DEBUG, "Outbox path is null, skipping native event collection.");
+      return;
+    }
+
+    final File outboxDir = new File(outboxPath);
+    if (!outboxDir.isDirectory()) {
```
romtsn: I know we do this a lot, but was wondering if we can reduce the number of I/O calls in this method a bit -- e.g. don't check for `isDirectory` but just go to listFiles() right away  (which will return null if it's not a dir). We could also just call `listFiles()` without the filename filter, and then iterate through it and check for relevant filename in a single for-loop

sentry-android-core/src/main/java/io/sentry/android/core/TombstoneIntegration.java
```diff
@@ -164,13 +180,65 @@ public boolean shouldReportHistorical() {
...
+      // * even for the crashing  we expect a much better stack-trace (+ symbolication)
+      // * tombstone adds additional exception meta-data to signal handler content
+      // * we add debug-meta for consistency since the Native SDK caches memory maps early
+      @Nullable List<SentryException> tombstoneExceptions = tombstoneEvent.getExceptions();
+      @Nullable DebugMeta tombstoneDebugMeta = tombstoneEvent.getDebugMeta();
+      @Nullable List<SentryThread> tombstoneThreads = tombstoneEvent.getThreads();
+      if (tombstoneExceptions != null
+          && !tombstoneExceptions.isEmpty()
+          && tombstoneDebugMeta != null
+          && tombstoneThreads != null) {
```
romtsn, line 287: just wondering if this is a necessary condition for the rest to be applied? Can there be a no-threads case for tombstones?

sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/TombstoneParser.java
romtsn: yeah, let's do that!

supervacuus: fixed in https://github.com/getsentry/sentry-java/pull/5037/commits/72d56443fedc5942629d9ec19bac8c0c4966b21f

sentry-android-core/src/main/java/io/sentry/android/core/NativeEventCollector.java
supervacuus: fixed in https://github.com/getsentry/sentry-java/pull/5037/commits/2fd6b287d2e835570b313d71d5475d28da3c43d1

sentry-android-core/src/main/java/io/sentry/android/core/TombstoneIntegration.java
supervacuus: From my POV, none of those can be missing in a tombstone, but if one is, we shouldn't merge any of them. Exceptions map to a thread (which native doesn't because threads are empty), both exceptions and threads map to debugMeta. 

So, for consistency, I wanted to keep them separate, because the least understandable outcome would be if those fields ended up inconsistent in the event, and you'd have to reason where each came from. 

But I agree that this could also be overly defensive. But either we check them like this, or we don't check them at all. Does this make sense?

romtsn: yeah, it makes sense! we can keep it

supervacuus:
> Looks great, thanks! I'm wondering if we can reuse your approach with BoundedInputStream in other parts of the SDK where we're reading envelopes.

It likely could be useful. Even in the current stable event parsing in the `sentry` module, there are multiple hints at "can we make this streaming?". I am not sure if my approach is _the_ answer to these questions, but it is _an_ answer.

The question is also whether to expose this fully and let client code interact directly with `BoundedInputStream`, or to provide a trivial evented streaming parser interface with callbacks that allow one to filter and collect from the envelope and the envelope items. The core of all my collection functions could be exposed for client code to supply lambdas, and one could provide the remaining "file iteration" machinery as an abstraction.

sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/TombstoneParser.java
```diff
@@ -91,14 +104,28 @@ private List<SentryThread> createThreads(
...
+
+      @Nullable
+      Boolean inApp =
+          SentryStackTraceFactory.isInApp(frame.getFunctionName(), inAppIncludes, inAppExcludes);
+
+      final boolean isInNativeLibraryDir =
+          nativeLibraryDir != null && frame.getFileName().startsWith(nativeLibraryDir);
+      inApp = (inApp != null && inApp) || isInNativeLibraryDir;
+
+      stackFrame.setInApp(inApp);
```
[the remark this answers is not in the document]
supervacuus: those do not overlap

sentry-android-core/src/main/java/io/sentry/android/core/TombstoneIntegration.java
```diff
@@ -169,8 +190,112 @@ public boolean shouldReportHistorical() {
...
+      // Delete from outbox so OutboxSender doesn't send it
+      boolean deletionSuccess = nativeEventCollector.deleteNativeEventFile(matchingNativeEvent);
+
+      if (deletionSuccess) {
+        final SentryEvent nativeEvent = matchingNativeEvent.getEvent();
+        mergeNativeCrashes(nativeEvent, tombstoneEvent);
+        addNativeAttachmentsToTombstoneHint(matchingNativeEvent, hint);
+        return nativeEvent;
+      }
+      return null;
```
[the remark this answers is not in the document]
supervacuus: > This causes both a tombstone event (unmerged) and the original native event (via OutboxSender, since the file remains on disk) to be sent, resulting in duplicate crash reports.

Yes, this is intentional: a failed delete would still trigger a native event. In such a situation, I would prefer to have the full information, with the tombstone on one side and the native SDK event on the other, and no merging between them (rather than a merged tombstone and a native event).

This is primarily a usability topic in the case of a rare event. If it later turns out that people prefer the duplication with a merged tombstone event, we can always change the behavior. This is far from a "high severity" detection.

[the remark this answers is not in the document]
supervacuus: Duplicate of: https://github.com/getsentry/sentry-java/pull/5037#discussion_r2760672378

sentry-android-core/src/main/java/io/sentry/android/core/TombstoneIntegration.java
```diff
@@ -169,8 +190,112 @@ public boolean shouldReportHistorical() {
...
+      // * even for the crashing  we expect a much better stack-trace (+ symbolication)
+      // * tombstone adds additional exception meta-data to signal handler content
+      // * we add debug-meta for consistency since the Native SDK caches memory maps early
+      @Nullable List<SentryException> tombstoneExceptions = tombstoneEvent.getExceptions();
+      @Nullable DebugMeta tombstoneDebugMeta = tombstoneEvent.getDebugMeta();
+      @Nullable List<SentryThread> tombstoneThreads = tombstoneEvent.getThreads();
+      if (tombstoneExceptions != null
+          && !tombstoneExceptions.isEmpty()
+          && tombstoneDebugMeta != null
+          && tombstoneThreads != null) {
```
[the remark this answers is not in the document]
supervacuus: > This loses the native SDK's thread data, which includes at least the crashing thread.

No, there is no native SDK thread data; the only stack trace is in the exception payload (which does not yet reference any thread). Threads being empty from a tombstone is highly unlikely because of the way we parse tombstones. An empty exceptions list means there are no exceptions and thus no crash metadata for the backend to work on, and as such is much worse.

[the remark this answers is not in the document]
supervacuus: ugh, the noise 😱 ... see: https://github.com/getsentry/sentry-java/pull/5037#discussion_r2760690476

sentry-android-core/src/test/java/io/sentry/android/core/internal/tombstone/TombstoneParserTest.kt
```diff
@@ -93,13 +103,22 @@ class TombstoneParserTest {
         assertNotNull(frame.function)
         assertNotNull(frame.`package`)
         assertNotNull(frame.instructionAddr)
+
+        if (thread.id == crashedThreadId) {
+          if (frame.isInApp!!) {
+            assert(
+              frame.function!!.startsWith(inAppIncludes[0]) ||
+                frame.filename!!.startsWith(nativeLibraryDir)
```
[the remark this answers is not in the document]
supervacuus: Nice catch! The Sentry API means source filename, whereas the tombstone API means library/module/package filename. But the problem was deeper: because using filename here should actually fail the assertion, and the only reason it doesn't is that all in-app frames start with Java/Kotlin `inAppIncludes`, which would be wrong. And it turned out: the supplied `nativeLibraryDir` was not in sync with the one used in the tombstone fixture.

I fixed the `nativeLibraryDir` in the affected tests and ensured they would fail if the directory was incorrect: https://github.com/getsentry/sentry-java/pull/5037/commits/0ec2068f4fb2f67d60bd190217f780c487007d8e

sentry-android-core/src/main/java/io/sentry/android/core/NativeEventCollector.java
```diff
@@ -0,0 +1,532 @@
...
+      return Math.min(inner.available(), (int) remaining);
+    }
+
+    @Override
+    public void close() throws IOException {
+      // Skip any remaining bytes to advance the underlying stream position,
+      // but don't close the underlying stream, because we might have other
+      // envelope items to read.
+      skipBytes(inner, remaining);
+    }
```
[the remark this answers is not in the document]
supervacuus: This is a very nice catch that tests do not uncover because of the internal buffers of the underlying `InputStream`. Added a test that actually triggers this with an envelope-item > 8KiB and a reset in the `close()`: https://github.com/getsentry/sentry-java/pull/5037/commits/017ac5faabb927dc022c29ed6c633b4b50dcea52

sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/TombstoneParser.java
```diff
@@ -91,14 +104,36 @@ private List<SentryThread> createThreads(
...
+      // inAppIncludes/inAppExcludes filter by Java/Kotlin package names, which don't overlap
+      // with native C/C++ function names (e.g., "crash", "__libc_init"). For native frames,
+      // isInApp() returns null, making nativeLibraryDir the effective in-app check.
+      @Nullable
+      Boolean inApp =
+          SentryStackTraceFactory.isInApp(frame.getFunctionName(), inAppIncludes, inAppExcludes);
+
+      final boolean isInNativeLibraryDir =
+          nativeLibraryDir != null && frame.getFileName().startsWith(nativeLibraryDir);
+      inApp = (inApp != null && inApp) || isInNativeLibraryDir;
```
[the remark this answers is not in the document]
supervacuus: We reuse the core functionality to align with the `inApp` include/exclude list application. However, we can override the behavior for empty classnames because the core behavior is actually for class names, whereas our interest is only in function names. In that case, Java function names should never be empty and thus aren't affected by a local override. An empty function name from a tombstone means _not `InApp`_.

sentry-android-core/src/main/java/io/sentry/android/core/TombstoneIntegration.java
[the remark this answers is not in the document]
supervacuus: jesus christ the noise 😱

sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/TombstoneParser.java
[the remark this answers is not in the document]
supervacuus: fixed here: https://github.com/getsentry/sentry-java/pull/5037/commits/67aee7b92555c916aa2693e3e573c774344ba62c

This is truly an edge case, as we do not assume that Java frames are empty for tombstones. But yeah, let's "fix" it.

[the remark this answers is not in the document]
supervacuus: no, sorry... the native frame without a name can still make sense to marked as InApp when the package is from the NativeLibraryDir, because the backend could symbolicate it successfully.

sentry-android-core/src/main/java/io/sentry/android/core/NativeEventCollector.java
```diff
@@ -0,0 +1,536 @@
...
+
+    for (final NativeEnvelopeMetadata metadata : nativeEnvelopes) {
+      final long timeDiff = Math.abs(tombstoneTimestampMs - metadata.getTimestampMs());
+      if (timeDiff <= TIMESTAMP_TOLERANCE_MS) {
+        options
+            .getLogger()
+            .log(SentryLevel.DEBUG, "Matched native event by timestamp (diff: %d ms)", timeDiff);
+        nativeEnvelopes.remove(metadata);
+        // Only load full event data when we have a match
+        return loadFullNativeEventData(metadata.getFile());
```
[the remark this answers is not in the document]
supervacuus: yup answered and linked 5 times, before

sentry-android-core/src/main/java/io/sentry/android/core/TombstoneIntegration.java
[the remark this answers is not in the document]
supervacuus: 😠 ditto

sentry-android-core/src/test/resources/envelopes/native-event.txt
```diff
@@ -0,0 +1,3 @@
+{"event_id":"9ec79c33ec9942ab8353589fcb2e04dc"}
+{"type":"event","length":123,"content_type":"application/json"}
+{"event_id":"9ec79c33ec9942ab8353589fcb2e04dc","timestamp":"2023-07-15T10:30:00.000Z","platform":"native","level":"fatal"}
```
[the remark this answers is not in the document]
supervacuus: fixed in https://github.com/getsentry/sentry-java/pull/5037/commits/3e3fc7318069aa9bbb19b1fc6b1442e2feb2fd9d

sentry-android-core/src/main/java/io/sentry/android/core/NativeEventCollector.java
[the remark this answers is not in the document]
supervacuus: > The caller receives null, so the tombstone event is sent without merge, while the native event file remains in the outbox for OutboxSender to send — resulting in a duplicate crash report.

That is exactly the intended fallback behavior: if we cannot fully deserialize the event, there is nothing to merge, so we send only the tombstone. It is possible that we were able to decode the timestamp, but the remaining event is somehow corrupted. So, this is a valid failure mode, and the current behavior is exactly what I would expect.

If the `OutboxSender` successfully transmits the event anyway, great. But more likely it will also fail (since it uses exactly the method of deserialization). So the worry about duplicate events is also overblown: it either transmits the event and we send both events (in the curiously rare situation where the same deserialization fails once but works in the other case), or it fails (most likely).

PR #5052, opened 2026-01-22
fix: establish native exception mechanisms

supervacuus:
## :scroll: Description
As discussed yesterday during the sync, let's establish native exception mechanisms right away.

allows us to cleanly differentiate which handler an event originated from. This is particularly helpful since we now also merge events from two different mechanisms: https://github.com/getsentry/sentry-java/pull/5037

adapted the affected unit tests.

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/NativeExceptionMechanism.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/NativeExceptionMechanism.java
@@ -4,9 +4,9 @@ import androidx.annotation.NonNull;
 
 /** Mechanism types for native crashes. */
 public enum NativeExceptionMechanism {
-  TOMBSTONE("tombstone"),
+  TOMBSTONE("Tombstone"),
   SIGNAL_HANDLER("signalhandler"),
-  TOMBSTONE_MERGED("tombstone_merged");
+  TOMBSTONE_MERGED("TombstoneMerged");
 
   private final @NonNull String value;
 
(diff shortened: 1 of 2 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/internal/tombstone/NativeExceptionMechanism.java
```diff
@@ -0,0 +1,21 @@
+package io.sentry.android.core.internal.tombstone;
+
+import androidx.annotation.NonNull;
+
+/** Mechanism types for native crashes. */
+public enum NativeExceptionMechanism {
+  TOMBSTONE("tombstone"),
+  SIGNAL_HANDLER("signalhandler"),
+  TOMBSTONE_MERGED("tombstone_merged");
```
romtsn:
```suggestion
  TOMBSTONE("Tombstone"),
  SIGNAL_HANDLER("signalhandler"),
  TOMBSTONE_MERGED("TombstoneMerged");
```
I think we're really inconsistent with naming across SDKs, but we tend to use camelCase in the Java SDK, so I'm thinking if we could change that? signalhandler probably has to stay as-is to not break existing filters, but the new ones we may change to follow camelCase

supervacuus: fixed in https://github.com/getsentry/sentry-java/pull/5052/commits/1ffaa057ec764be37f5dae60890df2e03a248377

PR #5062, opened 2026-01-27
feat(distribution): Add install_groups support

runningcode:
## Summary

- Add `installGroupsOverride` parameter to `UpdateCheckParams` for specifying install groups when checking for updates
- Add `installGroups` property to `UpdateInfo` to receive install groups from the API response
- Parse `install_groups` from the distribution API response

This matches the iOS implementation: https://github.com/getsentry/sentry-cocoa/pull/7278

### DIFF
--- a/sentry/src/main/java/io/sentry/UpdateInfo.java
+++ b/sentry/src/main/java/io/sentry/UpdateInfo.java
@@ -1,5 +1,6 @@
 package io.sentry;
 
+import java.util.List;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
@@ -13,6 +14,7 @@ public final class UpdateInfo {
   private final @NotNull String downloadUrl;
   private final @NotNull String appName;
   private final @Nullable String createdDate;
+  private final @Nullable List<String> installGroups;
 
   public UpdateInfo(
       final @NotNull String id,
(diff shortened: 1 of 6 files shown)


sentry/src/main/java/io/sentry/UpdateInfo.java
```diff
@@ -21,12 +23,24 @@ public UpdateInfo(
       final @NotNull String downloadUrl,
       final @NotNull String appName,
       final @Nullable String createdDate) {
+    this(id, buildVersion, buildNumber, downloadUrl, appName, createdDate, null);
```
runningcode: This existing constructor calls the new one so we don't break the API.

I'm also leaning towards just breaking the API since it is extremely unlikely that this class is being used directly.

sentry/src/main/java/io/sentry/UpdateInfo.java
```diff
@@ -13,20 +14,23 @@ public final class UpdateInfo {
...
+  private final @Nullable List<String> installGroups;
 
   public UpdateInfo(
       final @NotNull String id,
       final @NotNull String buildVersion,
       final int buildNumber,
       final @NotNull String downloadUrl,
       final @NotNull String appName,
-      final @Nullable String createdDate) {
+      final @Nullable String createdDate,
```
runningcode, line 25: This is a breaking change to the constructor. I don't think the constructor is used directly by consumers so it is fine.

PR #5065, opened 2026-01-29
proposal for streaming NativeEventCollector

supervacuus:
Proposal for an adaptation to the `NativeEventCollector` introduced in https://github.com/getsentry/sentry-java/pull/5037:

* has 2 phases: collection of native event timestamps for matching, and only full decode once a matching tombstone is found
* The collection only streams envelopes and events, skips (based on payload types and sizes), and exits early as much as possible (once we have the metadata we need); nothing besides a read buffer is allocated.
* Once a matching tombstone is found, we use the previous `extractNativeEventFromFile()` to load the full envelope and event.

I used the envelopes from the core test resources, copied them to the `sentry-android-core` test resource folder, and added a couple of native-specific envelopes for the tests.

This limits memory usage as much as possible since there are at best 10 `ApplicationExitInfo` with `EXIT_REASON_NATIVE` (likely fewer), which act as an upper bound to the number of envelopes/events we fully decode. 

It should also be much fairer on the CPU, since much less decoding/deserialization is required. I/O will be the same during collection; however, with this approach, we must read the matching envelopes from disk again during full decode.

#skip-changelog


sentry-android-core/src/main/java/io/sentry/android/core/NativeEventCollector.java
```diff
@@ -244,4 +463,66 @@ private boolean isRelevantFileName(final @Nullable String fileName) {
...
         && !fileName.startsWith(STARTUP_CRASH_MARKER_FILE);
   }
+
+  /**
+   * An InputStream wrapper that tracks reads within a bounded section of the stream. This allows
+   * callers to read/parse only what they need (e.g., extract a few JSON fields), then skip the
+   * remainder of the section on close to position the stream at the next envelope item. Does not
+   * close the underlying stream.
+   */
+  private static final class BoundedInputStream extends InputStream {
```
markushi, line 473: Nice, that's elegant!

PR #5077, opened 2026-02-05
feat(screenshot): Add screenshot masking using view hierarchy

### LINKED ISSUE #3286: Redact Screenshots via View Hierarchy
see https://github.com/getsentry/team-mobile/issues/178

romtsn:
## :scroll: Description

Adds masking support to error screenshots by reusing the Session Replay masking logic. This allows sensitive content (text, images) to be masked before attaching screenshots to error events.

- Add SentryMaskingOptions base class for shared masking configuration
- Add SentryScreenshotOptions for screenshot-specific masking settings
- Create MaskRenderer utility for shared mask rendering (used by both replay and screenshots)
- Add manifest metadata support for screenshot masking options
- Add snapshot tests with Dropbox Differ library for visual regression
- Update CLAUDE.md with dependency management guidelines

Masking requires the `sentry-android-replay` module to be present at runtime. Without it, screenshots are captured without masking.

some example events:

1. https://sentry-sdks.sentry.io/issues/7241812873/events/5652b2955b8d4e638be96a8bed0ddd2c/
2. https://sentry-sdks.sentry.io/issues/7235688870/events/9412e7165a4e4be98d3b9d9ceaea70b6/

Also verified that pixelCopy strategy still works fine after the change, replay here:
https://sentry-sdks.sentry.io/explore/replays/cb4a531e5851484cb547c93e31a9c9f3/

Closes #3286

Manually + automated

## :pencil: Checklist

- [] I updated the docs if needed.
- [] I updated the wizard if needed.

## :crystal_ball: Next steps
Docs and maybe wizard

### DIFF
--- a/sentry-android-core/src/test/java/io/sentry/android/core/ScreenshotEventProcessorTest.kt
+++ b/sentry-android-core/src/test/java/io/sentry/android/core/ScreenshotEventProcessorTest.kt
@@ -339,7 +339,7 @@ class ScreenshotEventProcessorTest {
   }
 
   @Test
-  fun `when masking is configured but replay is not available, screenshot is still captured without masking`() {
+  fun `when masking is configured but replay is not available, screenshot is not captured`() {
     val sut = fixture.getSut(attachScreenshot = true, isReplayAvailable = false)
     fixture.options.screenshot.setMaskAllText(true)
     val hint = Hint()
@@ -349,7 +349,7 @@ class ScreenshotEventProcessorTest {
     val event = fixture.mainProcessor.process(getEvent(), hint)
     sut.process(event, hint)
 
-    assertNotNull(hint.screenshot)
+    assertNull(hint.screenshot)
   }
 
   @Test


sentry-android-core/src/main/java/io/sentry/android/core/ScreenshotEventProcessor.java
```diff
@@ -89,25 +115,87 @@ public ScreenshotEventProcessor(
...
+      }
+
+      return mutableBitmap;
+    } catch (Throwable e) {
+      options.getLogger().log(SentryLevel.ERROR, "Failed to mask screenshot", e);
+      // Recycle the copy if we created one, to avoid memory leak
+      if (createdCopy && !mutableBitmap.isRecycled()) {
+        mutableBitmap.recycle();
+      }
+      return screenshot;
```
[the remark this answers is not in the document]
romtsn: @markushi wdyt? makes sense?

sentry-android-core/src/main/java/io/sentry/android/core/SentryScreenshotOptions.java
```diff
@@ -0,0 +1,53 @@
...
+   * {@inheritDoc}
+   *
+   * <p>When enabling image masking for screenshots, this also adds masking for WebView, VideoView,
+   * and media player views (ExoPlayer, Media3) since they may contain sensitive content.
+   */
+  @Override
+  public void setMaskAllImages(final boolean maskAllImages) {
+    super.setMaskAllImages(maskAllImages);
+    if (maskAllImages) {
+      addSensitiveViewClasses();
```
romtsn, line 37: I guess this is questionable if you enable masking images, we will also mask other media. But I didn't want to introduce new flag, so made it dependant on this one. We can document this, or introduce a new flag. Thoughts @markushi ? 

I think adding those classes by default is probably a no-go for screenshots, since it's a single frame screenshot, not a sequence like in replay, where potentially a lot more could be leaked

sentry-android-core/src/main/java/io/sentry/android/core/ScreenshotEventProcessor.java
```diff
@@ -89,25 +115,87 @@ public ScreenshotEventProcessor(
...
         captureScreenshot(
             activity, options.getThreadChecker(), options.getLogger(), buildInfoProvider);
     if (screenshot == null) {
       return event;
     }
 
+    // Apply masking if enabled and replay module is available
+    if (isMaskingEnabled()) {
+      final @Nullable View rootView =
+          activity.getWindow() != null
```
markushi: We should use `.peekDecorView();` instead of `.getDecorView()`

sentry-android-core/src/main/java/io/sentry/android/core/ScreenshotEventProcessor.java
```diff
@@ -89,25 +115,87 @@ public ScreenshotEventProcessor(
...
-    final Bitmap screenshot =
+    Bitmap screenshot =
         captureScreenshot(
             activity, options.getThreadChecker(), options.getLogger(), buildInfoProvider);
     if (screenshot == null) {
       return event;
     }
 
+    // Apply masking if enabled and replay module is available
+    if (isMaskingEnabled()) {
```
markushi, line 128: I don't this is guaranteed to be always executed on the main thread, right? We might need to enforce this, otherwise random crashes could occur while iterating the VH.

sentry-android-core/src/main/java/io/sentry/android/core/ScreenshotEventProcessor.java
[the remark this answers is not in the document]
markushi: > The safer behavior would be to skip attaching the screenshot entirely if masking was requested but couldn't be applied.  

Yes, I think we should play safe here and only attach screenshots if masking was successful

sentry-android-core/src/main/java/io/sentry/android/core/SentryScreenshotOptions.java
markushi: Yeah, I would just document it.

sentry-android-core/src/test/resources/Tongariro.jpg
markushi, line 1: do we need any license attribution for this file?

romtsn: good question, but we already have it in sentry-android-replay/src/test/resources/ and sentry/src/test/resources/ so I'd say no. Also, looks like it'd been taken from this website https://pixabay.com/images/search/tongariro/ which claims the license is [free](https://pixabay.com/service/license-summary/) for use (we don't distribute it)

sentry-android-core/src/test/java/io/sentry/android/core/SentryScreenshotOptionsTest.kt
```diff
@@ -0,0 +1,136 @@
...
+
+  @Test
+  fun `calling setMaskAllImages true multiple times does not duplicate classes`() {
+    val options = SentryScreenshotOptions()
+    options.setMaskAllImages(true)
+    options.setMaskAllImages(true)
+    options.setMaskAllImages(true)
+
+    // CopyOnWriteArraySet should prevent duplicates
+    assertEquals(6, options.maskViewClasses.size)
```
szijpeter: let's not forget about `PreviewView` here
```suggestion
    assertEquals(7, options.maskViewClasses.size)
```

and in other tests like 
- `setMaskAllImages true adds ImageView and sensitive view classes`
- `enabling both maskAllText and maskAllImages adds all expected classes`
etc.

I'm wondering if declaring an immutable collection of these class names would be a good idea? 🤔

sentry-android-core/src/test/java/io/sentry/android/core/SentryScreenshotOptionsTest.kt
```diff
@@ -0,0 +1,122 @@
...
+
+    assertTrue(options.maskViewClasses.contains(SentryMaskingOptions.IMAGE_VIEW_CLASS_NAME))
+    assertTrue(options.maskViewClasses.contains(SentryMaskingOptions.WEB_VIEW_CLASS_NAME))
+    assertTrue(options.maskViewClasses.contains(SentryMaskingOptions.VIDEO_VIEW_CLASS_NAME))
+    assertTrue(
+      options.maskViewClasses.contains(SentryMaskingOptions.ANDROIDX_MEDIA_VIEW_CLASS_NAME)
+    )
+    assertTrue(
+      options.maskViewClasses.contains(SentryMaskingOptions.ANDROIDX_MEDIA_VIEW_CLASS_NAME)
+    )
```
szijpeter, line 48: dupe

```suggestion
    assertTrue(
      options.maskViewClasses.contains(SentryMaskingOptions.CAMERAX_PREVIEW_VIEW_CLASS_NAME)
    )
    assertTrue(
      options.maskViewClasses.contains(SentryMaskingOptions.ANDROIDX_MEDIA_VIEW_CLASS_NAME)
    )
```

sentry-android-core/src/test/java/io/sentry/android/core/SentryScreenshotOptionsTest.kt
romtsn: @szijpeter thanks for your vigilance, addressed this 👍 

> I'm wondering if declaring an immutable collection of these class names would be a good idea? 

I guess we could do that for the predefined classes, but I'd keep it as-is for now - don't expect this list to change a lot overtime, so we're probably good for the observable future

sentry/src/main/java/io/sentry/SentryMaskingOptions.java
```diff
@@ -0,0 +1,129 @@
...
+
+  @NotNull
+  public Set<String> getUnmaskViewClasses() {
+    return this.unmaskViewClasses;
+  }
+
+  public void addUnmaskViewClass(final @NotNull String className) {
+    this.unmaskViewClasses.add(className);
+    this.maskViewClasses.remove(className);
+  }
```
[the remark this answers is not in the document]
romtsn: yeah this actually had been a bug I'd say, so we fixed it for both with this change

sentry-android-core/src/main/java/io/sentry/android/core/ScreenshotEventProcessor.java
markushi, line 92: if any masking is configured, but replay is not available - we should skip straight away and not attach any screenshot right?
```suggestion
    }

    if (!isReplayAvailable && !options.getScreenshot().getMaskViewClasses().isEmpty()) {
        return event;
    }
```

sentry-android-core/src/main/java/io/sentry/android/core/ScreenshotEventProcessor.java
```diff
@@ -71,6 +90,15 @@ public ScreenshotEventProcessor(
...
       return event;
     }
+    if (!isReplayAvailable && !options.getScreenshot().getMaskViewClasses().isEmpty()) {
+      if (!isReplayModuleAbsenceLogged.getAndSet(true)) {
+        options
+            .getLogger()
+            .log(SentryLevel.WARNING, "Screenshot masking requires sentry-android-replay module");
+      }
+      return event;
+    }
```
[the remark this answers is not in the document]
romtsn: that is intended

PR #5093, opened 2026-02-11
Fix thread leak caused by eager creation of `SentryExecutorService` in `SentryOptions`

### LINKED ISSUE #5079: Sentry Spring Boot starter leaves daemon thread running on undeploy
### Integration

sentry-spring-boot-starter-jakarta

### Java Version

25

### Version

8.27.0

### Steps to Reproduce

1. Create a minimal Spring Boot application with the Sentry Spring Boot starter enabled.
2. Deploy the application as a WAR to Apache Tomcat.
3. Start Tomcat and verify the application initializes normally.
4. Undeploy or redeploy the application (or shut down Tomcat with the app deployed).
5. Observe Tomcat logs during undeploy.

### Expected Result

- Executor services created by the Sentry Spring Boot starter are shut down when the application context is closed.
- No application-created threads (daemon or non-daemon) remain associated with the web application classloader after undeploy.
- Tomcat does not log clearReferencesThreads warnings or report classloader leaks.

### Actual Result

- During application undeploy, Tomcat logs the following warning:

```
The web application [sentry-thread-demo-0.0.1-SNAPSHOT] appears to have started a thread named
[SentryExecutorServiceThreadFactory-0] but has failed to stop it.
This is very likely to create a memory leak.
```

- The thread `SentryExecutorServiceThreadFactory-0` remains alive after the Spring application context has been shut down.
- This can be directly observed using a profiler or thread dump taken after undeploy, where the Sentry-created thread is still present.
- Although the thread is configured as a daemon thread, it continues running and retains references to the web application classloader.
- As a result, Tomcat’s classloader leak detection is triggered, and repeated redeploys risk accumulating classloader-related memory leaks.

adinauer:
## :scroll: Description

When creating (non empty) `SentryOptions` we previously eagerly created `SentryExecutorService` which in turn creates a thread named `SentryExecutorServiceThreadFactory-0`.

There were cases where we created options that ended up unused but we failed to clean those up, e.g.:
- logging integrations that created `SentryOptions` and due to lower priority of the call to `Sentry.init` ended up unused
- `SentryProperties` (a subclass of `SentryOptions`) were created even if the SDK wasn't enabled but we never cleaned up because the SDK was never enabled

Fixes #5079 

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
 ### Fixes
 
 - When merging tombstones with Native SDK, use the tombstone message if the Native SDK didn't explicitly provide one. ([#5095](https://github.com/getsentry/sentry-java/pull/5095))
+- Fix thread leak caused by eager creation of `SentryExecutorService` in `SentryOptions` ([#5093](https://github.com/getsentry/sentry-java/pull/5093))
+  - There were cases where we created options that ended up unused but we failed to clean those up.
 
 ### Dependencies
 


sentry-android-core/src/main/java/io/sentry/android/core/SentryPerformanceProvider.java
```diff
@@ -165,6 +165,7 @@ private void createAndStartContinuousProfiler(
       return;
     }
 
+    final @NotNull SentryExecutorService startupExecutorService = new SentryExecutorService();
```
adinauer, line 168: This creates a new thread for app startup profiling. My current understanding is that this is later reused and not just used for app startup profiling. This means we're not really leaking it since it ends up being used.

If there's cases where app startup profiler is abandoned, we'll need some sort of shutdown hook to also shut down the `SentryExecutorService`. LMK if I should implement that as well just to be safe here.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidProfiler.java
romtsn, line 102: a bit of a breaking change for RN, but should be fine since it's marked Internal anyway: https://github.com/getsentry/sentry-react-native/blob/74979ac1511f9110e88ba92af2f04ec892bb89e1/packages/core/android/src/main/java/io/sentry/react/RNSentryModuleImpl.java#L1018

romtsn: cc @antonis @alwx

antonis: Thank you for the heads up @romtsn 🙇 
As you said this seems more like an internal implementation breakage (which we should handle with the Android bump) and not a user facing thing.

sentry-android-core/src/main/java/io/sentry/android/core/SentryPerformanceProvider.java
romtsn: I think it's fine as-is. Plus, afaik we'll need to deprecate/delete the transaction profiler soon (with the next major likely or span-only), so I wouldn't pour much into it

romtsn: ah I see this also affects continuous profiler too.. but I don't think we abandon it anywhere (unless process termination), so we hsould be good.

Btw, speaking of transaction profiler - I guess we should change it to accept a lambda, too (line 216)?

adinauer: > ah I see this also affects continuous profiler too.. but I don't think we abandon it anywhere (unless process termination), so we hsould be good.

I'd assume once we get rid of transaction profiler, there wouldn't be much reason to abandon it.
I'll go through some things again to check for cases but I assume we can tackle those as a follow up since this isn't new behaviour but already existed previously.

> Btw, speaking of transaction profiler - I guess we should change it to accept a lambda, too (line 216)?

Yeah, will change that too.

adinauer: Oops, I created the ctor but didn't use it. It's being used now.

adinauer: We agreed to ignore the edge case of an abandoned app start profiler for now since it only happens when switching between continuous and transaction profiling or when switching from enabled app startup profiling to disabled. And it'll only be leaking a single thread for a single app run. With plans to remove transaction based profiling in the future, this will become even more of an edge case.

sentry-android-core/src/main/java/io/sentry/android/core/SentryPerformanceProvider.java
```diff
@@ -165,6 +165,7 @@ private void createAndStartContinuousProfiler(
...
     final @NotNull IContinuousProfiler appStartContinuousProfiler =
         new AndroidContinuousProfiler(
             buildInfoProvider,
```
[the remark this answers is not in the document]
adinauer: We accept this edge case for now.
https://github.com/getsentry/sentry-java/pull/5093#discussion_r2798336068

Issue #4400, opened 2025-05-13
Session Replay is unable to mask content when using the Camera


chris-mitchell:
### Integration

sentry-android

### Build System

Gradle

### AGP Version

8.7.3

### Proguard

Enabled

### Version

8.11.1

### Steps to Reproduce

1. Enable Sentry's Session Replay Feature
2. Launch the Camera from your app similar to https://developer.android.com/media/camera/camera-intents
3. Trigger an error / crash
4. View the result in Sentry

### Expected Result

Contents from the Camera would be masked as the user is capturing their image

### Actual Result

Contents from the Camera are shown in the Replay

We don't have a way to filter this since we don't control the app or view class used to capture the camera. Perhaps if we launch an Activity that isn't defined in our Manifest we could block it entirely?

chris-mitchell:
Good question - this error was for the Camera Intents approach.  However, we seem to use both in our app.  We use Camera Intents in some legacy areas, and we use Camera X via MLKit.

romtsn:
> Good question - this error was for the Camera Intents approach. However, we seem to use both in our app. We use Camera Intents in some legacy areas, and we use Camera X via MLKit.

thanks, I believe the issue you were seeing is related to Camera X, because when you launch a camera intent we have no way to capture it as part of replay since it's in a different app process.

romtsn:
@chris-mitchell I've tried reproducing this both using intents and Camera X (from one of the [samples](https://github.com/android/camera-samples/)) but wasn't able to. The replay just shows a black screen and that's it:

<img width="720" height="1105" alt="Image" src="https://github.com/user-attachments/assets/ff40b24f-9a1a-48f4-ae27-ba2b92874741" />

Would you mind providing a reproducible sample/giving more details into your setup?

romtsn:
(forgot to mention - for camera intent it actually cannot event capture and just shows missing segments because camera app is a different process that we're not able to capture)

romtsn:
@chris-mitchell I'm closing this issue due to the lack of reproduction but feel free to comment here with more details. Thanks!

romtsn:
@lucas-zimerman reproduced this in Maui using https://github.com/thomasgalliker/CameraScanner.Maui (which under the hood still uses androidx.camera, but maybe it's a different configuration so worth to look into it to see if it's reproducible)

szijpeter:
I've looked into this a bit and I think I've found a reproducer path.

[...]

PR #5097, opened 2026-02-12
Session Replay: Add PreviewView to default masked view classes

szijpeter:
## :scroll: Description
[PreviewView](https://developer.android.com/reference/androidx/camera/view/PreviewView) is the view used to display the camera feed for CameraX's [Preview](https://developer.android.com/reference/androidx/camera/core/Preview) use case. Adding it to Replay option defaults aligns with existing sensitive defaults (`TextView`, `ImageView`, `WebView`, `VideoView`, player views) in [`SentryReplayOptions`](https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/SentryReplayOptions.java).

It's a low-risk, sensible default imo.

* resolves: #4400

Reproduction path:

1. Use `PreviewView` in [COMPATIBLE](https://developer.android.com/reference/androidx/camera/view/PreviewView.ImplementationMode#COMPATIBLE) mode 
2. Keep screenshot strategy on the **default path** (PixelCopy), i.e. do **not** set Canvas.

👉🏼 https://github.com/szijpeter/sentry-java/tree/repro/4400

Why these two matter:
- `COMPATIBLE` forces the `TextureView` path where camera pixels are composited into the normal view hierarchy and can show up in captures.
- `PERFORMANCE` typically uses `SurfaceView`, which often gives black/non-captured camera content and can hide the bug (false negative). note that it might still fall back to `TextureView` in certain cases.
- setting screenshot strategy to `canvas` is a different capture mode and masks broadly; it does not rely on the defaults 

| Repro | Repro + Fix |
| ------------- | ------------- |
| repro  | repro+fix |

## Workarounds for older SDK versions

- explicitly mask `PreviewView` in app code via [`addMaskViewClass`](https://javadoc.io/static/io.sentry/sentry/8.16.0/io/sentry/SentryReplayOptions.html#addMaskViewClass(java.lang.String))
- and/or set screenshot strategy to `canvas`

## :pencil: Checklist

## :crystal_ball: Next steps
adding screenshot tests? could be worth it, but I feel like maybe out of scope for this PR

### DIFF
--- a/sentry/api/sentry.api
+++ b/sentry/api/sentry.api
@@ -7938,3 +7938,4 @@ public final class io/sentry/vendor/gson/stream/MalformedJsonException : java/io
 	public fun <init> (Ljava/lang/String;Ljava/lang/Throwable;)V
 	public fun <init> (Ljava/lang/Throwable;)V
 }
+


szijpeter:
> @szijpeter amazing investigation and detailed report, thanks a bunch! approving right away and going to merge right after 👍

thanks @romtsn, happy to help :)

there was a small mistake, I've accidentally removed an empty line from the api file which made the `apiCheck` task [fail](https://github.com/getsentry/sentry-java/actions/runs/21981028860/job/63503631529#step:6:944).
should be fixed now (https://github.com/getsentry/sentry-java/pull/5097/commits/3b1668f78935f5331084b841e8841e68f2c42de3)

PR #5117, opened 2026-02-25
test(android): add proguard tombstone test cases

supervacuus:
## :scroll: Description

This PR sits on top of #5116, but I added it separately because it addresses a different topic: tombstone frames that are ProGuard-obfuscated.

* The tests only check if the obfuscated frames affect the baseline (and I record a different set of Java frames for the snapshot).
* I also modified the build script of the Android sample, so I actually get meta properties for a release build, so I can upload ProGuard mappings for the sample. 

The former should be merged into `master`. We don't have to merge the latter to master, but I think it will be could be useful for @markushi in https://linear.app/getsentry/issue/ANDROID-265/improve-ux-for-mixed-stacktraces (https://github.com/getsentry/sentry/pull/107318), because it allows cross-checking UI changes for proguard.

The PR also allows us to verify that the ProGuard debug-meta is correctly appended to tombstone and ANR events when a corresponding properties file exists in the APK. If a mapping file has been uploaded with `sentry-cli` or the SAGP, the backend correctly triggers ProGuard deobfuscation, but it only deobfuscates the `module`, not the `function`:

image

Not sure if that is expected behavior ([event](https://sentry-sdks.sentry.io/issues/7291642259/events/cac10bc146624741888afa9350049d32/)).

#skip-changelog
(since this is currently only an internal test addition)

## :pencil: Checklist


sentry-samples/sentry-samples-android/build.gradle.kts
```diff
@@ -7,6 +7,7 @@ plugins {
   id("com.android.application")
   alias(libs.plugins.kotlin.android)
   alias(libs.plugins.kotlin.compose)
+  alias(libs.plugins.sentry)
```
markushi, line 10: 👍 I always wondered why we didn't do that, but never got around changing this, thanks! 😅

sentry-samples/sentry-samples-android/build.gradle.kts
```diff
@@ -124,7 +131,8 @@ dependencies {
   implementation(projects.sentryAndroid)
   implementation(projects.sentryAndroidFragment)
   implementation(projects.sentryAndroidTimber)
-  implementation(projects.sentryCompose)
+  // implementation(projects.sentryCompose) // TODO: re-enable, disabled due to R8 duplicate class
```
markushi: do you still have the error lying around somewhere? most likely it's because our gradle plugin "auto-install"s dependencies by default, for the purpose of this example app I think we can disable it

```
sentry {
	autoInstallation {
		enabled = false
	}
}
```

supervacuus: That is the error:

```
> Task :sentry-samples:sentry-samples-android:minifyReleaseWithR8 FAILED
ERROR: /Users/mischan/.gradle/caches/8.14.3/transforms/b98a7878125a49c4a3e65866e68c9856/transformed/classes.jar: R8: Type io.sentry.compose.SentryComposeHelper$Companion is defined multiple times: /Users/mischan/.gradle/caches/8.14.3/transforms/b98a7878125a49c4a3e65866e68c9856/transformed/classes.jar:io/sentry/compose/SentryComposeHelper$Companion.class, /Users/mischan/.gradle/caches/8.14.3/transforms/78a14c3b7a4ec2dc675d192df1aaff02/transformed/sentry-compose-release-runtime.jar:io/sentry/compose/SentryComposeHelper$Companion.class

[Incubating] Problems report is available at: file:///Users/mischan/devel/sentry-java/build/reports/problems/problems-report.html

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':sentry-samples:sentry-samples-android:minifyReleaseWithR8'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.R8Task$R8Runnable
   > Compilation failed to complete, origin: /Users/mischan/.gradle/caches/8.14.3/transforms/b98a7878125a49c4a3e65866e68c9856/transformed/classes.jar:io/sentry/compose/SentryComposeHelper$Companion.class
```

supervacuus: And yes, disabling auto-installation fixes it.

sentry-android-core/src/test/java/io/sentry/android/core/internal/tombstone/TombstoneParserTest.kt
```diff
@@ -566,6 +475,104 @@ class TombstoneParserTest {
...
+      if (thread.id == crashedThreadId) {
+        assert(thread.isCrashed == true)
+        hasCrashedThread = true
+      }
+      assert(thread.stacktrace!!.frames!!.isNotEmpty())
+
+      for (frame in thread.stacktrace!!.frames!!) {
+        assertNotNull(frame.function)
+        if (frame.platform == "java") {
+          assertNotNull(frame.module)
```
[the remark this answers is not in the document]
supervacuus: I don't follow this reasoning. `extractJavaModuleName()` returns null only for bare names without dots (e.g., `"myMethod"`). R8-obfuscated names like `"E3.A"`, `"O0.a"`, `"X0.d"`, etc. still contain dots, so the module is extracted fine (e.g., `module="E3"`, `function="A"`).

PR #5118, opened 2026-02-26
feat(core): [Global Attributes 1] Add scope-level attributes API

adinauer:
## :scroll: Description

Add a new API to set attributes on scopes (`IScope`, `IScopes`, `Sentry`). Attributes are key-value pairs (`SentryAttribute`) that can be set, removed, and queried from scopes. The `CombinedScopeView` merges attributes across scope layers (global, isolation, current) with the current scope taking precedence.

This lays the foundation for scope-level attributes that can be automatically attached to logs, metrics, and other events.

Scope-level attributes allow users to set contextual key-value pairs that are automatically propagated to logs and metrics, similar to how tags work but using the structured `SentryAttribute` type. A follow-up stacked PR will wire these attributes into `LoggerApi` and `MetricsApi`.

- Unit tests for `Scope` attribute operations (set, remove, getAttributes, setAttributes)
- Unit tests for `CombinedScopeView` attribute merging across scope layers

## :pencil: Checklist

## :crystal_ball: Next steps

- Stacked PR to wire scope attributes into `LoggerApi` and `MetricsApi`, including extracted factory methods (`SentryAttributeType.inferFrom`, `SentryLogEventAttributeValue.fromAttribute`)


adinauer:
**PR Stack (Global Attributes):**
1. [#5118](https://github.com/getsentry/sentry-java/pull/5118) — Add scope-level attributes API
2. [#5120](https://github.com/getsentry/sentry-java/pull/5120) — Wire scope attributes into LoggerApi and MetricsApi
3. [#5121](https://github.com/getsentry/sentry-java/pull/5121) — Showcase scope attributes in Spring Boot 4 samples
4. [#5122](https://github.com/getsentry/sentry-java/pull/5122) — Detect integer attribute type correctly for all integer Number subtypes
5. [#5124](https://github.com/getsentry/sentry-java/pull/5124) — Support collections and arrays in log attribute type inference

sentry/src/main/java/io/sentry/HubAdapter.java
```diff
@@ -395,6 +395,26 @@ public void reportFullyDisplayed() {
     return Sentry.getCurrentScopes().metrics();
   }
 
+  @Override
+  public void setAttribute(final @Nullable String key, final @Nullable Object value) {
```
alexander-alderman-webb, line 399: does the `key` need to be `Nullable` for any reason?

It looks like `null` values are ignored anyway further down.

adinauer: This is a general usability thing where we ideally accept null and handle it rather than causing `NullPointerException` for customers. At least in APIs that are often called with input that might just be piped through.

Issue #5193, opened 2026-03-15
Provide a way to set IScopesStorage using SentryOptions


Fabccc:
### Problem Statement

I want to set a scope storage compatible with Quarkus reactive core (Vertx), because the DefaultScopeStorage backed by ThreadLocal variables does not meet the "no pinning" condition, and a single thread can handle multiple requests, so it will cause weird behaviour. Currently I have to use reflection to override the field value, since it's marked private static. I also have to re-create a scope from the global scope, and it's just a duplicate.

### Solution Brainstorm

I think we should be able to set a custom IScopesStorage implementation in SentryOptions (passed to the init method). Inside the "Sentry#initForOpenTelemetryMaybe", we could maybe check for an existing scopeStorage in the SentryOptions parameters. If there is a non null value, then set the "scopesStorage" field value to the value, otherwise, fallback to classic behaviour. It should not impact current users setup.

adinauer:
Hey @Fabccc thanks for opening this issue.

We'll try to PR the required changes soonish.

Fabccc:
Hey @adinauer ! thank you for responding that fast and sending the patch !

PR #5199, opened 2026-03-16
feat(core): Add configurable IScopesStorageFactory to SentryOptions

adinauer:
## :scroll: Description

Adds a new `IScopesStorageFactory` interface and a corresponding `scopesStorageFactory` option on `SentryOptions`. When set, the custom factory takes precedence over the default auto-detection logic in `Sentry.initScopesStorage()`.

This allows users to provide a custom `IScopesStorage` implementation without reflection hacks, which is needed for frameworks where the default `ThreadLocal`-backed `DefaultScopesStorage` is incompatible (e.g. reactive/non-pinning thread models).

Changes:
- New `IScopesStorageFactory` functional interface with a single `create()` method
- `SentryOptions.getScopesStorageFactory()` / `setScopesStorageFactory()` (marked `@Experimental`)
- `Sentry.initScopesStorage()` checks the custom factory first, then falls back to existing auto-detection

Fixes https://github.com/getsentry/sentry-java/issues/5193

`Sentry.initScopesStorage()` decides the storage implementation internally with no user-facing override. Users who need a custom `IScopesStorage` (e.g. backed by Vert.x context instead of `ThreadLocal`) had to use reflection to override the private static field.

- Unit tests in `SentryOptionsTest` — default is null, getter/setter works, can reset to null
- Unit tests in `SentryTest` — custom factory is invoked and `init()` called on the result, null factory falls back to default, custom storage is functional end-to-end

## :pencil: Checklist

## :crystal_ball: Next steps

- Document how to use this for Quarkus/Vert.x reactive applications


sentry/src/main/java/io/sentry/Sentry.java
```diff
@@ -439,10 +439,15 @@ private static void initFatalLogger(final @NotNull SentryOptions options) {
 
   private static void initScopesStorage(SentryOptions options) {
     getScopesStorage().close();
-    if (SentryOpenTelemetryMode.OFF == options.getOpenTelemetryMode()) {
+    final @NotNull LoadClass loadClass = new LoadClass();
+    final @NotNull ILogger logger = NoOpLogger.getInstance();
+    if (options.getScopesStorageFactory() != null) {
+      scopesStorage = options.getScopesStorageFactory().create(loadClass, logger);
```
adinauer: Passing noop logger doesn't make sense

sentry/src/main/java/io/sentry/IScopesStorageFactory.java
```diff
@@ -0,0 +1,11 @@
+package io.sentry;
+
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+/** Factory for creating custom {@link IScopesStorage} implementations. */
+@ApiStatus.Experimental
+public interface IScopesStorageFactory {
```
adinauer, line 8: Opted to keep this separate from `ScopesStorageFactory` since passing `LoadClass` and `ILogger` doesn't seem necessary.

sentry/src/main/java/io/sentry/IScopesStorageFactory.java
```diff
@@ -0,0 +1,11 @@
...
+  @NotNull
+  IScopesStorage create();
```
adinauer: Should we be passing in options here?

adinauer: Probably not necessary as long as it has to be configured in options callback of `Sentry.init` anyways. May make more sense if/when we support SPI.

adinauer:
@sentry review

adinauer:
cursor review

PR #5214, opened 2026-03-19
feat(replay): add beforeErrorSampling callback to Session Replay

romtsn:
## :scroll: Description

Add a `BeforeErrorSamplingCallback` to `SentryReplayOptions` that lets developers filter which errors trigger replay capture. The callback runs **before** the `onErrorSampleRate` dice roll — returning `false` skips `captureReplay()` entirely. Fail-open behavior: if the callback throws, replay capture proceeds normally.

Currently, when an error occurs, `SentryClient` unconditionally calls `captureReplay()` — developers have no way to skip replay capture for specific error types (e.g. handled exceptions, specific error classes).

This mirrors the same feature added in the React Native SDK ([getsentry/sentry-react-native#5393](https://github.com/getsentry/sentry-react-native/pull/5393)).

* related: https://github.com/getsentry/sentry/issues/86905

5 new unit tests in `SentryClientTest.kt`:
- `beforeErrorSampling` returning `false` skips `captureReplay`
- `beforeErrorSampling` returning `true` proceeds with `captureReplay`
- `beforeErrorSampling` not set proceeds with `captureReplay` (default behavior unchanged)
- `beforeErrorSampling` throwing exception proceeds with `captureReplay` (fail-open)
- `beforeErrorSampling` receives correct event and hint

## :pencil: Checklist

## :crystal_ball: Next steps

- Update documentation with usage examples

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Features
 
-- Add `beforeErrorSampling` callback to Session Replay ([#5214](https://github.com/getsentry/sentry-java/pull/5214))
+- Android: Add `beforeErrorSampling` callback to Session Replay ([#5214](https://github.com/getsentry/sentry-java/pull/5214))
   - Allows filtering which errors trigger replay capture before the `onErrorSampleRate` is checked
   - Returning `false` skips replay capture entirely for that error; returning `true` proceeds with the normal sample rate check
   - Example usage:


CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Add `beforeErrorSampling` callback to Session Replay ([#5214](https://github.com/getsentry/sentry-java/pull/5214))
```
```suggestion
- Android: Add `beforeErrorSampling` callback to Session Replay ([#5214](https://github.com/getsentry/sentry-java/pull/5214))
```

sentry/src/main/java/io/sentry/SentryClient.java
```diff
@@ -231,7 +231,25 @@ private boolean shouldApplyScopeData(final @NotNull CheckIn event, final @NotNul
...
+              .log(
+                  SentryLevel.ERROR,
+                  "The beforeErrorSampling callback threw an exception. Proceeding with replay capture.",
+                  e);
+          shouldCaptureReplay = true;
+        }
+      }
+      if (shouldCaptureReplay) {
+        options.getReplayController().captureReplay(event.isCrashed());
+      }
```
markushi, line 252: No too sure about this, but should we report lost replays as well here?
Something along those lines ~
```suggestion
      } else {
      	if (options.getReplayController().isRecording()) {
      		 options
      		 	.getClientReportRecorder()
            	.recordLostEvent(DiscardReason.BEFORE_SEND, DataCategory.Replay);
      	}
      }
```

markushi (approved):
LGTM, once small concern, but that doesn't affect overall functionality.

sentry/src/main/java/io/sentry/SentryClient.java
romtsn: good question! but i think this doesn't apply here, because there's technically no "replay" at this point (I know we're buffering it, but until a trigger occurs there's technically no replay running). Also cross-checked with the JS implementation and they don't seem to do that either, so I'd keep it simple for now

PR #5251, opened 2026-03-31
feat(profiling): Add Android ProfilingManager (Perfetto) support

43jay:
## 📜 Description

Adds support for Android's `ProfilingManager` API (API 35+), which captures continuous profiles (profile chunks) using OS-level Perfetto stack sampling instead of the legacy `Debug.startMethodTracingSampling` engine.

**Behavior by API level:**
- **API 35+** — Perfetto profiling via `ProfilingManager` is used automatically. No configuration needed.
- **API < 35** — the legacy `Debug`-based profiler is used, controlled by the new `enableLegacyProfiling` option.

## ⚙️ New options

| Option | Manifest key | Default | Description |
|---|---|---|---|
| `enableLegacyProfiling` | `io.sentry.profiling.enable-legacy-profiling` | `true` | Enables the legacy `Debug`-based profiler on API < 35. Set to `false` to disable profiling on those devices. Has no effect on API 35+. Will be deprecated in the next major and removed after. |

## ⚠️ Limitations

- Perfetto profiling requires **API 35+**. On API < 35 with `enableLegacyProfiling = false`, no profiling data is collected (no silent fallback).
- Only **continuous profiling** (profile chunks) is supported via `ProfilingManager` — transaction-based (legacy) profiling is unaffected.
- **App-start profiling** is not supported on the Perfetto path (`ProfilingManager` doesn't support it).
- Thread names may be missing in captured traces (`ProfilingManager` doesn't include `linux.process_stats` data).

## 💚 How did you test it?

- Unit tests
- Manual testing on Pixel Fold AVD (API 35) — verified Perfetto chunks captured with `content_type: "perfetto"` and inspected in [Perfetto UI](https://ui.perfetto.dev/).

## 📝 Checklist


sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
```diff
@@ -0,0 +1,399 @@
...
+          .getExecutorService()
+          .submit(
+              () -> {
+                if (isClosed.get()) {
+                  return;
+                }
+                scopes.captureProfileChunk(builder.build(options));
+              });
+    } catch (Throwable e) {
+      options.getLogger().log(SentryLevel.DEBUG, "Failed to send profile chunk.", e);
```
43jay, line 428: ERROR

sentry/src/main/java/io/sentry/SentryEnvelopeItemHeader.java
```diff
@@ -147,6 +219,9 @@ public void serialize(final @NotNull ObjectWriter writer, final @NotNull ILogger
     if (itemCount != null) {
       writer.name(JsonKeys.ITEM_COUNT).value(itemCount);
     }
+    if (metaLength != null) {
+      writer.name(JsonKeys.META_LENGTH).value(metaLength);
+    }
```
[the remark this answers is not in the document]
43jay: fixed

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
```diff
@@ -0,0 +1,399 @@
...
+            return;
+          }
+          break;
+      }
+      if (!isRunning()) {
+        logger.log(SentryLevel.DEBUG, "Started Profiler.");
+        startInternal();
+      }
+    }
+  }
```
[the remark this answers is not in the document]
43jay: fixed

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
```diff
@@ -0,0 +1,399 @@
...
+  public void close(final boolean isTerminating) {
+    try (final @NotNull ISentryLifecycleToken ignored = lock.acquire()) {
+      activeTraceCount = 0;
+      shouldStop = true;
+      if (isTerminating) {
+        stopInternal(false);
+        isClosed.set(true);
+      }
+    }
+  }
```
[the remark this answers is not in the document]
43jay: don't believe this is valid. 

`close(isTerminating=true)` does indeed call stopInternal immediately ✅

`close(isTerminating=false)` was flagged by this comment.
It's currently called by LifecycleWatcher [when app backgrounded fires](https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/LifecycleWatcher.java#L117)

presumably to let the profiler continue for the remainder of the current chunk then stop (shouldStop=true)

1) matches legacy behaviour
2) seems reasonable to catch the chunk around app backgrounded (force stopping at this point will mean sentry profiler never captures stacks around app backgrounding events => 👎

=> **will not address**

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
```diff
@@ -0,0 +1,569 @@
...
+      chunkId = SentryId.EMPTY_ID;
+      return;
+    }
+
+    final @NotNull IScopes scopes = resolveScopes();
+    final @NotNull SentryOptions options = scopes.getOptions();
+
+    final @NotNull Map<String, ProfileMeasurement> measurements = chunkMeasurements.stop();
+
+    final @Nullable File traceFile = perfettoProfiler.endAndCollect();
```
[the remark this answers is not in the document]
43jay: same as https://github.com/getsentry/sentry-java/pull/5251#discussion_r3060574710

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
```diff
@@ -0,0 +1,569 @@
...
+    this.executorServiceSupplier = executorServiceSupplier;
+    this.perfettoProfilerSupplier = perfettoProfilerSupplier;
+  }
+
+  @Override
+  public void startProfiler(
+      final @NotNull ProfileLifecycle profileLifecycle,
+      final @NotNull TracesSampler tracesSampler) {
+    try (final @NotNull ISentryLifecycleToken ignored = lock.acquire()) {
+      shouldStop = false;
```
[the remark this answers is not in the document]
43jay: https://docs.sentry.io/platforms/android/profiling/#enabling-ui-profiling

^MANUAL and TRACE are mutualy exclusive, enforced in `Scopes`. still addressed; simple fix => [63704](https://github.com/getsentry/sentry-java/pull/5251/changes/63704856f4872275601790403683725593a3a395)

sentry/src/test/java/io/sentry/SentryEnvelopeItemTest.kt
```diff
@@ -596,6 +597,69 @@ class SentryEnvelopeItemTest {
...
+          ProfileChunk.PLATFORM_ANDROID,
+        )
+        .setContentType("perfetto")
+        .build(fixture.options)
+
+    val item = SentryEnvelopeItem.fromPerfettoProfileChunk(profileChunk, fixture.serializer)
+
+    assertEquals(SentryItemType.ProfileChunk, item.header.type)
+    assertEquals("application/octet-stream", item.header.contentType)
+    assertNotNull(item.header.metaLength)
```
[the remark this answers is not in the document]
43jay: fixed

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -335,16 +337,25 @@ private static void setupProfiler(
...
-                options.getProfilingTracesHz(),
-                () -> options.getExecutorService()));
+            options.isUseProfilingManager()
+                ? new PerfettoContinuousProfiler(
+                    buildInfoProvider,
+                    options.getLogger(),
+                    frameMetricsCollector,
+                    () -> options.getExecutorService(),
+                    () ->
+                        new PerfettoProfiler(context.getApplicationContext(), options.getLogger()))
```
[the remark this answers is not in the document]
43jay: fixed

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
```diff
@@ -68,13 +68,29 @@ public class AndroidContinuousProfiler
   private final AutoClosableReentrantLock lock = new AutoClosableReentrantLock();
   private final AutoClosableReentrantLock payloadLock = new AutoClosableReentrantLock();
 
-  public AndroidContinuousProfiler(
+  public static AndroidContinuousProfiler createLegacy(
```
markushi: I don't think we need that, I'd keep the public constructor as-is instead. Instead let's add a comment to outline that this profiler is considered legacy.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
```diff
@@ -152,21 +169,24 @@ public void startProfiler(
     }
   }
 
-  private void initScopes() {
```
markushi, line 155: I also don't think we need this change at all. Let's keep the legacy code as-is.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
markushi: Yeah, this should be guarded by API level checks. Did you mean to do this in this PR? https://github.com/getsentry/sentry-java/pull/5290
Please directly address it within this PR instead.

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
```diff
@@ -0,0 +1,569 @@
...
+public class PerfettoContinuousProfiler
+    implements IContinuousProfiler, RateLimiter.IRateLimitObserver {
+  private static final long MAX_CHUNK_DURATION_MILLIS = 60000;
+
+  private final @NotNull ILogger logger;
+  private final @NotNull LazyEvaluator.Evaluator<ISentryExecutorService> executorServiceSupplier;
+  private final @NotNull BuildInfoProvider buildInfoProvider;
+  private final @NotNull LazyEvaluator.Evaluator<PerfettoProfiler> perfettoProfilerSupplier;
+
+  private @Nullable PerfettoProfiler perfettoProfiler = null;
```
markushi, line 82:
```suggestion
  private @Nullable volatile PerfettoProfiler perfettoProfiler = null;
```

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
```diff
@@ -0,0 +1,569 @@
...
+   * <p>Performance data is collected by the {@link CompositePerformanceCollector}'s Timer thread
+   * every 100ms and returned as a list on {@code stop()}.
+   */
+  private static class ChunkMeasurementCollector {
+    private final @NotNull SentryFrameMetricsCollector frameMetricsCollector;
+    private @Nullable String frameMetricsListenerId = null;
+    private @Nullable CompositePerformanceCollector performanceCollector = null;
+    private @Nullable String chunkId = null;
+
+    private final @NotNull ConcurrentLinkedDeque<ProfileMeasurementValue>
```
markushi, line 449: That's quite some duplicate code similar to `AndroidProfiler`, ideally we can re-use the `ChunkMeasurementCollector` in both places.

markushi (dismissed):
Looking good overall! I left a few comments.

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
```diff
@@ -0,0 +1,569 @@
...
+        measurements.put(
+            ProfileMeasurement.ID_FROZEN_FRAME_RENDERS,
+            new ProfileMeasurement(
+                ProfileMeasurement.UNIT_NANOSECONDS, frozenFrameRenderMeasurements));
+      }
+      if (!screenFrameRateMeasurements.isEmpty()) {
+        measurements.put(
+            ProfileMeasurement.ID_SCREEN_FRAME_RATES,
+            new ProfileMeasurement(ProfileMeasurement.UNIT_HZ, screenFrameRateMeasurements));
+      }
```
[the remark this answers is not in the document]
43jay: fixed -> https://github.com/getsentry/sentry-java/pull/5251/changes/268a1e0c624379c6e214fe172cfccd25011daa71

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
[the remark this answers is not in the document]
43jay: fixed -> https://github.com/getsentry/sentry-java/pull/5251/changes/35b2bcb5caba57529a40daf9386fbf8a273d5228

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
43jay: nps. I started working in this file before creating PerfettoContProfiler so that's where the changes came from

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
43jay: hmm, that's strange. I wasn't aware of that PR. It looks like cursor took matters into its own hands there.. i used the web tool but it was taking ages and the solution didn't make sense ultimately so i thought I canned it.

sentry/api/sentry.api
```diff
@@ -2065,9 +2065,11 @@ public final class io/sentry/ProfileChunk : io/sentry/JsonSerializable, io/sentr
 	public fun equals (Ljava/lang/Object;)Z
 	public fun getChunkId ()Lio/sentry/protocol/SentryId;
 	public fun getClientSdk ()Lio/sentry/protocol/SdkVersion;
+	public fun getContentType ()Ljava/lang/String;
 	public fun getDebugMeta ()Lio/sentry/protocol/DebugMeta;
 	public fun getEnvironment ()Ljava/lang/String;
 	public fun getMeasurements ()Ljava/util/Map;
+	public fun getMetaLength ()I
```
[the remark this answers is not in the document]
43jay: fixed c072809c3d

sentry/src/main/java/io/sentry/SentryEnvelopeItemHeader.java
```diff
@@ -147,6 +216,10 @@ public void serialize(final @NotNull ObjectWriter writer, final @NotNull ILogger
     if (itemCount != null) {
       writer.name(JsonKeys.ITEM_COUNT).value(itemCount);
     }
+    final @Nullable Integer resolvedMetaLength = getMetaLength();
+    if (resolvedMetaLength != null) {
+      writer.name(JsonKeys.META_LENGTH).value(resolvedMetaLength);
+    }
     writer.name(JsonKeys.LENGTH).value(getLength());
     if (unknown != null) {
```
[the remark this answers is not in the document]
43jay: fixed 1a6742c13ff56ef88eed70c143137b2f42f51fd1

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
43jay: fixed a1e703e2d0278d3730237ecee555deb73613233a

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
43jay: fixed, see prev comment

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
43jay: see commit 35b2bcb

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
43jay: @markushi  I don't believe `volatile` is needed here.

The `perfettoProfiler` is only ever written under a lock. locks provide a superset of what `volatile` guarantees. 

Actually i intentionally removed `volatile` from everywhere in the new file - i agree with the tendency to consider it a bad pattern in java.

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
```diff
@@ -0,0 +1,573 @@
...
+            ProfileMeasurement.ID_MEMORY_FOOTPRINT,
+            new ProfileMeasurement(ProfileMeasurement.UNIT_BYTES, memoryUsageMeasurements));
+      }
+      if (!nativeMemoryUsageMeasurements.isEmpty()) {
+        measurements.put(
+            ProfileMeasurement.ID_MEMORY_NATIVE_FOOTPRINT,
+            new ProfileMeasurement(ProfileMeasurement.UNIT_BYTES, nativeMemoryUsageMeasurements));
+      }
+    }
+  }
```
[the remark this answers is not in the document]
43jay: same comment as https://github.com/getsentry/sentry-java/pull/5251#discussion_r3099208187

sentry-samples/sentry-samples-android/src/main/AndroidManifest.xml
```diff
@@ -163,6 +164,10 @@
     <meta-data
       android:name="io.sentry.traces.profiling.start-on-app-start"
       android:value="true" />
+    <!-- Enable ProfilingManager (Perfetto) on API 35+ -->
+    <meta-data
+      android:name="io.sentry.profiling.use-profiling-manager"
+      android:value="true" />
```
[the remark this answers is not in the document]
43jay: @markushi wyt? Use the new API in sample app or leave in the old API?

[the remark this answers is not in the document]
markushi: IMHO this is fine, let's keep on using the new APIs

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
markushi: Oh I see, the reads are under the lock as well, all good then

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
markushi: Reverted the constructor change (already on HEAD). Added a legacy comment above the class.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
markushi: Verified — `AndroidContinuousProfiler` is identical to `main`, the `initScopes` rename was already on main (not a change from this PR).

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
markushi: Already addressed — the code now guards with `buildInfoProvider.getSdkInfoVersion() >= Build.VERSION_CODES.VANILLA_ICE_CREAM` and logs a warning on lower API levels.

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
markushi: All reads/writes to `perfettoProfiler` are inside the `AutoClosableReentrantLock`. `ReentrantLock.unlock()` → `lock()` already provides the happens-before guarantee, so `volatile` is not needed for correctness. Happy to add it as a defensive signal if you prefer though.

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
markushi: Let's keep it as-is for now, as in the long run we can hopefully drop the legacy `AndroidProfiler`

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
markushi: is accessed within the same lock, so volatile is not needed here.

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
markushi: Done, the measurement collection code has been extracted into a ChunkMeasurementCollector that is shared between both profilers.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
markushi: Added a Javadoc comment marking this profiler as legacy. Constructor is kept as-is.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
markushi: Kept the legacy code as-is, no changes.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
markushi: Done, the API level guard is now directly in this PR. PerfettoContinuousProfiler is only instantiated on API 35+, with a warning log on lower levels.

sentry-android-core/src/test/java/io/sentry/android/core/ContinuousProfilerTestCases.kt
```diff
@@ -0,0 +1,194 @@
...
+
+fun IContinuousProfiler.testCloseWithoutTerminatingStopsAfterChunk(mocks: ProfilerMocks) {
+  startProfiler(ProfileLifecycle.MANUAL, mocks.tracesSampler)
+  startProfiler(ProfileLifecycle.TRACE, mocks.tracesSampler)
+  assertTrue(isRunning)
+  close(false)
+  assertTrue(isRunning)
+  mocks.executor.runAll()
+  assertFalse(isRunning)
+}
```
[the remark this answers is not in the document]
markushi: As we kept the old AndroidContinuousProfiler as is and have no changes planned, I think this is fine as-is.

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -645,6 +645,13 @@ public class SentryOptions {
    */
   private boolean startProfilerOnAppStart = false;
 
+  /**
+   * When true, the SDK uses Android's {@code ProfilingManager} (Perfetto-based stack sampling) on
+   * API 35+ devices. On older devices where ProfilingManager is not available, no profiling data is
+   * collected — the legacy {@code Debug}-based profiler is not used as a fallback.
+   */
+  private boolean useProfilingManager = false;
```
romtsn: hm, I'm wondering if we should just kill this flag altogether and always use Perfetto on API 35+ and fallback to the old profiler on lower API levels? I guess the current status quo is bad enough to consider this a "fix" rather than making it an opt-in?

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
```diff
@@ -373,6 +374,77 @@ private static void ensureAttachmentSizeLimit(
...
+    final @Nullable File traceFile = profileChunk.getTraceFile();
+
+    if (traceFile == null || !traceFile.exists()) {
+      throw new SentryEnvelopeException(
+          String.format(
+              "Dropping perfetto profile chunk, because the trace file '%s' doesn't exist",
+              traceFile != null ? traceFile.getName() : "null"));
+    }
+
+    // Perfetto profile chunks are serialized as [JSON metadata][raw .pftrace bytes] with no
```
romtsn, line 402: could've also used msgpack analogous to the replay payload, right? But if this is done already on the backend side, then I guess we're good as-is :)

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
```diff
@@ -373,6 +374,77 @@ private static void ensureAttachmentSizeLimit(
...
+    SentryEnvelopeItemHeader itemHeader =
+        new SentryEnvelopeItemHeader(
+            SentryItemType.ProfileChunk,
+            () -> cachedItem.getBytes().length,
+            profileChunk.getContentType(),
+            traceFile.getName(),
+            null,
+            profileChunk.getPlatform(),
+            null,
+            (Callable<Integer>) metaLength::get);
```
romtsn: I guess we should still use a lambda () -> here, because method refs are only available with desugaring and on API 24+ from what I recall.

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
```diff
@@ -0,0 +1,637 @@
...
+  @VisibleForTesting
+  @Nullable
+  Future<?> getStopFuture() {
+    return stopFuture;
+  }
+
+  @VisibleForTesting
+  public int getActiveTraceCount() {
+    return activeTraceCount;
+  }
```
romtsn:
```suggestion
```
these look to be unused

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
```diff
@@ -0,0 +1,637 @@
...
+              SentryLevel.DEBUG,
+              "Profile chunk finished, but profiler was already restarted or closed. Skipping.");
+          return;
+        }
+        logger.log(SentryLevel.DEBUG, "Profile chunk finished. Starting a new one.");
+        startInternal();
+      }
+    } else {
+      logger.log(SentryLevel.DEBUG, "Profile chunk finished.");
+    }
```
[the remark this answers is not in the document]
romtsn: I think this is valid?

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
```diff
@@ -0,0 +1,637 @@
...
+  }
+
+  private void sendChunk(
+      final @NotNull ProfileChunk.Builder builder,
+      final @NotNull IScopes scopes,
+      final @NotNull SentryOptions options) {
+    try {
+      options
+          .getExecutorService()
```
romtsn:
```suggestion
      executorServiceSupplier
          .evaluate()
```

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
```diff
@@ -0,0 +1,637 @@
...
+          .submit(
```
romtsn: I think we potentially are calling this already on the executor service thread (in case of the `stopInternal` scheduled call) , and I recall having problems with that (calling submit from within the same executor). I worked it around like this in replay, which we should do here as well: https://github.com/getsentry/sentry-java/blob/dcb15870e50781280dc5c3826f8c47b5df1cea07/sentry-android-replay/src/main/java/io/sentry/android/replay/util/ReplayExecutorService.kt#L18-L22

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoProfiler.java
```diff
@@ -0,0 +1,218 @@
...
+@ApiStatus.Internal
+@RequiresApi(api = Build.VERSION_CODES.VANILLA_ICE_CREAM)
+public class PerfettoProfiler {
+
+  // Bundle keys matching ProfilingManager constants
+  private static final String KEY_DURATION_MS = "KEY_DURATION_MS";
+  private static final String KEY_FREQUENCY_HZ = "KEY_FREQUENCY_HZ";
+
+  /** Fixed sampling frequency for Perfetto stack sampling. Not configurable by the developer. */
+  private static final int PROFILING_FREQUENCY_HZ = 100;
```
romtsn: should this be 101 or there's no lockstep sampling in perfetto?

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
```diff
@@ -335,16 +337,39 @@ private static void setupProfiler(
...
+        final @NotNull SentryFrameMetricsCollector frameMetricsCollector =
+            Objects.requireNonNull(
+                options.getFrameMetricsCollector(), "options.getFrameMetricsCollector is required");
+        if (options.isUseProfilingManager()) {
+          if (buildInfoProvider.getSdkInfoVersion() >= Build.VERSION_CODES.VANILLA_ICE_CREAM) {
+            options.setContinuousProfiler(
+                new PerfettoContinuousProfiler(
+                    options.getLogger(),
+                    frameMetricsCollector,
+                    () -> options.getExecutorService(),
```
romtsn: `l`: I know we hadn't done this for the legacy profiler, but wondering if we should use a separate executor for profiling to not get the main one clogged? similar to replay

romtsn (approved):
A couple of low-to-medium comments, but nothing blocking really (except the executor service submit on the same executor service thread, which can result in an endless loop I think).

The biggest question to me if we should get rid of the flag altogether and just use ProfilingManager by default on API 35+? Or we could keep it but make it `true` by default, just to have a backdoor to disable it, if something goes terribly wrong.

sentry/src/main/java/io/sentry/SentryOptions.java
markushi: Hmmm if we would do that people would still see crashes - I guess we'd then have to add another "enableLegacyProfiling" bool. I guess it should be best to default this to "false" then.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java
markushi: As it's mainly used for callbacks and at a very low frequency, sticking with the existing executor for now

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
markushi: Yes, that decision was made during discussions with the relay time. Keeping as-is.

romtsn:
## Decision: Option 4 — `enableLegacyProfiling` flag, ProfilingManager always-on for API 35+

After discussion in the SDK team (referencing telemetry context: ~49.5% of Android events are already from Android 15+ / API 35+, and the top profiling customer by volume is fully on the native Android SDK), we've aligned on **option 4**:

- **ProfilingManager** is the default profiler on API 35+ — no opt-in required.
- A new **`enableLegacyProfiling`** option (defaults to `true`) allows customers to keep legacy profiling on API < 35 during this transitional period.
- **`enableLegacyProfiling` is dropped in the next major** — at that point legacy profiling is gone entirely and ProfilingManager is the only path.

This avoids forcing an opt-in for the new (better) path, preserves continuity for sub-API-35 devices in the interim, and gives a clean deprecation timeline. Mixing profilers (options 2 or 3) was ruled out due to crash risk and sampling rate complexity.

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
```diff
@@ -0,0 +1,637 @@
...
+                    }
+                  },
+                  MAX_CHUNK_DURATION_MILLIS);
+    } catch (RejectedExecutionException e) {
+      logger.log(
+          SentryLevel.ERROR,
+          "Failed to schedule profiling chunk finish. Did you call Sentry.close()?",
+          e);
+      shouldStop = true;
+    }
```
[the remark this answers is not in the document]
romtsn: Not a real issue. When start() returns false, the profiler either:
  1. Was already started (no-op, started = true but no requestProfiling call happened a second time)
  2. profilingManager is null — nothing was ever started, nothing to clean up
  3. requestProfiling threw — the request failed, no profiling session is active

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
```diff
@@ -0,0 +1,637 @@
...
+
+    perfettoProfiler = perfettoProfilerFactory.get();
+    if (perfettoProfiler == null) {
+      return;
+    }
+    if (!perfettoProfiler.start(MAX_CHUNK_DURATION_MILLIS)) {
+      logger.log(
+          SentryLevel.ERROR,
+          "Failed to start Perfetto profiling. PerfettoProfiler.start() returned false.");
+      return;
```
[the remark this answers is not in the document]
romtsn: https://github.com/getsentry/sentry-java/pull/5251#discussion_r3490599136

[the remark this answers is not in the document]
romtsn: https://github.com/getsentry/sentry-java/pull/5251#discussion_r3490599136

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoProfiler.java
```diff
@@ -0,0 +1,222 @@
...
+
+  public boolean start(final long durationMs) {
+    if (started) {
+      logger.log(SentryLevel.WARNING, "PerfettoProfiler was already started.");
+      return false;
+    }
+    started = true;
+
+    if (profilingManager == null) {
+      logger.log(SentryLevel.WARNING, "ProfilingManager is not available.");
```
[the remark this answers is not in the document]
romtsn: https://github.com/getsentry/sentry-java/pull/5251#discussion_r3490599136

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
```diff
@@ -0,0 +1,637 @@
...
+            return;
+          }
+          shouldStop = true;
+          break;
+        case MANUAL:
+          shouldStop = true;
+          break;
+      }
+    }
+  }
```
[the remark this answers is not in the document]
romtsn: this is intentional to let it actually finish in background

sentry/src/main/java/io/sentry/ProfileChunk.java
```diff
@@ -22,6 +22,7 @@
 public final class ProfileChunk implements JsonUnknown, JsonSerializable {
   public static final String PLATFORM_ANDROID = "android";
   public static final String PLATFORM_JAVA = "java";
+  public static final String CONTENT_TYPE_PERFETTO = "application/x-perfetto-trace";
```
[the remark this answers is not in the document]
markushi: nope, header attribute != profile chunk attribute

sentry/src/main/java/io/sentry/SentryEnvelopeItem.java
```diff
@@ -385,6 +386,79 @@ private static void ensureAttachmentSizeLimit(
...
+            null,
+            profileChunk.getPlatform(),
+            null,
+            // avoid method refs on Android due to some issues with older AGP setups
+            // noinspection Convert2MethodRef
+            () -> metaLength.get());
+
+    // avoid method refs on Android due to some issues with older AGP setups
+    // noinspection Convert2MethodRef
+    return new SentryEnvelopeItem(itemHeader, () -> cachedItem.getBytes());
```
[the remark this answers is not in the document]
markushi: Nope if get length fails, the whole profile is likely faulty too already

0xadam-brown:
🎉 💯 🙌

PR #5280, opened 2026-04-10
fix(spring-jakarta): [Queue Instrumentation 10] Fork root scopes and skip when OTel is active in Kafka consumer

adinauer:
## PR Stack (Queue Instrumentation)

- #5249
- #5250
- #5253
- #5254
- #5255
- #5259
- #5260
- #5265
- #5274
- #5279
- #5280
- #5281
- #5282
- #5283
- #5288
- #5289
- #5291
- #5312
- #5313
- #5314
- #5315
- #5316
- #5318
- #5319
- #5320
- #5321
- #5322
- #5323
- #5324
- #5325
- #5327
- #5328
- #5330
- #5337
- #5338
- #5341
- #5343
- #5345
- #5347
- #5348
- #5352
- #5368

---

## :scroll: Description

Two changes to `SentryKafkaRecordInterceptor`:

1. **Fork root scopes** — Use `Sentry.forkedRootScopes()` instead of `scopes.forkedScopes()`. Each Kafka message is an independent request boundary, so it should start with a clean scope from root rather than inheriting whatever scope happens to be on the consumer thread. This matches the pattern used by `SentryWebFilter` for reactive request boundaries.

2. **Skip when OTel is active** — Add `isIgnored()` check using `SpanUtils.isIgnored()` on the trace origin. When OTel is active and the origin is in the ignored span origins list, the interceptor skips instrumentation entirely (no fork, no transaction, no ThreadLocal). This is consistent with how `SentryTracingFilter` handles OTel coexistence.

## :bulb: Motivation and Context

- Forking from current scopes on a Kafka consumer thread could inherit stale scope state from a previous message or unrelated context. Root scopes provide proper isolation per message.
- The OTel dedup via ignored origins ensures Sentry's Kafka instrumentation doesn't create duplicate spans when OTel is also instrumenting Kafka consumers.

## :green_heart: How did you test it?

- Updated existing tests to use `Mockito.mockStatic(Sentry::class.java)` for `forkedRootScopes`
- Added test: `does not create span when origin is ignored`
- All 13 tests pass

## :pencil: Checklist
- [ ] I added GH Issue ID _&_ Linear ID
- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [ ] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
[...]

PR #5279, opened 2026-04-10
fix(spring-jakarta): [Queue Instrumentation 9] Clean up stale ThreadLocal context in Kafka consumer interceptor

adinauer:
## PR Stack (Queue Instrumentation)

- #5249
- #5250
- #5253
- #5254
- #5255
- #5259
- #5260
- #5265
- #5274
- #5279
- #5280
- #5281
- #5282
- #5283
- #5288
- #5289
- #5291
- #5312
- #5313
- #5314
- #5315
- #5316
- #5318
- #5319
- #5320
- #5321
- #5322
- #5323
- #5324
- #5325
- #5327
- #5328
- #5330
- #5337
- #5338
- #5341
- #5343
- #5345
- #5347
- #5348
- #5352
- #5368

---

## :scroll: Description

Implement `clearThreadState()` and defensive cleanup in `intercept()` on `SentryKafkaRecordInterceptor` to prevent ThreadLocal leaks of `SentryRecordContext` (lifecycle token + transaction).

Spring Kafka's `RecordInterceptor` extends `ThreadStateProcessor`, which provides `clearThreadState()` — called in the poll loop's `finally` block, making it the most reliable cleanup hook. This catches edge cases where `success()`/`failure()` callbacks are skipped (e.g. `Error` thrown by listener).

Also adds defensive cleanup at the start of `intercept()` to handle any stale context from a previous record that was not properly cleaned up.

## :bulb: Motivation and Context

Review finding F-001 identified that `SentryRecordContext` stored in ThreadLocal is only cleaned in `success()`/`failure()`. While the realistic leak surface is narrow for synchronous listeners (Spring Kafka reliably calls these callbacks in most scenarios), implementing `clearThreadState()` is a trivially cheap safety net that Spring already calls on our instance.

## :green_heart: How did you test it?

- Added unit test: `clearThreadState cleans up stale context`
- Added unit test: `clearThreadState is no-op when no context exists`
- Added unit test: `intercept cleans up stale context from previous record`
- All existing tests continue to pass

## :pencil: Checklist
- [ ] I added GH Issue ID _&_ Linear ID
- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [ ] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

[...]

PR #5274, opened 2026-04-07
fix(spring-boot-jakarta): [Queue Instrumentation 8] Disable Sentry Kafka spans when OTel is active

adinauer:
## PR Stack (Queue Instrumentation)

- #5249
- #5250
- #5253
- #5254
- #5255
- #5259
- #5260
- #5265
- #5274
- #5279
- #5280
- #5281
- #5282
- #5283
- #5288
- #5289
- #5291
- #5312
- #5313
- #5314
- #5315
- #5316
- #5318
- #5319
- #5320
- #5321
- #5322
- #5323
- #5324
- #5325
- #5327
- #5328
- #5330
- #5337
- #5338
- #5341
- #5343
- #5345
- #5347
- #5348
- #5352
- #5368

---

## :scroll: Description

Disable Sentry's Kafka queue instrumentation when a Sentry OpenTelemetry integration (agent or agentless) is active, to avoid duplicate spans:

1. **Auto-configuration guard:** Add `@ConditionalOnMissingClass("io.sentry.opentelemetry.SentryAutoConfigurationCustomizerProvider")` to `SentryKafkaQueueConfiguration` so the Kafka bean post processors are not registered when OTel is present.
2. **Ignored span origins:** Add `auto.queue.spring_jakarta.kafka.producer` and `auto.queue.spring_jakarta.kafka.consumer` to `SpanUtils.ignoredSpanOriginsForOpenTelemetry()` for AGENT and AGENTLESS_SPRING modes as a second layer of protection.

## :bulb: Motivation and Context

OpenTelemetry provides its own Kafka instrumentation. When both Sentry and OTel instrument Kafka, duplicate spans are created. This disables the Sentry side when OTel handles it.

## :green_heart: How did you test it?

Unit tests in `SentryKafkaAutoConfigurationTest` — added a test verifying beans are not registered when OTel is present, and updated existing tests to use `FilteredClassLoader` to simulate OTel absence.

## :pencil: Checklist

- [ ] I added GH Issue ID _&_ Linear ID
- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [ ] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

## :crystal_ball: Next steps

#skip-changelog

[...]

PR #5265, opened 2026-04-07
feat(samples): [Queue Instrumentation 7] Add Kafka to Spring Boot 3 OTel sample apps

adinauer:
## PR Stack (Queue Instrumentation)

- #5249
- #5250
- #5253
- #5254
- #5255
- #5259
- #5260
- #5265
- #5274
- #5279
- #5280
- #5281
- #5282
- #5283
- #5288
- #5289
- #5291
- #5312
- #5313
- #5314
- #5315
- #5316
- #5318
- #5319
- #5320
- #5321
- #5322
- #5323
- #5324
- #5325
- #5327
- #5328
- #5330
- #5337
- #5338
- #5341
- #5343
- #5345
- #5347
- #5348
- #5352
- #5368

---

## :scroll: Description

Add Kafka queue tracing support to both OpenTelemetry Spring Boot 3 sample applications:

- **`sentry-samples-spring-boot-jakarta-opentelemetry`** (OTel Java agent)
- **`sentry-samples-spring-boot-jakarta-opentelemetry-noagent`** (agentless / Sentry OTel Spring)

Each sample gets:
- `KafkaController` — REST endpoint (`/kafka/produce`) that sends messages to a Kafka topic
- `KafkaConsumer` — `@KafkaListener` that receives and logs messages
- `application-kafka.properties` — Kafka configuration activated via `--spring.profiles.active=kafka`
- Kafka auto-configuration excluded by default to avoid startup errors when no broker is available

This mirrors the existing Kafka sample support added to the non-OTel Spring Boot 3 sample in an earlier stack PR.

## :bulb: Motivation and Context

The queue instrumentation stack adds Kafka tracing to the Spring Boot 3 integration. The OTel sample apps need the same Kafka producer/consumer setup so that queue tracing can be verified end-to-end in the OpenTelemetry context (both agent and agentless modes).

## :green_heart: How did you test it?

Ran both sample apps locally with `--spring.profiles.active=kafka` against a local Kafka broker and verified messages are produced and consumed with queue spans visible in Sentry.

## :pencil: Checklist

- [ ] I added GH Issue ID _&_ Linear ID
- [ ] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [ ] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
[...]

PR #5281, opened 2026-04-10
fix(spring-jakarta): [Queue Instrumentation 11] Guard entire span lifecycle in Kafka producer interceptor

adinauer:

- #5249
- #5250
- #5253
- #5254
- #5255
- #5259
- #5260
- #5265
- #5274
- #5279
- #5280
- #5281
- #5282
- #5283
- #5288
- #5289
- #5291
- #5312
- #5313
- #5314
- #5315
- #5316
- #5318
- #5319
- #5320
- #5321
- #5322
- #5323
- #5324
- #5325
- #5327
- #5328
- #5330
- #5337
- #5338
- #5341
- #5343
- #5345
- #5347
- #5348
- #5352
- #5368

---

## :scroll: Description

Wrap the entire span lifecycle in `SentryProducerInterceptor.onSend()` in a single try-catch so instrumentation can never break the customer's Kafka send. The `ProducerRecord` is always returned regardless of any exception in Sentry code.

Previously, only header injection was guarded. Span creation (`startChild`), `setData()`, and `finish()` were unguarded — any exception from these would propagate and fail the customer's Kafka send per the `ProducerInterceptor.onSend()` contract.

The inner try-catch around `injectHeaders` is removed since the outer catch now covers it.

Kafka's `ProducerInterceptor.onSend()` contract requires that interceptors not throw exceptions — any exception propagates and fails the send. Observability code must never cause customer message loss.

- All 8 existing `SentryProducerInterceptorTest` tests pass
- The change is purely defensive — wraps existing logic in try-catch without changing behavior

## :pencil: Checklist

## :crystal_ball: Next steps

Nothing.

#skip-changelog

### DIFF
--- a/sentry-spring-jakarta/src/main/java/io/sentry/spring/jakarta/kafka/SentryProducerInterceptor.java
+++ b/sentry-spring-jakarta/src/main/java/io/sentry/spring/jakarta/kafka/SentryProducerInterceptor.java
@@ -53,24 +53,25 @@ public final class SentryProducerInterceptor<K, V> implements ProducerIntercepto
       return record;
     }
 
-    final @NotNull SpanOptions spanOptions = new SpanOptions();
-    spanOptions.setOrigin(TRACE_ORIGIN);
-    final @NotNull ISpan span = activeSpan.startChild("queue.publish", record.topic(), spanOptions);
-    if (span.isNoOp()) {
-      return record;
-    }
-
-    span.setData(SpanDataConvention.MESSAGING_SYSTEM, "kafka");
-    span.setData(SpanDataConvention.MESSAGING_DESTINATION_NAME, record.topic());
-
     try {
-      injectHeaders(record.headers(), span);
-    } catch (Throwable ignored) {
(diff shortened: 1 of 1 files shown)


sentry-spring-jakarta/src/main/java/io/sentry/spring/jakarta/kafka/SentryProducerInterceptor.java
```diff
@@ -53,25 +53,26 @@ public SentryProducerInterceptor(final @NotNull IScopes scopes) {
...
+      span.setData(SpanDataConvention.MESSAGING_DESTINATION_NAME, record.topic());
 
-    try {
       injectHeaders(record.headers(), span);
+
+      span.setStatus(SpanStatus.OK);
+      span.finish();
     } catch (Throwable ignored) {
-      // Header injection must not break the send
+      // Instrumentation must never break the customer's Kafka send
```
lbloder, line 73: Same as in the original, should we log anything here?

PR #5284, opened 2026-04-10
fix(profiling): Reset shouldStop flag on startProfiler

43jay:
## :scroll: Description

`shouldStop` is never reset to `false` after `stopProfiler()`, so a stop/start cycle leaves `shouldStop=true`. When the next 60-second chunk timer fires, `stop(restartProfiler=true)` checks `!shouldStop` and does not restart — the profiler silently stops after one chunk instead of continuing indefinitely.

This affects both lifecycle modes:
- **MANUAL**: `Sentry.startProfiler()` → `Sentry.stopProfiler()` → `Sentry.startProfiler()` — second session stops after one chunk
- **TRACE**: all transactions finish (rootSpanCounter reaches 0, sets `shouldStop=true`) → new transaction starts → profiler starts but stops after one chunk

Fix: reset `shouldStop = false` at the top of `startProfiler()`, since it represents a new intent to profile.

Discovered while working on the Perfetto ProfilingManager integration (#5251). The same bug existed in both `AndroidContinuousProfiler` and `PerfettoContinuousProfiler`.

Unit test added: `manual profiler can be started again after a full start-stop cycle` — starts profiling, stops it, starts again, verifies the profiler continues running after the first chunk restart.

```
JAVA_HOME=$(/usr/libexec/java_home -v 17) ./gradlew :sentry-android-core:testDebugUnitTest --tests "io.sentry.android.core.AndroidContinuousProfilerTest.manual profiler can be started again after a full start-stop cycle"
```

## :pencil: Checklist

## :crystal_ball: Next steps

None — standalone bugfix.

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
@@ -146,6 +146,7 @@ public class AndroidContinuousProfiler
           break;
       }
       if (!isRunning()) {
+        shouldStop = false;
         logger.log(SentryLevel.DEBUG, "Started Profiler.");
         start();
       }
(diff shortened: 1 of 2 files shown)


romtsn (approved):
nice catch! please, add a changelog entry as well

PR #5293, opened 2026-04-14
fix(sentry): Recover object readers after deserialization errors

### LINKED ISSUE #5278: JsonObjectReader can get stuck when deserializer throws
### Integration

sentry

### Java Version

21

No

### Other Error Monitoring Solution Name

_No response_

### Version

8.38.0

### Steps to Reproduce

`JsonReader.nextName` can throw exceptions, I think this causes `nextMapOrNull` to spin forever, since the token isn't consumed. While I'm not sure what caused the "bad" input this all seems to start in `SentryEnvelopeItem.getLogs`.

What is presumably the same path can be reproduced with this snippet:
```java
final var reader = new JsonObjectReader(new StringReader("{\"x\": {}}"));
reader.nextMapOrNull(NoOpLogger.getInstance(), (_, _) -> {
  throw new IllegalStateException();
});
```
I'm using `nextMapOrNull` here, since that's where we saw this happen, but this probably happens for arrays as well.

### Expected Result

`nextMapOrNull` should, at some point, return.

### Actual Result

`nextMapOrNull` **never** returns, instead it spins forever.

adinauer:
## :scroll: Description
Fix `JsonObjectReader` and `MapObjectReader` so failed value deserialization no longer hangs
or corrupts subsequent reads.

The change makes both readers advance past failed values, keeps later valid entries readable,
and adds coverage for partial-consumption and truncated-input recovery paths.

Some collection deserialization helpers could get stuck retrying the same value forever when a
nested deserializer threw. In other cases, partial consumption left the reader in an invalid
state and broke later values.

This fixes GH-5278.

- Ran `./gradlew :sentry:test --tests="io.sentry.JsonObjectReaderTest" --tests="io.sentry.util.MapObjectReaderTest"`
- Added regression tests for skipped failed values, partially consumed nested values, and
  unrecoverable truncated-input recovery

## :pencil: Checklist

## :crystal_ball: Next steps
- Verify the changelog wording before merge


adinauer:
@sentry review

adinauer:
cursor review

sentry/src/main/java/io/sentry/JsonObjectReader.java
```diff
@@ -278,11 +349,80 @@ public void setLenient(boolean lenient) {
 
   @Override
   public void skipValue() throws IOException {
+    markValueConsumed();
```
romtsn: is this on purpose before `jsonReader.skipValue()`? just looking at the above and the pattern there to call this _after_ jsonReader's work.

romtsn (approved):
just one question, but LGTM otherwise!

sentry/src/main/java/io/sentry/JsonObjectReader.java
adinauer: updated

PR #5333, opened 2026-04-23
feat(replay): Capture SurfaceView content (experimental)

### LINKED ISSUE #4862: Capture maps (google, mapbox, etc.) in Session Replay
### Description

There's a discord thread about mapbox not working https://www.npmjs.com/package/@rnmapbox/maps. That's probably due to them doing rendering in native directly to opengl. My impression was that it's possible to capture this still, because I have seen google maps in some replays. But we'd have to investigate mapbox.

### LINKED ISSUE #5257: Session Replay: support Unity (SurfaceView/OpenGL) rendering
## Summary

Session Replay on Android shows a black screen when the app uses Unity for rendering. Unity renders via a `SurfaceView` backed by OpenGL in a separate process. The current `PixelCopy`-based screenshot mechanism does not capture `SurfaceView` content, resulting in a fully black frame. This blocks replay utility for apps that embed Unity scenes (e.g. for game-like or interactive lesson surfaces).

## Current behavior

* Android replay capture uses `PixelCopy` to screenshot the window hierarchy.
* `PixelCopy` does not capture `SurfaceView` contents because `SurfaceView` renders to a separate hardware-composited layer.
* Unity games/scenes embedded in an Android app use `SurfaceView` + OpenGL, so they appear black in any captured replay frame.
* Attempting capture via `screenshotty` (an alternative screenshot library) produces the same result.
* The Sentry SDK is not initialized inside the Unity process; the Java/Android SDK captures the host app side only.

## Gap

Any Android app that embeds Unity content — increasingly common for interactive or game-adjacent product surfaces — gets unusable replays. The black region can cover most or all of the viewport, making the replay meaningless for debugging UX issues in Unity-rendered screens.

This is the same underlying constraint as [JAVA-221](https://linear.app/getsentry/issue/JAVA-221/capture-maps-google-mapbox-etc-in-session-replay) (maps rendered via native OpenGL), but Unity is a distinct integration path with higher prevalence in consumer apps.

## Options

1. **MediaProjection-based capture** — use Android's `MediaProjection` API to capture the composited screen output. This includes `SurfaceView` content and would solve Unity, maps, and other native-render cases. Requires a runtime permission prompt (`FOREGROUND_SERVICE_MEDIA_PROJECTION`) and is more invasive/complex to implement safely. Privacy implications need careful handling.
2. **VirtualDisplay + SurfaceView mirror** — create a `VirtualDisplay` that mirrors the target surface. Technically complex; not guaranteed to work across Unity's multi-process setup.
3. **Placeholder/redaction overlay** — when a `SurfaceView` is detected in the view hierarchy, render a labeled placeholder block in the replay frame (similar to how we mask sensitive content). This doesn't capture Unity content but makes replays honest about what's missing instead of just showing black.
4. **SDK-side Unity integration** — coordinate with the Unity SDK (`getsentry/sentry-unity`) to have Unity emit its own frame snapshots or DOM-equivalent events that can be stitched into the replay. High effort; requires cross-SDK protocol work.

## Recommendation

Option 3 (placeholder overlay) is the lowest-risk near-term improvement — it converts a confusing black-screen replay into an honest, annotated gap. Option 1 (MediaProjection) is the right long-term fix for full capture fidelity, but carries permission/privacy tradeoffs that need product alignment before implementing.

Related: [JAVA-221](https://linear.app/getsentry/issue/JAVA-221/capture-maps-google-mapbox-etc-in-session-replay) (maps / native OpenGL rendering same root cause)

Action taken on behalf of Simon Zhong.

romtsn:
## :scroll: Description

Adds an experimental option `SentryReplayOptions.isCaptureSurfaceViews` (default `false`). When enabled, `PixelCopyStrategy` captures each visible `SurfaceView` on screen via `PixelCopy.request(surfaceView, ...)` and composites the result onto the screenshot using `PorterDuff.DST_OVER`, so `SurfaceView` content no longer renders as a transparent hole in the replay.

Since `SurfaceView` redraws do not trigger `ViewTreeObserver.OnDrawListener`, `ScreenshotRecorder` also bypasses the `contentChanged` guard when `SurfaceView`s are present, so frames are re-captured at the configured frame rate instead of reusing the previous screenshot.

Example replay: https://sentry-sdks.sentry.io/explore/replays/0c28984d7c24464aa204495b0d95c53b/

Additionally it also resolves the issue with capturing maps that we've had (#4862), verified with Mapbox: https://sentry-sdks.sentry.io/explore/replays/a52feeea52a84aa0ab85f9fd9fcc706c/?playlistEnd=2026-04-24T10%3A36%3A08&playlistStart=2026-03-25T10%3A36%3A08&project=5428559&query=&referrer=%2Fexplore%2Freplays%2F%3AreplaySlug%2F&t=32

`SurfaceView` (used by Unity, ExoPlayer, maps, camera previews, etc.) renders to its own `Surface` that is composited by SurfaceFlinger outside the View hierarchy. `PixelCopy.request(window, ...)` only captures the Window surface, so anything drawn on a `SurfaceView` was missing from Session Replay — a common complaint for apps embedding Unity or other 3D/video content.

The feature is gated behind an experimental flag and defaults to `false` to preserve existing behavior and avoid the extra per-frame cost (one additional PixelCopy per visible SurfaceView + a compositing pass) for apps that don't need it.

Closes #5257
Closes #4862

Manually verified on a Unity-as-a-Library Android sample app: with the flag off, the Unity area is transparent/black in the replay (unchanged); with the flag on, the Unity frames appear in the replay video and animate at the configured frame rate. Also verified that non-SurfaceView screens behave identically to before. See example replay linked above.

## :pencil: Checklist


sentry-android-replay/src/main/java/io/sentry/android/replay/screenshot/PixelCopyStrategy.kt
```diff
@@ -115,6 +129,148 @@ internal class PixelCopyStrategy(
...
+            tmpDstRect,
+            capture.bitmap,
+            capture.x,
+            capture.y,
+            windowX,
+            windowY,
+            config.scaleFactorX,
+            config.scaleFactorY,
+          )
+          capture.bitmap.recycle()
```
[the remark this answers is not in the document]
romtsn: filed https://github.com/getsentry/sentry-java/issues/5340 to follow up on this separately

sentry-android-replay/src/main/java/io/sentry/android/replay/screenshot/PixelCopyStrategy.kt
```diff
@@ -115,6 +129,148 @@ internal class PixelCopyStrategy(
...
+      // holder.surface can be null before the surface is created — guard against NPE.
+      val surface = surfaceView?.holder?.surface
+      if (surfaceView == null || surface == null || !surface.isValid) {
+        onCaptureComplete()
+        continue
+      }
+
+      var svBitmap: Bitmap? = null
+      try {
+        svBitmap =
```
markushi, line 186: I'm wondering if we could have a cached Bitmap here, which matches the screens size.
`Pixelcopy.request` always seems to scale the surface to the bitmap size, and then we would scale it  make when drawing onto the final bitmap. 

On the other hand, for traditional use cases there will be most likely 1, max. 2 surfaces per screen I guess.

markushi (approved):
Great stuff!

romtsn:
tested this thoroughly in the sample app with both canvas and pixel copy.

PixelCopy: https://sentry-sdks.sentry.io/explore/replays/6483de83680841a8a5237817b9a743e6/

Canvas: https://sentry-sdks.sentry.io/explore/replays/1ec81acba14b41f88e0c622931093f1c

everything looks as before, so it's good to merge.

sentry-android-replay/src/main/java/io/sentry/android/replay/screenshot/PixelCopyStrategy.kt
romtsn: yeah, the cost is not that high, but I agree would be better to cache those :) however, we still need to land #5340 first, otherwise the cached bitmap would potentially be corrupt  by multiple threads writing. I'll add a comment there to follow-up on this too, thanks for pointing out 👍

PR #5337, opened 2026-04-27
fix(otel): [Queue Instrumentation 34] Fix OTel messaging span mapping correctness

adinauer:
## PR Stack (Queue Instrumentation)

- #5249
- #5250
- #5253
- #5254
- #5255
- #5259
- #5260
- #5265
- #5274
- #5279
- #5280
- #5281
- #5282
- #5283
- #5288
- #5289
- #5291
- #5312
- #5313
- #5314
- #5315
- #5316
- #5318
- #5319
- #5320
- #5321
- #5322
- #5323
- #5324
- #5325
- #5327
- #5328
- #5330
- #5337
- #5338
- #5341
- #5343
- #5345
- #5347
- #5348
- #5352
- #5368

---

## :scroll: Description

Follow-up fixes to the OTel messaging span mapping introduced in [Queue Instrumentation 33]:

- **Gate ordering** — when `enableQueueTracing` is on, check `messaging.system` before `http.request.method` so OTel AWS SDK SQS spans (which carry both attributes) are classified as queue spans, not HTTP spans.
- **`create` mapping** — map the `"create"` operation type to `queue.create` instead of `queue.publish`, matching the spec.
- **`send` mapping coverage** — add explicit test coverage for `"send"` → `queue.publish`.
- **`deliver` mapping coverage** — add explicit test coverage for `"deliver"` → `queue.process`.
- **`settle` mapping** — add explicit `"settle"` → `queue.settle` arm.
- **Javadoc** — clarify `enableQueueTracing` getter/setter docs.
- **Sample cleanup** — remove verbose explanatory comment block from sample `application-kafka.properties`.

## :bulb: Motivation and Context

Review of Queue Instrumentation 33 surfaced several correctness and documentation issues (F002, F004, F007, F008 from internal review). This PR addresses them as a batch of small targeted fixes.

## :green_heart: How did you test it?

- Added unit tests for gate ordering (messaging wins over HTTP when flag on; HTTP wins when flag off)
- Added unit tests for `create` → `queue.create`, `send` → `queue.publish`, `deliver` → `queue.process`, and `settle` → `queue.settle` mappings
- Ran `./gradlew spotlessApply apiDump`
- Ran `./gradlew :sentry-opentelemetry:sentry-opentelemetry-core:test --tests "SpanDescriptionExtractorTest"`
- All 30 `SpanDescriptionExtractorTest` tests pass

## :pencil: Checklist
- [ ] I added GH Issue ID _&_ Linear ID
[...]

PR #5330, opened 2026-04-23
test(spring-boot-jakarta): [Queue Instrumentation 32] Filter OTel in Kafka auto-config negative tests

adinauer:
## PR Stack (Queue Instrumentation)

- #5249
- #5250
- #5253
- #5254
- #5255
- #5259
- #5260
- #5265
- #5274
- #5279
- #5280
- #5281
- #5282
- #5283
- #5288
- #5289
- #5291
- #5312
- #5313
- #5314
- #5315
- #5316
- #5318
- #5319
- #5320
- #5321
- #5322
- #5323
- #5324
- #5325
- #5327
- #5328
- #5330
- #5337
- #5338
- #5341
- #5343
- #5345
- #5347
- #5348
- #5352
- #5368

---

## :scroll: Description

Fix the Kafka auto-config negative regression tests so they validate the `@ConditionalOnClass` gate they claim to guard.

`SentryKafkaQueueConfiguration` is gated by three annotations:

1. `@ConditionalOnClass({KafkaTemplate, SentryKafkaProducerInterceptor})`
2. `@ConditionalOnProperty("sentry.enable-queue-tracing" = "true")`
3. `@ConditionalOnMissingClass("io.sentry.opentelemetry.SentryAutoConfigurationCustomizerProvider")`

The regression tests `does not register Kafka BPPs when sentry-kafka is not present` and `...when spring-kafka is not present` used `FilteredClassLoader` to hide `SentryKafkaProducerInterceptor` / `KafkaTemplate` respectively, asserting the BPPs are absent. But `sentry-opentelemetry-agentless-spring` is on `testImplementation`, so `SentryAutoConfigurationCustomizerProvider` is always on the test classpath — gate #3 was already blocking the config independent of gate #1. The tests passed for the wrong reason: if someone removed a class from `@ConditionalOnClass`, they would still pass.

Extend both `FilteredClassLoader`s to additionally filter `SentryAutoConfigurationCustomizerProvider`, so gate #3 is satisfied and only the gate under test can be the blocker. With this change, temporarily removing `"io.sentry.kafka.SentryKafkaProducerInterceptor"` from `@ConditionalOnClass` causes the `sentry-kafka not present` test to correctly fail — confirmed locally before restoring the gate.

Addresses [review8.md](https://github.com/getsentry/sentry-java/pull/5291#discussion_r3129873509) finding **R8-F005**.

## :bulb: Motivation and Context

[...]

PR #5328, opened 2026-04-23
fix(kafka): [Queue Instrumentation 31] Write enqueued-time header as plain decimal

adinauer:
## PR Stack (Queue Instrumentation)

- #5249
- #5250
- #5253
- #5254
- #5255
- #5259
- #5260
- #5265
- #5274
- #5279
- #5280
- #5281
- #5282
- #5283
- #5288
- #5289
- #5291
- #5312
- #5313
- #5314
- #5315
- #5316
- #5318
- #5319
- #5320
- #5321
- #5322
- #5323
- #5324
- #5325
- #5327
- #5328
- #5330
- #5337
- #5338
- #5341
- #5343
- #5345
- #5347
- #5348
- #5352
- #5368

---

## :scroll: Description

Fix the `sentry-task-enqueued-time` Kafka header being serialized in scientific notation (e.g. `1.776933649613E9`). The producer interceptor was writing the value via `String.valueOf(double)`, which switches to scientific notation once `|d| >= 10^7` — and epoch seconds are always `~1.77 × 10^9`.

Route the value through `DateUtils.doubleToBigDecimal(...).toString()` — the same helper already used to serialize epoch-seconds timestamps in `SentryTransaction`, `SentrySpan`, `SentryLogEvent`, `SentryMetricsEvent`, `ProfileChunk`, etc. At the pinned scale of 6, `BigDecimal.toString()` produces plain decimal form (`1776938295.692000`) for every realistic epoch-seconds magnitude.

Addresses [review7.md](https://github.com/getsentry/sentry-java/pull/5283#discussion_r3122702573) finding **R7-F017** — the only remaining open review comment across the Queue Instrumentation stack (1–30).

## :bulb: Motivation and Context

PR #5283 ("Align enqueue time with Python") was meant to make the Java Kafka queue instrumentation interoperable with other Sentry SDKs by switching the header payload to epoch seconds. However, the format it actually shipped — Java's default `double` `toString` — uses scientific notation for all realistic timestamps, which:

- Python's `sentry_sdk` (and Ruby, PHP, .NET) Celery consumers parse the header with plain-decimal expectations, so they would fail to parse Java-produced headers.
- The Java consumer (`SentryKafkaConsumerTracing.receiveLatency`) uses `Double.parseDouble`, which accepts both plain and scientific forms — so internal round-trips pass and none of the existing tests caught the bug.

This PR closes the gap without a core API change by reusing the canonical `DateUtils.doubleToBigDecimal` helper.

[...]

PR #5318, opened 2026-04-21
test(spring-boot-jakarta): [Queue Instrumentation 22] Cover spring-kafka class-absence gate

adinauer:
## PR Stack (Queue Instrumentation)

- #5249
- #5250
- #5253
- #5254
- #5255
- #5259
- #5260
- #5265
- #5274
- #5279
- #5280
- #5281
- #5282
- #5283
- #5288
- #5289
- #5291
- #5312
- #5313
- #5314
- #5315
- #5316
- #5318
- #5319
- #5320
- #5321
- #5322
- #5323
- #5324
- #5325
- #5327
- #5328
- #5330
- #5337
- #5338
- #5341
- #5343
- #5345
- #5347
- #5348
- #5352
- #5368

---

## :scroll: Description

`SentryKafkaQueueConfiguration` in `SentryAutoConfiguration` is gated on two classes being present on the classpath:

```java
@ConditionalOnClass(name = {
  "org.springframework.kafka.core.KafkaTemplate",
  "io.sentry.kafka.SentryKafkaProducerInterceptor"
})
```

Only the `SentryKafkaProducerInterceptor`-missing half of that gate had test coverage. This PR adds a `FilteredClassLoader(KafkaTemplate::class.java)` test that asserts neither `SentryKafkaProducerBeanPostProcessor` nor `SentryKafkaConsumerBeanPostProcessor` is registered when spring-kafka is missing, even with `sentry.enable-queue-tracing=true`.

## :bulb: Motivation and Context

Round out the class-absence gate coverage in `SentryKafkaAutoConfigurationTest` so the autoconfig contract for spring-kafka being absent is explicitly pinned by a test.

## :green_heart: How did you test it?

- `./gradlew :sentry-spring-boot-jakarta:test --tests='io.sentry.spring.boot.jakarta.SentryKafkaAutoConfigurationTest'` — 6/6 pass (incl. new `does not register Kafka BPPs when spring-kafka is not present`).
- `./gradlew spotlessApply apiDump` — clean.

## :pencil: Checklist

- [ ] I added GH Issue ID _&_ Linear ID
- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [x] I updated the docs if needed.
- [x] I updated the wizard if needed.
- [x] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

[...]

PR #5338, opened 2026-04-27
fix(kafka): [Queue Instrumentation 35] Inject trace headers without real active span

adinauer:

- #5249
- #5250
- #5253
- #5254
- #5255
- #5259
- #5260
- #5265
- #5274
- #5279
- #5280
- #5281
- #5282
- #5283
- #5288
- #5289
- #5291
- #5312
- #5313
- #5314
- #5315
- #5316
- #5318
- #5319
- #5320
- #5321
- #5322
- #5323
- #5324
- #5325
- #5327
- #5328
- #5330
- #5337
- #5338
- #5341
- #5343
- #5345
- #5347
- #5348
- #5352
- #5368

---

## :scroll: Description

Restructure `SentryKafkaProducer.send()` to inject trace headers even when there is no real active span, and simplify the overall flow to match the `SentryFeignClient` / `SentryOkHttpInterceptor` pattern.

**Before:** When `scopes.getSpan()` returned null (background workers, `@Scheduled` jobs, startup publishers), `send()` short-circuited to the bare delegate — no `sentry-trace`, `baggage`, or `sentry-task-enqueued-time` headers were injected. An active `NoOpSpan` follows the same propagation-only behavior. Distributed tracing silently broke for these common Kafka patterns.

**After:** Three clean branches matching the Feign/OkHttp/RestTemplate convention:
- `isIgnored` → pure delegate (no headers, no span)
- No real active span (`null` or `NoOpSpan`) → inject headers from `PropagationContext`, no publish span
- Real active span → start child span, inject headers, delegate with wrapped callback

Also simplifies the implementation:
- Rename `injectHeaders` → `maybeInjectHeaders` with encapsulated try/catch (matches Feign's `maybeAddTracingHeaders`)
- Remove outer try/catch around span setup
- Keep child no-op spans on the wrapped-callback path (finishing a no-op child span is harmless)
- Remove redundant `isFinished()` guards — `Span.finish()` is already idempotent via `compareAndSet`

Addresses review finding R10-F001: producer drops trace headers when no active span exists. Background workers and `@Scheduled` jobs are a primary Kafka producer pattern; losing distributed tracing for them contradicts the Queues product expectations.

- Updated existing unit tests to assert headers are injected when no active span exists
- Added matching coverage for active `NoOpSpan` propagation-only behavior
- Updated no-op child span test to verify callback wrapping still works
- Updated header-injection-failure test for new `maybeInjectHeaders` behavior
- Ran `./gradlew :sentry-kafka:test --tests "io.sentry.kafka.SentryKafkaProducerTest"`
- All 16 `SentryKafkaProducerTest` tests pass

## :pencil: Checklist

## :crystal_ball: Next steps

Continue addressing remaining review findings from the Queue Instrumentation review.


sentry-kafka/src/test/kotlin/io/sentry/kafka/SentryKafkaProducerTest.kt
```diff
@@ -0,0 +1,345 @@
...
+
+    producer.send(record)
+
+    assertEquals(0, tx.spans.size)
+    verify(delegate).send(eq(record), isNull())
+    assertEquals(null, record.headers().lastHeader(SentryTraceHeader.SENTRY_TRACE_HEADER))
+  }
+
+  @Test
+  fun `injects headers but creates no span when no active span`() {
```
lbloder, line 222: We could do the same test for `NoOpSpan`

sentry-spring-jakarta/src/main/java/io/sentry/spring/jakarta/kafka/SentryKafkaRecordInterceptor.java
lbloder, line 72: what happens if the the `delegate` RecordInterceptor changes the record? E.g. by adding a baggage header, that would mean we might continue the wrong trace. However, if we let the delegate run first and continue from that record the timing might be off, depending on how long the delegate takes.

sentry-kafka/src/test/kotlin/io/sentry/kafka/SentryKafkaProducerTest.kt
adinauer: added to this PR

sentry-spring-jakarta/src/main/java/io/sentry/spring/jakarta/kafka/SentryKafkaRecordInterceptor.java
adinauer: Yes, timing would be off. I'd say the chance of adding relevant headers in a consumer is very low so let's keep it as implemented.

lbloder (approved):

lbloder (approved):

PR #5342, opened 2026-04-28
feat(android): Add standalone app start tracing

### LINKED ISSUE #5046: Include app start spans for all Android app launches
### Description

Instead of only attaching app start spans to the first ui.load transaction, we should make sure any app launch has the duration captured as a span, regardless of whether or not the user is also using other ui.load transactions. The iOS equivalent of this feature is here: https://github.com/getsentry/sentry-cocoa/issues/6883

buenaflor:
## :scroll: Description

Adds standalone Android app-start transaction support behind the opt-in `enableStandaloneAppStartTracing` option and the `io.sentry.standalone-app-start-tracing.enable` manifest key.

This PR also covers the non-activity startup path so cold starts triggered by broadcasts or foreground services can emit an `App Start Cold/Warm` transaction without waiting for an activity. The legacy flag-off behavior is preserved: activity app-start data remains nested under the `ui.load` transaction.

Additionally this PR adds a new attribute:
- `app.vitals.start.screen`

Resolves: #5046

Standalone app-start traces make app-start performance visible independently from activity load transactions, and they let Android report app starts where no activity is launched. The implementation keeps the default behavior unchanged and adds tests for the new option, manifest metadata, transaction context trace reuse, app-start metrics, event processing, and activity lifecycle behavior.

## App-start creation paths

This PR preserves the existing flag-off behavior and adds a separate standalone app-start branch when `enableStandaloneAppStartTracing` is enabled.

| Scenario | How app start is created | Resulting trace shape |
|---|---|---|
| Activity launch, flag ON | `ActivityLifecycleIntegration` starts a standalone `App Start` transaction and the normal activity `ui.load` transaction on the same trace. | Two sibling transactions: `app.start` for app-start phases, and `ui.load` for TTID/TTFD. The `ui.load` transaction does not contain an `app.start.cold/warm` child. |
| Activity launch, flag OFF | Legacy activity lifecycle tracing creates app start as a child span of the activity transaction. | One `ui.load` transaction with nested `app.start.cold/warm`, plus app-start phase spans under that child. |
| Non-activity process start, flag ON, Gradle plugin instrumentation present | The Gradle plugin records `Application.onCreate` start/end through `AppStartMetrics.onApplicationCreate/PostCreate`. When no activity starts, `ActivityLifecycleIntegration` emits from that completed app-start span. | One standalone `App Start` transaction and no `ui.load`; includes `process.load` and `application.load` phase spans. |
| Non-activity process start, flag ON, no Gradle plugin on API 35+ | `AppStartMetrics` reads `ApplicationStartInfo` from `ActivityManager` to determine start type and use the platform `APPLICATION_ONCREATE` timestamp as the app-start end time. | One standalone `App Start` transaction and no `ui.load`; includes `process.load`, but not `application.load` because the SDK does not have plugin instrumentation for that phase. |
| Non-activity process start, flag ON, no Gradle plugin on API <35 | `ApplicationStartInfo` is unavailable, so `AppStartMetrics` defaults the no-activity process start to cold and uses the class-loaded timestamp fallback for the app-start end time. | One standalone `App Start` transaction and no `ui.load`; includes `process.load` only and remains classified as cold. |
| Non-activity process start followed by activity | The non-activity `App Start` transaction stores its trace ID for a later activity. | The later activity `ui.load` reuses the stored trace ID and does not create a second standalone app-start transaction. |
| Non-activity process start, flag OFF | The non-activity listener is not installed. | Broadcasts/services emit no app-start transaction. |

- Targeted unit coverage for the option, manifest metadata, app-start metric resolution, event processing, lifecycle integration, and transaction context trace reuse.
- Manual end-to-end harness runs from `scripts/test-standalone-app-start.sh`, summarized below.

### End-to-end Sentry verification

| # | What it tests | Locally emitted shape | Trace ID link |
|---|---|---|---|
| **1a** | Launcher activity cold start, standalone flag ON. Standalone `App Start` emits alongside sibling `MainActivity ui.load`; shared trace ID; `ui.load` has no `app.start.*` child. | `app.start` (`app_start_cold=1156ms`) + `activity.load` x2 + `process.load` + `application.load`; sibling `ui.load` (`time_to_initial_display=1156ms`) | [`52154e91...`](https://sentry-sdks.sentry.io/performance/trace/52154e914456477da4847313b0e2cb82/?project=5428559&statsPeriod=24h) |
| **1c** | Launcher activity cold start, standalone flag OFF. Legacy shape preserved: single `ui.load` with `app.start.cold` nested inside. | `ui.load` (`time_to_initial_display=1536ms`) -> nested `app.start.cold` (`app_start_cold=1537ms`) -> `process.load` + `activity.load` x2 | [`d3f0e049...`](https://sentry-sdks.sentry.io/performance/trace/d3f0e04926ef44518e399cfd3a8e2137/?project=5428559&statsPeriod=24h) |
| **2a** | Broadcast cold, tier-1 end-time resolution with Gradle-plugin timing simulation. | `app.start` (`app_start_cold=1408ms`) + `process.load` + `application.load`; no `ui.load` | [`81ae80f6...`](https://sentry-sdks.sentry.io/performance/trace/81ae80f661f94c48aac9eb4f841b64d5/?project=5428559&statsPeriod=24h) |
| **2b** | Broadcast cold, tier-2 end-time resolution via Android 15 `ApplicationStartInfo`. | `app.start` (`app_start_cold=525ms`) + `process.load` only; no `ui.load` | [`0dcd120e...`](https://sentry-sdks.sentry.io/performance/trace/0dcd120ec9c04f1a83f787e511233fc0/?project=5428559&statsPeriod=24h) |
| **2c** | Broadcast cold, tier-3 fallback on API 33 using `CLASS_LOADED_UPTIME_MS`. | `app.start` (`app_start_cold=614ms`) + `process.load` only; correctly classified Cold | [`ac8a9189...`](https://sentry-sdks.sentry.io/performance/trace/ac8a91892cf042c4a2f492d1ab87cec5/?project=5428559&statsPeriod=24h) |
| **2d** | Foreground service cold start through the non-activity path. | `app.start` (`app_start_cold=4605ms`) + `process.load` + `application.load`; no `ui.load` | [`905262ce...`](https://sentry-sdks.sentry.io/performance/trace/905262ce5ab6442d972c9440953ef0e9/?project=5428559&statsPeriod=24h) |
| **2e** | Broadcast then launcher: trace reuse without duplicate standalone transaction. | Same trace: broadcast `app.start` (`app_start_cold=1751ms`) + later sibling `ui.load` (`time_to_initial_display=4563ms`); no duplicate standalone | [`5da4a88f...`](https://sentry-sdks.sentry.io/performance/trace/5da4a88f12314350afbfe5c028097bc4/?project=5428559&statsPeriod=24h) |
| **2f** | Broadcast + standalone flag OFF. | No transaction emitted | -- |

## :pencil: Checklist


sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
```diff
@@ -161,6 +170,31 @@ public void setAppLaunchedInForeground(final boolean appLaunchedInForeground) {
...
+  }
+
+  /** Trace ID from a non-activity app start transaction, to be reused by a later activity. */
+  public @Nullable SentryId getAppStartTraceId() {
+    return appStartTraceId;
+  }
+
+  public void setAppStartTraceId(final @Nullable SentryId traceId) {
+    this.appStartTraceId = traceId;
+  }
```
buenaflor: later on this should generally be fixed to be trace connected by session and not only the app start + ui.load transactions

buenaflor: can be improved later on with https://linear.app/getsentry/project/make-traces-more-meaningful-out-of-the-box-android-6e7c1f6a7c51/overview

sentry-android-core/src/main/java/io/sentry/android/core/PerformanceAndroidEventProcessor.java
```diff
@@ -216,9 +225,7 @@ private boolean hasAppStartSpan(final @NotNull SentryTransaction txn) {
     }
 
     final @Nullable SpanContext context = txn.getContexts().getTrace();
-    return context != null
-        && (context.getOperation().equals(APP_START_COLD)
-            || context.getOperation().equals(APP_START_WARM));
+    return context != null && context.getOperation().equals(STANDALONE_APP_START_OP);
```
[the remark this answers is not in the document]
buenaflor: no one will create a root transaction with `app.start.cold/warm`

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
```diff
@@ -357,41 +405,45 @@ public void registerLifecycleCallbacks(final @NotNull Application application) {
...
+   */
+  private void handleNoActivityStartIfNeededOnMain() {
     if (activeActivitiesCounter.get() == 0) {
       appLaunchedInForeground.setValue(false);
 
+      // Non-activity starts have no Activity signal for the pre-API 35 warm/cold heuristic.
+      // If ApplicationStartInfo did not resolve the type, classify the process start as cold.
+      if (appStartType == AppStartType.UNKNOWN) {
+        appStartType = AppStartType.COLD;
+      }
```
buenaflor, line 494: so basically pre-API 35 we don't really have a way to reliably tell in headless starts afaict - are we ok in classifying these as cold starts?

buenaflor:
@cursor review

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
buenaflor: cc @markushi

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -124,6 +129,11 @@ public void register(final @NotNull IScopes scopes, final @NotNull SentryOptions
     timeToFullDisplaySpanEnabled = this.options.isEnableTimeToFullDisplayTracing();
 
     application.registerActivityLifecycleCallbacks(this);
+
+    if (performanceEnabled && this.options.isEnableStandaloneAppStartTracing()) {
+      AppStartMetrics.getInstance().setOnNoActivityStartedListener(this::onNoActivityStarted);
+    }
```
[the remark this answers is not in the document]
buenaflor: this is not an issue, hybrids should not have native tracing enabled anyway.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -779,6 +839,16 @@ WeakHashMap<Activity, ISpan> getTtfdSpanMap() {
     }
   }
 
+  private @Nullable ISpan getAppStartParent(final @NotNull Activity activity) {
+    if (appStartTransaction != null) {
+      return appStartTransaction;
```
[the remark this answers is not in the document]
buenaflor: unlikely to be an issue

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
```diff
@@ -161,6 +169,28 @@ public void setAppLaunchedInForeground(final boolean appLaunchedInForeground) {
     this.appLaunchedInForeground.setValue(appLaunchedInForeground);
   }
 
+  public void setOnNoActivityStartedListener(final @Nullable OnNoActivityStartedListener listener) {
+    this.noActivityStartedListener = listener;
+    // On API 35+ registerLifecycleCallbacks resolves appStartType early and skips scheduling
+    // the idle check (because noActivityStartedListener is still null at that point). When the
+    // listener is set later during Sentry.init(), re-schedule if the idle check hasn't fired yet.
```
0xadam-brown: m: It sounds like we could have a situation where we've scheduled two IdleHandlers. That'd cause handleNoActivityStartIfNeededOnMain() to run twice, potentially creating two standalone transactions (AFAICT). 

Could we gate based on the scheduling of the handler rather than its being executed? 

(...nvm if I was misled by the comment and we can't really have two IdleHandlers scheduled simultaneously.)

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
```diff
@@ -49,6 +50,10 @@
  */
 @ApiStatus.Internal
 public class AppStartMetrics extends ActivityLifecycleCallbacksAdapter {
+  public interface OnNoActivityStartedListener {
```
0xadam-brown: l: Fwiw, I find the concept of "headless start" easier to understand at a glance than "no Activity start". If you feel similarly, we could rename throughout. (Up to you.)

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
```diff
@@ -161,6 +169,28 @@ public void setAppLaunchedInForeground(final boolean appLaunchedInForeground) {
...
+  public void setOnNoActivityStartedListener(final @Nullable OnNoActivityStartedListener listener) {
+    this.noActivityStartedListener = listener;
+    // On API 35+ registerLifecycleCallbacks resolves appStartType early and skips scheduling
+    // the idle check (because noActivityStartedListener is still null at that point). When the
+    // listener is set later during Sentry.init(), re-schedule if the idle check hasn't fired yet.
+    if (listener != null
+        && isCallbackRegistered
+        && firstIdle == -1
+        && appStartType != AppStartType.UNKNOWN) {
+      scheduleNoActivityStartCheckOnMain();
```
0xadam-brown: l: Can we get rid of this call and always invoke `scheduleNoActivityStartCheckOnMain()` in `registerLifecycleCallbacks()`? 

That'd eliminate the side-effect here, and it'd let us strip out the closely-coupled conditions guarding both `scheduleNoActivityStartCheckOnMain()` call sites.

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
```diff
@@ -161,6 +169,28 @@ public void setAppLaunchedInForeground(final boolean appLaunchedInForeground) {
...
+    }
+  }
+
+  /** Trace ID from a non-activity app start transaction, to be reused by a later activity. */
```
0xadam-brown: m: Thoughts about limiting reuse based on an expiration period for the appStartTraceId (eg, 15 seconds)? 

Eg, if a broadcast starts an app and the user only opens the app minutes later, the two events are probably unrelated, which suggests they shouldn't share a trace ID.

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
0xadam-brown: Makes sense to me. Looks like the IdleHandler is scheduled once in registerLifecycleCallbacks() or via Sentry.init(), both of which are triggered by Application.onCreate(). 

That^^ means the process should always be newly forked whenever we get here.

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
```diff
@@ -401,6 +456,62 @@ private void checkCreateTimeOnMain() {
...
+
+      resolveNonActivityAppStartEndTime();
+
+      if (noActivityStartedListener != null) {
+        noActivityStartedListener.onNoActivityStarted();
+      }
+    }
+  }
+
+  private void resolveNonActivityAppStartEndTime() {
```
0xadam-brown: l: A observation + a non-blocking question:

1. **Activity appStart end time** -> when the first frame drawn.

2. **Non-Activity appStart end time** -> when Application.onCreate() ends (with more or less fidelity depending on which fallback we use below). 

Those^^ are of course quite a bit different, given that (1) includes more than just Application.onCreate(). Are we okay with that difference, or do we have plans to – say – use byte code instrumentation to track the completion of calls to headless components (BroadcastReceiver, Service, etc.)?

Probably no perfect answer here – just curious about our reasoning.

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -788,12 +858,63 @@ WeakHashMap<Activity, ISpan> getTtfdSpanMap() {
...
+    txnOptions.setBindToScope(false);
+    txnOptions.setStartTimestamp(startTime);
+    setSpanOrigin(txnOptions);
+
+    final @NotNull TransactionContext txnContext =
+        new TransactionContext(
+            STANDALONE_APP_START_NAME,
+            TransactionNameSource.COMPONENT,
+            STANDALONE_APP_START_OP,
+            null);
```
0xadam-brown, line 1002: Any sampling synchronization concerns here? Eg, do we need to make sure whatever is applied to app.start is also applied to ui.load for all sampler types (including custom samplers)?

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
buenaflor: hm don't have a strong opinion here, wdyt @markushi

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
buenaflor: yeah I agree, I was actually working in my head with `headless` but not sure why I kept pushing NoActivity in the code

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
buenaflor: It's not in this PR but we will also include the [reason](https://developer.android.com/reference/android/app/ApplicationStartInfo#getReason()) as [app.vitals.start.reason](https://getsentry.github.io/sentry-conventions/attributes/app/#app-vitals-start-reason) why a certain app has started and reflect that in the mobile vitals dashboard - so it'll be clearer why some app starts are shorter than others.

So the shape what users will see in the dashboard (and query themselves) will be something in the direction of: 

```
Screen: empty (because headless)
Reason: broadcast
Start Duration: 100ms
Type: Cold

-> user knows this specifically was started headless 

vs

Screen: MyActivity
Reason: launcher
Start Duration: 1000ms
Type: Cold

```
so definitely we don't want to be ambiguous here and leave the interpretation up to the user

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
buenaflor: possibly there is since both could be generating different `sample_rand` - I'll have a look

this should also be solved by https://github.com/getsentry/sentry-java/pull/5383 but we cannot rely on it yet at this point

buenaflor: current behaviour is both `app.start` and `ui.load` transactions generate different `sampleRand` and it could happen that sampling is not uniform here. 

@romtsn @markushi not sure if it's worth addressing this right now since this will be fixed by the new trace improvements

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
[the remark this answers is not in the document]
buenaflor: we currently don't care about contributes ttid/ttfd (no product around it) so we can ignore this imo

sentry-android-core/src/test/java/io/sentry/android/core/ActivityLifecycleIntegrationTest.kt
```diff
@@ -882,6 +1100,102 @@ class ActivityLifecycleIntegrationTest {
...
+
+    val ttidSpan =
+      uiLoadTransaction.children.single { it.operation == ActivityLifecycleIntegration.TTID_OP }
+    assertTrue(ttidSpan.isFinished)
+    assertTrue(appStartTransaction.isFinished)
+    assertEquals(ttidSpan.finishDate, appStartTransaction.finishDate)
+    assertEquals(
+      ttidSpan.measurements[MeasurementValue.KEY_TIME_TO_INITIAL_DISPLAY]!!.value,
+      AppStartMetrics.getInstance().appStartTimeSpan.durationMs,
+    )
```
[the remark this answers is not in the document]
buenaflor: not a real issue

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
```diff
@@ -663,6 +665,53 @@ public void setEnablePerformanceV2(final boolean enablePerformanceV2) {
...
+  /**
+   * Enables or disables standalone app start tracing.
+   *
+   * <p>When enabled, app start is sent as its own transaction instead of an {@code app.start.*}
+   * child span on the first Activity transaction.
+   *
+   * <p>The SDK reports app start through these paths:
+   *
+   * <ul>
+   *   <li>With an Activity: the SDK sends an {@code App Start} transaction with operation {@code
```
runningcode: nit: not sure we want the `{@code` here and the next few lines in the javadoc since these uses don't refer to code.

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
```diff
@@ -663,6 +665,53 @@ public void setEnablePerformanceV2(final boolean enablePerformanceV2) {
...
+   *       app.start}, plus a separate {@code ui.load} transaction for the Activity. Both
+   *       transactions share the same trace ID.
+   *   <li>Headless app start: for launches started by something like a broadcast receiver, service,
+   *       or content provider without an Activity, the SDK sends only the standalone app-start
+   *       transaction.
+   *       <ul>
+   *         <li>On Android API 35 and newer, the SDK can use {@code ApplicationStartInfo} to
```
runningcode: Is this the device's sdk or the `compileSdk` ? I ask because sometimes changing the `compileSdk` can affect runtime behavior.

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
```diff
@@ -49,6 +50,10 @@
  */
 @ApiStatus.Internal
 public class AppStartMetrics extends ActivityLifecycleCallbacksAdapter {
+  public interface HeadlessAppStartListener {
```
runningcode, line 53: is this only meant for internal consumption?

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -788,12 +858,63 @@ WeakHashMap<Activity, ISpan> getTtfdSpanMap() {
...
+    final @Nullable SentryDate startTime = appStartTimeSpan.getStartTimestamp();
+    final @Nullable SentryDate endTime = appStartTimeSpan.getProjectedStopTimestamp();
+    if (startTime == null || endTime == null) {
+      return;
+    }
+
+    final TransactionOptions txnOptions = new TransactionOptions();
+    txnOptions.setBindToScope(false);
+    txnOptions.setStartTimestamp(startTime);
+    setSpanOrigin(txnOptions);
```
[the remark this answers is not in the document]
buenaflor: good point, what kind of trace origin should we use here?

sentry-android-core/src/main/java/io/sentry/android/core/PerformanceAndroidEventProcessor.java
```diff
@@ -84,9 +86,21 @@ public SentryEvent process(@NotNull SentryEvent event, @NotNull Hint hint) {
...
-        if (appStartMetrics.shouldSendStartMeasurements()) {
+        // For headless starts, appLaunchedInForeground is false, so only headless standalone app
+        // start transactions bypass the foreground check, not the duplicate-send guard.
+        final @Nullable SpanContext traceContext = transaction.getContexts().getTrace();
+        final boolean isStandaloneAppStartTxn =
+            traceContext != null && STANDALONE_APP_START_OP.equals(traceContext.getOperation());
+        final boolean isHeadlessStandaloneAppStartTxn =
+            traceContext != null
+                && isStandaloneAppStartTxn
+                && !traceContext.getData().containsKey(APP_START_SCREEN_DATA);
```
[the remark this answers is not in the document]
buenaflor: not an issue, the absence of a screen name is the indicator that this app start has no screen destination and therefore is a headless start

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
romtsn: this actually got me thinking that there's a case of a deferred Sentry.init (e.g. when our customers want to ask for a user consent before initializing the SDK), and that Sentry.init could've been called from an Activity.

I ran this idea through codex and this is what it's suggested:
```
• I’d handle it by treating “no observed Activity” as headless only if the process is not foreground.

  For deferred init inside an Activity, AppStartMetrics missed the earlier onActivityCreated, but ContextUtils.isForegroundImportance() should still report IMPORTANCE_FOREGROUND. So the headless check can do:

  if (activeActivitiesCounter.get() == 0) {
    if (ContextUtils.isForegroundImportance()) {
      appLaunchedInForeground.setValue(true);
      shouldSendStartMeasurements = false;
      activeActivitiesCounter.compareAndSet(0, 1); // avoid later destroy going negative
      return;
    }

    appLaunchedInForeground.setValue(false);
    if (appStartType == AppStartType.UNKNOWN) {
      appStartType = AppStartType.COLD;
    }
    ...
  }

  The important behavior is: if SDK init is late and the app is already foreground, we should not emit a headless standalone app-start transaction. We missed the true app start, so dropping that measurement is
  better than reporting a fake background/headless cold start.

 I’d add a regression test for: provider disabled, registerLifecycleCallbacks() happens late, process importance is foreground, headless listener is set, looper idles. Expected: listener is not called and no
  headless app.start is created.
```
wdyt?

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
```diff
@@ -401,6 +464,62 @@ private void checkCreateTimeOnMain() {
...
+    }
+
+    // Priority 2: API 35+ ApplicationStartInfo (cached from registerLifecycleCallbacks)
+    if (cachedStartInfo != null && Build.VERSION.SDK_INT >= Build.VERSION_CODES.VANILLA_ICE_CREAM) {
+      try {
+        final @NotNull Map<Integer, Long> timestamps = cachedStartInfo.getStartupTimestamps();
+        final @Nullable Long onCreateNanos =
+            timestamps.get(ApplicationStartInfo.START_TIMESTAMP_APPLICATION_ONCREATE);
+        if (onCreateNanos != null && onCreateNanos > 0) {
+          // ApplicationStartInfo documents "clock monotonic" timestamps in nanoseconds,
```
romtsn: I guess it's `SystemClock.uptimeNanos()` from [here](https://cs.android.com/android/platform/superproject/+/android-latest-release:frameworks/base/core/java/android/app/ActivityThread.java;l=8138?q=finishAttachApplication) so we could drop this complication?

sentry/src/main/java/io/sentry/TransactionContext.java
```diff
@@ -84,6 +84,25 @@ public TransactionContext(
     this.baggage = TracingUtils.ensureBaggage(null, samplingDecision);
   }
 
+  /**
+   * Creates {@link TransactionContext} that shares a trace ID with another transaction. Used for
+   * standalone app start transactions that need to belong to the same trace as the activity
+   * transaction.
+   */
+  @ApiStatus.Internal
```
romtsn: Could we open it up? I recall having a need in this so our customers could easily stitch multiple txs together, this would prevent them from using internal APIs

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@
 
 ### Features
 
+- Add `enableStandaloneAppStartTracing` option to send app start as a standalone transaction instead of attaching it as a child span of the first activity transaction ([#5342](https://github.com/getsentry/sentry-java/pull/5342))
+  - Disabled by default; opt in via `SentryAndroidOptions.setEnableStandaloneAppStartTracing(true)` or manifest meta-data `io.sentry.standalone-app-start-tracing.enable`
```
```suggestion
  - Disabled by default; opt in via `options.isEnableStandaloneAppStartTracing = true` or manifest meta-data `io.sentry.standalone-app-start-tracing.enable`
```
I guess we should prefer kotlin going forward :)

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@
 
 ### Features
```
romtsn, line 5: this is also a part of the already release version, I guess we should move it

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -637,22 +685,23 @@ private void onFirstFrameDrawn(final @Nullable ISpan ttfdSpan, final @Nullable I
     final @NotNull AppStartMetrics appStartMetrics = AppStartMetrics.getInstance();
     final @NotNull TimeSpan appStartTimeSpan = appStartMetrics.getAppStartTimeSpan();
     final @NotNull TimeSpan sdkInitTimeSpan = appStartMetrics.getSdkInitTimeSpan();
+    final @Nullable SentryDate firstFrameEndDate =
+        options != null && ttidSpan != null ? options.getDateProvider().now() : null;
```
romtsn: not sure why are we checking the `ttid span` here?

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -788,12 +858,63 @@ WeakHashMap<Activity, ISpan> getTtfdSpanMap() {
...
+    }
+
+    final @Nullable SentryDate startTime = appStartTimeSpan.getStartTimestamp();
+    final @Nullable SentryDate endTime = appStartTimeSpan.getProjectedStopTimestamp();
+    if (startTime == null || endTime == null) {
+      return;
+    }
+
+    final TransactionOptions txnOptions = new TransactionOptions();
+    txnOptions.setBindToScope(false);
```
romtsn, line 993: any specific reason not to bind it to scope here?

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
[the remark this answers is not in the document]
romtsn: `auto.app.start`? we even have it in docs, wouldn't have expected that 😅 https://github.com/getsentry/sentry-docs/blob/de23c1b80aa10fef96d07784d14fd04e6d4df7bb/develop-docs/sdk/telemetry/traces/trace-origin.mdx?plain=1#L44

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
romtsn: good catch @0xadam-brown! 

is it too much work to address this now @buenaflor? Given the state of session-based traces, I'm not sure if we're not going to receive more pushback and drag it for god knows how long, so wouldn't rely on it, if possible 😅

romtsn (commented):
Left some comments, overall nicely done! Let's see where do we go after today's sync.

And one more point I'd like to maybe document is the 1 minute threshold warm starts that we'd discussed in DM. Perhaps we should leave a comment that it's not gonna work as expected for standalone?

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
buenaflor: I think it is some work since it would basically require all the changes done regarding the session based traces, e.g changes to the core java sdk and how we sample in startTransaction since it also has implications on DSC and things like that

sentry/src/main/java/io/sentry/TransactionContext.java
buenaflor: we can open it up but not guarantee sampleRand consistency which only works with the manual continueTrace hack right now, wdyt?

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -239,33 +251,68 @@ private void startTracing(final @NotNull Activity activity) {
...
-
-          // in case there's already an end time (e.g. due to deferred SDK init)
-          // we can finish the app-start span
-          finishAppStartSpan();
+          if (options.isEnableStandaloneAppStartTracing() && !isFollowingHeadlessAppStart) {
+            final TransactionOptions appStartTransactionOptions = new TransactionOptions();
+            appStartTransactionOptions.setBindToScope(false);
+            appStartTransactionOptions.setStartTimestamp(appStartTime);
+            appStartTransactionOptions.setAppStartTransaction(appStartSamplingDecision != null);
+            setSpanOrigin(appStartTransactionOptions);
```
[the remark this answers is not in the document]
buenaflor: @romtsn I guess we wanna keep `auto.ui.activity` for the normal foreground app start? or should we change it so both are `auto.app.start` right now?

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
buenaflor: we can drop it, I guess it's just a defensive stance but we don't really need it

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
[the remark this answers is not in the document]
buenaflor: discussed in other flagged review

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
buenaflor: not sure if it makes much sense, typically the headless transaction will start and end at a defined time (it's not an idle transaciton like ui.load)

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
buenaflor: okay we decided we'll just (for now) use `continueTrace` to achieve this

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
romtsn: @buenaflor is that the case we'd discussed about the 1 minute threshold and warm starts classification?

sentry/src/main/java/io/sentry/TransactionContext.java
romtsn: yeah fine to not do that after the continueTrace decision we've made yday. I just wanted to have a nicer API to continue a trace 😅

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
buenaflor: The topic here is if we should trace connect headless app starts and ui.load if there is a large gap between them<br><br>currently the implementation does trace connect them regardless of how far apart the headless start and ui.load is

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
romtsn: hm, I guess our customers could create custom spans and add them to the standalone tx when it's on the scope? On the other hand when https://github.com/getsentry/sentry-java/issues/5294 lands, there's probably no need for it, since we could use the explicit API to customize app launch?

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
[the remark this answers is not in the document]
romtsn: origin is per root/transaction right? shall we have auto.app.start for the appstart tx, and auto.ui.activity for the following ui.load?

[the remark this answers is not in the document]
buenaflor: ah yeah definitely

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
buenaflor: yeah makes sense

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
romtsn: ah alright, then I'd suggest the same 1 minute threshold here to connect them, otherwise set them apart, wdyt?

sentry/src/main/java/io/sentry/TransactionContext.java
buenaflor: yeah, since we now use continueTrace we don't actually need this anymore and can remove it

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
buenaflor: yeah that should be the idiomatic API to use for adding custom spans

buenaflor:
Updated:
- Added/kept the headless-to-ui.load trace connection behavior, including shared trace/sampleRand and a 1-minute cutoff for unrelated later activities -> if a headless app start and a ui.load is more than 1 minute apart then they won't be trace connected.
- Fixed standalone app-start transaction origin so `app.start` uses `auto.app.start`, while `ui.load` remains `auto.ui.activity`

I also ran the E2E tests again described in the description

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -316,6 +397,61 @@ private void setSpanOrigin(final @NotNull SpanOptions spanOptions) {
...
+    return uiLoadStartTime.diff(appStartEndTime) <= APP_START_TO_UI_LOAD_CONTINUATION_MAX_GAP_NANOS;
+  }
+
+  /**
+   * Builds a {@link TransactionContext} for the ui.load transaction that shares the standalone
+   * app.start trace (same traceId and sampleRand) while staying a sibling (no parentSpanId), rather
+   * than a child. The continued baggage keeps sampling decisions on the same sampleRand. Returns
+   * null if the trace cannot be continued, so callers can fall back.
+   */
+  private @Nullable TransactionContext continueUiLoadTrace(
```
buenaflor, line 418: `Sentry.continueTrace(...)` models standard distributed trace continuation, which makes the new transaction a child of the incoming span so we cannot really use it directly. For app start + ui.load, we need sibling transactions in the same trace, so we build the TransactionContext from the parsed propagation data while intentionally clearing parentSpanId

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
buenaflor: Updated the impl to not trace connect app start and ui load span if they are more than 1 minute apart

sentry-samples/sentry-samples-android/src/main/AndroidManifest.xml
```diff
@@ -234,10 +225,6 @@
       android:name="io.sentry.performance-v2.enable"
       android:value="true" />
 
-    <meta-data
```
romtsn: should we keep this on for the sample app?

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/TestBroadcastReceiver.java
romtsn, line 1: i think this one may also be useful to test this scenario in the future?

buenaflor: Re-added it

sentry-samples/sentry-samples-android/src/main/AndroidManifest.xml
buenaflor: Updated it so standalone app start is on by default for the sample

sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
romtsn: I think it would be great to have a public API method for this (like Sentry.connectRootSpans or something), but that's for later

romtsn (approved):
LGTM! great work

runningcode (approved):
Just some general comments on code, I'm not so familiar with app tracing though.

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
```diff
@@ -401,6 +502,57 @@ private void checkCreateTimeOnMain() {
...
+        listener.onHeadlessAppStart();
+      }
+    }
+  }
+
+  private void resolveHeadlessAppStartEndTime() {
+    // Priority 1: Gradle plugin instrumented onApplicationPostCreate
+    if (applicationOnCreate.hasStopped()) {
+      final long stopUptimeMs =
+          applicationOnCreate.getStartUptimeMs() + applicationOnCreate.getDurationMs();
```
runningcode, line 518: what is upTime in this case? is it the app's upTime or the system's upTime?

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
```diff
@@ -401,6 +502,57 @@ private void checkCreateTimeOnMain() {
...
+          // ActivityThread.handleBindApplication), the same base as TimeSpan, so we can use it
+          // directly as an uptime-millis value without re-anchoring.
+          final long onCreateUptimeMs = TimeUnit.NANOSECONDS.toMillis(onCreateNanos);
+          if (applicationOnCreate.hasStarted() && applicationOnCreate.hasNotStopped()) {
+            applicationOnCreate.setStoppedAt(onCreateUptimeMs);
+          }
+          stopHeadlessAppStartAt(onCreateUptimeMs);
+          return;
+        }
+      } catch (Throwable ignored) {
```
runningcode, line 539: In general it is a bad practice to catch `Throwable` and ignore it. This could catch OOMs or StackOverflow exceptions. I will bring this up in the next SDK sync

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
buenaflor: yes only internally

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
buenaflor: it's system in this case

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
buenaflor: In this case it's the device's, updated it

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
buenaflor: updated it 👍

PR #5345, opened 2026-04-29
fix(spring-jakarta): [Queue Instrumentation 38] Close leaked Kafka interceptor scope

adinauer:
## PR Stack (Queue Instrumentation)

- #5249
- #5250
- #5253
- #5254
- #5255
- #5259
- #5260
- #5265
- #5274
- #5279
- #5280
- #5281
- #5282
- #5283
- #5288
- #5289
- #5291
- #5312
- #5313
- #5314
- #5315
- #5316
- #5318
- #5319
- #5320
- #5321
- #5322
- #5323
- #5324
- #5325
- #5327
- #5328
- #5330
- #5337
- #5338
- #5341
- #5343
- #5345
- #5347
- #5348
- #5352
- #5368

---

## :scroll: Description

Ensure `SentryKafkaRecordInterceptor` does not leak its forked scopes when interceptor setup throws.

The interceptor now stores the lifecycle token in `currentContext` immediately after `makeCurrent()`, before `continueTrace()` or `startTransaction()` can fail. If setup fails, the existing cleanup path can still close the token instead of leaving the forked scopes current on the consumer thread.

## :bulb: Motivation and Context

A review comment pointed out that `continueTrace()` or `startTransaction()` could throw after `makeCurrent()` succeeded but before the lifecycle token was saved into the thread-local. In that case `failure()` and `clearThreadState()` would see no context to finish, and the lifecycle token would never be closed.

This change keeps the cleanup path intact and logs setup failures without breaking Kafka consumer processing.

## :green_heart: How did you test it?

- Ran `./gradlew spotlessApply apiDump`
- Ran `./gradlew :sentry-spring-jakarta:test --tests='*SentryKafkaRecordInterceptorTest*'`

## :pencil: Checklist
- [ ] I added GH Issue ID _&_ Linear ID
- [ ] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [ ] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

## :crystal_ball: Next steps

Continue tightening the remaining Queue Instrumentation error paths.

#skip-changelog

> ⚠️ **Merge this PR using a merge commit** (not squash). Only the collection branch is squash-merged into main.

PR #5343, opened 2026-04-29
fix(spring-boot-jakarta): [Queue Instrumentation 37] Skip Kafka autoconfig for OTel agent

adinauer:
## PR Stack (Queue Instrumentation)

- #5249
- #5250
- #5253
- #5254
- #5255
- #5259
- #5260
- #5265
- #5274
- #5279
- #5280
- #5281
- #5282
- #5283
- #5288
- #5289
- #5291
- #5312
- #5313
- #5314
- #5315
- #5316
- #5318
- #5319
- #5320
- #5321
- #5322
- #5323
- #5324
- #5325
- #5327
- #5328
- #5330
- #5337
- #5338
- #5341
- #5343
- #5345
- #5347
- #5348
- #5352
- #5368

---

## :scroll: Description

Prevent Spring Kafka auto-configuration from installing Sentry Kafka bean post-processors when the Sentry OpenTelemetry Java agent is active.

This change also updates the Jakarta OTel coexistence system tests to assert the intended behavior: OTel queue spans still reach Sentry, but the Sentry `SpringKafka` integration stays disabled.

## :bulb: Motivation and Context

The Kafka auto-config gate only checked for `SentryAutoConfigurationCustomizerProvider`, which covered the app-classpath OTel integration but missed the Java agent path. As a result, the agent sample still installed `SentryKafkaProducerBeanPostProcessor`, and the coexistence test failed for the wrong reason.

Adding `AgentMarker` to the conditional fixes the real suppression bug and aligns the tests with the current OTel messaging mapping behavior.

## :green_heart: How did you test it?

- Ran `./gradlew spotlessApply apiDump`
- Ran `./gradlew :sentry-spring-boot-jakarta:test --tests='*SentryKafkaAutoConfigurationTest' -x :sentry-spring-jakarta:compileJava -x :sentry-spring-jakarta:jar`
- Ran `./gradlew :sentry-spring-boot-jakarta:compileTestKotlin :sentry-samples:sentry-samples-spring-boot-jakarta-opentelemetry:compileTestKotlin :sentry-samples:sentry-samples-spring-boot-jakarta-opentelemetry-noagent:compileTestKotlin -x :sentry-spring-jakarta:compileJava -x :sentry-spring-jakarta:jar`

## :pencil: Checklist
- [ ] I added GH Issue ID _&_ Linear ID
- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
[...]

PR #5291, opened 2026-04-14
fix(test): [Queue Instrumentation 16] Enable Kafka profile in system runner

adinauer:
## PR Stack (Queue Instrumentation)

- #5249
- #5250
- #5253
- #5254
- #5255
- #5259
- #5260
- #5265
- #5274
- #5279
- #5280
- #5281
- #5282
- #5283
- #5288
- #5289
- #5291
- #5312
- #5313
- #5314
- #5315
- #5316
- #5318
- #5319
- #5320
- #5321
- #5322
- #5323
- #5324
- #5325
- #5327
- #5328
- #5330
- #5337
- #5338
- #5341
- #5343
- #5345
- #5347
- #5348
- #5352
- #5368

---

## :scroll: Description

Finalize the top Kafka stack PR with the Spring Boot Jakarta review fix.

Changes:
- define module lists for Kafka broker requirements and Kafka Spring profile requirements
- start Kafka for modules that require it
- set `SPRING_PROFILES_ACTIVE=kafka` for modules that need Kafka endpoints enabled
- require `io.sentry.kafka.SentryKafkaProducerInterceptor` before activating Spring Boot Jakarta Kafka queue auto-configuration
- add a regression test proving Kafka bean post-processors are skipped when `sentry-kafka` is absent from the classpath

## :bulb: Motivation and Context

Kafka queue system tests for `sentry-samples-spring-boot-jakarta` were running without the
`kafka` profile, so `/kafka` endpoints were not active and tests failed with 404s.

This PR also fixes the late stack regression where Spring Boot Jakarta queue auto-configuration
could activate without the optional `sentry-kafka` module being present.

## :green_heart: How did you test it?

- Ran `./gradlew spotlessApply apiDump`
- Ran `./gradlew :sentry-spring-boot-jakarta:test --tests='*SentryKafkaAutoConfigurationTest'`
- Verified the system runner script changes and Spring Boot Jakarta Kafka auto-config regression test

## :pencil: Checklist
- [ ] I added GH Issue ID _&_ Linear ID
- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [ ] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

## :crystal_ball: Next steps

#skip-changelog

[...]

PR #5289, opened 2026-04-13
feat(samples): [Queue Instrumentation 15] Add opt-in Kafka console e2e coverage

adinauer:
## PR Stack (Queue Instrumentation)

- #5249
- #5250
- #5253
- #5254
- #5255
- #5259
- #5260
- #5265
- #5274
- #5279
- #5280
- #5281
- #5282
- #5283
- #5288
- #5289
- #5291
- #5312
- #5313
- #5314
- #5315
- #5316
- #5318
- #5319
- #5320
- #5321
- #5322
- #5323
- #5324
- #5325
- #5327
- #5328
- #5330
- #5337
- #5338
- #5341
- #5343
- #5345
- #5347
- #5348
- #5352
- #5368

---

## :scroll: Description

Add console-sample Kafka coverage for queue instrumentation and make the sample’s Kafka flow opt-in so existing system tests stay stable when Kafka is not available.

This PR includes:
- a dedicated `KafkaShowcase` flow in the console sample
- a no-arg producer interceptor constructor path used by Kafka config loading
- a consumer demo path so both producer and consumer queue spans are exercised
- a system test path that enables Kafka only when `SENTRY_SAMPLE_KAFKA_BOOTSTRAP_SERVERS` is set

## :bulb: Motivation and Context

After extracting `sentry-kafka` into a standalone module, we need end-to-end coverage outside Spring to validate queue span creation and trace propagation for plain `kafka-clients` usage.

## :green_heart: How did you test it?

- Ran `./gradlew spotlessApply apiDump`
- Added/updated tests in `sentry-kafka` for producer and consumer interceptors
- Added console sample system test coverage for Kafka-enabled flow

## :pencil: Checklist
- [ ] I added GH Issue ID _&_ Linear ID
- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [ ] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

## :crystal_ball: Next steps

#skip-changelog

> ⚠️ **Merge this PR using a merge commit** (not squash). Only the collection branch is squash-merged into main.

PR #5347, opened 2026-04-29
fix(test): [Queue Instrumentation 39] Remove stale Kafka container before startup

adinauer:

- #5249
- #5250
- #5253
- #5254
- #5255
- #5259
- #5260
- #5265
- #5274
- #5279
- #5280
- #5281
- #5282
- #5283
- #5288
- #5289
- #5291
- #5312
- #5313
- #5314
- #5315
- #5316
- #5318
- #5319
- #5320
- #5321
- #5322
- #5323
- #5324
- #5325
- #5327
- #5328
- #5330
- #5337
- #5338
- #5341
- #5343
- #5345
- #5347
- #5348
- #5352
- #5368

---

## :scroll: Description

Clean up stale Kafka system-test containers before starting the local broker.

`SystemTestRunner.start_kafka_broker()` now force-removes the named container up front, even when the current runner instance did not start the previous broker. `stop_kafka_broker()` stays ownership-aware and still only tears down brokers started by the current run.

A review comment on the original Kafka console sample PR pointed out that interrupted runs could leave behind a stopped container named `sentry-java-system-test-kafka`. The runner tried to clean that up before startup, but the cleanup path returned early whenever `kafka_started_by_runner` was false, so the next `docker run --name ...` failed with a container-name conflict.

This change fixes the stale-container path without changing the reuse behavior for an already running broker on `localhost:9092`.

I also looked into replacing Redpanda with the official Apache Kafka image. That is possible, but it is not a drop-in swap because the current startup command is Redpanda-specific, so I left that as a separate follow-up.

- Ran `python3 -m py_compile test/system-test-runner.py`
- Ran a mocked Python verification that `start_kafka_broker()` now issues `docker rm -f sentry-java-system-test-kafka` before `docker run`

## :pencil: Checklist

## :crystal_ball: Next steps

If we want the most official broker image for these tests, switch the runner to `apache/kafka` in a separate PR.

#skip-changelog

### DIFF
--- a/test/system-test-runner.py
+++ b/test/system-test-runner.py
@@ -442,6 +442,7 @@ class SystemTestRunner:
 
         if self.module_requires_kafka_profile(sample_module):
             env["SPRING_PROFILES_ACTIVE"] = "kafka"
+            env["SENTRY_ENABLE_QUEUE_TRACING"] = "true"
             print("Enabling Spring profile: kafka")
         else:
             env.pop("SPRING_PROFILES_ACTIVE", None)
--- a/test/system-test-runner.py
+++ b/test/system-test-runner.py
@@ -227,13 +227,21 @@ class SystemTestRunner:
                 time.sleep(1)
         return False
 
+    def remove_kafka_broker_container(self) -> None:
+        subprocess.run(
+            ["docker", "rm", "-f", KAFKA_CONTAINER_NAME],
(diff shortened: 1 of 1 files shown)


lbloder (approved):
Changes approved, we need to look into the build failures though, i suggest we do that at the last one in the stack.

PR #5349, opened 2026-04-29
ref(feedback): Rename Dialog to Form across feedback APIs

### LINKED ISSUE #5304: Feedback: Rename FeedbackWidget to FeedbackForm
* Deprecate “widget” term → new API uses Form
* Rename FeedbackWidget to FeedbackForm ([History](<https://github.com/getsentry/sentry-react-native/pull/4547>))

romtsn:
Rename `SentryUserFeedbackDialog` to `SentryUserFeedbackForm` as the primary feedback UI class. The old `SentryUserFeedbackDialog` is kept as a deprecated subclass for backward compatibility — existing user code continues to compile and work unchanged.

**Public API: `Sentry.showUserFeedbackForm()`**

New `Sentry.showUserFeedbackForm()` methods replace the deprecated `Sentry.showUserFeedbackDialog()` overloads, which now delegate to the new ones.

**Internal API renames**

Internal-only types and methods are renamed directly (no deprecation shim needed): `IDialogHandler` → `IFormHandler`, `showDialog()` → `showForm()`, `setDialogHandler()`/`getDialogHandler()` → `setFormHandler()`/`getFormHandler()`, `AndroidUserFeedbackIDialogHandler` → `AndroidUserFeedbackFormHandler`.

Closes #5304

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/SentryUserFeedbackDialog.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/SentryUserFeedbackDialog.java
@@ -62,8 +62,7 @@ public final class SentryUserFeedbackDialog extends SentryUserFeedbackForm {
      */
     @Deprecated
     public Builder(
-        final @NotNull Context context,
-        final @Nullable SentryUserFeedbackForm.OptionsConfiguration configuration) {
+        final @NotNull Context context, final @Nullable OptionsConfiguration configuration) {
       super(context, configuration);
     }
 
@@ -80,7 +79,7 @@ public final class SentryUserFeedbackDialog extends SentryUserFeedbackForm {
     public Builder(
         final @NotNull Context context,
         final int themeResId,
-        final @Nullable SentryUserFeedbackForm.OptionsConfiguration configuration) {
+        final @Nullable OptionsConfiguration configuration) {
       super(context, themeResId, configuration);
     }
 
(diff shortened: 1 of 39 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/SentryUserFeedbackDialog.java
```diff
@@ -1,380 +1,114 @@
...
-import io.sentry.SentryIntegrationPackageStorage;
-import io.sentry.SentryLevel;
-import io.sentry.SentryOptions;
-import io.sentry.protocol.Feedback;
 import io.sentry.protocol.SentryId;
-import io.sentry.protocol.User;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-public final class SentryUserFeedbackDialog extends AlertDialog {
```
markushi, line 24: nit: Should we keep this and mark it as deprecated instead? I'm fine with removing, as I suspect nobody used it anyway.
~~
```
@Deprecated
public final class SentryUserFeedbackDialog extends SentryUserFeedbackForm {}
```

markushi (approved):

sentry-android-core/src/main/java/io/sentry/android/core/SentryUserFeedbackDialog.java
romtsn: that's already done :)

markushi (approved):

PR #5372, opened 2026-05-06
ci: Fix Spring Boot matrix version updates

adinauer:
## :scroll: Description
Fix the Spring Boot 2.x, 3.x, and 4.x matrix workflows so their version override matches the TOML version catalog format.

The workflows now preserve whitespace around the assignment and replace the quoted Spring Boot version value in `gradle/libs.versions.toml`.

The existing `sed` expressions looked for keys without spaces around `=`, but the version catalog uses entries like `springboot2 = "2.7.18"`. As a result, matrix jobs did not actually update the Spring Boot version under test.

- Ran the updated `sed` expressions against a temporary copy of `gradle/libs.versions.toml` and verified Spring Boot 2, 3, and 4 versions were updated.

## :pencil: Checklist

## :crystal_ball: Next steps

#skip-changelog

### DIFF
--- a/.github/workflows/spring-boot-2-matrix.yml
+++ b/.github/workflows/spring-boot-2-matrix.yml
@@ -62,7 +62,7 @@ jobs:
 
       - name: Update Spring Boot 2.x version
         run: |
-          sed -i 's/^springboot2=.*/springboot2=${{ matrix.springboot-version }}/' gradle/libs.versions.toml
+          sed -i 's/^\(springboot2[[:space:]]*=[[:space:]]*\)".*"/\1"${{ matrix.springboot-version }}"/' gradle/libs.versions.toml
           echo "Updated Spring Boot 2.x version to ${{ matrix.springboot-version }}"
 
       - name: Exclude android modules from build
--- a/.github/workflows/spring-boot-2-matrix.yml
+++ b/.github/workflows/spring-boot-2-matrix.yml
@@ -141,6 +141,7 @@ jobs:
             --build "true"
 
       - name: Test sentry-samples-spring-boot-opentelemetry-noagent
+        if: ${{ !startsWith(matrix.springboot-version, '2.1.') && !startsWith(matrix.springboot-version, '2.2.') }}
         run: |
           python3 test/system-test-runner.py test \
             --module "sentry-samples-spring-boot-opentelemetry-noagent" \
--- a/.github/workflows/spring-boot-2-matrix.yml
+++ b/.github/workflows/spring-boot-2-matrix.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        springboot-version: [ '2.1.0', '2.2.5', '2.4.13', '2.5.15', '2.6.15', '2.7.0', '2.7.18' ]
+        springboot-version: [ '2.4.13', '2.5.15', '2.6.15', '2.7.0', '2.7.18' ]
 
     name: Spring Boot ${{ matrix.springboot-version }}
     env:
@@ -63,9 +63,6 @@ jobs:
       - name: Update Spring Boot 2.x version
         run: |
           springboot_version="${{ matrix.springboot-version }}"
-          if [[ "$springboot_version" =~ ^2\.[0-3]\. ]]; then
-            springboot_version="${springboot_version}.RELEASE"
-          fi
           if [[ ! "$springboot_version" =~ ^2\.7\. ]]; then
             echo "ORG_GRADLE_PROJECT_excludeGraphql=true" >> "$GITHUB_ENV"
             echo "ORG_GRADLE_PROJECT_excludeKafka=true" >> "$GITHUB_ENV"
@@ -141,7 +138,6 @@ jobs:
             --build "true"
 
       - name: Test sentry-samples-spring-boot-opentelemetry-noagent
-        if: ${{ !startsWith(matrix.springboot-version, '2.1.') && !startsWith(matrix.springboot-version, '2.2.') }}
         run: |
           python3 test/system-test-runner.py test \
             --module "sentry-samples-spring-boot-opentelemetry-noagent" \
--- a/.github/workflows/spring-boot-2-matrix.yml
+++ b/.github/workflows/spring-boot-2-matrix.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        springboot-version: [ '2.7.17', '2.7.18' ]
+        springboot-version: [ '2.1.0', '2.2.5', '2.4.13', '2.5.15', '2.6.15', '2.7.0', '2.7.18' ]
 
     name: Spring Boot ${{ matrix.springboot-version }}
     env:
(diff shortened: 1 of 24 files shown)


alexander-alderman-webb (approved):

sentry-samples/sentry-samples-spring-boot-opentelemetry-noagent/build.gradle.kts
```diff
@@ -103,7 +120,19 @@ tasks.jar {
 
 tasks.startScripts { dependsOn(tasks.shadowJar) }
 
-configure<SourceSetContainer> { test { java.srcDir("src/test/java") } }
+configure<SourceSetContainer> {
+  main {
+    if (!includeGraphql) {
```
runningcode, line 124: For the best gradle caching it would be best to make these 3 separate source sets.
This way they can each get their own cache entry and be compiled separately.
This depends however on what their dependencies look like but I imagine that if they are easily excluded like this that the sources are independent.

EDIT: after scrolling further down and seeing that they also have different imports maybe I need to understand better why we need this. could you explain it?

sentry-samples/sentry-samples-spring-boot-opentelemetry-noagent/build.gradle.kts
```diff
@@ -103,7 +120,19 @@ tasks.jar {
...
+      java.exclude("**/graphql/**")
+      resources.exclude("graphql/**")
+    }
+    if (!includeKafka) {
+      java.exclude("**/queues/kafka/**")
+      resources.exclude("application-kafka.properties")
+    }
+  }
+  test { java.srcDir("src/test/java") }
```
runningcode: nit: if this is the implicit default, why do we need this?
Does spring boot or something else change it?

sentry-samples/sentry-samples-spring-boot-opentelemetry-noagent/build.gradle.kts
```diff
@@ -121,7 +150,15 @@ tasks.register<Test>("systemTest").configure {
   minHeapSize = "128m"
   maxHeapSize = "1g"
 
-  filter { includeTestsMatching("io.sentry.systemtest*") }
+  filter {
```
runningcode, line 151: Same as above, if we use 3 separate test source sets/tasks we can compile and execute/cache them independently.

sentry-samples/sentry-samples-spring-boot-opentelemetry/build.gradle.kts
```diff
@@ -15,22 +15,34 @@ group = "io.sentry.sample.spring-boot"
...
 configure<JavaPluginExtension> {
-  sourceCompatibility = JavaVersion.VERSION_17
-  targetCompatibility = JavaVersion.VERSION_17
+  sourceCompatibility = JavaVersion.VERSION_11
+  targetCompatibility = JavaVersion.VERSION_11
 }
 
 tasks.withType<KotlinCompile>().configureEach {
-  compilerOptions.jvmTarget = org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_17
+  compilerOptions.jvmTarget = org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_11
```
runningcode: nit: should we import this?

sentry-spring-boot/src/main/java/io/sentry/spring/boot/SentryAutoConfiguration.java
```diff
@@ -203,11 +200,12 @@ static class ContextTagsEventProcessorConfiguration {
     @Configuration(proxyBeanMethods = false)
     @Import(SentryGraphqlAutoConfiguration.class)
     @Open
-    @ConditionalOnClass({
-      SentryGraphqlExceptionHandler.class,
-      DataFetcherExceptionResolverAdapter.class,
-      GraphQLError.class
-    })
+    @ConditionalOnClass(
+        name = {
```
runningcode, line 204: Should we add a comment on why we changed these to strings? It breaks type safety.

sentry-samples/sentry-samples-spring-boot/build.gradle.kts
```diff
@@ -102,7 +119,19 @@ tasks.jar {
...
+    if (!includeGraphql) {
+      java.exclude("**/graphql/**")
+      resources.exclude("graphql/**")
+    }
+    if (!includeKafka) {
+      java.exclude("**/queues/kafka/**")
+      resources.exclude("application-kafka.properties")
+    }
+  }
+  test { java.srcDir("src/test/java") }
```
runningcode: nit: is this the default?

.github/workflows/spring-boot-2-matrix.yml
```diff
@@ -62,8 +62,13 @@ jobs:
...
       - name: Update Spring Boot 2.x version
         run: |
-          sed -i 's/^springboot2=.*/springboot2=${{ matrix.springboot-version }}/' gradle/libs.versions.toml
-          echo "Updated Spring Boot 2.x version to ${{ matrix.springboot-version }}"
+          springboot_version="${{ matrix.springboot-version }}"
+          if [[ ! "$springboot_version" =~ ^2\.7\. ]]; then
+            echo "ORG_GRADLE_PROJECT_excludeGraphql=true" >> "$GITHUB_ENV"
+            echo "ORG_GRADLE_PROJECT_excludeKafka=true" >> "$GITHUB_ENV"
+          fi
+          sed -i 's/^\(springboot2[[:space:]]*=[[:space:]]*\)".*"/\1"'"$springboot_version"'"/' gradle/libs.versions.toml
```
runningcode: Since this PR is about fixing the fact that this `sed` wasn't replacing versions correctly, how do we prevent changes or refactoring of `libs.versions.toml` file from breaking this? Can we make this fail if no replacement is found?

sentry-samples/sentry-samples-spring-boot-opentelemetry-noagent/build.gradle.kts
adinauer: Since older versions of Spring Boot 2 don't have GraphQL support, we're excluding it from the sample in the matrix build.

sentry-samples/sentry-samples-spring-boot-opentelemetry-noagent/build.gradle.kts
adinauer: Probably redundant, can try to remove

sentry-samples/sentry-samples-spring-boot-opentelemetry/build.gradle.kts
adinauer: sure

sentry-spring-boot/src/main/java/io/sentry/spring/boot/SentryAutoConfiguration.java
adinauer: AFAIR If classes don't exist during build, it fails. Which is the case for older Spring Boot 2 versions.

.github/workflows/spring-boot-2-matrix.yml
adinauer: Yeah we should

sentry-spring-boot/src/main/java/io/sentry/spring/boot/SentryAutoConfiguration.java
adinauer: Yeah just confirmed, also we have conditional annotations using Strings in multiple places. Don't think we need a comment for it.

.github/workflows/spring-boot-2-matrix.yml
adinauer: updated

sentry-samples/sentry-samples-spring-boot-opentelemetry/build.gradle.kts
adinauer: updated

runningcode (approved):

PR #5380, opened 2026-05-06
chore: Add Dependabot

runningcode:
## Summary
- Adds `gradle` package ecosystem to Dependabot config with the Gradle Plugin Portal registry so plugin/dependency updates are tracked automatically.
- Updates both `gradle` and `github-actions` ecosystems to run daily.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,6 +12,8 @@ updates:
       - gradle-plugin-portal
     schedule:
       interval: "daily"
+    ignore:
+      - dependency-name: "org.springframework.boot*"
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,6 +1,18 @@
 version: 2
+registries:
+  gradle-plugin-portal:
+    type: maven-repository
+    url: https://plugins.gradle.org/m2
+    username: dummy # Required by dependabot
+    password: dummy # Required by dependabot
 updates:
+  - package-ecosystem: "gradle"
+    directory: "/"
+    registries:
+      - gradle-plugin-portal
+    schedule:
+      interval: "daily"
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: weekly
+      interval: "daily"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,7 +14,11 @@ updates:
       interval: "daily"
     ignore:
       - dependency-name: "org.springframework.boot*"
+    commit-message:
+      prefix: "chore(deps)"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,6 +1,18 @@
 version: 2
+registries:
+  gradle-plugin-portal:
+    type: maven-repository
+    url: https://plugins.gradle.org/m2
+    username: dummy # Required by dependabot
+    password: dummy # Required by dependabot
 updates:
+  - package-ecosystem: "gradle"
```
0xadam-brown, line 9: l/m: Thoughts about using [groups](https://docs.github.com/en/code-security/tutorials/secure-your-dependencies/optimizing-pr-creation-version-updates#prioritizing-meaningful-updates) to reduce the number of PRs generated (and the pain of sequential merges) + ensure we only get a single PR for deps that need to be bumped together?

We could even have dependabot [group everything into a single PR](https://carrick.eu/blog/faster-dependabot-updates-with-groups/)...

.github/dependabot.yml
```diff
@@ -1,6 +1,18 @@
...
+    directory: "/"
+    registries:
+      - gradle-plugin-portal
+    schedule:
```
0xadam-brown, line 13: I think we need a custom commit message to match our `chore(deps)` convention:

```
 commit-message: 
    prefix: "chore"
    prefix-development: "chore"
```

.github/dependabot.yml
```diff
@@ -1,6 +1,18 @@
...
+      interval: "daily"
```
0xadam-brown, line 14: l: My vote would be for weekly and we can see about increasing cadence as we go, esp as we were getting by with nothing.

.github/dependabot.yml
runningcode: good point, will update

.github/dependabot.yml
runningcode: im curious what other think here. i would favor the daily at first so we can catch up and then we can adjust it later once we've caught up.

.github/dependabot.yml
runningcode: That's a good point. I think the issue here is that some of our version numbers are explicitly set to be older versions in order to ensure compatibility like the spring boot.

0xadam-brown: Ah, good to exclude those versions so we don't get pinged for them repeatedly.

runningcode: I added the exclusion for spring boot. I know there are others. We can add them as they come in.

.github/dependabot.yml
runningcode: added

0xadam-brown (approved):

Issue #5340, opened 2026-04-28
Session Replay: race between PixelCopy.request and executor compositing/masking on shared screenshot bitmap


romtsn:
## Summary

The PixelCopy-based replay strategy (`PixelCopyStrategy`) shares a single `screenshot` `Bitmap` between two writers:

1. The system PixelCopy worker on the RenderThread, which writes to `screenshot` when the next frame's `PixelCopy.request(window, screenshot, …)` performs its blit.
2. The replay executor (single-threaded), which reads/writes `screenshot` inside `applyMaskingAndNotify` (mask render) and `compositeSurfaceViewsAndMask` (SurfaceView composite).

If a frame's executor work is still in progress when the next recorder cycle fires `PixelCopy.request`, the system PixelCopy worker overwrites parts of `screenshot` while the executor is still touching it. The executor being single-threaded prevents executor↔executor overlap, but does not gate the system PixelCopy worker.

The same shape exists in the no-SurfaceView mask path; the SurfaceView capture feature ([#5333](<https://github.com/getsentry/sentry-java/pull/5333>)) just makes it easier to hit because compositing adds a draw pass to the executor side.

## What this looks like in practice

Visual:

* **Torn frames** — emitted `screenshot` ends up with a strip of old composited content (mask + SurfaceView already applied) and a strip of newly captured raw window content. Visually, a horizontal seam in the replay frame.
* **Mask shift / partial mask** — if the next `PixelCopy.request` overwrites window pixels while `MaskRenderer.renderMasks` is still rendering, masks can land on stale pixels or in the wrong region for one frame.
* **SurfaceView ghosting** — composited SurfaceView pixels from frame N can persist into frame N+1 in regions the new PixelCopy didn't clear before the executor's recycle ran.

Native crashes (likely related to [#4696](<https://github.com/getsentry/sentry-java/issues/4696>)):

* **SIGABRT in** `libhwui.so` with the typical `__android_log_assert` → `art::Runtime::Abort` chain. PixelCopy runs on the RenderThread on top of libhwui; when its hardware blit lands on a `Bitmap` whose backing hardware buffer is being concurrently drawn into by a Canvas on the executor, libhwui's `LOG_ALWAYS_FATAL`/`CHECK` invariants fail and the process aborts.
[...]

Issue #4154, opened 2025-02-10
Replay: internal screenshot format


vaind:
### Description

I'm switching Flutter to using `onScreenshotRecorded(Bitmap)` and I've noticed a degradation in visual quality. I believe it is because the original code, which used the `File` based API, saved the screenshot to PNG, while the native Bitmap implementation uses JPG.

Previously, with Flutter generated PNG:  https://sentry-sdks.sentry.io/replays/2e9baae198384b6387e4e3f497070382/?project=5428562&query=release%3Aoss.krtirtho.spotube.dev%403.9.0-dev%2B38&referrer=%2Freplays%2F%3AreplaySlug%2F&statsPeriod=90d&yAxis=count%28%29&t=0

New, with Android JPG from Bitmap: https://sentry-sdks.sentry.io/replays/0a8ff56225b74dbf93b519043224fe54/?project=5428562&query=release%3Aoss.krtirtho.spotube.dev%403.9.0-dev%2B38&referrer=%2Freplays%2F%3AreplaySlug%2F&statsPeriod=90d&t_main=tags&yAxis=count%28%29&t=2

You can clearly see JPG-specific artifacts in the new version. Interestingly, the new version also results in a video that is larger than previously, at the same resolution, although some of that can be attributed to a slightly different content because the testing app uses random images in button backgrounds.

Have you considered PNG as the screenshot format when saving to file? If not, I'd suggest we evaluate this in from the performance perspective, as the quality should be a no-brainer.

Issue #2093, opened 2022-06-13
Build sample app against AGP matrix on CI


romtsn:
### Description

Similar to https://github.com/getsentry/sentry-android-gradle-plugin/blob/1deb7a6da144fdf6e55c86634afa9387132ec80c/.github/workflows/test-publish-dry-run.yaml#L15-L40

This would help catch regressions with preview versions of AGP, similar to this one https://github.com/getsentry/sentry-android-gradle-plugin/pull/321#partial-pull-merging

PR #5386, opened 2026-05-08
feat(replay): Add ReplayFrameObserver for snapshot testing

runningcode:
## Summary

- Add an experimental `ReplayFrameObserver` functional interface that fires right before a replay frame is stored to disk, receiving the masked bitmap, timestamp, and screen name
- Expose `snapshotObserver` on `ReplayIntegration` so callers can set it after SDK init
- Copy the bitmap before passing it to the observer so consumers own the copy and don't need to worry about the replay system's internal bitmap lifecycle
- Add `ReplaySnapshotTest` UI integration test that captures masked replay frames on Sauce Labs
- Enable `useTestStorageService` and `*.png` artifact collection in the Sauce Labs config
- Move `ReplaySnapshotTest` to a conditional `androidTestReplay` source set so it compiles only when `APPLY_SENTRY_INTEGRATIONS=true`

Relates to JAVA-504


sentry-android-integration-tests/sentry-uitest-android/src/androidTestReplay/java/io/sentry/uitest/android/ReplaySnapshotTest.kt
```diff
@@ -0,0 +1,74 @@
...
+import org.hamcrest.CoreMatchers.`is`
+import org.junit.Assume.assumeThat
+import org.junit.Before
+
+class ReplaySnapshotTest : BaseUiTest() {
+
+  @Before
+  fun setup() {
+    // GH Actions emulators don't support capturing screenshots for replay
+    @Suppress("KotlinConstantConditions")
```
runningcode, line 24: I copied this from `ReplayTest` but why are we even running these on emulators in gh actions?

sentry/src/main/java/io/sentry/SentryReplayOptions.java
```diff
@@ -17,6 +17,30 @@
...
+   * callback receives the frame after masking has been applied.
+   *
+   * <p>The frame bitmap is passed via a {@link Hint} using the key {@link
+   * TypeCheckHint#REPLAY_FRAME_BITMAP}. On Android, retrieve it with: {@code hint.getAs(
+   * TypeCheckHint.REPLAY_FRAME_BITMAP, Bitmap.class)}.
+   *
+   * <p>The callback runs on a background thread (replay executor). Do not recycle the bitmap — it
+   * may be reused by the replay system.
+   */
+  @ApiStatus.Experimental
```
markushi: IMHO I would mark this as `@ApiStatus.Internal` instead, and if we see demand, we could open it up.
```suggestion
  @ApiStatus.Internal
```

sentry/src/main/java/io/sentry/SentryReplayOptions.java
```diff
@@ -17,6 +17,30 @@
 
 public final class SentryReplayOptions extends SentryMaskingOptions {
 
+  /**
+   * Callback that is invoked right before a replay frame is stored to disk. This allows
+   * intercepting frames for testing (e.g., screenshot comparison tests) or custom processing. The
+   * callback receives the frame after masking has been applied.
+   *
+   * <p>The frame bitmap is passed via a {@link Hint} using the key {@link
```
markushi: 💯

sentry-android-integration-tests/sentry-uitest-android/src/androidTest/java/io/sentry/uitest/android/ReplaySnapshotTest.kt
```diff
@@ -0,0 +1,74 @@
...
+      File(
+          Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DOWNLOADS),
+          "sauce_labs_custom_screenshots",
+        )
+        .apply {
+          deleteRecursively()
+          mkdirs()
+        }
+    val frameReceived = CountDownLatch(1)
+    val capturedScreens = CopyOnWriteArrayList<String>()
```
markushi: not that it matters too much for tests, but that could be a set as well, right?
```suggestion
    val capturedScreens = mutableSetOf<String>()
```

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
```diff
@@ -308,6 +311,16 @@ public class ReplayIntegration(
...
+      val callback = options.sessionReplay.beforeStoreFrame
+      if (callback != null) {
+        try {
+          val hint = Hint()
+          hint.set(TypeCheckHint.REPLAY_FRAME_BITMAP, bitmap)
+          callback.execute(hint, frameTimeStamp, screen)
+        } catch (e: Throwable) {
+          options.logger.log(ERROR, "Error in beforeStoreFrame callback", e)
+        }
+      }
```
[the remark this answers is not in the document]
markushi: Correct, but that's already tracked here: https://github.com/getsentry/sentry-java/issues/5340

markushi (dismissed):
LGTM! Let's discuss first if the API should be experimental for internal, but other than that no blockers.

sentry/src/main/java/io/sentry/SentryReplayOptions.java
0xadam-brown: h: Unless someone has asked for this callback, my vote would be for Internal rather than Experimental to avoid a footgun.

Eg, looks like we're hoping for read-only use cases, but we don't protect against writes, meaning users could modify the screens we save for session replay. And we lock on the bitmap after handing it off, meaning deadlock becomes a possibility (at least analytically). We could hand off a copy but I imagine we'd still want to recycle it after the callback completes to avoid memory issues, and that means warning the user accordingly. Etc.  

Thoughts?

sentry/src/main/java/io/sentry/SentryReplayOptions.java
```diff
@@ -17,6 +17,30 @@
...
+   * TypeCheckHint#REPLAY_FRAME_BITMAP}. On Android, retrieve it with: {@code hint.getAs(
+   * TypeCheckHint.REPLAY_FRAME_BITMAP, Bitmap.class)}.
+   *
+   * <p>The callback runs on a background thread (replay executor). Do not recycle the bitmap — it
+   * may be reused by the replay system.
+   */
+  @ApiStatus.Experimental
+  public interface BeforeStoreFrameCallback {
```
0xadam-brown: l: Thoughts about keeping the nomenclature in line with our "screenshot recorded" language?

Eg, renaming `BeforeStoreScreenshotCallback` (if we want to guarantee the callback is fired immediately before writing to disk), or `ScreenshotRecordedCallback` (if we want to keep our options open about exactly when we invoke it) .

sentry/src/main/java/io/sentry/SentryReplayOptions.java
```diff
@@ -17,6 +17,30 @@
...
+  /**
+   * Callback that is invoked right before a replay frame is stored to disk. This allows
+   * intercepting frames for testing (e.g., screenshot comparison tests) or custom processing. The
+   * callback receives the frame after masking has been applied.
+   *
+   * <p>The frame bitmap is passed via a {@link Hint} using the key {@link
+   * TypeCheckHint#REPLAY_FRAME_BITMAP}. On Android, retrieve it with: {@code hint.getAs(
+   * TypeCheckHint.REPLAY_FRAME_BITMAP, Bitmap.class)}.
+   *
+   * <p>The callback runs on a background thread (replay executor). Do not recycle the bitmap — it
```
0xadam-brown: l: We could beef this up to protect against misuse if we think it'd help: 

```kotlin
 * <p>**Bitmap lifecycle:** The bitmap is owned by the replay system. It should be deemed 
 * read-only and valid solely for the duration of the callback. Do not write to it or recycle it. Do
 * not synchronize on it, store a reference to it, or access it after this method returns – copy
 * the pixel data (e.g., compress to a file) within this method if you need it later.
 *
 * <p>The callback runs on a background thread (the replay executor).
```

...or whatever makes sense.

0xadam-brown (commented):
Thanks for this @runningcode. A few comments. 

Big picture, did we think about defining a new `ReplayScreenshotObserver` interface inside the replay module (akin to `ScreenshotRecorderCallback`) rather than routing through the (universally available) `SentryReplayOptions`? 

An interface in the replay module would let us avoid the `Hint` indirection, and it'd avoid the tension of putting `@ApiStatus.Internal` on an `*Options` member. 

Thoughts?

sentry/src/main/java/io/sentry/SentryReplayOptions.java
0xadam-brown: ...looks like I was late to this game and we've updated to internal - good deal. (We should update the other sites to match 👍 )

sentry-android-integration-tests/sentry-uitest-android/src/androidTest/java/io/sentry/uitest/android/ReplaySnapshotTest.kt
runningcode: I switched it to a `CopyOnWriteArraySet` the reason is that the `ReplaySnapshotObserver` runs on a background thread.
It isn't strictly necessary here but it gives us a bit of safety.

sentry-android-replay/build.gradle.kts
```diff
@@ -61,6 +61,8 @@ android {
 
   buildFeatures { buildConfig = true }
 
+  configurations.all { resolutionStrategy.force(libs.jetbrains.annotations.get()) }
```
runningcode: I need to dig deeper in to why we need to add this to all the modules that add `libs.jetbrains.annotations` another day.

sentry-android-replay/src/main/java/io/sentry/android/replay/SessionReplayOptions.kt
```diff
@@ -29,3 +31,18 @@ public var SentryReplayOptions.maskAllImages: Boolean
...
+ * Observer that is notified when a replay snapshot is captured. The snapshot bitmap has masking
+ * already applied.
+ *
+ * **Bitmap lifecycle:** The bitmap is owned by the replay system and may be reused. Do not store a
+ * reference to it or access it after this method returns — copy the pixel data (e.g., compress to a
+ * file) within this method if you need it later. Do not recycle the bitmap.
+ *
+ * The callback runs on a background thread (the replay executor).
+ */
+@ApiStatus.Experimental
```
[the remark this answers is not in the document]
runningcode: we had a slack discussion where we decided it should be experimental since customers use it

sentry/src/main/java/io/sentry/SentryReplayOptions.java
runningcode: Just to close the loop on this, we discussed in slack that a customer asked for this, therefore it will be kept as experimental.

sentry/src/main/java/io/sentry/SentryReplayOptions.java
runningcode: Sounds good, updated to use `ReplayScreenshotObserver` as you suggested!

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
```diff
@@ -313,10 +313,14 @@ public class ReplayIntegration(
...
-          observer.onSnapshotCaptured(bitmap, frameTimeStamp, screen)
-        } catch (e: Throwable) {
-          options.logger.log(ERROR, "Error in ReplaySnapshotObserver", e)
+        val copy = bitmap.copy(bitmap.config!!, false)
+        if (copy != null) {
+          try {
+            observer.onSnapshotCaptured(copy, frameTimeStamp, screen)
+          } catch (e: Throwable) {
+            options.logger.log(ERROR, "Error in ReplaySnapshotObserver", e)
+            copy.recycle()
```
runningcode, line 325: I would not have thought of calling copy.recycle() if we get an exception here. AI is smart.

sentry-android-integration-tests/sentry-uitest-android/src/androidTestReplay/java/io/sentry/uitest/android/ReplaySnapshotTest.kt
```diff
@@ -52,6 +52,7 @@ class ReplaySnapshotTest : BaseUiTest() {
         val file = File(snapshotsDir, "${name}_$frameTimestamp.png")
         file.outputStream().use { out -> bitmap.compress(Bitmap.CompressFormat.PNG, 100, out) }
       }
+      bitmap.recycle()
```
runningcode: with the bitmap copy, we now have to make sure that we call `bitmap.recycle()` as should consumers of our API.

sentry/src/main/java/io/sentry/SentryReplayOptions.java
runningcode: i've updated it to copy the bitmap. let me know if you think the new docs make sense along with that change to copy it

sentry-android-integration-tests/sentry-uitest-android/src/androidTestReplay/java/io/sentry/uitest/android/ReplaySnapshotTest.kt
```diff
@@ -0,0 +1,68 @@
...
+
+  @Before
+  fun setup() {
+    // GH Actions emulators don't support capturing screenshots for replay
+    @Suppress("KotlinConstantConditions")
+    assumeThat(BuildConfig.ENVIRONMENT != "github", `is`(true))
+  }
+
+  @Test
+  fun captureComposeReplayFrameSnapshots() {
```
runningcode, line 29: just FYI, this is a pretty contrived test but it just makes sure we can capture a replay using the new snapshot observer api

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
```diff
@@ -122,6 +123,8 @@ public class ReplayIntegration(
   private val lifecycleLock = AutoClosableReentrantLock()
   private val lifecycle = ReplayLifecycle()
 
+  @Volatile public var snapshotObserver: ReplaySnapshotObserver? = null
```
runningcode: this awkward API is the consequence of moving the API to the `sentry-andorid-replay` api

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
```diff
@@ -308,6 +311,18 @@ public class ReplayIntegration(
     var screen: String? = null
     scopes?.configureScope { screen = it.screen?.substringAfterLast('.') }
     captureStrategy?.onScreenshotRecorded(bitmap) { frameTimeStamp ->
+      val observer = snapshotObserver
+      if (observer != null) {
+        val copy = bitmap.copy(bitmap.config!!, false)
```
runningcode, line 317: we copy the bitmap now

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
```diff
@@ -308,6 +311,18 @@ public class ReplayIntegration(
...
+        if (copy != null) {
```
runningcode, line 318: no point in calling the API if we have a null bitmap

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
0xadam-brown: ❗- Ah, I thought we were going the `SentryReplayOptions` route to avoid allowing folks to mutate configs after `Sentry.init()` returns (+ for enhanced visibility and better ergonomics)?

0xadam-brown: `@Experimental`

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
0xadam-brown, line 317: We should have our `try` block below enclose `bitmap.copy()`, esp given the `!!` 👍

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
```diff
@@ -308,6 +311,18 @@ public class ReplayIntegration(
...
+          try {
+            observer.onSnapshotCaptured(copy, frameTimeStamp, screen)
+          } catch (e: Throwable) {
+            options.logger.log(ERROR, "Error in ReplaySnapshotObserver", e)
+            copy.recycle()
+          }
+        }
```
0xadam-brown: Do we want a `finally` block where we call `copy.recycle()` on behalf of our users? 

That'd prevent accidental OOMs, but at the cost of the user not being able to use the bitmap after the callback returns. 

No strong opinions.

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
```diff
@@ -308,6 +311,18 @@ public class ReplayIntegration(
...
+      }
       addFrame(bitmap, frameTimeStamp, screen)
     }
```
0xadam-brown, line 330: ❗- It's not in the diff, so apologies for missing it the first time round, but I see there's an `onScreenshotRecorded(File, Long)` method as well as the `Bitmap` version above. I imagine we need to call the observer in both....

sentry-android-replay/src/main/java/io/sentry/android/replay/SessionReplayOptions.kt
```diff
@@ -29,3 +31,17 @@ public var SentryReplayOptions.maskAllImages: Boolean
...
   set(value) = setMaskAllImages(value)
+
+/**
+ * Observer that is notified when a replay snapshot is captured. The snapshot bitmap has masking
+ * already applied.
+ *
+ * **Bitmap lifecycle:** The bitmap is a copy owned by the caller. You may store it or use it on
+ * another thread. Call [Bitmap.recycle] when you no longer need it to free native memory promptly.
+ *
+ * The callback runs on a background thread (the replay executor).
```
0xadam-brown: Worth emphasizing that folks should keep processing quick or hand off to another thread. (Otherwise they'll slow down the SDK's replay pipeline.)

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
0xadam-brown, line 317: Newbie question, but are we able to observe the performance hit of copying on our end? 

Esp if we want to expand our use-cases beyond testing or debugging at some point, it'd be good to know whether we need to optimize.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Features
 
 - Add support to configure reporting historical ANRs via `AndroidManifest.xml` using the  `io.sentry.anr.report-historical` attribute ([#5387](https://github.com/getsentry/sentry-java/pull/5387))
+- Session Replay: Add `ReplaySnapshotObserver` for observing captured replay frames ([#5386](https://github.com/getsentry/sentry-java/pull/5386))
```
0xadam-brown: Worth noting that this is for use in snapshot tests and for debugging purposes and isn't currently optimized for production use.

0xadam-brown (commented):
Thx for the updates! A few comments (two more impactful one's I've flagged with ❗) – let me know if you have any questions.

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
runningcode: Oh, then I misunderstood, this is the same API as in your branch which is what I thought we agreed to: https://github.com/getsentry/sentry-java/commit/a97de37c59614341e5b4c4ccd406b120bc50cfa5#diff-cb8ccc45cef66aca485d205c9e99c17c114b6319ab9b7d52173b52ee72b047a8R126

We just need to make it `public` here instead of private in your branch because it needs to be usable from other packages i.e. other apps.

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
runningcode: I think the only types of errors that can happen when copying the bitmap are memory errors and we shouldn't try to catch those.
As for the `!!` I don't see how bitmap.config can be null if it is a valid bitmap. Do you know how it could be null?

runningcode: Memory-wise, each bitmap is roughly 900kb and we produce one frame per second.

sentry-android-integration-tests/sentry-uitest-android/src/androidTestReplay/java/io/sentry/uitest/android/ReplaySnapshotTest.kt
romtsn: i think the purpose of this tests running in gh actions is split in two parts:
- Run our SDK on a test app which uses newer AGP versions to catch issues like https://github.com/getsentry/sentry-java/issues/2093 early
- Compile the test app against only `sentry-android-core` without integrations to catch issues like https://github.com/getsentry/sentry-java/issues/2738 early

That said, we probably don't even need to run the app, just compiling it would be sufficient to catch these two potential issues. Actually running the tests was a nice addition to also verify the SDK behaviour at runtime, considering the two things above (e.g. if R8 strips out some code over-aggressively it'd crash at runtime)

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
romtsn: javadoc says:
> If the bitmap's internal config is in one of the public formats, return that config, otherwise return null.

that said, I think in our case it will never be null since we always use `ARGB_8888`

romtsn: 900kb sounds like a lot, last time check it was way less 😅 I guess it depends on complexity of app's UI as well as the device used. We'll probably have to look into this as part of SDK Perf overhead initiative, and we even have an issue for that: https://github.com/getsentry/sentry-java/issues/4154. I'd imagine we could use webp without losing anything, but that's for later

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
romtsn: yep, that one is used by Flutter exclusively, but would be good to align here and also call the observer 👍

sentry-android-replay/build.gradle.kts
```diff
@@ -71,6 +73,7 @@ kotlin { explicitApi() }
 dependencies {
   api(projects.sentry)
 
+  compileOnly(libs.jetbrains.annotations)
```
romtsn: not sure if I missed something, but this one turns out to be unused?

CHANGELOG.md
romtsn: agree, and could we also add a short code snippet showcasing the usage? (in Kotlin would be enough)

sentry/src/main/java/io/sentry/SentryReplayOptions.java
```diff
@@ -36,6 +36,28 @@ public interface BeforeErrorSamplingCallback {
...
+   * applied) is passed via a {@link Hint} using the key {@link TypeCheckHint#REPLAY_FRAME_BITMAP}.
+   *
+   * <p>On Android, retrieve the bitmap with: {@code hint.getAs(TypeCheckHint.REPLAY_FRAME_BITMAP,
+   * Bitmap.class)}.
+   *
+   * <p>The callback runs on a background thread (replay executor). The bitmap is a copy owned by
+   * the caller. Call {@code Bitmap.recycle()} when done to free native memory.
+   */
+  @ApiStatus.Experimental
+  public interface ReplaySnapshotObserver {
```
romtsn: Not sure, but the word "Snapshot" may be confusing, but I also don't really have a better word for it 😅  (other than "frame" or "screenshot", which are also not the best)

sentry/src/main/java/io/sentry/SentryReplayOptions.java
```diff
@@ -36,6 +36,28 @@ public interface BeforeErrorSamplingCallback {
...
+    /**
+     * Called when a replay snapshot is captured.
```
romtsn: I guess important to note here that it's already a masked snapshot/frame. This could potentially read like "captured, but not yet masked" (especially, if we introduce another callback in the future, like onCapturedButNotMasked", like in https://github.com/getsentry/sentry/issues/86905)

CHANGELOG.md
runningcode: Do you mean a snippet here in the CHANGELOG ?

sentry-android-replay/build.gradle.kts
runningcode: It is used for the `@ApiStatus.Experimental` annotation.

sentry/src/main/java/io/sentry/SentryReplayOptions.java
runningcode: Actually this is a good point, since we are using snapshots for a different produce. I like frame better.

CHANGELOG.md
romtsn: yes! you can find some example down there in the already released versions

sentry-android-replay/build.gradle.kts
romtsn: but that one is in `sentry` and not in `sentry-android-replay`?

runningcode: Oh i see, this is leftover from a previous refactoring! thanks!

romtsn (approved):
LGTM. I think would be still good to call `onMaskedFrameCaptured` in the other `onScreenshotRecorded()`

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
runningcode: Ok added below. Note, we read the file and decode it as a bitmap.

PR #5388, opened 2026-05-08
Remove testBuildType mechanism from UI test modules

runningcode:
## Summary
- Disable the debug variant unconditionally in `sentry-uitest-android` and `sentry-uitest-android-benchmark` since debug and release were configured identically (same minification, proguard rules, signing)
- Hardcode `testBuildType = "release"` and remove the `-DtestBuildType=release` system property mechanism
- Combine separate Gradle invocations into single commands in the Makefile and docs

#skip-changelog

### DIFF
--- a/.github/workflows/agp-matrix.yml
+++ b/.github/workflows/agp-matrix.yml
@@ -90,7 +90,7 @@ jobs:
           disable-spellchecker: true
           emulator-options: -no-snapshot-save -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
           disk-size: 4096M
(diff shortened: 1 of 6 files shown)


0xadam-brown (approved):
Claude tells me: 

> The debug and release build types in this PR's two uitest modules were configured identically (same minification, same proguard rules, same signing config), and every
  single call site always passed -DtestBuildType=release to override the default. The System.getProperty("testBuildType", "debug") mechanism gave the appearance of a configurable choice, but in practice nobody ever used it as "debug" — it was dead flexibility.
> 
> The PR removes that indirection so the build config matches how it was actually used.

Good deal 👍

runningcode:
Yup that's correct. In addition, it also increases build times and cache sizes because if you switch variants then you have to rebuild it.

PR #5383, opened 2026-05-07
feat(core): Session based Traces for Mobile

adinauer:
## PR Stack (Session based Traces for Mobile)

- #5383
- #5398
- #5402
- #5417
- #5429
- #5430

---

Collection PR for the Session based Traces for Mobile stack. Squash-merge this once all stack PRs are merged.

PR #5398, opened 2026-05-11
feat(android): [Session based Traces for Mobile 1B] Add session trace lifecycle

adinauer:

- #5383
- #5398
- #5402
- #5417
- #5429
- #5430

---

## :scroll: Description
Adds opt-in Android session trace lifecycle support and simplifies the trace model by keeping session lifecycle behavior at the option/transaction level instead of storing lifecycle state on `PropagationContext`.

This PR:
- adds `enableSessionTraceLifecycle` configuration and Android wiring
- starts a new propagation context when Android starts a new session
- remaps eligible root transactions onto the current session trace
- preserves non-trace span/transaction context state while replacing only the trace id and clearing the parent span id
- keeps continued traces with a parent span out of session trace remapping
- removes the temporary propagation context lifecycle mode and `forceNewTrace` transaction flag

Session-based traces on mobile need independently started root transactions to join the ambient app-session trace. Keeping this behavior behind an SDK option and deriving the opt-out from `parentSpanId` avoids per-context lifecycle state while preserving bounded/continued distributed traces.

This is the new first PR in the Session based Traces for Mobile stack: `1B`.

- `./gradlew :sentry:compileJava :sentry:compileTestKotlin :sentry-opentelemetry:sentry-opentelemetry-core:compileJava`
- `./gradlew :sentry:test --tests 'io.sentry.ScopesTest' --tests 'io.sentry.ScopeTest' --tests 'io.sentry.util.TracingUtilsTest' :sentry-opentelemetry:sentry-opentelemetry-core:test`
- `./gradlew :sentry:test --tests 'io.sentry.ScopesTest.when session trace lifecycle is enabled, root transaction uses current propagation context' --tests 'io.sentry.ScopesTest.forceNewTrace does not override continued trace with parent span' spotlessApply apiDump`
- `./gradlew :sentry:compileJava :sentry:compileTestKotlin`
- `./gradlew :sentry:test --tests 'io.sentry.TransactionContextTest'`
- `./gradlew :sentry:compileJava :sentry:compileTestKotlin :sentry:test --tests 'io.sentry.ScopesTest.continued trace with parent span is not remapped to session trace' --tests 'io.sentry.TransactionContextTest'`

## :pencil: Checklist

## :crystal_ball: Next steps

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/SentryGestureListener.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/SentryGestureListener.java
@@ -203,7 +203,7 @@ public final class SentryGestureListener implements GestureDetector.OnGestureLis
 
     if (!(options.isTracingEnabled() && options.isEnableUserInteractionTracing())) {
       if (isNewInteraction) {
-        if (options.isEnableAutoTraceIdGeneration()) {
+        if (options.isEnableAutoTraceIdGeneration() && !options.isEnableSessionTraceLifecycle()) {
           TracingUtils.startNewTrace(scopes);
         }
         activeUiElement = target;
--- a/sentry-android-core/src/test/java/io/sentry/android/core/ActivityLifecycleIntegrationTest.kt
+++ b/sentry-android-core/src/test/java/io/sentry/android/core/ActivityLifecycleIntegrationTest.kt
@@ -1414,6 +1414,30 @@ class ActivityLifecycleIntegrationTest {
   }
 
   @Test
+  fun `does not start a new trace if performance is disabled and session trace lifecycle is enabled`() {
+    val sut = fixture.getSut()
+    val activity = mock<Activity>()
+    fixture.options.tracesSampleRate = null
+    fixture.options.isEnableAutoTraceIdGeneration = true
+    fixture.options.isEnableSessionTraceLifecycle = true
+
+    val argumentCaptor: ArgumentCaptor<ScopeCallback> =
+      ArgumentCaptor.forClass(ScopeCallback::class.java)
+    val scope = Scope(fixture.options)
+    val propagationContextAtStart = scope.propagationContext
+    whenever(fixture.scopes.configureScope(argumentCaptor.capture())).thenAnswer {
+      argumentCaptor.value.run(scope)
+    }
+
+    sut.register(fixture.scopes, fixture.options)
+    sut.onActivityCreated(activity, fixture.bundle)
+
+    // once for the screen
+    verify(fixture.scopes).configureScope(any())
+    assertSame(propagationContextAtStart, scope.propagationContext)
+  }
+
+  @Test
   fun `sets the activity as the current screen`() {
     val sut = fixture.getSut()
     val activity = mock<Activity>()
--- a/sentry/src/main/java/io/sentry/SentryOptions.java
+++ b/sentry/src/main/java/io/sentry/SentryOptions.java
@@ -1436,8 +1436,8 @@ public class SentryOptions {
   }
 
   /**
-   * Enables or disables session trace lifecycle. When enabled, root transactions can reuse the
-   * current session propagation context unless they force a new trace.
+   * Enables or disables session trace lifecycle. When enabled, root transactions without a parent
+   * span can reuse the current session propagation context.
    *
    * <p>This option is intended for Android/mobile SDKs where trace boundaries are managed by the
    * SDK session lifecycle. Do not enable it for JVM backend, desktop, or other non-session-managed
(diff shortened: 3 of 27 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/SentryGestureListener.java
```diff
@@ -203,7 +203,7 @@ private void startTracing(final @NotNull UiElement target, final @NotNull Gestur
 
     if (!(options.isTracingEnabled() && options.isEnableUserInteractionTracing())) {
       if (isNewInteraction) {
-        if (options.isEnableAutoTraceIdGeneration()) {
+        if (options.isEnableAutoTraceIdGeneration() && !options.isEnableSessionTraceLifecycle()) {
```
buenaflor, line 206: imo it's a good option name but I think maybe we should call it a bit differently since it clashes with span first's `traceLifecycle` option

maybe just `isEnableSessionTraces`? wdyt

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -290,6 +290,9 @@ public class SentryOptions {
   /** Whether to enable or disable automatic session tracking. */
   private boolean enableAutoSessionTracking = true;
 
+  /** Whether root transactions should reuse the current session trace lifecycle. */
```
buenaflor, line 293: imo this should be more generic since trace re-use  is not only specific to spans/transactions, wdyt?

```suggestion
  /** Whether telemetry should reuse the current session trace lifecycle. */
```

sentry-android-core/src/test/java/io/sentry/android/core/ActivityLifecycleIntegrationTest.kt
```diff
@@ -1413,6 +1413,30 @@ class ActivityLifecycleIntegrationTest {
...
+    }
+
+    sut.register(fixture.scopes, fixture.options)
+    sut.onActivityCreated(activity, fixture.bundle)
+
+    // once for the screen
+    verify(fixture.scopes).configureScope(any())
+    assertSame(propagationContextAtStart, scope.propagationContext)
+  }
+
```
buenaflor, line 1439: I don't think this should be tied to performance since you'd wanna have telemetry still be session connected aside from spans e.g errors, logs

e.g Flutter and Browser don't gate this behind performance

sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/SentryGestureListener.java
adinauer: Sure, I'll change the name.

adinauer: Changed in https://github.com/getsentry/sentry-java/pull/5430

sentry/src/main/java/io/sentry/SentryOptions.java
adinauer: It does only affect head of trace transactions (no `parentSpanID`) so the current version sounds more correct to me.

buenaflor: which traceId do errors, metrics, logs inherit then?

adinauer: Those should take `traceId` from the active transaction if there is any on scope. But since that should be identical to the scopes `traceId` the distinction would only matter for non head of trace transactions where the new behaviour doesn't kick in.

Probably simplest to change the wording to what you suggested rather than explain SDK internals here.

sentry-android-core/src/test/java/io/sentry/android/core/ActivityLifecycleIntegrationTest.kt
adinauer: This change makes sure we don't generate new traces for each Activity if performance is disabled but `enableSessionTraces` is enabled. The alternative would be starting a new trace on each session plus one trace per Activity.

sentry/src/main/java/io/sentry/SentryOptions.java
buenaflor: ah okay, btw should we scope this option to the android sdk? (not sure about the protocol here when we add new options)

adinauer: Since this is used in core SDK, we can't move it to `AndroidOptions`. We can add a comment to the JavaDoc to explain this is only intended for Android. In docs we can simply not mention it for JVM.

adinauer: In the future (new major) we'll likely make this default enabled in `AndroidOptionsInitializer`.

sentry-android-core/src/test/java/io/sentry/android/core/ActivityLifecycleIntegrationTest.kt
buenaflor: ah I misunderstood, all good

sentry/src/main/java/io/sentry/SentryOptions.java
buenaflor: makes sense 👍

adinauer:

PR #5402, opened 2026-05-11
fix(core): [Session based Traces for Mobile 2] Keep transaction baggage mutable

adinauer:

- #5383
- #5398
- #5402
- #5417
- #5429
- #5430

---

## :scroll: Description
Keeps transaction baggage mutable when remapping a root transaction onto the current session trace. This lets each transaction populate its own DSC fields even if the ambient session propagation baggage was already frozen by an earlier outgoing request without an active transaction.

Adds regression coverage for:
- starting a new session
- producing outgoing headers without an active transaction, freezing the session baggage
- starting a transaction in the session trace and verifying outgoing transaction baggage
- finishing that transaction, then starting a second transaction and verifying its baggage too

Session trace lifecycle can freeze ambient scope baggage before any transaction starts. If a later transaction reuses that frozen baggage directly, outgoing transaction propagation can miss transaction-specific DSC such as `sentry-transaction`, `sentry-sampled`, `sentry-sample_rate`, and `sentry-sample_rand`.

This keeps the session trace id and sample rand while allowing each remapped transaction to complete its own outgoing baggage.

- `./gradlew :sentry:test --tests 'io.sentry.BaggageTest' --tests 'io.sentry.ScopesTest.session trace transaction baggage is populated after scope baggage is frozen'`

## :pencil: Checklist

## :crystal_ball: Next steps

#skip-changelog

### DIFF
--- a/sentry/src/main/java/io/sentry/Baggage.java
+++ b/sentry/src/main/java/io/sentry/Baggage.java
@@ -233,8 +233,8 @@ public final class Baggage {
         baggage.sampleRate,
         sampleRand,
         baggage.thirdPartyHeader,
-        baggage.mutable,
-        baggage.shouldFreeze,
+        true,
+        false,
         baggage.logger);
   }
 
--- a/sentry/src/test/java/io/sentry/ScopesTest.kt
+++ b/sentry/src/test/java/io/sentry/ScopesTest.kt
@@ -21,6 +21,7 @@ import io.sentry.test.createTestScopes
 import io.sentry.test.initForTest
 import io.sentry.util.HintUtils
 import io.sentry.util.StringUtils
+import io.sentry.util.TracingUtils
 import java.io.File
 import java.nio.file.Files
 import java.util.Queue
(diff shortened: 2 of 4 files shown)


sentry/src/test/java/io/sentry/ScopesTest.kt
```diff
@@ -1865,6 +1866,47 @@ class ScopesTest {
...
+    val sessionTraceId = AtomicReference<SentryId>()
+    val sessionSampleRand = AtomicReference<Double>()
+    val headersWithoutTransaction =
+      TracingUtils.traceIfAllowed(scopes, "https://sentry.io/hello", emptyList(), null)
+    assertNotNull(headersWithoutTransaction)
+    scopes.configureScope { scope ->
+      sessionTraceId.set(scope.propagationContext.traceId)
+      sessionSampleRand.set(scope.propagationContext.sampleRand)
+      assertFalse(scope.propagationContext.baggage!!.isMutable)
+    }
```
buenaflor, line 1888: should we also add a test that tracesSampler sampleRates are respected? not sure if that should be here or another PR

e.g tracesSampler: if transaction A: 0.8 else 0.2

- then a request associated with transaction A propagates transaction A sample_rate (0.8) in the baggage
- then a request associated with transaction B propagates transaction B sample_rate (0.2) in the baggage

buenaflor (approved):
looks good, only got a question about an additional test case

sentry/src/main/java/io/sentry/Baggage.java
```diff
@@ -233,8 +233,8 @@ public Baggage(final @NotNull Baggage baggage) {
         baggage.sampleRate,
         sampleRand,
         baggage.thirdPartyHeader,
-        baggage.mutable,
-        baggage.shouldFreeze,
+        true,
```
adinauer, line 236: Undoing this "unfreeze" change in https://github.com/getsentry/sentry-java/pull/5429 again since we can rely on `Baggage` still being mutable in head of trace scenario when creating a transaction.

adinauer:

PR #5417, opened 2026-05-12
fix(core): [Session based Traces for Mobile 3] Require active session for lifecycle

adinauer:

- #5383
- #5398
- #5402
- #5417
- #5429
- #5430

---

## :scroll: Description
Only remaps root transactions onto the session propagation context when there is an active session. If `enableSessionTraceLifecycle` is enabled before a session starts, root transactions keep their own transaction context instead of reusing the ambient scope trace.

Adds regression coverage for enabling session trace lifecycle without an active session, while preserving the existing behavior once a session is active.

Session trace lifecycle should be tied to an actual session lifecycle. Without this guard, enabling the option before a session exists can cause unrelated root transactions to share the ambient scope propagation context.

This still supports manually-created sessions because the behavior depends on whether a session is active, not whether auto session tracking is enabled.

- `./gradlew :sentry:test --tests "io.sentry.ScopesTest"`

## :pencil: Checklist

## :crystal_ball: Next steps

#skip-changelog

### DIFF
--- a/sentry/src/main/java/io/sentry/Baggage.java
+++ b/sentry/src/main/java/io/sentry/Baggage.java
@@ -221,21 +221,23 @@ public final class Baggage {
 
   @ApiStatus.Internal
   static @NotNull Baggage copyWithOverrides(
-      final @NotNull Baggage baggage,
+      final @Nullable Baggage baggage,
       final @NotNull SentryId traceId,
       final @Nullable Double sampleRand) {
+    final @NotNull Baggage source =
+        baggage == null ? new Baggage(NoOpLogger.getInstance()) : baggage;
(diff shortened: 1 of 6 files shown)


buenaflor (approved):
so if there's no active session then we'll just fallback to the previous mode (new transaction = new trace) right?

adinauer:
> so if there's no active session then we'll just fallback to the previous mode (new transaction = new trace) right?

@buenaflor yes exactly,`enableSessionTraceLifecycle` option needs to be enabled, can't have `parentSpanId` set for the transaction and a session must be there as well to trigger this new behaviour.

adinauer:

PR #5426, opened 2026-05-13
feat(core): Add API to clear scope feature flags

### LINKED ISSUE #5422: Add API to clear feature flag buffer
## Problem

When users switch accounts in-app, previously recorded feature flag evaluations from the old user persist in the buffer and get attached to error events for the new user. There is no public API to clear or reset the feature flag buffer.

**Example:**
- UserA has: `experimentA=control`, `experimentB=experiment`
- UserB has: `experimentB=experiment`, `experimentC=experiment`
- After switching from UserA → UserB, error events report: `experimentA=control`, `experimentB=experiment`, `experimentC=experiment`
- Expected: only UserB's flags (`experimentB=experiment`, `experimentC=experiment`)

### Why existing APIs don't work

- **`Scope.clear()`** does not reset the `featureFlags` field ([Scope.java#L564-L578](https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/Scope.java#L564))
- **`pushIsolationScope()`** clones the existing scope (including flags) via `isolationScope.clone()`, so stale flags carry over
- **`addFeatureFlag()`** updates existing entries by name but cannot remove flags that don't apply to the new user
- **`FeatureFlagBuffer`** and `IFeatureFlagBuffer` have no `clear()` method

### Proposed changes

1. Add `clear()` to `IFeatureFlagBuffer` and `FeatureFlagBuffer`
2. Include `featureFlags` in `Scope.clear()` (reset to a fresh buffer)
3. Optionally expose `clearFeatureFlags()` on `IScope` for targeted use

### Precedent

The Python SDK already supports this — `FlagBuffer.clear()` exists and `Scope.clear()` resets flags to `None` ([sentry-python feature_flags.py#L30](https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/feature_flags.py#L30), [scope.py#L756](https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/scope.py#L756)).

Action taken on behalf of Michael Chai.

adinauer:
## :scroll: Description
Add `IScope.clearFeatureFlags()` to clear feature flag evaluations stored on a scope.

This also makes `Scope.clear()` reset the feature flag buffer, matching the existing expectation that clearing a scope returns it to its default state.

Fixes #5422
Refs JAVA-512

Previously, feature flag evaluations could only be added or overwritten by name. There was no way to remove evaluations that no longer apply, which could leave stale feature flags attached to later error events after an app account switch.

- `./gradlew :sentry:test --tests 'io.sentry.ScopeTest' --tests 'io.sentry.featureflags.FeatureFlagBufferTest' --tests 'io.sentry.featureflags.SpanFeatureFlagBufferTest'`

## :pencil: Checklist

## :crystal_ball: Next steps
Docs can be updated separately if we want to document the new scope-level clearing API.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Features
 
+- Add API to clear feature flags from scopes ([#5426](https://github.com/getsentry/sentry-java/pull/5426))
 - Add support to configure reporting historical ANRs via `AndroidManifest.xml` using the  `io.sentry.anr.report-historical` attribute ([#5387](https://github.com/getsentry/sentry-java/pull/5387))
 
 ## 8.41.0
--- a/sentry/api/sentry.api
+++ b/sentry/api/sentry.api
@@ -273,6 +273,7 @@ public final class io/sentry/CombinedScopeView : io/sentry/IScope {
 	public fun clear ()V
 	public fun clearAttachments ()V
 	public fun clearBreadcrumbs ()V
+	public fun clearFeatureFlags ()V
 	public fun clearSession ()V
 	public fun clearTransaction ()V
 	public fun clone ()Lio/sentry/IScope;
@@ -901,6 +902,7 @@ public abstract interface class io/sentry/IScope {
 	public abstract fun clear ()V
 	public abstract fun clearAttachments ()V
 	public abstract fun clearBreadcrumbs ()V
+	public abstract fun clearFeatureFlags ()V
 	public abstract fun clearSession ()V
 	public abstract fun clearTransaction ()V
 	public abstract fun clone ()Lio/sentry/IScope;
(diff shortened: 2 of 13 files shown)


runningcode (approved):
LGTM!
Should we document either in the javadoc or in or docs when we expect customer to explicitly call this? i.e. when switching users or clearing data?

adinauer:
> Should we document either in the javadoc or in or docs when we expect customer to explicitly call this? i.e. when switching users or clearing data?

In theory they should already be calling `scope.clear()` when a user logs out, so feature flags are also automatically cleared now.

I can check whether we say something about clearing in docs.

adinauer:
Couldn't find explicit mention of clearFeatureFlags in other SDK docs. Most other things you can clear on scope directly also don't have direct mentions, so IMO it's fine to skip this.

PR #5428, opened 2026-05-13
feat(android): Parse memory and GC info from ANR thread dumps

### LINKED ISSUE #2801: Parse currently available memory from ANRv2 thread dump
### Description

The ANR thread dump actually contains what is the current memory size. size until OOME and GC metrics, which we could parse and send as part of contexts and even add it as suspect root cause if it's low.

```
Free memory 3107KB
Free memory until GC 3107KB
Free memory until OOME 187MB
Total memory 7592KB
Max memory 192MB
...
Total time waiting for GC to complete: 8.054ms
Total GC count: 1
Total GC time: 11.807ms
Total blocking GC count: 1
Total blocking GC time: 11.873ms
Total pre-OOME GC count: 0
```

markushi:
## :scroll: Description

Parses memory and GC metrics from ANRv2 dumps (ApplicationExitInfo traces) and attaches them as a new `art` context on ANR events.

New classes:
- `ArtContext` — serializable context for ART runtime memory/GC fields (bytes and milliseconds), added to `Contexts`
- `ArtContextParser` — parses lines like `Free memory 3107KB`, `Total GC time: 11.807ms` using simple suffix-based parsing (counterpart to Android's `PrettySize` and `PrettyDuration` output)

The parsed data is attached as `event.contexts.art` with attribute names aligned to [sentry-conventions#382](https://github.com/getsentry/sentry-conventions/pull/382).

Resolves: https://github.com/getsentry/sentry-java/issues/2801

ANR thread dumps from ART contain memory and GC statistics that are currently ignored. This information is valuable for diagnosing memory pressure-related ANRs.

- Unit tests
- Integration tests

## :pencil: Checklist


markushi:
@sentry review

sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
```diff
@@ -723,6 +724,36 @@ public void applyPostEnrichment(
...
+
+      if (memoryInfo.getFreeMemoryUntilOOMEBytes() != null) {
+        device.setFreeMemory(memoryInfo.getFreeMemoryUntilOOMEBytes());
+      }
+      if (memoryInfo.getFreeMemoryBytes() != null) {
+        device.setUsableMemory(memoryInfo.getFreeMemoryBytes());
+      }
+      if (memoryInfo.getMaxMemoryBytes() != null) {
+        device.setMemorySize(memoryInfo.getMaxMemoryBytes());
+      }
```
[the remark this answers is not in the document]
markushi: That's true, moved the parsed values to the new `ArtContext`

sentry-android-core/src/main/java/io/sentry/android/core/internal/threaddump/ArtContextParser.java
```diff
@@ -0,0 +1,126 @@
...
+  private static final String TOTAL_PRE_OOME_GC_COUNT_PREFIX = "Total pre-OOME GC count: ";
+
+  private @Nullable ArtContext artContext;
+
+  @Nullable
+  ArtContext getArtContext() {
+    return artContext;
+  }
+
+  void parseLine(final @NotNull String text) {
```
0xadam-brown, line 39: Do we have a source URL for the substrings we parse against that we could link to in a Javadoc? (eg, [here](https://android.googlesource.com/platform/art/+/master/runtime/gc/heap.cc#1282)?) That'd let folks know where to look if things change + where they can double-check semantics.

sentry-android-core/src/test/java/io/sentry/android/core/internal/threaddump/ArtContextParserTest.kt
```diff
@@ -0,0 +1,102 @@
...
+
+  @Test
+  fun `ignores unrelated lines`() {
+    val parser = ArtContextParser()
+    parser.parseLine("some random line")
+    parser.parseLine("DALVIK THREADS (29):")
+    parser.parseLine("")
+    assertNull(parser.artContext)
+  }
+}
```
0xadam-brown, line 130: I see our other context types (eg, `Device`, etc.) often have tests for their copy constructors (see, eg, `DeviceTest`, `GpuTest`, etc.). Worth adding one for `ArtContext` if we think the others were worthwhile 👍

sentry/src/test/resources/json/art_context.json
```diff
@@ -0,0 +1,13 @@
...
+  "gc_blocking_count": 1,
+  "gc_blocking_time": 11.873,
+  "gc_pre_oome_count": 0,
+  "gc_waiting_time": 8.054,
+  "free_memory": 3181568,
+  "free_memory_until_gc": 3181568,
+  "free_memory_until_oome": 196083712,
+  "total_memory": 7774208,
+  "max_memory": 201326592
+}
```
0xadam-brown, line 13: Could be nice to test the preservation of fields in the `unknown` map as well. (Up to you.)

sentry-android-core/src/main/java/io/sentry/android/core/internal/threaddump/ArtContextParser.java
```diff
@@ -0,0 +1,126 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+final class ArtContextParser {
+
+  private static final long KB = 1024;
+  private static final long MB = 1024 * KB;
+  private static final long GB = 1024 * MB;
+
+  private static final String FREE_MEMORY_PREFIX = "Free memory ";
```
0xadam-brown, line 19: Do we have a sense for how frequently these strings change, and how stable they are across OEMs? Just wondering how much we need to monitor the presence of unparseable / null values here vs our ability to set it and forget it. 

(Doesn't have to hold up this PR, regardless.)

sentry/src/main/java/io/sentry/protocol/ArtContext.java
```diff
@@ -0,0 +1,331 @@
...
+import io.sentry.util.Objects;
+import io.sentry.vendor.gson.stream.JsonToken;
+import java.io.IOException;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Context containing ART (Android Runtime) specific information. This is only relevant for Android
```
0xadam-brown, line 19: Any concerns about the fact that we have an Android-specific context living in the core sentry-java module rather than in sentry-android-core? 

(This is partly a newbie question about whether it's feasible + sensible to relocate to sentry-android-core, and I don't know the answer to that. An alternative could be to generalize the naming, but that has its own challenges given the specificity of this class's fields.)

sentry/src/main/java/io/sentry/protocol/ArtContext.java
```diff
@@ -0,0 +1,331 @@
...
+ * and may be null on other platforms.
+ */
+public final class ArtContext implements JsonUnknown, JsonSerializable {
+  public static final String TYPE = "art";
```
0xadam-brown, line 23: Newbie question: Would "runtime" make more sense as the type, with other platforms supplying their own `*Context` type? (or "runtime.art" if our conventions permit multipart keys)

"runtime" seems a better match for the abstraction level of things like "gpu" and "device", but perhaps Relay doesn't permit that sort of key-value "polymorphism" (ie, single key but different value types based on the platform).

0xadam-brown (commented):
Really nice. Just a few comments; none have to block us.

sentry/src/main/java/io/sentry/protocol/ArtContext.java
buenaflor: @0xadam-brown we do allow multipart keys and we generally should try to follow/extend OTEL conventions (if it fits) and only add our own stuff if there really is no equivalent (this hasn't always been the case but recently we decided to do enforce this).

In this case I suggest we do `art.gc.xyz` and `art.memory.xyz`. OTEL defines a generic `process.runtime.name/version/description` attribute but for runtime specifics they favour namespacing it under the runtime directly e.g

`dotnet.gc.*`, `jvm.gc.*` (we already do this for JVM)

buenaflor: jvm example: https://getsentry.github.io/sentry-conventions/attributes/jvm/

sentry/src/main/java/io/sentry/protocol/ArtContext.java
markushi: Good point, but yes, this is "fine" as-is - all protocol definitions live in the base module, as we also have use-cases where serialized envelopes from other SDKs get parsed and re-transmitted, so having (de)serialization supports is a good measure here.

sentry-android-core/src/main/java/io/sentry/android/core/internal/threaddump/ArtContextParser.java
markushi: Yeah, that's some good input! I'm not aware of any "flavors" across vendors. It's probably most practicable to have a look at some customer data as soon as the feature rolls out.

sentry-android-core/src/main/java/io/sentry/android/core/internal/threaddump/ArtContextParser.java
markushi: 👍 I've added the counterpart source links to the headers of `parsePrettySize` and `parseTimeMs`

sentry/src/main/java/io/sentry/protocol/ArtContext.java
0xadam-brown: Thanks for the helpful context @buenaflor – looks good to me 🥇

0xadam-brown (approved):
Thanks for the updates – lgtm 👍

runningcode (approved):
Is there a way we can avoid having the `ArtContext` in the java sdk?

markushi:
> Is there a way we can avoid having the `ArtContext` in the java sdk?

As of now we keep all protocol definitions within the core module, as the core SDK _could_ be used as a proxy and end up parsing / sending envelope files produced by other SDKs (like `sentry-native`)

PR #5429, opened 2026-05-13
fix(core): [Session based Traces for Mobile 4] No longer unfreeze Baggage

adinauer:

- #5383
- #5398
- #5402
- #5417
- #5429
- #5430

---

## :scroll: Description
Avoids reusing the session propagation context's frozen baggage when remapping root transactions onto the session trace. The remapped transaction now keeps its own transaction baggage as the source and only overrides the session `traceId` and `sampleRand`.

This preserves the session trace linkage without creating a mutable copy from already-frozen propagation baggage.

Session trace lifecycle needs root transactions to share the session trace ID and sampling random value. The previous approach did that by copying the scope propagation baggage as mutable transaction baggage, effectively unfreezing baggage that had already been frozen during header propagation.

Keeping the transaction baggage as the source avoids that unfreeze behavior while still reusing the session trace identifiers required for session-based traces.

- `./gradlew :sentry:test --tests='io.sentry.BaggageTest' --tests='io.sentry.TransactionContextTest'`

## :pencil: Checklist

## :crystal_ball: Next steps

#skip-changelog

### DIFF
--- a/sentry/src/main/java/io/sentry/Baggage.java
+++ b/sentry/src/main/java/io/sentry/Baggage.java
@@ -221,21 +221,23 @@ public final class Baggage {
 
   @ApiStatus.Internal
   static @NotNull Baggage copyWithOverrides(
-      final @NotNull Baggage baggage,
+      final @Nullable Baggage baggage,
       final @NotNull SentryId traceId,
       final @Nullable Double sampleRand) {
+    final @NotNull Baggage source =
+        baggage == null ? new Baggage(NoOpLogger.getInstance()) : baggage;
     final @NotNull ConcurrentHashMap<String, String> keyValues =
-        new ConcurrentHashMap<>(baggage.keyValues);
+        new ConcurrentHashMap<>(source.keyValues);
     keyValues.put(DSCKeys.TRACE_ID, traceId.toString());
 
     return new Baggage(
         keyValues,
-        baggage.sampleRate,
+        source.sampleRate,
         sampleRand,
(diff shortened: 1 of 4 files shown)


sentry/src/main/java/io/sentry/Baggage.java
```diff
@@ -221,21 +221,23 @@ public Baggage(final @NotNull Baggage baggage) {
...
         keyValues,
-        baggage.sampleRate,
+        source.sampleRate,
         sampleRand,
-        baggage.thirdPartyHeader,
-        true,
-        false,
-        baggage.logger);
+        source.thirdPartyHeader,
+        source.mutable,
```
adinauer, line 238: No longer unfreezing here since we can rely on `Baggage` still being mutable when creating a new transaction in head of trace scenario.

sentry/src/main/java/io/sentry/TransactionContext.java
```diff
@@ -44,7 +44,7 @@ public static TransactionContext fromPropagationContext(
       final @NotNull TransactionContext transactionContext) {
     final @NotNull Baggage baggage =
         Baggage.copyWithOverrides(
-            propagationContext.getBaggage(),
+            transactionContext.getBaggage(),
```
adinauer, line 47: Better to use the transactions own `Baggage` and copy over desired values from `Scope` instead of the other way around.

adinauer:

PR #5446, opened 2026-05-18
feat(tombstones): Add option to attach raw tombstone as protobuf

romtsn:
## :scroll: Description

Adds a new `attachRawTombstone` option (default `false`) that attaches the raw tombstone protobuf from `ApplicationExitInfo.getTraceInputStream()` as an attachment on native crash events.

Changes:
- `Hint.java`: Added `tombstone` field with getter/setter
- `Attachment.java`: Added `fromTombstone(byte[])` factory method (`tombstone.pb`, `application/x-protobuf`)
- `SentryClient.java`: Retrieves tombstone from hint and adds to envelope attachments
- `SentryAndroidOptions.java`: Added `attachRawTombstone` boolean option with getter/setter
- `TombstoneIntegration.java`: Buffers raw bytes from the trace InputStream before parsing, attaches when option is enabled
- `ManifestMetadataReader.java`: Added `io.sentry.tombstone.attach-raw` manifest meta-data option

### Usage

**Kotlin (programmatic):**
```kotlin
SentryAndroid.init(context) { options ->
    options.isAttachRawTombstone = true
}
```

**AndroidManifest.xml:**
```xml
<meta-data
    android:name="io.sentry.tombstone.attach-raw"
    android:value="true" />
```

Allows users to get the raw tombstone protobuf alongside the parsed native crash event, useful for deeper post-mortem analysis.

Manually tested on a physical device (HMD Pulse, Android 15) by triggering a native crash in the sample app and verifying the tombstone attachment is present.

https://sentry-sdks.sentry.io/issues/7229399848/events/1e81d1e1ed9a400b78e1be7811776a18/

Downloaded the raw tombstone file and verified it can be correctly parsed and viewed via `protoc`:

[tombstone.txt](https://github.com/user-attachments/files/27974268/tombstone.txt)

## :pencil: Checklist

## :crystal_ball: Next steps

- Update docs

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
@@ -238,6 +238,12 @@ public final class SentryAndroidOptions extends SentryOptions {
    */
   private boolean attachAnrThreadDump = false;
 
+  /**
+   * Controls whether to send raw tombstone as an attachment with plain text. The tombstone is being
+   * attached from {@link ApplicationExitInfo#getTraceInputStream()}, if available.
+   */
+  private boolean attachTombstone = false;
+
   private boolean enablePerformanceV2 = true;
 
   private @Nullable SentryFrameMetricsCollector frameMetricsCollector;
@@ -641,6 +647,14 @@ public final class SentryAndroidOptions extends SentryOptions {
 
   public void setAttachAnrThreadDump(final boolean attachAnrThreadDump) {
     this.attachAnrThreadDump = attachAnrThreadDump;
+  }
+
+  public boolean isAttachTombstone() {
+    return attachTombstone;
+  }
+
+  public void setAttachTombstone(final boolean attachTombstone) {
+    this.attachTombstone = attachTombstone;
   }
 
   /**
--- a/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
@@ -242,7 +242,7 @@ public final class SentryAndroidOptions extends SentryOptions {
    * Controls whether to send raw tombstone as an attachment with plain text. The tombstone is being
    * attached from {@link ApplicationExitInfo#getTraceInputStream()}, if available.
    */
-  private boolean attachTombstone = false;
+  private boolean attachRawTombstone = false;
 
   private boolean enablePerformanceV2 = true;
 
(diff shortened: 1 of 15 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
```diff
@@ -238,6 +238,12 @@ public interface BeforeCaptureCallback {
    */
   private boolean attachAnrThreadDump = false;
 
+  /**
+   * Controls whether to send raw tombstone as an attachment with plain text. The tombstone is being
```
supervacuus: nit: this says "attachment with plain text", but it is actually protobuf-encoded (and correctly identified as s such via the `content-type`, likely a copypasta from the ANR thread dump option.

sentry-android-core/src/main/java/io/sentry/android/core/TombstoneIntegration.java
```diff
@@ -150,22 +152,26 @@ public boolean shouldReportHistorical() {
...
+                .getLogger()
+                .log(
+                    SentryLevel.WARNING,
+                    "No tombstone InputStream available for ApplicationExitInfo from %s",
+                    DateTimeFormatter.ISO_INSTANT.format(
+                        Instant.ofEpochMilli(exitInfo.getTimestamp())));
+            return null;
+          }
+
+          rawTombstone = NativeEventUtils.readBytes(tombstoneInputStream);
```
supervacuus: l: Not a blocker, but we now unconditionally read the tombstone fully, even if the option is not turned on.

supervacuus (approved):
This looks good, only minor comments. 

Just one general assessment: the tombstone also includes raw memory views into regions pointed to by register contents. These are potentially PII-relevant and should move the tombstone attachment into a PII category similar to minidumps. This might also raise storage retention as a follow-up.

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java
romtsn: https://github.com/getsentry/sentry-java/pull/5446/commits/be90d3053cf818722933d93300257320a5fc4b18

sentry-android-core/src/main/java/io/sentry/android/core/TombstoneIntegration.java
romtsn: addressed in https://github.com/getsentry/sentry-java/pull/5446/commits/169ac4fedab460d8c871ad384b6c9ffc1e5dbb1a, no if the option is not turned on we pass the traceInputStream, otherwise we pre-buffer it for the attachment

romtsn:
@supervacuus thanks for the review 🙏

PR #5448, opened 2026-05-19
perf: Replace java.net.URI with custom string parsing in Dsn

runningcode:
## Summary

Replaces `java.net.URI` parsing in the `Dsn` constructor with lightweight manual
string parsing, and hardens and tests the parser. This is a critical startup path performance improvement.

Here are perfetto traces of the `retrieveParsedDsn()` method as measured on a Pixel 3. I've repeated the measurement several times and I always get an order of magnitude difference.

This is on master:
image
This is on this branch:
image

### What changed

- **Parsing:** Replace `new URI(dsnString).normalize()` with manual
  `indexOf`/`substring` parsing, avoiding the `java.net.URI` string parser.
  `retrieveParsedDsn()` runs on the main thread during `Sentry.init()`, so this
  keeps a known-slow API off the init path. The only remaining `URI` use is
  constructing the endpoint from already-parsed components
  (`new URI(scheme, null, host, port, path, null, null)`), which does not
  re-parse a string.
- **Correctness:** Strip URI fragments (`#…`) alongside query strings so they no
  longer leak into the project id, and support bracketed IPv6 literal hosts
  (e.g. `https://key@[::1]:9000/1`) — both behaviors the old `URI` parser handled.

JAVA-532

Relates to: https://linear.app/getsentry/issue/JAVA-532
Also relates to: https://linear.app/getsentry/issue/JAVA-516/move-dsn-parsing-off-the-main-thread-during-init

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -238,6 +238,7 @@ camerax-camera2 = { module = "androidx.camera:camera-camera2", version.ref = "ca
 camerax-lifecycle = { module = "androidx.camera:camera-lifecycle", version.ref = "camerax" }
 camerax-view = { module = "androidx.camera:camera-view", version.ref = "camerax" }
 
+google-truth = { module = "com.google.truth:truth", version = "1.4.5" }
 hsqldb = { module = "org.hsqldb:hsqldb", version = "2.6.1" }
 javafaker = { module = "com.github.javafaker:javafaker", version = "1.0.2" }
 kotlinx-coroutines-test = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-test", version.ref = "coroutines" }
--- a/sentry/src/main/java/io/sentry/Dsn.java
+++ b/sentry/src/main/java/io/sentry/Dsn.java
@@ -53,54 +53,100 @@ final class Dsn {
     return sentryUri;
   }
 
+  // Avoids java.net.URI for DSN parsing, which is slow on Android.
   Dsn(@Nullable String dsn) throws IllegalArgumentException {
     try {
       final String dsnString = Objects.requireNonNull(dsn, "The DSN is required.").trim();
       if (dsnString.isEmpty()) {
         throw new IllegalArgumentException("The DSN is empty.");
       }
-      final URI uri = new URI(dsnString).normalize();
-      final String scheme = uri.getScheme();
+
+      // Extract scheme
+      final int schemeEnd = dsnString.indexOf("://");
+      if (schemeEnd < 0) {
+        throw new IllegalArgumentException("Invalid DSN: missing scheme.");
+      }
+      final String scheme = dsnString.substring(0, schemeEnd);
       if (!("http".equalsIgnoreCase(scheme) || "https".equalsIgnoreCase(scheme))) {
         throw new IllegalArgumentException("Invalid DSN scheme: " + scheme);
       }
 
-      String userInfo = uri.getUserInfo();
-      if (userInfo == null || userInfo.isEmpty()) {
+      // Extract userinfo (public key and optional secret key)
+      final int authStart = schemeEnd + 3;
+      final int atIndex = dsnString.indexOf('@', authStart);
+      if (atIndex < 0) {
         throw new IllegalArgumentException("Invalid DSN: No public key provided.");
       }
-      String[] keys = userInfo.split(":", -1);
-      publicKey = keys[0];
-      if (publicKey == null || publicKey.isEmpty()) {
+      final String userInfo = dsnString.substring(authStart, atIndex);
+      if (userInfo.isEmpty()) {
         throw new IllegalArgumentException("Invalid DSN: No public key provided.");
       }
-      secretKey = keys.length > 1 ? keys[1] : null;
-      String uriPath = uri.getPath();
-      if (uriPath.endsWith("/")) {
-        uriPath = uriPath.substring(0, uriPath.length() - 1);
+      final int colonIndex = userInfo.indexOf(':');
+      if (colonIndex < 0) {
+        publicKey = userInfo;
+        secretKey = null;
+      } else {
+        publicKey = userInfo.substring(0, colonIndex);
+        secretKey = userInfo.substring(colonIndex + 1);
       }
-      int projectIdStart = uriPath.lastIndexOf("/") + 1;
-      String path = uriPath.substring(0, projectIdStart);
-      if (!path.endsWith("/")) {
-        path += "/";
+      if (publicKey.isEmpty()) {
+        throw new IllegalArgumentException("Invalid DSN: No public key provided.");
       }
-      this.path = path;
-      projectId = uriPath.substring(projectIdStart);
+
+      // Extract host, optional port, and path+projectId
+      final int hostStart = atIndex + 1;
+
+      // Strip query string if present
+      final int queryIndex = dsnString.indexOf('?', hostStart);
+      final String hostAndPath =
+          queryIndex < 0
+              ? dsnString.substring(hostStart)
+              : dsnString.substring(hostStart, queryIndex);
+
+      final int firstSlash = hostAndPath.indexOf('/');
+      if (firstSlash < 0) {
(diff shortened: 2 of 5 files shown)


sentry/src/main/java/io/sentry/Dsn.java
```diff
@@ -53,54 +53,100 @@ URI getSentryUri() {
     return sentryUri;
   }
 
+  // Avoids java.net.URI for DSN parsing, which is slow on Android.
   Dsn(@Nullable String dsn) throws IllegalArgumentException {
     try {
       final String dsnString = Objects.requireNonNull(dsn, "The DSN is required.").trim();
```
0xadam-brown: l: Thoughts about extracting logic into static methods for legibility? Something like: 

```java
        try {
          final String s  = requireValid(dsn);
          final String scheme  = extractScheme(s);

          final String userInfo  = extractUserInfo(s);
          final int keyColon  = userInfo.indexOf(':');
          this.publicKey = keyColon < 0 ? userInfo : userInfo.substring(0, keyColon);
          this.secretKey = keyColon < 0 ? null : userInfo.substring(keyColon + 1);
          if (this.publicKey.isEmpty()) {
              throw new IllegalArgumentException("Invalid DSN: No public key provided.");
          }
  
          final String hostPort = extractHostPort(s);
          final int portColon = hostPort.indexOf(':');
          final String host = portColon < 0 ? hostPort : hostPort.substring(0, portColon);
          final int port = portColon < 0 ? -1 : Integer.parseInt(hostPort.substring(portColon + 1));
  
          final String rawPath = extractNormalizedPath(s);
          this.path = extractPathSegment(rawPath);
          final String projectId = extractProjectId(rawPath);
          this.projectId = projectId;

          this.sentryUri = buildSentryUri(scheme, host, port, this.path, projectId);
          this.orgId = extractOrgId(host);
      } catch (Throwable e) {
          throw new IllegalArgumentException(e);
      }
```

Claude thinks HotSpot, dex2oat, etc. would inline the invocations, meaning no practical overhead.

Up to you, of course.

sentry/src/main/java/io/sentry/Dsn.java
```diff
@@ -53,54 +53,100 @@ URI getSentryUri() {
...
       }
-      secretKey = keys.length > 1 ? keys[1] : null;
-      String uriPath = uri.getPath();
-      if (uriPath.endsWith("/")) {
-        uriPath = uriPath.substring(0, uriPath.length() - 1);
+
+      // Extract host, optional port, and path+projectId
+      final int hostStart = atIndex + 1;
+
+      // Strip query string if present
```
0xadam-brown: Strip any fragment too?

```java
 final int fragmentIndex = dsnString.indexOf('#', hostStart);
```

URI would've done it. Not sure how much we care about saving folks from obvious errors they're unlikely to make, but including it will keep parity.

sentry/src/main/java/io/sentry/Dsn.java
```diff
@@ -53,54 +53,100 @@ URI getSentryUri() {
...
+      final String hostPort = hostAndPath.substring(0, firstSlash);
+      final int portColon = hostPort.indexOf(':');
+      final String host;
+      final int port;
+      if (portColon < 0) {
+        host = hostPort;
+        port = -1;
+      } else {
+        host = hostPort.substring(0, portColon);
+        port = Integer.parseInt(hostPort.substring(portColon + 1));
```
0xadam-brown: l: We could give a more informative error message if the port is malformed / not an int: 

```java
final String portStr = hostPort.substring(portColon + 1);
  try {
      port = Integer.parseInt(portStr);
  } catch (NumberFormatException e) {
      throw new IllegalArgumentException("Invalid DSN: invalid port: " + portStr);
  }
```

(Presumably rare in practice, and cleaner to do if we extract logic into static methods. Also up to you.)

sentry/src/test/java/io/sentry/DsnTest.kt
```diff
@@ -145,4 +145,71 @@ class DsnTest {
     val dsn = Dsn("http://key@localhost:9000/456")
     assertNull(dsn.orgId)
   }
+
+  @Test
+  fun `when dsn is null, throws exception`() {
+    assertFailsWith<IllegalArgumentException> { Dsn(null) }
```
0xadam-brown: l: My vote would be for asserting particular error messages in these tests, as they're part of our implicit API. (Looks like we sometimes do it above.)

sentry/src/main/java/io/sentry/Dsn.java
0xadam-brown: ...ah, cursorbot beat me to it.

0xadam-brown (approved):
A few optionals; otherwise lgtm

sentry/src/test/java/io/sentry/DsnTest.kt
runningcode: Good idea. I've updated the tests to add google truth to make this easier and also improve the error messages.

sentry/src/main/java/io/sentry/Dsn.java
runningcode: added better parsing in a separate method for parsing the port.

sentry/src/main/java/io/sentry/Dsn.java
runningcode: I don't think we support this in our DSN's but I will fix it anyways

sentry/src/main/java/io/sentry/Dsn.java
runningcode: Good idea i've extracted some of the logic to helper methods. let me know what you think

gradle/libs.versions.toml
```diff
@@ -238,6 +238,7 @@ camerax-camera2 = { module = "androidx.camera:camera-camera2", version.ref = "ca
 camerax-lifecycle = { module = "androidx.camera:camera-lifecycle", version.ref = "camerax" }
 camerax-view = { module = "androidx.camera:camera-view", version.ref = "camerax" }
 
+google-truth = { module = "com.google.truth:truth", version = "1.4.5" }
```
0xadam-brown, line 241: l: No strong opinions about introducing a new assertion library for tests + it's helpful for verifying error msgs, so I'm okay with this even though we'll only have a few tests using Truth.

Anyone else can feel free to chime in if they think otherwise...

sentry/src/main/java/io/sentry/Dsn.java
0xadam-brown: Lgtm, and much easier to track – thx 🥇

0xadam-brown (approved):
Thanks for the updates – lgtm 👍

sentry/src/main/java/io/sentry/Dsn.java
```diff
@@ -17,98 +18,148 @@ final class Dsn {
...
       }
-      this.path = path;
-      projectId = uriPath.substring(projectIdStart);
+
+      final String hostPort = hostAndPath.substring(0, firstSlash);
+      final int portColon = portSeparatorIndex(hostPort);
+      final String host = portColon < 0 ? hostPort : hostPort.substring(0, portColon);
+      final int port = portColon < 0 ? -1 : parsePort(hostPort.substring(portColon + 1));
+
+      final String rawPath = stripTrailingSlash(collapseSlashes(hostAndPath.substring(firstSlash)));
```
romtsn, line 93: theoretically this could've been a single for-loop with a StringBuilder -- would've spared us a couple of more `substring` calls, but up to you if you'd like to optimize it further.

romtsn (approved):
one nit, but LGTM otherwise! great gains

sentry/src/main/java/io/sentry/Dsn.java
runningcode: i had claude code that up and it was hard to understand what the code was doing. it would also only save 3 strings if i understand your suggestion correctly so probably not worth it. good eye though!

gradle/libs.versions.toml
runningcode: I already added it to SAGP last year. I think it has much better error messages and enforces much better practices

PR #5466, opened 2026-05-22
feat(android-sqlite): Add SentrySQLiteDriver (JAVA-275)

0xadam-brown:
## :scroll: Description

Introduces `SentrySQLiteDriver` for wrapping and instrumenting `androidx.sqlite.SQLiteDriver`. Like our existing `SentrySupportSQLiteOpenHelper`, the new wrapper produces a span per executed SQL statement. 

Example use:

```kotlin
  Room.databaseBuilder(context, AppDatabase::class.java, "myapp.db")
      .setDriver(SentrySQLiteDriver.create(AndroidSQLiteDriver()))
      .build()
```

Room 2.7 introduced the `SQLiteDriver` API as a replacement for `SupportSQLiteOpenHelper`; Room 3.0+ makes use of `SQLiteDriver` mandatory.

A key motivation behind Google's introduction of `SQLiteDriver` was Kotlin Multiplatform compatibility. This PR makes `SentrySQLiteDriver` available on Android only, but the wrapper has been packaged so that we can lift it into our KMP module in the future without having to break clients.

Addresses JAVA-275.

## ⚠️ Callouts

**[1]** Unlike `SentrySupportSQLiteOpenHelper`, the `SentrySQLiteDriver` is _**not**_ automatically wrapped via the sentry-android-gradle-plugin. (We can add byte code support later if we want – or do it now if there's a strong interest.)

**[2]** API is *not* marked as `@Experimental`. (Super small surface area: a `create(SQLiteDriver)` constructor; future additions are non-breaking; sufficient alignment with `SentrySupportSQLiteOpenHelper` data model. That said, chime in if you think we should add it.)

**[3]** Behavior differences vs `SentrySupportSQLiteHelper`

<details>
<summary>Click to expand</summary>

### Behavior differences

| Aspect | `SentrySupportSQLiteOpenHelper` (old) | `SentrySQLiteDriver` (new) | Evaluation |
|---|---|---|---|
| Span duration | Whole `performSql` call (incl. app time during cursor materialization) | Accumulated `step()` db time only – app time between steps excluded | **New is more accurate.** Span duration is now limited to SQLite work, not the read loop plus arbitrary app-side row processing, I/O, or GC pauses between rows. Cleaner p95s for db performance dashboards; teams using span duration as a proxy for end-to-end read time will see it shrink. |
| Cursor iteration | Only first `getCount`/`onMove`/`fillWindow` timed; later rows untimed | Every `step()` contributes to cumulative span time | **New is more complete.** A 10k-row read now reflects the full cumulative db cost rather than just the first window fill. The old path silently under-reported large reads. |
| Span wall-clock anchor | Captured *before* the operation runs | Captured at first `step()` of the cycle | **Visually, new span sits slightly later in trace** New anchor tracks when SQLite actually started working; old anchor included setup/dispatch overhead in the span timeline.|
| `db.name` derivation | Reads open helper `databaseName` (for Room, the **builder name**, e.g., `"tracks"` from `databaseBuilder(ctx, MyDb, "tracks")`). | Reads `File(fileName).name` — the **on-disk filename** of the path Room passes to `driver.open()` (e.g., `"tracks.db"`). | **The same db can show up under two different `db.name` values during migration.** Both paths report data correctly, but will attribute it to different sources.|
| Multi-statement scripts | `execSQL("CREATE TABLE …; INSERT …; INSERT …;")` produces **one span** whose description is the full script. | The Driver API compiles one statement per `prepare(...)`, so multi-statement scripts must be split by Room (or the caller) into separate prepare/step cycles → **one span per statement**. | **New is more accurate but more verbose.** Migration scripts and seed scripts that previously appeared as one bundled span will now appear as N smaller spans, each with its own timing and description. Useful for finding the slow statement; expect span counts to rise for these code paths. |

</details>

**[4]** Risk of duplicate spans with Room's bridge adapter

`SentrySQLiteDriver` protects internally against duplicate span creation should a developer try to double-wrap it or any of its components. Room and SQLDelight ensure that use of the driver and `SentrySupportSQLiteOpenHelper` are mutually exclusive at the API level in virtually all instances, save for one: 

Room 2.7+ (but not Room 3.0+) exposes a bridge adapter (`SupportSQLiteDriver`) that lets users delegate to a `SQLiteDriver` from an existing `SupportSQLiteOpenHelper`. Double-wrapping both components is possible if folks aren't mindful. 

I'll be updating the [Sentry Docs](https://docs.sentry.io/platforms/android/integrations/room-and-sqlite/) to warn users against wrapping both adapter components. Another option is to log an error, as mentioned [here](https://github.com/getsentry/sentry-java/pull/5466#discussion_r3318215398).

<details>
<summary>Updates to Sentry Docs: Click to expand</summary>

## Avoiding duplicate spans with Room 2.7+

AndroidX ships an adapter class, `SupportSQLiteDriver`, that lets developers bridge an existing `SupportSQLiteOpenHelper` to a `SQLiteDriver` that Room 2.7+ accepts. **Do not wrap both the open helper and the driver.** (Remember that the Sentry Android Gradle Plugin will wrap the open helper for you at the byte code level if enabled.) If you double-wrap, you'll produce duplicate spans for every SQL statement:

```kotlin
// AVOID — this configuration produces duplicate spans for every SQL statement.

// Step 1: Wrap the open helper manually or via the Sentry Android Gradle Plugin.
val sentryWrappedHelper: SupportSQLiteOpenHelper =
    SentrySupportSQLiteOpenHelper.create(
        FrameworkSQLiteOpenHelperFactory().create(configuration)
    )

// Step 2: Pass the wrapped helper to the Room 2.7+ adapter.
val driver: SQLiteDriver = SupportSQLiteDriver(sentryWrappedHelper)

// Step 3: Also (wrongly!) wrap the driver. All spans will now be duplicated.
val sentryWrappedDriver: SQLiteDriver = SentrySQLiteDriver.create(driver)

Room.databaseBuilder(context, MyDb::class.java, "mydb")
    .setDriver(sentryWrappedDriver)
    .build()
```
</details>

Unit tests cover:
- `SentrySQLiteDriver` delegation and wrapping
- `SentrySQLiteConnection` delegation and wrapping
- `SentrySQLiteStatement` span creation, cancellation, and success/error tagging
- `SQLiteSpanRecorder` span lifecycle (start/finish/cancel)

I also dog-fooded `SentrySQLiteDriver` on my own example app via Maven local artifact + I verified spans are displayed in Sentry UI. 

⚠️ Possibilities I did without:

1. Testing against a real Room db in this PR (I did so manually in my example app, but this PR doesn't include a test that wires up Room)
2. Instrumentation tests
3. Add a sample to [sentry-samples-android](https://github.com/getsentry/sentry-java/tree/main/sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android) (not really a test, but it would exercise the code paths in an actual Android environment).

The legacy open helper didn't have the above, so I followed suit. Let me know if you think any is worth it and I'll be happy to address (eg, real Room db test, which would require a test dependency on androidx.sqlite:sqlite-bundled but wouldn't require Robolectric).

## :pencil: Checklist

## :crystal_ball: Next steps

1. Update the [Room & SQLite Sentry docs](https://docs.sentry.io/platforms/android/integrations/room-and-sqlite/).
2. If there's sufficient demand, we could support auto-wrapping `SQLiteDriver` via the sentry-android-gradle-plugin at some point (not currently on the roadmap).

#skip-changelog


sentry-android-sqlite/src/main/java/io/sentry/android/sqlite/SQLiteSpanManager.kt
```diff
@@ -4,20 +4,18 @@ import android.database.CrossProcessCursor
...
+import io.sentry.sqlite.dbMetadataFromDatabaseName
 
 internal class SQLiteSpanManager(
   private val scopes: IScopes = ScopesAdapter.getInstance(),
-  private val databaseName: String? = null,
+  databaseName: String? = null,
 ) {
-  private val stackTraceFactory = SentryStackTraceFactory(scopes.options)
+
+  private val spanHelper = SQLiteSpanHelper(scopes, dbMetadataFromDatabaseName(databaseName))
```
0xadam-brown: Helper lets us share span population logic between the old and new worlds. Behavior on the legacy path is unchanged.

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SentrySQLiteDriver.kt
```diff
@@ -0,0 +1,66 @@
...
+ * instrument at different layers, so combining them will produce duplicate spans for every SQL
+ * statement.
+ *
+ * @param delegate The [SQLiteDriver] instance to delegate calls to.
+ */
+public class SentrySQLiteDriver private constructor(private val delegate: SQLiteDriver) :
+  SQLiteDriver {
+
+  init {
+    SentryIntegrationPackageStorage.getInstance().addIntegration("SQLiteDriver")
```
0xadam-brown, line 35: ⚠️ Note that the integration name differs from what we use with the  support open helper ("SQLite"). I presume that's what developers would expect and it follows the pattern of ANR vs ANR2, but let me know if you think otherwise.

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SentrySQLiteStatement.kt
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Wraps a [SQLiteStatement] and records a single Sentry span covering all [step] calls for the
+ * statement's lifetime (until [step] iteration is complete or the statement is [reset] or
+ * [closed][close]).
+ *
+ * Span duration is purposefully restricted to accumulated database time, i.e., each [step] call is
+ * individually timed and the durations are summed. Time the application spends between steps (e.g.,
+ * processing rows, sleeping, or doing I/O) is intentionally excluded so the span accurately
+ * represents how long SQLite itself was working.
```
0xadam-brown: ### Alternative approaches for measuring span duration

This is the most impactful design decision of this PR from a user's perspective. Let's think through our options. I've implemented alternative (3) below; I'm on the fence between (3) and (4).

#### An example SQLiteStatement call sequence

```kotlin
    val statement = sqliteConnection.prepare("SELECT * ...")

    while (statement.step()) {
      val id = statement.getLong(0)
      val sender = statement.getText(1)
      val bodyBytes = statement.getBlob(2)
      val body = converter.fromEncryptedBytes(bodyBytes)
      results.add(Message(id, sender, body))
    }
```

#### A few terms

- **Preparation time**: Time the SQLite engine spends parsing and optimizing a statement before executing it (represented by the sqliteConnection.prepare() call in the example above)

- **Execution time**: Time the SQLite engine spends executing and advancing a statement, plus loading results into native memory (calls to step() in the example).

- **Consumption time**: Time spent bringing statement results into the ART-managed heap and accessing them at the application level (calls to getLong(), getText() and getBlob() in the example). Includes things like type conversion, JNI boundary crossing, copying native bytes to ART's heap, GC sweeps during consumption, etc.

- **Application time**: Time the application spends processing statement results while the statement is still ongoing (here, converter.fromEncryptedBytes(bodyBytes)). Can involve arbitrary work, including parsing, decryption, etc.

#### Alternatives

1. **Only track the initial step() call**: Includes all execution time for single-step() statements; only captures part of execution time for multi-step() statements. Superficially similar to what we do with `SupportSQLiteOpenHelper` (which is why I bothered to include it here), but materially different in that the open helper's initial call can load an entire cursor window, whereas step() amortizes loading across invocations. 

2. **Track the entire duration between the initial and final step() calls**: Includes execution time + consumption time + application time. Matches easily with the `ISpan` API, which expects a start and end timestamp for the span. But application time can make span duration misleading (see dropdown below).

<details>
<summary>Why we don't want to include application time: Click to expand</summary>

#### Application time can be long and it's not db-related

> E.g., app queries 200 rows from a SQLite db, where each row contains an encrypted column that the app decrypts between step() calls. (See converter.fromEncryptedBytes(bodyBytes) from the example query above).
> 
> If each step() takes ~0.1ms and each decryption takes ~2ms, cumulative execution time is ~20ms, but execution + application time is ~420ms. Developers will think they have an inefficient query when in fact 95% of the span is unrelated decryption overhead.
</details>

3. **Track only time spent in each step() call (what this PR implements):** Limited to execution time. This or (4) are presumably what most developers would want from a `db.sql.query` span. I opted for (3) because it's simpler to implement, gives an (arguably) purer look at query + SQLite version performance, and leaves open the possibility of adding a separate `db.sql.consume` span to cover (4) in the future. That said, I'm open to being convinced....

4. **Track time spent in all SQLiteStatement methods (step() + getBlob(), getLong(), etc.)**: Includes execution time + consumption time. My second choice. 

#### Another axis

To any of the above, we could consider adding preparation time, which we'd track via SentrySQLiteConnection.prepare(). I've omitted it because:

- it's computed once per statement construction, but each statement can be executed multiple times, meaning we'd either have to attribute the cost asymmetrically to the first execution, or misleadingly to all; and 
- we often exclude statement preparation when computing spans for our legacy open helper, including whenever the statement returns a cursor.

We could add a separate `db.sql.preparation` span later if we want to track preparation time in isolation.

sentry-android-sqlite/src/test/java/io/sentry/sqlite/SentrySQLiteConnectionTest.kt
```diff
@@ -0,0 +1,63 @@
...
+    val alreadyInstrumented = SentrySQLiteStatement(fixture.mockStatement, spanRecorder, "SELECT 1")
+    whenever(fixture.mockConnection.prepare("SELECT 1")).thenReturn(alreadyInstrumented)
+
+    val statement = sut.prepare("SELECT 1")
+
+    assertSame(alreadyInstrumented, statement)
+  }
+
+  @Test
+  fun `all calls are propagated to the delegate`() {
```
0xadam-brown, line 54: Delegation tests (here and below) copy existing tests for the legacy open helper. Let me know if you think they're overkill.

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SentrySQLiteDriver.kt
```diff
@@ -0,0 +1,66 @@
+package io.sentry.sqlite
```
0xadam-brown, line 1: As mentioned in the PR description, I've created a separate package for the wrapper without an `.android.` infix and scrubbed the package of any Android dependencies. 

The point of Google's introducing the new driver was to support KMP, so I suspect we may want to lift the wrapper to sentry-kotlin-multiplatform at some point.

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SentrySQLiteDriver.kt
```diff
@@ -0,0 +1,66 @@
...
+ * ```
+ * val database = Room.databaseBuilder(context, MyDatabase::class.java, "dbName")
+ *     .setDriver(SentrySQLiteDriver.create(AndroidSQLiteDriver()))
+ *     .build()
+ * ```
+ *
+ * **Warning:** Do not use [SentrySQLiteDriver] together with
+ * [io.sentry.android.sqlite.SentrySupportSQLiteOpenHelper] on the same database file. Both wrappers
+ * instrument at different layers, so combining them will produce duplicate spans for every SQL
+ * statement.
```
0xadam-brown: I plan to repeat a similar warning in the Sentry Docs (update forthcoming). 

⚠️ In addition, we could consider logging an error if we detect that both the "SQLite" and "SQLiteDriver" integrations have been registered. I haven't implemented that b/c it's somewhat involved, so chime in if you want it. Otherwise I'll stick with the verbal warning, especially as the risk of double-wrapping disappears once folks are on Room 3.0+.

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SentrySQLiteDriver.kt
0xadam-brown: ...a new README captures the policy.

markushi: 💯 , smart choice!

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SentrySQLiteDriver.kt
markushi: Whilst we surface this information to developers on the issue details page, the main use case is for us (sentry) to get an idea of which integrations are used and to track adoption.
So using a different name here is actually preferred.

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SentrySQLiteDriver.kt
```diff
@@ -0,0 +1,66 @@
...
+ * ```
+ *
+ * **Warning:** Do not use [SentrySQLiteDriver] together with
+ * [SentrySupportSQLiteOpenHelper][io.sentry.android.sqlite.SentrySupportSQLiteOpenHelper] on the
+ * same database file. Both wrappers instrument at different layers, so combining them will produce
+ * duplicate spans for every SQL statement.
+ *
+ * @param delegate The [SQLiteDriver] instance to delegate calls to.
+ */
+public class SentrySQLiteDriver private constructor(private val delegate: SQLiteDriver) :
```
markushi: Once we lift this up for kmp, we'll probably need to extend this, like e.g. passing a `scope / options / logger` instance, instead of relying on `ScopesAdapter.getInstance()` like we do below. But IMHO that's fine for now and won't cause a breaking change.

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SentrySQLiteStatement.kt
markushi: Good callout! To be honest I would actually lean towards 2: tracking everything including application time, as this would result in non-synthetic span end timestamps. But your concerns are more than valid, I would even expect Application Time to be the main culprit of many wrongly classified "slow queries". I see a few options, not sure if they're feasible:
1. create a child span for "application time", this will cause our backend to calculate a `span.self_time` for the parent span, which then can be queried and analyzed
2. extend https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/SpanDataConvention.java to simply add the application timings as a property

1. seems to be the way to go, as it allows all sorts of queries out of the box - but comes with the cost of generating extra spans. 
@romtsn any thoughts on this?

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SentrySQLiteStatement.kt
```diff
@@ -0,0 +1,81 @@
...
+    } catch (e: Throwable) {
+      accumulatedDbNanos += nanoTimeProvider() - beforeNanos
+      recordSpan(SpanStatus.INTERNAL_ERROR, e)
+      throw e
+    }
+  }
+
+  override fun reset() {
+    if (closed) {
+      return delegate.reset()
```
markushi, line 57: any reason why this is guarded by `if (closed)`, shouldn't we always call the delegate?

markushi (commented):
Looks very promising already, left a few comments. Once we've decided what execution phases spans should cover, I'll have a second look.

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SentrySQLiteStatement.kt
0xadam-brown: We actually call the delegate irrespective of the value of `closed` (but a quick reading can be misleading – I actually had to re-read to make sure I hadn't goofed 😅). When `closed` is true, we call the delegate but skip creating a span; when false, we delegate + create.

sentry-android-sqlite/src/main/java/io/sentry/android/sqlite/SQLiteSpanManager.kt
```diff
@@ -45,33 +43,18 @@ internal class SQLiteSpanManager(
...
       }
-      span = scopes.span?.startChild("db.sql.query", sql, startTimestamp, Instrumenter.SENTRY)
-      span?.spanContext?.origin = TRACE_ORIGIN
+      span = spanHelper.startSpan(sql, startTimestamp)
       span?.status = SpanStatus.OK
       result
     } catch (e: Throwable) {
-      span = scopes.span?.startChild("db.sql.query", sql, startTimestamp, Instrumenter.SENTRY)
-      span?.spanContext?.origin = TRACE_ORIGIN
+      span = spanHelper.startSpan(sql, startTimestamp)
```
romtsn: wait, I think this doesn't make sense, or does it? If an sql operation fails we start a new span... for throwing an exception? I guess this was overlooked, we should drop this 😅

romtsn: ah, dammit it had been introduced during merge conflict resolution. We should def. drop this

sentry-android-sqlite/src/main/java/io/sentry/sqlite/DbMetadata.kt
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Resolves metadata from the [fileName] argument to
+ * [SQLiteDriver.open][androidx.sqlite.SQLiteDriver.open].
+ */
+internal fun dbMetadataFromFileName(fileName: String): DbMetadata {
+  if (fileName == IN_MEMORY_DB_FILENAME) {
+    return DbMetadata(name = null, system = DB_SYSTEM_IN_MEMORY)
+  }
+
+  val trimmed = fileName.trimEnd('/', '\\')
```
romtsn: could we spread `FILE_NAME_PATH_SEPARATORS` here?

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SentrySQLiteStatement.kt
romtsn: as discussed internally let's for the current approach (option 3) and adapt Room instrumentation in SAGP, if necessary.

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SQLiteSpanRecorder.kt
romtsn, line 1: I really like that you've tried to split responsibilities here, but was wondering if we could somehow reduce the number of helper classes? 😅 We now have a Manager, a Recorder and a Helper -- that's too many "ers" for my taste, haha, and it's a bit difficult to make a distinction between them. 

I understand that it might be not possible due to backwards-compat with the existing sqlite instrumentation, but perhaps there's a way?

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SQLiteSpanRecorder.kt
```diff
@@ -0,0 +1,45 @@
...
+    status: SpanStatus,
+    throwable: Throwable? = null,
+  ) {
+    try {
+      val span = spanHelper.startSpan(sql, startTimestamp) ?: return
+      throwable?.let { span.throwable = it }
+      spanHelper.applyDataToSpan(span)
+      val endTimestamp = SentryLongDate(startTimestamp.nanoTimestamp() + durationNanos)
+      span.finish(status, endTimestamp)
+    } catch (t: Throwable) {
```
romtsn: I think this is a bit overly-safe here since everything here is our own code, I'd rather prefer to let it crash so we can know about it and fix it 😅

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SentrySQLiteStatement.kt
```diff
@@ -0,0 +1,81 @@
...
+      stepsComplete = !delegate.step()
+      accumulatedDbNanos += nanoTimeProvider() - beforeNanos
+      if (stepsComplete) {
+        recordSpan(SpanStatus.OK)
+      }
+      !stepsComplete
+    } catch (e: Throwable) {
+      accumulatedDbNanos += nanoTimeProvider() - beforeNanos
+      recordSpan(SpanStatus.INTERNAL_ERROR, e)
+      throw e
```
romtsn, line 51: wondering if we have to set `stepsComplete` to `true` here... or can this statement be reused after an error arises here?

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SentrySQLiteStatement.kt
```diff
@@ -0,0 +1,81 @@
...
+ * processing rows, sleeping, or doing I/O) is intentionally excluded so the span accurately
+ * represents how long SQLite itself was working.
+ *
+ * Not thread-safe: assumes sequential access within each SQL statement (normal SQLite usage).
+ */
+internal class SentrySQLiteStatement(
+  private val delegate: SQLiteStatement,
+  private val spanRecorder: SQLiteSpanRecorder,
+  private val sql: String,
+  private val nanoTimeProvider: () -> Long = { System.nanoTime() },
```
romtsn, line 22: Any reason we are not using SentryDate here, too? just simplification? i'm fine either way, just curious

romtsn (commented):
Looks great, nice analysis and research, kudos!

Some more things I'd want to clarify before approving:
- do we actually need to bump `androidx.sqlite` version that we compile against in the module to support this new stuff?
- what's the migration path? do we want to keep them side-by-side in the same module, or would it make sense to introduce `sentry-android-sqlite3` (or whatever fits best) to keep them separate? would probably also simplify the future KMP work (if ever)
- you said you're yet to test it out with SAGP and auto-instrumentation, which is great! I think I would want to also make SAGP instrumentation of the new Driver a part of the success criteria, because this is the main method for our users to get room/sqlite instrumented (obviously it doesn't have to block this PR)

sentry-android-sqlite/src/main/java/io/sentry/sqlite/DbMetadata.kt
0xadam-brown: Good catch! Detekt fails on spread operators, but  `trimEnd { it in FILE_NAME_PATH_SEPARATORS }` will do the trick.

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SentrySQLiteStatement.kt
0xadam-brown: Just to avoid extra SentryDate allocations for each step() call (SentryDateProvider.now() allocates a SentryNanotimeDate + a Date + a Calendar, and we'd need two now() calls per step).

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SentrySQLiteStatement.kt
0xadam-brown: Good question. 

> or can this statement be reused after an error arises here?

It can indeed...at least in theory.

If one of the delegate androidx.sqlite implementations throws during `step()`, the statement remains valid and re-steppable. (Implementations track `isClosed` flags but not error states, and no impl closes if `step()` throws.) 

The current approach matches that^^ behavior. But it won't really matter in practice. If Room encounters an exception during `step()` it automatically calls `reset()`, which (rightly) sets `stepsComplete = false`.  So if we were to set `stepsComplete = true` in the catch block here, it'd be immediately set back to false by Room. (Note: SQLDelight doesn't use `SQLiteDriver` yet.) 

My mild vote is to leave the implementation as-is b/c it tracks the (implicit) contract of `step()` that permits re-stepping after an exception is thrown, even if that'll almost never happen in the wild.

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SQLiteSpanRecorder.kt
0xadam-brown: Two options (chime in if you prefer a third): 

1. Merge `SQLiteSpanHelper` into `SQLiteSpanRecorder` at the cost of making the resulting API wider than it needs to be for either consumer.
2. Do (1) but no longer share logic across the `.android.sqlite` and `.sqlite` packages (lets us keep the `SQLiteSpanRecorder` API tailored to its now sole consumer).

I'll do (1), see whether (2) calls to me, and then make a decision on my end. Feel free to wait for me to post my next commit if you don't have a strong opinion in the meantime....

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SQLiteSpanRecorder.kt
0xadam-brown: Nixing the try-catch!

sentry-android-sqlite/src/main/java/io/sentry/android/sqlite/SQLiteSpanManager.kt
0xadam-brown: Hmm... I actually think this was intentional. (It's been with us for a while and it's a pattern used elsewhere when an action that should produce a plain vanilla span throws instead, see eg sentry-jdbc, sentry-okhttp, etc.)

Note that `spanHelper` uses a lazy start pattern here, in that it's documenting the error span after the fact and immediately finishes the span in `finally`, rather than leaving it open.

So presumably we want to keep this? (+ the similar `INTERNAL_ERROR` span in the `catch` clause of the new `SentrySQLiteStatement.span()` method).

That said, let me know if I goofed and we really do want to strip the error span out.

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SentrySQLiteStatement.kt
romtsn: makes sense!

sentry-android-sqlite/src/main/java/io/sentry/android/sqlite/SQLiteSpanManager.kt
romtsn: ah, indeed, I missed that we're using the startTimestamp that we retrieved at the beginning of the method 🙈 

I guess it reads a bit confusing (like we're starting the span inside `catch {}` as opposed to starting it before the sql op), I'd expect something like:
```kotlin
val span = spanHelper.startSpan(sql, now)
try {
  val result = operation()
  span?.status = SpanStatus.OK
  result
} catch (e: Throwable) {
  span?.status = INTERNAL_ERROR
  span?.throwable = e
} finally {
  ...
  span?.finish()
}
```
which is more natural. But the current approach is also fine by me 👍

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SentrySQLiteStatement.kt
romtsn: sounds good, let's leave it as-is!

0xadam-brown:
Thanks for the excellent feedback / comments @romtsn. Answers to your questions + one question about single vs multiple modules on my end...

_**> do we actually need to bump `androidx.sqlite` version that we compile against in the module to support this new stuff?**_

~~Not for this PR: we're already on androidx.sqlite:sqlite:2.5.2 and `SQLiteDriver` landed in 2.5.0.~~ 

Scratch that^^. I've now bumped to 2.6.2 to pick up the introduction `SQLiteDriver.hasConnectionPool()` in 2.6.0. (I originally thought I'd give that fix its own PR, but it was trivial so I've included it here for completeness.)

Fyi, I'll wait to merge all the PRs together so we make sure they land in the same release.

_**> what's the migration path? do we want to keep them side-by-side in the same module, or would it make sense to introduce u`sentry-android-sqlite3` (or whatever fits best) to keep them separate? would probably also simplify the future KMP work (if ever)**_

This ended up being an important question (thx!). I propose keeping the driver in the existing `sentry-android-sqlite` module (see discussion below).

Here are the migration steps I had in mind:

1. Namespace the driver in a way that lets it (eventually) be lifted into sentry-kotlin-multiplatform + keep its public ABI compatible with KMP common. (Already completed via this PR.)
2. Prepare for lifting (e.g., update driver to use TimeSource rather than System.nanoTime(); have someone create KMP common versions of ISpan, etc. – this latter work is the real blocker atm). 
3. Lift all of `io.sentry.sqlite` into a new sentry-kotlin-platform module (eg, `sentry-kmp-sqlite`) without changing its public ABI.
4. Remove the driver from `sentry-android-sqlite` (or a new `sentry-sqlite` module if we go that route) + update the existing module's build.gradle to pull in the driver so consumers don't have to change their coordinate save for bumping the version:

```groovy
// build.gradle in sentry-android-sqlite (if the driver lives with the open helper) or 
// sentry-sqlite (if we give the driver its own module)
dependencies {
    api("io.sentry:sentry-kmp-sqlite") 
}
```

Creating a new module doesn't meaningfully simplify future KMP work because we'll always have to relocate the driver somewhere + redirect from the existing module's build.gradle script. (Eg, if we want to publish the driver in a JAR for non-Android JVM consumers / desktop, we could create a new module then and redirect from sentry-android-sqlite; if we want to go straight for common KMP, we do likewise, just in sentry-kotlin-multiplatform.) See the expandable "migration path to KMP" discussion below for more details.

`sentry-compose` seems reasonable prior art here, and it doesn't have a separate module for its KMP-compatible code. So my vote would be to keep the driver in sentry-android-sqlite for now to aid open helper -> driver migration, and then break it out if and when the need arises.

cc @markushi 

<details>
  <summary>Step-by-step migration path to KMP: Click to expand</summary>

### Migration path

| # | Step  | **Path A: Use existing module** | **Path B: Create `sentry-sqlite` module now** |
|---|---|---|---|
| 1 | Untangle shared helpers (`SQLiteSpanHelper`, `DbMetadata`) from the open helper path | **Must do** - trivial | ✅  **Already done** |
| 2 | Isolate driver + shared code into a standalone, Android-free compilation unit with its own coordinate | **Must do** - this is the module split itself | ✅  **Already done** |
| 3 | **Relocate driver into `sentry-kotlin-multiplatform`** (new module/source set there) | **Required** | **Required** |
| 4 | Extend the KMP SDK's _common_ API to cover what the wrapper needs: child span with explicit start/finish timestamps, `setData` (db.system/db.name/blocked_main_thread/call-stack), span status, trace origin, scope access, integration registration | **Required** | **Required** |
| 5 | Rewrite the Sentry-touching code (`SQLiteSpanHelper`/`SQLiteSpanRecorder`/registration) against the KMP common API; drop all JVM-only `io.sentry.*` types (`ISpan`, `IScopes`, etc.) | **Required** | **Required** |
| 6 | Replace `System.nanoTime()` with a multiplatform monotonic clock (`kotlin.time.TimeSource.Monotonic`) | **Required** | **Required** |
| 7 | Convert build to `kotlin.multiplatform`, add targets, set up GMM publishing + `.craft`/release wiring **in the sentry-kmp repo** | **Required** | **Required** |
| 8 | Redirect the sentry-java artifact toward the new KMP coordinate; migration note for existing Android/JVM users | **Required** | **Required**|

### User experience

| | Path A | Path B |
|---|---|---|
| Add today (typical Android user) | Often zero new dep; familiar artifact | A new, separate artifact (two during Room 2.7 transition) |
| Code today | `io.sentry.sqlite.SentrySQLiteDriver` | Identical |
| ProGuard / stack-trace fidelity | Automatic (AAR) | Manual `META-INF/proguard` — small risk of silent degradation |
| Off-Android JVM use | Impossible | Works |
| Greenfield driver-only classpath | Carries unused open-helper code | Cleaner (driver-only) |

KMP migration shape for both paths is identical: no code change so long as we can keep the package name in sentry-kotlin-multiplatform; dependency redirect means only version bump needed on existing coordinate.

</details>

_**> you said you're yet to test it out with SAGP and auto-instrumentation, which is great! I think I would want to also make SAGP instrumentation of the new Driver a part of the success criteria, because this is the main method for our users to get room/sqlite instrumented (obviously it doesn't have to block this PR)**_

Good deal. I'll plan to implement SAGP auto-instrumentation of `RoomDatabase.Builder.setDriver()` as part of the current project. (I'll also double-check the behavior of our current auto-generated `db.sql.room` spans, and we can tweak if needed.)

sentry-android-sqlite/src/main/java/io/sentry/android/sqlite/SQLiteSpanManager.kt
0xadam-brown: Yeah, this is due to my ambition to share span building logic btw the open helper and driver. I'll see whether I can't make the API more obvious.

0xadam-brown: Updated to avoid splitting between `startSpan()` and `finishSpan()` calls. Now there's just one w/a (hopefully) less misleading name:

```diff
@@ -43,19 +41,11 @@ internal class SQLiteSpanManager(
       if (result is CrossProcessCursor) {
         return SentryCrossProcessCursor(result, this, sql) as T
       }
-      span = spanHelper.startSpan(sql, startTimestamp)
-      span?.status = SpanStatus.OK
+      spans.recordSpan(sql, startTimestamp, SpanStatus.OK)
       result
     } catch (e: Throwable) {
-      span = spanHelper.startSpan(sql, startTimestamp)
-      span?.status = SpanStatus.INTERNAL_ERROR
-      span?.throwable = e
+      spans.recordSpan(sql, startTimestamp, SpanStatus.INTERNAL_ERROR, e)
       throw e
-    } finally {
-      span?.let {
-        spanHelper.applyDataToSpan(it)
-        it.finish()
-      }
     }
   }
 }
```

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SQLiteSpanRecorder.kt
0xadam-brown: Did (1) and it was able to tailor the API appropriately to each consumer by overloading a single `recordSpan()` method. Satisfied on my end. Let me know if you'd like any additional changes.

romtsn: no, that looks great, thanks for addressing this!

sentry-android-sqlite/src/main/java/io/sentry/android/sqlite/SQLiteSpanManager.kt
romtsn: yep, recordSpan is way more intuitive, thanks!

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SentrySQLiteDriver.kt
```diff
@@ -0,0 +1,69 @@
...
+    } catch (t: Throwable) {
+      ScopesAdapter.getInstance()
+        .options
+        .logger
+        .log(
+          SentryLevel.ERROR,
+          "Failed to instrument SQLite connection; returning uninstrumented connection.",
+          t,
+        )
+      connection
```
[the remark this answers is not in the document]
0xadam-brown: Leaving this as-is b/c error path isn't realistic and our common practice is not to wrap logging in catch clauses.

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SentrySQLiteDriver.kt
```diff
@@ -0,0 +1,75 @@
...
+
+  init {
+    SentryIntegrationPackageStorage.getInstance().addIntegration("SQLiteDriver")
+  }
+
+  override val hasConnectionPool: Boolean
+    get() =
+      try {
+        delegate.hasConnectionPool
+      } catch (_: LinkageError) {
```
0xadam-brown, line 42: I see that this is the only reference to a `LinkageError` in our code base. Do we normally handle these sorts of issues differently? 

*** "These sorts of issues" = A `compileOnly` dependency against an API that's added methods as its versions evolve. In this case, androidx.sqlite 2.6.0 added `SQLiteDriver.hasConnectionPool`. Because we don't include our own androidx.sqlite dep at runtime, we need to account for users who bring in versions with and without `hasConnectionPool`.

romtsn: I see we had a similar one here: https://github.com/getsentry/sentry-java/pull/4597 which we just fixed by bumping the version, but apparently we don't account for older versions, so I guess it just breaks there?

romtsn (approved):
LGTM! great work

0xadam-brown:
Making driver interface `internal` so we can merge + keep building. Will restore to public + add CHANGELOG entry prior to release.

PR #5473, opened 2026-05-27
fix(replay): Associate trace IDs with replay segments

romtsn:
## :scroll: Description

This change enables associating trace IDs with replay segments, allowing replays to be searched and filtered by the trace IDs of transactions that occurred during recording.

### Key Changes:
- Added `registerTraceId()` method to `ReplayController` interface and implementations
- Modified `BaseCaptureStrategy` to collect trace IDs in a `ConcurrentHashMap` and include them in generated replay segments
- Updated `SentryClient.captureTransaction()` to register the transaction's trace ID with the replay controller when a transaction is captured
- Modified `ReplayIntegration` to forward trace ID registration to the active capture strategy
- Updated `ReplaySegment` to include trace IDs in the replay payload
- Added validation to ignore empty trace IDs

### Implementation Details:
- Trace IDs are collected in `currentTraceIds` set during a segment's lifetime
- After a segment is created, the trace IDs are cleared for the next segment
- Empty trace IDs (`SentryId.EMPTY_ID`) are filtered out
- The feature gracefully handles cases where replay is not recording

This enables better correlation between replays and transactions. When a transaction occurs while replay is recording, its trace ID is now associated with the replay segment, making it possible to:
- Search for replays by trace ID
- Link replays to specific transactions in the UI
- Improve debugging workflows by connecting performance data with visual recordings

Example replay: https://sentry-sdks.sentry.io/explore/replays/?project=5428559&statsPeriod=24h&query=trace:f2b899ed30f0476ebea8a7678998cc8b

I'm able to search by trace id:

Google Chrome 2026-05-28 21 31 21

Closes #5346

Added comprehensive unit tests covering:
- `registerTraceId` includes trace IDs in the next segment
- `registerTraceId` clears trace IDs after segment creation
- `registerTraceId` ignores empty trace IDs
- `registerTraceId` does nothing when replay is not started
- `registerTraceId` forwards to capture strategy when recording
- `captureTransaction` registers trace ID with replay controller

All tests pass and verify the trace ID registration flow end-to-end.

## :pencil: Checklist

https://claude.ai/code/session_012wjHQtsEPzcrxSMCufxrDY

### DIFF
--- a/sentry-android-replay/src/main/java/io/sentry/android/replay/capture/BaseCaptureStrategy.kt
+++ b/sentry-android-replay/src/main/java/io/sentry/android/replay/capture/BaseCaptureStrategy.kt
@@ -33,6 +33,7 @@ import io.sentry.transport.ICurrentDateProvider
 import java.io.File
 import java.util.Date
 import java.util.Deque
+import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.ConcurrentLinkedDeque
 import java.util.concurrent.Executors
 import java.util.concurrent.ScheduledExecutorService
@@ -96,6 +97,7 @@ internal abstract class BaseCaptureStrategy(
   override var replayType by persistableAtomic<ReplayType>(propertyName = SEGMENT_KEY_REPLAY_TYPE)
 
   protected val currentEvents: Deque<RRWebEvent> = ConcurrentLinkedDeque()
+  protected val currentTraceIds: MutableSet<String> = ConcurrentHashMap.newKeySet()
 
   override fun start(segmentId: Int, replayId: SentryId, replayType: ReplayType?) {
     cache = replayCacheProvider?.invoke(replayId) ?: ReplayCache(options, replayId)
@@ -135,8 +137,10 @@ internal abstract class BaseCaptureStrategy(
     screenAtStart: String? = this.screenAtStart,
     breadcrumbs: List<Breadcrumb>? = null,
     events: Deque<RRWebEvent> = this.currentEvents,
-  ): ReplaySegment =
-    createSegment(
+  ): ReplaySegment {
+    val traceIds = currentTraceIds.toList().ifEmpty { null }
+    currentTraceIds.clear()
+    return createSegment(
       scopes,
       options,
       duration,
@@ -152,7 +156,9 @@ internal abstract class BaseCaptureStrategy(
       screenAtStart,
       breadcrumbs,
       events,
+      traceIds,
     )
+  }
 
   override fun onConfigurationChanged(recorderConfig: ScreenshotRecorderConfig) {
     this.recorderConfig = recorderConfig
@@ -164,6 +170,12 @@ internal abstract class BaseCaptureStrategy(
       if (rrwebEvents != null) {
         currentEvents += rrwebEvents
       }
+    }
+  }
+
+  override fun registerTraceId(traceId: SentryId) {
+    if (traceId != SentryId.EMPTY_ID) {
+      currentTraceIds.add(traceId.toString())
     }
   }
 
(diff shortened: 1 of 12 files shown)


sentry-android-replay/src/main/java/io/sentry/android/replay/capture/BaseCaptureStrategy.kt
```diff
@@ -135,8 +139,14 @@ internal abstract class BaseCaptureStrategy(
...
     events: Deque<RRWebEvent> = this.currentEvents,
-  ): ReplaySegment =
-    createSegment(
+  ): ReplaySegment {
+    val traceIds =
+      synchronized(traceIdsLock) {
+        val ids = currentTraceIds.toList().ifEmpty { null }
+        currentTraceIds.clear()
+        ids
+      }
```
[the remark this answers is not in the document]
romtsn: this is fine because those trace_ids are going to be irrelevant to the following segments anyway

sentry-android-replay/src/main/java/io/sentry/android/replay/capture/BaseCaptureStrategy.kt
```diff
@@ -54,6 +54,8 @@ internal abstract class BaseCaptureStrategy(
 ) : CaptureStrategy {
   internal companion object {
     private const val TAG = "CaptureStrategy"
+    // https://github.com/getsentry/sentry-javascript/blob/30eb68fff5077211c30c61ba74625e66ab514870/packages/replay-internal/src/coreHandlers/handleAfterSendEvent.ts#L41
```
0xadam-brown, line 57: 👍

sentry/src/main/java/io/sentry/SentryClient.java
```diff
@@ -1043,6 +1043,13 @@ public void captureSession(final @NotNull Session session, final @Nullable Hint
       sentryId = SentryId.EMPTY_ID;
     }
 
+    if (!sentryId.equals(SentryId.EMPTY_ID)) {
+      final @Nullable SpanContext trace = transaction.getContexts().getTrace();
+      if (trace != null) {
+        options.getReplayController().registerTraceId(trace.getTraceId());
```
0xadam-brown, line 1049: l: I imagine you considered and decided against it, but as a newbie I'm curious to hear your thoughts about inverting our dependency btw SentryClient and ReplayIntegration, eg:

- creating an `@ApiStatus.Internal` callback inside SentryClient (eg, `onTransactionSent(SentryTransaction)`) where the current registerTraceId() call lives, and let the replay integration subscribe to it
- using an event processor (assuming that's sane and not way too heavy-weight)

0xadam-brown (approved):

sentry/src/main/java/io/sentry/SentryClient.java
romtsn: good point! event processors are unfortunately being called to early for this -- we want to register trace ids for txs that have been successfully sent over the wire. 

The `onTransactionSent` callback is a great idea -- but I'd like to approach this holistically and also introduce such hooks for events, logs and all other types of telemetry, so I filed #5484 to address it!

P.S. While I was typing the above, I realized that the callback approach might be challenging w.r.t the hybrid SDKs, because they just call `captureEnvelope` as opposed to calling the respective methods for different types of telemetry. It's possible to have `afterSendEnvelope` and unpack envelopes to see their contents, but it quickly becomes ugly I think.

0xadam-brown: Thanks for filing #5484 – nice! 

> It's possible to have afterSendEnvelope and unpack envelopes to see their contents, but it quickly becomes ugly I think.

Agreed – that would indeed be ugly. Is the trace ID in the envelope header? That'd be non-ugly to grab, but exposing `afterSendEnvelope` might feel like a one-off for convenience, rather than something we really believe in, given the lack of types to query against. 

If we want to believe, maybe the hybrids call the native hook, having mapped to native types? A bit more work, but more to make use of going forward.

Something to think about...

romtsn: yeah sounds like quite a bit more work and kind of diverging from our thin wrapper approach where we only expose lightweight types between hybrid<->native SDKs... but maybe it's a good thing for the future -- would have to brainstorm more use-cases for this kind of wiring

PR #5489, opened 2026-06-01
fix(replay): Keep replay recording during animations

### LINKED ISSUE #5404: Session Replay: Continuous animations prevent screenshot capture
### Integration

sentry-android

### Build System

Gradle

### AGP Version

N/A

### Proguard

Enabled

No

### Version

8.38.0

### Steps to Reproduce

When an Activity contains continuous animations (e.g. a looping bounce animation on a UI element), Session Replay fails to capture new screenshot frames. The replay shows user gestures/interactions correctly, but the visual stream freezes on the last successfully captured frame, resulting in a stale/stuck image for the duration of the animated screen.

Steps to Reproduce

- Create an Activity with a continuously looping animation (e.g. a small image bouncing up and down via an infinite ObjectAnimator or similar).
- Initialize Sentry with Session Replay enabled.
- Navigate to the animated Activity.
- Observe the replay, the screen will appear frozen while gesture events continue to be recorded.
- Navigate away from the animated Activity — captures resume.

### Expected Result

Session Replay should capture screenshot frames even when continuous animations are running. The 1fps capture should still produce valid frames regardless of ongoing invalidations from animations.

### Actual Result

The replay visual stream freezes for the entire duration the user is on a screen with continuous animations. Logcat with options.debug = true and options.sessionReplay.debug = true shows:

```
Capturing a frame.
Capturing screenshot, isCapturing: true
Capturing screenshot, contentChanged: true, lastCaptureSuccessful: false
Posting the capture runnable again, frame rate is 1 fps.
Failed to determine view hierarchy, not capturing
```

romtsn:
## :scroll: Description
Update Session Replay's PixelCopy strategy so continuously invalidating screens do not freeze after the first unstable capture. The strategy now skips only the first unstable PixelCopy/view-hierarchy pair, then resumes emitting frames while the screen keeps drawing.

Add Android sample screens for Lottie, Compose canvas, and classic View animations so the behavior can be tested manually in the sample app. The sample app remains configured to use the Canvas replay screenshot strategy.

Continuous animations can cause PixelCopy output and the view hierarchy snapshot to be taken from slightly different moments. The previous debounce treated this as persistent instability and stopped emitting updated frames, which made replay visuals freeze on animated screens.

Fixes #5404

./gradlew :sentry-android-replay:testDebugUnitTest --tests io.sentry.android.replay.screenshot.PixelCopyStrategyTest
./gradlew spotlessApply apiDump
./gradlew :sentry-samples:sentry-samples-android:installDebug

Example replay: https://sentry-sdks.sentry.io/explore/replays/e6d4d1d93cff44cead71d10ba9d54c59/?t=9

## :pencil: Checklist
- [x] No new PII added or SDK only sends newly added PII if sendDefaultPII is enabled.

## :crystal_ball: Next steps


sentry-android-replay/src/main/java/io/sentry/android/replay/screenshot/PixelCopyStrategy.kt
```diff
@@ -40,6 +40,10 @@ internal class PixelCopyStrategy(
   private val markContentChanged: () -> Unit = {},
 ) : ScreenshotStrategy {
 
+  private companion object {
+    const val MAX_UNSTABLE_CAPTURES_TO_SKIP = 1
```
romtsn, line 48: could've been a boolean, but I decided that a counter is more natural and we may want to increase it in the future if we see problems with this

0xadam-brown, line 48: l: Bonus points for a quick KDoc explaining what an unstable capture is + why we want to cap the number of unstable captures we skip.

0xadam-brown (approved):
Implementation looks good.

Any concerns about masking slips during animation? 

Nicely, I see [our docs flag the possibility](https://docs.sentry.io/platforms/android/session-replay/#screenshot-strategy):

> This strategy has slightly lower performance overhead but may result in masking misalignments due to the asynchronous nature of the PixelCopy API. We recommend using this strategy for apps that do not have strict PII requirements or do not require masking functionality.

Not sure if we want the extra security, but we could also consider:

1. increasing the mask size when recording unstable captures, or
2. covering the view's entire range of possible positions during the unstable window, ie, the union of where it was (in the pixels, time T) and where the hierarchy thinks it is (T+delta).

Up to you.

PR #5501, opened 2026-06-05
perf(android): [SDK Overhead Reduction 2] Replace reflective OptionsContainer with direct subclass

adinauer:
## PR Stack (SDK Overhead Reduction)

- #5499
- #5500
- #5501
- #5502
- #5587
- #5589
- #5591
- #5598
- #5599
- #5600
- #5601
- #5602
- #5603
- #5609

---

## :scroll: Description

Replace `OptionsContainer.create(SentryAndroidOptions.class)` — which uses `getDeclaredConstructor().newInstance()` — with a direct `SentryAndroidOptionsContainer` subclass that returns `new SentryAndroidOptions()` without reflection.

To enable subclassing, `OptionsContainer` is made non-final (`@Open`) with a `protected` no-arg constructor. The `clazz` field becomes `@Nullable` with a null check in `createInstance()` for the reflective path (used by non-Android callers).

## :bulb: Motivation and Context

Part of the SDK Overhead Reduction stack. Avoids reflective constructor invocation on the Android cold-start path. While the improvement is below the noise floor in isolation (AR-71 in autoresearch report), it removes unnecessary reflection and is a clean, risk-free change.

## :green_heart: How did you test it?

- `./gradlew :sentry:test` — all tests pass
- `./gradlew :sentry-android-core:testDebugUnitTest` — all tests pass
- `./gradlew spotlessApply apiDump` — API change is only the expected non-final + protected ctor

## :pencil: Checklist
- [ ] I added GH Issue ID _&_ Linear ID
- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [ ] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

## :crystal_ball: Next steps

More SDK overhead reduction PRs in this stack.

#skip-changelog

> ⚠️ **Merge this PR using a merge commit** (not squash). Only the collection branch is squash-merged into main.

PR #5500, opened 2026-06-05
perf(core): [SDK Overhead Reduction 1] Skip java.specification.version lookup on Android

adinauer:
## PR Stack (SDK Overhead Reduction)

- #5499
- #5500
- #5501
- #5502
- #5587
- #5589
- #5591
- #5598
- #5599
- #5600
- #5601
- #5602
- #5603
- #5609

---

## :scroll: Description

In the `Platform.java` static initializer, when running on Android, skip the `System.getProperty("java.specification.version")` lookup and `Double.valueOf` parse, directly setting `isJavaNinePlus = false`. Android is never Java 9+, so this work is unnecessary on the Android cold-start path.

The JVM code path is unchanged.

## :bulb: Motivation and Context

Part of a series of SDK overhead reduction changes identified through automated Perfetto trace benchmarking. While this individual change is below the noise floor in isolation (~0.6% improvement), it removes unnecessary work from the Android startup critical path and is a clean, risk-free optimization.

## :green_heart: How did you test it?

- `./gradlew :sentry:test` — all tests pass
- `./gradlew spotlessApply apiDump` — no API surface changes
- Benchmarked via Perfetto traces (AR-72 in autoresearch report)

## :pencil: Checklist
- [ ] I added GH Issue ID _&_ Linear ID
- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [ ] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

## :crystal_ball: Next steps

More SDK overhead reduction PRs in this stack.

#skip-changelog

> ⚠️ **Merge this PR using a merge commit** (not squash). Only the collection branch is squash-merged into main.

PR #5502, opened 2026-06-05
perf(android): [SDK Overhead Reduction 3] Reuse SentryExecutorService for DeviceInfoUtil pre-caching

adinauer:

- #5499
- #5500
- #5501
- #5502
- #5587
- #5589
- #5591
- #5598
- #5599
- #5600
- #5601
- #5602
- #5603
- #5609

---

## :scroll: Description

Replace `Executors.newSingleThreadExecutor()` with `options.getExecutorService()` in `DefaultAndroidEventProcessor` for the `DeviceInfoUtil` pre-caching task. This avoids creating a dedicated thread + executor during SDK init when the SDK already has an executor service available. Also removes the `executorService.shutdown()` call since we don't own the shared executor.

Part of the SDK Overhead Reduction stack. During Android SDK init, `DefaultAndroidEventProcessor` was creating its own `SingleThreadExecutor` just to submit one task for pre-caching device info. The SDK's executor service (`SentryExecutorService`) is already available at this point and can handle this work without the overhead of spinning up an additional thread pool.

- `./gradlew :sentry-android-core:compileDebugUnitTestJavaWithJavac` — compiles successfully

## :pencil: Checklist

## :crystal_ball: Next steps

#skip-changelog

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java
@@ -23,8 +23,6 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.RejectedExecutionException;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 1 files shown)


adinauer:
Turns out this is actually slowing down init.
The question is, do we still want to update this to avoid creating a new short lived thread?

Screenshot 2026-06-08 at 16 36 00

romtsn:
the change looks good to me (still would prefer not to spawn threads when unnecessary), but i'd be curious where the slowness comes from. Could you check the actual flamegraph in perfetto?

adinauer:
> Could you check the actual flamegraph in perfetto?

Will do, likely tomorrow.

adinauer:
@romtsn I just checked the perfetto traces and it basically sums up to:
while this avoids creating a short lived thread it adds work to the single threaded shared SentryExecutorService on SentryOptions and causes other queued startup work to be delayed

Depending on the device and its load this might add a significant chunk to startup. Maybe loading this lazily is an option, I'll investigate a bit more.

adinauer:
A quick investigation suggests we'd be running the risk of ANRs when doing this work lazily on first use since it might run on main thread then.

I still wanted to investigate whether we could add async event processing. This would be a good candidate to move but might also be a breaking change for customers who expect the info to already be present when their callbacks kick in.

romtsn:
> @romtsn I just checked the perfetto traces and it basically sums up to: while this avoids creating a short lived thread it adds work to the single threaded shared SentryExecutorService on SentryOptions and causes other queued startup work to be delayed

hm, do we wait for SentryExecutorService on the main thread somewhere? I guess even if other work gets queued up (but on a bg thread) it shouldn't affect the main thread startup time?

adinauer:
Closing this for now as it makes more sense to move this to lazy invocation if/when we have async processing.

PR #5503, opened 2026-06-05
fix(replay): Fix Compose masking on obfuscated/minified builds

romtsn:
## 📜 Description

Fixes Compose view masking on obfuscated/minified (R8) builds, plus adds a CI mechanism to catch such regressions early.

- **Masking fix**: the SDK accesses internal Compose `LayoutNode` members to walk the node tree. On Compose < 1.10 these are looked up via reflection by their mangled names (`getChildren$ui_release`, `getOuterCoordinator$ui_release`, `getCollapsedSemantics$ui_release`). Since they aren't referenced directly, R8 could strip or rename them, silently breaking masking on minified builds. Added consumer proguard keep rules for them.
- **Fail-fast in CI**: added an internal `SentryReplayDebug.failFast` switch (gated on the `io.sentry.replay.compose.fail-fast` system property) that re-throws the exceptions `ComposeViewHierarchyNode` normally swallows. It's enabled in the sample app and the on-device `ReplayTest`/`ReplaySnapshotTest`, so our release/obfuscated builds running on real devices in Sauce Labs crash instead of silently under-masking. Defaults off, so customers are unaffected.
- Guard the on-device replay tests against GitHub-hosted emulators, which can't capture screenshots reliably.

## 💡 Motivation and Context

On obfuscated/minified builds, R8 could rename/strip the reflected Compose internals, which silently degraded or disabled Compose masking. Because the SDK swallows these errors to avoid crashing customer apps, the regression was invisible in our CI until now.

## 💚 How did you test it?

Unit tests + on-device `ReplayTest`/`ReplaySnapshotTest` running on release/obfuscated builds on real devices (Sauce Labs), which now fail fast if masking breaks.

Example replay: https://sentry-sdks.sentry.io/explore/replays/6783b073aef442a8b4ad8731bf6901fa

## 📝 Checklist

### DIFF
--- a/sentry-android-replay/src/main/java/io/sentry/android/replay/util/SentryReplayDebug.kt
+++ b/sentry-android-replay/src/main/java/io/sentry/android/replay/util/SentryReplayDebug.kt
@@ -1,0 +1,26 @@
+package io.sentry.android.replay.util
+
+/**
+ * Internal, undocumented escape hatch used to make Session Replay fail fast instead of silently
+ * degrading masking when an exception is swallowed (e.g. unsupported/obfuscated Compose internals).
+ *
+ * It is intended to be enabled only in our own sample/UI-test apps that run on real devices in CI
+ * (which are release/obfuscated builds, so [io.sentry.android.replay.BuildConfig.DEBUG] can't be
+ * used), so that regressions surface as crashes rather than under-masked replays. Customers should
+ * never set this.
+ *
+ * Enable via:
+ * ```
+ * System.setProperty("io.sentry.replay.compose.fail-fast", "true")
+ * ```
+ */
+internal object SentryReplayDebug {
+  private const val FAIL_FAST_PROPERTY = "io.sentry.replay.compose.fail-fast"
+
+  /**
+   * Read live (not cached) so it's only evaluated on the error path and unit tests can toggle it
+   * between cases.
+   */
+  val failFast: Boolean
+    get() = "true".equals(System.getProperty(FAIL_FAST_PROPERTY), ignoreCase = true)
+}
--- a/sentry-android-replay/src/main/java/io/sentry/android/replay/viewhierarchy/ComposeViewHierarchyNode.kt
+++ b/sentry-android-replay/src/main/java/io/sentry/android/replay/viewhierarchy/ComposeViewHierarchyNode.kt
@@ -22,6 +22,7 @@ import io.sentry.SentryLevel
 import io.sentry.SentryMaskingOptions
 import io.sentry.android.replay.SentryReplayModifiers
 import io.sentry.android.replay.util.ComposeTextLayout
+import io.sentry.android.replay.util.SentryReplayDebug
 import io.sentry.android.replay.util.boundsInWindow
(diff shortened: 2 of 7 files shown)


sentry-android-replay/src/main/java/io/sentry/android/replay/util/SentryReplayDebug.kt
```diff
@@ -0,0 +1,26 @@
+package io.sentry.android.replay.util
+
+/**
+ * Internal, undocumented escape hatch used to make Session Replay fail fast instead of silently
+ * degrading masking when an exception is swallowed (e.g. unsupported/obfuscated Compose internals).
+ *
+ * It is intended to be enabled only in our own sample/UI-test apps that run on real devices in CI
+ * (which are release/obfuscated builds, so [io.sentry.android.replay.BuildConfig.DEBUG] can't be
+ * used), so that regressions surface as crashes rather than under-masked replays. Customers should
```
runningcode, line 9: Theoretically customers _could_ want to test this in their own CI setups.

sentry-android-replay/src/main/java/io/sentry/android/replay/viewhierarchy/ComposeViewHierarchyNode.kt
runningcode, line 134: not related to your PR but we shouldn't catch `Throwable` as it could be an OOM/stackoverflow and then we're masking/obfuscating the source of those. it is better to catch only exceptions we know to be thrown. In this case it would probably be better to just switch to `Exception` at the very least so we don't catch OOM/stackoverflows. I can make a PR to make this change.

runningcode (approved):
That's a good way to enhance our tests for this!

sentry-android-replay/src/main/java/io/sentry/android/replay/viewhierarchy/ComposeViewHierarchyNode.kt
romtsn: In this case I think it's on purpose to catch LinkageError/AbstractMethodError/NoSuchMethodError, since this is a backward/forward compatibility codepath.

There was also a deliberate decision to do this across the SDK to also cover gaming SDKs which throw Errors (afair), see #1806 for more context. 

That said, I think you are right, and we should still audit the catch blocks throughout the SDK and see where they do more harm than good. Thanks for opening #5513 👍

sentry-android-replay/src/main/java/io/sentry/android/replay/util/SentryReplayDebug.kt
romtsn: good point! I'll keep it internal for now though, and see if anybody is interested. They could still set it via System.setProperty

sentry-android-replay/src/main/java/io/sentry/android/replay/viewhierarchy/ComposeViewHierarchyNode.kt
runningcode: Sounds good and thanks for that context. I also added it to our agenda for wednesday.

PR #5507, opened 2026-06-08
fix(replay): Fix VerifyError in Compose masking under DexGuard/R8 obfuscation

### LINKED ISSUE #5497: Session Replay: VerifyError in ComposeViewHierarchyNode (androidx.compose.ui.geometry.Rect) under DexGuard/R8 obfuscation
### Integration

sentry-android

### Build System

Gradle

### AGP Version

Release build with DexGuard 9.13.2 (aggressive obfuscation/optimization)

### Proguard

Enabled

No

### Version

8.25.0 (the affected code path is unchanged on main/8.43.0 — see below)

### Steps to Reproduce

Steps to Reproduce:
1. Compose app (androidx.compose.ui 1.7.2), Session Replay enabled.
2. Release build with aggressive obfuscation (DexGuard / R8 full mode).
3. Launch the app; Replay captures the first Compose screen.

### Expected Result

Replay captures the Compose screen without crashing.

### Actual Result

Fatal VerifyError at class load when Replay traverses the Compose tree:
`java.lang.VerifyError: Verifier rejected class io.sentry.android.replay.viewhierarchy.<obf>:
  ... <m>(androidx.compose.ui.node.LayoutNode, <obf>, int, boolean, io.sentry.SentryOptions) failed to verify:
  [0x65] 'this' argument 'Unresolved Reference: androidx.compose.ui.geometry.Rect' not instance of 'Reference: <obf>'
    at io.sentry.android.replay.util.ViewsKt.traverse(Views.kt:36)
    at io.sentry.android.replay.screenshot.PixelCopyStrategy...onPixelCopyFinished(...)
    at android.view.PixelCopy$1.lambda$onCopyFinished$0(PixelCopy.java:193)
`

Deobfuscated, the rejected method maps to ComposeViewHierarchyNode.fromComposeNode(...).

What I've found so far:
- fromComposeNode → boundsInWindow() (util/Nodes.kt) → localBoundingBoxOf(...), which returns androidx.compose.ui.geometry.Rect, the type the verifier reports as unresolved.
- Our best guess is that this Compose type reference isn't surviving the app's obfuscation, so ART can't verify the class and rejects it at load. We haven't confirmed the exact mechanism — sharing it in case it's familiar to you.
- For what it's worth, this doesn't look like the Compose 1.10 / #5189 case: the app is on Compose 1.7.2 and the unresolved symbol is geometry.Rect, not getChildren. Possibly the same family as the obfuscation VerifyError in gradle-plugin #618 and #5302.

Note on upgrading: The boundsInWindow → localBoundingBoxOf → Rect reference is still present on main/8.43.0, and sentry-android-replay/proguard-rules.pro doesn't appear to keep the Compose geometry/layout/node classes, so upgrading alone may not resolve it. Flagging in case a keep-rule or guarding change is needed; we'll defer to you on the right fix.

romtsn:
## 📜 Description

`ComposeViewHierarchyNode` traversal computed node bounds via `LayoutCoordinates.boundsInWindow(...)`, which returned an `android.graphics.Rect` while the surrounding code treated the value as `androidx.compose.ui.geometry.Rect` — mixing the two `Rect` types within `fromComposeNode(...)`.

Under aggressive obfuscation/optimization (DexGuard 9.13.2 / R8 full mode) this could be rejected at class verification time with a `VerifyError` (`'Unresolved Reference: androidx.compose.ui.geometry.Rect' not instance of ...`), crashing Session Replay the moment it traversed the first Compose screen.

Fix:
- `boundsInWindow(...)` now consistently returns `androidx.compose.ui.geometry.Rect` (float-based), so a single `Rect` type flows through the method.
- Added an `internal fun Rect.toRect(): android.graphics.Rect` extension to convert only at the boundary where the view-hierarchy node actually needs `android.graphics.Rect`.

## 💡 Motivation and Context

Fixes #5497 — Session Replay `VerifyError` in `ComposeViewHierarchyNode` under DexGuard/R8 obfuscation.

## 💚 How did you test it?

Existing Session Replay Compose unit tests; built the replay module in release. The crash itself only reproduces under aggressive obfuscation (DexGuard / R8 full mode), where the mixed `Rect` types tripped the verifier; collapsing to a single `Rect` type removes the ambiguity.

Example replay: https://sentry-sdks.sentry.io/explore/replays/6783b073aef442a8b4ad8731bf6901fa

## 📝 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Session Replay: Fix `VerifyError` in Compose masking under DexGuard/R8 obfuscation ([#5507](https://github.com/getsentry/sentry-java/pull/5507))
+
 ## 8.43.1
 
 ### Fixes
--- a/sentry-android-replay/src/main/java/io/sentry/android/replay/util/Nodes.kt
+++ b/sentry-android-replay/src/main/java/io/sentry/android/replay/util/Nodes.kt
@@ -2,8 +2,8 @@
 
 package io.sentry.android.replay.util
 
-import android.graphics.Rect
 import androidx.compose.ui.geometry.Offset
+import androidx.compose.ui.geometry.Rect
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.ColorProducer
 import androidx.compose.ui.graphics.painter.Painter
@@ -176,7 +176,7 @@ internal fun LayoutCoordinates.boundsInWindow(rootCoordinates: LayoutCoordinates
   val boundsBottom = bounds.bottom.fastCoerceIn(0f, rootHeight)
 
   if (boundsLeft == boundsRight || boundsTop == boundsBottom) {
-    return Rect()
+    return Rect(0.0f, 0.0f, 0.0f, 0.0f)
   }
 
   val topLeft = root.localToWindow(Offset(boundsLeft, boundsTop))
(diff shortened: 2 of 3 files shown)


runningcode (approved):
Nice!

PR #5514, opened 2026-06-08
chore(android-sqlite): Skip wrapping SupportSQLiteDriver bridge to avoid duplicate spans

0xadam-brown:
## :scroll: Description

PR protects against generating duplicate `db.sql.query` spans when using the `androidx.sqlite.driver.SupportSQLiteDriver` bridge by refusing to wrap the support driver and requiring users to wrap the `SupportSQLiteOpenHelper` it consumes instead.

The bridge is the sole Room (or SQLDelight) API that leaves open the possibility of using `SentrySQLiteDriver` and `SentrySupportSQLiteOpenHelper` together with the same db file – a possibility that would lead to redundant spans. 

### Example misuse this PR guards against

```kotlin
// Step 1: Developer wraps their open helper with Sentry, either manually or
// via the Sentry Android Gradle Plugin.
val sentryWrappedHelper: SupportSQLiteOpenHelper =
    SentrySupportSQLiteOpenHelper.create(
        FrameworkSQLiteOpenHelperFactory().create(configuration)
    )

// Step 2: Developer builds the support driver around that wrapped helper.
val driver: SQLiteDriver = SupportSQLiteDriver(sentryWrappedHelper)

// Step 3: Developer (wrongly) wraps the bridge with Sentry as well. All
// spans would be duplicated were it not for this PR, which returns 
// SupportSQLiteDrivers unwrapped.
val sentryWrappedDriver: SQLiteDriver = SentrySQLiteDriver.create(driver)

Room.databaseBuilder(context, MyDb::class.java, "mydb")
    .setDriver(sentryWrappedDriver)
    .build()
```

Addresses JAVA-275

## ⚠️ Callouts 

**[1]** **I've covered the no-span edge case via documentation:** It's possible to use `SupportSQLiteDriver` without wrapping the open helper passed via its constructor, resulting in no spans being produced. That a temporary concern, as `SupportSQLiteDriver` goes away with Room 3.0+; the blast radius should also be limited, as most of our users rely on the Sentry Android Gradle Plugin to auto-wrap the open helper. I've flagged the issue in the CHANGELOG and the `SentrySQLiteDriver` KDoc.

**[2]** **Support driver has open helper span semantics:** Users who rely on `SupportSQLiteDriver` will have their spans auto-generated via the open helper, and so the span semantics will match its behavior rather than that of SQLiteDriver (for more details about SQLiteDriver's span policy, see [here](https://github.com/getsentry/sentry-java/pull/5466#discussion_r3317374806)). (Wrapping the SQLiteDriver would've required reflection, which I wanted to avoid for performance reasons and to make it easier to eventually lift the driver into KMP common.)

**[3]** **Changes in this PR are needed if we want to use ASM auto-wrapping**: I've a forthcoming PR that introduces ASM auto-wrapping of `SQLiteDriver` via the Sentry Android Gradle Plugin. The approach in this PR makes that auto-wrapping safe, as SAGP will always convert the `SQLiteDriver` passed to `RoomDatabase.Builder.setDriver(SQLiteDriver)` into `SentrySQLiteDriver.create(SQLiteDriver)`. If the provided driver is a `SupportSQLiteDriver`, this PR ensures `SentrySQLiteDriver.create(SQLiteDriver)` will return it unwrapped.

**[1]** Created a new "Bridge" section in the Android sample app that wires up the `SupportSQLiteDriver` and provides it with both a `SentrySupportSQLiteOpenHelper` and a `SentrySQLiteDriver`:

Bridge section

**[2]** Verified via the Sentry UI that spans are not duplicated: 

_Baseline: Existing behavior of SentrySupportSQLiteOpenHelper_

Open helper baseline

_This PR: Behavior of SupportSQLiteDriver_

Bridge - no duplicate spans

## :pencil: Checklist


sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/sqlite/SampleDatabases.kt
```diff
@@ -156,22 +259,47 @@ object SampleDatabases {
...
+          openHelperRoomDb(appContext).songDao().also { it.primeWriter() }.count()
+        }
+        runWarmUpStep("SQLDelight", failures) {
+          SampleSQLDelightDatabase(sqlDelightDriver(appContext))
+            .songQueries
+            .countSongs()
+            .executeAsOne()
+        }
+        warmUpErrors = failures.joinToString("\n") { "Warm-up failed: $it" }
+        warmUpComplete = true
```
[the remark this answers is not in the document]
0xadam-brown: Addressed in 4aacc7079da54dfa53bdb4beae8411ff5d8a4db8.

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SentrySQLiteDriver.kt
```diff
@@ -68,8 +67,23 @@ public class SentrySQLiteDriver private constructor(private val delegate: SQLite
...
+    /**
+     * Name of the bridge adapter often used with Room 2.7+. It implements the `SQLiteDriver`
+     * interface and its constructor consumes a `SupportSQLiteOpenHelper`. (Users of the Sentry
+     * Android Gradle Plugin will have the `SupportSQLiteOpenHelper` wrapped for them
+     * automatically.) We deliberately avoid wrapping the adapter to prevent duplicate spans.
+     *
+     * String (rather than an `is` check) lets us avoid a compile-time dependency on
+     * androidx.sqlite:sqlite-framework.
+     */
+    private const val SUPPORT_SQLITE_DRIVER_FQN = "androidx.sqlite.driver.SupportSQLiteDriver"
```
markushi, line 79: Just as a safety net we should add the class to proguard as well, to ensure this runtime check works as expected. (`-keepnames` to the `proguard-rules.pro` file). This might be covered by the androidx libs already, but I'd still add it nevertheless.

sentry-android-sqlite/src/test/java/androidx/sqlite/driver/SupportSQLiteDriver.kt
```diff
@@ -0,0 +1,18 @@
+package androidx.sqlite.driver
+
+import androidx.sqlite.SQLiteConnection
+import androidx.sqlite.SQLiteDriver
+
+/**
+ * Minimal stub of `androidx.sqlite.driver.SupportSQLiteDriver` (which lives in
```
markushi, line 7: 👍

markushi (approved):
LGTM!

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SentrySQLiteDriver.kt
romtsn: +1, we should add the rule here just in case

romtsn (approved):

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SentrySQLiteDriver.kt
0xadam-brown: Good callout. Can't imagine AndroidX doesn't keep the names of its public API, but doesn't hurt to make sure. Updated.

0xadam-brown: Just checked with Claude and (fwiw) it wasn't able to find ProGuard rules in the AAR or otherwise, so I may have been wrong and we really did need this. Thx again!

PR #5520, opened 2026-06-10
ref(core): Reduce unnecessary boxing and redundant null checks (JAVA-554)

runningcode:
## :scroll: Description

Small, behavior-preserving cleanups applied consistently across the core `sentry` module:

- Replace boxed `Long.valueOf(a).compareTo(b)` with `Long.compare(a, b)` (and `order.compareTo(...)` on a boxed `Long` field), avoiding unnecessary boxing.
- Drop the redundant `x != null && x instanceof Type` guards, since `instanceof` already returns `false` for `null`.
- Remove other unnecessary boxing flagged by inspection: pass primitives to `String.format`, cast `long`→`double` instead of `Double.valueOf`, and use `Double.parseDouble` for the Java version check.

Files touched:
- `SentryDate.java`, `SentryNanotimeDate.java` — `Long.compare`
- `internal/eventprocessor/EventProcessorAndOrder.java` — `Long.compare`
- `SpanFactoryFactory.java`, `ScopesStorageFactory.java`, `util/LifecycleHelper.java`, `protocol/Contexts.java` — redundant null check removal
- `CircularFifoQueue.java`, `DateUtils.java`, `util/Platform.java` — unnecessary boxing removal

Started as a small refactor in `SentryDate`/`SentryNanotimeDate` and applied the same patterns everywhere else they appeared in the core module. Pure readability/cleanup, no behavior change.

Resolves JAVA-554.

## :pencil: Checklist

## :crystal_ball: Next steps

None.

### DIFF
--- a/sentry/src/main/java/io/sentry/DateUtils.java
+++ b/sentry/src/main/java/io/sentry/DateUtils.java
@@ -115,7 +115,7 @@ public final class DateUtils {
    * @return date rounded down to milliseconds
    */
   public static Date nanosToDate(final long nanos) {
-    final Double millis = nanosToMillis(Double.valueOf(nanos));
+    final Double millis = nanosToMillis((double) nanos);
     return getDateTime(millis.longValue());
   }
 
@@ -137,7 +137,7 @@ public final class DateUtils {
    * @return seconds
    */
   public static double nanosToSeconds(final long nanos) {
-    return Double.valueOf(nanos) / (1000.0 * 1000.0 * 1000.0);
+    return (double) nanos / (1000.0 * 1000.0 * 1000.0);
   }
 
   /**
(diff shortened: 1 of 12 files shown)


sentry/src/main/java/io/sentry/DateUtils.java
```diff
@@ -115,7 +115,7 @@ public static double nanosToMillis(final double nanos) {
    * @return date rounded down to milliseconds
    */
   public static Date nanosToDate(final long nanos) {
-    final Double millis = nanosToMillis(Double.valueOf(nanos));
+    final Double millis = nanosToMillis((double) nanos);
```
0xadam-brown, line 118: l: We could avoid boxing altogether via `return getDateTime((long) nanosToMillis((double) nanos)`

(Maybe the compiler saves us anyway, but the updated version would guarantee as much.)

0xadam-brown (approved):
One more box to remove if you want it; otherwise lgtm 👍

markushi (approved):
LGTM!

sentry/src/main/java/io/sentry/DateUtils.java
runningcode: Nice find. i will do this in a follow up given we want to make a release.

PR #5524, opened 2026-06-10
perf(core): Use fixed-delay scheduling for performance collector (JAVA-555)

runningcode:
## :scroll: Description

Switch the transaction collection timer in `DefaultCompositePerformanceCollector` from `Timer.scheduleAtFixedRate` to `Timer.schedule` (fixed-delay scheduling).

Fixes JAVA-555.

`scheduleAtFixedRate` schedules tasks relative to the *initial* start time, so after a delay or GC pause it fires a burst of rapid catch-up executions to "make up" for missed runs. The old code defended against this with a 10ms skip check, since collections spaced closer than that are not meaningful. In addition, it is using `System.currentTimeMillis()` which isn't a monotonic clock meaning it could go forwards and then guard is broken allowing all the tasks to run.

`schedule` uses fixed-delay scheduling: each collection is scheduled 100ms after the previous one *finishes*. Catch-up bursts can no longer happen, so the guard and its supporting state are redundant. The result is more even collection spacing and simpler code.

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,13 +6,16 @@
 
 - Reduce unboxing in `DateUtils.nanosToDate` ([#5523](https://github.com/getsentry/sentry-java/pull/5523))
 
+### Fixes
+
+- Fix performance collector scheduling many tasks in a row ([#5524](https://github.com/getsentry/sentry-java/pull/5524))
+
 ## 8.43.2
 
 ### Improvements
 
 - Improve SDK init performance by replacing `java.net.URI` with custom string parsing for DSN ([#5448](https://github.com/getsentry/sentry-java/pull/5448))
 - Remove unnecessary boxing to improve performance ([#5520](https://github.com/getsentry/sentry-java/pull/5520))
-- Use fixed-delay scheduling for the performance collector to avoid catch-up collection bursts ([#5524](https://github.com/getsentry/sentry-java/pull/5524))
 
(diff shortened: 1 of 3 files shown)


sentry/src/main/java/io/sentry/DefaultCompositePerformanceCollector.java
```diff
@@ -112,16 +111,8 @@ public void run() {
             new TimerTask() {
               @Override
               public void run() {
-                long now = System.currentTimeMillis();
```
runningcode, line 115: The issue is that `scheduleAtFixedRate` could schedule thousands of tasks in a row if the system goes in to deep sleep but then we guard against it using this `System.currentTimeMillis()` which isn't a monotonic clock meaning it could go forwards and then guard is broken allowing thousands of tasks to run.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 
 ### Fixes
 
+- Fix performance collector scheduling many tasks in a row ([#5524](https://github.com/getsentry/sentry-java/pull/5524))
```
romtsn: need to rebase on main and move this to Unreleased again

romtsn (approved):

CHANGELOG.md
runningcode: done

PR #5538, opened 2026-06-12
chore(samples-android): Add optional SAGP build mode

0xadam-brown:
## :scroll: Description

Adds an optional `useSagp` flag to Android sample app builds that, when true, applies the Sentry Android Gradle Plugin. 

Defaults to false, which matches existing build behavior. (Flag omitted == false)

```
./gradlew :sentry-samples:sentry-samples-android:installDebug -PuseSagp=true
```

See the new README for more details.

Lets us verify SAGP functionality against our Android sample app demos (e.g., auto-instrumentation of SQLite queries, etc.).

Manually. 

## :pencil: Checklist

## :crystal_ball: Next steps

- Use mavenLocal SAGP artifacts in preference to libs.versions.toml ([#5539](https://github.com/getsentry/sentry-java/pull/5539))

### DIFF
--- a/sentry-samples/sentry-samples-android/README.md
+++ b/sentry-samples/sentry-samples-android/README.md
@@ -1,0 +1,57 @@
+# Sentry Sample Android App
+
+Sample application demonstrating how to use the Sentry Android SDK, including core functionality (error reporting, tracing, session replay,
+profiling) and integrations (Compose, OkHttp, etc.).
+
+## How to run it?
+
+Install the app on your device or emulator:
+
+```
+./gradlew :sentry-samples:sentry-samples-android:installDebug
+```
+
+or simply open the project in Android Studio and run the `sentry-samples-android` configuration.
+
+You can also apply the [Sentry Android Gradle Plugin](https://github.com/getsentry/sentry-android-gradle-plugin) (SAGP) when building (not applied by default):
+
+```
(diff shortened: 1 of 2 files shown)


romtsn (approved):
LGTM, but anything against just adding it by default and omitting the flag altogether? The features of the plugin can be then independently tweaked via its options

0xadam-brown:
> LGTM, but anything against just adding it by default and omitting the flag altogether? The features of the plugin can be then independently tweaked via its options

I went the flag route b/c i) I wanted to be conservative (in case others were relying on SAGP not being applied), and ii) it makes it easy for us to swizzle between manual vs auto-instrumentation (something I'm planning on taking advantage of for SQLite in a follow-on PR).

If we always apply SAGP, we'd need to introduce one or more flags for enabling / disabling auto-instrumentation or else rely on folks to locally adjust the SAGP config block. Either is certainly doable.

**Four options:**

1. Keep the -PuseSagp flag (ie, don't apply SAGP by default but let folks opt in).
2. Invert the flag to -PskipSagp (ie, apply SAGP by default but let folks opt out).
3. Go with instrumentation-specific flags + always apply SAGP.
4. Require folks to update instrumentation locally + always apply SAGP.

I'll merge as-is for now. But let me know if you prefer one of the above and I'll be happy to post a follow-up PR. (Personally I like option (2) b/c it preserves the toggle – which I imagine will be useful – but gives the default it sounds like most people may want.)

PR #5539, opened 2026-06-12
chore(samples-android): Support mavenLocal for builds that apply the SAGP

0xadam-brown:
Follow-on to https://github.com/getsentry/sentry-java/pull/5538

## :scroll: Description

Adds wiring that lets us prefer mavenLocal SAGP artifacts, when present, for Android sample app builds that set `-PuseSagp=true`. If no local artifact is found, we fall back to `libs.versions.toml`.

Lets us use local / under-development versions of SAGP with the Android sample app without having to locally modify `settings.gradle` every time out.

Manually.

## :pencil: Checklist


gradle/libs.versions.toml
```diff
@@ -29,6 +29,7 @@ otelInstrumentationAlpha = "2.26.0-alpha"
 otelSemanticConventions = "1.40.0"
 otelSemanticConventionsAlpha = "1.40.0-alpha"
 retrofit = "2.9.0"
+sagp = "6.6.0"
```
runningcode: should we use `6.10.0`?

sentry-samples/sentry-samples-android/build.gradle.kts
```diff
@@ -230,3 +234,46 @@ abstract class ToggleNativeLoggingTask : Exec() {
...
+      ?.toURI()
+      ?.let { File(it).name }
+      ?.removePrefix("sentry-android-gradle-plugin-")
+      ?.removeSuffix(".jar")
+
+  val sagpLocalJar =
+    sagpVersion?.let {
+      File(
+        System.getProperty("user.home"),
+        ".m2/repository/io/sentry/sentry-android-gradle-plugin/$it/sentry-android-gradle-plugin-$it.jar",
```
runningcode: Do you know if we can rely on the artifact always being here? This seems prone to breaking.

Is this configuration cache compatible?

settings.gradle.kts
```diff
@@ -2,6 +2,15 @@ enableFeaturePreview("TYPESAFE_PROJECT_ACCESSORS")
 
 pluginManagement {
     repositories {
+        // Prefer local SAGP artifact if one exists; otherwise fall back to libs.versions.toml.
+        if (providers.gradleProperty("useSagp").orNull.equals("true", ignoreCase = true)) {
```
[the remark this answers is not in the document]
runningcode: this is true.

sentry-samples/sentry-samples-android/build.gradle.kts
```diff
@@ -230,3 +234,46 @@ abstract class ToggleNativeLoggingTask : Exec() {
       )
   }
 }
+
+fun Project.logSagpOrigin() {
+  // A locally published SAGP in ~/.m2 silently shadows the released artifact (see README,
+  // "Testing an unpublished SAGP build"); we log so developers don't wonder what's going on.
```
runningcode: This seems prone to breaking. I would just make sure to use a unique version for publishing and then you are sure that you are using it.
If you are still unsure you can check a build scan to verify where it came from.

gradle/libs.versions.toml
0xadam-brown: Updated

settings.gradle.kts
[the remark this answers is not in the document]
0xadam-brown: Updated in 864bda30613dee4d7e90e60958c5be2e7a7a714f.

sentry-samples/sentry-samples-android/build.gradle.kts
0xadam-brown: Fyi, I've adopted your sane suggestion of relying on the unique coordinate published to mavenLocal. That let me strip out the cleverness here and the logging method 👍

sentry-samples/sentry-samples-android/build.gradle.kts
0xadam-brown: (see [above](https://github.com/getsentry/sentry-java/pull/5539#discussion_r3412985386))

0xadam-brown:
Addressed @runningcode's (always helpful) comments in the latest commit. Ready for another / final pass.

settings.gradle.kts
```diff
@@ -1,10 +1,19 @@
 enableFeaturePreview("TYPESAFE_PROJECT_ACCESSORS")
 
 pluginManagement {
-    repositories {
-        mavenCentral()
-        gradlePluginPortal()
+  repositories {
+    // Prefer local SAGP artifact if one exists; otherwise fall back to libs.versions.toml.
+    if (providers.gradleProperty("useSagp").orNull.equals("true", ignoreCase = true)) {
```
runningcode: We might want to enable mavenLocal for other reasons than just testing SAGP. In that case I think we should name this `enableMavenLocal` or something like that.


Code wise, it might also be a bit simpler if you check for the presence of the property instead of a true false. Scroll to the kotlin/groovy example here: https://docs.gradle.org/current/userguide/build_environment.html#setting_a_project_property

```suggestion
    if (providers.gradleProperty("useSagp").isPresent) {
```

runningcode (approved):
Looks great! Just one comment worth considering but can be merged as is.

settings.gradle.kts
0xadam-brown: > Code wise, it might also be a bit simpler if you check for the presence of the property instead of a true false.

Nice! Updated.

> We might want to enable mavenLocal for other reasons than just testing SAGP. In that case I think we should name this enableMavenLocal or something like that.

I've punted for now. I want to keep the diff narrow and I'm (slightly) worried the enableMavenLocal flag will require more explanation than it's worth (it only applies to plugins, not dependencies in general), and it either requires us to continue to specify included groups for safety, or we increase the risk the build includes a stale plugin. If we have a second use, let's go for it then 👍

PR #3427, opened 2024-05-16
Fix faulty span.frame_delay calculation for early app start spans

markushi:
## :scroll: Description

The return value of `frameMetricsCollector.getLastKnownFrameStartTimeNanos()` was ignored, thus creating large frame delay values.

I also moved data collection post initial `tracer.finish`, ensuring a finish timestamp is set and spans / txns are already trimmed. This is especially useful for idle `ui.load` transactions, as otherwise the frame delay will be longer than the txn duration.

## :pencil: Checklist

- [ ] I reviewed the submitted code.
- [ ] I added tests to verify the changes.
- [ ] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] Review from the native team if needed.
- [ ] No breaking change or entry added to the changelog.
- [ ] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

PR #5543, opened 2026-06-12
chore(android-sqlite): Repair start times of spans generated by SentrySQLiteDriver

0xadam-brown:
## :scroll: Description

Repairs the `nanoTimetamp` of the `SentryNanotimeDate`s used as start times for the spans generated by `SentrySQLiteDriver`. 

For more details, see the discussion [here](https://github.com/getsentry/sentry-java/pull/5504#issuecomment-4679631245). This PR implements approach [1] ("Offset SQLite spans based nano duration between parent and child start times").

Without those repairs, all spans within a given wall clock millisecond are displayed by Sentry UI as starting at that same millisecond and are re-ordered arbitrarily. Often that's quite confusing as actual BEGIN -> EXECUTE STATEMENT -> END sequences can appear as EXECUTE STATEMENT -> END -> BEGIN (etc.).

JAVA-275

### Before

Before - R3

([Link](https://sentry-sdks.sentry.io/explore/traces/trace/03687888e02e42b084878b912094c9ec/?cursor=0%3A150%3A0&fov=0.11925692974129336%2C19.303783807108257&node=span-b49f7331113b4aa9&project=5428559&source=traces&statsPeriod=1h&targetId=b49f7331113b4aa9&timestamp=1781292263))

### After

After 2

([Link](https://sentry-sdks.sentry.io/explore/traces/trace/e31450e09ab34260adbf8dc81ae7079d/?fov=0%2C19.295166015625&node=span-ffa78c9a366648e6&project=5428559&source=traces&statsPeriod=1h&targetId=60f5c5cba09f4912&timestamp=1781267783))

### Before

Before - Room2

([Link](https://sentry-sdks.sentry.io/explore/traces/trace/9568c28cd9ca44b8bf2479acbd5dafbf/?cursor=0%3A150%3A0&node=span-c1945133853e4cfb&project=5428559&source=traces&statsPeriod=1h&targetId=c1945133853e4cfb&timestamp=1781295128))

### After

After 3

([Link](https://sentry-sdks.sentry.io/explore/traces/trace/2551a0b2967f4f32b947630faf9381dd/?cursor=0%3A150%3A0&node=span-36b40e265f8547c9&project=5428559&source=traces&statsPeriod=1h&targetId=2d1ea6b6e28f439d&timestamp=1781268738))

- Unit tests
- Via the sample app + verifying span appearance in Sentry UI (see screenshots above)

## :pencil: Checklist

## :crystal_ball: Next steps

Consider implementing nesting of spans as mentioned [here](https://github.com/getsentry/sentry-java/pull/5504#issuecomment-4691614092).


sentry-android-sqlite/src/main/java/io/sentry/sqlite/SQLiteSpanInstrumentation.kt
```diff
@@ -28,26 +29,33 @@ internal class SQLiteSpanInstrumentation(
...
 
-  /** Records a `db.sql.query` span from [startTimestamp] to [startTimestamp] + [durationNanos]. */
-  fun recordSpan(
+  /**
+   * Records a `db.sql.query` span from [startTimestamp] to the moment of invocation.
+   *
+   * "Coarse" in that it doesn't provide a nanosecond precision for [SentryNanotimeDate]
+   * [startTimestamp]s.
+   */
+  fun recordCoarseSpan(
```
0xadam-brown: Chime in if you can think of a better name. Not in love with this, but it'll do if it has to...

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SQLiteSpanInstrumentation.kt
```diff
@@ -74,6 +82,38 @@ internal class SQLiteSpanInstrumentation(
...
+   *
+   * Repair ensures proper ordering and lets the spans stagger:
+   * ```
+   * Parent span                 ├█████████████┤
+   * BEGIN IMMEDIATE TRANSACTION  ├████┤         0.25 ms
+   * INSERT INTO `my_db` …              ├██┤     0.10 ms
+   * END TRANSACTION                     ├███┤   0.18 ms
+   * ```
+   */
+  private fun SentryDate.repairPrecision(baseline: SentryDate?): SentryDate =
```
0xadam-brown: This arguably belongs on `SentryDate`, especially as the latter's API is already enmeshed in the broken precision of `SentryDate.nanoTimestamp`. But I struggled to get it to work. I considered:

**[1]** adding a `SentryDate.laterDateByDiff()` that returns a `SentryLongDate` (declared as a `SentryDate`). 

...but that seemed heavyweight for a single call site (new public method, tests against all implementing classes), esp as the only thing it buys us is a replacement for the already minimal line 112. Plus, bumping the minSdkVersion to 26 will let us move off of `SentryNanotimeDate`, so any investments in it will be short-lived. 

**[2]** adding a `repairPrecision()` method solely to `SentryNanotimeDate` (allowing us to relocate all of the `repair` method). 

...but that'd also be a new public API, and it'd mean we'd have to cast at the call site, which would be against the grain of how we typically use `SentryDate`.

So I left things as you see them, figuring we could shuffle bits around if we ever get a second call site. 

Let me know if you have any opinions.

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SQLiteSpanInstrumentation.kt
```diff
@@ -28,36 +30,44 @@ internal class SQLiteSpanInstrumentation(
...
   fun recordSpan(
     sql: String,
     startTimestamp: SentryDate,
+    durationNanos: Long,
     status: SpanStatus,
     throwable: Throwable? = null,
   ) {
-    recordSpan(sql, startTimestamp, endTimestamp = null, status, throwable)
+    val parent = scopes.span ?: return
+    val nanoPrecisionStart = startTimestamp.repairPrecision(baseline = parent.startDate)
```
0xadam-brown: Repairing the incoming timestamp is the only interesting change this PR makes. (The diff looks a bit bigger because I switched the order of the first and second methods and renamed the second recordCoarseSpan.)

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SentrySQLiteStatement.kt
runningcode, line 47: There's a lot of catching throwables in this PR. Could you note what type of exceptions are expected here? I think it will help us later for our audit.

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SentrySQLiteDriver.kt
runningcode, line 42: Do you kno why do we only catch `LinkageError` here but `Throwable` in other places?
I prefer catching the narrowest `Exception` as is best practice in java but I'm just curious as to the decision.

And should we log this to help dev's debug this issue?

sentry-android-sqlite/src/main/java/io/sentry/android/sqlite/SQLiteSpanManager.kt
```diff
@@ -41,10 +41,10 @@ internal class SQLiteSpanManager(
       if (result is CrossProcessCursor) {
         return SentryCrossProcessCursor(result, this, sql) as T
       }
-      spans.recordSpan(sql, startTimestamp, SpanStatus.OK)
+      spans.recordCoarseSpan(sql, startTimestamp, SpanStatus.OK)
       result
     } catch (e: Throwable) {
-      spans.recordSpan(sql, startTimestamp, SpanStatus.INTERNAL_ERROR, e)
+      spans.recordCoarseSpan(sql, startTimestamp, SpanStatus.INTERNAL_ERROR, e)
       throw e
```
runningcode, line 57: This refactor makes it a bit hard to review just the time spans change

sentry-android-sqlite/src/test/java/io/sentry/sqlite/SQLiteSpanInstrumentationTest.kt
```diff
@@ -162,32 +283,176 @@ class SQLiteSpanInstrumentationTest {
...
+    val start = sut.startTimestamp()
+    sut.recordSpan("SELECT 1", start, 1_000_000, SpanStatus.OK)
+
+    val span = fixture.sentryTracer.children.first()
+    assertEquals("sqlite", span.data[SpanDataConvention.DB_SYSTEM_KEY])
+    assertEquals("tracks.db", span.data[SpanDataConvention.DB_NAME_KEY])
+  }
+
+  @Test
+  fun `fromDatabaseName sets db name from databaseName when using recordSpan`() {
```
runningcode: what's the value of this test?

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SQLiteSpanInstrumentation.kt
```diff
@@ -28,36 +30,44 @@ internal class SQLiteSpanInstrumentation(
    */
   fun startTimestamp(): SentryDate = scopes.options.dateProvider.now()
```
runningcode: If we require nanosecond precision, why can't we use a class here that requires nanosecond precision?

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SQLiteSpanInstrumentation.kt
```diff
@@ -74,26 +84,44 @@ internal class SQLiteSpanInstrumentation(
...
+   * ```
+   * Parent span                 ├█████████████┤
+   * BEGIN IMMEDIATE TRANSACTION  ├████┤         0.25 ms
+   * INSERT INTO `my_db` …              ├██┤     0.10 ms
+   * END TRANSACTION                     ├███┤   0.18 ms
+   * ```
+   */
+  private fun SentryDate.repairPrecision(baseline: SentryDate?): SentryDate =
+    if (baseline is SentryNanotimeDate) {
+      SentryLongDate(baseline.laterDateNanosTimestampByDiff(this))
```
runningcode: Seems expensive to create yet another object for every span. Creating a `SentryNanodateDate` object is already expensive and now we're creating another one.

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SQLiteSpanInstrumentation.kt
runningcode: The date rabbithole gets deeper.

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SentrySQLiteStatement.kt
```diff
@@ -9,9 +9,10 @@ import io.sentry.SpanStatus
...
  *
- * Span duration is restricted to accumulated database time, i.e., each [step] call is individually
- * timed and the durations are summed. Time the application spends between steps (e.g., processing
- * rows, sleeping, or doing I/O) is intentionally excluded.
+ * Span duration is purposefully restricted to accumulated database time, i.e., each [step] call is
+ * individually timed and the durations are summed. Time the application spends between steps (e.g.,
+ * processing rows, sleeping, or doing I/O) is intentionally excluded so the span accurately
+ * represents how long SQLite itself was working.
  *
  * Not thread-safe: assumes sequential access within each SQL statement (normal SQLite usage).
```
runningcode, line 15: good to note

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SentrySQLiteStatement.kt
0xadam-brown: Happy to do that in a follow-on 👍  

(Fyi, these files briefly included in the diff after I merged the underlying branch to `main` and before I had a chance to rebase here. Apologies! The current diff is now up-to-date.)

runningcode (commented):
Thanks for looking at this tricky problem. I might be missing some context and I realize that the dates in our codebase are a mess.

Is it possible to pass around timestamp offsets as `long` everywhere? This would keep maximum precision and improve performance.

I realize that we lose type safety here, but creating two date objects in `recordSpan` in order to do arithmetic seems quite heavy when these transactions require nanosecond precision.

Otherwise if using a `long` isn't possible, what stops us from using a more precise date object like `SentryNanotimeDate` instead of `SentryDate`?

sentry-android-sqlite/src/main/java/io/sentry/android/sqlite/SQLiteSpanManager.kt
0xadam-brown: (see comment [here](https://github.com/getsentry/sentry-java/pull/5543#discussion_r3419179193))

0xadam-brown:
@runningcode – thanks much for your comments. I'll reply presently. 

Fyi, looks like you reviewed this in the few minutes between my merging the underlying branch + retargeting this PR. Apologies! Fyi, I've re-pushed. What you saw that was relevant to this PR is in the first commit (unchanged). What little is new (just making the key repair method `internal` so I can write better tests against it) is in the second commit. No need to look again until I've replied to your comments.

0xadam-brown:
@runningcode – great questions! Answers below

> Is it possible to pass around timestamp offsets as `long` everywhere? This would keep maximum precision and improve performance.

Yes! but there are few extra pieces we need to account for. In particular, three constraints show up if we want to record a span start time with ns-precision:

1. we need a wall clock instant in order to peg the timestamp to the epoch;
2. we need a nanosecond to compute the offset from the wall clock instant (the instant + offset is our timestamp's ns-precise value); and
3. we have to pass a SentryDate to the span creation API b/c it requires as much.

At present we're getting (1) and half of (2) from the parent span (which is created by other code we don't control, so we can ignore its cost). We get the other half of (2) from a SentryDate we request from the globally available SentryDateProvider... 

...and it's the creation of that provided date that's wasteful, not the creation of the SentryLongDate in the repair method. (Requirement (3) means the creation of a SentryDate has to happen _somewhere_, and we know we need a SentryLongDate b/c it's the only way on Android to capture a ns-precise timestamp. So the best we can do is to create a single SentryLongDate and skip the SentryNanotimeDate from the provider.)

The nub, then, becomes extracting the nanos value from the parent without the (otherwise superfluous) SentryNanotimeDate. That's tricky b/c SentryNanotimeDate doesn't expose its nanos field directly and the relevant diff API expects another SentryDate instance. 

But @markushi has [shown the way](https://github.com/getsentry/sentry-java/pull/3427) in SpanFrameMetricsCollector 🎉.  We can copy the approach there and allocate a single SentryNanotimeDate when SQLiteSpanInstrumentation is initiated (cf. [here](https://github.com/getsentry/sentry-java/blob/773a0df13db7654c534b8775902021af234fa3d5/sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java#L36) and [here](https://github.com/getsentry/sentry-java/blob/773a0df13db7654c534b8775902021af234fa3d5/sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java#L320)).

...and that should do it! 

Thanks for challenging this point. I didn't think past the expectations of the diff API in my original approach, and so I didn't question the idea of getting the SentryDate from the provider. Good stuff 💯 

> Otherwise if using a `long` isn't possible, what stops us from using a more precise date object like `SentryNanotimeDate` instead of `SentryDate`?

Fyi, we're already using SentryNanotimeDate everywhere on Android. Part of what makes that class confusing is that its nanoTimestamp() method is misnamed. The 'timestamp' part is correct; the 'nano' part is sometimes wrong on the JVM and always wrong an Android. IMO, it'd be better if it were called instant() or epochalTimestamp(), b/c its return value represents the wall clock instant ticked off from the Unix epoch at whatever resolution the platform allows.  

Note: on Java 9+ Instant.now() actually provides sub-ms precision (sometimes microsecond, sometimes nanosecond, depending on the JVM, IIRC). (See the SentryInstantDate implementation, which relies on Instant.) So if the world consisted solely of SentryInstantDate running on JVMs targeting Java 9+, we'd see a microsecond- or ns-precise epochal timestamp returned by nanoTimestamp() and the name would make sense. The problem is pre-Java 9 environments still see ms-precision, as does anything before Android 33, as that was the first Android API to introduce sub-ms precision with Instant.now(). (Ten years from now when our minSdk is 33 😅 , we can get rid of SentryNanotimeDate and use SentryInstanceDate everywhere.)

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SQLiteSpanInstrumentation.kt
```diff
@@ -28,36 +30,45 @@ internal class SQLiteSpanInstrumentation(
    */
   fun startTimestamp(): SentryDate = scopes.options.dateProvider.now()
 
-  /** Records a `db.sql.query` span from [startTimestamp] to the moment of invocation. */
+  /** Records a `db.sql.query` span from [startTimestamp] to [startTimestamp] + [durationNanos]. */
   fun recordSpan(
     sql: String,
     startTimestamp: SentryDate,
```
runningcode: I guess, more specifically why does this have to be a `SentryDate` not a `SentyNanodateTime` ?

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SQLiteSpanInstrumentation.kt
0xadam-brown: (addressed [below](https://github.com/getsentry/sentry-java/pull/5543#issuecomment-4717937873))

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SentrySQLiteDriver.kt
0xadam-brown: Good question. This is the fix to the general `compileOnly` problem. In our case, we compile against a specific version of SQLite, but we don't put that version on the owning app's classpath, so we can't be certain about what we'll see at runtime. The `hasConnectionPool` property is a recent addition to this SQLite interface, so we have to protect against its not actually existing on the delegate.

Sharing your sensibilities about `Throwable`, I didn't want to catch more than what we needed.

(As mentioned in our sync, I'll create PR extracting this logic into a utility method so we can document the problem.)

EDIT:
> And should we log this to help dev's debug this issue?

I don't think so. If we catch, we fall back to the behavior of the library default method (return false). That's well-behaved from the developer's perspective rather than deviant.

sentry-android-sqlite/src/test/java/io/sentry/sqlite/SQLiteSpanInstrumentationTest.kt
0xadam-brown: Not much. I'll see if there's anything worth rescuing, otherwise I'll zap it (and its siblings).

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SQLiteSpanInstrumentation.kt
```diff
@@ -97,3 +108,39 @@ internal class SQLiteSpanInstrumentation(
...
+ * INSERT INTO `my_db` …              ├██┤     0.30 ms
+ * END TRANSACTION                     ├███┤   0.33 ms
+ * ```
+ */
+internal fun SentryDate.repairPrecision(anchor: SentryDate?): SentryDate =
+  if (anchor is SentryNanotimeDate) {
+    // Compute a new timestamp with nanosecond precision by using the anchor as the epoch instant
+    // and adding to it the diff of this.nanos - anchor.nanos.
+    SentryLongDate(anchor.laterDateNanosTimestampByDiff(this))
+  } else {
```
runningcode: When does the else case happen?

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SQLiteSpanInstrumentation.kt
```diff
@@ -97,3 +108,39 @@ internal class SQLiteSpanInstrumentation(
...
+ * BEGIN IMMEDIATE TRANSACTION  ├████┤         0.02 ms
+ * INSERT INTO `my_db` …              ├██┤     0.30 ms
+ * END TRANSACTION                     ├███┤   0.33 ms
+ * ```
+ */
+internal fun SentryDate.repairPrecision(anchor: SentryDate?): SentryDate =
+  if (anchor is SentryNanotimeDate) {
+    // Compute a new timestamp with nanosecond precision by using the anchor as the epoch instant
+    // and adding to it the diff of this.nanos - anchor.nanos.
+    SentryLongDate(anchor.laterDateNanosTimestampByDiff(this))
```
runningcode: I think we've had this exact problem you are describing in this PR given this `laterDateNanosTimestampByDiff` exists. Maybe @adinauer can shed some light on the overall issue in the PR?

runningcode:
If we are already using SentryNanodateTime everywhere on Android as you say, can we enforce that using types if it would simplify the code?

runningcode (approved):
I'll approve this since it is an improvement. We should probably have another reviewer take a look given the complexity.

0xadam-brown:
> If we are already using SentryNanodateTime everywhere on Android as you say, can we enforce that using types if it would simplify the code?

That's a great question. The short answer is no b/c, when I said we're using SentryNanotimeDate everywhere on Android, I was being hand-wavy. We occasionally use SentryLongDate too. 

SentryNanotimeDate is used "everywhere" in the sense that the universally-available SentryDateProvider (`scopes.options.dateProvider`) on Android always exposes a SentryAndroidDateProvider, and that in turn always vends a SentryNanotimeDate. But nothing stops Android code (or a user) from hand-rolling SentryLongDate's. And sometimes we do just that (eg, `ActivityLifecycleIntegration` does so for some app start transactions).

That's^^ the global reason. W/r/t local declarations, analyzing whether SentryNanotimeDate is safe or SentryDate is needed one method at a time is probably too burdensome in most scenarios (esp when the source of those dates originates from a `SentryDate` declaration we can't change, eg, `scopes.span`). But folks could always do so if it ever makes sense.

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SQLiteSpanInstrumentation.kt
0xadam-brown: (See [here](https://github.com/getsentry/sentry-java/pull/5543#issuecomment-4728231739).)

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SQLiteSpanInstrumentation.kt
0xadam-brown: (See [here](https://github.com/getsentry/sentry-java/pull/5543#issuecomment-4728231739).)

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SQLiteSpanInstrumentation.kt
0xadam-brown: Updated in 776deac39ba92a65ba8bd431c8aae3a043d8e665. We no longer allocate when creating the start timestamp. Now we only create the (inevitable) two `SentryLongDate` instances for the start and end timestamps required by the Span API.

sentry-android-sqlite/src/test/java/io/sentry/sqlite/SQLiteSpanInstrumentationTest.kt
0xadam-brown: Removed in 776deac39ba92a65ba8bd431c8aae3a043d8e665.

sentry-android-sqlite/src/main/java/io/sentry/android/sqlite/SQLiteSpanManager.kt
```diff
@@ -3,17 +3,21 @@ package io.sentry.android.sqlite
 import android.database.CrossProcessCursor
 import android.database.SQLException
 import io.sentry.IScopes
+import io.sentry.ISpan
```
0xadam-brown, line 6: _**You can ignore this class's diff: I'm just restoring it to its previous state.**_

When implementing @runningcode's (excellent) idea to return a `Long` from `SQLiteSpanInstrumentation.startTimestamp()` rather than `SentryDate`, it became apparent that routing both the open helper and the driver through `SQLiteSpanInstrumentation` for the sake of DRY'ing out a few lines of span config code was more pain / confusion than it was worth. 

So I've: 

- Reverted `SQLiteSpanManager` (this class) to its exact state before I introduced `SentrySQLiteDriver`. (That guarantees no behavioral changes for the legacy path when we release the driver.)
- Simplified `SQLiteSpanInstrumentation` now that it no longer has to serve two clients. (The biggest change is that I was able to rip out a number of tests.)

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SQLiteSpanInstrumentation.kt
```diff
@@ -1,17 +1,26 @@
...
+import java.util.Date
 
 private const val SQLITE_TRACE_ORIGIN = "auto.db.sqlite"
 
-/** Shared span instrumentation for SQLite. */
+/**
+ * Sentinel for extracting a [SentryNanotimeDate]'s underlying [System.nanoTime] value via
+ * [SentryDate.diff].
+ */
+private val EMPTY_NANO_TIME = SentryNanotimeDate(Date(0), 0L)
```
0xadam-brown, line 21: It'd be nice to introduce an `@Internal` method on `SentryNanotimeDate` that exposes its nano time:

```java
@ApiStatus.Internal
public long getMonotonicNanos() {
  return nanos;
}
```

That'd^^ let us avoid the `EMPTY_NANO_TIME` hack here and in `SpanFrameMetricsCollector` where I stole this from (cf. [here](https://github.com/getsentry/sentry-java/blob/773a0df13db7654c534b8775902021af234fa3d5/sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java#L36) and [here](https://github.com/getsentry/sentry-java/blob/773a0df13db7654c534b8775902021af234fa3d5/sentry-android-core/src/main/java/io/sentry/android/core/SpanFrameMetricsCollector.java#L320)). But that's too big for this PR...

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SQLiteSpanInstrumentation.kt
```diff
@@ -20,44 +29,32 @@ internal class SQLiteSpanInstrumentation(
...
-    status: SpanStatus,
-    throwable: Throwable? = null,
-  ) {
-    recordSpan(sql, startTimestamp, endTimestamp = null, status, throwable)
-  }
+  fun startTimestamp(): Long =
+    // Try to retain nanosecond precision + avoid SentryDate allocation...
+    scopes.span?.childStartTimestampOrNull()
+      // ...otherwise fall back to millisecond precision + allocate.
+      ?: scopes.options.dateProvider.now().nanoTimestamp()
```
0xadam-brown, line 43: This is the big change in the latest commit: no longer automatically allocate a `SentryDate` for the start timestamp. Instead, return a Long with ns-precision if possible (and it usually will be). 

For more details, see [here](https://github.com/getsentry/sentry-java/pull/5543#issuecomment-4717937873).

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SQLiteSpanInstrumentation.kt
```diff
@@ -85,15 +82,43 @@ internal class SQLiteSpanInstrumentation(
...
+ *
+ * Nanosecond precision ensures proper ordering and lets the spans stagger:
+ * ```
+ * Parent span                 ├█████████████┤
+ * BEGIN IMMEDIATE TRANSACTION  ├████┤         0.02 ms
+ * INSERT INTO `my_db` …              ├██┤     0.30 ms
+ * END TRANSACTION                     ├███┤   0.33 ms
+ * ```
+ */
+internal fun ISpan.childStartTimestampOrNull(): Long? {
```
markushi: `nit`: I know naming is hard, would renaming it to `computeNanoStartTimestampForChild()` make it a bit more clear? Not sure if it's any better. but at least `compute` could give away that the value is based on the current time.

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SQLiteSpanInstrumentation.kt
```diff
@@ -85,15 +82,43 @@ internal class SQLiteSpanInstrumentation(
...
+ * Parent span                 ├█████████████┤
+ * END TRANSACTION              ├███┤          0.33 ms
+ * BEGIN IMMEDIATE TRANSACTION  ├████┤         0.02 ms
+ * INSERT INTO `my_db` …        ├██┤           0.30 ms
+ *                              ↑
+ *               (All spans share the same ms baseline
+ *             even though their execution was staggered)
+ * ```
+ *
+ * Nanosecond precision ensures proper ordering and lets the spans stagger:
```
markushi, line 106: Love this, especially for timing, makes it all super clear!

markushi (approved):
Looks great, thanks for following this rabbit hole down to the root cause.

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SQLiteSpanInstrumentation.kt
0xadam-brown: Good point! Updated in fee24613dcf5d516f43f2e27e5d9bf606a9acd17.

sentry-android-sqlite/src/main/java/io/sentry/sqlite/SQLiteSpanInstrumentation.kt
0xadam-brown: (Resolving so I can merge. Feel free to continue to convo if folks care to.)

PR #5545, opened 2026-06-15
fix(samples): allow leak canary for non-debug builds

markushi:
## :scroll: Description

This way the sample app won't crash when running profiling session via the Android Studio Profiler.

#skip-changelog


romtsn (approved):

0xadam-brown (approved):
🐦

Issue #5542, opened 2026-06-12
Attachments duplication


codelimit:
### Integration

sentry-android

### Build System

Gradle

### AGP Version

9.5.1

### Proguard

Enabled

### Other Error Monitoring Solution

No

### Version

8.43.2

### Steps to Reproduce

For sentry_samples_android app
1. Declare native function in **io/sentry/samples/android/NativeSample.java**
`public static native void captureException();`
2. Implement native function in **native-sample.cpp**
```
JNIEXPORT void JNICALL Java_io_sentry_samples_android_NativeSample_captureException(
    JNIEnv *env, jclass cls) {
  sentry_value_t event = sentry_value_new_event();
  sentry_value_t exc = sentry_value_new_exception("Exception", "Error message.");
  sentry_value_set_stacktrace(exc, NULL, 0);
  sentry_event_add_exception(event, exc);
  sentry_capture_event(event);
}
```
3. Add button to trigger the method in **io/sentry/samples/android/MainActivity.kt**:
```
  item {
    SentryTraced("capture_exception_native") {
      OutlinedButton(
        onClick = {
          NativeSample.captureException();
        },
        modifier = Modifier,
      ) {
        Text("Capture Exception Native", maxLines = 2, overflow = TextOverflow.Ellipsis)
      }
    }
  }
```
4. Build and run app, click "Capture Exception Native"
5. Check event's attachments

### Expected Result

Attachments are not duplicated

### Actual Result

Attachment duplication:

<img width="3076" height="638" alt="Image" src="https://github.com/user-attachments/assets/2b0d94df-4b84-4577-98aa-ef08b6d8d45e" />

codelimit:
As I understood, sentry native captures the exception, dumps the envlope to outbox directory. Then, EnvelopeFileObserver picks up and process envelope. Envelope has two items: event and attachment. When OutboxSender process event item, it enriches it with attachment from global scope (this is the first attachment instance), and then, sends second item, which is original attachment from sentry native envelope (second attachment instance).

codelimit:
The reproduction is even simplier: just click "Native Capture" button in "sentry samples android" app and check the issue page. It has two attachments.

runningcode:
Thanks for filing as well as the repro steps. We will take a look at this.

PR #5211, opened 2026-03-18
feat: Sync file attachments to native

bitsandfoxes:
## :scroll: Description
Relies on https://github.com/getsentry/sentry-native/pull/1584

Attachments added to the scope should get synced to `sentry-native`. I.e. a Unity games has 
```
SentrySdk.ConfigureScope(scope => scope.AddAttachment("special-logfile.log");
```
and the game crashes, that attachment should be available in the resulting sentry event.

## :bulb: Motivation and Context
Resolves https://github.com/getsentry/sentry-java/issues/4758

## :green_heart: How did you test it?
Unit tests. Unity SDK will have e2e tests.

## :pencil: Checklist

- [x] I added GH Issue ID _&_ Linear ID
- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [ ] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

## :crystal_ball: Next steps
This needs to be used and implemented in the hybrid SDKs.
Adding support to sync `bytes` as attachments.

PR #5548, opened 2026-06-16
fix(android): Stop duplicating attachments on native events (JAVA-559)

runningcode:
## :scroll: Description

Native (NDK) events no longer duplicate scope attachments.

When an attachment is added to the scope on the Java side, it is synced down to the native SDK via `NdkScopeObserver`. A native event captured through `sentry_capture_event` is then written to the outbox **already carrying that attachment as its own envelope item**. When `OutboxSender` re-ingests the cached envelope, it calls `captureEvent` with a hint that is both `Cached` and `ApplyScopeData`, so `SentryClient` re-applied the Java scope's attachments on top — sending each attachment twice.

The fix skips re-applying scope attachments when the event comes from a cached envelope (`!HintUtils.hasType(hint, Cached.class)`). The guard is intentionally narrow: it only gates `addScopeAttachmentsToHint`, leaving the rest of scope-data enrichment (contexts, tags, breadcrumbs, level, user) applied as before, which native crashes rely on.

Fixes attachment duplication on native/NDK events.

- GH: #5542
- Linear: JAVA-559

- Added a unit test in `SentryClientTest` asserting that a `Cached` + `ApplyScopeData` hint does not add scope attachments to the envelope.
- Reproduced manually in `sentry-samples-android`: tapping "Native Capture" with a scope attachment configured showed `sentry.png` twice before the fix and once after.

## :pencil: Checklist

## :crystal_ball: Next steps

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 
 ### Fixes
 
+- Fix attachments being duplicated on native events that carry scope attachments ([#5548](https://github.com/getsentry/sentry-java/pull/5548))
 - Fix performance collector scheduling many tasks in a row ([#5524](https://github.com/getsentry/sentry-java/pull/5524))
 
 ## 8.43.2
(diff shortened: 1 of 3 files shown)


romtsn (approved):
nice! probably a regression introduced with https://github.com/getsentry/sentry-java/pull/5211

PR #5550, opened 2026-06-16
perf(android): Replace Date with unix timestamp in SentryNanotimeDate (JAVA-533)

runningcode:
## :scroll: Description

`SentryNanotimeDate` stored a `java.util.Date` but only ever read its epoch millis (in `compareTo` and via `DateUtils.dateToNanos`). This replaces the stored `Date` with a `long unixDate` (milliseconds since the epoch), and the default constructor now calls `System.currentTimeMillis()` instead of `DateUtils.getCurrentDateTime()` (which allocates a `Calendar`).

This is **behavior-preserving**: `Date.getTime()` already returns epoch millis, and the UTC `Calendar` used by `getCurrentDateTime()` only affected calendar *field* access — never the absolute instant the class actually used. So `Calendar.getInstance(UTC).getTime().getTime() == System.currentTimeMillis()`.

`SentryNanotimeDate` is now marked `@ApiStatus.Internal` — it is the legacy `Date`+`nanoTime` precision workaround and is not intended for direct use by consumers (use `SentryLongDate`/`SentryInstantDate` to build a `SentryDate`). Its constructor changed from `(Date, long)` to `(long unixDate, long nanos)`, and the now-unused internal `DateUtils.dateToNanos(Date)` is removed. Per the repo's API policy, `@ApiStatus.Internal` types are not part of the public contract.

JAVA-533, part of the "Reduce SDK init time / allocations" effort. On Android, `SentryNanotimeDate` is **always** the active date provider (`SentryAndroidDateProvider` never switches to `Instant`), so `now()` runs this on every span/transaction timestamp. Removing the per-timestamp `Calendar` + `Date` allocation reduces CPU and GC pressure that scales with tracing volume.

### Performance

**JVM microbenchmark** (JDK 17, isolating the differing part of the constructor; both paths also call `System.nanoTime()`, excluded):

| Path | ns/op | bytes allocated/op |
|---|---|---|
| Old — `Calendar.getInstance(UTC).getTime()` | ~56 ns | **528 B** |
| New — `System.currentTimeMillis()` | ~14 ns | **0 B** |

≈ **4× faster** and **zero allocation** (the old path allocates a `GregorianCalendar` + a `Date` per call).

**On-device confirmation** — ART method trace, Pixel 3 (Android 12), cold start to idle. Same scenario, per-cold-start call counts:

| Method | Baseline (`Date`) | Change (`long`) |
|---|---|---|
| `SentryNanotimeDate` objects created | 16 | 16 (unchanged) |
| `DateUtils.getCurrentDateTime()` | 8 | 1 |
| `Calendar.getInstance(TimeZone)` | 8 | 2 |
| `GregorianCalendar.<init>` | 9 | 6 |

The eliminated `getCurrentDateTime`/`Calendar` calls map to the no-arg `SentryNanotimeDate()` constructions — the allocation chain is gone from the timestamp path while the number of `SentryNanotimeDate` objects is unchanged.

Average time to construct a `SentryNanotimeDate()` in the (instrumented) trace was **~372 µs baseline → ~24 µs change**. ⚠️ These are **instrumentation-inflated** (ART records every method entry/exit) and exaggerate both the absolute value and the ratio — the real per-call figure is the **~56 ns → ~14 ns** microbenchmark above. The trace's value here is the *call-count* evidence that the `Calendar`/`GregorianCalendar` allocations actually disappear on a real device.

Startup *time* is unaffected (only a handful of these calls during cold start); the win is allocation/GC reduction that scales with tracing volume.

Updated existing unit tests for the new constructor signature; full `sentry` and `sentry-android-core` test suites pass. Behavior verified equivalent (epoch-millis identical to the previous `Date`-based path).

## :pencil: Checklist

## :crystal_ball: Next steps

Other internal callers already migrated to the `long` constructor. Hybrid SDKs that constructed `SentryNanotimeDate` directly should adopt the new signature (or switch to `SentryLongDate`); the class is now `@ApiStatus.Internal`.

### DIFF
--- a/sentry/api/sentry.api
+++ b/sentry/api/sentry.api
@@ -3558,6 +3558,7 @@ public final class io/sentry/SentryMetricsEvents$JsonKeys {
 public final class io/sentry/SentryNanotimeDate : io/sentry/SentryDate {
 	public fun <init> ()V
 	public fun <init> (JJ)V
+	public fun <init> (Ljava/util/Date;J)V
 	public fun compareTo (Lio/sentry/SentryDate;)I
 	public synthetic fun compareTo (Ljava/lang/Object;)I
 	public fun diff (Lio/sentry/SentryDate;)J
--- a/sentry/api/sentry.api
+++ b/sentry/api/sentry.api
@@ -382,7 +382,6 @@ public final class io/sentry/DataCategory : java/lang/Enum {
 }
 
 public final class io/sentry/DateUtils {
-	public static fun dateToNanos (Ljava/util/Date;)J
 	public static fun dateToSeconds (Ljava/util/Date;)D
 	public static fun doubleToBigDecimal (D)Ljava/math/BigDecimal;
 	public static fun getCurrentDateTime ()Ljava/util/Date;
@@ -3558,7 +3557,7 @@ public final class io/sentry/SentryMetricsEvents$JsonKeys {
 
 public final class io/sentry/SentryNanotimeDate : io/sentry/SentryDate {
 	public fun <init> ()V
-	public fun <init> (Ljava/util/Date;J)V
+	public fun <init> (JJ)V
 	public fun compareTo (Lio/sentry/SentryDate;)I
 	public synthetic fun compareTo (Ljava/lang/Object;)I
 	public fun diff (Lio/sentry/SentryDate;)J
--- a/sentry/src/main/java/io/sentry/SentryNanotimeDate.java
+++ b/sentry/src/main/java/io/sentry/SentryNanotimeDate.java
@@ -1,5 +1,6 @@
 package io.sentry;
 
+import java.util.Date;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
@@ -24,6 +25,14 @@ public final class SentryNanotimeDate extends SentryDate {
 
   public SentryNanotimeDate() {
     this(System.currentTimeMillis(), System.nanoTime());
+  }
+
+  /**
+   * @deprecated use {@link SentryNanotimeDate#SentryNanotimeDate(long, long)} instead.
+   */
+  @Deprecated
+  public SentryNanotimeDate(final @NotNull Date date, final long nanos) {
+    this(date.getTime(), nanos);
   }
 
   public SentryNanotimeDate(final long unixDateMillis, final long nanos) {
--- a/sentry/src/main/java/io/sentry/SentryNanotimeDate.java
+++ b/sentry/src/main/java/io/sentry/SentryNanotimeDate.java
@@ -31,7 +31,7 @@ public final class SentryNanotimeDate extends SentryDate {
    * @deprecated use {@link SentryNanotimeDate#SentryNanotimeDate(long, long)} instead.
    */
   @Deprecated
-  @SuppressWarnings("InlineMeSuggester")
+  @SuppressWarnings({"InlineMeSuggester", "JavaUtilDate"})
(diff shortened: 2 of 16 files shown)


sentry/api/sentry.api
```diff
@@ -3558,7 +3557,7 @@ public final class io/sentry/SentryMetricsEvents$JsonKeys {
 
 public final class io/sentry/SentryNanotimeDate : io/sentry/SentryDate {
 	public fun <init> ()V
-	public fun <init> (Ljava/util/Date;J)V
+	public fun <init> (JJ)V
```
runningcode, line 3560: This is a breaking API change. I would hope that nobody is using this API/object though.

sentry/src/main/java/io/sentry/SentryNanotimeDate.java
```diff
@@ -1,32 +1,33 @@
...
  *
  * <p>This is a workaround for older versions of Java (before 9) and Android API (lower than 26)
- * that allows for higher precision than {@link Date} alone would.
+ * that allows for higher precision than a millisecond timestamp alone would.
  */
+@ApiStatus.Internal
 public final class SentryNanotimeDate extends SentryDate {
 
-  private final @NotNull Date date;
+  private final long unixDate;
```
markushi: `nit`: we usually always suffix fields with their unit. It's internal here, so the chances is misinterpretation are quite low. 
```suggestion
  private final long unixDateMillis;
```

markushi (approved):
LGTM!

sentry/src/main/java/io/sentry/SentryNanotimeDate.java
runningcode: good idea

sentry/api/sentry.api
0xadam-brown, line 3560: h: ⚠️ We'll be breaking anyone who used SentryNanotimeDate. (I'm not as worried about the DateUtils removal b/c that was already marked as `@Internal`; flagging it in the CHANGELOG seems sufficient if that's our normal practice.) 

I suspect there aren't many folks who use SentryNanotimeDate directly, but good hygiene would provide a bridge for existing callers + strip out the Date constructor with the next major (and mark it as deprecated in the meantime).

Thoughts?

sentry/src/main/java/io/sentry/SentryNanotimeDate.java
0xadam-brown: l: Consider `timestampMs` or `unixTimestampMs` (etc.).

Fwiw, I'm tripping up over "Date" b/c it makes me think of the Java type and civil time (days/months/years). It'd be great to have a name that highlights this field's role as our absolute time measure pegged to the Unix epoch, esp as that's a big part of what distinguishes it from `nanos`.

0xadam-brown (commented):
So good to see us getting rid of `Date`. Nice! 💯 

A few quick comments, one of which could block us. Let me know what you think.

sentry/api/sentry.api
romtsn: I was about to post the same -- I agree, i think we should mark the Date ctor deprecated and drop it in the next major, just to be safe (I know we accidentally forgot to mark the class as `@Internal` from the start, so it's on us, sorry!).

sentry/src/main/java/io/sentry/SentryNanotimeDate.java
runningcode: I already incorporated Max's suggestion here so I'm going to go with that: https://github.com/getsentry/sentry-java/pull/5550#discussion_r3427712194

sentry/api/sentry.api
runningcode: Sounds good. Added it in.

PR #5552, opened 2026-06-16
feat(android): Report app start reason on standalone app start transaction

buenaflor:
## :scroll: Description
Reports **why the OS started the process** on the standalone app start transaction, using `ApplicationStartInfo.getReason()` (Android 15 / API 35+).

- Adds `AppStartMetrics.getAppStartReason()`, which maps the already-cached `ApplicationStartInfo` reason to a stable lowercase string (`launcher`, `launcher_recents`, `start_activity`, `broadcast`, `service`, `content_provider`, `job`, `alarm`, `push`, `backup`, `boot_complete`, `other`). Returns `null` on API < 35, when no start info was resolved, or for an unmapped value.
- Attaches it as `app.vitals.start.reason` trace data on the standalone `app.start` transaction in both paths: foreground/activity (`onActivityCreated`) and headless/non-activity (`onHeadlessAppStart`).
- Also registers a `StandaloneAppStart` marker in the SDK metadata `integrations` when the feature is enabled (internal only, no changelog) so we can tell from events that standalone app start tracing was active.

Only standalone app start transactions are affected; this rides the existing `enableStandaloneAppStartTracing` opt-in and adds no new option.

Builds on standalone app start tracing (#5342). The reason is a high-value dimension for standalone starts: it distinguishes a real user launch (`launcher` / `start_activity`) from headless starts (`broadcast`, `service`, `content_provider`, `job`, …) — exactly the disambiguation the headless app start path was built for. The reason (why the process started) is orthogonal to cold/warm (`getStartType()`), so both are reported.

`ApplicationStartInfo` is already cached in `AppStartMetrics` from the standalone work, so this only reads one additional field (`getReason()`) off the same object — no extra system calls.

Unit tests in `sentry-android-core`:
- `AppStartMetricsTestApi35`: reason → string mapping (API 35), `null` when no start info, `null` for an unmapped reason value.
- `ActivityLifecycleIntegrationTest`: `app.vitals.start.reason` present on the foreground standalone transaction (API 35), absent when unavailable, and present on the headless standalone transaction.

Note: the headless-vs-foreground detection in `PerformanceAndroidEventProcessor` keys off the *absence* of `app.vitals.start.screen`; the new distinct `app.vitals.start.reason` key does not affect it (existing tests still pass).

## :pencil: Checklist

## :crystal_ball: Next steps
- Confirm with the product/spec side whether the reason should also be exposed as a queryable **tag** (in addition to span data), and lock in the exact string values dashboards/alerts will rely on.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
   - Emits a transaction named `App Start` with op `app.start`, carrying the existing app start measurements and phase spans (`process.load`, `contentprovider.load`, `application.load`, activity lifecycle spans) as direct children of the root
   - The standalone transaction shares the same `traceId` as the first `ui.load` activity transaction so they remain linked in the trace view
   - Also covers non-activity starts (broadcast receivers, services, content providers)
-  - On Android 15+ (API 35), the standalone transaction reports why the OS started the process via `app.vitals.start.reason` trace data (e.g. `launcher`, `broadcast`, `service`, `content_provider`), derived from `ApplicationStartInfo.getReason()` ([#5552](https://github.com/getsentry/sentry-java/pull/5552))
+  - On Android 15+ (API 35), the standalone `app.start` transaction reports why the OS started the process via `app.vitals.start.reason` trace data (e.g. `launcher`, `broadcast`, `service`, `content_provider`), derived from `ApplicationStartInfo.getReason()`. You can search and group by this attribute in the Trace Explorer. ([#5552](https://github.com/getsentry/sentry-java/pull/5552))
 
 ### Improvements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
   - Emits a transaction named `App Start` with op `app.start`, carrying the existing app start measurements and phase spans (`process.load`, `contentprovider.load`, `application.load`, activity lifecycle spans) as direct children of the root
   - The standalone transaction shares the same `traceId` as the first `ui.load` activity transaction so they remain linked in the trace view
   - Also covers non-activity starts (broadcast receivers, services, content providers)
+  - On Android 15+ (API 35), the standalone transaction reports why the OS started the process via `app.vitals.start.reason` trace data (e.g. `launcher`, `broadcast`, `service`, `content_provider`), derived from `ApplicationStartInfo.getReason()` ([#5552](https://github.com/getsentry/sentry-java/pull/5552))
 
 ### Improvements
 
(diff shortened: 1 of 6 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/ActivityLifecycleIntegration.java
```diff
@@ -140,6 +141,7 @@ public void register(final @NotNull IScopes scopes, final @NotNull SentryOptions
 
     if (performanceEnabled && this.options.isEnableStandaloneAppStartTracing()) {
       AppStartMetrics.getInstance().setHeadlessAppStartListener(this::onHeadlessAppStart);
+      addIntegrationToSdkVersion("StandaloneAppStart");
```
buenaflor, line 144: needed for tracking in looker

runningcode (approved):
Looks great! I feel a bit mixed about the added `ActivityLifecycleIntegrationTest.kt` since we're just setting the value and then getting it.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
   - Emits a transaction named `App Start` with op `app.start`, carrying the existing app start measurements and phase spans (`process.load`, `contentprovider.load`, `application.load`, activity lifecycle spans) as direct children of the root
   - The standalone transaction shares the same `traceId` as the first `ui.load` activity transaction so they remain linked in the trace view
   - Also covers non-activity starts (broadcast receivers, services, content providers)
+  - On Android 15+ (API 35), the standalone transaction reports why the OS started the process via `app.vitals.start.reason` trace data (e.g. `launcher`, `broadcast`, `service`, `content_provider`), derived from `ApplicationStartInfo.getReason()` ([#5552](https://github.com/getsentry/sentry-java/pull/5552))
```
romtsn: maybe worth mentioning that customers can search/group by this property in trace viewer? (they can, right?)

romtsn (approved):
dope!

buenaflor:
> I feel a bit mixed about the added ActivityLifecycleIntegrationTest.kt since we're just setting the value and then getting it.

Ah yeah, the mapping itself is covered in `AppStartMetricsTestApi35`, so these activity integration tests are really just checking the wiring that the reason ends up on the standalone `app.start` transaction under the right key, for both the foreground and headless paths

CHANGELOG.md
buenaflor: Good point, updated (and yes they can) 👍

PR #5565, opened 2026-06-18
chore(util): Add CompileOnlyCompat for compile-only linkage errors

0xadam-brown:
_**I'm not convinced this PR is worthwhile: see the "Callouts" section below.**_

## :scroll: Description

Introduces an internal helper for calling compileOnly APIs that may be missing or binary-incompatible at runtime.

Adopted in SentrySQLiteDriver and SentryLayoutNodeHelper to replace ad-hoc try/catch blocks.

We frequently use `compileOnly` dependencies to instrument code so that users can remain free to chose their preferred version. That means the shape of the dependency we compile against may not match what shows up at runtime. In particular, if the dependency has an API that adds methods over time, users can crash if we don't use `try-catch` blocks when calling those methods. The utilities introduced here help us raise awareness about the issue and centralize our approach.

## ⚠️ Callouts

Not in love with this PR (my vote would be to close it and wait for more examples to show up, if ever), but I'm posting b/c I said I would in our sync + I wanted folks to have a chance to weigh in. IMO, it's heavyweight for what ended up being just two call sites. And there's probably no way to keep everyone happy with the approach. 

A number of reasons for that: 

**[1]** If we want these utilities to be generally available, they need to live in the core `sentry` module. That module is Java-only, and that means making Kotlin call sites look reasonably idiomatic can be tricky.

**[2]** We can't easily make them Kotlin-only as there's no "core"/"util" Kotlin module that all other Kotlin modules depend on. (The closest is `sentry-kotlin-extension`, but those are externally-facing and higher-level than the abstractions here.)

**[3]** Extracting `try-catch` logic into a method requires using lambdas / SAMs, and that means an additional object allocation relative to an inline `try-catch`. (We can't use an inline function b/c of [1]/[2].)

**[4]** Sometimes fallbacks are themselves lambdas rather than a simple default value. That means two lambda args, which further complicates call sites. (Eg, my first approach was `T withFallbackIfAbsent(T fallback, Callable<T> callable)`, but the additional Callable for the fallback case made that less-than-lovely. Plus, putting the fallback first was misleading in the case of booleans, as it looks like you're opting out of some under-the-hood functionality: `withFallbackIfAbsent(false) { delegate.doSomething() }`.)

Sigh.

New unit tests + existing tests still pass.

## :pencil: Checklist

### DIFF
--- a/sentry/src/main/java/io/sentry/util/CompileOnlyCompat.java
+++ b/sentry/src/main/java/io/sentry/util/CompileOnlyCompat.java
@@ -1,0 +1,72 @@
+package io.sentry.util;
+
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Helpers for safely calling methods from {@code compileOnly} dependencies that may not exist at
+ * runtime.
+ *
+ * <p>When a dependency is {@code compileOnly}, the app supplies the actual version. That version
+ * may predate the API we compiled against, which means it may not have methods we call from our
+ * source code if the API has added methods over time. Calling a missing method throws a {@link
+ * LinkageError} (e.g., {@link NoSuchMethodError}, {@link AbstractMethodError}, etc.). This class
+ * lets us centralize the try/catch-with-fallback pattern.
+ *
+ * <p>This is not for {@code implementation} dependencies with transitive version conflicts on the
+ * classpath; use local {@code try/catch} handling for those cases instead.
+ */
+@ApiStatus.Internal
+public final class CompileOnlyCompat {
+
+  /**
+   * Functional interface for a call that returns a non-null value and may throw a {@link
+   * LinkageError}.
+   *
+   * <p>Use it when calling from Kotlin:
+   *
+   * <pre>{@code
+   * CompileOnlyCall { delegate.hasConnectionPool }.ifAbsent(false)
+   * }</pre>
+   */
+  @FunctionalInterface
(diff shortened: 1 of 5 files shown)


sentry/src/main/java/io/sentry/util/CompileOnlyCompat.java
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  private CompileOnlyCompat() {}
+
+  private static <T> @NotNull T run(
+      final @NotNull CompileOnlyCall<T> call, final @NotNull Fallback<T> fallback) {
+    try {
+      return call.call();
+    } catch (LinkageError e) {
+      return fallback.call(e);
```
0xadam-brown, line 69: Note that we're not logging. Three reasons for that:

1. Falling back when the method isn't present is correct behavior and expected from the consumer's side in cases where there's a default implementation for the method and our fallback matches its behavior (this will almost always be the case). 
2. Wiring a logger through this Java API would make it even more ungainly.
3. Callers can always log themselves via the `Fallback` lambda.

romtsn (commented):
@0xadam-brown thanks for opening this up -- while it looks indeed nicer on the kotlin side, I think the downsides that you listed outweigh the benefit of this utility. So I agree with you about closing this until further examples show up that would justify having this utility 👍

0xadam-brown:
Thanks for taking a look @romtsn – and makes sense to me as well. Closing and we can revisit if we see more examples show up.

PR #5587, opened 2026-06-22
perf(android): [SDK Overhead Reduction 4] Use default timezone directly

adinauer:

- #5499
- #5500
- #5501
- #5502
- #5587
- #5589
- #5591
- #5598
- #5599
- #5600
- #5601
- #5602
- #5603
- #5609

---

## :scroll: Description

Replace the normal Android device timezone path with `TimeZone.getDefault()` instead of constructing a `Calendar` just to read its timezone.

One compatibility detail is preserved: Android 13+ `Calendar.getInstance(locale)` can honor a locale Unicode `tz` extension (for example `en-US-u-tz-usnyc`). For those rare locales, the code keeps the Calendar path so the SDK continues reporting the same timezone as before. On Android N through Android 12, and for normal locales without a `tz` extension, the locale did not affect the timezone value.

Part of the SDK Overhead Reduction stack. In the common case, device context collection only needs the process default timezone. `TimeZone.getDefault()` is available on Android API 21 and Java 8, and avoids constructing a `Calendar` and reading locale configuration just to retrieve the same timezone.

The previous API 24+ branch used the first `Configuration` locale, but Android N through Android 12 did not use that locale to select the timezone. Android 13 added `tz` Unicode extension support to `Calendar.getInstance(locale)`, so this PR preserves that behavior with a narrow fallback while keeping the fast direct-default path for all other cases.

- `./gradlew :sentry-android-core:testDebugUnitTest --tests io.sentry.android.core.DeviceInfoUtilTest` — all tests pass
  - includes regression coverage for an Android 13+ locale with a Unicode `tz` extension (`en-US-u-tz-usnyc`)

## :pencil: Checklist

## :crystal_ball: Next steps


lbloder (approved):
LGTM 👍

adinauer:
Cursor review

adinauer:
@sentry review

sentry-android-core/src/main/java/io/sentry/android/core/DeviceInfoUtil.java
```diff
@@ -257,14 +257,16 @@ private void setDeviceIO(
   @SuppressWarnings("NewApi")
```
runningcode, line 257: If we have a proper check for the `Build.VERSION_CODE` why do we need to suppress this?

sentry-android-core/src/main/java/io/sentry/android/core/DeviceInfoUtil.java
```diff
@@ -257,14 +257,16 @@ private void setDeviceIO(
...
   @NotNull
   private TimeZone getTimeZone() {
```
runningcode, line 259: This method is probably worth a comment about why this built like this and any gotchas.

runningcode (commented):
This one is a bit confusing and I'm not sure about the edge cases. Do we have an idea what the performance benefit is here?

adinauer:
Benchmark results invoking the util on a newer Android version (API level 34) that does not enter the inner if regarding "tz":

```
 ┌─────────────────────────────────────────┬─────────────┬─────────────┬────────┐
 │ Metric                                  │ Old/base    │ New/head    │ Delta  │
 ├─────────────────────────────────────────┼─────────────┼─────────────┼────────┤
 │ Benchmark thread wall time              │ 6803.058 ms │ 1624.340 ms │ -76.1% │
 ├─────────────────────────────────────────┼─────────────┼─────────────┼────────┤
 │ Benchmark thread CPU time               │ 6421.307 ms │ 1532.483 ms │ -76.1% │
 ├─────────────────────────────────────────┼─────────────┼─────────────┼────────┤
 │ CPU per call                            │ 0.3211 ms   │ 0.0766 ms   │ -76.1% │
 ├─────────────────────────────────────────┼─────────────┼─────────────┼────────┤
 │ App process CPU during benchmark window │ 9187.463 ms │ 2716.467 ms │ -70.4% │
 ├─────────────────────────────────────────┼─────────────┼─────────────┼────────┤
 │ Scheduler slices                        │ 1037        │ 223         │ -78.5% │
 └─────────────────────────────────────────┴─────────────┴─────────────┴────────┘
 ```

Results for inner code triggered, e.g. by "en-US-u-tz-usnyc":

```
 ┌─────────────────────────────────────────┬───────────────┬───────────────┬───────┐
 │ Metric                                  │ Old/base rare │ New/head rare │ Delta │
 ├─────────────────────────────────────────┼───────────────┼───────────────┼───────┤
 │ Benchmark thread wall time              │ 14359.059 ms  │ 15315.136 ms  │ +6.7% │
 ├─────────────────────────────────────────┼───────────────┼───────────────┼───────┤
 │ Benchmark thread CPU time               │ 13549.090 ms  │ 14585.745 ms  │ +7.7% │
 ├─────────────────────────────────────────┼───────────────┼───────────────┼───────┤
 │ CPU per call                            │ 0.6775 ms     │ 0.7293 ms     │ +7.7% │
 ├─────────────────────────────────────────┼───────────────┼───────────────┼───────┤
 │ App process CPU during benchmark window │ 19780.192 ms  │ 20366.205 ms  │ +3.0% │
 └─────────────────────────────────────────┴───────────────┴───────────────┴───────┘
```

Both benchmarks ran 20k iterations on the util.

**Comment by adinauer** [comment]
Some more info on the "tz":

```
tz is a Unicode locale extension key meaning timezone.

 Example:

 `en-US-u-tz-usnyc`

 Breakdown:

 - en-US = English, United States
 - u = start of Unicode locale extensions
 - tz = timezone extension key
 - usnyc = timezone type, here New York

 So en-US-u-tz-usnyc roughly means: “English (US), with preferred timezone New York.”

 It’s different from the device/system timezone. It’s metadata embedded in the locale tag, mainly for APIs that format calendars/dates according to locale preferences. Android 13+ Calendar.getInstance(locale) can honor this extension, which is why PR #5587 keeps the Calendar path when
 locale.getUnicodeLocaleType("tz") != null.
```

sentry-android-core/src/main/java/io/sentry/android/core/DeviceInfoUtil.java
adinauer: Because we're checking against `BuildInfoProvider` instead of directly checking `Build.VERSION.SDK_INT`

PR #5589, opened 2026-06-22
perf(core): [SDK Overhead Reduction 5] Replace Calendar with Date in DateUtils

adinauer:

- #5499
- #5500
- #5501
- #5502
- #5587
- #5589
- #5591
- #5598
- #5599
- #5600
- #5601
- #5602
- #5603
- #5609

---

## :scroll: Description

Replace `Calendar.getInstance(TIMEZONE_UTC).getTime()` with `new Date()` in `DateUtils.getCurrentDateTime()`, and replace the `Calendar` conversion in `DateUtils.getDateTime(long)` with `new Date(millis)`.

`Date` stores epoch milliseconds and has no timezone state, so the returned values are unchanged. The UTC `Calendar` was only doing unnecessary allocation and calendar field computation before returning the same epoch millis as a `Date`.

Also update a stale Breadcrumb memoization comment that referenced Calendar allocation.

Part of the SDK Overhead Reduction stack. `DateUtils` is used in hot event, breadcrumb, and session timestamp paths. Constructing `Calendar` for these helpers is unnecessary because no calendar fields are used. The change keeps Android/JDK compatibility and avoids introducing `java.time` or desugaring requirements.

- `./gradlew :sentry:test --tests io.sentry.DateUtilsTest` — all tests pass

## :pencil: Checklist

## :crystal_ball: Next steps


lbloder (approved):
LGTM 👍

adinauer:
Cursor review

adinauer:
@sentry review

sentry/src/main/java/io/sentry/DateUtils.java
```diff
@@ -24,10 +21,9 @@ private DateUtils() {}
    *
    * @return the UTC Date
    */
-  @SuppressWarnings("JdkObsolete")
+  @SuppressWarnings("JavaUtilDate")
   public static @NotNull Date getCurrentDateTime() {
-    final Calendar calendar = Calendar.getInstance(TIMEZONE_UTC);
-    return calendar.getTime();
+    return new Date();
```
runningcode, line 26: Why was the previous code doing it this way?
Doesn't a Date() object create a Calendar object under the hood anyways.
What do the benchmarks show in terms of speed for this change?

adinauer: > Doesn't a Date() object create a Calendar object under the hood anyways.

No, just stores a `long` for current time millis.

> Why was the previous code doing it this way?

Just pasting agent output here:

Historically, it looks like it was done to force “UTC Date” semantics while optimizing an earlier, much slower implementation.

 From PR #1111 (Ref: using Calendar to generate Dates), the motivation was:

 │ using Calendar instead of Dates directly
 │ this would avoid the Date/String parsing and improve the performance #1081

 Before that refactor, getCurrentDateTime() effectively did:

 ```java
   getTimestampIsoFormat(new Date()) // format as UTC string
   getDateTime(timestampIsoFormat)    // parse back into Date
 ```

 That was expensive, especially on low-end Android devices. The Calendar.getInstance(UTC).getTime() version was introduced as a faster way to avoid the Date -> String -> Date roundtrip while preserving the intended “UTC” behavior.

 But the underlying assumption was slightly off: Date itself has no timezone. It’s just epoch millis. Timezone only matters when formatting/parsing or when using calendar fields. So for “current instant”, this:

 ```java
   Calendar.getInstance(TIMEZONE_UTC).getTime()
 ```

 is equivalent in value to:

 ```java
   new Date()
 ```

 but slower.

adinauer: benchmark results:

 1 thread

 ```text
   mainGetCurrentDateTime   55.956 ns/op   528.001 B/op
   prGetCurrentDateTime     17.097 ns/op    24.000 B/op

   mainGetDateTimeMillis    98.190 ns/op   528.001 B/op
   prGetDateTimeMillis       2.360 ns/op    24.000 B/op
 ```

 4 threads

 ```text
   mainGetCurrentDateTime    70.010 ns/op   528.000 B/op
   prGetCurrentDateTime     101.737 ns/op    24.000 B/op

   mainGetDateTimeMillis    106.893 ns/op   528.000 B/op
   prGetDateTimeMillis        2.480 ns/op    24.000 B/op
 ```

Slowdown is likely due the benchmark hammering `System.currentTimeMillis()` from multiple threads because the version passing a `long` to ctor is much faster.

Anyways it seems very unrealistic for any code to be this excessive on `System.currentTimeMillis` calls for it to matter.

runningcode: Looks great! thanks fro the explanation and benchmarks!

runningcode (approved):
Much simpler and faster! Great

PR #4707, opened 2025-09-09
Use Kotlin stdlib 1.9.24 dependency for Android modules

markushi:
## :scroll: Description
Unsure about 
- `ktor` (switched to 1.9.24)
- UI test dependencies (switched to 1.9.24)
- samples (stuck with compiler version)

## :green_heart: How did you test it?
Manually, using the [socialite app ](https://github.com/android/socialite/) @ [Kotlin 1.9.x](https://github.com/android/socialite/commits/5263dffbe8ba1426c6062fde13b2188e1e5d0e04)
 and Kotlin 2.x (main branch).

## :pencil: Checklist

- [ ] I added tests to verify the changes.
- [ ] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [ ] Review from the native team if needed.
- [ ] No breaking change or entry added to the changelog.
- [ ] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

PR #5590, opened 2026-06-22
ref(android): Remove kotlinStdLibVersionAndroid pin

runningcode:
## :scroll: Description

Removes the `Config.kotlinStdLibVersionAndroid` constant (`1.9.24`) and the explicit version argument from the `kotlin(Config.kotlinStdLib, ...)` dependency declarations in the Android modules. The stdlib version now aligns with the Kotlin Gradle plugin (2.2.0) automatically.

Affected modules: `sentry-okhttp`, `sentry-ktor-client`, `sentry-android-sqlite`, `sentry-android-timber`, `sentry-android-distribution`, `sentry-android-replay`, `sentry-launchdarkly-android`.

The pin overrode the stdlib version managed by the Kotlin Gradle plugin, which is discouraged. Letting the plugin manage the stdlib version keeps it consistent with the compiler and removes a stale constant.

`./gradlew spotlessApply apiDump` builds successfully; no API changes.

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Release `MediaMuxer` when a replay segment has no encodable frames to avoid a resource leak ([#5583](https://github.com/getsentry/sentry-java/pull/5583))
 
+### Dependencies
+
+- Align the Kotlin stdlib version in Android modules with the Kotlin Gradle plugin instead of pinning it to 1.9.24 ([#5590](https://github.com/getsentry/sentry-java/pull/5590))
+
 ## 8.44.1
 
 ### Fixes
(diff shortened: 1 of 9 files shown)


romtsn (commented):
we actually pinned it on purpose, see https://github.com/getsentry/sentry-java/pull/4707

reason was that RN or dotnet had had a compile time issue with Kotlin 2.+. I don't know if it's still the case, but I'm a bit hesitant about this change

0xadam-brown (commented):
Thanks for the hygiene update, @runningcode!

h: Any worries that the original concerns that led to us pinning still apply? (see the CHANGELOG from #4707) 

Eg, with the Kotlin Gradle Plugin currently at 2.x+, our compiled classes could have `@Metadata` annotations a 1.x+ compiler wouldn't recognize and therefore would fail on.

EDIT: and I just saw Roman left the same comment – sorry for the duplicate...

PR #5592, opened 2026-06-22
fix(replay): Fix network detail response body size being unknown for gzip-compressed responses

romtsn:
## :scroll: Description

For gzip-compressed responses, OkHttp strips the `Content-Length` header during transparent decompression, so `response.body.contentLength()` returns `-1`. This caused `NetworkRequestData.responseBodySize` to always be unknown for replay network details on such responses.

This PR adds `originalByteCount` to `NetworkBody`, populated from the actual byte array length in `NetworkBodyParser.fromBytes`. `NetworkDetailCaptureUtils.createRequestOrResponseInternal` now uses this as a fallback when the passed `bodySize` is `null` or `-1`.

The fix piggybacks on the existing `peekBody` call — no additional network I/O. For responses larger than the peek limit (`MAX_NETWORK_BODY_SIZE`), the capped byte count is reported (lower bound) instead of unknown.

Example replay: https://sentry-sdks.sentry.io/explore/replays/95bf32448ef94fafb41b5164ccde5d19

Screenshot 2026-06-22 17 04 56

Session replay network details show response body size. For gzip-compressed HTTP responses (common for JSON APIs), OkHttp's transparent decompression strips `Content-Length`, making the size always unknown (`-1`). The Sentry UI shows "It is possible the network transfer size is smaller due to compression", so reporting the decompressed size is the correct behavior.

Fixes SDK-1006

- Added unit tests for `NetworkBodyParser` covering `originalByteCount` for both truncated and non-truncated bodies
- Added unit tests for `NetworkDetailCaptureUtils` covering the bodySize fallback logic (unknown → uses byte count, explicit → preserved, body capture off → unchanged)
- Verified existing tests pass

## :pencil: Checklist

## :crystal_ball: Next steps

### DIFF
--- a/sentry/src/main/java/io/sentry/util/network/NetworkBody.java
+++ b/sentry/src/main/java/io/sentry/util/network/NetworkBody.java
@@ -16,15 +16,24 @@ public final class NetworkBody {
 
   private final @Nullable Object body;
   private final @Nullable List<NetworkBodyWarning> warnings;
+  private final long originalByteCount;
 
   public NetworkBody(final @Nullable Object body) {
-    this(body, null);
+    this(body, null, -1);
   }
 
   public NetworkBody(
       final @Nullable Object body, final @Nullable List<NetworkBodyWarning> warnings) {
+    this(body, warnings, -1);
+  }
+
+  public NetworkBody(
+      final @Nullable Object body,
+      final @Nullable List<NetworkBodyWarning> warnings,
+      final long originalByteCount) {
     this.body = body;
     this.warnings = warnings;
+    this.originalByteCount = originalByteCount;
   }
 
   public @Nullable Object getBody() {
@@ -33,6 +42,10 @@ public final class NetworkBody {
 
   public @Nullable List<NetworkBodyWarning> getWarnings() {
     return warnings;
+  }
+
+  public long getOriginalByteCount() {
+    return originalByteCount;
   }
 
   // Based on
--- a/sentry/src/main/java/io/sentry/util/network/NetworkBody.java
+++ b/sentry/src/main/java/io/sentry/util/network/NetworkBody.java
@@ -27,7 +27,7 @@ public final class NetworkBody {
     this(body, warnings, -1);
   }
 
-  public NetworkBody(
+  NetworkBody(
       final @Nullable Object body,
       final @Nullable List<NetworkBodyWarning> warnings,
       final long originalByteCount) {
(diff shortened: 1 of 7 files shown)


sentry/src/main/java/io/sentry/util/network/NetworkBody.java
```diff
@@ -16,15 +16,24 @@ public final class NetworkBody {
...
 
   public NetworkBody(
       final @Nullable Object body, final @Nullable List<NetworkBodyWarning> warnings) {
+    this(body, warnings, -1);
+  }
+
+  public NetworkBody(
+      final @Nullable Object body,
+      final @Nullable List<NetworkBodyWarning> warnings,
+      final long originalByteCount) {
```
0xadam-brown, line 33: m: Thoughts about making this package private?

I worry about folks not knowing what to do with the constructor's `originalByteCount` param. (They have to dig into our implementation to find out.) 

Alternatively, we could make the constructor private and expose a static factory method to be (arguably) even more explicit about what's going on:

```java
final NetworkBody parsed = parse(content, contentType, isTruncated, logger);
return parsed.withSourceByteLength(bytes.length);
```

```java
private NetworkBody(Object body, List<NetworkBodyWarning> warnings, long sourceByteArrayLength)

// package-private
NetworkBody withSourceByteLength(long sourceByteArrayLength) {
  return new NetworkBody(body, warnings, sourceByteArrayLength);
}
```

...or whatever.

0xadam-brown (commented):
Thanks for the fix! 

One comment that could avoid a new public API surface (doesn't have to block us); otherwise lgtm.

sentry/src/main/java/io/sentry/util/network/NetworkBody.java
romtsn: good catch, I made it package-private 👍

sentry/src/main/java/io/sentry/util/network/NetworkBody.java
```diff
@@ -35,6 +44,10 @@ public NetworkBody(
     return warnings;
   }
 
+  public long getOriginalByteCount() {
```
0xadam-brown: m: Thoughts about making this guy package private too? (Sorry for not mentioning that explicitly the first time!)

romtsn: indeed 🤦

0xadam-brown (approved):
🙌

PR #5595, opened 2026-06-22
perf(android): Hit-test gestures without getLocationOnScreen

runningcode:
## :scroll: Description
`ViewUtils.findTarget` (run on every tap and at every scroll start) called `View.getLocationOnScreen` for every visited view to hit-test it. `getLocationOnScreen` walks from the view **up to the root** each call, so hit-testing the whole tree was **O(N·depth)** per gesture, repeatedly re-walking the same ancestor chains.

This replaces that with the technique `ViewGroup` itself uses to dispatch touch events: map the touch **point down** into each child's local coordinate space as we descend (offset by the parent's scroll and the child's `left/top`, then apply the child's inverse matrix for transformed views). Each view is then a cheap `O(1)` bounds check against its own `[0,0,width,height]`, making the whole traversal **O(N)**.

Notes:
- The locators still receive the original decor-view-relative `x,y`, because `ComposeGestureTargetLocator` hit-tests against window coordinates.
- Because the traversal now stays in window/local space instead of mixing in screen space, hit-testing is also slightly more correct when the window isn't at the screen origin (status bar offset, split-screen).
- Keeps the BFS order, so the selected target for overlapping clickable views is unchanged.

The `sentry-compose` locator gets the same `LinkedList`→`ArrayDeque` cleanup (it already hit-tests in window space, so no coordinate change there).

JAVA-534 / #5481. Follow-up to #5594 — that PR removed the queue-node allocations; this one removes the per-view `getLocationOnScreen` call, replacing an `O(N·depth)` traversal with `O(N)`.

Tradeoff: carrying the per-view local coordinates through the BFS needs a small holder object per node, so this re-introduces a per-node allocation that #5594 removed. That is an intentional trade — avoiding the repeated `O(depth)` ancestor walk per view in exchange for a short-lived gen-0 object.

A micro-benchmark (see testing notes) measures a **~4–5×** speedup of the traversal on large view trees, scaling down to ~1.7× on trivial ones — the win is proportional to how many views sit under the touch.

- Rewrote the gesture test harness (`ViewHelpers`) to mock local geometry; all existing `SentryGestureListenerClickTest` / `SentryGestureListenerScrollTest` cases pass unchanged.
- Added `ViewUtilsTest` coverage that exercises the actual transform — a child offset within its parent is hit when the point maps inside it and missed when it maps outside, even though the point is inside the decor view.
- On-device (Pixel 3, Android 12): scroll target still resolves (`Scroll target found: scrolling_container`) and a click on a button near the bottom of the layout resolves correctly (`view.id: scrolling_crash`), confirming the transform is correct for a real non-trivial offset.

**Timing — micro-benchmark.** Profiler-based measurement didn't work: on the test device (Pixel 3, Android 12) instrumented method tracing produces empty traces, fine-grained sampling overflows the buffer, and at the only working sampling rate (1 ms) `findTarget` (~0.3 ms) is below the resolution — run-to-run variance dwarfs the difference. So I wrote a JVM micro-benchmark instead (not committed — it was a one-off).

What it does: builds one real, attached, laid-out `View` tree and times both hit-testing strategies over it — **old** (`getLocationOnScreen` per view) vs **new** (point-transform-during-descent) — 800 iterations × 6 rounds each, alternating order with warmup, on a release/R8 build. Only the per-view bounds computation differs between the two paths (the part this PR changes); the locator loop is excluded since it's identical for both. The tree is built so every node contains the touch point, forcing a full N-node traversal, and is swept across four sizes. Median of two runs:

| Tree (views under touch) | old: `getLocationOnScreen` | new: point-transform | speedup |
|---|---|---|---|
| ~9 (a simple screen) | ~9–12 µs | ~5–7 µs | ~1.7× |
| ~61 | ~45–54 µs | ~12–15 µs | ~3.7× |
| ~161 | ~143–175 µs | ~35–44 µs | ~4.0× |
| ~301 | ~298–376 µs | ~61–84 µs | ~4.5–4.9× |

So `getLocationOnScreen` costs ~1 µs/view, the new transform ~0.2 µs/view → a steady ~4–5× on the traversal, with absolute savings scaling from a few µs on a trivial screen to ~250 µs on a dense 300-view one. This is a per-tap / per-scroll-start main-thread cost (well under one frame even at the high end), not a frame-rate or startup change. Speedup ratios reproduced within ~8% across runs; absolute numbers drift with device thermals.

## :pencil: Checklist

## :crystal_ball: Next steps
None.

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/ViewUtils.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/ViewUtils.java
@@ -1,13 +1,14 @@
 package io.sentry.android.core.internal.gestures;
 
 import android.content.res.Resources;
+import android.graphics.Matrix;
 import android.view.MotionEvent;
 import android.view.View;
 import android.view.ViewGroup;
 import io.sentry.android.core.SentryAndroidOptions;
 import io.sentry.internal.gestures.GestureTargetLocator;
 import io.sentry.internal.gestures.UiElement;
-import java.util.LinkedList;
+import java.util.ArrayDeque;
 import java.util.List;
 import java.util.Queue;
 import org.jetbrains.annotations.ApiStatus;
@@ -17,30 +18,53 @@ import org.jetbrains.annotations.Nullable;
 @ApiStatus.Internal
 public final class ViewUtils {
 
-  private static final int[] coordinates = new int[2];
-
   /**
-   * Verifies if the given touch coordinates are within the bounds of the given view.
+   * Verifies if the given touch coordinates, expressed in the view's own local coordinate space,
+   * are within the bounds of the given view.
    *
    * @param view the view to check if the touch coordinates are within its bounds
-   * @param x - the x coordinate of a {@link MotionEvent}
-   * @param y - the y coordinate of {@link MotionEvent}
+   * @param localX - the x coordinate of the touch, relative to the view's top-left corner
+   * @param localY - the y coordinate of the touch, relative to the view's top-left corner
    * @return true if the touch coordinates are within the bounds of the view, false otherwise
    */
   private static boolean touchWithinBounds(
-      final @Nullable View view, final float x, final float y) {
+      final @Nullable View view, final float localX, final float localY) {
     if (view == null) {
       return false;
     }
 
-    view.getLocationOnScreen(coordinates);
-    int vx = coordinates[0];
-    int vy = coordinates[1];
+    final int w = view.getWidth();
+    final int h = view.getHeight();
 
-    int w = view.getWidth();
-    int h = view.getHeight();
+    return !(localX < 0 || localX > w || localY < 0 || localY > h);
+  }
 
-    return !(x < vx || x > vx + w || y < vy || y > vy + h);
+  /**
+   * Maps a touch point expressed in the parent's local coordinate space into the child's local
+   * coordinate space. This mirrors how {@link ViewGroup} dispatches touch events to its children
+   * and lets us hit-test the whole tree with a single downward traversal, instead of calling {@link
+   * View#getLocationOnScreen(int[])} (which walks up to the root) for every view.
+   */
+  private static @NotNull ViewWithLocation mapToChild(
+      final @NotNull View child,
+      final float parentX,
+      final float parentY,
+      final int parentScrollX,
+      final int parentScrollY) {
+    float childX = parentX + parentScrollX - child.getLeft();
+    float childY = parentY + parentScrollY - child.getTop();
+
+    final @Nullable Matrix matrix = child.getMatrix();
+    if (matrix != null && !matrix.isIdentity()) {
+      final Matrix inverse = new Matrix();
+      if (matrix.invert(inverse)) {
+        final float[] point = {childX, childY};
+        inverse.mapPoints(point);
+        childX = point[0];
+        childY = point[1];
+      }
+    }
+    return new ViewWithLocation(child, childX, childY);
   }
 
   /**
@@ -48,8 +72,8 @@ public final class ViewUtils {
    * given {@code viewTargetSelector}.
    *
    * @param decorView - the root view of this window
-   * @param x - the x coordinate of a {@link MotionEvent}
-   * @param y - the y coordinate of {@link MotionEvent}
+   * @param x - the x coordinate of a {@link MotionEvent}, relative to the decor view
+   * @param y - the y coordinate of {@link MotionEvent}, relative to the decor view
    * @param targetType - the type of target to find
    * @return the {@link View} that contains the touch coordinates and complements the {@code
    *     viewTargetSelector}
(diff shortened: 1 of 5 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/ViewUtils.java
```diff
@@ -17,39 +18,62 @@
...
-    int h = view.getHeight();
+    return !(localX < 0 || localX > w || localY < 0 || localY > h);
+  }
 
-    return !(x < vx || x > vx + w || y < vy || y > vy + h);
+  /**
+   * Maps a touch point expressed in the parent's local coordinate space into the child's local
+   * coordinate space. This mirrors how {@link ViewGroup} dispatches touch events to its children
+   * and lets us hit-test the whole tree with a single downward traversal, instead of calling {@link
+   * View#getLocationOnScreen(int[])} (which walks up to the root) for every view.
```
0xadam-brown, line 46: 🥇

sentry-android-core/src/test/java/io/sentry/android/core/internal/gestures/ViewUtilsTest.kt
```diff
@@ -80,6 +86,44 @@ class ViewUtilsTest {
...
+        gestureTargetLocators = listOf(AndroidViewGestureTargetLocator(LazyEvaluator { true }))
+      }
+
+    // (120, 220) maps to (20, 20) in the child's space -> inside its 50x50 bounds.
+    assertNotNull(ViewUtils.findTarget(options, decorView, 120f, 220f, UiElement.Type.CLICKABLE))
+
+    // (90, 220) maps to (-10, 20) in the child's space -> outside, despite being inside the decor.
+    assertNull(ViewUtils.findTarget(options, decorView, 90f, 220f, UiElement.Type.CLICKABLE))
+  }
+
```
0xadam-brown, line 111: l: Nice! Beyond covering the top/left path, might be worth adding quick cases for parent scrollX/scrollY and a non-identity child matrix. The gesture listener mocks won't catch regressions there and they'd cover the other two mapToChild branches.

0xadam-brown (approved):
Great work! (And love the [helper doc](https://claude.ai/code/artifact/c824d0db-58ff-4897-9fef-1583a252759a).) 

One optional comment; otherwise lgtm

sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/ViewUtils.java
```diff
@@ -17,39 +18,62 @@
...
+    if (matrix != null && !matrix.isIdentity()) {
+      final Matrix inverse = new Matrix();
+      if (matrix.invert(inverse)) {
+        final float[] point = {childX, childY};
+        inverse.mapPoints(point);
+        childX = point[0];
+        childY = point[1];
+      }
+    }
+    return new ViewWithLocation(child, childX, childY);
```
romtsn, line 67: i guess the next step to improve this could be to have a pool of reusable `ViewWithLocation`s to reduce the number of allocations? but that could be done separately (if it makes sense at all)

romtsn (approved):
dope! I wish my Algebraic geometry teacher had told me what would I need matrices for 😅

sentry-android-core/src/test/java/io/sentry/android/core/internal/gestures/ViewUtilsTest.kt
runningcode: added that test below! good point

sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/ViewUtils.java
runningcode: i always dreamed of the day I would need to use a pool to fix a real performance problem. it's a good point. can add a ticket for a follow up investigation to see if this makes sense.

runningcode: Claude says:

❯ Would it make sense to add a ticket for a follow up to add a pool for the ViewWithLocation objects?

⏺ Honestly, I'd lean against it — and I wouldn't file the ticket speculatively.

  Reasoning:

  - These are textbook ephemeral allocations. Each ViewWithLocation is ~24 bytes (one ref + two floats) and dies immediately after the
  traversal. ART's young-gen allocates these with a bump pointer (TLAB) and collects them almost for free. Pooling trades that for a managed
  pool with reset/lifecycle logic, thread-safety questions, and live objects the GC now has to keep scanning — on Android, hand-rolled pools for
  small short-lived objects often net-lose.
  - The count is bounded and small — one per view on the hit-test path, and the traversal prunes whole subtrees as soon as a view doesn't
  contain the touch, so it's nowhere near the full tree. This fires at interaction frequency (taps/scrolls), not per-frame render.
  - The real win in this PR was removing the per-view getLocationOnScreen walk (O(depth) up to the root, each call). The wrapper allocation is
  noise next to that.

PR #5605, opened 2026-06-23
feat(extend-app-start): [1/4] Add IAppStartExtender bridge

buenaflor:
## PR Stack (Extend App Start)

- #5604
- #5605
- #5606
- #5608
- #5628

---

## :scroll: Description

Adds the naming-stable **core contract** for the app start extension API:

- `IAppStartExtender` (`@ApiStatus.Internal`) + a `NoOpAppStartExtender` default.
- Wired into `SentryOptions` via a getter/setter, mirroring `fullyDisplayedReporter`.

Inert until later PRs — returns `NoOpSpan` / no-ops until the Android implementation (`[2/4]`) and the public `Sentry` facade (`[4/4]`) land. Splitting the contract from the machinery keeps the API naming isolated and easy to review.

## :bulb: Motivation and Context

First PR of the stack implementing the app start extension API (part of getsentry/sentry-java#5553), which lets apps extend the app start measurement to cover extra launch-time work (mirrors sentry-cocoa's `extendAppLaunch()`). Splitting into stacked PRs keeps each change small and reviewable.

## :green_heart: How did you test it?

Unit tests (TDD):
- `SentryOptionsTest` — default `appStartExtender` is `NoOpAppStartExtender`; setter works.
- `NoOpAppStartExtenderTest` — singleton; `getExtendedAppStartSpan()` returns `NoOpSpan`; `extendAppStart()` / `finishAppStart()` no-op.

`./gradlew :sentry:apiCheck :sentry:spotlessJavaCheck :sentry:spotlessKotlinCheck` pass.

## :pencil: Checklist

- [x] I added GH Issue ID _&_ Linear ID
- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [ ] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

## :crystal_ball: Next steps

`[2/4]` — extract the `AppStartExtension` component in `sentry-android-core`.

> ⚠️ **Merge this PR using a merge commit** (not squash). Only the collection branch is squash-merged into main.

#skip-changelog

PR #5604, opened 2026-06-23
feat(extend-app-start): App start extension API

buenaflor:
## PR Stack (Extend App Start)

- #5604
- #5605
- #5606
- #5608
- #5628

---

Tested scenarios:

1 · Non-headless — extended after first frame (finish called ~5s in) → OK, app_start_cold = 6511 ms
https://sentry-sdks.sentry.io/explore/traces/trace/cb4ce5b65c42445da7d87d80fd49896a/

2 · Non-headless — finish never called → DEADLINE_EXCEEDED, measurement suppressed
https://sentry-sdks.sentry.io/explore/traces/trace/87bb5f4916f04837b5b4b6a31b021d0a/

3 · Headless — extended (finish called) → OK, app_start_cold = 6342 ms
https://sentry-sdks.sentry.io/explore/traces/trace/0563146982d44e05ac67000cc0089f6e/

4 · Headless — finish never called → DEADLINE_EXCEEDED, measurement suppressed
https://sentry-sdks.sentry.io/explore/traces/trace/6aa847bb646c4810b2991ad6b1c2687d/

---

Collection PR for the **Extend App Start** stack. Squash-merge this into `main` once all stack PRs are merged.

Implements the app start extension API — `Sentry.extendAppStart()` / `Sentry.finishAppStart()` / `Sentry.getExtendedAppStartSpan()` — which lets apps extend the app start measurement to cover extra launch-time work.

Closes getsentry/sentry-java#5553 https://github.com/getsentry/sentry-java/issues/3581

Issue #5553, opened 2026-06-16
Add API to extend app start (Android)


buenaflor:
Add an API to the Sentry Android SDK to extend the app start measurement past its default end point and instrument the extended region with child spans.

```kotlin
// Simple: extend the app start window, then finish
Sentry.extendAppStart()
initializeRemoteConfig()
Sentry.finishExtendedAppStart()
```

```kotlin
// With child spans: instrument the extended region
Sentry.extendAppStart()
val span = Sentry.getExtendedAppStartSpan()

val configSpan = span.startChild("app.init", "fetch remote config")
fetchRemoteConfig()
configSpan.finish()

Sentry.finishExtendedAppStart()
```

**Reference:** getsentry/sentry-cocoa#6886 (Done, `SentryExtendedAppLaunchManager`). **Siblings:** getsentry/sentry-dart#3767, getsentry/sentry-react-native#6303 — shared contract, per-platform deltas below.

## API

* `extendAppStart(): void` — creates the extended app start span (op `app.start.extended`, description `"Extended App Start"`).
* `finishExtendedAppStart` — finishes it (returns void); triggers finalization.
* `getExtendedAppStartSpan(): ISpan` — returns the span for attaching children; a no-op span when there's no active extension.

> `finishExtendedAppStart` is equivalent to calling `.finish()` on the returned span.

## Lifecycle

* `extendAppStart`**:** creates the span (start = call time). No-ops if app start already finished, if none is in progress, or on repeat calls (first wins).
* `getExtendedAppStartSpan`**:** returns the extended span, else a no-op span.
* `finishExtendedAppStart`**:** no-ops if not extended or called twice. Doesn't finalize directly — the parent (auto-generated, `waitForChildren`) finalizes when its last child finishes and trims its end to that child. Open children of the extended span are finished `cancelled` at the finish-call time.

## Duration & measurement

* Parent duration = end of the last child to finish (trim-to-last-child), **floored at the default app start end** — i.e. `end = max(last child end, default app start end)`. Extending can only push the end *later*; finishing before the default end keeps the default (never shorter than a non-extended app start). The extended span is just one participant.
* App start measurement = that final duration (process start → last child), set at **finalization**. The normal completion-time measurement must be suppressed in extended mode so the early (non-extended) value doesn't win.

## Timeout

[...]

PR #5606, opened 2026-06-23
feat(extend-app-start): [2/4] Extract AppStartExtension component

buenaflor:
## PR Stack (Extend App Start)

- #5604
- #5605
- #5606
- #5608
- #5628

---

## :scroll: Description

Pulls the Android extender out of the already-large `AppStartMetrics` into a focused, testable `AppStartExtension` component.

| Class | Change |
|---|---|
| **`AppStartExtension`** (new) | `implements IAppStartExtender`. Owns the lock-guarded extend lifecycle and holds the eager App Start `ITransaction` + its child `ISpan`. |
| **`AppStartMetrics`** | Holds the component and exposes the extend gate `canExtendAppStart()`. |
| **`AndroidOptionsInitializer`** | Registers the component as the options' app start extender. |

**Notes**

- `AppStartExtension` has no scopes, so it can't create the transaction/span itself — the integration builds them and **returns** an `ExtendedAppStart` from the listener.
- `getExtendedEndTime()` is deadline-aware: `null` on `DEADLINE_EXCEEDED`, so the vital is suppressed rather than inflated.
- `canExtendAppStart()` gate = measurements not sent + no activity + first frame not drawn. The foreground check is ignored so headless starts can extend as well.
- **Inert on its own**: no listener is registered until `[3/4]`, so `extendAppStart()` is a no-op and `getExtendedAppStartSpan()` returns `NoOpSpan`.

Part of the app start extension API stack (getsentry/sentry-java#5553). A dedicated component keeps the new concern out of `AppStartMetrics` and is easy to isolate and test.

Unit tests in `AppStartExtensionTest`, `AppStartMetricsTest`, and `AndroidOptionsInitializerTest` cover the extend gate, listener firing, inert behavior, idempotent finish, deadline suppression, and wiring. `apiCheck`, spotless, and the `:sentry-android-core` unit suite pass.

## :pencil: Checklist

## :crystal_ball: Next steps

`[3/4]` registers the listener, eagerly creates the App Start transaction + child, and extends/suppresses the app start vital in `PerformanceAndroidEventProcessor`.

#skip-changelog

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AppStartExtension.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AppStartExtension.java
@@ -60,7 +60,7 @@ public final class AppStartExtension implements IAppStartExtender {
             .log(SentryLevel.WARNING, "App start is already being extended.");
         return;
       }
-      if (!metrics.isAppStartWindowOpen()) {
+      if (!metrics.canExtendAppStart()) {
         Sentry.getCurrentScopes()
             .getOptions()
             .getLogger()
@@ -112,7 +112,6 @@ public final class AppStartExtension implements IAppStartExtender {
     try (final @NotNull ISentryLifecycleToken ignored = lock.acquire()) {
       final @Nullable ITransaction transaction = extendedTransaction;
       if (transaction != null && !transaction.isFinished()) {
-        // End at the extended span's finish if it ran past endTimestamp, so the txn covers it.
         final @Nullable ISpan span = extendedSpan;
         final @Nullable SentryDate spanEnd = span == null ? null : span.getFinishDate();
         final @NotNull SentryDate end =
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AppStartExtension.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AppStartExtension.java
@@ -5,8 +5,8 @@ import io.sentry.ILogger;
 import io.sentry.ISentryLifecycleToken;
 import io.sentry.ISpan;
 import io.sentry.ITransaction;
+import io.sentry.NoOpLogger;
 import io.sentry.NoOpSpan;
-import io.sentry.Sentry;
 import io.sentry.SentryDate;
 import io.sentry.SentryLevel;
 import io.sentry.SpanStatus;
@@ -42,6 +42,10 @@ public final class AppStartExtension implements IAppStartExtender {
   private final @NotNull AppStartMetrics metrics;
   private final @NotNull AutoClosableReentrantLock lock = new AutoClosableReentrantLock();
 
+  // Set once at SDK init via setLogger(), read later when an extension is requested. Defaults to a
+  // no-op because this component is created before SentryOptions (and its logger) exist.
+  private volatile @NotNull ILogger logger = NoOpLogger.getInstance();
+
   private @Nullable ExtendAppStartListener extendAppStartListener;
   private @Nullable ISpan extendedSpan;
   private @Nullable ITransaction extendedTransaction;
@@ -54,11 +58,15 @@ public final class AppStartExtension implements IAppStartExtender {
     this.extendAppStartListener = listener;
   }
 
+  void setLogger(final @NotNull ILogger logger) {
+    this.logger = logger;
+  }
+
   @Override
   public void extendAppStart() {
     try (final @NotNull ISentryLifecycleToken ignored = lock.acquire()) {
       if (extendedSpan != null) {
-        getLogger().log(SentryLevel.WARNING, "App start is already being extended.");
+        logger.log(SentryLevel.WARNING, "App start is already being extended.");
         return;
       }
       // Ignore the foreground check: headless app starts (broadcast/service) run in a
@@ -66,10 +74,9 @@ public final class AppStartExtension implements IAppStartExtender {
       // extension once an activity was created, the first frame was drawn, or measurements were
       // already sent.
       if (!metrics.isAppStartWindowOpen()) {
-        getLogger()
-            .log(
-                SentryLevel.WARNING,
-                "Cannot extend app start: the app start window has already passed.");
+        logger.log(
+            SentryLevel.WARNING,
+            "Cannot extend app start: the app start window has already passed.");
         return;
       }
       final @Nullable ExtendAppStartListener listener = extendAppStartListener;
@@ -162,9 +169,5 @@ public final class AppStartExtension implements IAppStartExtender {
       extendedSpan = null;
       extendedTransaction = null;
     }
-  }
-
-  private static @NotNull ILogger getLogger() {
-    return Sentry.getCurrentScopes().getOptions().getLogger();
   }
 }
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AppStartExtension.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AppStartExtension.java
@@ -94,9 +94,7 @@ public final class AppStartExtension implements IAppStartExtender {
   public @NotNull ISpan getExtendedAppStartSpan() {
     try (final @NotNull ISentryLifecycleToken ignored = lock.acquire()) {
       final @Nullable ISpan span = extendedSpan;
-      // Read the finish date, not isFinished(): finishing the extended span runs the event
-      // processor re-entrantly before the span's finished flag is set, but the finish timestamp is
-      // already in place. Matches getExtendedEndTime().
+      // Read the finish date, not isFinished() — see getExtendedEndTime() for the reentrancy reason.
       if (span != null && span.getFinishDate() == null) {
         return span;
       }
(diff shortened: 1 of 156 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/AppStartExtension.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+  }
+
+  public boolean isActive() {
+    try (final @NotNull ISentryLifecycleToken ignored = lock.acquire()) {
+      return extendedTransaction != null && !extendedTransaction.isFinished();
+    }
+  }
+
+  public void finishTransaction(final @NotNull SentryDate endTimestamp) {
```
buenaflor, line 147: this will be used in the next PR 3/4

sentry-android-core/src/main/java/io/sentry/android/core/AppStartExtension.java
```diff
@@ -0,0 +1,148 @@
...
+        span.finish(SpanStatus.OK);
+      }
+    }
+  }
+
+  @Override
+  public @NotNull ISpan getExtendedAppStartSpan() {
+    try (final @NotNull ISentryLifecycleToken ignored = lock.acquire()) {
+      final @Nullable ISpan span = extendedSpan;
+      if (span != null && !span.isFinished()) {
```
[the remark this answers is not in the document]
buenaflor: Valid mechanism — confirmed in `Span.finish()`: the finish timestamp is set before `spanFinishedCallback` runs the waitForChildren transaction re-entrantly, and `finished` flips only afterward. Although no current caller hits `getExtendedAppStartSpan()` during that window (the event processor only calls `getExtendedEndTime()`/`isExtended()`, and the lock serializes external callers so they always see `finished == true`), the fix is a cheap, consistent hardening. Switched to `getFinishDate() == null` to match `getExtendedEndTime()` in 83e2058480, with a regression test for the reentrancy window.

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
```diff
@@ -336,6 +339,21 @@ public long getClassLoadedUptimeMs() {
...
+  /**
+   * Whether the app start window is still open, i.e. an app start can be extended: measurements
+   * haven't been sent yet, no activity has been created, and the first frame hasn't been drawn. The
+   * foreground check is ignored so headless app starts (broadcast/service) can also be extended.
+   */
+  public boolean isAppStartWindowOpen() {
+    return shouldSendStartMeasurements(true)
+        && activeActivitiesCounter.get() == 0
+        && !firstDrawDone.get();
+  }
```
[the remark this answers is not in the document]
buenaflor: The observation about the flags is accurate — `onActivityDestroyed()` resets `shouldSendStartMeasurements`/`firstDrawDone` for warm-start re-measurement, so the gate reads open again after the last activity is destroyed. But the reopened window isn't reachable in practice: `extendAppStart()` has no internal caller (verified across the repo — only the external `Sentry.extendAppStart()` API), and its contract is to be called once from `Application.onCreate`, which does not re-run on a warm start (return from background). Config-change recreation is already excluded via `!isChangingConfigurations()`. Allowing a genuine warm start to extend is also intended. Resolving as not reachable given the call contract; if we ever expose a path that re-invokes `extendAppStart()` post-startup we'd revisit with a dedicated consumed-window latch.

sentry-android-core/src/main/java/io/sentry/android/core/AppStartExtension.java
```diff
@@ -0,0 +1,149 @@
...
+    }
+  }
+
+  @Override
+  public void finishExtendedAppStart() {
+    try (final @NotNull ISentryLifecycleToken ignored = lock.acquire()) {
+      final @Nullable ISpan span = extendedSpan;
+      if (span != null && !span.isFinished()) {
+        span.finish(SpanStatus.OK);
+      }
```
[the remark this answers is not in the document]
buenaflor: False positive — not a real issue, for two independent reasons:

1. **No reentrant caller.** The scenario (span.finish() → event processor → calls back into finishExtendedAppStart) does not exist. finishExtendedAppStart has no internal caller — it is only invoked via the external Sentry.finishExtendedAppStart() API. The event processor on the finish path calls isExtended()/getExtendedEndTime()/onAppStartSpansSent(), never finishExtendedAppStart(). So it is never re-entered during span.finish().

2. **span.finish() is idempotent.** Span.finish() self-guards: `if (finished || !isFinishing.compareAndSet(false, true)) return;`. Even within the reentrant window (finished still false, finishDate set), isFinishing is already true, so a second finish() call returns immediately — no double-finish, status/timestamp not overwritten.

This differs from getExtendedEndTime()/getExtendedAppStartSpan(), which were changed because they are read methods whose return value feeds downstream logic (getExtendedEndTime is literally called during the reentrant window to compute the vital; getExtendedAppStartSpan could hand out a finishing span). finishExtendedAppStart returns nothing and performs an idempotent action, so the isFinished()-vs-getFinishDate() distinction has no observable effect here. isFinished() is the natural guard in front of an idempotent finish() call; switching it would imply a reentrancy hazard that does not apply. Resolving.

sentry-android-core/src/main/java/io/sentry/android/core/AppStartExtension.java
```diff
@@ -0,0 +1,171 @@
...
+
+  @Override
+  public @NotNull ISpan getExtendedAppStartSpan() {
+    try (final @NotNull ISentryLifecycleToken ignored = lock.acquire()) {
+      final @Nullable ISpan span = extendedSpan;
+      // Mirrors getExtendedEndTime(): the finish date is set before isFinished() flips.
+      if (span != null && span.getFinishDate() == null) {
+        return span;
+      }
+      return NoOpSpan.getInstance();
```
romtsn: I'm wondering if a better API would actually be to make this method nullable and return null here? That way it would be explicit that there's no extended span, otherwise NoOpSpan might give a false impression that there's a span actually running?

sentry-android-core/src/main/java/io/sentry/android/core/AppStartExtension.java
```diff
@@ -0,0 +1,171 @@
...
+    @Nullable
+    ExtendedAppStart onExtendAppStartRequested();
+  }
+
+  private final @NotNull AppStartMetrics metrics;
+  private final @NotNull AutoClosableReentrantLock lock = new AutoClosableReentrantLock();
+
+  private @Nullable ExtendAppStartListener extendAppStartListener;
+  private @Nullable ISpan extendedSpan;
+  private @Nullable ITransaction extendedTransaction;
```
romtsn, line 54: would be great to have some docs on why we need to hold a ref to both the span and the transaction (I'm guessing because we want to finish the tx and set the vitals in a centralized place?), because for now it's a bit unclear

romtsn (approved):
one medium comment about the extended span API and a low-prio for having docs, but LGTM otherwise 👍

sentry-android-core/src/main/java/io/sentry/android/core/AppStartExtension.java
buenaflor: hm indeed, yeah that is def. clearer

sentry-android-core/src/main/java/io/sentry/android/core/AppStartExtension.java
buenaflor: in short:

- Span = what the user controls, they finish it with finishExtendedAppStart(), and it defines the extended duration.
- Transaction = the container that gets sent, the SDK (potentially) finishes it at the first-frame/headless end and attaches the screen to it.

it's awkward to get the other one if you only have one reference of either span or transaction

will add the docs

sentry-android-core/src/main/java/io/sentry/android/core/AppStartExtension.java
buenaflor: would you also make the public api nullable?

buenaflor: the non-null version would look like this

```java
public static @NotNull ISpan getExtendedAppStartSpan() {
  final @Nullable ISpan span =
      getCurrentScopes().getOptions().getAppStartExtender().getExtendedAppStartSpan();
  return span != null ? span : NoOpSpan.getInstance();
}
```

romtsn: yeah I'd also make it nullable so it follows `getSpan()` closely :)

sentry-android-core/src/main/java/io/sentry/android/core/AppStartExtension.java
buenaflor: Updated it

sentry-android-core/src/main/java/io/sentry/android/core/AppStartExtension.java
buenaflor: Updated it

PR #5607, opened 2026-06-23
fix(replay): Release MediaMuxer when the encoder fails to start

runningcode:
## :scroll: Description
The `MediaMuxer` is opened eagerly in `SimpleVideoEncoder`'s constructor, but its `release()` was only reachable on paths that assume `start()` succeeded. Two cases leaked it when the encoder failed to start:

- `ReplayCache.createVideoOf` constructed the encoder and called `start()` in a single expression (`SimpleVideoEncoder(...).also { it.start() }`). If `start()` threw, the assignment to the `encoder` field never happened, so the only caller of `release()` could never run.
- `SimpleVideoEncoder.release()` released the muxer as the last statement of the `try` block, after draining and stopping the codec. Draining a codec that never started throws, which jumped to the `catch` and skipped the muxer release.

The fix releases the encoder if `start()` throws (then rethrows), and always releases the muxer from a `finally` block so it is freed even when draining/stopping the codec fails.

This surfaced as a `CloseGuard` "A resource was acquired at attached stack trace but never released" warning pointing at the `MediaMuxer` allocation in `SimpleMp4FrameMuxer`. It is a real resource leak on devices where the encoder fails to start, not just a test artifact. Complements #5583, which closed the no-frames and never-started paths.

Added a regression test (`ReplayCacheTest.releases the muxer when the encoder fails to start`) that forces `start()` to fail via the test `MediaCodec` shadow and asserts, through `ShadowCloseGuard.getErrors()`, that no `MediaMuxer` leak is reported. Verified the test fails without the fix (with the exact `Explicit termination method 'release' not called` error) and passes with it. The full `ReplayCacheTest` suite is green.

## :pencil: Checklist

## :crystal_ball: Next steps
None.

### DIFF
--- a/sentry-android-replay/src/main/java/io/sentry/android/replay/video/SimpleVideoEncoder.kt
+++ b/sentry-android-replay/src/main/java/io/sentry/android/replay/video/SimpleVideoEncoder.kt
@@ -287,12 +287,14 @@ internal class SimpleVideoEncoder(
       onClose?.invoke()
       drainCodec(true)
       mediaCodec.stop()
+    } catch (e: RuntimeException) {
+      options.logger.log(DEBUG, "Failed to properly release video encoder", e)
+    } finally {
+      // always release the native resources, even if draining/stopping the codec above threw (e.g.
+      // when the encoder failed to fully start), otherwise they leak (CloseGuard warning)
       mediaCodec.release()
       surface?.release()
-
       frameMuxer.release()
-    } catch (e: Throwable) {
-      options.logger.log(DEBUG, "Failed to properly release video encoder", e)
     }
   }
 }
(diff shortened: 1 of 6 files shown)


romtsn (approved):
sorry this slipped, looks great!

sentry-android-replay/src/main/java/io/sentry/android/replay/video/SimpleVideoEncoder.kt
```diff
@@ -289,10 +289,12 @@ internal class SimpleVideoEncoder(
...
       mediaCodec.release()
       surface?.release()
-
-      frameMuxer.release()
     } catch (e: Throwable) {
       options.logger.log(DEBUG, "Failed to properly release video encoder", e)
+    } finally {
+      // always release the muxer, even if draining/stopping the codec above threw (e.g. when the
+      // encoder failed to fully start), otherwise its MediaMuxer leaks (CloseGuard warning)
+      frameMuxer.release()
```
romtsn: btw, wondering if we need to also move surface.release() and mediaCodec.release() in to the finally-block?

runningcode: Good point. Worth moving those especially if they can't throw.

romtsn (approved):
nice!

PR #5611, opened 2026-06-24
fix: use System.nanoTime() for cron check-in duration measurement

### LINKED ISSUE #5579: Cron check-in durations are measured with the wall clock
Audit finding **B4 — actual bug, LOW-MEDIUM**.

Cron check-in durations are computed as `currentTimeMillis() - start` (wall clock):

* `sentry/src/main/java/io/sentry/util/CheckInUtils.java:64`
* Same pattern in `SentryCheckInAdvice` across `sentry-spring`, `sentry-spring-jakarta`, and `sentry-spring-7`.

Cron jobs run long, giving wide exposure to wall-clock steps over the measured interval → wrong or negative durations reported.

Source: [JAVA-557](<https://linear.app/getsentry/issue/JAVA-557/clock-usage-audit-wall-vs-monotonic-time-issues-across-the-sdk>) §B4.

tsushanth:
Fixes #5579

## Problem

Cron check-in durations were computed by subtracting two `System.currentTimeMillis()` values. The wall clock is not monotonic — it is subject to NTP steps, leap-second smearing, and DST transitions. For long-running cron jobs this gives wide exposure to clock jumps, which can produce incorrect or negative duration values in the check-in payload.

## Fix

Switch the start/end capture to `System.nanoTime()`, which is guaranteed monotonic and is the correct choice for measuring elapsed time. `DateUtils.nanosToSeconds()` (already present in the SDK) is used to convert the nanosecond delta to the seconds value expected by `setDuration()`.

Changed files:
- `sentry/src/main/java/io/sentry/util/CheckInUtils.java`
- `sentry-spring/src/main/java/io/sentry/spring/checkin/SentryCheckInAdvice.java`
- `sentry-spring-jakarta/src/main/java/io/sentry/spring/jakarta/checkin/SentryCheckInAdvice.java`
- `sentry-spring-7/src/main/java/io/sentry/spring7/checkin/SentryCheckInAdvice.java`

Existing duration tests (`assertNotNull(doneCheckIn.duration)`) continue to pass unchanged; the delta of two `nanoTime()` calls is always non-negative and non-zero for any real workload

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixes
 
+- Use `System.nanoTime()` for cron check-in duration measurement to avoid incorrect durations from wall-clock adjustments ([#5611](https://github.com/getsentry/sentry-java/pull/5611))
 - Fix crash when `getHistoricalProcessStartReasons` is called from an isolated or wrong-userId process ([#5597](https://github.com/getsentry/sentry-java/pull/5597))
 - Release `MediaMuxer` when a replay segment has no encodable frames to avoid a resource leak ([#5583](https://github.com/getsentry/sentry-java/pull/5583))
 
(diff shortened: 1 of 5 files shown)


runningcode (approved):
Thank you for the contribution. This looks great!

adinauer (approved):
LGTM, thank you for the help!

PR #5615, opened 2026-06-24
test(replay): Fix flaky ComposeMaskingOptionsTest

### LINKED ISSUE #5585: Flaky test: ComposeMaskingOptionsTest "when sentry-unmask modifier is set unmasks the node"
## Summary

`ComposeMaskingOptionsTest > when sentry-unmask modifier is set unmasks the node` fails intermittently on `main` with:

```
java.lang.AssertionError: Node with text null should be masked
    at io.sentry.android.replay.viewhierarchy.ComposeMaskingOptionsTest.kt:238
```

(`:sentry-android-replay:testReleaseUnitTest`, Java 17 build job.)

## Evidence it's flaky, not a regression

* Same test, same assertion failed on two unrelated commits while commits in between passed:
  * Jun 22, commit `0c118e90` — [https://github.com/getsentry/sentry-java/actions/runs/27941528687/job/82675843497](<https://github.com/getsentry/sentry-java/actions/runs/27941528687/job/82675843497>)
  * Jun 18, commit `f6192aac` — [https://github.com/getsentry/sentry-java/actions/runs/27750608535](<https://github.com/getsentry/sentry-java/actions/runs/27750608535>)
* Both failures are preceded by a JVM instrumentation-agent error in the same run:<br>`*** java.lang.instrument ASSERTION FAILED ***: "!errorOutstanding" ... transform method call failed at .../JPLISAgent.c line: 876`

## Likely mechanism

In `ComposeViewHierarchyNode.fromComposeNode` (`ComposeViewHierarchyNode.kt:178-189`), masking is gated on visibility:

```kotlin
val isVisible =
  !SentryLayoutNodeHelper.isTransparent(node) &&
    (semantics == null || !semantics.contains(InvisibleToUser)) &&
    visibleRect.height > 0 && visibleRect.width > 0
...
val shouldMask = isVisible && semantics.shouldMask(isImage = false, options)
```

`visibleRect` comes from `node.coordinates.boundsInWindow(...)`. If a node's bounds resolve to empty (`width`/`height == 0`), `isVisible` is false and the node is not masked even with `maskAllText = true` — exactly the failing assertion (the activity-title node, whose layout isn't a `ComposeTextLayout`, hence `text == null`).

Under Robolectric, Compose layout/coordinate resolution after `shadowOf(Looper.getMainLooper()).idle()` is timing-sensitive; when layout hasn't fully settled (occasionally perturbed by the instrumentation-agent transform failures above), `boundsInWindow` returns empty bounds → `isVisible = false` → node not masked → intermittent failure.

## Possible fixes

* Await/assert non-empty bounds for the laid-out nodes before running the masking assertions.
* Make the masking assertions robust to the non-Compose activity-title node.
* Investigate the `java.lang.instrument` transform failure, which may be aggravating layout timing.

Unrelated to the MediaMuxer leak tracked in getsentry/sentry-java#5584 / PR #5583.

runningcode:
Fixes #5585

## Root cause

`ComposeMaskingOptionsTest` intermittently failed (~10% of full-class runs) with:

```
java.lang.AssertionError: Node with text null should be masked
    at ComposeMaskingOptionsTest.kt:238
```

I reproduced it locally and instrumented the collected nodes. The flaky node is **not** the activity-title `TextView` (that node uses `AndroidTextLayout` and is always laid out and visible). The flaky nodes are the **Compose** nodes (`Text`, `TextField`, `Button`): under Robolectric, a single `Looper.idle()` does not reliably complete a Compose recomposition + layout pass, so the entire `AndroidComposeView` subtree occasionally keeps zero-size bounds. Zero-bounds nodes are reported as not visible (`Views.isVisibleToUser` / the Compose `isVisible` gate), so masking skips them and `assertTrue(shouldMask)` fails on the first one (whose `ComposeTextLayout` is null → "text null").

Production behavior is correct (we shouldn't mask zero-size, invisible views), so this is a test-only timing issue.

## Fix

Before collecting the view hierarchy, drive several `idle` + `requestLayout` + `measure`/`layout` passes on the `AndroidComposeView` so its content reliably settles with non-empty bounds.

## Verification

- Before: reproduced the failure repeatedly (~10% of full-class runs).
- After: **97 consecutive full-class runs** passed (`--rerun-tasks`), the full `:sentry-android-replay:testReleaseUnitTest` suite is green, and spotless is clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Internal
 
 - Speed up gesture target hit-testing by mapping the touch point into local coordinates instead of calling `getLocationOnScreen` per view ([#5595](https://github.com/getsentry/sentry-java/pull/5595))
+- Fix flaky `ComposeMaskingOptionsTest` by settling Compose layout under Robolectric before collecting the view hierarchy ([#5615](https://github.com/getsentry/sentry-java/pull/5615))
 
 ## 8.44.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Release `MediaMuxer` when a replay segment has no encodable frames to avoid a resource leak ([#5583](https://github.com/getsentry/sentry-java/pull/5583))
 
+### Internal
+
+- Speed up gesture target hit-testing by mapping the touch point into local coordinates instead of calling `getLocationOnScreen` per view ([#5595](https://github.com/getsentry/sentry-java/pull/5595))
+
 ## 8.44.1
 
 ### Fixes
--- a/sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/ViewUtils.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/ViewUtils.java
@@ -1,13 +1,14 @@
 package io.sentry.android.core.internal.gestures;
 
 import android.content.res.Resources;
+import android.graphics.Matrix;
 import android.view.MotionEvent;
 import android.view.View;
 import android.view.ViewGroup;
 import io.sentry.android.core.SentryAndroidOptions;
 import io.sentry.internal.gestures.GestureTargetLocator;
 import io.sentry.internal.gestures.UiElement;
-import java.util.LinkedList;
+import java.util.ArrayDeque;
 import java.util.List;
 import java.util.Queue;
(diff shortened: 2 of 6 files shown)


runningcode:
Closed in favor of #5616 — this PR's head ref got stuck after a force-push (it kept showing unrelated JAVA-534 gesture files) and couldn't be reopened. #5616 contains only the flaky-test fix.

PR #5625, opened 2026-06-24
fix(changelog): Move app start reason to 8.45.0

romtsn:
slipped through to a wrong version

_#skip-changelog_


adinauer (approved):

0xadam-brown:
Thanks for this 🥇!

PR #5626, opened 2026-06-25
feat: add trace_metric_byte to DataCategory enum

### LINKED ISSUE #5618: Wire trace_metric_byte client outcome reporting
The `trace_metric_byte` data category is not tracked as a client outcome in sentry-java. The `trace_metric` outcome may be present, but the byte-count counterpart is missing.

- Wire `trace_metric_byte` client outcome reporting
- See https://develop.sentry.dev/sdk/telemetry/client-reports/ for data categories

tsushanth:
Fixes #5618

The `DataCategory` enum was missing a `trace_metric_byte` entry, which is defined in the Sentry SDK developer docs as a byte-count category for trace metrics. Without it, client outcome reporting cannot correctly attribute discarded trace metric bytes, so those losses go untracked.

This change adds `TraceMetricByte("trace_metric_byte")` to `DataCategory`, following the same pattern as `TraceMetric`, `LogItem`, `LogByte`, and other categories. The existing `DataCategory.valueOf(StringUtils.camelCase(catItem))` path in `RateLimiter` will automatically resolve `"trace_metric_byte"` to the new enum value when it appears in rate-limit response headers.


runningcode (approved):
Looks good. Can you run API dump to re-generate the api? https://github.com/getsentry/sentry-java/blob/main/CONTRIBUTING.md#binary-compatibility-validation

tsushanth:
Done — added `TraceMetricByte` to the API dump file manually (alphabetically after `TraceMetric` in the `DataCategory` block).

adinauer (approved):
Thanks for the contribution! Changes look good to me but while reviewing I noticed some gaps in both this and the pre-existing `log_byte` handling:
- `EventProcessor` discards aren't tracked
- Rate limiter doesn't apply the `*_byte` categories to the respective feature

I'll open PRs to address these.

PR #5627, opened 2026-06-25
fix: guard executor shutdown in BaseCaptureStrategy.stop()

### LINKED ISSUE #5564: Fix leaked persisting executor in BaseCaptureStrategy
Each `ReplayIntegration.start()` creates a new `SessionCaptureStrategy` or `BufferCaptureStrategy`. Each strategy owns a lazy `persistingExecutor` in `BaseCaptureStrategy`, but `stop()` never shuts it down. When `stop()` runs on the main thread, it resets delegated properties via `persistableAtomic*`, which initializes the executor — leaving one leaked `SentryReplayPersister-*` thread per start/stop cycle.

Discovered via Kiln Android benchmarks after [kiln#73](<https://github.com/getsentry/kiln/pull/73>) changed replay from a single long session to repeated start/stop loops: memory and CPU climbed monotonically, GC allocation rate for `sdk-full` reached ~9× baseline.

**Root cause:** `BaseCaptureStrategy.stop()` initializes but never shuts down the `persistingExecutor` on main-thread stop paths.

**Recommended fix (PR1):**

* Reuse the executor across stop/start cycles, or perform a non-blocking shutdown after stop-time property resets.
* Do not call `ReplayExecutorService.shutdown()` on the stop path — it blocks and risks ANRs.
* Store the raw delegate or an explicit lazy holder so shutdown only runs if the executor was initialized.

**Follow-up:**

* Add a regression test asserting no growth in `SentryReplayPersister-*` threads across many start/stop cycles.
* Rerun Kiln charts after fix; if GC blocking persists, investigate async bitmap/surface cleanup separately.

Action taken on behalf of nelson.osacky.

tsushanth:
Fixes #5564

## Problem

Each `ReplayIntegration.start()` constructs a fresh `SessionCaptureStrategy` or `BufferCaptureStrategy`. Both inherit `BaseCaptureStrategy`, which owns a `persistingExecutor` that was previously declared as a Kotlin `lazy` delegate.

`stop()` resets the delegated properties `segmentTimestamp` and `currentReplayId`. Their setters call `runInBackground`, which — when `options.threadChecker.isMainThread()` is true — accesses `persistingExecutor`, silently initialising the lazy. `stop()` never shuts the executor down, so one `SentryReplayPersister-*` thread is abandoned on every cycle. Kiln benchmark data confirmed monotonically growing thread counts and a GC allocation rate roughly 9× baseline after the repeated start/stop pattern introduced in kiln#73.

## Fix

Replace the `lazy` delegate with an explicit nullable holder (`persistingExecutorHolder`). The custom `get()` mirrors the old lazy behaviour (create-on-first-access) while making initialisation detectable. At the end of `stop()`, if the holder is non-null the executor is shut down via `shutdownNow()` — non-blocking, safe to call on the main thread — and the holder is cleared so a subsequent `start()` gets a fresh executor without any residual state.

`ReplayExecutorService.shutdown()` is intentionally not used here: it blocks for `options.shutdownTimeoutMillis` and risks an ANR when invoked on the main thread. `shutdownNow()` interrupts any in-flight persistence write and discards the queue, which is acceptable because `cache.close()` is called immediately before in the same `stop()` body.

## Test

Added `stop shuts down persisting executor so no SentryReplayPersister threads leak across cycles` to `SessionCaptureStrategyTest`. The test stubs `threadChecker.isMainThread()` to `true` (forcing the persisting executor path), runs three start/stop cycles, and asserts that no `SentryReplayPersister-*` threads remain alive after a short drain window.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+<<<<<<< HEAD
 ### Behavioral Changes
 
 - `SentryOkHttpInterceptor::intercept` now throws `IOException`. This is a source-only and Java-only breaking change ([#5654](https://github.com/getsentry/sentry-java/pull/5654))
@@ -11,6 +12,7 @@
 - Don't start a redundant UI interaction transaction when a transaction is already bound to the Scope ([#5491](https://github.com/getsentry/sentry-java/issues/5491))
   - Previously, `SentryGestureListener` always started a UI transaction and only afterwards skipped binding it to the Scope when a manually-bound transaction already existed, leaving the new transaction to be dropped as an idle transaction without children.
 - Fix potential NPE within `Scope.endSession()` ([#5657](https://github.com/getsentry/sentry-java/pull/5657))
+- Fix memory leak in `ReplayIntegration` due to persisting executor not being shut down ([#5627](https://github.com/getsentry/sentry-java/pull/5627))
 
 ### Performance
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,6 @@
 
 ## Unreleased
 
-<<<<<<< HEAD
 ### Behavioral Changes
 
 - `SentryOkHttpInterceptor::intercept` now throws `IOException`. This is a source-only and Java-only breaking change ([#5654](https://github.com/getsentry/sentry-java/pull/5654))
--- a/sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
+++ b/sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
@@ -111,6 +111,11 @@ public class ReplayIntegration(
     val delegate = Executors.newSingleThreadScheduledExecutor(ReplayExecutorServiceThreadFactory())
     ReplayExecutorService(delegate, options)
   }
+  private val persistingExecutor by lazy {
+    val delegate =
+      Executors.newSingleThreadScheduledExecutor(ReplayPersistingExecutorServiceThreadFactory())
+    ReplayExecutorService(delegate, options)
+  }
 
   internal val isEnabled = AtomicBoolean(false)
   internal val isManualPause = AtomicBoolean(false)
@@ -192,6 +197,7 @@ public class ReplayIntegration(
               scopes,
               dateProvider,
               replayExecutor,
+              persistingExecutor,
               replayCacheProvider,
             )
           } else {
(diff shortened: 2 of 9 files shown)


runningcode:
Thanks for contributing, could you run spotlessApply to fix the formatting failure? Then we'll review this more closely.

romtsn (changes requested):
@tsushanth thanks for your contribution, that's very much appreciated! I'm thinking if there's potentially a better approach, which would be to pull `persistingExecutor` up to the `ReplayIntegration` level and then pass it as a ctor argument to the respective CaptureStrategy. 

It could then have the same lifecycle as the `replayExecutor` (that is, shut it down inside `ReplayIntegration.close()`) and would survive multiple start/stop calls, potentially saving us the cost of creating a new executor every time we start a new recording.

tsushanth:
Refactored per the feedback — `persistingExecutor` is now owned by `ReplayIntegration` instead of `BaseCaptureStrategy`:

- Declared as a `lazy` field in `ReplayIntegration` using the same pattern as `replayExecutor` (thread factory moved there too)
- Passed as a constructor argument to `SessionCaptureStrategy` and `BufferCaptureStrategy`, which forward it up to `BaseCaptureStrategy`
- `BufferCaptureStrategy.convert()` hands both executors to the new `SessionCaptureStrategy` to preserve ordering
- `ReplayIntegration.close()` shuts down `persistingExecutor` alongside `replayExecutor`
- `BaseCaptureStrategy.stop()` no longer touches the executor — the holder/lazy workaround and per-stop shutdown are gone

Tests updated to pass a mock `persistingExecutor` matching the existing mock pattern.

tsushanth:
Refactored as suggested @romtsn — `persistingExecutor` is now owned by `ReplayIntegration` and passed as a constructor argument to the capture strategies. It shares the same lifecycle as `replayExecutor`: created lazily on first use, survives `stop()`/`start()` cycles, and shut down in `ReplayIntegration.close()`.

Also ran `spotlessApply` to fix the formatting failure @runningcode flagged.

tsushanth:
Hi @romtsn — good suggestion! I've updated the PR to follow exactly that approach: `persistingExecutor` is now created as a lazy property on `ReplayIntegration` (same pattern as `replayExecutor`) and passed down to `BaseCaptureStrategy` as a constructor argument. It's shut down inside `ReplayIntegration.close()` alongside `replayExecutor`, so it survives start/stop cycles and is only torn down when the integration itself closes. Happy to adjust anything further.

romtsn (approved):
@tsushanth thanks again, I've updated the changelog and moved the tests for this regression into `ReplayIntegrationTest`. Looks good to me now!

PR #5635, opened 2026-06-25
perf(core): [Init Reflection 1] Probe class availability without initializing

runningcode:
## PR Stack (Init Reflection)

- #5634 — collection
- #5635 — [1] Probe class availability without initializing
- #5636 — [2] Cache class lookups and collapse double probes
- #5637 — [3] Gate Compose class probes behind their features

---

Part of [JAVA-587](https://linear.app/getsentry/issue/JAVA-587/reduce-reflection-cost-on-the-sdk-init-path)

## :scroll: Description

`LoadClass.loadClass` used `Class.forName(name)`, which **initializes** the class. Since this method is used purely to probe whether an optional integration is on the classpath during `SentryAndroid.init`, that eagerly runs unrelated static initializers — the customer trace shows `androidx.compose.ui.node.Owner.<clinit>` and `FragmentLifecycleIntegration.<clinit>` executing under the availability check.

This switches to `Class.forName(name, false, classLoader)` so the class is loaded but only initialized lazily on first real use (e.g. when it's actually instantiated).

First of three stacked PRs reducing reflection cost on the init path, from the customer-provided Perfetto trace in the *Reduce SDK init time [Android]* project (JAVA-586 area).

New `LoadClassTest` including a guard asserting that probing a class does **not** run its static initializer; existing init/integration tests pass unchanged.

## :pencil: Checklist

## :crystal_ball: Next steps

PR 2 caches lookups and collapses double-probes; PR 3 gates the Compose probes behind their features.

## :stopwatch: Pixel 3 benchmark (ART method trace → Perfetto trace_processor)

Probing a class that has a (deliberately heavy) static initializer:

| | old `Class.forName(name)` | new `Class.forName(name, false, loader)` |
|---|---:|---:|
| target-class `<clinit>` invocations under the probe | 1 | **0** |

The static initializer runs under the old probe and is entirely skipped under the new one. In the production trace this was `androidx.compose.ui.node.Owner.<clinit>` and `FragmentLifecycleIntegration.<clinit>` running during init. (Method tracing inflates the absolute `<clinit>` time, so only the invocation count is reported.)

## :stopwatch: Cold-start macrobenchmark (Pixel 3, `sentry-samples-android`)

Jetpack Macrobenchmark, `StartupMode.COLD` + `CompilationMode.Full()`, 20 iterations. A new `android.os.Trace` marker (`SentryInitProbes`, a no-op unless tracing is active) lets a `TraceSectionMetric` measure just the 6 `isClassAvailable` probes instead of the ~560 ms whole start — so the sub-millisecond effect clears the noise floor.

| `SentryInitProbes` duration | median | min | max |
|---|---:|---:|---:|
| old (`forName` initializes) | 0.833 ms | 0.725 | 1.056 |
| **new (no-init)** | **0.434 ms** | 0.399 | 0.549 |

**Δ median ≈ −399 µs** per init on this app (timber + fragment + replay present), with non-overlapping ranges. The saving is the deferred static initializers — dominated by `ReplayIntegration.<clinit>` — which now run lazily on first use instead of under the probe. At whole-startup granularity (`timeToInitialDisplay` ≈ 563 ms ± 50 ms) the change is below run-to-run noise; it's a critical-path micro-optimization, not a measurable end-to-end startup win on its own.

### DIFF
--- a/sentry/src/main/java/io/sentry/util/LoadClass.java
+++ b/sentry/src/main/java/io/sentry/util/LoadClass.java
@@ -12,17 +12,23 @@ import org.jetbrains.annotations.Nullable;
 public class LoadClass {
 
   /**
-   * Try to load a class via reflection
+   * Loads and initializes a class via reflection. Use this when you intend to actually use the
+   * class (e.g. instantiate it or invoke its methods). The returned class is fully initialized, so
+   * its static initializers run. To merely check whether a class is on the classpath, use {@link
+   * #isClassAvailable} instead, which avoids running those initializers.
    *
    * @param clazz the full class name
    * @param logger an instance of ILogger
    * @return a Class&lt;?&gt; if it's available, or null
    */
   public @Nullable Class<?> loadClass(final @NotNull String clazz, final @Nullable ILogger logger) {
+    return loadClass(clazz, logger, true);
+  }
+
+  private @Nullable Class<?> loadClass(
+      final @NotNull String clazz, final @Nullable ILogger logger, final boolean initialize) {
     try {
-      // Don't initialize the class just to probe for availability; it gets initialized lazily on
-      // first use. This avoids running unrelated static initializers during SDK init.
-      return Class.forName(clazz, false, LoadClass.class.getClassLoader());
+      return Class.forName(clazz, initialize, LoadClass.class.getClassLoader());
     } catch (ClassNotFoundException e) {
       if (logger != null) {
         logger.log(SentryLevel.INFO, "Class not available: " + clazz);
@@ -39,8 +45,19 @@ public class LoadClass {
     return null;
   }
 
+  /**
+   * Probes whether a class is on the classpath without initializing it. Use this for availability
+   * checks (e.g. deciding whether to register an integration); the class is not initialized, so its
+   * static initializers do not run until something actually uses it. This keeps SDK init cheap by
+   * not triggering unrelated initializers. If you need to use the class, use {@link #loadClass}
+   * instead.
+   *
+   * @param clazz the full class name
+   * @param logger an instance of ILogger
+   * @return true if the class is on the classpath
+   */
   public boolean isClassAvailable(final @NotNull String clazz, final @Nullable ILogger logger) {
-    return loadClass(clazz, logger) != null;
+    return loadClass(clazz, logger, false) != null;
   }
 
   public boolean isClassAvailable(
@@ -48,6 +65,15 @@ public class LoadClass {
     return isClassAvailable(clazz, options != null ? options.getLogger() : null);
   }
 
+  /**
+   * Like {@link #isClassAvailable}, but defers the (non-initializing) availability check until the
+   * result is first read. Use this when the check itself should not run during SDK init but only
+   * later, on first access.
+   *
+   * @param clazz the full class name
+   * @param logger an instance of ILogger
+   * @return a lazily-evaluated availability check
+   */
   public LazyEvaluator<Boolean> isClassAvailableLazy(
       final @NotNull String clazz, final @Nullable ILogger logger) {
     return new LazyEvaluator<>(() -> isClassAvailable(clazz, logger));
--- a/sentry/src/main/java/io/sentry/util/LoadClass.java
+++ b/sentry/src/main/java/io/sentry/util/LoadClass.java
@@ -20,7 +20,9 @@ public class LoadClass {
    */
   public @Nullable Class<?> loadClass(final @NotNull String clazz, final @Nullable ILogger logger) {
     try {
-      return Class.forName(clazz);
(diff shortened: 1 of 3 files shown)


sentry/src/main/java/io/sentry/util/LoadClass.java
```diff
@@ -20,7 +20,9 @@ public class LoadClass {
    */
   public @Nullable Class<?> loadClass(final @NotNull String clazz, final @Nullable ILogger logger) {
     try {
-      return Class.forName(clazz);
+      // Don't initialize the class just to probe for availability; it gets initialized lazily on
+      // first use. This avoids running unrelated static initializers during SDK init.
+      return Class.forName(clazz, false, LoadClass.class.getClassLoader());
```
adinauer: In `SentryNdk` the static init block will now run much later, potentially causing ANRs.

Previously `Class.forName` triggered `static block` early and gave it time to run `SentryNdk.loadNativeLibraries()` in the background before we waited for `loadLibraryLatch.await` in the `init` method.

Now this may cause applications to spend more time waiting on main thread. Worst case would be 2s additional wait on main thread.

adinauer: This also leads to a very theoretical case of classes being handed back that then fail when creating an instance. This could bite us on the OTel span factory and scopes storage. One fix here could be to just catch `Throwable` in `SpanFactoryFactory` and `ScopesStorageFactory` instead of all the individual exceptions where we might then miss ExceptionInInitializerError/NoClassDefFoundError/LinkageError.

But looking at the finding above, it may make sense to allow controlling `true`/`false` for the `Class.forName` call from the caller.

This however would increase complexity of caching results since we might have invoked `Class.forName` with `false` already and cached the result, then another caller might want `true` but since we already cached it it won't do it.

runningcode: Thanks for the information! I didn't know this. The idea here was to use this performance enhancement in places that are just checking for the presence of classes.

I propose adjusting only the `isClassAvailable` method and leaving the `loadClass` method untouched.

That being said, if we are relying on NDK integration loading in the background and then blocking on it that sound like a potential for an ANR with or without this change.

adinauer: > That being said, if we are relying on NDK integration loading in the background and then blocking on it that sound like a potential for an ANR with or without this change.

The way this is implemented on `main`, gives the SDK a chance to perform other work before waiting for NDK to be loaded. I don't think we could remove the await since we're relying on NDK to perform some work before we trigger outbox sender. Not sure what'd happen if NDK was still moving files after we started sending out.

@markushi has already implemented some optimizations around NDK init, so he may have more details here.

>  propose adjusting only the isClassAvailable method and leaving the loadClass method untouched.

Could you performance test, whether calling `Class.forName` again for classes that can be found has a real performance cost? Maybe we could just cache whether a class exists (`isClassAvailable`) and make `loadClass` check if the class has already been cached as not found. My idea is that it may only have a large performance impact to call `Class.forName` if the class doesn't exist because then there's no cache. This might behave differently for different JDK versions and Android versions.

The simpler solution would probably be what you proposed and not use `loadClass` from `isClassAvailable`, if that's what you mean.

runningcode: Here's a benchmark I did. Let me know if you'd like me to commit it or see the benchmark.
 Otherwise the code has been adjusted so only the `isClassAvailable` changes in behavior and `loadClass` stays the same.
 
 ## Cold-start macrobenchmark (Pixel 3, `sentry-samples-android`)

  Jetpack Macrobenchmark, `StartupMode.COLD` + `CompilationMode.Full()`, 20 iterations. I added an `android.os.Trace` marker
  (`SentryInitProbes`, a no-op unless tracing is active) around the 6 `isClassAvailable` probes in `SentryAndroid.init`, so a
  `TraceSectionMetric` measures just that block instead of the ~560 ms whole start — which is what lets a sub-millisecond effect clear the noise
  floor.

  | `SentryInitProbes` duration | median | min | max |
  |---|---:|---:|---:|
  | old (`forName` initializes) | 0.833 ms | 0.725 | 1.056 |
  | **new (no-init)** | **0.434 ms** | 0.399 | 0.549 |

  **Δ median ≈ −399 µs** per init on this app (timber + fragment + replay present), with **non-overlapping** distributions (new max 0.549 < old
  min 0.725). The saving is the deferred static initializers — dominated by `ReplayIntegration.<clinit>` — which now run lazily on first use
  instead of under the availability probe.

romtsn (approved):
looks great, thanks for doing additional benchmarks! I guess using bytecode manipulation to inject these class names would still gain us even more savings, but after this lands, it's not that crucial anymore

adinauer (approved):
LGTM

PR #5636, opened 2026-06-25
perf(core): [Init Reflection 2] Cache class lookups and collapse double probes

runningcode:
## PR Stack (Init Reflection)

- #5634 — collection
- #5635 — [1] Probe class availability without initializing
- #5636 — [2] Cache class lookups and collapse double probes
- #5637 — [3] Gate Compose class probes behind their features

---

Part of [JAVA-587](https://linear.app/getsentry/issue/JAVA-587/reduce-reflection-cost-on-the-sdk-init-path)

## :scroll: Description

Two related reflection-cost reductions on top of [Init Reflection 1]:

- **Cache `LoadClass` results.** Whether a class is on the classpath doesn't change during the process lifetime, so results (including "not available") are cached in a static map. Repeated probes of the same class skip `Class.forName` entirely — and skip the `ClassNotFoundException` thrown for absent classes.
- **Collapse double-probes.** `SpanFactoryFactory` and `ScopesStorageFactory` called `isClassAvailable(X)` and then `loadClass(X)` — two `Class.forName` calls for the same class. They now call `loadClass(X)` once and null-check.

Second of three stacked PRs reducing reflection cost on the init path (customer-provided Perfetto trace, *Reduce SDK init time [Android]*).

Existing `LoadClassTest`, factory, and `Sentry` init tests pass; the cache returns identical availability results.

## :pencil: Checklist

## :crystal_ball: Next steps

PR 3 gates the Compose class probes behind the features that consume them.

## :stopwatch: Pixel 3 benchmark (untraced timing)

Repeatedly probing the same absent class (2000 calls × 6 rounds; cache hits after the first miss):

| | old (uncached `forName` each call) | new (cached) |
|---|---:|---:|
| per probe | 25,872 ns | **137 ns** |

~**188× faster** for repeat probes — once the first miss caches `NOT_AVAILABLE`, later probes skip both the `Class.forName` call and the `ClassNotFoundException`. (Traced counts are noisy here due to background class loading, so this is measured untraced.)

### DIFF
--- a/sentry/src/main/java/io/sentry/util/LoadClass.java
+++ b/sentry/src/main/java/io/sentry/util/LoadClass.java
@@ -12,17 +12,23 @@ import org.jetbrains.annotations.Nullable;
 public class LoadClass {
 
   /**
-   * Try to load a class via reflection
+   * Loads and initializes a class via reflection. Use this when you intend to actually use the
+   * class (e.g. instantiate it or invoke its methods). The returned class is fully initialized, so
+   * its static initializers run. To merely check whether a class is on the classpath, use {@link
+   * #isClassAvailable} instead, which avoids running those initializers.
    *
    * @param clazz the full class name
    * @param logger an instance of ILogger
    * @return a Class&lt;?&gt; if it's available, or null
    */
   public @Nullable Class<?> loadClass(final @NotNull String clazz, final @Nullable ILogger logger) {
+    return loadClass(clazz, logger, true);
+  }
+
+  private @Nullable Class<?> loadClass(
+      final @NotNull String clazz, final @Nullable ILogger logger, final boolean initialize) {
     try {
-      // Don't initialize the class just to probe for availability; it gets initialized lazily on
-      // first use. This avoids running unrelated static initializers during SDK init.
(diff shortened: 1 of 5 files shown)


sentry/src/main/java/io/sentry/util/LoadClass.java
```diff
@@ -19,11 +31,18 @@ public class LoadClass {
...
   public @Nullable Class<?> loadClass(final @NotNull String clazz, final @Nullable ILogger logger) {
+    final @Nullable Object cached = CLASSES.get(clazz);
+    if (cached != null) {
+      return cached == NOT_AVAILABLE ? null : (Class<?>) cached;
+    }
     try {
       // Don't initialize the class just to probe for availability; it gets initialized lazily on
       // first use. This avoids running unrelated static initializers during SDK init.
-      return Class.forName(clazz, false, LoadClass.class.getClassLoader());
+      final Class<?> loadedClass = Class.forName(clazz, false, LoadClass.class.getClassLoader());
```
adinauer: See my comments on https://github.com/getsentry/sentry-java/pull/5635#discussion_r3481764621

runningcode:
There's no point to the cache since we don't call these more than once.

PR #5638, opened 2026-06-25
build: Remove redundant mavenCentral repository declarations

runningcode:
## :scroll: Description
Removes the module-level `repositories { mavenCentral() }` block from all 25 build files that declared it — `sentry-reactor` and 24 sample modules.

`settings.gradle.kts` already declares `google()`, `mavenCentral()` and `mavenLocal()` centrally via `dependencyResolutionManagement` for every project, so these per-module blocks were redundant. With Gradle's default `PREFER_PROJECT` mode, declaring `mavenCentral()` in a module actually *narrowed* that module to only mavenCentral and ignored the central repositories; removing the block lets it fall back to the central superset, with no effect on resolution.

This is a follow-up to the earlier redundant build-script cleanups (#5624, #5633).

## :pencil: Checklist

## :crystal_ball: Next steps

#skip-changelog

### DIFF
--- a/sentry-reactor/build.gradle.kts
+++ b/sentry-reactor/build.gradle.kts
@@ -62,8 +62,6 @@ tasks.withType<JavaCompile>().configureEach {
   }
 }
 
-repositories { mavenCentral() }
-
 tasks.jar {
   manifest {
     attributes(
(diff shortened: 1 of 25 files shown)


0xadam-brown (approved):
⚡

PR #5639, opened 2026-06-25
chore(ci): ignore UI test suite timeout failures

romtsn:
## Summary
- Treat SauceLabs "Suite timed out" the same as "Process crashed" — exit 0 in the `Verify Test Results` step so infrastructure timeouts don't block PRs.
- Example failure: https://github.com/getsentry/sentry-java/actions/runs/28164799834/job/83414025923

## Test plan
- [x] Verified the timeout log message matches `Suite timed out` from the failing run

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,6 +49,8 @@ jobs:
         run: curl -sL https://sentry.io/get-cli/ | bash
 
       - name: Upload Snapshots to Sentry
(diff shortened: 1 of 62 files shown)


runningcode:
If these don't block merging, wouldn't this just hide the fact that these are happening more often? 

Otherwise a PR might go green because this failure was ignored rather than rerun because it could have been a real failure.

PR #5641, opened 2026-06-25
perf(android): Defer SentryFrameMetricsCollector thread startup

runningcode:
Resolves [JAVA-535](https://linear.app/getsentry/issue/JAVA-535/defer-sentryframemetricscollector-thread-startup)

## :scroll: Description

`SentryFrameMetricsCollector` created and started its `HandlerThread` in the constructor and then called `new Handler(handlerThread.getLooper())`. `getLooper()` **blocks** the caller until the new thread's looper is ready — and the collector is constructed on the **main thread** during `SentryAndroid.init` (`loadDefaultAndMetadataOptions`).

The `handler` is only ever used by `trackCurrentWindow()`, which no-ops until a listener is registered via `startCollection()`. So this starts the `HandlerThread` lazily on the first `startCollection()` (profiling / `SpanFrameMetricsCollector`). The constructor still registers the activity-lifecycle callback (current-window tracking is unchanged) but no longer blocks. Apps that never collect frame metrics never start the thread at all.

Reusing the SDK's shared executor isn't viable here: `addOnFrameMetricsAvailableListener` requires a dedicated non-main-thread `Handler` (Looper), which the shared `ScheduledExecutorService` doesn't provide — so the right move is to defer the dedicated thread, not share one.

From the customer-provided Perfetto trace (`sentryinvest`): the main thread blocked ~5.75 ms on `HandlerThread.getLooper()` during init via `SentryFrameMetricsCollector.<init>`.

`SentryFrameMetricsCollectorTest` (incl. a new test asserting `handler` is null after construction and non-null after `startCollection`); full `sentry-android-core` suite passes.

**Pixel 3 (Android 12) benchmark — ART method trace → Perfetto trace_processor:**

| construction | `HandlerThread.getLooper()` | blocking `Object.wait()` |
|---|---:|---:|
| old (pre-change) | 1 | 1 |
| new (deferred) | **0** | **0** |

The synchronous main-thread wait for the frame-metrics looper is entirely removed from construction.

**Cold-start A/B — `sentry-samples-android` (release, Pixel 3, 15 cold starts each via `am start -W -S`, `TotalTime` ms, first cold-disk run dropped):**

| | median | mean | min | stdev |
|---|---:|---:|---:|---:|
| with change | 434 | 434 | 416 | 9 |
| without change | 441 | 459 | 421 | 49 |

Directionally faster on every statistic, but **within end-to-end cold-start noise**. Note this sample auto-inits Sentry *and* calls `Sentry.startProfiler()` in `onCreate`, so frame collection (the lazy-thread trigger) fires during startup either way — the change just moves *when* within startup. So this isn't a strong end-to-end showcase; the deterministic proof is the ART trace above. The real win is for apps that don't start frame collection at startup (the thread is then never created at init).

> ⚠️ **Caveat:** these runs were captured in a very warm room (~34 °C), so thermal throttling may have affected the absolute timings and likely contributed to the elevated variance / spikes in the "without change" run (stdev 49 vs 9).

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
@@ -14,12 +14,14 @@ import android.view.FrameMetrics;
 import android.view.Window;
 import androidx.annotation.RequiresApi;
 import io.sentry.ILogger;
+import io.sentry.ISentryLifecycleToken;
 import io.sentry.SentryLevel;
 import io.sentry.SentryOptions;
 import io.sentry.SentryUUID;
 import io.sentry.android.core.BuildInfoProvider;
 import io.sentry.android.core.ContextUtils;
 import io.sentry.android.core.SentryFramesDelayResult;
+import io.sentry.util.AutoClosableReentrantLock;
 import io.sentry.util.Objects;
 import java.lang.ref.WeakReference;
 import java.lang.reflect.Field;
@@ -45,7 +47,8 @@ public final class SentryFrameMetricsCollector implements Application.ActivityLi
   private final @NotNull Set<Window> trackedWindows = new CopyOnWriteArraySet<>();
 
   private final @NotNull ILogger logger;
-  private @Nullable Handler handler;
+  private volatile @Nullable Handler handler;
+  private final @NotNull AutoClosableReentrantLock handlerLock = new AutoClosableReentrantLock();
   private @Nullable WeakReference<Window> currentWindow;
   private final @NotNull Map<String, FrameMetricsCollectorListener> listenerMap =
       new ConcurrentHashMap<>();
@@ -113,12 +116,8 @@ public final class SentryFrameMetricsCollector implements Application.ActivityLi
     }
     isAvailable = true;
 
-    HandlerThread handlerThread =
-        new HandlerThread("io.sentry.android.core.internal.util.SentryFrameMetricsCollector");
-    handlerThread.setUncaughtExceptionHandler(
-        (thread, e) -> logger.log(SentryLevel.ERROR, "Error during frames measurements.", e));
-    handlerThread.start();
-    handler = new Handler(handlerThread.getLooper());
+    // The frame metrics HandlerThread is started lazily on the first startCollection() call.
+    // Starting it here would block the main thread on HandlerThread.getLooper() during SDK init.
 
     // We have to register the lifecycle callback, even if no profile is started, otherwise when we
     // start a profile, we wouldn't have the current activity and couldn't get the frameMetrics.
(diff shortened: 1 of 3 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
```diff
@@ -113,12 +116,8 @@ public SentryFrameMetricsCollector(
     }
     isAvailable = true;
 
-    HandlerThread handlerThread =
-        new HandlerThread("io.sentry.android.core.internal.util.SentryFrameMetricsCollector");
-    handlerThread.setUncaughtExceptionHandler(
-        (thread, e) -> logger.log(SentryLevel.ERROR, "Error during frames measurements.", e));
-    handlerThread.start();
-    handler = new Handler(handlerThread.getLooper());
+    // The frame metrics HandlerThread is started lazily on the first startCollection() call.
```
runningcode, line 119: I feel like this comment isn't useful here since this isn't where one would look for it. I'll remove it after a review unless anyone thinks otherwise.

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
```diff
@@ -281,12 +280,34 @@ public void onActivityDestroyed(@NotNull Activity activity) {}
...
     final String uid = SentryUUID.generateSentryId();
     listenerMap.put(uid, listener);
     trackCurrentWindow();
     return uid;
   }
 
+  /**
+   * Lazily starts the background HandlerThread used to receive frame metrics. Deferred out of the
+   * constructor because {@link HandlerThread#getLooper()} blocks the caller (the main thread during
+   * SDK init) until the thread is ready, and the handler is only needed once collection starts.
```
0xadam-brown, line 293: 🥇

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
0xadam-brown: Agreed ⚡

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
```diff
@@ -281,12 +280,34 @@ public void onActivityDestroyed(@NotNull Activity activity) {}
...
+    try (final @NotNull ISentryLifecycleToken ignored = handlerLock.acquire()) {
+      if (handler == null) {
+        final HandlerThread handlerThread =
+            new HandlerThread("io.sentry.android.core.internal.util.SentryFrameMetricsCollector");
+        handlerThread.setUncaughtExceptionHandler(
+            (thread, e) -> logger.log(SentryLevel.ERROR, "Error during frames measurements.", e));
+        handlerThread.start();
+        handler = new Handler(handlerThread.getLooper());
+      }
+    }
```
0xadam-brown, line 308: m: Thoughts about catching `Exception`? (eg, looks like there's an off chance `handlerThread.getLooper()` could return null + cause the `Handler` constructor to throw an NPE if the thread dies before its looper is prepared). 

(I realize we didn't do so before, but now we'll be starting the handler thread outside of `Sentry.init()` at a time when developers may be less likely to have their own try-catch or when it's harder to diagnose.)

0xadam-brown (commented):
Nice! That's a real win 🧵 ⚡ 💯  

One question for your consideration; otherwise lgtm.

romtsn (approved):
love how easy and impactful these changes are 🚀

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
runningcode: My thoughts: if we weren't doing this before, adding a try/catch will only slow down performance and isn't recoverable anyways.

PR #5645, opened 2026-06-25
perf(core): Drop per-instance lock from SentryId and SpanId

runningcode:
## :scroll: Description

`SentryId` and `SpanId` stored their string value behind a `LazyEvaluator<String>`. Each `LazyEvaluator` allocates an `AutoClosableReentrantLock` (a `ReentrantLock` subclass, including its internal `Sync` object) plus a capturing lambda — on **every** instance.

This change replaces the `LazyEvaluator<String>` in both classes with a plain `volatile String` guarded by a double-checked `synchronized (this)` block:

- Eager (string-arg) constructors now assign the value directly — no lazy wrapper, no lock.
- The no-arg / `UUID` constructors still defer UUID-string generation; `SentryId` keeps the `UUID` reference so `toSentryIdString` stays lazy.
- Synchronization is retained (not a lock-free race) because UUID generation is non-idempotent — two racing threads must not produce different ids for the same instance.

No public API change (`apiDump` produces no diff).

One `SentryId` is created per event/transaction and one `SpanId` per span, so these are among the most frequently allocated objects in the SDK. Paying for a per-instance `ReentrantLock` + lambda to guard a single `String` is pure overhead, and the eager string-arg constructors got no laziness benefit at all. Follow-up to the SDK Overhead Reduction work (#5499).

Resolves JAVA-589.

Existing `SpanIdTest` and `SentryIdTest` cover the lazy-generation and normalization behavior (UUID generated/normalized only once, not on init, etc.). The full `:sentry` unit test suite passes (3372/3373; the one failing `SentryIdTest` case is a pre-existing test-isolation artifact that fails identically on `main` and is unrelated to this change).

### Benchmark

Measured against `main` using the real compiled classes on the same JVM (Temurin 17.0.16). Allocation is measured via `ThreadMXBean.getThreadAllocatedBytes` (exact TLAB accounting), with objects stored into a live array so the JIT can't scalar-replace them. Throughput is single-threaded, best-of-10 measured rounds after 5 warmup rounds of 1M ops each.

| Scenario | Bytes/op (before → after) | Δ bytes | Throughput ops/ms (before → after) | Speedup |
|---|---|---|---|---|
| `SpanId(String)` eager | 104 → **16** | **−88 (−85%)** | 93,580 → **458,935** | **4.9×** |
| `SpanId()` lazy + `toString` | 232 → **144** | **−88 (−38%)** | 41,587 → **65,556** | **1.6×** |
| `SentryId(String)` eager | 104 → **24** | **−80 (−77%)** | 138,296 → **312,029** | **2.3×** |
| `SentryId()` lazy + `toString` | 320 → **208** | **−112 (−35%)** | 26,472 → **34,084** | **1.3×** |
| `SentryId(UUID)` lazy + `toString` | 408 → **304** | **−104 (−25%)** | 4,430 → **4,579** | **1.03×** |

The ~88-byte reduction on eager `SpanId` accounts exactly for the removed object graph: `LazyEvaluator` (24B) + `AutoClosableReentrantLock` (16B) + its internal `NonfairSync`/AQS (32B) + the capturing lambda (16B). Eager constructors (hot deserialization/propagation paths) win biggest. The lazy paths still shed the lock machinery; `SentryId(UUID)` throughput barely moves because that path is dominated by `UUID.randomUUID()` (a `SecureRandom` draw), not the lock.

## :pencil: Checklist

## :crystal_ball: Next steps

Other `LazyEvaluator`/`AutoClosableReentrantLock` usages that guard genuinely light objects can be reviewed similarly as part of the SDK overhead reduction effort.

### DIFF
--- a/sentry/src/main/java/io/sentry/SpanId.java
+++ b/sentry/src/main/java/io/sentry/SpanId.java
@@ -2,24 +2,35 @@ package io.sentry;
 
 import static io.sentry.util.StringUtils.PROPER_NIL_UUID;
 
-import io.sentry.util.LazyEvaluator;
 import java.io.IOException;
 import java.util.Objects;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 public final class SpanId implements JsonSerializable {
   public static final SpanId EMPTY_ID =
       new SpanId(PROPER_NIL_UUID.replace("-", "").substring(0, 16));
 
-  private final @NotNull LazyEvaluator<String> lazyValue;
+  private volatile @Nullable String value;
 
   public SpanId(final @NotNull String value) {
-    Objects.requireNonNull(value, "value is required");
-    this.lazyValue = new LazyEvaluator<>(() -> value);
+    this.value = Objects.requireNonNull(value, "value is required");
   }
 
-  public SpanId() {
-    this.lazyValue = new LazyEvaluator<>(SentryUUID::generateSpanId);
+  public SpanId() {}
+
+  private @NotNull String getValue() {
+    String result = value;
+    if (result == null) {
+      synchronized (this) {
+        result = value;
+        if (result == null) {
+          result = SentryUUID.generateSpanId();
+          value = result;
+        }
+      }
+    }
+    return result;
   }
 
   @Override
@@ -27,17 +38,17 @@ public final class SpanId implements JsonSerializable {
     if (this == o) return true;
     if (o == null || getClass() != o.getClass()) return false;
     SpanId spanId = (SpanId) o;
-    return lazyValue.getValue().equals(spanId.lazyValue.getValue());
+    return getValue().equals(spanId.getValue());
   }
 
   @Override
   public int hashCode() {
-    return lazyValue.getValue().hashCode();
+    return getValue().hashCode();
   }
 
   @Override
   public String toString() {
-    return lazyValue.getValue();
+    return getValue();
   }
 
   // JsonElementSerializer
@@ -45,7 +56,7 @@ public final class SpanId implements JsonSerializable {
   @Override
   public void serialize(final @NotNull ObjectWriter writer, final @NotNull ILogger logger)
       throws IOException {
-    writer.value(lazyValue.getValue());
+    writer.value(getValue());
   }
 
   // JsonElementDeserializer
--- a/sentry/src/main/java/io/sentry/protocol/SentryId.java
+++ b/sentry/src/main/java/io/sentry/protocol/SentryId.java
@@ -6,7 +6,6 @@ import io.sentry.JsonSerializable;
 import io.sentry.ObjectReader;
 import io.sentry.ObjectWriter;
 import io.sentry.SentryUUID;
-import io.sentry.util.LazyEvaluator;
 import io.sentry.util.StringUtils;
 import io.sentry.util.UUIDStringUtils;
 import java.io.IOException;
@@ -19,19 +18,15 @@ public final class SentryId implements JsonSerializable {
   public static final SentryId EMPTY_ID =
       new SentryId(StringUtils.PROPER_NIL_UUID.replace("-", ""));
 
-  private final @NotNull LazyEvaluator<String> lazyStringValue;
+  private volatile @Nullable String value;
+  private final @Nullable UUID uuid;
 
   public SentryId() {
     this((UUID) null);
   }
 
   public SentryId(@Nullable UUID uuid) {
-    if (uuid != null) {
-      this.lazyStringValue =
-          new LazyEvaluator<>(() -> normalize(UUIDStringUtils.toSentryIdString(uuid)));
-    } else {
-      this.lazyStringValue = new LazyEvaluator<>(SentryUUID::generateSentryId);
-    }
+    this.uuid = uuid;
   }
 
   public SentryId(final @NotNull String sentryIdString) {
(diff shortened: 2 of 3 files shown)


sentry/src/main/java/io/sentry/protocol/SentryId.java
```diff
@@ -19,19 +18,15 @@ public final class SentryId implements JsonSerializable {
   public static final SentryId EMPTY_ID =
       new SentryId(StringUtils.PROPER_NIL_UUID.replace("-", ""));
 
-  private final @NotNull LazyEvaluator<String> lazyStringValue;
+  private volatile @Nullable String value;
```
runningcode, line 21: i don't like the name of this field `value` along with the `getValue()` method but to change the `getValue()` would break the API so probably better to keep it like this.

sentry/src/main/java/io/sentry/protocol/SentryId.java
```diff
@@ -42,29 +37,43 @@ public SentryId(final @NotNull String sentryIdString) {
...
-    } else {
-      this.lazyStringValue = new LazyEvaluator<>(() -> normalized);
+    this.uuid = null;
+    this.value = normalized.length() == 36 ? normalize(normalized) : normalized;
+  }
+
+  private @NotNull String getValue() {
+    String result = value;
+    if (result == null) {
+      synchronized (this) {
```
runningcode, line 47: using synchronized here still provides thread safety without allocating any extra objects.

sentry/src/main/java/io/sentry/protocol/SentryId.java
0xadam-brown: l: Still true, or was this comment from a previous version?

sentry/src/main/java/io/sentry/protocol/SentryId.java
```diff
@@ -42,29 +37,43 @@ public SentryId(final @NotNull String sentryIdString) {
               + "or 36 characters long (completed UUID). Received: "
               + sentryIdString);
     }
-    if (normalized.length() == 36) {
-      this.lazyStringValue = new LazyEvaluator<>(() -> normalize(normalized));
-    } else {
-      this.lazyStringValue = new LazyEvaluator<>(() -> normalized);
+    this.uuid = null;
+    this.value = normalized.length() == 36 ? normalize(normalized) : normalized;
```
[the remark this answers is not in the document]
0xadam-brown: (Looks like this is an existing issue; approving on that basis, but obviously feel free to update.)

0xadam-brown (approved):
Lgtm 👍

sentry/src/main/java/io/sentry/protocol/SentryId.java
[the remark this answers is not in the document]
runningcode: Yeah, this is fixed.

sentry/src/main/java/io/sentry/protocol/SentryId.java
runningcode: Yes still true.

sentry/src/main/java/io/sentry/SpanId.java
```diff
@@ -2,50 +2,61 @@
...
   }
 
-  public SpanId() {
-    this.lazyValue = new LazyEvaluator<>(SentryUUID::generateSpanId);
+  public SpanId() {}
+
+  private @NotNull String getValue() {
+    String result = value;
+    if (result == null) {
+      synchronized (this) {
```
markushi, line 25: I remember we intentionally replaced all synchronized calls a while ago to support [Java Virtual threads](https://github.com/getsentry/sentry-java/issues/3312), maybe @adinauer can shed some light on this?

markushi: Oh, looks like it works fine with Java 24+: https://mikemybytes.com/2025/04/09/java24-thread-pinning-revisited/#synchronized-revisited

runningcode: That is interesting. But is that only relevant if we are using virtual threads here?

romtsn: I thought we'd introduced it to also support re-entrance, but I might be wrong 🙈 @adinauer probably knows better

adinauer: `synchronized` in Java is re-entrant so there's no problem there.
We need to be careful where we replace the `AutoClosableReentrantLock` with `synchronized`.
This does look OK to me.

sentry/src/main/java/io/sentry/protocol/SentryId.java
adinauer: We replaced all `synchronized` blocks a while back with `AutoClosableReentrantLock` due to virtual threads being pinned. From what I understand today doing non blocking tasks inside `synchronized` should be fine but we need to watch out, not to replace it everywhere since that might re-introduce the thread pinning problem.

adinauer (approved):
The changes do look OK to me but I think your benchmarks are biased towards IDs that are actually being used. The point of the `LazyEvaluator` was to optimize cases where we don't actually end up using the IDs. I'm curious whether we even need the lazy part + lock anymore since we've moved away from `SecureRandom`. In case you take the time to benchmark, we might be able to conclude we can **reduce the complexity by just eagerly creating the String**.

Approving so you can decide on whether to merge as is or invest further time.

runningcode:
Thank you. Yeah the point of this PR is that the lock guarding the creation of the string was actually much heavier than the string itself in terms of object allocations.
I will merge as is and consider just eagerly creating the string.

PR #5658, opened 2026-06-29
refactor: do not start redundant UI event transaction when one is already on Scope

### LINKED ISSUE #5491: Do not start UI event transactions if there already is one on Scope
### Description

At the moment we start a UI event transaction and then check if there's already an existing one on Scope - if yes, we do not set the UI event transaction to the Scope. But this makes it redundant anyway, because if the UI event transaction is not on the Scope, it will not have any children, and we do not send such idle transactions without children.

We could improve it and not start the transaction at all if there's something on the Scope already.

mvanhorn:
## :scroll: Description

`SentryGestureListener.startTracing` always called `scopes.startTransaction(...)` for a UI interaction and only afterwards, in `applyScope`, declined to bind the new transaction to the Scope when one was already bound there. The unbound UI transaction then gathered no children and was dropped as an idle transaction without children.

This change reads the Scope's currently bound transaction before building the `TransactionOptions`. If a transaction is already bound, it logs a debug message and returns early without starting a new UI transaction. The existing handling for the listener's own `activeTransaction` (reschedule-finish and stop-previous) is unchanged, so only the externally-bound case short-circuits.

When a user-bound transaction is active, the SDK created and immediately discarded a UI transaction on every interaction. Not starting it avoids that wasted work. See `sentry-android-core/src/main/java/io/sentry/android/core/internal/gestures/SentryGestureListener.java`.

Fixes #5491

* resolves: JAVA-541

Added a unit test in `SentryGestureListenerTracingTest` asserting that `scopes.startTransaction` is never invoked and the bound transaction is left untouched when a transaction is already on the Scope. Could not run `./gradlew :sentry-android-core:testDebugUnitTest` locally because no Java runtime is available in this environment.

## :pencil: Checklist

## :crystal_ball: Next steps

AI was used for assistance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Fixes
 
+- Don't start a redundant UI interaction transaction when a transaction is already bound to the Scope ([#5491](https://github.com/getsentry/sentry-java/issues/5491))
+  - Previously, `SentryGestureListener` always started a UI transaction and only afterwards skipped binding it to the Scope when a manually-bound transaction already existed, leaving the new transaction to be dropped as an idle transaction without children.
 - Fix potential NPE within `Scope.endSession()` ([#5657](https://github.com/getsentry/sentry-java/pull/5657))
 
 ### Performance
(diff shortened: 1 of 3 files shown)


romtsn (approved):
@mvanhorn nice, this looks great, thank you for your contribution!

PR #5660, opened 2026-06-29
fix(android-fragment): support detach/attach navigation in fragment tracing

romtsn:
## Summary

Fixes a silent data quality issue where detach/attach tab navigation (and similar patterns) left `ui.load` spans open until the 30-second activity transaction deadline, producing badly inflated performance data.

### Root cause

`SentryFragmentLifecycleCallbacks` started spans in `onFragmentCreated` and stopped them in `onFragmentStarted` (a concession to ViewPager2 which locks background fragments to STARTED). For detach/attach navigation — manual tab switching (`transaction.detach(current).attach(next).commit()`), ViewPager v1 with `FragmentPagerAdapter`, and custom navigation frameworks that preserve fragment state — `onFragmentCreated` is **never called** for off-screen fragments. Spans would hang silently until the deadline.

### Fix

Three targeted changes, all leveraging the fact that `startTracing` / `stopTracing` are already idempotent:

| Hook | Change | Why |
|---|---|---|
| `onFragmentViewCreated` | +`startTracing` | Enables narrower `viewCreated → resumed` span for detach/attach paths. No-op if `onFragmentCreated` already started a span. |
| `onFragmentResumed` | +`stopTracing` | Ends the span on the detach/attach path where `onFragmentStarted` may not fire. No-op for the normal path where `onFragmentStarted` already closed the span. |
| `onFragmentViewDestroyed` | +`stopTracing` | Failsafe: closes any span for a fragment whose view is destroyed before reaching STARTED or RESUMED. |

The normal lifecycle path (`onFragmentCreated → onFragmentViewCreated → onFragmentStarted → onFragmentResumed`) is completely unaffected — still produces exactly one `ui.load` span, closed at `onFragmentStarted`.

## What's new for customers

Apps using detach/attach navigation will now see correct, narrower `ui.load` spans (`viewCreated → resumed`) instead of timeout-inflated spans.
## Tests

Five new unit tests covering:
- Detach/attach path: `onFragmentViewCreated` starts a span when tracing enabled
- Normal path: `onFragmentViewCreated` is a no-op when span already running
- Detach/attach path: `onFragmentResumed` stops the span
- Normal path: `onFragmentResumed` does not double-finish after `onFragmentStarted`
- Failsafe: `onFragmentViewDestroyed` stops a span that never reached STARTED/RESUMED

Example trace: https://sentry-sdks.sentry.io/explore/traces/trace/edb7de2bba454d70a4c13eee5bce660c/

As you can see TabFragmentB now correctly shows up after re-attaching (but wouldn't have without this fix):
Google Chrome 2026-06-30 17 25 15

## Not verified

`./gradlew spotlessApply apiDump` could not run (no JDK in sandbox) — CI will apply formatting and regenerate `.api` files if needed.


sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
```diff
@@ -93,17 +93,35 @@ public class SentryFragmentLifecycleCallbacks(
     savedInstanceState: Bundle?,
   ) {
     addBreadcrumb(fragment, FragmentLifecycleState.VIEW_CREATED)
+
+    // For detach/attach navigation (e.g. manual tab switching, ViewPager v1 with
+    // FragmentPagerAdapter, custom navigation frameworks), onFragmentCreated is never called for
+    // off-screen fragments that are re-attached. Starting here enables a narrower
+    // "view created -> resumed" span for those paths. startTracing is idempotent, so for the
+    // normal onFragmentCreated -> onFragmentViewCreated path this is a no-op.
```
0xadam-brown, line 94: nit: Feel free to prune Jr's comments if you think they're getting too verbose.  (Up to you, of course.)

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
```diff
@@ -93,17 +93,35 @@ public class SentryFragmentLifecycleCallbacks(
...
+    if (fragment.isAdded) {
```
0xadam-brown: l: We could internalize the `if (fragment.isAdded) {...` checks inside `startTracing()` since they protect the integrity of `ui.load` initiation + we always apply them.

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/DetachAttachTabsActivity.kt
```diff
@@ -0,0 +1,50 @@
+package io.sentry.samples.android
+
+import android.os.Bundle
+import android.view.View
+import android.widget.TextView
+import androidx.appcompat.app.AppCompatActivity
+import androidx.fragment.app.Fragment
+import androidx.fragment.app.commit
+
+class DetachAttachTabsActivity : AppCompatActivity(R.layout.activity_detach_attach_tabs) {
```
0xadam-brown, line 10: Thanks for the sample 💯 

m: Unfortunately I'm not able to see or click the tabs on my emulator (Pixel 10, API 36.1):

Image

...seems they're being rendered under the action bar.

0xadam-brown (commented):
Nice! A few quick comments; otherwise lgtm 👍

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
runningcode: would it also make sense to add the `scopes.options.isEnableScreenTracking` inside there too?

runningcode (approved):
Fragments are so confusing! Thanks for the fix.

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
romtsn: I think in this case they are on point, so I'd keep them!

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
romtsn: encapsulated this into `startTracing` in https://github.com/getsentry/sentry-java/pull/5660/changes/5d6efb487557f885b2141b387ba8b49d99635ae5

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/DetachAttachTabsActivity.kt
romtsn: Used a NoActionBar theme for this activity in https://github.com/getsentry/sentry-java/pull/5660/changes/5d6efb487557f885b2141b387ba8b49d99635ae5 the buttons should be visible now
image

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
```diff
@@ -153,14 +163,19 @@ public class SentryFragmentLifecycleCallbacks(
...
 
   private fun startTracing(fragment: Fragment) {
-    if (!isPerformanceEnabled || isRunningSpan(fragment)) {
+    if (!fragment.isAdded || !isPerformanceEnabled || isRunningSpan(fragment)) {
       return
     }
 
+    val fragmentName = getFragmentName(fragment)
+
+    if (scopes.options.isEnableScreenTracking) {
```
0xadam-brown, line 172: h: Oops - looks like we'll lose screen tracking in cases where performance isn't enabled. We could put this check first to fix:

```kotlin
 if (fragment.isAdded) { // Or invert and return early.
      if (scopes.options.isEnableScreenTracking) {
        scopes.configureScope { it.screen = getFragmentName(fragment) }
      }
     // ...rest of the startTracing() method
}
```

romtsn: oh true, I'm surprised none of the bots (or tests!) caught this, seems pretty obvious 🙈

romtsn: fixed in https://github.com/getsentry/sentry-java/pull/5660/changes/aaeb61090e2dc29eea76210ef77ae69d2abdef06

sentry-android-fragment/src/test/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacksTest.kt
```diff
@@ -251,6 +257,115 @@ class SentryFragmentLifecycleCallbacksTest {
...
+      fixture.fragment,
+      view = mock(),
+      savedInstanceState = null,
+    )
+
+    verify(fixture.scope).screen = "androidx.fragment.app.Fragment"
+  }
+
+  @Test
+  fun `When performance is disabled, it should still update screen name`() {
```
0xadam-brown, line 295: 🏅

sentry-android-fragment/src/main/java/io/sentry/android/fragment/SentryFragmentLifecycleCallbacks.kt
0xadam-brown: >  I'm surprised none of the bots (or tests!) caught this, seems pretty obvious 🙈

Story of my life 😅

0xadam-brown (approved):
Thanks for the updates – lgtm 💯

PR #5670, opened 2026-07-01
perf: Avoid per-transaction Timer thread in SentryTracer (JAVA-596)

### LINKED ISSUE #5663: ANR: per-transaction `java.util.Timer` in `SentryTracer` spawns a thread on the calling thread
### Integration

sentry-android

### Build System

Gradle

### AGP Version

8.43.1

### Proguard

Enabled

Firebase Crashlytics

### Version

34.14.0

### Steps to Reproduce

Setup — minimal app with user-interaction tracing on:

SentryAndroid.init(this) { options ->
    options.dsn = "<your-dsn>"
    options.tracesSampleRate = 1.0
    options.isEnableUserInteractionTracing = true   // the path under test
}

Add any clickable/scrollable view with a resolvable view id (e.g. a Button
with android:id, or a RecyclerView) so SentryGestureListener resolves a target
and starts a transaction on the gesture.

A) Observe the thread-per-transaction behavior (any device):
1. Set a breakpoint in the java.util.Timer constructor, or in
   SentryTracer.<init> at the `new Timer(true)` line.
2. Tap the button N times (or scroll the list, pausing >idleTimeout between
   gestures so each interaction opens a fresh transaction).
3. Observe the breakpoint hits once per interaction → one Timer constructor →
   one Thread.start() per transaction.
   Equivalent without a debugger: dump threads before and after a burst of
   taps (adb shell kill -3 <pid>, or Thread.getAllStackTraces().size) — the
   thread count rises with the number of in-flight transactions, and you'll
   see transient "Timer-N" threads being created and torn down.

B) Surface the main-thread blocking (reproduces the ANR signature):
1. Run on a low-end / heavily loaded device (or an emulator throttled with
   few cores + memory pressure; `adb shell am start ...` while a stress app
   runs in the background). Thread creation cost is what we're amplifying.
2. Rapidly tap/scroll to fire many user-interaction transactions in quick
   succession. Each one does Thread.start() (nativeCreate) on the main thread.
3. Under enough memory/CPU pressure, one of those native thread spawns blocks
   on the runtime lock long enough to trip the ANR watchdog — matching the
   reported stack (SentryGestureListener.onUp → startTransaction →
   SentryTracer.<init> → new Timer → Thread.nativeCreate →
   art::ConditionVariable::WaitHoldingLocks).

### Expected Result

Creating Traces should be performant and have minimal allocation. Tracing performance of application should not consume runtime resource aggressively. 

### Actual Result

We're seeing production ANRs where the main thread is blocked inside the SentryTracer constructor, specifically creating the per-transaction java.util.Timer. The Timer constructor starts a dedicated thread, and that thread spawn (Thread.nativeCreate) is what blocks. It happens on every user-interaction transaction (tap/scroll/swipe), so it fires on the main thread on a hot path.

This is the same underlying concern as #2130 (closed as not planned), but we now have custom ANR with 1 second duration report showing it has a real, production impact, and we think there's a cleaner fix than the shared-Timer idea that was originally proposed.

### ANR stack trace

The main thread is blocked here while handling a touch event:

```
at nativeCreate (Thread.java)
at start (Thread.java:1433)
at <init> (Timer.java:197)
at <init> (Timer.java:166)
at <init> (SentryTracer.java:102)
at createTransaction (DefaultSpanFactory.java:15)
at createTransaction (Scopes.java:987)
at startTransaction (Scopes.java:917)
at startTransaction (Sentry.java:1224)
at startTransaction (ScopesAdapter.java:300)
at startTracing (SentryGestureListener.java:265)
at onUp (SentryGestureListener.java:84)
at handleTouchEvent (SentryWindowCallback.java:77)
at dispatchTouchEvent (SentryWindowCallback.java:58)
... (DecorView / ViewRootImpl input dispatch)
```

The native frame underneath shows the thread being parked on a runtime lock during thread creation:

```
at art::ConditionVariable::WaitHoldingLocks
```

- **~100% on low-end devices.**
- **Concentrated in emerging markets**
- **Always on a gesture.**
- The absolute user count is modest for us because we collect that custom shorten ANR data only from Beta.

### Why this is more than the object cost

The `Timer` object itself is cheap (a small `TaskQueue` heap + the thread wrapper). The expensive part is that **every `Timer` constructor calls `Thread.start()`**, and we do that once per transaction. `java.util.Timer` is explicitly designed for one thread to service thousands of scheduled tasks — so creating one per transaction is the opposite of how it's meant to be used.

Each `SentryTracer` only ever schedules two tasks (idle timeout + deadline timeout) and already tracks/cancels them individually (`idleTimeoutTask`, `deadlineTimeoutTask`, `cancelIdleTimer()`, `cancelDeadlineTimer()`). The only thing that requires a per-instance `Timer` today is the `timer.cancel()` in `finish()`. So the structure is already 90% of the way to using a shared scheduler.

### Proposed fix (different from #2130)

#2130 suggested reusing a single `Timer`. The likely downside there: cancelled `TimerTask`s stay in the timer's heap until their scheduled fire time, so under high transaction volume the queue grows until `purge()` runs — a memory wart.

Instead, the SDK **already has the right primitive**: `ISentryExecutorService` / `SentryExecutorService` — a single daemon-thread `ScheduledThreadPoolExecutor(1)`, created lazily off the main thread, reachable via `options.getExecutorService()`, and already used for other delayed work (with a `prewarm()` for queue pre-allocation). `SentryTracer` is the last place still on raw `java.util.Timer`.

Concretely:

1. In `SentryExecutorService`, call `setRemoveOnCancelPolicy(true)` on the `ScheduledThreadPoolExecutor`. This evicts cancelled tasks immediately, which sidesteps the heap-growth problem that made the shared-`Timer` approach unattractive — and benefits every other caller too.
2. In `SentryTracer`, replace the `Timer`/`TimerTask` pair with `Future<?>` handles from `executorService.schedule(runnable, delayMillis)`:
   - keep the existing cancel-previous-then-reschedule logic, using `future.cancel(false)`;
   - in `finish()`, cancel only the tracer's own two futures (drop `timer.cancel()`, since the executor is owned by `SentryOptions`);
   - guard scheduling against `RejectedExecutionException` for the shutdown window.

runningcode:
Fixes JAVA-596
Closes #5663

## :scroll: Description

`SentryTracer` created a `java.util.Timer` (`new Timer(true)`) for every transaction configured with an idle or deadline timeout. `new Timer(...)` spawns a dedicated thread **synchronously on the calling thread** — often the main thread on Android — and each such transaction got its own throwaway thread.

This change schedules the idle/deadline timeouts on a **dedicated, shared** `ISentryExecutorService` instead:

- New `SentryOptions.getTimerExecutorService()` — a single `SentryExecutorService` created in `activate()`, kept **separate from the main executor** so timeout callbacks (which finish transactions) don't contend with cached-event sending.
- `SentryTracer` schedules cancellable `Future`s (`idleTimeoutFuture` / `deadlineTimeoutFuture`) on that executor. It does **not** cache a reference to the executor: it tracks a `timersEnabled` flag and looks the executor up from the options each time it schedules, so it always uses the instance currently held by the options (e.g. the fresh one installed after an SDK restart) rather than a stale reference. On finish only the futures are cancelled.
- Lifecycle: the timer executor is closed only on a **full** SDK shutdown (`Scopes.close` when not restarting). On restart it is intentionally left running so that pending idle/deadline timeouts of transactions started before the restart still fire and finish those transactions; the abandoned instance self-terminates once idle (`allowCoreThreadTimeOut`, 30s keep-alive).

At scale (screen loads, HTTP spans, user-interaction transactions) the per-transaction `Timer` was the dominant source of SDK thread churn, and the thread creation happened on the caller's thread. This is item #4 ("Note B") of the thread/executor audit (JAVA-570 / SDK-1347) and is the biggest thread-count win in that effort. Using a dedicated executor (rather than the main one) keeps transaction-finishing work off the executor that sends cached events.

- Updated `SentryTracerTest` idle/deadline coverage to the new `Future`-based scheduling; the "timer executor shut down" cases now close the dedicated executor to exercise the immediate-finish fallback.
- Added `ScopesTest` coverage for the executor lifecycle: a full close shuts the timer executor down, while a restart leaves it running so in-flight timeouts still fire.
- `ActivityLifecycleIntegrationTest` idle-timeout test now installs a real `timerExecutorService`.
- Full `:sentry:test` and `:sentry-android-core:testDebugUnitTest` pass.

## :pencil: Checklist

## :crystal_ball: Next steps

Follow-up audit items can fold the remaining per-`Timer` sites (`DefaultCompositePerformanceCollector`, `RateLimiter`, `LifecycleWatcher`) onto the same dedicated scheduler in a separate cleanup pass.

### DIFF
--- a/sentry/src/main/java/io/sentry/Sentry.java
+++ b/sentry/src/main/java/io/sentry/Sentry.java
@@ -354,7 +354,9 @@ public final class Sentry {
         }
 
         if (options.getTimerExecutorService().isClosed()) {
-          options.setTimerExecutorService(new SentryExecutorService(options, true));
+          options.setTimerExecutorService(
+              new SentryExecutorService(
+                  options, true, SentryExecutorService.TIMER_KEEP_ALIVE_SECONDS, TimeUnit.SECONDS));
         }
 
         // load lazy fields of the options in a separate thread
--- a/sentry/src/main/java/io/sentry/Sentry.java
+++ b/sentry/src/main/java/io/sentry/Sentry.java
@@ -353,6 +353,10 @@ public final class Sentry {
           options.setExecutorService(new SentryExecutorService(options));
         }
 
+        if (options.getTimerExecutorService().isClosed()) {
+          options.setTimerExecutorService(new SentryExecutorService(options, true));
+        }
+
         // load lazy fields of the options in a separate thread
         try {
           options.getExecutorService().submit(() -> options.loadLazyFields());
--- a/sentry/src/main/java/io/sentry/SentryExecutorService.java
+++ b/sentry/src/main/java/io/sentry/SentryExecutorService.java
@@ -22,6 +22,12 @@ public final class SentryExecutorService implements ISentryExecutorService {
    */
   private static final int MAX_QUEUE_SIZE = 271;
 
+  /**
+   * How long the timer executor's worker thread stays alive while idle before self-terminating, so
+   * an instance abandoned on SDK restart doesn't leak a live thread once its queue drains.
+   */
+  static final long TIMER_KEEP_ALIVE_SECONDS = 30;
+
   private final @NotNull ScheduledThreadPoolExecutor executorService;
   private final @NotNull AutoClosableReentrantLock lock = new AutoClosableReentrantLock();
 
@@ -39,14 +45,16 @@ public final class SentryExecutorService implements ISentryExecutorService {
     this(new ScheduledThreadPoolExecutor(1, new SentryExecutorServiceThreadFactory()), options);
   }
 
-  SentryExecutorService(final @Nullable SentryOptions options, final boolean removeOnCancelPolicy) {
+  SentryExecutorService(
+      final @Nullable SentryOptions options,
+      final boolean removeOnCancelPolicy,
+      final long keepAliveTime,
+      final @NotNull TimeUnit keepAliveTimeUnit) {
     this(options);
     // removes cancelled tasks from the work queue immediately instead of leaving them until their
     // scheduled time; useful for executors that frequently reschedule (e.g. transaction timeouts)
     executorService.setRemoveOnCancelPolicy(removeOnCancelPolicy);
-    // let the worker thread die when idle so an executor abandoned on SDK restart (its pending
-    // timeouts still fire) doesn't leak a live thread once its queue drains
-    executorService.setKeepAliveTime(10, TimeUnit.SECONDS);
+    executorService.setKeepAliveTime(keepAliveTime, keepAliveTimeUnit);
     executorService.allowCoreThreadTimeOut(true);
   }
 
--- a/sentry/src/main/java/io/sentry/SentryExecutorService.java
+++ b/sentry/src/main/java/io/sentry/SentryExecutorService.java
@@ -44,6 +44,10 @@ public final class SentryExecutorService implements ISentryExecutorService {
     // removes cancelled tasks from the work queue immediately instead of leaving them until their
     // scheduled time; useful for executors that frequently reschedule (e.g. transaction timeouts)
     executorService.setRemoveOnCancelPolicy(removeOnCancelPolicy);
+    // let the worker thread die when idle so an executor abandoned on SDK restart (its pending
+    // timeouts still fire) doesn't leak a live thread once its queue drains
+    executorService.setKeepAliveTime(10, TimeUnit.SECONDS);
+    executorService.allowCoreThreadTimeOut(true);
   }
 
   public SentryExecutorService() {
--- a/sentry/src/main/java/io/sentry/SentryExecutorService.java
+++ b/sentry/src/main/java/io/sentry/SentryExecutorService.java
@@ -39,6 +39,13 @@ public final class SentryExecutorService implements ISentryExecutorService {
     this(new ScheduledThreadPoolExecutor(1, new SentryExecutorServiceThreadFactory()), options);
   }
 
+  SentryExecutorService(final @Nullable SentryOptions options, final boolean removeOnCancelPolicy) {
+    this(options);
+    // removes cancelled tasks from the work queue immediately instead of leaving them until their
+    // scheduled time; useful for executors that frequently reschedule (e.g. transaction timeouts)
+    executorService.setRemoveOnCancelPolicy(removeOnCancelPolicy);
+  }
+
   public SentryExecutorService() {
     this(new ScheduledThreadPoolExecutor(1, new SentryExecutorServiceThreadFactory()), null);
   }
--- a/sentry/src/main/java/io/sentry/SentryOptions.java
+++ b/sentry/src/main/java/io/sentry/SentryOptions.java
@@ -318,6 +318,14 @@ public class SentryOptions {
   private @NotNull ISentryExecutorService executorService = NoOpSentryExecutorService.getInstance();
 
   /**
+   * Dedicated executor for scheduling transaction idle/deadline timeouts. Kept separate from {@link
+   * #executorService} so timeout callbacks (which finish transactions) don't contend with cached
+   * event sending.
+   */
+  private @NotNull ISentryExecutorService timerExecutorService =
+      NoOpSentryExecutorService.getInstance();
+
+  /**
    * Whether SpotlightIntegration has already been loaded via reflection. This prevents re-adding it
    * if the user removed it in their configuration callback and activate() is called again.
    */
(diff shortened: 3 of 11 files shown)


sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -1568,6 +1583,30 @@ public void setExecutorService(final @NotNull ISentryExecutorService executorSer
...
+  /**
+   * Sets the dedicated executor used to schedule transaction idle/deadline timeouts.
+   *
+   * @param timerExecutorService the timer executor service
+   */
+  @ApiStatus.Internal
+  @TestOnly
+  public void setTimerExecutorService(final @NotNull ISentryExecutorService timerExecutorService) {
+    if (timerExecutorService != null) {
+      this.timerExecutorService = timerExecutorService;
```
markushi, line 1611: `nit:` I'm wondering if we should close the old/stale one, or at least mention in the docs that the existing one needs to be closed manually.

sentry/src/main/java/io/sentry/Sentry.java
```diff
@@ -354,6 +354,10 @@ private static void init(final @NotNull SentryOptions options, final boolean glo
           options.getExecutorService().prewarm();
         }
 
+        if (options.getTimerExecutorService().isClosed()) {
```
markushi, line 356: shouldn't this be 
```suggestion
        if (options.getTimerExecutorService() instanceof NoOpSentryExecutorService) {
```
instead?

sentry/src/main/java/io/sentry/SentryExecutorService.java
```diff
@@ -49,6 +49,17 @@ public SentryExecutorService(final @Nullable SentryOptions options) {
...
   }
 
+  SentryExecutorService(final @Nullable SentryOptions options, final boolean removeOnCancelPolicy) {
+    this(options);
+    // removes cancelled tasks from the work queue immediately instead of leaving them until their
+    // scheduled time; useful for executors that frequently reschedule (e.g. transaction timeouts)
+    executorService.setRemoveOnCancelPolicy(removeOnCancelPolicy);
+    // let the worker thread die when idle so an executor abandoned on SDK restart (its pending
+    // timeouts still fire) doesn't leak a live thread once its queue drains
+    executorService.setKeepAliveTime(10, TimeUnit.SECONDS);
```
markushi: Any rationale why it's 10 seconds? My gut tells me a higher value (e.g. 60 seconds) could make more sense to avoid frequent restarts, but on the other hand I don't have a good way to back this argument 😅

markushi (approved):
Nice one, looks great to me. Can't wait until we get rid of all timer tasks.

sentry/src/main/java/io/sentry/SentryExecutorService.java
romtsn: `l`: should we also pass these as ctor arguments? Because currently it's kind of implicit that if you use this ctor with `removeOnCancelPolicy` then these two will be set automatically for you. Or we could also just javadoc it

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -37,10 +36,13 @@ public final class SentryTracer implements ITransaction {
...
-  private volatile @Nullable TimerTask idleTimeoutTask;
-  private volatile @Nullable TimerTask deadlineTimeoutTask;
+  private volatile @Nullable Future<?> idleTimeoutFuture;
+  private volatile @Nullable Future<?> deadlineTimeoutFuture;
 
-  private volatile @Nullable Timer timer = null;
+  // Shared executor used to schedule the timeout tasks. Null once the tracer is finished, at which
+  // point no more timeouts may be scheduled. It is never shut down here since it is shared
+  // SDK-wide.
+  private volatile @Nullable ISentryExecutorService timerExecutorService = null;
```
romtsn: `m`: is there a reason to keep a ref to the service? Or could we just always retrieve it from `options`?

romtsn (approved):
couple of minor things but LGTM otherwise! great improvement

sentry/src/main/java/io/sentry/Sentry.java
runningcode: Good point! but on line 332 we call options.activate which checks `instanceof NoOpSentryExecutorService`

sentry/src/main/java/io/sentry/SentryOptions.java
runningcode: This is `@TestOnly` and is the same as the method above on line 1582 so I don't think we need this but I'm open to discussion!

sentry/src/main/java/io/sentry/SentryTracer.java
runningcode: this is a good point. this was serving two purposes.
* a reference to the timerExecutorService
* a guard against scheduling more timouts

the first one can be replaced with `options` ill replace the second with a boolean for clarity.

sentry/src/main/java/io/sentry/SentryExecutorService.java
runningcode: I dont have any good heuristice here either. I will adjust to 30 seconds to be in the middle of our proposals.

runningcode: good point. will add as ctor args.

PR #5681, opened 2026-07-02
perf(core): Remove executor prewarm

runningcode:
## Summary

`prewarm()` scheduled 40 dummy tasks far in the future, cancelled them, and purged the queue, in order to pre-grow the single-threaded `ScheduledThreadPoolExecutor`'s work queue during `Sentry.init`. On-device measurement shows the cost it avoids is negligible-to-zero, so this removes it (and the now-unused constants) along with its two init call sites.

## Why it's safe to remove

prewarm's only job was to pre-grow the executor's work queue so a later resize wouldn't happen "in an unexpected area of the SDK". Measured on a Galaxy A55 (Android 16), growing the queue the full 16→54 (all three array resizes) costs **~8µs total** — and that is the ceiling: init only ever queues a handful of tasks, well under the default 16-slot capacity, so in practice **no resize happens and the saving is 0µs**.

`PrewarmBenchmarkTest`, n=1000 × 3 runs, median µs. Both paths schedule 40 tasks into an empty queue (identical per-insert work); they differ only in whether the backing array resizes:

| | Run 1 | Run 2 | Run 3 |
|---|---|---|---|
| grow 16→54 | 15 | 15 | 15 |
| preGrown 54 | 7 | 6 | 7 |
| **resize cost (delta)** | **~8** | **~9** | **~8** |

## It also can't help, and slightly regresses init

The executor is single-threaded and prewarm is submitted *ahead* of the first useful task (`loadLazyFields`, submitted right after in `Sentry.init`), so the 40-task loop can only delay that task, never speed it up. Thread creation and executor class-loading are paid by whichever task runs first, so `loadLazyFields` warms them identically with or without prewarm.

Measured end-to-end, the time until the first useful task finishes (from a fresh executor, n=400 × 3 runs) is **~100µs slower** with prewarm than without, at every percentile; a plain `submit(noop)` is indistinguishable from doing nothing:

| median µs | Run 1 | Run 2 | Run 3 |
|---|---|---|---|
| prewarm | 567 | 524 | 537 |
| no warming | 460 | 428 | 435 |

(`PrewarmBenchmarkTest` also records the per-op loop cost and the process-once class-loading cost, which is unavoidable regardless of prewarm; both confirm the above.)

## Changes

- Remove `prewarm()` from `ISentryExecutorService`, `SentryExecutorService` (and the now-unused `INITIAL_QUEUE_SIZE` / `dummyRunnable`), and `NoOpSentryExecutorService`.
- Remove the two init call sites (`Sentry.init` re-init branch and `SentryOptions.activate()`).
- Drop the prewarm unit test and mock overrides; add `PrewarmBenchmarkTest`.

Original prewarm PR: #4606 — its only stated reason was the queue-growth cost measured above.

### DIFF
--- a/sentry/api/sentry.api
+++ b/sentry/api/sentry.api
@@ -1134,7 +1134,6 @@ public abstract interface class io/sentry/ISentryClient {
 public abstract interface class io/sentry/ISentryExecutorService {
 	public abstract fun close (J)V
 	public abstract fun isClosed ()Z
-	public abstract fun prewarm ()V
 	public abstract fun schedule (Ljava/lang/Runnable;J)Ljava/util/concurrent/Future;
 	public abstract fun submit (Ljava/lang/Runnable;)Ljava/util/concurrent/Future;
 	public abstract fun submit (Ljava/util/concurrent/Callable;)Ljava/util/concurrent/Future;
@@ -1880,7 +1879,6 @@ public final class io/sentry/NoOpSentryExecutorService : io/sentry/ISentryExecut
 	public fun close (J)V
 	public static fun getInstance ()Lio/sentry/ISentryExecutorService;
 	public fun isClosed ()Z
-	public fun prewarm ()V
 	public fun schedule (Ljava/lang/Runnable;J)Ljava/util/concurrent/Future;
 	public fun submit (Ljava/lang/Runnable;)Ljava/util/concurrent/Future;
 	public fun submit (Ljava/util/concurrent/Callable;)Ljava/util/concurrent/Future;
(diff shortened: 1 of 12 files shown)


sentry/api/sentry.api
```diff
@@ -1133,7 +1133,6 @@ public abstract interface class io/sentry/ISentryClient {
 public abstract interface class io/sentry/ISentryExecutorService {
 	public abstract fun close (J)V
 	public abstract fun isClosed ()Z
-	public abstract fun prewarm ()V
```
0xadam-brown, line 1137: (Fine with this as these are marked `@ApiStatus.Internal`.)

0xadam-brown (approved):
Agreed that avoiding prewarming on the init path is a good move unless there's a compelling reason to do it. 

@romtsn: I see we introduced prewarming in https://github.com/getsentry/sentry-java/pull/4606 with a number of other perf fixes. Any concerns about removing it now?

romtsn:
@0xadam-brown yeah I have some concerns, there's a thread going on about this internally: https://sentry.slack.com/archives/C089VFRB8N9/p1782999439636989

romtsn (approved):

PR #5684, opened 2026-07-02
fix(android): Name the device-info caching thread

runningcode:
## :scroll: Description

`DefaultAndroidEventProcessor` pre-caches device info on a background thread using `Executors.newSingleThreadExecutor()` without a `ThreadFactory`. The JDK's default factory names such threads `pool-N-thread-M`, which is not identifiable as a Sentry thread.

This was the only thread-creation site in the SDK's production source that did not name its thread. This change supplies a `ThreadFactory` that names the (daemon) thread `SentryDeviceInfoCache`, matching the convention already used by `HostnameCache`, `SentryExecutorService`, and `AsyncHttpTransport`.

Named threads let customers identify which threads belong to the Sentry SDK when inspecting their apps (profilers, thread dumps, ANR traces). Every other thread the SDK spawns is already named; this closes the one remaining gap.

Existing `DefaultAndroidEventProcessor` unit tests continue to pass. The change only affects the name/daemon flag of the executor's thread, not its behavior.

## :pencil: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Name the device-info caching thread `SentryDeviceInfoCache` so all threads spawned by the SDK are identifiable ([#5684](https://github.com/getsentry/sentry-java/pull/5684))
+
 ## 8.47.0
 
 ### Behavioral Changes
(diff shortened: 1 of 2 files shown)


0xadam-brown (approved):
👍

PR #5733, opened 2026-07-07
fix(anr): extend main thread detection from thread dump

### LINKED ISSUE #5731: ANRv2 (AppExitInfo): main thread not detected when the OS thread dump names it after the process → ANR reported with no stack trace
### Integration

sentry-android

### Build System

Gradle

### AGP Version

N/A (observed server-side from ingested events)

### Proguard

Enabled

No

### Version

8.29.0 (logic unchanged on current main)

### Steps to Reproduce

1. Android app using the default ANRv2 path (`AnrV2Integration`, API 30+ / `ApplicationExitInfo`).
2. An ANR occurs on a device/ROM where `ApplicationExitInfo#getTraceInputStream()` returns a **native-format** thread dump. In that format, thread headers carry the kernel `comm` name (max 15 chars), so the **main thread is named after the process** (e.g. `"com.example.app"`), not `"main"`. Every thread is parsed via `BEGIN_UNMANAGED_NATIVE_THREAD_RE` (no `state`, all frames `platform:native`).

   Example header the parser receives (synthetic):

   ```
   "com.example.app" sysTid=1039
     native: #00 pc 0x... /apex/com.android.runtime/lib64/bionic/libc.so (__epoll_pwait)
     ...
     native: #NN pc 0x... /system/bin/app_process64 (main)
   ```

3. The SDK parses the dump and builds the ANR event.

### Expected Result

The `ApplicationNotResponding` exception is backfilled with the main thread's stack trace (the documented ANRv2 behavior), regardless of what the OS names the main thread. The main-thread stack is clearly identifiable, it roots at `__libc_init → main (app_process64) → AndroidRuntime::start`.

### Actual Result

The main thread is never identified, so the ANR is reported **with no stack trace on the exception**, even though the real main-thread stack is present in `event.threads`.

Root cause: main-thread detection is a strict name match:

- `ThreadDumpParser` marks `main` only if `threadName.equals("main")`:
  https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/internal/threaddump/ThreadDumpParser.java#L210-L217
- `ApplicationExitInfoEventProcessor.findMainThread` also matches only `name.equals("main")`:
  https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java#L764-L775
- With a process-named main thread, `findMainThread` returns `null`, so the exception is built from a dummy thread with an empty stacktrace:
  https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java#L801-L808

Secondary symptom (same code path): for these native dumps, a thread whose section has no parseable frames still gets a stacktrace with `frames: []` — `parseStacktrace` always attaches a stacktrace and there is no `isEmpty` guard (contrast `SentryThreadFactory` / `SentryExceptionFactory`, which skip empty stacks). Ingestion then rejects it, since `stacktrace.frames` must be non-empty, producing an `invalid_data` / "expected a non-empty value" **processing error** on the event and `threads.values.N.stacktrace.frames` being nulled.

### Suggested fix directions
1. Detect the main thread robustly for `AppExitInfo` dumps instead of an exact `"main"` match — e.g. the thread whose stack roots at `app_process` / `AndroidRuntime::start`, or the thread whose tid equals the process id, or fall back to the process-named thread.
2. Do not attach a stacktrace with empty `frames` to a `SentryThread` (mirror the existing `!frames.isEmpty()` guards), so ingestion does not reject the event with a processing error.

markushi:
## :scroll: Description

Not every main thread has a `name="main"` set, so we ended up not correctly identifying it (on some Android versions/OEMs, native threads, etc...). 

To solve this we now rely on a common Linux/Android convention: The main thread kernel id (`sysTid`) is the same as the process id (`pid`). For managed Java threads, a fallback for the "main" thread name remains.

**Changes**

- Parses the process id from the thread dump header (`----- pid <pid> at ... -----`) and marks the thread whose `sysTid` equals it as the main thread.
- The main thread's name is normalized back to `"main"` so downstream consumers see a consistent name even when the OS renamed it.
- Threads without a stacktrace are skipped, as they are not actionable.

No public API change.

When the OS names the main thread after the process, ANRs were not attributed to the main thread and were reported without a stack trace. Matching on `sysTid == pid` is robust and does not depend on the (truncatable) process name.

Fixes JAVA-635
Fixes https://github.com/getsentry/sentry-java/issues/5731

- Added tests to `ThreadDumpParserTest`
- Updated existing tests

## :pencil: Checklist

## :crystal_ball: Next steps

--

[View Junior Session in Sentry](https://sentry.sentry.io/explore/conversations/slack%3AC089VFRB8N9%3A1783415141.214909/?project=4510944073809921)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ### Fixes
 
 - Fix main thread identification parsing for ApplicationExitInfo ANRs ([#5733](https://github.com/getsentry/sentry-java/pull/5733))
+- Do not send threads without stacktraces for ApplicationExitInfo ANRs ([#5733](https://github.com/getsentry/sentry-java/pull/5733))
 - Record byte-level client reports when event processors discard logs or trace metrics ([#5718](https://github.com/getsentry/sentry-java/pull/5718))
 - Name the device-info caching thread `SentryDeviceInfoCache` so all threads spawned by the SDK are identifiable ([#5684](https://github.com/getsentry/sentry-java/pull/5684))
 - Apply byte-category rate limits to log and trace metric envelope items ([#5716](https://github.com/getsentry/sentry-java/pull/5716))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ### Fixes
 
+- Detect the main thread in AnrV2 thread dumps when the OS renames it to the process name, by matching the thread whose `sysTid` equals the process id ([#5733](https://github.com/getsentry/sentry-java/pull/5733))
 - Record byte-level client reports when event processors discard logs or trace metrics ([#5718](https://github.com/getsentry/sentry-java/pull/5718))
 - Name the device-info caching thread `SentryDeviceInfoCache` so all threads spawned by the SDK are identifiable ([#5684](https://github.com/getsentry/sentry-java/pull/5684))
 - Apply byte-category rate limits to log and trace metric envelope items ([#5716](https://github.com/getsentry/sentry-java/pull/5716))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 
 ### Fixes
 
-- Detect the main thread in AnrV2 thread dumps when the OS renames it to the process name, by matching the thread whose `sysTid` equals the process id ([#5733](https://github.com/getsentry/sentry-java/pull/5733))
+- Fix main thread identification parsing for ApplicationExitInfo ANRs ([#5733](https://github.com/getsentry/sentry-java/pull/5733))
 - Record byte-level client reports when event processors discard logs or trace metrics ([#5718](https://github.com/getsentry/sentry-java/pull/5718))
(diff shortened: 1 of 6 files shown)


romtsn (approved):
this looks good! shall we also address the second part which is not to attach a thread if its stacktrace has no frames? Could be addressed in a separate PR ofc

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 
 ### Fixes
 
+- Fix main thread identification parsing for ApplicationExitInfo ANRs ([#5733](https://github.com/getsentry/sentry-java/pull/5733))
```
romtsn, line 12:
```suggestion
- Fix main thread identification parsing for ApplicationExitInfo ANRs ([#5733](https://github.com/getsentry/sentry-java/pull/5733))
- Do not send threads without stacktraces for ApplicationExitInfo ANRs ([#5733](https://github.com/getsentry/sentry-java/pull/5733))
```

romtsn (approved):

PR #5737, opened 2026-07-07
fix(core): Prevent recursion when a callback triggers another capture

runningcode:
## 📜 Description
Adds a core-level re-entrancy guard so a user callback (`beforeSend`, `beforeBreadcrumb`, `beforeSendLog`, and the other `beforeSend*` variants) that triggers another SDK capture no longer recurses into a `StackOverflowError`.

## 💡 Motivation and Context
These callbacks run synchronously on the caller thread with only a `try/catch` — nothing prevents re-entrant capture. So a callback that captures again — directly, or transitively through a logging integration that routes back into Sentry — loops forever:

`captureX()` → `executeBeforeX()` → callback logs/captures → (integration) → `captureX()` → …

This is the same class of bug as the logcat `StackOverflowError` (SDK-CRASHES-JAVA-3T3H, fixed per-integration in #5734). It also affects the **Timber** integration (`SentryTimberTree` fires `captureEvent` + `addBreadcrumb` + `logger().log()` on every Timber call) and any app code that captures from within a callback. Rather than patch each integration, this fixes the whole class once in the core.

**Solution**
`SentryCallbackReentrancyGuard` is a shared thread-local flag set **only** around each callback's `execute(...)`. The capture entry points (`SentryClient.captureEvent`/`captureTransaction`/`captureLog`, `Scope.addBreadcrumb`) drop a nested capture while the flag is active, breaking the loop across all capture types on the same thread. Captures made by event processors run outside the flag and are unaffected. The real `Log.*` output in integrations is untouched.

Dropping the nested capture (rather than sending it without running `beforeSend`) is deliberate — bypassing `beforeSend` would risk sending unscrubbed PII, which the SDK already guards against.

## 💥 Behavioral change
A capture a callback makes **intentionally** (e.g. a diagnostic event, or re-capturing an enriched event) is now dropped while that callback runs, and logged at `DEBUG`. This is the safe trade-off — such a nested capture is indistinguishable from recursion. Event processors are not affected. The per-integration guard added in #5734 becomes redundant (left in place; harmless).

## 💚 How did you test it?
- `SentryClientTest`: `beforeSend`→`captureEvent`, `beforeSendLog`→log, and cross-type (`beforeSend` emitting a log) — each verified to `StackOverflowError`/fail without the guard and pass with it, with the nested capture dropped.
- `ScopeTest`: `beforeBreadcrumb`→`addBreadcrumb`.
- `SentryTimberIntegrationTest`: end-to-end with a real `Sentry.init`, a planted `SentryTimberTree`, and a `beforeSend` that calls `Timber.e(...)` — verified to recurse without the guard and to run the callback exactly once with it.

Fixes SDK-CRASHES-JAVA-3T3H

JAVA-638

### DIFF
--- a/sentry/src/main/java/io/sentry/util/SentryCallbackReentrancyGuard.java
+++ b/sentry/src/main/java/io/sentry/util/SentryCallbackReentrancyGuard.java
@@ -1,6 +1,7 @@
 package io.sentry.util;
 
 import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.Nullable;
 
 /**
  * Thread-local re-entrancy guard that marks whether a user-supplied {@code before*} callback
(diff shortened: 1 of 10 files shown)


sentry/src/main/java/io/sentry/util/SentryCallbackReentrancyGuard.java
```diff
@@ -0,0 +1,54 @@
...
+  private SentryCallbackReentrancyGuard() {}
+
+  /** Whether a user callback is currently executing on this thread. */
+  public static boolean isActive() {
+    final @Nullable Integer current = depth.get();
+    return current != null && current > 0;
+  }
+
+  /** Marks that a user callback is starting to execute on this thread. */
+  public static void enter() {
```
adinauer: `m` we could return a `ISentryLifecycleToken` that implements `AutoCloseable` here so we can use it in try-with-resources statements and not forget to call `exit` in a finally block correctly.

adinauer (approved):
LGTM, we could consider also handling `EventProcessor`, `onOversizedEvent`, `captureEnvelope` and scope callbacks on things like `Scopes.captureException`.

sentry/src/main/java/io/sentry/util/SentryCallbackReentrancyGuard.java
runningcode: great idea!

Issue #5741, opened 2026-07-08
java.lang.NoSuchMethodError: java.lang.Math.floorDiv(JI)J


sysmat:
### Integration

sentry

### Java Version

8

### Other Error Monitoring Solution

No

### Version

8.47.0

### Steps to Reproduce

```
java.lang.NoSuchMethodError: java.lang.Math.floorDiv(JI)J
java.lang.NoSuchMethodError: java.lang.Math.floorDiv(JI)J
	at io.sentry.vendor.SentryIso8601Utils.epochDayToYearMonthDay(SentryIso8601Utils.java:293)
	at io.sentry.vendor.SentryIso8601Utils.formatTimestamp(SentryIso8601Utils.java:148)
	at io.sentry.DateUtils.getTimestampFromMillis(DateUtils.java:76)
	at io.sentry.DateUtils.getTimestamp(DateUtils.java:66)
	at io.sentry.clientreport.ClientReport.serialize(ClientReport.java:62)
	at io.sentry.JsonObjectSerializer.serialize(JsonObjectSerializer.java:56)
	at io.sentry.JsonObjectWriter.value(JsonObjectWriter.java:108)
	at io.sentry.JsonSerializer.serializeToString(JsonSerializer.java:284)
	at io.sentry.JsonSerializer.serialize(JsonSerializer.java:215)
	at io.sentry.SentryEnvelopeItem.lambda$fromClientReport$21(SentryEnvelopeItem.java:457)
	at io.sentry.SentryEnvelopeItem$CachedItem.getBytes(SentryEnvelopeItem.java:626)
	at io.sentry.SentryEnvelopeItem.lambda$fromClientReport$23(SentryEnvelopeItem.java:471)
	at io.sentry.SentryEnvelopeItem.getData(SentryEnvelopeItem.java:72)
	at io.sentry.SentryEnvelopeItem.getClientReport(SentryEnvelopeItem.java:480)
	at io.sentry.clientreport.ClientReportRecorder.recordLostEnvelopeItem(ClientReportRecorder.java:85)
	at io.sentry.clientreport.ClientReportRecorder.recordLostEnvelope(ClientReportRecorder.java:67)
	at io.sentry.transport.AsyncHttpTransport$EnvelopeSender.flush(AsyncHttpTransport.java:318)
	at io.sentry.transport.AsyncHttpTransport$EnvelopeSender.run(AsyncHttpTransport.java:247)
[...]

PR #5745, opened 2026-07-08
build: Add Java 8 API compatibility checks

adinauer:
## :scroll: Description
Add Java 8 runtime API compatibility coverage to JVM SDK modules using Animal Sniffer and the Codehaus Java 8 signature.

This applies `ru.vyarus.animalsniffer` to Java 8 `java-library` SDK modules that were missing it and adds:

```text
org.codehaus.mojo.signature:java18:1.0@signature
```

to those modules, with the signature version defined in `gradle/libs.versions.toml` as `java18Signature`. Modules that already used Animal Sniffer keep their existing gummy-bears signature and also get the Codehaus Java 8 signature. The plugin checks signatures independently, so the Java 8 check still catches plain JVM incompatibilities while the existing Android/desugaring coverage remains intact.

We recently hit a Java 8 runtime incompatibility from a method reference that compiled on a newer JDK but is unavailable on Java 8. Targeting Java 8 bytecode is not sufficient to catch this class of issue, so published JVM SDK modules should also be checked against the Java 8 runtime API surface.

This now runs through the existing `check` lifecycle and therefore through the existing Build workflow on PRs and `main`.

Refs GH-5741

```text
./gradlew spotlessApply apiDump
BUILD SUCCESSFUL
```

```text
./gradlew check --no-configuration-cache -Dorg.gradle.configureondemand=false
BUILD SUCCESSFUL
```

The `check` run includes Android modules and passed, confirming the duplicated JVM Animal Sniffer configuration does not break Android builds.

A stacked validation PR intentionally introduces newer JDK API references in multiple JVM modules to confirm the new checks fail where expected.

## :pencil: Checklist

#skip-changelog


adinauer:
Tested this using https://github.com/getsentry/sentry-java/pull/5746 and it failed for both `Math.floorDiv` and `Objects.requireNonNullElse` in different modules.

sentry-apache-http-client-5/build.gradle.kts
```diff
@@ -25,6 +26,8 @@ dependencies {
   errorprone(libs.nopen.checker)
   errorprone(libs.nullaway)
 
+  signature("org.codehaus.mojo.signature:java18:${libs.versions.java18Signature.get()}@signature")
```
runningcode: why not use the version catalog for the dependency as well (instead of just the version) ?

sentry-apache-http-client-5/build.gradle.kts
```diff
@@ -33,6 +36,8 @@ dependencies {
   testImplementation(libs.mockito.kotlin)
 }
 
+tasks { check { dependsOn(animalsnifferMain) } }
```
runningcode: Check the build scan to make sure this doesn't eagerly resolve the task.

This would definitely not eagerly resolve the task:
```suggestion
tasks.check.configureEach { dependsOn(animalsnifferMain) }
```

adinauer: After reworking this PR into a plugin, it should no longer be an issue.

build-logic/src/main/kotlin/io/sentry/gradle/SentryAnimalSnifferPlugin.kt
```diff
@@ -0,0 +1,61 @@
...
+        SentryAnimalSnifferExtension::class.java,
+      )
+
+    project.addSignatureDependency("java18-signature")
+
+    project.tasks.matching { it.name == "check" }.configureEach {
+      dependsOn("animalsnifferMain")
+    }
+
+    project.afterEvaluate {
```
runningcode: why do we need the `afterEvaluate` here? in general `afterEvaluate` is bad for performance and build consistency and should be avoided.

build-logic/src/main/kotlin/io/sentry/gradle/SentryAnimalSnifferPlugin.kt
```diff
@@ -0,0 +1,61 @@
...
+      project.tasks.named("animalsnifferMain", SourceTask::class.java).configure {
+        exclude(extension.mainExcludes.get())
+        val ignoredClasses = extension.ignoredClasses.get()
+        if (ignoredClasses.isNotEmpty()) {
+          javaClass.getMethod("setIgnoreClasses", Iterable::class.java).invoke(this, ignoredClasses)
```
runningcode: why do we need to use reflection here?

adinauer: updated

gradle/libs.versions.toml
```diff
@@ -11,6 +12,7 @@ coroutines = "1.6.1"
 espresso = "3.7.0"
 feign = "11.6"
 gummyBears = "0.12.0"
+java18Signature = "1.0"
```
runningcode: nit: is this java 1.8 or java 18? it is hard to tell. we should add a comment or some other way to distinguish

build-logic/src/main/kotlin/io/sentry/gradle/SentryAnimalSnifferPlugin.kt
```diff
@@ -1,44 +1,43 @@
...
-import org.gradle.api.model.ObjectFactory
-import org.gradle.api.provider.ListProperty
 import ru.vyarus.gradle.plugin.animalsniffer.AnimalSniffer
 
-abstract class SentryAnimalSnifferExtension @Inject constructor(objects: ObjectFactory) {
-  val ignoredClasses: ListProperty<String> = objects.listProperty(String::class.java)
-  val mainExcludes: ListProperty<String> = objects.listProperty(String::class.java)
+open class SentryAnimalSnifferExtension(private val project: Project) {
+  fun ignoreClasses(vararg classes: String) {
+    project.tasks.named("animalsnifferMain", AnimalSniffer::class.java).configure {
```
runningcode: Extension classes aren't supposed to have build logic. They are only supposed to hold configuration properties. What you had before this commit was perfect.

What was the error you were seeing when you removed `afterEvaluate`?

I'm trying to understand why the logic is like this.

runningcode (approved):
I left some comments and then did the refactor myself. Go back to the previous commit if you are curious. There's just one nit left but I'm giving it a 👍

gradle/libs.versions.toml
adinauer: Renamed to java8 to avoid confusion.

build-logic/src/main/kotlin/io/sentry/gradle/SentryAnimalSnifferPlugin.kt
adinauer: I don't think there was an error, clanker seems to have made unnecessary changes, thanks for fixing!

PR #5755, opened 2026-07-11
fix: drop app-start transaction when deadline timer fires doze-delayed

### LINKED ISSUE #5752: Incorrect app start transaction duration when app goes in sleep
As reported by a customer we're seeing the following behavior:

1. App is launched
2. A cold start is detected and an `ui.load` transaction is created (`app.start.cold` + `ttid`)
3. The app is backgrounded before the first frame draws (and the activity only progresses through the `onCreate` + `onStart` phases, no `onDestroy`) — so TTID never happens, and no lifecycle cleanup on our side runs.
4. The device enters Doze, which freezes our deadline timer so it doesn't fire at +30s.
5. Hours later the device wakes up, the deadline finally fires when the app is resumed, and our `forceFinish` stamps the spans with the current time.
6. The resulting transactions spans over multiple hours.

I would argue that we should drop any foreground launches which have a doze in-between, as it just skews the real "user perceived" launches.

On the other hand, a real user was probably waiting for a few seconds for the app to come alive, but it didn't, so the user bailed.

Related issue: getsentry/sentry-java#5530

mvanhorn:
## 📜 Description

`SentryTracer` now records the wall-clock time when the deadline timeout is scheduled and, when the timer fires far later than scheduled (more than 2x the configured `deadlineTimeout`), drops the transaction instead of force-finishing it with `now()`. The drop follows the existing `dropIfNoChildren` pattern in `finish()`: DEBUG log, clear the scope's transaction, cancel timers, no `captureTransaction`. Timers that fire within the slack behave exactly as before (forceFinish with DEADLINE_EXCEEDED).

## 💡 Motivation and Context

When an Android app cold-starts and is backgrounded before first frame, the `ui.load` transaction's 30s deadline timer (a `java.util.Timer`) is frozen during Doze and fires only hours later when the device wakes. `forceFinish` then stamps the root span and all children with the current time, producing app-start transactions spanning hours and skewing user-perceived launch metrics. "Timer fired far later than scheduled" is the exact failure signature, so the check stays in platform-neutral core with no Android API and no public API change.

Resolves getsentry/sentry-java#5752

## 💚 How did you test it?

New unit tests in `SentryTracerTest`: a deadline timeout that fires within the slack still force-finishes with DEADLINE_EXCEEDED; a simulated late fire (scheduled timestamp shifted beyond 2x deadline) drops the transaction, clears it from the scope, and captures nothing. Ran the sentry module test suite.

## :pencil: Checklist

- [X] Public API changes reviewed by another Mobile SDK team member or implemented according to the [develop docs](<https://develop.sentry.dev/>) spec.

## 🔮 Next steps


mvanhorn:
Fixed the failing checks in 3bc4794: the root span is now finished when the deadline timer fires too late (so the dropped transaction reads as finished), the two logger-mock tests enable options.isDebug so DiagnosticLogger actually forwards to the mock, and the test file is spotless-formatted. Verified locally on JDK 17: spotlessCheck plus the full SentryTracerTest suite pass.

mvanhorn:
Brought the branch up to date with main (c134c2e7). It was 69 commits behind with conflicts in CHANGELOG.md and SentryTracer.java, both resolved, so it's mergeable again.

The fix itself is unchanged from 3bc4794.

runningcode:
Hi @mvanhorn sorry for the delay, we've just prioritized a project looking at how we measure time and set deadline timers in the SDK. It will be a big refactor and we will reconsider this PR if the refactor doesn't fix this issue.

PR #5756, opened 2026-07-13
fix(core): Prevent lost wakeups in batch processors

### LINKED ISSUE #5739: Lost-wakeup race in the Logs/Metrics batch processor can strand queued events past the flush interval
### Description

The `add -> maybeSchedule` fast path reads the `hasScheduled` flag outside `scheduleLock`, while the running flush resets that flag inside `scheduleLock` after a separate `queue.isEmpty()` check. This is a check-then-act TOCTOU: an event offered by a producer thread in the window after the flush reads `isEmpty() == true` but before it writes `hasScheduled = false` is neither observed by the flush (not rescheduled) nor able to schedule itself (the fast path sees `hasScheduled` still `true`). The event is left queued with `hasScheduled == false` and no pending flush. Affects both `LoggerBatchProcessor` and `MetricsBatchProcessor` (byte-identical scheduling logic).

### Root cause

* `sentry/src/main/java/io/sentry/logger/LoggerBatchProcessor.java`: `maybeSchedule` reads `hasScheduled` outside the lock; `flush()` re-checks `isEmpty()` then sets `hasScheduled = false` inside the lock.
* `sentry/src/main/java/io/sentry/metrics/MetricsBatchProcessor.java`: same structure.

The stranded event is only sent on a subsequent `add()` or an explicit `flush()`/`close()`. On Android this is partly masked because `AndroidLoggerBatchProcessor`/`AndroidMetricsBatchProcessor` force a flush on `onBackground()`.

*Note: reproduced as a deterministic interleaving of the state machine (same volatile flag,* `ReentrantLock`*,* `ConcurrentLinkedQueue`*), not a live Android run.*

### Suggested fix

Make the flag reset and queue re-check safe against the lock-free producer path: after draining, set `hasScheduled = false` first and then re-check `queue.isEmpty()` under the lock (rescheduling if non-empty), or remove the lock-free fast path so `maybeSchedule` always consults the queue under `scheduleLock`.

adinauer:
## 📜 Description

Use an atomic scheduling state for the logs and metrics batch processors. A producer now claims delayed flush scheduling with compare-and-set, while a completed flush releases that claim before checking whether more events arrived.

This removes the lock and `Future` lifecycle checks that allowed an event to be queued without a pending flush. Regression tests verify that both processors schedule another flush after the previous flush has run.

## 💡 Motivation and Context

An event added while a flush was transitioning from an empty queue to resetting its scheduling flag could remain queued indefinitely. The running flush did not observe it, and the producer saw scheduling as already claimed.

Fixes getsentry/sentry-java#5739

## 💚 How did you test it?

* `./gradlew spotlessApply apiDump`
* `./gradlew :sentry:test --tests 'io.sentry.logger.LoggerBatchProcessorTest' --tests 'io.sentry.metrics.MetricsBatchProcessorTest'`

## :pencil: Checklist

- [X] Public API changes reviewed by another Mobile SDK team member or implemented according to the [develop docs](<https://develop.sentry.dev/>) spec.

## 🔮 Next steps

None.

### DIFF
--- a/sentry/src/main/java/io/sentry/logger/LoggerBatchProcessor.java
+++ b/sentry/src/main/java/io/sentry/logger/LoggerBatchProcessor.java
@@ -4,7 +4,6 @@ import com.jakewharton.nopen.annotation.Open;
 import io.sentry.DataCategory;
 import io.sentry.ISentryClient;
 import io.sentry.ISentryExecutorService;
-import io.sentry.ISentryLifecycleToken;
 import io.sentry.SentryExecutorService;
 import io.sentry.SentryLevel;
 import io.sentry.SentryLogEvent;
@@ -12,15 +11,14 @@ import io.sentry.SentryLogEvents;
 import io.sentry.SentryOptions;
 import io.sentry.clientreport.DiscardReason;
 import io.sentry.transport.ReusableCountLatch;
-import io.sentry.util.AutoClosableReentrantLock;
 import io.sentry.util.JsonSerializationUtils;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.Future;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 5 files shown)


markushi (approved):
LGTM!

sentry/src/main/java/io/sentry/logger/LoggerBatchProcessor.java
```diff
@@ -97,33 +93,28 @@ public void close(final boolean isRestarting) {
...
-        } catch (RejectedExecutionException e) {
-          hasScheduled = false;
-          options
-              .getLogger()
-              .log(SentryLevel.WARNING, "Logs batch processor flush task rejected", e);
-        }
-      }
+    final int flushAfterMs = immediately ? 0 : FLUSH_AFTER_MS;
+    try {
+      executorService.schedule(new BatchRunnable(), flushAfterMs);
```
runningcode, line 108: is `hasScheduled` supposed to be reset to false at some point?

runningcode (commented):
MetricsBatchProcessor and LoggerBatchProcessor share a lot of complicated scheduling logic, is there a way we can share it?

PR #5765, opened 2026-07-15
fix(profiling): Disable transaction-based profiling when enableLegacyProfiling is false

markushi:
#### Summary

Broadens `enableLegacyProfiling` so that setting it to `false` also disables **transaction-based** profiling (`profilesSampleRate` / `profilesSampler`), not just continuous profiling.

#### Why

Transaction-based profiling always relies on the legacy `Debug`-based profiler — it is *not* supported by the ProfilingManager (Perfetto) backend. Previously `enableLegacyProfiling` only controlled continuous profiling, which led to a confusing combination:

> If someone sets `profilesSampleRate` and `enableLegacyProfiling=false`, we would still run the old legacy profiler, as it's in transaction-based profiling mode.

Per discussion, it's sensible that `enableLegacyProfiling` disables transaction-based profiling too.

#### Changes

- `AndroidOptionsInitializer.setupProfiler`: when the transaction-based path is selected but `enableLegacyProfiling` is `false`, no-op both profilers, close any app-start profiler, and log a warning guiding users towards `profileSessionSampleRate` (continuous profiling).
- Updated `SentryOptions.isEnableLegacyProfiling` / `setEnableLegacyProfiling` javadoc.
- Added unit tests.
- Docs updated in a matching `sentry-docs` change (Perfetto profiling / `enableLegacyProfiling` option).

#skip-changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,8 +34,6 @@
 - Session Replay: Fix error-to-replay linkage in `buffer` mode ([#5754](https://github.com/getsentry/sentry-java/pull/5754))
 - Prevent logs and metrics from remaining queued after a flush scheduling race ([#5756](https://github.com/getsentry/sentry-java/pull/5756))
 - Fix main thread identification for tombstone (native crash) events ([#5742](https://github.com/getsentry/sentry-java/pull/5742))
-- Setting `enableLegacyProfiling` to `false` now also disables transaction-based profiling (`profilesSampleRate`/`profilesSampler`) ([#5765](https://github.com/getsentry/sentry-java/pull/5765))
-  - Transaction-based profiling always relies on the legacy profiler and is not supported by the ProfilingManager (Perfetto) backend. Use `profileSessionSampleRate` for continuous profiling instead.
 
 ### Dependencies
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -34,6 +34,8 @@
 - Session Replay: Fix error-to-replay linkage in `buffer` mode ([#5754](https://github.com/getsentry/sentry-java/pull/5754))
 - Prevent logs and metrics from remaining queued after a flush scheduling race ([#5756](https://github.com/getsentry/sentry-java/pull/5756))
 - Fix main thread identification for tombstone (native crash) events ([#5742](https://github.com/getsentry/sentry-java/pull/5742))
+- Setting `enableLegacyProfiling` to `false` now also disables transaction-based profiling (`profilesSampleRate`/`profilesSampler`) ([#5765](https://github.com/getsentry/sentry-java/pull/5765))
```
romtsn: not sure if we need this one since it targets the integration branch (which is not yet merged)?

romtsn (approved):
one changelog nit, LGTM otherwise!

CHANGELOG.md
markushi: oh yeah true, let me merge this then and update the integration branch instead

Issue #5769, opened 2026-07-16
Add `scope.SetEnvironment`


bitsandfoxes:
# Proposal: Allow setting `Environment` on the Scope

Unblocks
- https://github.com/getsentry/sentry-dotnet/issues/5387
- https://github.com/getsentry/sentry-unity/pull/2764

## Summary

Add `getEnvironment()` / `setEnvironment(String)` to sentry-java's `IScope`, so other SDKs (hybrid SDKs like .NET, Godot, Unity, Unreal, React Native?) and end users can set the `Environment` on the scope at runtime. This lifts the restriction of having to provide it strictly at init via `SentryOptions.setEnvironment(...)`.

This brings `sentry-java` to parity with `sentry-cocoa` and `sentry-native`. These already expose a scope-/runtime-level environment setter.

## Motivation

Today `environment` in sentry-java lives only on `SentryOptions`, fixed at init time. Any hybrid SDKs layered on top either manage their own `environment` or provide a way for users to modify it at runtime. This creates the  need to sync it into the Java layer at runtime. That way events, and persisted (background/ANR/NDK) events, carry the correct value. 

Both reference SDKs already support this:

| SDK    | API                                                          | Level        | Getter? |
|--------|--------------------------------------------------------------|--------------|---------|
| cocoa  | `- (void)setEnvironment:(NSString *)environment` on `Scope`  | scope        | no      |
| native | `sentry_set_environment(const char *environment)` (+ `_n`)   | global scope | no      |
| java   | **(proposed)** `IScope.setEnvironment(String)` + `getEnvironment()` | scope  | yes     |

## Proposed API (sentry-java)

```java
public interface IScope {
  // ...
  @Nullable String getEnvironment();

  /**
   * Sets the Scope's environment. Takes precedence over SentryOptions#getEnvironment()
   * when applied to events. Mainly used by hybrid SDKs to sync their environment.
   */
  void setEnvironment(@Nullable String environment);
}
```

Reached by users/hybrid SDKs the same way as other scope mutations:

```java
Sentry.configureScope(scope -> scope.setEnvironment("staging"));
```

Unlike cocoa, we could also add a **getter**, consistent with other Java scope fields like `getTransactionName`, and `getUser` e.t.c. and required by the combined-scope view.

### API References

**sentry-cocoa**
[...]

PR #5772, opened 2026-07-16
feat: added `Get/Set Environment` on `Scope`

bitsandfoxes:
## :scroll: Description
Adds a getter and setter to the scope that allows users to configure the `Environment` at runtime.

See https://github.com/getsentry/sentry-java/issues/5769

## :pencil: Checklist


sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
```diff
@@ -196,6 +196,7 @@ private void backfillScope(
     setRequest(event);
     setUser(event);
     setScopeTags(event);
+    setScopeEnvironment(event);
     setBreadcrumbs(event);
     setExtras(event);
     setContexts(event);
```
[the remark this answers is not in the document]
bitsandfoxes: Historical events intentionally skip all persisted scope data, not only `Environment`.

sentry/src/main/java/io/sentry/Sentry.java
```diff
@@ -579,6 +579,10 @@ private static void notifyOptionsObservers(final @NotNull SentryOptions options)
                     options.findPersistingScopeObserver();
                 if (scopeCache != null) {
                   scopeCache.resetCache();
+                  final @Nullable String environment = globalScope.getEnvironment();
+                  if (environment != null) {
+                    scopeCache.setEnvironment(environment);
+                  }
```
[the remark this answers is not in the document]
bitsandfoxes: The intended lifecycle of this is enriching of historical exists. So this is not intended to to be able to seed the next session.

sentry/src/main/java/io/sentry/Sentry.java
```diff
@@ -579,6 +579,10 @@ private static void notifyOptionsObservers(final @NotNull SentryOptions options)
                     options.findPersistingScopeObserver();
                 if (scopeCache != null) {
                   scopeCache.resetCache();
+                  final @Nullable String environment = globalScope.getEnvironment();
+                  if (environment != null) {
+                    scopeCache.setEnvironment(environment);
```
romtsn, line 604: `l`: can we maybe add a comment describing why we do this for environment now? Something like
```
// scope environment is configuration, not telemetry state: re-persist it after the reset like release/dist above
```

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@
 - Prevent concurrent PixelCopy access during Session Replay masking and bitmap cleanup ([#5808](https://github.com/getsentry/sentry-java/pull/5808))
 - Release `MediaMuxer` when the replay video encoder fails to start to avoid a resource leak ([#5607](https://github.com/getsentry/sentry-java/pull/5607))
 
+### Features
+
+- Added `Environment` to the `Scope`, allowing you to overwrite the `Environment` set on the options ([#5772](https://github.com/getsentry/sentry-java/pull/5772))
+
```
romtsn, line 159:
```suggestion
### Features

- Added `Environment` to the `Scope`, allowing you to overwrite the `Environment` set on the options ([#5772](https://github.com/getsentry/sentry-java/pull/5772))

### Improvements

- Skip building Android manifest metadata debug log messages when debug logging is disabled, reducing allocations during SDK init ([#5790](https://github.com/getsentry/sentry-java/pull/5790))

### Fixes

- Prevent concurrent PixelCopy access during Session Replay masking and bitmap cleanup ([#5808](https://github.com/getsentry/sentry-java/pull/5808))
- Release `MediaMuxer` when the replay video encoder fails to start to avoid a resource leak ([#5607](https://github.com/getsentry/sentry-java/pull/5607))

```

romtsn: I think i'd reorder it to the top

romtsn (commented):
overall looks good to me, nice job!

There's one place where I think we need to update the usage of `environment` too, and that is Baggage/DSC (Baggage.java:523, :560). That's the envelope header relay uses for dynamic-sampling rules and trace-level environment in the UI, so we should get it aligned with the rest of telemetry I believe?

markushi:
*(ai summary)*

Thanks for the review! Merged `main` in and pushed.

**Baggage/DSC (`Baggage.java:523`, `:560`)** — done. `setValuesFromScope` prefers the scope environment; `setValuesFromTransaction` takes it as a new parameter.

Note that while testing this I found the first version was still wrong: `scopes.configureScope(cb)` gives the default write scope (isolation), not the combined view, so an environment on the **global** scope — what `Sentry.init` persists — never reached the DSC. All three call sites now use `ScopeType.COMBINED`, same as event capture. Regression tests added.

**Also fixed the same gap in logs, metrics and profiles:**
- `LoggerApi`/`MetricsApi` set `sentry.environment` from the combined scope, so a log's attribute can't disagree with its own DSC.
- Profiles get it in `SentryClient.captureProfileChunk`/`captureTransaction` rather than in the profilers, which build their objects on a profiler thread where the current scope isn't the configured one. This covers the Android and JVM profiler in one place.
- `Scopes.captureProfileChunk` passed `getScope()` (current only) and had the same defect; now passes the combined view.

**`Sentry.java` comment** — added. **Changelog** — the duplicated sections had landed in the released 8.51.0 block; removed, one entry now under `Unreleased` → `Features`.

Two things left deliberately unchanged, happy to go either way:

- **`CombinedScopeView.startSession()`** reads the combined environment while the session lives on the default write scope. `CombinedScopeView` only has two read rules — scalars take the first match current → isolation → global, collections merge — so reading "owner, then global" would give `environment` a third rule for one call site. The `user` asymmetry inside `new Session(...)` is real but points the other way: `Scope.startSession()` has no combined view to consult.
- **`Scope.clear()`** notifies observers for `level`/`environment`, and `PersistingScopeObserver` is shared by all three scopes, so clearing the isolation scope drops a value the global scope still holds. This came from the earlier round (`879e15a`), so I didn't reverse it silently.

PR #5779, opened 2026-07-16
build: Bump AGP to 9.2.1 and migrate Android modules

### LINKED ISSUE #5777: Bump AGP to 9.2.0 in sentry-java

runningcode:
We need to bump AGP to 9.2 because compile SDK introduced a breaking change (`37.0` vs `37`) which makes it very difficult to compile using AGP 8.X.

AGP 9.X introduced several breaking changes that we try to mitigate in this PR.

* AGP 9.X sets the java target compile level to 11 from 8, so we override this to keep the same level.
* AGP 9.X includes a bundles kotlin plugin, we disable this with `android.builtInKotlin=false`
* AGP 9.X has a new DSL which we disable with `android.builtInKotlin=false`
* AGP 9.X disables unit tests for debug variants. we therefore set the test variant to release in modules where debug is disabled.
* Fix and ignore some AGP 9.X lint warnings

Fixes getsentry/sentry-java#5777<br>Linear: getsentry/sentry-java#5777 (getsentry/sentry-java#5777)

#skip-changelog

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)

[https://claude.ai/code/session_01EmE8hdaj9H9K61opK2PZ6U](<https://claude.ai/code/session_01EmE8hdaj9H9K61opK2PZ6U>)

### DIFF
--- a/.github/workflows/agp-matrix.yml
+++ b/.github/workflows/agp-matrix.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        agp: [ '8.7.0','8.8.0','8.9.0' ]
+        agp: [ '9.0.0', '9.1.1', '9.2.1' ]
         integrations: [ true, false ]
 
     name: AGP Matrix Release - AGP ${{ matrix.agp }} - Integrations ${{ matrix.integrations }}
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -113,6 +113,20 @@ allprojects {
 subprojects {
     apply { plugin("io.sentry.spotless") }
 
+    // AGP 9.2 bundles lint 9.2.1, which flags compileSdk 36 as outdated because 37 is available.
+    // We intentionally stay on compileSdk 36 until the API 37 bump (#5768), so silence that check
+    // for every Android module (library and application).
+    pluginManager.withPlugin("com.android.library") {
+        extensions.configure<com.android.build.gradle.BaseExtension> {
+            lintOptions { disable("GradleDependency") }
+        }
+    }
+    pluginManager.withPlugin("com.android.application") {
+        extensions.configure<com.android.build.gradle.BaseExtension> {
+            lintOptions { disable("GradleDependency") }
+        }
+    }
+
     plugins.withId(Config.QualityPlugins.detektPlugin) {
         configure<DetektExtension> {
             buildUponDefaultConfig = true
@@ -165,6 +179,18 @@ subprojects {
 
                 sourceCompatibility = JavaVersion.VERSION_1_8
                 targetCompatibility = JavaVersion.VERSION_1_8
+            }
+        }
+
+        // AGP 9 defaults Android modules to Java 11. Pin the published library modules back
+        // to Java 8 so their bytecode stays consumable by Java 8 projects, mirroring the
+        // java-library pin above.
+        plugins.withId("com.android.library") {
+            configure<com.android.build.gradle.BaseExtension> {
+                compileOptions {
+                    sourceCompatibility = JavaVersion.VERSION_1_8
+                    targetCompatibility = JavaVersion.VERSION_1_8
+                }
             }
         }
 
--- a/sentry-android-core/build.gradle.kts
+++ b/sentry-android-core/build.gradle.kts
@@ -1,5 +1,6 @@
 import net.ltgt.gradle.errorprone.errorprone
 import org.jetbrains.kotlin.config.KotlinCompilerVersion
+import org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_1_8
 
 plugins {
   id("com.android.library")
@@ -33,7 +34,11 @@ android {
     getByName("release") { consumerProguardFiles("proguard-rules.pro") }
   }
 
-  kotlin { compilerOptions.jvmTarget = org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_1_8 }
+  // AGP 9 only generates unit tests for the testBuildType. CI disables the debug
+  // variant, so unit tests must target release to run at all.
+  testBuildType = "release"
+
+  kotlin { compilerOptions.jvmTarget = JVM_1_8 }
 
   testOptions {
     animationsDisabled = true
--- a/sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
@@ -91,7 +91,7 @@ public final class SentryFrameMetricsCollector implements Application.ActivityLi
   }
 
   @SuppressWarnings("deprecation")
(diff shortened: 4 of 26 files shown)


.github/workflows/agp-matrix.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        agp: [ '8.7.0','8.8.0','8.9.0' ]
+        agp: [ '9.0.0', '9.1.1', '9.2.1' ]
```
runningcode, line 20: reminder: we publish the entry-android artifact **compiled against 9.2.1** as set in `Config.kt` below. this test **compiles the sentry-android artifact against different versions** of AGP meaning the tests **do not represent our published artifact**

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
```diff
@@ -91,7 +91,7 @@ public SentryFrameMetricsCollector(
   }
 
   @SuppressWarnings("deprecation")
-  @SuppressLint({"NewApi", "PrivateApi"})
+  @SuppressLint({"NewApi", "PrivateApi", "DiscouragedPrivateApi"})
```
runningcode, line 94: this is the `Choreographer.class.getDeclaredField("mLastFrameTimeNanos");` on line 144

sentry-android-distribution/src/test/java/io/sentry/android/distribution/UpdateResponseParserTest.kt
```diff
@@ -8,8 +8,10 @@ import org.junit.Assert.assertTrue
 import org.junit.Before
 import org.junit.Test
 import org.junit.runner.RunWith
+import org.robolectric.annotation.Config
 
 @RunWith(AndroidJUnit4::class)
+@Config(sdk = [35])
```
runningcode, line 14: we should investigate and/or update robolectric to fix the failure this was causing in a separate PR

sentry-android-integration-tests/sentry-uitest-android-benchmark/build.gradle.kts
```diff
@@ -64,12 +65,12 @@ android {
...
   }
 
-  kotlin { compilerOptions.jvmTarget = org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_1_8 }
+  kotlin { compilerOptions.jvmTarget = JvmTarget.JVM_11 }
 
   lint {
     warningsAsErrors = true
     checkDependencies = true
-    // Suppress OldTargetApi: lint 8.13.1 expects API 37 but we target 36
+    // Suppress OldTargetApi: lint 9.2.1 expects API 37 but we target 36
```
runningcode, line 73: we should remove this once we bump the SDK

sentry-android-integration-tests/sentry-uitest-android-critical/build.gradle.kts
```diff
@@ -31,7 +32,7 @@ android {
       proguardFiles(getDefaultProguardFile("proguard-android-optimize.txt"), "proguard-rules.pro")
     }
   }
-  kotlin { compilerOptions.jvmTarget = org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_1_8 }
+  kotlin { compilerOptions.jvmTarget = JvmTarget.JVM_11 }
```
runningcode, line 35: AGP defaults to 11 now. Since these are our UI tests, it doesn't affect the compiled artifact.

build.gradle.kts
```diff
@@ -113,6 +113,20 @@ allprojects {
 subprojects {
     apply { plugin("io.sentry.spotless") }
 
+    // AGP 9.2 bundles lint 9.2.1, which flags compileSdk 36 as outdated because 37 is available.
+    // We intentionally stay on compileSdk 36 until the API 37 bump (#5768), so silence that check
+    // for every Android module (library and application).
+    pluginManager.withPlugin("com.android.library") {
```
runningcode, line 119: we should remove this once we bump to compile/target sdk 37

sentry-apache-http-client-5/build.gradle.kts
```diff
@@ -10,10 +11,10 @@ plugins {
   alias(libs.plugins.gradle.versions)
 }
 
-tasks.withType<KotlinCompile>().configureEach {
-  compilerOptions.jvmTarget = org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_1_8
-  compilerOptions.languageVersion = org.jetbrains.kotlin.gradle.dsl.KotlinVersion.KOTLIN_1_9
-  compilerOptions.apiVersion = org.jetbrains.kotlin.gradle.dsl.KotlinVersion.KOTLIN_1_9
+kotlin {
```
runningcode, line 14: just a random fix here, this should be declared on the kotlin extension not the task like all the other modules

runningcode:
I'm guessing the failing snapshot test diff is due to a bump in layoutlib which mustve changed the default theme from dark to light: https://github.com/getsentry/sentry-java/pull/5779/checks?check_run_id=87875970201

I went ahead and approved it. I don't think it is worth investigating

sentry-android-core/build.gradle.kts
```diff
@@ -33,7 +34,11 @@ android {
     getByName("release") { consumerProguardFiles("proguard-rules.pro") }
   }
 
-  kotlin { compilerOptions.jvmTarget = org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_1_8 }
+  // AGP 9 only generates unit tests for the testBuildType. CI disables the debug
+  // variant, so unit tests must target release to run at all.
+  testBuildType = "release"
```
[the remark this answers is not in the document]
romtsn: is that what we want? Probably need to update AGENTS.md and skills as well, because we recommend `testDebugUnitTest` all over the place

build.gradle.kts
```diff
@@ -168,6 +182,18 @@ subprojects {
             }
         }
 
+        // AGP 9 defaults Android modules to Java 11. Pin the published library modules back
+        // to Java 8 so their bytecode stays consumable by Java 8 projects, mirroring the
+        // java-library pin above.
+        plugins.withId("com.android.library") {
```
romtsn, line 188: just sanity-checking if this also covers sentry-compose (which is a kmp module)? I guess it should, right?

romtsn (approved):
one question and one comment, but unblocking ✅

build.gradle.kts
runningcode: Yeah, sentry-compose does have the `com.android.library` applied which means that the default java compile language version was bumped from 8 to 11. We have to bring it back down for compatibility.

sentry-android-core/build.gradle.kts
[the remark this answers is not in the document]
runningcode: i'll go back and update the `AGENTS.md`. i don't see any reason that we have a difference between local and CI here. actually now that you mention this, I should just make `shouldSkipDebugVariant` always return true so the debug is always skipped.

PR #5790, opened 2026-07-20
perf(android): mini optimization: Guard manifest metadata debug logs behind isEnabled

### LINKED ISSUE #5700: Reduce logging allocation overhead on the init path
## Problem

Two compounding issues:

1. `ManifestMetadataReader`'s read helpers build the debug message eagerly — `logger.log(DEBUG, key + " read: " + value)` (`sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java:750-800`). With ~101 metadata reads per init, that is ~100+ StringBuilder/String allocations (plus explicit boxing in `readDouble`, `:794-796`) thrown away when `debug=false` (the default), because `DiagnosticLogger` filters *after* the string was built at the call site.
2. Init overall emits 150+ `logger.log(DEBUG, ...)` calls; the varargs overloads allocate an `Object[]` per call even when filtered.

## Proposal

* Guard the metadata read helpers with one `options.isDebug()` / `logger.isEnabled(DEBUG)` check, or pass format + args instead of concatenating.
* Consider an `isEnabled` fast path at the hottest init call sites, or making the `debug=false` logger path allocation-free.

Pure allocation/GC-pressure win; not individually measurable on cold start (getsentry/sentry-java#5679 noise floor), verify via ART allocation trace.

Related: [JAVA-652](https://linear.app/getsentry/issue/JAVA-652/opt-out-for-android-manifest-parsing) / getsentry/sentry-java#5477 would remove the runtime manifest parse entirely; this is the cheap runtime fix in the meantime.

runningcode:
## 📜 Description

This is a mini optimization, this just reduces GC churn in the manifest reading which is a very expensive part of the startup.

This wraps each of the six helpers in a `logger.isEnabled(SentryLevel.DEBUG)` guard so the message is only built when debug logging is actually enabled. The guard is exactly the condition `DiagnosticLogger.log` already applies internally, so behavior is unchanged.

Closes getsentry/sentry-java#5700

## 💚 How did you test it?

Existing `ManifestMetadataReaderTest` coverage exercises these helpers. The change is behavior-preserving (only defers message construction), and `sentry-android-core` compiles with `apiDump` producing no `.api` changes.

## :pencil: Checklist

## 🔮 Next steps

None.

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
@@ -779,7 +779,9 @@ final class ManifestMetadataReader {
       final @NotNull String key,
       final boolean defaultValue) {
     final boolean value = metadata.getBoolean(key, defaultValue);
-    logger.log(SentryLevel.DEBUG, key + " read: " + value);
+    if (logger.isEnabled(SentryLevel.DEBUG)) {
+      logger.log(SentryLevel.DEBUG, key + " read: " + value);
+    }
     return value;
   }
 
@@ -789,7 +791,9 @@ final class ManifestMetadataReader {
       final @NotNull String key,
       final @Nullable String defaultValue) {
     final String value = metadata.getString(key, defaultValue);
-    logger.log(SentryLevel.DEBUG, key + " read: " + value);
+    if (logger.isEnabled(SentryLevel.DEBUG)) {
+      logger.log(SentryLevel.DEBUG, key + " read: " + value);
+    }
     return value;
   }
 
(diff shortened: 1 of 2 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
```diff
@@ -779,7 +779,9 @@ private static boolean readBool(
       final @NotNull String key,
       final boolean defaultValue) {
     final boolean value = metadata.getBoolean(key, defaultValue);
-    logger.log(SentryLevel.DEBUG, key + " read: " + value);
+    if (logger.isEnabled(SentryLevel.DEBUG)) {
```
runningcode, line 782: We could provide a new api to build the strings inside a lambda like Timber does. I wonder if customers have asked for this.

romtsn: not sure if anyone's asked, but sounds like a good idea anyway! Although there's potentially a lambda allocation that's gonna be added on top

romtsn (approved):

PR #5791, opened 2026-07-20
perf(core): Batch and coalesce scope-persistence disk writes

runningcode:
## Summary

Scope persistence wrote to disk on **every** scope mutation, which a customer cold-start trace ([JAVA-628](https://linear.app/getsentry/issue/JAVA-628)) showed to be the single largest SDK cost during startup — larger than init itself. On the Sentry executor thread, during a 1.7 s startup window:

- `FileObjectQueue#add` ×107 — every breadcrumb was a synchronous, fsync'd `QueueFile` append (the file was opened in `rwd` mode).
- `PersistingScopeObserver#store` ×310 — `setContexts`/`setTrace`/`setUser`/`setTags`/… each rewrote their whole file on every change, even though only the latest value matters.
- `JsonSerializer#serialize` ×421.

(The trace was captured with method-level tracing, which inflates per-slice durations; the call counts are the reliable signal, so the wall-clock attributions from the ticket are omitted here.)

I did a benchmark to compare the performance improvement of this PR and an alternative.
Tape as shipped is what we have on `main`, tape without fsync is this PR and JSONL is replacing tape with a single json file.
image (2)

## What changed

Instead of writing eagerly on each mutation, mutations are recorded as pending state and flushed by a single task submitted to the Sentry executor:

- **Latest-wins per file** — each scope field keeps only its most recent pending value, collapsing the many redundant rewrites into a handful.
- **Coalescing without a timer** — at most one flush is queued at a time. The executor is single-threaded, so every mutation that arrives while the flush sits in the queue is folded into the same write. Coalescing therefore tracks executor load rather than a fixed delay: it's strongest exactly during startup, when the queue is deepest, and on an idle app the flush runs almost immediately.
- **Batched breadcrumbs** — breadcrumbs are buffered and appended together behind a **single fsync**. `QueueFile` gains an opt-in buffered-write mode (`synchronousWrites(false)`) plus a `sync()` method; only the breadcrumb queue opts in, the ANR-profile queues keep synchronous writes.
- **Ordered breadcrumb operations** — adds and clears share one queue (a `CLEAR_MARKER` sentinel, mirroring the existing `DELETE_MARKER` for field writes) rather than a list plus a separate flag, so a clear cannot wipe a breadcrumb added after it. Batching split what used to be one ordered stream of executor tasks into independent state; keeping both in one queue makes the ordering intrinsic again.
- Mutations no longer do I/O on the calling thread. Previously a mutation already on the Sentry executor thread ran its write inline; now it only records pending state.

## Tradeoff

Persistence exists to enrich crash/ANR events on the **next** launch and was already asynchronous, so nothing changes about the "data is only needed if the process dies" contract. Because the flush is submitted rather than delayed, pending state is written as soon as the executor drains to it — there is no added time-based loss window. On-disk format is unchanged; restore is unaffected.

## Notes

- Reuses the existing `enableScopePersistence` option; no new public option.
- `requestFlush()` guards against the executor rejecting the flush task. `SentryExecutorService` returns a `CancelledFuture` instead of throwing once its queue is full, which would otherwise leave the pending flag set and silently stop scope persistence for the rest of the process.
- `resetCache()` deliberately leaves pending mutations alone: they only ever hold values from the current process, so dropping them would lose scope state set during init rather than clearing the previous run's data.
- The trace also flagged a 24 ms main-thread `Scopes#addBreadcrumb` call (9.3 ms of it inside `PersistingScopeObserver#addBreadcrumb`, suspected lazy `FileObjectQueue` init). That is intentionally **not** addressed here.

### DIFF
--- a/sentry/src/main/java/io/sentry/cache/PersistingScopeObserver.java
+++ b/sentry/src/main/java/io/sentry/cache/PersistingScopeObserver.java
@@ -38,7 +38,6 @@ import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicBoolean;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
-import org.jetbrains.annotations.TestOnly;
 
 public final class PersistingScopeObserver extends ScopeObserverAdapter {
 
@@ -235,7 +234,7 @@ public final class PersistingScopeObserver extends ScopeObserverAdapter {
       return;
     }
     try {
-      final @NotNull Future<?> future = options.getExecutorService().submit(this::flushOnExecutor);
+      final @NotNull Future<?> future = options.getExecutorService().submit(this::flush);
       if (future.isCancelled()) {
         // the executor rejects tasks without throwing once its queue is full, so clear the flag or
         // no later mutation would ever be able to queue a flush again
@@ -247,18 +246,26 @@ public final class PersistingScopeObserver extends ScopeObserverAdapter {
     }
   }
 
-  private void flushOnExecutor() {
-    runSafely(this::flushPending);
-    // clear the flag before re-checking, otherwise a mutation landing between the drain and the
-    // clear would see a flush still queued and be left with nobody to write it
-    hasPendingFlush.set(false);
-    if (!pendingWrites.isEmpty() || !pendingBreadcrumbs.isEmpty()) {
-      requestFlush();
+  /**
+   * Writes all coalesced scope state to disk. Runs as the queued flush task on the Sentry executor,
+   * which is also the only place it may run: it does I/O, and it owns clearing {@link
+   * #hasPendingFlush} once the write is done.
+   */
+  private void flush() {
+    try {
+      runSafely(this::writePending);
+    } finally {
+      // clear the flag before re-checking, otherwise a mutation landing between the drain and the
+      // clear would see a flush still queued and be left with nobody to write it. In a finally so
+      // an unexpected throw can't leave the flag set and stop persistence for the whole process.
+      hasPendingFlush.set(false);
+      if (!pendingWrites.isEmpty() || !pendingBreadcrumbs.isEmpty()) {
+        requestFlush();
+      }
     }
   }
 
-  /** Writes all coalesced scope state to disk. Does I/O; must run off the caller/main thread. */
-  private void flushPending() {
+  private void writePending() {
     for (final @NotNull String fileName : new ArrayList<>(pendingWrites.keySet())) {
       final @Nullable Object entity = pendingWrites.remove(fileName);
       if (entity == null) {
@@ -297,15 +304,6 @@ public final class PersistingScopeObserver extends ScopeObserverAdapter {
         options.getLogger().log(ERROR, "Failed to sync breadcrumbs file queue", e);
       }
     }
-  }
-
-  /**
-   * Synchronously writes any pending scope state to disk. Does I/O on the calling thread, so it's
-   * only meant for tests and shutdown, not the hot path.
-   */
-  @TestOnly
-  void flush() {
-    runSafely(this::flushPending);
   }
 
   private void runSafely(final @NotNull Runnable task) {
--- a/sentry/src/main/java/io/sentry/cache/PersistingScopeObserver.java
+++ b/sentry/src/main/java/io/sentry/cache/PersistingScopeObserver.java
@@ -34,6 +34,7 @@ import java.util.Map;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicBoolean;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
@@ -42,15 +43,6 @@ import org.jetbrains.annotations.TestOnly;
 public final class PersistingScopeObserver extends ScopeObserverAdapter {
 
   private static final Charset UTF_8 = Charset.forName("UTF-8");
-
-  /**
-   * How long scope mutations are coalesced before being flushed to disk. Rather than writing on
-   * every mutation, we keep only the latest value per file (and buffer breadcrumbs) and flush them
-   * together after this delay. This trades a small data-loss window (mutations from the last
-   * ~{@value #FLUSH_AFTER_MS} ms before the process dies) for far fewer disk writes and fsyncs,
-   * which is significant during startup.
-   */
-  static final long FLUSH_AFTER_MS = 100;
 
   /** Sentinel value marking a file that should be deleted rather than written on the next flush. */
   private static final Object DELETE_MARKER = new Object();
@@ -136,7 +128,7 @@ public final class PersistingScopeObserver extends ScopeObserverAdapter {
   // Breadcrumbs buffered since the last flush, appended together behind a single fsync.
   private final @NotNull Queue<Breadcrumb> pendingBreadcrumbs = new ConcurrentLinkedQueue<>();
   private final @NotNull AtomicBoolean pendingBreadcrumbsClear = new AtomicBoolean(false);
-  private final @NotNull AtomicBoolean hasScheduledFlush = new AtomicBoolean(false);
+  private final @NotNull AtomicBoolean hasPendingFlush = new AtomicBoolean(false);
 
   public PersistingScopeObserver(final @NotNull SentryOptions options) {
     this.options = options;
@@ -153,7 +145,7 @@ public final class PersistingScopeObserver extends ScopeObserverAdapter {
       return;
     }
     pendingBreadcrumbs.offer(crumb);
-    scheduleFlush();
+    requestFlush();
   }
 
   @Override
@@ -167,7 +159,7 @@ public final class PersistingScopeObserver extends ScopeObserverAdapter {
       // drop breadcrumbs buffered before the clear; anything added after it is enqueued again
       pendingBreadcrumbs.clear();
       pendingBreadcrumbsClear.set(true);
-      scheduleFlush();
+      requestFlush();
     }
   }
 
@@ -226,31 +218,42 @@ public final class PersistingScopeObserver extends ScopeObserverAdapter {
     }
     // latest value wins; a null entity means the file should be deleted on the next flush
     pendingWrites.put(fileName, entity == null ? DELETE_MARKER : entity);
-    scheduleFlush();
+    requestFlush();
   }
 
-  @SuppressWarnings("FutureReturnValueIgnored")
-  private void scheduleFlush() {
-    if (!hasScheduledFlush.compareAndSet(false, true)) {
-      // a flush is already scheduled; it will pick up the latest pending state
+  /**
+   * Queues a flush unless one is already queued. Coalescing comes from the flush task sitting in
+   * the executor queue: every mutation that arrives before it runs is folded into the same write.
+   * The executor is single-threaded, so during startup — when mutations are frequent and the queue
+   * is deep — that window covers many mutations.
+   */
+  private void requestFlush() {
+    if (!hasPendingFlush.compareAndSet(false, true)) {
+      // a flush is already queued; it will pick up the latest pending state
       return;
     }
     try {
-      options.getExecutorService().schedule(this::flushOnExecutor, FLUSH_AFTER_MS);
+      final @NotNull Future<?> future = options.getExecutorService().submit(this::flushOnExecutor);
+      if (future.isCancelled()) {
+        // the executor rejects tasks without throwing once its queue is full, so clear the flag or
+        // no later mutation would ever be able to queue a flush again
+        hasPendingFlush.set(false);
+      }
     } catch (Throwable e) {
-      hasScheduledFlush.set(false);
-      options.getLogger().log(ERROR, "Scope persistence flush could not be scheduled", e);
+      hasPendingFlush.set(false);
+      options.getLogger().log(ERROR, "Scope persistence flush could not be submitted", e);
     }
   }
 
   private void flushOnExecutor() {
     runSafely(this::flushPending);
-    hasScheduledFlush.set(false);
-    // reschedule if mutations arrived while we were flushing
+    // clear the flag before re-checking, otherwise a mutation landing between the drain and the
+    // clear would see a flush still queued and be left with nobody to write it
+    hasPendingFlush.set(false);
     if (!pendingWrites.isEmpty()
         || !pendingBreadcrumbs.isEmpty()
         || pendingBreadcrumbsClear.get()) {
-      scheduleFlush();
+      requestFlush();
     }
   }
 
@@ -353,6 +356,9 @@ public final class PersistingScopeObserver extends ScopeObserverAdapter {
    * I/O and should be called from a background thread.
    */
   public void resetCache() {
+    // NOTE: pending mutations are deliberately left alone. They only ever hold values from the
+    // current process, which is exactly what this reset is clearing the way for; dropping them
+    // would lose scope state set during init.
     // since it keeps a reference to the file and we cannot delete it, breadcrumbs we just clear
     try {
       final @NotNull ObjectQueue<Breadcrumb> queue = breadcrumbsQueue.getValue();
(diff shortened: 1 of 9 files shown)


sentry/src/main/java/io/sentry/cache/tape/QueueFile.java
```diff
@@ -130,13 +130,22 @@ public final class QueueFile implements Closeable, Iterable<byte[]> {
   /** A number of elements at which this queue will wrap around (ring buffer). */
   private final int maxElements;
 
+  /**
+   * When {@code false}, writes are buffered by the OS and callers must call {@link #sync()} to make
+   * them durable. This lets callers batch many adds behind a single fsync instead of paying one
```
0xadam-brown: nit: 'make them durable' -> 'flush them to disk'? (IIRC fsync still can't guarantee durability due to file system differences, but it gets us as close as we can.)

(Similarly below.)

sentry/src/main/java/io/sentry/cache/tape/QueueFile.java
```diff
@@ -130,13 +130,22 @@ public final class QueueFile implements Closeable, Iterable<byte[]> {
...
+   * fsync per write.
+   */
+  private final boolean synchronousWrites;
```
0xadam-brown, line 153: l: renaming `bufferedWrites` and flipping the boolean would make it easier to document this. Up to you, of course.

sentry/src/main/java/io/sentry/cache/PersistingScopeObserver.java
```diff
@@ -106,143 +126,188 @@ public void toStream(Breadcrumb value, OutputStream sink) throws IOException {
                 });
           });
 
+  // Latest pending value per file (or DELETE_MARKER), coalesced until the next flush.
+  private final @NotNull Map<String, Object> pendingWrites = new ConcurrentHashMap<>();
+  // Breadcrumb adds and clears buffered since the last flush, applied together behind a single
```
0xadam-brown: nit: I'm having trouble interpreting "adds and clears buffered". Maybe: "Tracks requests to add and clear breadcrumbs since last flush..."?

sentry/src/main/java/io/sentry/cache/PersistingScopeObserver.java
```diff
@@ -106,143 +126,188 @@ public void toStream(Breadcrumb value, OutputStream sink) throws IOException {
...
+   */
+  private void requestFlush() {
+    if (!hasPendingFlush.compareAndSet(false, true)) {
+      // a flush is already queued; it will pick up the latest pending state
       return;
     }
-
     try {
-      options.getExecutorService().submit(() -> runSafely(task));
+      final @NotNull Future<?> future = options.getExecutorService().submit(this::flushOnExecutor);
```
0xadam-brown: h: Given your PR description, should we be batching over 100 ms windows, rather than just whenever writes are slower than requests to enqueue?

sentry/src/main/java/io/sentry/cache/PersistingScopeObserver.java
```diff
@@ -106,143 +126,188 @@ public void toStream(Breadcrumb value, OutputStream sink) throws IOException {
...
+    // clear the flag before re-checking, otherwise a mutation landing between the drain and the
+    // clear would see a flush still queued and be left with nobody to write it
+    hasPendingFlush.set(false);
+    if (!pendingWrites.isEmpty() || !pendingBreadcrumbs.isEmpty()) {
+      requestFlush();
     }
   }
 
+  /** Writes all coalesced scope state to disk. Does I/O; must run off the caller/main thread. */
+  private void flushPending() {
```
0xadam-brown: nit: Consider `flushEnqueuedValues()`. (Fwiw, I tripped over `flushPending()` a few times because I interpreted it as meaning a flush request was pending.)

sentry/src/test/java/io/sentry/cache/tape/QueueFileTest.kt
```diff
@@ -72,6 +73,21 @@ class QueueFileTest {
     assertArrayEquals(queue.peek(), expected)
   }
 
+  @Test
+  fun bufferedWritesSurviveReopenAfterSync() {
```
0xadam-brown: nit: Maybe `syncPersistsBufferedWrites`? 

(Fwiw, I'm having trouble figuring out what it means for a write to "survive reopen" – educate me if I'm missing something!)

sentry/src/main/java/io/sentry/cache/PersistingScopeObserver.java
```diff
@@ -106,143 +126,188 @@ public void toStream(Breadcrumb value, OutputStream sink) throws IOException {
...
   public void addBreadcrumb(@NotNull Breadcrumb crumb) {
-    serializeToDisk(
-        () -> {
-          try {
-            breadcrumbsQueue.getValue().add(crumb);
-          } catch (IOException e) {
-            options.getLogger().log(ERROR, "Failed to add breadcrumb to file queue", e);
-          }
-        });
+    if (!options.isEnableScopePersistence()) {
```
0xadam-brown, line 147: l: Thoughts about creating an `enqueue()` method for breadcrumbs? That'd let us match what we do for pending writes. (The difference raised questions in my head / required me to jump through the implementations before I could convince myself there wasn't a meaningful difference.)

sentry/src/main/java/io/sentry/cache/PersistingScopeObserver.java
```diff
@@ -106,143 +126,188 @@ public void toStream(Breadcrumb value, OutputStream sink) throws IOException {
...
+        hasPendingFlush.set(false);
+      }
     } catch (Throwable e) {
-      options.getLogger().log(ERROR, "Serialization task could not be scheduled", e);
+      hasPendingFlush.set(false);
+      options.getLogger().log(ERROR, "Scope persistence flush could not be submitted", e);
+    }
+  }
+
+  private void flushOnExecutor() {
```
0xadam-brown: l: Thoughts about merging this and `flushPending()`?  (Or calling `flushPending()` -> `doFlushPending()` and renaming this `flushPending()` if passing a big lambda to `runSafely` looks awkward.)

It's not obvious why this method needs to advertise that it's being run on our executor, as nothing in it seems to require it. (But let me know if I'm mistaken.)

sentry/src/main/java/io/sentry/cache/PersistingScopeObserver.java
```diff
@@ -74,7 +90,11 @@ public final class PersistingScopeObserver extends ScopeObserverAdapter {
                 // update where the new format was introduced
                 file.delete();
 
-                queueFile = new QueueFile.Builder(file).size(options.getMaxBreadcrumbs()).build();
```
0xadam-brown, line 77: m: We should log an error if the queue file is corrupted.

QueueFile's Javadoc warns about file corruption if power goes out when writing a segment. Buffering means the risk window is larger.  (Would've been much nicer to already have a baseline we can compare against, but we can at least keep an eye on things going forward.)

0xadam-brown (commented):
Thanks for this @runningcode! A few comments for your consideration. 

And two big-picture thoughts...

**Thread-safety**: 

I see that `QueueFile` isn't thread-safe. Worth double-checking that we don't expect the buffered write patterns to aggravate any concurrency concerns that were already there but perhaps less common (totally hand-wavy on my part – just flagging that I haven't done a thread-safety deep dive in my review).

**Testing / monitoring**:

Given that this is core infrastructure, do you plan to perform any manual tests / add tests to the sample app, etc.?

sentry/src/main/java/io/sentry/cache/PersistingScopeObserver.java
runningcode: we `catch IOException e` on the next line and then we log `Failed to create breadcrumbs queue`

is that what you mean? Or otherwise how/where should we detect that the queue file is corrupted?

sentry/src/main/java/io/sentry/cache/PersistingScopeObserver.java
runningcode: do you mean change the API from `addBreadcrumb` to enqueBreadcrumb` or adding a new method?

I shied away from breaking API changes here because this would break the `IScopeObserver`. `NdkScopeObserver` inherits from that and then I just shied away from changing it.
setBreadcrumbs below is poorly named as well. I filed a ticket for that for our next major version: https://linear.app/getsentry/issue/JAVA-664/replace-misleading-iscopeobserversetbreadcrumbs-with-clearbreadcrumbs

sentry/src/main/java/io/sentry/cache/tape/QueueFile.java
runningcode: thanks for pointing this out! I think synchronous describes this situation better. Otherwise we'd have to get in to the nitty gritty of debating if it should be called `queuedWrites` or `bufferedWrites` and how does it behave if `queuedWrites` is false? imo synchronous and not synchronous describes it better.

sentry/src/main/java/io/sentry/cache/PersistingScopeObserver.java
runningcode: oops that was from an earlier implementation. ive updated the PR description.

In an earlier commit, it batch breadcrumbs over 100ms windows. I removed that for simplicity.

sentry/src/main/java/io/sentry/cache/PersistingScopeObserver.java
runningcode: This split is here because of the `@TestOnly flush()` on line 308 which cannot touch the `hasPendingFlush` state.
You make a good point, an alternative would be to have a single method with a boolean field (`flush(boolean fromExecutor)`) where it doesn't update the state depending on the boolean. I don't like boolean methods though.

Agreed that this is not the best design but I'm not sure how to improve it.

sentry/src/main/java/io/sentry/cache/PersistingScopeObserver.java
runningcode: I rewrote it so there's only a single `flush()` function. WDYT?

runningcode (commented):
Thanks for the thorough review, @0xadam-brown ! I've replied to your points and fixed the rest in the PR.

sentry/src/main/java/io/sentry/cache/PersistingScopeObserver.java
```diff
@@ -106,140 +125,184 @@ public void toStream(Breadcrumb value, OutputStream sink) throws IOException {
...
-  @SuppressWarnings("FutureReturnValueIgnored")
-  private void serializeToDisk(final @NotNull Runnable task) {
+  private void enqueue(final @NotNull String fileName, final @Nullable Object entity) {
     if (!options.isEnableScopePersistence()) {
       return;
     }
-    if (SentryExecutorService.isSentryExecutorThread()) {
-      // we're already on the sentry executor thread, so we can just execute it directly
-      runSafely(task);
+    // latest value wins; a null entity means the file should be deleted on the next flush
```
romtsn, line 220: neat!

sentry/src/main/java/io/sentry/cache/PersistingScopeObserver.java
```diff
@@ -106,140 +125,184 @@ public void toStream(Breadcrumb value, OutputStream sink) throws IOException {
...
+      // an unexpected throw can't leave the flag set and stop persistence for the whole process.
+      hasPendingFlush.set(false);
+      if (!pendingWrites.isEmpty() || !pendingBreadcrumbs.isEmpty()) {
+        requestFlush();
+      }
+    }
+  }
+
+  private void writePending() {
+    for (final @NotNull String fileName : new ArrayList<>(pendingWrites.keySet())) {
```
romtsn: `l`: could've been a `pendingWrites.iterator()`?

sentry/src/main/java/io/sentry/cache/PersistingScopeObserver.java
```diff
@@ -106,140 +125,184 @@ public void toStream(Breadcrumb value, OutputStream sink) throws IOException {
...
+      runSafely(this::writePending);
+    } finally {
+      // clear the flag before re-checking, otherwise a mutation landing between the drain and the
+      // clear would see a flush still queued and be left with nobody to write it. In a finally so
+      // an unexpected throw can't leave the flag set and stop persistence for the whole process.
+      hasPendingFlush.set(false);
+      if (!pendingWrites.isEmpty() || !pendingBreadcrumbs.isEmpty()) {
+        requestFlush();
+      }
+    }
```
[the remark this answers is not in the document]
romtsn: probably valid, but i don't think it's _that_ severe, probably just risking losing breadcrumbs added in the first few ms after `init`

sentry/src/main/java/io/sentry/cache/tape/QueueFile.java
romtsn, line 1: `l`: probs need to update javadoc of this class since atomicity cannot be guaranteed anymore when synchronousWrites is disabled

romtsn (approved):
couple of minor things, but looks good, great improvement!

sentry/src/main/java/io/sentry/cache/PersistingScopeObserver.java
0xadam-brown: Ah, I had in mind logging an error here in connection with having caught the original IOException that led us to believe the queue file was corrupted (see the inline comment immediately above). 

(The logging statement in the catch clause below is only triggered if we fail to recreate a new queue after discovering the first was corrupted.)

sentry/src/main/java/io/sentry/cache/PersistingScopeObserver.java
0xadam-brown: > do you mean change the API from addBreadcrumb to enqueBreadcrumb` or adding a new method?

Gotcha – agree we don't want to change the public API. I actually meant creating a new private method like this: 

```java
private void enqueueBreadcrumb(@NotNull Breadcrumb crumb) {
  if (!options.isEnableScopePersistence()) {
    return;
  }
  pendingBreadcrumbs.offer(crumb);
  requestFlush();
}
```

...and then calling that method here + in `setBreadcrumbs()`.

That'd^^ match the pattern we use with pending writes, which call an `enqueue()` method with a similar implementation.

sentry/src/main/java/io/sentry/cache/tape/QueueFile.java
runningcode: the operations are all still atomic we only lose the guarantee of durability as in we might not be able to write everything before power loss if we batch up the writes. this was always the case though

runningcode: i'll update the javadoc

sentry/src/main/java/io/sentry/cache/tape/QueueFile.java
```diff
@@ -788,15 +811,24 @@ public Builder size(int size) {
...
      * file at a time.
      */
     public QueueFile build() throws IOException {
-      RandomAccessFile raf = initializeFromFile(file);
+      RandomAccessFile raf = initializeFromFile(file, synchronousWrites);
       QueueFile qf = null;
       try {
-        qf = new QueueFile(file, raf, zero, size);
+        qf = new QueueFile(file, raf, zero, size, synchronousWrites);
         return qf;
```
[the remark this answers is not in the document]
runningcode: IIUC the `setLength` is only used to expand the file if needed in which case we want it to be done synchronously.

sentry/src/main/java/io/sentry/cache/PersistingScopeObserver.java
runningcode: ohhh i see what you mean. its a good point to reduce duplication. but to be honest, i don't want to wait on another CI round.

sentry/src/main/java/io/sentry/cache/PersistingScopeObserver.java
runningcode: I don't think I caused a regression on this IIUC. You're just saying you would add logging in general? I can do that as a follow up if you feel strongly but I will resolve this for now to merge this PR.

sentry/src/main/java/io/sentry/cache/PersistingScopeObserver.java
```diff
@@ -106,140 +125,190 @@ public void toStream(Breadcrumb value, OutputStream sink) throws IOException {
...
     }
-
     try {
-      options.getExecutorService().submit(() -> runSafely(task));
+      final @NotNull Future<?> future = options.getExecutorService().submit(this::flush);
+      if (future.isCancelled()) {
+        // the executor rejects tasks without throwing once its queue is full, so clear the flag or
+        // no later mutation would ever be able to queue a flush again
+        hasPendingFlush.set(false);
+      }
```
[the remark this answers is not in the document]
runningcode: This isn't a reachable bug but has the potential to introduce bugs. Following up on this here: https://github.com/getsentry/sentry-java/pull/5874

Issue #5704, opened 2026-07-03
Run the Integration.register() loop off the main thread (two-phase init)


runningcode:
## Problem

`Sentry.init` registers all integrations synchronously on the calling thread (`sentry/src/main/java/io/sentry/Sentry.java:374-385`) — the main thread under auto-init. With ~17 default Android integrations this is the biggest remaining *block* of main-thread init work once the per-integration items (NDK etc.) are fixed.

## Proposal

Keep a minimal synchronous core — options, scopes, `SentryClient` binding, `UncaughtExceptionHandlerIntegration` — and submit the rest of the registration loop to `options.getExecutorService()`, with an opt-in marker for integrations that genuinely need the main thread (`AppLifecycleIntegration` already self-posts when off-main).

`Sentry.captureX` immediately after `init()` keeps working: the client is bound synchronously and events queue on the transport.

## Constraints

* The deliberate startup-crash blocking flush (`sentry-android-core/.../SendCachedEnvelopeIntegration.java:151-161`) must still block `init()` — needs an explicit carve-out.
* Integration ordering guarantees (NDK before outbox watcher, AppLifecycle before ANR, ANR before Replay — see comments in `AndroidOptionsInitializer.installDefaultIntegrations`) must be preserved. The single-threaded executor keeps submission order, but mixed sync/async registration could break it; audit each.
* Session auto-start + `isForegroundImportance()` at the end of `SentryAndroid.init` (`SentryAndroid.java:184-203`) depend on integration state (LifecycleWatcher) — decide whether they move too.

Concrete follow-up to getsentry/sentry-java#5475 ("Move more init code async").

Drayl42:
## One switch to run the entire SDK init off the main thread (diagram attached)

Hi Sentry team,

We've been following the startup-overhead work in recent releases closely, and the two-phase init described in this issue is exactly the direction we've been hoping for. We'd like to suggest taking it one step further.

### Motivation: 
[...]

PR #5792, opened 2026-07-20
perf(core): Create outbox and cache dirs lazily instead of during init (JAVA-613)

runningcode:
## :scroll: Description

`Sentry.initConfigurations` created the outbox and cache directories synchronously via `mkdirs()` on the init thread — which on Android is the main thread. This moves both off the synchronous init path by creating each directory lazily in its consumer:

- **Cache dir** — created at the start of `EnvelopeCache.storeInternal`, so the first envelope write (on the transport thread) creates it. Covers envelopes, sessions, crash markers, and the Android ANR/tombstone markers, which all live under `cacheDirPath`.
- **Outbox dir** — created in `EnvelopeFileObserverIntegration.startOutboxSender` (runs on the executor) before `startWatching()`, and before writing the startup-crash marker in `AndroidEnvelopeCache`. The native SDK already creates the outbox dir itself during `sentry_init`, so NDK crash writes are unaffected.

### :warning: Behavior change for outbox writers

Previously `Sentry.init` created the outbox dir synchronously, so it was guaranteed to exist once `init` returned. It is now created on the SDK executor, so anything writing envelopes into the outbox itself must create the dir first (`new File(outboxPath).mkdirs()`). I checked RN, Flutter, Unity and none of them should be affected by this. Native was fixed in 0.16.0 and that is already merged.

This can cause ANRs as we are blocking the main thread with disk I/O during startup.

Relying on integration tests for this. Some of them failed with this integration. I also asked JR which other SDKs might be affected by this.

## :pencil: Checklist

## :crystal_ball: Next steps

Verify the init-time improvement as a batch via Perfetto/macrobenchmark (individually below noise).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Behavioral Changes
+
+- The outbox and cache directories are no longer created by `Sentry.init` ([#5792](https://github.com/getsentry/sentry-java/pull/5792))
+  - They are now created lazily by whichever component first writes into them, off the init thread. As a result, the directories at `SentryOptions.getOutboxPath()` and `SentryOptions.getCacheDirPath()` are not guaranteed to exist once `Sentry.init` returns.
+  - If you write envelopes into the outbox path yourself instead of going through the SDK — as hybrid SDKs do for `captureEnvelope` — create the directory first, e.g. `new File(outboxPath).mkdirs()`.
+
 ### Improvements
 
 - Skip building Android manifest metadata debug log messages when debug logging is disabled, reducing allocations during SDK init ([#5790](https://github.com/getsentry/sentry-java/pull/5790))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 
 ### Performance
 
+- Create the outbox and cache directories lazily in their consumers instead of during SDK init, moving the `mkdirs()` calls off the init (main) thread ([#5792](https://github.com/getsentry/sentry-java/pull/5792))
 - Reduce the number of SDK threads: `LifecycleWatcher` now schedules the session-end task on the shared timer executor instead of creating a dedicated `java.util.Timer` thread ([#5819](https://github.com/getsentry/sentry-java/pull/5819))
 - Speed up deserialization of arbitrary JSON objects by typing numbers without throwing exceptions ([#5783](https://github.com/getsentry/sentry-java/pull/5783))
 
--- a/sentry-android-core/src/main/java/io/sentry/android/core/cache/AndroidEnvelopeCache.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/cache/AndroidEnvelopeCache.java
@@ -48,7 +48,8 @@ public final class AndroidEnvelopeCache extends EnvelopeCache {
       final @NotNull ICurrentDateProvider currentDateProvider) {
     super(
         options,
-        Objects.requireNonNull(options.getCacheDirPath(), "cacheDirPath must not be null"),
+        new LazyDirectory(
+            Objects.requireNonNull(options.getCacheDirPath(), "cacheDirPath must not be null")),
         options.getMaxCacheItems());
     this.currentDateProvider = currentDateProvider;
   }
--- a/sentry-android-core/src/main/java/io/sentry/android/core/cache/AndroidEnvelopeCache.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/cache/AndroidEnvelopeCache.java
@@ -18,7 +18,6 @@ import io.sentry.cache.EnvelopeCache;
 import io.sentry.transport.ICurrentDateProvider;
 import io.sentry.util.FileUtils;
 import io.sentry.util.HintUtils;
-import io.sentry.util.LazyDirectory;
 import io.sentry.util.Objects;
 import java.io.File;
 import java.io.FileNotFoundException;
@@ -48,8 +47,7 @@ public final class AndroidEnvelopeCache extends EnvelopeCache {
       final @NotNull ICurrentDateProvider currentDateProvider) {
     super(
         options,
-        new LazyDirectory(
-            Objects.requireNonNull(options.getCacheDirPath(), "cacheDirPath must not be null")),
+        Objects.requireNonNull(options.getCacheDirPath(), "cacheDirPath must not be null"),
         options.getMaxCacheItems());
     this.currentDateProvider = currentDateProvider;
   }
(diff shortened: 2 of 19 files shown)


runningcode:
Native needs to be merged first: https://github.com/getsentry/sentry-native/pull/1889

sentry/src/main/java/io/sentry/cache/EnvelopeCache.java
```diff
@@ -385,7 +388,7 @@ public void discard(final @NotNull SentryEnvelope envelope) {
         fileNameMap.put(envelope, fileName);
       }
 
-      return new File(directory.getAbsolutePath(), fileName);
+      return directory.resolve(fileName);
     }
   }
```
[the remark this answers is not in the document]
runningcode: This seems like an existing issue.

sentry-android-core/src/main/java/io/sentry/android/core/cache/AndroidEnvelopeCache.java
```diff
@@ -106,7 +106,14 @@ private void writeStartupCrashMarkerFile() {
           .log(DEBUG, "Outbox path is null, the startup crash marker file will not be written");
       return;
     }
-    final File crashMarkerFile = new File(outboxPath, STARTUP_CRASH_MARKER_FILE);
+    // The outbox dir is no longer created during Sentry.init, so create it here in case the native
+    // SDK (which normally creates it) is disabled.
+    final File outboxDir = new File(outboxPath);
```
0xadam-brown, line 111: **Longer-term side note:** Seems like `Sentry.init()` would benefit from an `onCompleted` callback so we can i) do work off the main thread and ii) provide users with guarantees about what will be in place when we invoke the callback. 

That'd^^ let us centralize our policies + avoid having to check for existence (etc.) in multiple places.

(Obviously way out of scope here, but something for us to think about.)

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -1104,7 +1104,11 @@ String getCacheDirPathWithoutDsn() {
   }
 
   /**
-   * Returns the outbox path if cacheDirPath is set
+   * Returns the outbox path if cacheDirPath is set.
+   *
+   * <p>The directory is created lazily by the SDK on a background thread, so it is not guaranteed
```
0xadam-brown, line 1109: l: Thoughts about introducing a new `getOutboxPathOrCreate()` method that always returns the path but also creates the directory if it doesn't currently exist? (Presumably we'd want a new `getCacheDirPathOrCreate()` for symmetry.)

That'd^^ make our code's intentions more explicit + make it easier to track which call sites need the directory to actually exist vs those that just need the path string.

Up to you...

sentry/src/main/java/io/sentry/util/LazyDirectory.java
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  /** Returns the directory without touching the filesystem. */
+  public @NotNull File getFile() {
+    return file;
+  }
+
+  /** Returns the directory, creating it and any missing parents if it does not exist yet. */
+  public @NotNull File getOrCreate() {
+    FileUtils.createDirectory(file);
```
0xadam-brown, line 35: m: We're ignoring a possible `false` return value (which indicates that the directory couldn't be created). Seems we should either return null from this method or update the Javadoc to advertise it's a best-effort-only approach.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 
 ### Performance
 
+- Create the outbox and cache directories lazily in their consumers instead of during SDK init, moving the `mkdirs()` calls off the init (main) thread ([#5792](https://github.com/getsentry/sentry-java/pull/5792))
```
0xadam-brown, line 22: m: We should also mention that direct outbox writers now need to create the outbox directory before writing envelopes.

0xadam-brown (approved):
Thanks @runningcode!

A few comments, but not blockers. 

Fwiw, I asked my LLM whether the PR accounts for all possible code paths that may need to create the directory, now that we're not doing so in `init()`, and it said it does. But I'm not familiar with those paths at all, so feel free to ask for another look if you're also not sure.

sentry-android-core/src/main/java/io/sentry/android/core/cache/AndroidEnvelopeCache.java
runningcode: sounds interesting! probably related to this two phase init idea: https://github.com/getsentry/sentry-java/issues/5704

sentry/src/main/java/io/sentry/util/LazyDirectory.java
runningcode: This is a very good catch. To be fair, the previous code would also ignore the return value of `mkdirs` when creating the dir.
`LazyDirectory` is just a basic class. Adding a logger to it would be overkill. The error should surface when trying to write later on. Ill make a comment to make this explicit.

sentry/src/main/java/io/sentry/SentryOptions.java
runningcode: I think we should use another mechanism for this. A mechanism which we don't have. (Something like DI which is I think what our Scopes does. Either way, I don't think it's a good idea to introduce disk IO to the SentryOptions class.

CHANGELOG.md
runningcode: I feel like this might confuse consumers of the SDK but I'll do this. I think it makes sense to put in the release notes.

runningcode (commented):
Thanks for the review

sentry/src/main/java/io/sentry/SentryOptions.java
0xadam-brown: Nw. "A mechanism which we don't have" was my feeling too....

sentry/src/main/java/io/sentry/util/FileUtils.java
```diff
@@ -36,6 +37,21 @@ public static boolean deleteRecursively(@Nullable File file) {
...
+   * <p>Callers are expected to log a failure: a missing directory otherwise surfaces later as an
+   * unrelated-looking write error.
+   *
+   * @param directory the directory to create
+   * @return true if the directory exists once this returns, false if it could not be created
+   */
+  public static boolean createDirectory(final @NotNull File directory) {
+    // mkdirs() also returns false when another thread created the directory first, so re-check
+    // instead of reporting a failure the caller would act on by skipping its write.
+    return directory.isDirectory() || directory.mkdirs() || directory.isDirectory();
```
romtsn, line 52: not gonna lie this reads super confusing 😂

sentry/src/main/java/io/sentry/Sentry.java
```diff
@@ -616,19 +616,14 @@ private static void initConfigurations(final @NotNull SentryOptions options) {
...
+    // The outbox and cache dirs are created lazily by their consumers (envelope cache, outbox file
+    // observer, native SDK) off the init thread, so we don't stat/mkdir them here.
+    if (options.getOutboxPath() == null) {
       logger.log(SentryLevel.INFO, "No outbox dir path is defined in options.");
     }
 
     final String cacheDirPath = options.getCacheDirPath();
     if (cacheDirPath != null) {
-      final File cacheDir = new File(cacheDirPath);
-      cacheDir.mkdirs();
```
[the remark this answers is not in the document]
romtsn: Has this been actually resolved? I think for that we'd need to create the dir in `processDirectory()`?

romtsn (approved):
Unsure whether one cursor comment has already been resolved or not, but lgtm otherwise!

sentry/src/main/java/io/sentry/Sentry.java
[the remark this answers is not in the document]
runningcode: Creating the directory doesn't really change any behavior here since there's nothing to process inside the directory if we freshly create it.

My understanding is that the `SendCachedEnvelopeFireAndForgetIntegration` refires on every connection change so the cached envelopes will get sent up the next time after the directory is created.

sentry/src/main/java/io/sentry/util/FileUtils.java
runningcode: agreed :(

PR #5795, opened 2026-07-21
feat(android): Add internal non-terminating envelope capture 

buenaflor:
## :scroll: Description

Add `InternalSentrySdk.captureEnvelopeNonTerminating` for hybrid runtimes such as Flutter where an unhandled exception does not terminate the process.

The new path keeps the current session alive with the same SID, increments its error count, persists a pending-unhandled marker in the existing session cache, and finalizes it as `unhandled` only when the session naturally ends. Existing `Abnormal` and `Crashed` terminal states remain authoritative, and the existing `captureEnvelope(byte[], boolean)` behavior used by other SDKs is unchanged.

Flutter currently forwards `handled=false` events through the terminating hybrid capture path. This marks the session as `crashed` and may start a replacement session even though the Flutter process continues running, which lowers crash-free session rates incorrectly.

This follows the session protocol's `unhandled` status for exceptions that are unhandled by the application but do not terminate the process.

- Added coverage for pending session mutation and serialization, natural/abnormal/crashed terminal precedence, previous-session recovery, SID-aware cache start/end handling, null SID handling, and terminating capture compatibility.

## :pencil: Checklist

## :crystal_ball: Next steps

- Update the Flutter Android bridge to use `captureEnvelopeNonTerminating` for non-terminating unhandled events.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
@@ -40,7 +40,7 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.OutputStreamWriter;
 import java.io.Writer;
-import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -54,9 +54,6 @@ import org.jetbrains.annotations.Nullable;
 /** Sentry SDK internal API methods meant for being used by the Sentry Hybrid SDKs. */
 @ApiStatus.Internal
 public final class InternalSentrySdk {
-
-  @SuppressWarnings("CharsetObjectCanBeUsed")
-  private static final Charset UTF_8_CHARSET = Charset.forName("UTF-8");
 
   /**
    * @return a copy of the current scopes's topmost scope, or null in case the scopes is disabled
@@ -174,7 +171,58 @@ public final class InternalSentrySdk {
   @Nullable
   public static SentryId captureEnvelope(
       final @NotNull byte[] envelopeData, final boolean maybeStartNewSession) {
-    return captureEnvelope(envelopeData, maybeStartNewSession, true);
+    final @NotNull IScopes scopes = ScopesAdapter.getInstance();
+    final @NotNull SentryOptions options = scopes.getOptions();
+
+    try (final InputStream envelopeInputStream = new ByteArrayInputStream(envelopeData)) {
+      final @NotNull ISerializer serializer = options.getSerializer();
+      final @Nullable SentryEnvelope envelope =
+          options.getEnvelopeReader().read(envelopeInputStream);
+      if (envelope == null) {
+        return null;
+      }
+
+      final @NotNull List<SentryEnvelopeItem> envelopeItems = new ArrayList<>();
+
+      // determine session state based on events inside envelope
+      @Nullable Session.State status = null;
+      boolean crashedOrErrored = false;
+      for (SentryEnvelopeItem item : envelope.getItems()) {
+        envelopeItems.add(item);
+
+        final SentryEvent event = item.getEvent(serializer);
+        if (event != null) {
+          if (event.isCrashed()) {
+            status = Session.State.Crashed;
+          }
+          if (event.isCrashed() || event.isErrored()) {
+            crashedOrErrored = true;
+          }
+        }
+      }
+
+      // update session and add it to envelope if necessary
+      final @Nullable Session session =
+          updateSession(scopes, options, status, crashedOrErrored, false);
+      if (session != null) {
+        final SentryEnvelopeItem sessionItem = SentryEnvelopeItem.fromSession(serializer, session);
+        envelopeItems.add(sessionItem);
+        deleteCurrentSessionFile(
+            options,
+            // should be sync if going to crash or already not a main thread
+            !maybeStartNewSession || !scopes.getOptions().getThreadChecker().isMainThread());
+        if (maybeStartNewSession) {
+          scopes.startSession();
+        }
+      }
+
+      final SentryEnvelope repackagedEnvelope =
+          new SentryEnvelope(envelope.getHeader(), envelopeItems);
+      return scopes.captureEnvelope(repackagedEnvelope);
+    } catch (Throwable t) {
+      options.getLogger().log(SentryLevel.ERROR, "Failed to capture envelope", t);
+    }
+    return null;
   }
 
   /**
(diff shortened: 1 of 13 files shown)


buenaflor:
@cursor review

buenaflor:
@sentry review

buenaflor:
@sentry review

buenaflor:
@cursor review

sentry/src/main/java/io/sentry/cache/EnvelopeCache.java
```diff
@@ -115,8 +116,22 @@ private boolean storeInternal(final @NotNull SentryEnvelope envelope, final @Not
...
+                && currentSession != null
+                && currentSession.isPendingUnhandled()
+                && endingSession.getSessionId() != null
+                && currentSession.getSessionId() != null
+                && !Objects.equals(endingSession.getSessionId(), currentSession.getSessionId())
+                && endingSession.getStarted() != null
+                && currentSession.getStarted() != null
+                && currentSession.getStarted().after(endingSession.getStarted());
+        if (!preservePendingSession && !currentSessionFile.delete()) {
+          options.getLogger().log(WARNING, "Current envelope doesn't exist.");
```
buenaflor, line 133: Instead of deleting `session.json` unconditionally, we preserve it when a delayed `End(A)` would delete a newer pending-unhandled session B, confirmed by a different SID and later start time. For example, if `End(A)` is queued before B’s pending-unhandled state is written, processing `End(A)` must not delete B’s file. All other cases retain the existing deletion behavior, so this currently only affects state persisted by `captureEnvelopeNonTerminating()`.

sentry/src/main/java/io/sentry/Session.java
```diff
@@ -21,7 +21,8 @@ public enum State {
     Ok,
     Exited,
     Crashed,
-    Abnormal
+    Abnormal,
+    Unhandled
   }
 
   /** started timestamp */
```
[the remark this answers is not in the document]
buenaflor: relay supports `unhandled`

sentry-android-core/src/test/java/io/sentry/android/core/InternalSentrySdkTest.kt
```diff
@@ -38,6 +39,7 @@ import java.util.concurrent.atomic.AtomicReference
 import kotlin.test.BeforeTest
 import kotlin.test.Test
 import kotlin.test.assertEquals
+import kotlin.test.assertFalse
```
runningcode: do you know why this test was added with kotlin.test but the others used google truth? we should prefer google truth because it is easier to debug.

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
```diff
@@ -213,6 +218,90 @@ public static SentryId captureEnvelope(
...
+   * crashed}.
+   *
+   * <p>Same as {@link #captureEnvelope(byte[], boolean)}, this method will not enrich events, run
+   * {@code beforeSend}, or sample — the caller is responsible for that.
+   *
+   * @param envelopeData the serialized envelope data
+   * @return the id of the captured envelope, or null if capture failed
+   */
+  @Nullable
+  public static SentryId captureEnvelopeNonTerminating(final @NotNull byte[] envelopeData) {
```
runningcode, line 246: is there any part of the logic here that we can share/duplicate with the `captureEnvelope` method?
For example maybe they both forward to a private method that has a `boolean terminating = true/false`?

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
```diff
@@ -213,6 +218,90 @@ public static SentryId captureEnvelope(
...
+                          .getLogger()
+                          .log(INFO, "Session is null on captureEnvelopeNonTerminating");
+                    }
+                  });
+            });
+      }
+
+      // Capture the original envelope as-is (no session item attached).
+      return scopes.captureEnvelope(envelope);
+    } catch (Throwable t) {
```
runningcode: Do you know what kinds of exceptions we might expect here? I think we should either explicitly document them for when we remove `catch Throwable` or not add new `catch Throwable`

sentry/src/main/java/io/sentry/cache/EnvelopeCache.java
```diff
@@ -306,12 +334,37 @@ private void updateCurrentSession(
...
+  }
+
+  private @Nullable Session readSessionFromDisk(final @NotNull File sessionFile) {
+    if (!sessionFile.exists()) {
+      return null;
+    }
+    try (final Reader reader =
+        new BufferedReader(new InputStreamReader(new FileInputStream(sessionFile), UTF_8))) {
+      return serializer.getValue().deserialize(reader, Session.class);
+    } catch (Throwable e) {
```
runningcode: Same comment here as the other `Throwable`

buenaflor:
Superseded by a stack that splits this into three reviewable PRs, rebased onto current `main`:

- #5915 (collection)
- #5916 — `Session` model: `State.Unhandled` + the pending-unhandled marker
- #5917 — `EnvelopeCache`: session-id-aware session file writes
- #5918 — `InternalSentrySdk.captureEnvelopeNonTerminating`

All review feedback from this PR is carried over:

- @runningcode's `catch (Throwable)` comments — narrowed to `catch (Exception)` in both `InternalSentrySdk` capture paths and `EnvelopeCache.readSessionFromDisk`, so `OutOfMemoryError` and friends propagate instead of being swallowed. `Exception` is the tightest catch available, since `SentryEnvelopeItem.getEvent` declares `throws Exception`.
- @runningcode's code-sharing comment — both capture paths now share a private `readEnvelope` helper. I skipped the `boolean terminating` flag: the session handling either side of it is fully disjoint (repackage + session item + delete file + `startSession` vs. mutate in place + persist + capture as-is).
- @runningcode's Truth comment — the new tests now use Truth, with `testImplementation(libs.google.truth)` added to `sentry-android-core`. Pre-existing `kotlin.test` assertions in that file are left alone.
- Copilot's deprecated `Date(2023, 10, 1)` comment — fixed, now `DateUtils.getDateTime("2023-10-01T00:00:00.000Z")`.
- The bot's `SessionEnd` stale-file comment — fixed; the preserve guard requires a non-null parse on both sides, so parse failures still delete. Covered by `malformed SessionEnd deletes current session` and `unreadable current session on SessionEnd is deleted`.
- The bot's HIGH-severity claim that relay rejects `"unhandled"` — not valid. Relay's `SessionStatus` includes `Unhandled` (`as_str` → `"unhandled"`), and the [session protocol](https://develop.sentry.dev/sdk/telemetry/sessions/) has listed it as a valid terminal status since 1.6.0.
- Copilot's enum-compatibility comment — keeping the enum constant. There is no `switch`/`when` over `Session.State` in the codebase; consumers go through `isTerminated()`, which correctly treats `Unhandled` as terminal. The proposed workaround (`Exited` in memory, `"unhandled"` on the wire) would split the model from the protocol.

Closing in favour of the stack.

PR #5813, opened 2026-07-22
fix(replay): Skip buffer-mode replay capture when rate-limited (DART-313)

runningcode:
## 📜 Description

In buffer (on-error) session replay mode, the recorder intentionally keeps running while the SDK is rate-limited so the rolling buffer stays warm. However, when an error triggered `captureReplay`, the strategy still encoded the current segment **and every buffered segment** and handed the resulting envelopes to the transport — which then discarded them because of the active rate limit. That wasted CPU, disk I/O, and `MediaMuxer` file descriptors on replays that could never be sent.

This adds a guard at the top of `BufferCaptureStrategy.captureReplay`: if the `Replay` (or `All`) data category is rate-limited, we skip encoding and capturing entirely. This mirrors the guard session mode already applies via `onRateLimitChanged`/`checkCanRecord`. The buffer keeps rolling cheaply in the background and captures normally once the rate limit expires.

## 💡 Motivation and Context

Follow-up to the session replay resource-leak work in getsentry/sentry-java#5583 and getsentry/sentry-java#5607. Those fixed the `MediaMuxer` FD leak during encoding; this closes the remaining item called out in [getsentry/sentry-dart#3528](<https://github.com/getsentry/sentry-dart/issues/3528>) (getsentry/sentry-dart#3528): while rate-limited, the Android SDK kept creating and encoding segments that were only going to be discarded.

## 💚 How did you test it?

Added a unit test (`captureReplay does nothing when rate-limited`) asserting that, even with a buffered segment present, no segment is sent to the transport and no `replayId` is written to the scope while rate-limited.

## :pencil: Checklist

## 🔮 Next steps

Optionally extend the same guard to the `pause` and `onConfigurationChanged` encode paths in buffer mode if that churn proves significant.

### DIFF
--- a/sentry-android-replay/src/main/java/io/sentry/android/replay/capture/BufferCaptureStrategy.kt
+++ b/sentry-android-replay/src/main/java/io/sentry/android/replay/capture/BufferCaptureStrategy.kt
@@ -19,6 +19,7 @@ import io.sentry.android.replay.capture.CaptureStrategy.Companion.rotateEvents
 import io.sentry.android.replay.capture.CaptureStrategy.ReplaySegment
 import io.sentry.android.replay.util.ReplayRunnable
 import io.sentry.android.replay.util.sample
+import io.sentry.clientreport.DiscardReason.RATELIMIT_BACKOFF
 import io.sentry.protocol.SentryId
 import io.sentry.transport.ICurrentDateProvider
 import io.sentry.util.FileUtils
@@ -78,13 +79,6 @@ internal class BufferCaptureStrategy(
   }
 
   override fun captureReplay(isTerminating: Boolean, onSegmentSent: (Date) -> Unit) {
-    if (isReplayRateLimited()) {
-      // the segment envelopes would be dropped by the transport anyway, so don't waste resources
-      // encoding videos that will only be discarded
-      options.logger.log(INFO, "Replay is rate-limited, not capturing for event")
-      return
-    }
-
     val sampled = random.sample(options.sessionReplay.onErrorSampleRate)
 
     if (!sampled) {
@@ -106,6 +100,18 @@ internal class BufferCaptureStrategy(
         DEBUG,
         "Not capturing replay for crashed event, will be captured on next launch",
       )
+      return
+    }
+
+    if (isReplayRateLimited()) {
+      // the segment envelopes would be dropped by the transport anyway, so don't waste resources
+      // encoding videos that will only be discarded
+      options.logger.log(INFO, "Replay is rate-limited, not capturing for event")
+      // one lost event per flush, not per segment: the transport would have counted the current
+      // segment plus every buffered one, but a flush only ever loses a single replay from the
+      // user's perspective. Under-reporting here is preferable to making replay look like it
+      // dropped data it never held.
+      options.clientReportRecorder.recordLostEvent(RATELIMIT_BACKOFF, Replay)
       return
     }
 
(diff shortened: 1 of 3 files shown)


sentry-android-replay/src/main/java/io/sentry/android/replay/capture/BufferCaptureStrategy.kt
```diff
@@ -76,6 +78,13 @@ internal class BufferCaptureStrategy(
   }
 
   override fun captureReplay(isTerminating: Boolean, onSegmentSent: (Date) -> Unit) {
+    if (isReplayRateLimited()) {
```
romtsn: `h`: I think we're calling this a bit early -- should likely do it after the `if (isTerminating)` check, otherwise we'd be dropping replays for crashed events (which are anyway not being sent in this process, but only on next launch, where the rate limit likely will have already been lifted). 

Another side effect of not setting `isTerminating` would be that the following `convert()` call would convert to session mode and try to keep recording while the process is terminating. Probs, needs a test for this to not regress in the future too

sentry-android-replay/src/main/java/io/sentry/android/replay/capture/BufferCaptureStrategy.kt
```diff
@@ -76,6 +78,13 @@ internal class BufferCaptureStrategy(
...
+      // the segment envelopes would be dropped by the transport anyway, so don't waste resources
+      // encoding videos that will only be discarded
+      options.logger.log(INFO, "Replay is rate-limited, not capturing for event")
+      return
```
romtsn: `m`: another thing -- previously we'd record a client report with the reason `ratelimit_backoff` on the envelope being dropped, but now we're not doing it. I think after addressing the point above (moving this to after `if (isTerminating)`, we could also record the `ratelimit_backoff` client report for a genuinely lost event (which would be also called after sampling) inside the `if (isReplayRateLimited())` block.

romtsn (approved):
pre-approving, but 2 things I'd want to have addressed

sentry-android-replay/src/main/java/io/sentry/android/replay/capture/BufferCaptureStrategy.kt
runningcode: Good point!

sentry-android-replay/src/main/java/io/sentry/android/replay/capture/BufferCaptureStrategy.kt
```diff
@@ -109,6 +103,18 @@ internal class BufferCaptureStrategy(
...
 
+    if (isReplayRateLimited()) {
+      // the segment envelopes would be dropped by the transport anyway, so don't waste resources
+      // encoding videos that will only be discarded
+      options.logger.log(INFO, "Replay is rate-limited, not capturing for event")
+      // one lost event per flush, not per segment: the transport would have counted the current
+      // segment plus every buffered one, but a flush only ever loses a single replay from the
+      // user's perspective. Under-reporting here is preferable to making replay look like it
+      // dropped data it never held.
+      options.clientReportRecorder.recordLostEvent(RATELIMIT_BACKOFF, Replay)
```
runningcode, line 114: @romtsn is this what you were thinking? Is it okay that we only report 1 lost event here even if there are more in the buffer? This is something my clanker said was a possibility.

romtsn: yeah, theoretically we could've used `bufferedSegments.size + 1` here, but I agree it's better to under-report because for users it looks like a single replay lost, so let's keep it as-is

romtsn (approved):
double-checked everything, it all makes sense, approving!

PR #5814, opened 2026-07-22
perf: Schedule rate-limit notifications on shared executor (JAVA-653)

runningcode:
## :scroll: Description

`RateLimiter` created a `java.util.Timer` (a dedicated thread) whose thread stayed alive for the rest of the process once the SDK got rate limited. The "rate limit lifted" observer notification now runs on the shared timer executor (`SentryOptions#getTimerExecutorService`), already used for transaction timeouts, whose single worker thread is reused and self-terminates when idle.

- The notification is scheduled with the delay until the limit expires; to avoid an extra `getCurrentTimeMillis()` call the already-computed `retryAfterMillis` is passed through as the delay.
- Pending notification futures are tracked (completed ones pruned on each schedule) and cancelled on `close()`, preserving the old `Timer#cancel()` semantics.

Once an app got rate limited, this timer thread lived forever. The shared timer executor's worker is reused and idles out.

Existing `RateLimiterTest`, adapted from the Timer-mock verification to the executor/future model, plus `AsyncHttpTransportTest`.

## :pencil: Checklist

## :crystal_ball: Next steps

Related PRs in this effort: LifecycleWatcher (#5819), performance collector (#5816), HostnameCache (#5817), batch processors (#5818).

### DIFF
--- a/sentry/src/main/java/io/sentry/transport/RateLimiter.java
+++ b/sentry/src/main/java/io/sentry/transport/RateLimiter.java
@@ -23,12 +23,13 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Date;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.Timer;
-import java.util.TimerTask;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.Future;
+import java.util.concurrent.RejectedExecutionException;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
@@ -42,8 +43,9 @@ public final class RateLimiter implements Closeable {
   private final @NotNull Map<DataCategory, @NotNull Date> sentryRetryAfterLimit =
       new ConcurrentHashMap<>();
   private final @NotNull List<IRateLimitObserver> rateLimitObservers = new CopyOnWriteArrayList<>();
-  private @Nullable Timer timer = null;
-  private final @NotNull AutoClosableReentrantLock timerLock = new AutoClosableReentrantLock();
+  private final @NotNull List<Future<?>> notifyObserversFutures = new ArrayList<>();
+  private final @NotNull AutoClosableReentrantLock notifyFuturesLock =
+      new AutoClosableReentrantLock();
 
   public RateLimiter(
       final @NotNull ICurrentDateProvider currentDateProvider,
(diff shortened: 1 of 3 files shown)


sentry/src/test/java/io/sentry/transport/RateLimiterTest.kt
```diff
@@ -66,6 +66,8 @@ class RateLimiterTest {
 
     fun getSUT(): RateLimiter {
       val options = SentryOptions().apply { setLogger(NoOpLogger.getInstance()) }
+      // a real executor so scheduled rate-limit-lifted notifications actually run
+      options.setTimerExecutorService(SentryExecutorService(options))
```
romtsn: do we need to close it in teardown perhaps, so it doesn't leak across tests?

sentry/src/main/java/io/sentry/transport/RateLimiter.java
```diff
@@ -312,19 +315,25 @@ private void applyRetryAfterOnlyIfLonger(
 
       notifyRateLimitObservers();
 
-      try (final @NotNull ISentryLifecycleToken ignored = timerLock.acquire()) {
-        if (timer == null) {
-          timer = new Timer(true);
+      // notify observers again once the rate limit is lifted, using the shared timer executor
+      // instead of a dedicated Timer thread
+      try (final @NotNull ISentryLifecycleToken ignored = notifyFuturesLock.acquire()) {
```
romtsn, line 320: `l`: unsure if we need to protect this with checking for `instanceof NoOpSentryExecutorService` to avoid allocating FutureTasks for nothing (`isDone` also returns `false` when no-op, so pruning wouldn't do anything), but we're not really doing that anywhere else I believe, so I'm fine with keeping it as-is.

romtsn (approved):
couple of nits, but LGTM 🎉

sentry/src/main/java/io/sentry/transport/RateLimiter.java
runningcode: Yeah it is a good point but also quite an edge case. Since we aren't doing it anywhere else, I don't think we should do it here.

sentry/src/test/java/io/sentry/transport/RateLimiterTest.kt
runningcode: Good point!

sentry/src/main/java/io/sentry/transport/RateLimiter.java
```diff
@@ -312,19 +315,25 @@ private void applyRetryAfterOnlyIfLonger(
...
+        while (iterator.hasNext()) {
+          if (iterator.next().isDone()) {
+            iterator.remove();
+          }
+        }
+        try {
+          notifyObserversFutures.add(
+              options
+                  .getTimerExecutorService()
+                  .schedule(this::notifyRateLimitObservers, delayMillis));
```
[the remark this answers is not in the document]
runningcode: Created https://github.com/getsentry/sentry-java/pull/5851 to follow up on this

PR #5817, opened 2026-07-22
perf: Let HostnameCache worker thread idle out (JAVA-653)

runningcode:
## :scroll: Description

`HostnameCache`'s executor stays around for the entire process lifetime (that's how `ThreadPoolExecutors` work by default) even though the cache only refreshes once every 5 hours. It now uses a `ThreadPoolExecutor` (core = max = 1) with a 30s keep-alive and `allowCoreThreadTimeOut(true)`, so the worker thread self-terminates when idle and is recreated on the next refresh.

Alternatively, we can also use an existing executor for this but this is the smallest change here. Let me know if you think we should go that route.

There are some new unit tests, but otherwise relying on ART/JVM to do their thing correctly.

## :pencil: Checklist

### DIFF
--- a/sentry/src/main/java/io/sentry/HostnameCache.java
+++ b/sentry/src/main/java/io/sentry/HostnameCache.java
@@ -6,9 +6,10 @@ import java.net.InetAddress;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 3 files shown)


sentry/src/main/java/io/sentry/HostnameCache.java
```diff
@@ -87,6 +90,18 @@ private HostnameCache() {
...
+    // A single thread executor whose worker thread times out while idle, so no thread is kept
+    // alive between the infrequent cache refreshes.
+    final @NotNull ThreadPoolExecutor executor =
+        new ThreadPoolExecutor(
+            1,
+            1,
+            THREAD_KEEP_ALIVE_SECONDS,
+            TimeUnit.SECONDS,
+            new LinkedBlockingQueue<>(),
+            new HostnameCacheThreadFactory());
```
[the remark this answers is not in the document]
runningcode: This can only happen if the SDK is shut down and not a regression from this PR.

sentry/src/main/java/io/sentry/HostnameCache.java
```diff
@@ -34,6 +35,9 @@ public final class HostnameCache {
   /** Time before the get hostname operation times out (in ms). */
   private static final long GET_HOSTNAME_TIMEOUT = TimeUnit.SECONDS.toMillis(1);
 
+  /** How long the worker thread may stay idle before it self-terminates. */
+  private static final long THREAD_KEEP_ALIVE_SECONDS = 30;
```
0xadam-brown, line 39: l: Do we want to reduce this to, say, 5 or 10 seconds? 

AFAICT, we'll only need this for try-after-failure scenarios. After failure, handleCacheUpdateFailure() sets the retry window to 1 second, meaning the next retry is likely almost immediate after that 1 second expires. 

Obviously not a blocker b/c 30s is much better than the process lifetime.

0xadam-brown (approved):
Nice! One comment, no blockers 👍

sentry/src/main/java/io/sentry/HostnameCache.java
runningcode: I think it won't make a big difference so I don't want to wait on CI again and delay the release.

PR #5819, opened 2026-07-22
perf: Schedule session end on shared executor (JAVA-653)

runningcode:
## :scroll: Description

`LifecycleWatcher` created a `java.util.Timer` (a dedicated thread) on the first app-background transition, and that thread lived for the rest of the process. The session-end task now runs on the shared timer executor (`SentryOptions#getTimerExecutorService`), already used for transaction timeouts, whose single worker thread is reused and self-terminates when idle.

The task is scheduled/cancelled as the app moves between background and foreground via `Future#cancel(false)` (the same pattern `SentryTracer` uses for idle/deadline timeouts). If scheduling fails (executor already shut down), the session is ended immediately instead of being left open.

On Android this timer thread was created on the first background transition and never went away. Reusing the shared timer executor removes it.

Existing `LifecycleWatcherTest` (updated to assert on the scheduled `Future` instead of the removed `Timer`/`TimerTask` internals) and `SessionTrackingIntegrationTest`. Both fixtures now wire a real executor into options so scheduled tasks run.

## :pencil: Checklist

## :crystal_ball: Next steps

Related PRs in this effort: RateLimiter (#5814), performance collector (#5816), HostnameCache (#5817), batch processors (#5818).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Performance
+
+- Reduce the number of SDK threads: `LifecycleWatcher` now schedules the session-end task on the shared timer executor instead of creating a dedicated `java.util.Timer` thread ([#5819](https://github.com/getsentry/sentry-java/pull/5819))
+
 ## 8.50.0
 
 ### Android 17 support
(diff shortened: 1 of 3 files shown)


romtsn (approved):
nice!

PR #5826, opened 2026-07-23
fix(replay): Dispatch session-deadline stop off the replay worker thread

runningcode:
## :scroll: Description

When a session replay reaches its duration deadline (default 1h), the resulting `stop()` was invoked inline from the replay worker thread, inside the `add_frame` task in `SessionCaptureStrategy.onScreenshotRecorded`. Because `ReplayExecutorService.submit()` runs tasks synchronously when the caller is already on a `SentryReplayIntegration-*` thread, `stop()` encoded the final video segment and recursively deleted the replay cache **while holding the replay lifecycle lock** (`AutoClosableReentrantLock`).

A foreground `start()` (`LifecycleWatcher.onForeground` → `startSession`) runs on the **main thread** and must acquire that same lock, so it parked long enough to trigger a background ANR.

This change dispatches the deadline `stop()` through `options.executorService` (a non-worker thread). `stop()` then submits the segment encoding **asynchronously** onto the single-threaded replay executor (preserving ordering), and the lifecycle lock is held only for the fast state transition — matching the existing timer-executor `endSession` stop path.

Fixes the background ANR reported in getsentry/sentry-dart#3556 (Linear: DART-323), whose stack shows the main thread parked on the replay lifecycle lock in `ReplayIntegration.start()` on foreground. The deadline stop re-entering the lifecycle lock from the worker thread is the only path where the segment encode runs synchronously under the lock; every other `start`/`stop`/`pause`/`resume` caller already submits the encode asynchronously.

- GH Issue: getsentry/sentry-dart#3556
- Linear: DART-323

Added a regression test in `SessionCaptureStrategyTest` using a deferred executor that asserts the deadline `stop()` is dispatched off the calling thread (not run inline) and only runs once the executor drains. Updated the existing deadline test's fixture to use an immediate executor so it still exercises the dispatched stop.

## :pencil: Checklist

## :crystal_ball: Next steps

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixes
 
+- Fix background ANR when a session replay reaches its duration deadline: the resulting `stop()` (and its segment encoding) no longer runs on the replay worker thread while holding the replay lifecycle lock, which could block a foreground `start()` on the main thread ([#5826](https://github.com/getsentry/sentry-java/pull/5826))
 - Pin the published Sentry Android SDK's AAR metadata `minCompileSdk` to our `minSdk` (`21`) instead of AGP 9's new default of the SDK's own `compileSdk` (`37`), so apps that depend on the SDK aren't forced to raise their `compileSdk` ([#5823](https://github.com/getsentry/sentry-java/pull/5823))
 
 ### Performance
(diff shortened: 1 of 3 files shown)


romtsn (approved):
looks good!

PR #5827, opened 2026-07-23
feat(feedback): Support runtime enable/disable of shake-to-report

### LINKED ISSUE #5486: Support runtime enable/disable of shake-to-report
Sentry's configuration is set synchronously at `init()`. Customers who fetch a feature flag asynchronously at startup (e.g. to restrict feedback to internal users only) cannot use that flag to toggle shake-to-report on the very first session — the config has already been applied before the async result returns.

## Request

Expose a runtime API to enable or disable the shake-to-report / user feedback feature after initialization, so it can be toggled dynamically based on async signals like feature flags or user roles.

Action taken on behalf of Angelo de Voer.

markushi:
## 📜 Description

Adds a runtime API to toggle shake-to-report after `Sentry.init()`:

* `Sentry.feedback().enableOnShake()`
* `Sentry.feedback().disableOnShake()`
* `Sentry.feedback().isOnShakeEnabled()`

`SentryFeedbackOptions.isUseShakeGesture()` now only determines the initial state; the runtime API overrides it afterwards. Propagation works similar to `IFormHandler` / `ReplayController` by introducing a new `SentryFeedbackOptions.IShakeController`, implemented by `FeedbackShakeIntegration` on Android and by a logging no-op elsewhere.

This also required re-wiring on-demand dialogs vs. the automatic ones, so that no duplicate dialogs are shown and no duplicate detectors run at the same time:

* `FeedbackShakeIntegration` is always installed and registers itself as the `IShakeController`, even when the option starts out disabled — otherwise there would be nothing to enable later. It arms/disarms the detector in `enableOnShake()` / `disableOnShake()`.
* Any visible form reports itself and its host activity to the integration (`onFormVisible` in `onStart`, `onFormGone` in `onStop`, with `onDetachedFromWindow` as a safety net for teardown without a dismiss). Detection is suppressed for the activity hosting a form and only re-armed once no form remains there, so neither a form on a backgrounded activity nor one of two simultaneously visible forms can re-enable detection under a visible one. This replaces the previous `onFormClose` swap-and-restore dance, which was fragile and could leak the host activity.
* `SentryUserFeedbackForm` only starts a detector of its own for the per-form opt-in case (`useShakeGesture` enabled via a configurator while it is globally off); otherwise it defers to the integration, including if shake-to-report gets enabled globally between form creation and the shake.
* User-supplied `onFormOpen` / `onFormClose` / `onSubmitSuccess` / `onSubmitError` callbacks are now guarded, so a crashing callback can't take down the app or skip SDK-side cleanup.

## 💡 Motivation and Context

Having a runtime option to enable / disable the feedback.  <br>Closes getsentry/sentry-java#5486 (getsentry/sentry-java#5486)

## 💚 How did you test it?

Unit tests (`FeedbackShakeIntegrationTest`, `SentryUserFeedbackFormTest`, `FeedbackApiTest`, `SentryOptionsTest`) + manual testing with the Android sample app.

## :pencil: Checklist

## 🔮 Next steps

* Docs update for the new runtime API (docs repo)


sentry-android-core/src/main/java/io/sentry/android/core/SentryUserFeedbackForm.java
```diff
@@ -62,7 +62,9 @@ public class SentryUserFeedbackForm extends AlertDialog {
   private void maybeStartShakeDetection(final @NotNull Context context) {
     final @NotNull SentryFeedbackOptions globalFeedbackOptions =
         Sentry.getCurrentScopes().getOptions().getFeedbackOptions();
-    if (!resolvedFeedbackOptions.isUseShakeGesture() || globalFeedbackOptions.isUseShakeGesture()) {
+
+    if (!resolvedFeedbackOptions.isUseShakeGesture()
+        || globalFeedbackOptions.getShakeController().isEnabled()) {
```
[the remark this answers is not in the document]
markushi: resolved by consolidating to a single shake detector running at all times

sentry/src/main/java/io/sentry/IFeedbackApi.java
```diff
@@ -15,6 +16,29 @@ void show(
       final @Nullable SentryId associatedEventId,
       final @Nullable SentryFeedbackOptions.OptionsConfigurator configurator);
 
+  /**
+   * Enables showing the feedback form when a shake gesture is detected, overriding {@link
+   * SentryFeedbackOptions#isUseShakeGesture()}. Only supported on Android; no-op on other
```
runningcode, line 20: would it be possible to provide different interfaces to different platforms?

sentry-android-core/src/main/java/io/sentry/android/core/SentryUserFeedbackForm.java
```diff
@@ -332,14 +368,44 @@ protected void onStart() {
...
     final @NotNull SentryFeedbackOptions feedbackOptions = options.getFeedbackOptions();
+    // Pause shake-to-report while this form is visible, so a shake can't stack a second
+    // form on top of it
+    feedbackOptions.getShakeController().pauseDetection(true);
     final @Nullable Runnable onFormOpen = feedbackOptions.getOnFormOpen();
     if (onFormOpen != null) {
-      onFormOpen.run();
+      try {
+        onFormOpen.run();
+      } catch (Throwable e) {
```
runningcode: nit: same question as above

sentry-android-core/src/main/java/io/sentry/android/core/SentryUserFeedbackForm.java
```diff
@@ -310,7 +338,15 @@ public void setOnDismissListener(final @Nullable OnDismissListener listener) {
     if (onFormClose != null) {
       super.setOnDismissListener(
           dialog -> {
-            onFormClose.run();
+            // User-provided callback: a crash in it must not take down the app or skip the
+            // cleanup and the user's own dismiss listener below
+            try {
+              onFormClose.run();
+            } catch (Throwable e) {
```
runningcode: even if it is a user provided callback, i don't think this is a good idea. a user-provided callback could still trigger a stackoverflow or OOMs

sentry-android-core/src/main/java/io/sentry/android/core/FeedbackShakeIntegration.java
```diff
@@ -81,34 +103,34 @@ public void register(final @NotNull IScopes scopes, final @NotNull SentryOptions
...
+  @Override
+  public void pauseDetection(final boolean paused) {
+    this.paused = paused;
+  }
+
+  @Override
+  public void close() throws IOException {
+    disable();
+  }
+
```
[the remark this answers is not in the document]
runningcode: This seems legit, why did we remove that from the `onActivityResumed` ?

runningcode (commented):
Looks good, just some small things!

sentry/src/main/java/io/sentry/IFeedbackApi.java
```diff
@@ -15,6 +16,29 @@ void show(
...
+   */
+  void disableFeedbackOnShake();
+
+  /**
+   * Whether showing the feedback form on a shake gesture is currently enabled. Always {@code false}
+   * on non-Android platforms.
+   *
+   * @return true if the feedback form is shown when a shake gesture is detected
+   */
+  @ApiStatus.Internal
```
romtsn: `m`: since we're using this one in the sample app (and I imagine there's a use-case for this for customers too), shall we unmark it as Internal and make it public?

sentry/src/main/java/io/sentry/IFeedbackApi.java
romtsn: I think we usually keep a single interface and reserve a right to implement it for JVM in the future (e.g. for compose desktop) :)

sentry-android-core/src/main/java/io/sentry/android/core/SentryUserFeedbackForm.java
markushi: Fixed in e5d6f39 — narrowed to `catch (Exception)` so `Error`s (StackOverflow/OOM) propagate. Applied to all four callback guards, including `onSubmitSuccess`/`onSubmitError` which had the same problem.

sentry-android-core/src/main/java/io/sentry/android/core/SentryUserFeedbackForm.java
markushi: Same fix as the thread below — narrowed to `catch (Exception)` in e5d6f39.

sentry/src/main/java/io/sentry/IFeedbackApi.java
markushi: Going with a single interface as @romtsn suggests, so the JVM side stays open for e.g. Compose Desktop later.

sentry/src/main/java/io/sentry/IFeedbackApi.java
markushi: Done in e5d6f39 — dropped `@ApiStatus.Internal` and renamed it to `isOnShakeEnabled()` so it matches `enableOnShake()`/`disableOnShake()`.

sentry-android-core/src/main/java/io/sentry/android/core/FeedbackShakeIntegration.java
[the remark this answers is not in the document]
markushi: Fixed in e5d6f39, but by scoping the pause to the activity the form is showing on rather than restoring the `onActivityResumed` cleanup, which would have cleared the pause when returning to an activity whose dialog is still visible.

sentry-android-core/src/main/java/io/sentry/android/core/SentryUserFeedbackForm.java
```diff
@@ -60,9 +60,14 @@ public class SentryUserFeedbackForm extends AlertDialog {
...
   private void maybeStartShakeDetection(final @NotNull Context context) {
+    // Only an explicit per-form opt-in starts a detector for this form. When shake is
+    // configured globally (via the option or the runtime toggle), FeedbackShakeIntegration
+    // already shows a form on shake and this form defers to it.
     final @NotNull SentryFeedbackOptions globalFeedbackOptions =
         Sentry.getCurrentScopes().getOptions().getFeedbackOptions();
-    if (!resolvedFeedbackOptions.isUseShakeGesture() || globalFeedbackOptions.isUseShakeGesture()) {
+    if (!resolvedFeedbackOptions.isUseShakeGesture()
+        || globalFeedbackOptions.isUseShakeGesture()
+        || globalFeedbackOptions.getShakeController().isEnabled()) {
```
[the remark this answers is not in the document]
markushi: Intentional, keeping as is. `resolvedFeedbackOptions` is a copy of the global options, so when the global `isUseShakeGesture` is true its per-form value is true as well — it is indistinguishable from a genuine per-form opt-in unless a configurator explicitly set it.

That means dropping the `isUseShakeGesture()` term would make every form shown after `disableOnShake()` start its own detector, resurrecting the feature the user just turned off (and leaking a detector + HandlerThread per `show()`, since the per-form detector is only torn down on activity destroy). An explicit runtime disable winning over an inherited flag is the behaviour we want here.

romtsn (approved):

sentry-android-core/src/main/java/io/sentry/android/core/SentryUserFeedbackForm.java
```diff
@@ -332,14 +370,44 @@ protected void onStart() {
...
+  public void onDetachedFromWindow() {
+    super.onDetachedFromWindow();
+    // Safety net for teardown without a dismiss (e.g. the host activity is destroyed while the
+    // form is still showing): onStop never fires then, but the window is still detached —
+    // without this, shake-to-report would stay paused forever.
+    Sentry.getCurrentScopes()
+        .getOptions()
+        .getFeedbackOptions()
+        .getShakeController()
+        .setOnShakePaused(false);
```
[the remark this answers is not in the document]
markushi: The race as described is not reachable: `Dialog.dismissDialog()` runs `removeViewImmediate(mDecor)` (which dispatches `onDetachedFromWindow` synchronously) and then `onStop()` back-to-back on the main thread, before the dismiss listener is even sent — nothing can show a form in between, since that needs the main thread too.

The adjacent case *is* real though, without any timing race: two forms can be visible at once (e.g. `Sentry.feedback().showForm()` while a form is already up — the integration only suppresses shake-triggered forms). Tearing down either one used to clear the single shared ref and re-arm detection under the still-visible one.

Fixed in 99dd1c5 by tracking visible forms individually: each form reports itself plus its host activity, and detection is only re-armed once no form remains on the current activity. Reporting the same form gone twice (onStop + onDetachedFromWindow) is now a no-op the second time. Covered by `dismissing one of two visible forms keeps detection suppressed` and `reporting the same form gone twice re-arms detection only once`.

sentry-android-core/src/main/java/io/sentry/android/core/FeedbackShakeIntegration.java
```diff
@@ -81,34 +108,58 @@ public void register(final @NotNull IScopes scopes, final @NotNull SentryOptions
...
+  public boolean isOnShakeEnabled() {
+    return enabled;
+  }
+
+  @Override
+  public void setOnShakePaused(final boolean paused) {
+    if (paused) {
+      final @Nullable Activity activity = CurrentActivityHolder.getInstance().getActivity();
+      formActivityRef = activity == null ? null : new WeakReference<>(activity);
+      stopShakeDetection();
```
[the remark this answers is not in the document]
markushi: Good catch, this was real. `IShakeController` lives in the `sentry` core module and cannot reference `android.app.Activity`, which is why it reached for `CurrentActivityHolder` in the first place.

Fixed in 99dd1c5 by moving pausing off the cross-platform interface: `SentryUserFeedbackForm` (same module as the integration) now calls `FeedbackShakeIntegration.onFormVisible(host, form)` / `onFormGone(form)` directly and passes its own host activity, so suppression can never be attributed to the wrong activity. `IShakeController` keeps only enable/disable/isEnabled. Covered by `a form reports the activity it is showing on, not the current one`.

sentry-android-core/src/main/java/io/sentry/android/core/SentryUserFeedbackForm.java
```diff
@@ -60,9 +60,12 @@ public class SentryUserFeedbackForm extends AlertDialog {
...
   private void maybeStartShakeDetection(final @NotNull Context context) {
+    // Only start shake detection if it's enabled within the options,
+    // and not already running globally
     final @NotNull SentryFeedbackOptions globalFeedbackOptions =
         Sentry.getCurrentScopes().getOptions().getFeedbackOptions();
-    if (!resolvedFeedbackOptions.isUseShakeGesture() || globalFeedbackOptions.isUseShakeGesture()) {
+    if (!resolvedFeedbackOptions.isUseShakeGesture()
+        || globalFeedbackOptions.getShakeController().isOnShakeEnabled()) {
       return;
     }
```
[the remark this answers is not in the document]
markushi: Yes that's by cesign, as the those dialogs have an option configuration block, they can disable it anyway, so that's fine.

PR #5828, opened 2026-07-23
feat(feedback): Add screenshot attachment button to user feedback widget

### LINKED ISSUE #5485: Add file attachment support to user feedback widget
The user feedback widget (shake-to-report) has no built-in file-picker UI. Users who want to attach screenshots or screen recordings must implement a custom file-picker on top of the existing attachments API, adding development effort and ongoing maintenance overhead.

## Request

Add a native file-picker / gallery-access component directly into the feedback modal, so users can attach screenshots and recordings out of the box without any custom UI work.

## Reference

https://docs.luciq.ai/product-guides-and-integrations/product-guides/bug-reporting/report-types-and-content

Action taken on behalf of Angelo de Voer.

markushi:
## 📜 Description

Adds an "Add a screenshot" button to the Android user feedback form that opens the  <br>androidx photo picker (`ActivityResultContracts.PickVisualMedia`, images only) and sends  <br>the picked image as an attachment in the same envelope as the feedback.

https://github.com/user-attachments/assets/46d9e1f5-3610-424d-b225-f4195a19a592

image

* New `SentryFeedbackOptions`: `enableAttachScreenshot` (default `true`, matching the JS SDK),  <br>`addScreenshotButtonLabel`, `removeScreenshotButtonLabel`, `screenshotTooLargeMessageText`, plus the  <br>`io.sentry.feedback.enable-attach-screenshot` manifest flag
* `androidx.activity:activity:1.8.2` is added as **compileOnly** — no new transitive  <br>dependency; the button is hidden (with a warning log) when androidx.activity is missing  <br>or the host Activity is not a `ComponentActivity`
* The image is size-checked against `maxAttachmentSize` at pick time

### Fixed along the way

The form was not scrollable, so in landscape or at large font scales the send and cancel<br>buttons were pushed off screen and the dialog could not be dismissed. It is now wrapped in<br>a `ScrollView`, and the three full width buttons were replaced with the framework dialog<br>button bar (cancel and send in a right aligned bar, screenshot as a borderless full width<br>action). No changelog entry, as it is bundled with this feature.

## 💡 Motivation and Context

* resolves: getsentry/sentry-java#5485

## 💚 How did you test it?

Unit tests + Manual testing.

## :pencil: Checklist

- [ ] Public API changes reviewed by another Mobile SDK team member or implemented according to the [develop docs](<https://develop.sentry.dev/>) spec.


sentry-android-core/src/main/java/io/sentry/android/core/SentryUserFeedbackForm.java
```diff
@@ -340,6 +386,126 @@ protected void onStart() {
...
+              () ->
+                  readUriBytes(
+                      resolver,
+                      imageUri,
+                      Sentry.getCurrentScopes().getOptions().getMaxAttachmentSize()),
+              "screenshot." + ext,
+              mime,
+              "event.attachment",
+              false));
+    } catch (Throwable t) {
```
runningcode, line 549: I decided I'm always going to comment whenever we add new usages of this. Sorry to bring it up again.

1. How will we know that this feature is working in production if we swallow all errors?
2. Since we're dealing with images, files and attachments how we we know we aren't swallowing a permissions issue or OOM?

sentry-android-core/src/main/java/io/sentry/android/core/SentryUserFeedbackForm.java
```diff
@@ -340,6 +386,126 @@ protected void onStart() {
...
+
+  private static long getUriSize(final @NotNull ContentResolver resolver, final @NotNull Uri uri) {
+    try (final @Nullable Cursor cursor = resolver.query(uri, null, null, null, null)) {
+      if (cursor != null && cursor.moveToFirst()) {
+        final int sizeIndex = cursor.getColumnIndex(OpenableColumns.SIZE);
+        if (sizeIndex != -1 && !cursor.isNull(sizeIndex)) {
+          return cursor.getLong(sizeIndex);
+        }
+      }
+    } catch (Throwable ignored) {
```
runningcode: same comment here as above.

sentry/src/main/java/io/sentry/util/FileUtils.java
```diff
@@ -100,12 +102,39 @@ public static byte[] readBytesFromFile(String pathname, long maxFileLength)
...
+        final ByteArrayOutputStream outputStream = new ByteArrayOutputStream()) {
       byte[] bytes = new byte[1024];
       int length;
       int offset = 0;
+      long totalLength = 0;
       while ((length = inputStream.read(bytes)) != -1) {
+        totalLength += length;
+        if (totalLength >= maxLength) {
+          throw new IOException(
+              String.format(
```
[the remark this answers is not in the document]
runningcode: Seems legit

sentry-android-core/build.gradle.kts
```diff
@@ -115,6 +118,7 @@ dependencies {
   testImplementation(libs.androidx.test.ext.junit)
   testImplementation(libs.androidx.test.runner)
   testImplementation(libs.awaitility.kotlin)
+  testImplementation(libs.google.truth)
```
runningcode, line 121: yes for using google truth!

sentry-android-core/src/main/java/io/sentry/android/core/SentryUserFeedbackForm.java
```diff
@@ -39,6 +50,15 @@ public class SentryUserFeedbackForm extends AlertDialog {
   private @Nullable SentryShakeDetector shakeDetector;
   private @Nullable Application.ActivityLifecycleCallbacks shakeLifecycleCallbacks;
 
+  private @NotNull LoadClass loadClass = new LoadClass();
+  private @Nullable SentryFeedbackPhotoPicker photoPicker;
+  private @Nullable Uri selectedImageUri;
+
+  // for testing
+  void setLoadClass(final @NotNull LoadClass loadClass) {
```
runningcode: can we pass this in to the constructor instead of having a dedicated `setLoadClass` method?

sentry-android-core/src/main/java/io/sentry/android/core/SentryUserFeedbackForm.java
```diff
@@ -191,6 +211,29 @@ protected void onCreate(Bundle savedInstanceState) {
...
+
+    // The button is made visible in onStart, once the photo picker is registered successfully
+    btnAddScreenshot.setVisibility(View.GONE);
+    btnAddScreenshot.setOnClickListener(
+        v -> {
+          if (selectedImageUri == null) {
+            if (photoPicker != null) {
+              try {
+                photoPicker.launch();
+              } catch (Throwable t) {
```
runningcode, line 239: gonna make the same comment that I did for the other usages of `Throwable`

sentry-android-core/src/test/java/io/sentry/android/core/SentryUserFeedbackFormTest.kt
```diff
@@ -59,15 +77,22 @@ class SentryUserFeedbackFormTest {
       associatedEventId: SentryId? = null,
       configuration: SentryUserFeedbackForm.OptionsConfiguration? = null,
       configurator: SentryFeedbackOptions.OptionsConfigurator? = null,
+      context: Context = application,
+      loadClass: LoadClass? = null,
```
runningcode: when would this be null? can we make it `NotNull`?

sentry-android-core/src/main/java/io/sentry/android/core/SentryFeedbackPhotoPicker.java
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * Launches the androidx photo picker to attach an image to user feedback. All androidx.activity
+ * references are isolated in this class, so it must only be loaded after {@link #isAvailable} has
+ * returned true.
+ */
+final class SentryFeedbackPhotoPicker {
+
+  interface OnImagePicked {
+    void onImagePicked(@Nullable Uri uri);
```
runningcode: what does it mean when the `uri` is null in this interface?

sentry/src/main/java/io/sentry/SentryFeedbackOptions.java
```diff
@@ -256,6 +279,28 @@ public void setUseShakeGesture(final boolean useShakeGesture) {
...
   }
 
+  /**
+   * Displays a button on the feedback form that allows the user to attach a screenshot from the
+   * device gallery. Only supported on Android, and requires the androidx.activity dependency and a
+   * ComponentActivity host; the button is hidden otherwise. Defaults to true.
+   *
+   * @return true if the attach screenshot button is shown
+   */
+  public boolean isEnableScreenshot() {
```
runningcode: do you think this would be a better name?
```suggestion
  public boolean isAttachScreenshotEnabled() {
```

sentry/src/main/java/io/sentry/SentryFeedbackOptions.java
```diff
@@ -256,6 +279,28 @@ public void setUseShakeGesture(final boolean useShakeGesture) {
...
+  }
+
+  /**
+   * Sets whether the attach screenshot button is shown on the feedback form. Only supported on
+   * Android, and requires the androidx.activity dependency and a ComponentActivity host; the button
+   * is hidden otherwise. Defaults to true.
+   *
+   * @param enableScreenshot true if the attach screenshot button is shown
+   */
+  public void setEnableScreenshot(final boolean enableScreenshot) {
```
runningcode:
```suggestion
  public void setPhotoAttachmentEnabled(final boolean photoAttachmentEnabled) {
```

sentry-android-core/src/test/java/io/sentry/android/core/SentryUserFeedbackFormTest.kt
```diff
@@ -59,15 +77,22 @@ class SentryUserFeedbackFormTest {
...
     ): SentryUserFeedbackForm =
-      SentryUserFeedbackForm(application, 0, associatedEventId, configuration, configurator)
+      SentryUserFeedbackForm(context, 0, associatedEventId, configuration, configurator).also {
+        if (loadClass != null) {
```
runningcode: when would this be null? if it is null, would the tests still be effective?

runningcode (commented):
Thanks for adding this!

sentry-android-core/src/main/java/io/sentry/android/core/SentryUserFeedbackForm.java
markushi: Yeah good point, I've moved into the options, so it's much easier to test and got rid of the field nullability

sentry/src/main/java/io/sentry/SentryFeedbackOptions.java
markushi: Good point, I was on the verge of naming it differently, but then sticked to `enableScreenshot` from JS: https://github.com/getsentry/sentry-javascript/blob/110f6c443cc660f589341fce2004f068332a1010/packages/feedback/src/core/integration.ts#L96C5-L96C21

I'll rename it anyway, also use "screenshot" as the term everywhere (even if we're using a "photo"-picker).

sentry-android-core/src/main/java/io/sentry/android/core/SentryFeedbackPhotoPicker.java
markushi: The Uri is `null` if an app user decides to bail and does not select any screenshot. I've moved this check further up, and made it `@NotNull` within our consuming paths downstream.

sentry-android-core/src/main/java/io/sentry/android/core/SentryUserFeedbackForm.java
```diff
@@ -340,6 +379,125 @@ protected void onStart() {
...
+    if (screenshotPicker != null) {
+      btnAddScreenshot.setVisibility(View.VISIBLE);
+    } else {
+      btnAddScreenshot.setVisibility(View.GONE);
+      options
+          .getLogger()
+          .log(
+              SentryLevel.WARNING,
+              "Feedback screenshot button won't be shown. It requires the androidx.activity "
+                  + "dependency and the feedback form being shown from a ComponentActivity.");
```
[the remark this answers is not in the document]
runningcode: oh the android lifecycle!

sentry-android-core/src/main/java/io/sentry/android/core/SentryUserFeedbackForm.java
markushi: Switched to using the newly added `ExceptionUtils.rethrowIfFatal`, so OOM and linkage errors propagate. `getUriSize` now logs the throwable too, it was silent before.

sentry-android-core/src/main/java/io/sentry/android/core/SentryUserFeedbackForm.java
markushi: Switched to using the newly added `ExceptionUtils.rethrowIfFatal`, so OOM and linkage errors propagate. `getUriSize` now logs the throwable too, it was silent before.

sentry-android-core/src/main/java/io/sentry/android/core/SentryUserFeedbackForm.java
markushi: Switched to using the newly added `ExceptionUtils.rethrowIfFatal`, so OOM and linkage errors propagate. `getUriSize` now logs the throwable too, it was silent before.

sentry-android-core/src/main/java/io/sentry/android/core/SentryUserFeedbackForm.java
```diff
@@ -340,6 +379,125 @@ protected void onStart() {
...
+          new Attachment(
+              () ->
+                  readUriBytes(
+                      resolver,
+                      imageUri,
+                      Sentry.getCurrentScopes().getOptions().getMaxAttachmentSize()),
+              "screenshot." + ext,
+              mime,
+              "event.attachment",
+              false));
```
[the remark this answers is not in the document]
markushi: The picker grant holds for the process lifetime and the provider runs in-process before caching, so this window doesn't open. A failed read drops just the attachment, logged at `ERROR` by `JsonSerializer`.

sentry-android-core/src/main/java/io/sentry/android/core/SentryFeedbackScreenshotPicker.java
```diff
@@ -0,0 +1,61 @@
...
+            .getActivityResultRegistry()
+            .register(
+                "sentry_user_feedback_screenshot_picker",
+                new ActivityResultContracts.PickVisualMedia(),
+                (@Nullable Uri uri) -> {
+                  if (uri != null) {
+                    callback.onScreenshotPicked(uri);
+                  }
+                });
+    return new SentryFeedbackScreenshotPicker(launcher);
```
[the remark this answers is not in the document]
markushi: False positive — this is the 3-arg `register(key, contract, callback)` overload, which has no `LifecycleOwner` and no STARTED check. That exception only comes from the 4-arg overload. The picker tests register from a RESUMED `ComponentActivity` and pass.

sentry-android-core/src/main/java/io/sentry/android/core/SentryUserFeedbackForm.java
runningcode: Thanks for the fix! Do we know what the exception that is thrown when the device doesn't have a screenshot picker? I would favor just catching that one but I'll accept this as well.

sentry-android-core/src/main/java/io/sentry/android/core/SentryUserFeedbackForm.java
```diff
@@ -388,12 +438,160 @@ protected void onStart() {
...
     final @Nullable FeedbackShakeIntegration integration = getFeedbackShakeIntegration();
     if (integration != null) {
       integration.onDialogGone(this);
     }
   }
 
+  private void maybeRegisterScreenshotPicker(final @NotNull SentryOptions options) {
+    // Clear any previously selected image so subsequent show() calls start with a fresh form
+    final @NotNull Button btnAddScreenshot =
+        findViewById(R.id.sentry_dialog_user_feedback_btn_add_screenshot);
```
runningcode, line 454: i feel i should mention that using this view system is considered deprecated by google now 😢

sentry/src/main/java/io/sentry/SentryFeedbackOptions.java
```diff
@@ -261,6 +291,28 @@ public void setUseShakeGesture(final boolean useShakeGesture) {
...
   }
 
+  /**
+   * Displays a button on the feedback form that allows the user to attach a screenshot from the
+   * device gallery. Only supported on Android, and requires the androidx.activity dependency and a
+   * ComponentActivity host; the button is hidden otherwise. Defaults to true.
+   *
+   * @return true if the attach screenshot button is shown
+   */
+  public boolean isEnableAttachScreenshot() {
```
runningcode, line 301: If we prefix with `get` instead of `is` we get a nicer synthetic getter in Kotlin like `attachScreenshotEnabled`
```suggestion
  public boolean getAttachScreenshotEnabled() {
```

sentry-android-core/src/main/java/io/sentry/android/core/SentryUserFeedbackForm.java
```diff
@@ -388,12 +438,160 @@ protected void onStart() {
...
+      return;
+    }
+    final @Nullable Activity activity = getActivity(getContext());
+    if (activity != null && isScreenshotPickerAvailable(options)) {
+      try {
+        screenshotPicker =
+            SentryFeedbackScreenshotPicker.register(
+                activity, uri -> onScreenshotPicked(options, btnAddScreenshot, uri));
+      } catch (LinkageError e) {
+        // This is where androidx.activity is linked for the first time. It is a compileOnly
```
runningcode, line 468: Can we check this in `isScreenshotPickerAvailable` ?

sentry-android-core/src/main/java/io/sentry/android/core/SentryUserFeedbackForm.java
```diff
@@ -203,6 +218,38 @@ protected void onCreate(Bundle savedInstanceState) {
...
+        findViewById(R.id.sentry_dialog_user_feedback_btn_add_screenshot);
+
+    // The button is made visible in onStart, once the screenshot picker is registered successfully
+    btnAddScreenshot.setVisibility(View.GONE);
+    btnAddScreenshot.setOnClickListener(
+        v -> {
+          if (selectedImageUri == null) {
+            if (screenshotPicker != null) {
+              try {
+                screenshotPicker.launch();
```
runningcode, line 231: The docs for ScreenShotPicker suggest we should only load the class once `SentryUserFeedbackForm#isScreenshotPickerAvailable` returns true and we do that. So it seems to me that it can only be not null if we already know the class is available so I'm wondering why we have to wrapp all the calls in try catch?

My suggestion to improve this would be can we find a way to check all the prerequisites beforehand so we don't have to wrap every call with a complicated try/catch block?

runningcode (approved):
looks better! i think we can simplify the stuff around calling the picker!

sentry-android-core/src/main/java/io/sentry/android/core/SentryUserFeedbackForm.java
markushi: Whilst the class could be available, we have no control of the actual version of the library, as it's app defined. So in theory a breaking API change / or change in behavior could cause a crash here.

sentry/src/main/java/io/sentry/SentryFeedbackOptions.java
```diff
@@ -626,9 +738,28 @@ public String toString() {
...
         + '}';
   }
 
+  @ApiStatus.Internal
+  public @NotNull LoadClass getLoadClass() {
+    return loadClass;
+  }
+
+  @ApiStatus.Internal
+  public @NotNull void setLoadClass(final @NotNull LoadClass loadClass) {
```
romtsn:
```suggestion
  public void setLoadClass(final @NotNull LoadClass loadClass) {
```
i guess don't need it on a `void` method :)

sentry/src/main/java/io/sentry/SentryFeedbackOptions.java
```diff
@@ -78,6 +86,18 @@ public final class SentryFeedbackOptions {
...
 
+  /** The label of the button to attach a screenshot. Defaults to "Add a screenshot". */
+  private @NotNull CharSequence addScreenshotButtonLabel = "Add a screenshot";
+
+  /** The label of the button to remove an attached screenshot. Defaults to "Remove screenshot". */
+  private @NotNull CharSequence removeScreenshotButtonLabel = "Remove screenshot";
+
+  /**
+   * The message displayed when the selected screenshot exceeds the maximum attachment size.
+   * Defaults to "Image is too large".
```
romtsn:
```suggestion
   * Defaults to "Screenshot is too large".
```

sentry/src/main/java/io/sentry/SentryFeedbackOptions.java
```diff
@@ -461,6 +513,64 @@ public void setSuccessMessageText(final @NotNull CharSequence successMessageText
...
+   * @param removeScreenshotButtonLabel the label of the button to remove an attached screenshot
+   */
+  public void setRemoveScreenshotButtonLabel(
+      final @NotNull CharSequence removeScreenshotButtonLabel) {
+    this.removeScreenshotButtonLabel = removeScreenshotButtonLabel;
+  }
+
+  /**
+   * The message displayed when the selected screenshot exceeds the maximum attachment size.
+   * Defaults to "Image is too large".
```
romtsn:
```suggestion
   * Defaults to "Screenshot is too large".
```

sentry-android-core/src/test/java/io/sentry/android/core/SentryUserFeedbackFormTest.kt
```diff
@@ -213,4 +234,227 @@ class SentryUserFeedbackFormTest {
...
+    sut.findViewById<EditText>(R.id.sentry_dialog_user_feedback_edt_description).setText("message")
+    sut.findViewById<Button>(R.id.sentry_dialog_user_feedback_btn_send).performClick()
+
+    val hintCaptor = argumentCaptor<Hint>()
+    verify(fixture.mockFeedbackApi).capture(any<Feedback>(), hintCaptor.capture())
+    assertThat(hintCaptor.firstValue.attachments).isEmpty()
+  }
+
+  @Test
+  fun `feedback is captured with an image attachment when the size of the image is unknown`() {
```
romtsn, line 387: shall we also test that we drop an over-sized image?

romtsn (approved):
nice work!

sentry/src/main/java/io/sentry/SentryFeedbackOptions.java
markushi: Fully agree that this would be nicer, but I'd rather keep it this way in order to be consistent with all other option getters/setters.

sentry-android-core/src/main/java/io/sentry/android/core/SentryUserFeedbackForm.java
```diff
@@ -365,6 +414,7 @@ protected void onStart() {
     edtMessage.setError(null);
 
     final @NotNull SentryOptions options = Sentry.getCurrentScopes().getOptions();
+    maybeRegisterScreenshotPicker(options);
```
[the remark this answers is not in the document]
markushi: that's fine, we should only read the flag only once during construction time

sentry-android-core/src/main/java/io/sentry/android/core/SentryFeedbackScreenshotPicker.java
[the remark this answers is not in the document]
markushi: I don't think there's a better way for this right now, keeping as-is, as it's an edge case anyway.

sentry-android-core/src/main/java/io/sentry/android/core/SentryUserFeedbackForm.java
markushi: We check for class availability first anyway, so this catch is already an edge case where the API doesn't behave as assumed / compiled against. As it's out of our control it really could be any kind of exception here.

PR #5832, opened 2026-07-24
test(replay): [Data Collection 16] Keep Replay independent

adinauer:
## PR Stack (Data Collection)

- #5759
- #5760
- #5761
- #5799
- #5800
- #5801
- #5802
- #5803
- #5804
- #5805
- #5806
- #5807
- #5810
- #5811
- #5812
- #5831
- #5832
- #5834
- #5937
- #5983
- #6036
- #6038
- #6064
- #6065
- #6075
- #6076
- #6122

---

## :scroll: Description

Keep Session Replay network privacy settings independent from Data Collection.

Add regression coverage proving that restrictive Data Collection body and header settings do not affect explicitly enabled Replay network details. Replay continues to use its existing URL allow/deny lists, body capture option, and request/response header allow lists.

## :bulb: Motivation and Context

`sendDefaultPii` has no effect on Session Replay today. Wiring Data Collection into Replay would introduce new behavior and combine controls with opposing defaults: Replay network details are opt-in through URL allow lists, while Data Collection uses its own defaults for supported integrations.

Keeping the features independent avoids privacy-sensitive behavior changes for existing Replay users. The Data Collection documentation will call out that Replay remains governed by Replay-specific options.

Refs #5666

## :green_heart: How did you test it?

- `./gradlew :sentry-okhttp:test --tests='io.sentry.okhttp.SentryOkHttpInterceptorTest'`
- `./gradlew spotlessApply apiDump`
- `git diff --check`

## :pencil: Checklist

- [x] I added GH Issue ID _&_ Linear ID
- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [x] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.
- [x] Public API changes reviewed by another Mobile SDK team member or implemented according to the [develop docs](https://develop.sentry.dev/) spec.

## :crystal_ball: Next steps

Document that Session Replay is not controlled by Data Collection and continues to use Replay-specific privacy options.

#skip-changelog

> ⚠️ **Merge this PR using a merge commit** (not squash). Only the collection branch is squash-merged into main.

PR #5831, opened 2026-07-24
fix(android): [Data Collection 15] Scope device info cache to SDK options

adinauer:
## PR Stack (Data Collection)

- #5759
- #5760
- #5761
- #5799
- #5800
- #5801
- #5802
- #5803
- #5804
- #5805
- #5806
- #5807
- #5810
- #5811
- #5812
- #5831
- #5832
- #5834
- #5937
- #5983
- #6036
- #6038
- #6064
- #6065
- #6075
- #6076
- #6122

---

## :scroll: Description

Scope `DeviceInfoUtil` to each `SentryAndroidOptions` instance instead of keeping one process-global singleton.

The existing `DeviceInfoUtil.getInstance(context, options)` entry point remains, but now delegates to a lazily initialized utility owned by the supplied options. Repeated calls with the same options reuse cached device information, while different SDK initializations keep their collection policy, logger, connection provider, root-check setting, and additional-context settings isolated.

Remove the test-only global reset because fresh options now provide test isolation naturally.

## :bulb: Motivation and Context

The user information policy introduced in the preceding stack PR exposed a pre-existing lifecycle problem: Android auto-init followed by manual init could continue reading the first initialization's options from the process-global `DeviceInfoUtil`. That could attach `device.id` despite the active initialization setting `dataCollection.userInfo=false`, and could also retain stale additional-context and root-check behavior.

Owning the cache from `SentryAndroidOptions` gives every SDK initialization one internally consistent utility without a static map or process-lifetime references to old options.

Refs #5666

## :green_heart: How did you test it?

- `./gradlew spotlessApply apiDump`
- `./gradlew :sentry-android-core:testDebugUnitTest --tests='io.sentry.android.core.DeviceInfoUtilTest' --tests='io.sentry.android.core.DefaultAndroidEventProcessorTest' --tests='io.sentry.android.core.ApplicationExitInfoEventProcessorTest' --tests='io.sentry.android.core.InternalSentrySdkTest' :sentry-android-core:apiCheck :sentry-android-core:spotlessJavaCheck :sentry-android-core:spotlessKotlinCheck`
- `git diff --check`

## :pencil: Checklist

- [x] I added GH Issue ID _&_ Linear ID
- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [x] Review from the native team if needed.
[...]

PR #5812, opened 2026-07-22
feat(user): [Data Collection 14] Apply user information policy

adinauer:
## PR Stack (Data Collection)

- #5759
- #5760
- #5761
- #5799
- #5800
- #5801
- #5802
- #5803
- #5804
- #5805
- #5806
- #5807
- #5810
- #5811
- #5812
- #5831
- #5832
- #5834
- #5937
- #5983
- #6036
- #6038
- #6064
- #6065
- #6075
- #6076
- #6122

---

## :scroll: Description

Apply the user information Data Collection policy to automatic identity enrichment across core, Android, and Spring integrations.

The policy controls automatic `{{auto}}` IP addresses, Spring request and Spring Security usernames, and SDK-generated Android installation identifiers written to `user.id` and `device.id`. Explicit `userInfo=false` suppresses these automatic values, while application-supplied user data and custom `SentryUserProvider` values remain untouched.

When Data Collection is absent, existing behavior is preserved: `sendDefaultPii` continues to control automatic IP and Spring user enrichment, while Android installation identifiers remain enabled. The unused private `TraceContext` user-ID helper is removed without changing generated trace context data.

## :bulb: Motivation and Context

Automatic user identity fields can contain personal or stable identifying data. This wires the existing `dataCollection.userInfo` option to all supported Java and Android collectors while preserving the legacy bridge for applications that have not configured Data Collection.

Refs #5666

## :green_heart: How did you test it?

- `./gradlew spotlessApply apiDump`
- `./gradlew :sentry:apiCheck`
- Core resolver and event processor tests
- Android event processor, device information, exit-info, and internal SDK tests
- Spring, Spring Jakarta, and Spring 7 request/Security user-provider tests
- `git diff --check`

## :pencil: Checklist

- [x] I added GH Issue ID _&_ Linear ID
- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [x] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.
- [x] Public API changes reviewed by another Mobile SDK team member or implemented according to the [develop docs](https://develop.sentry.dev/) spec.

## :crystal_ball: Next steps

Complete the remaining Data Collection configuration, documentation, and migration work.

#skip-changelog

[...]

PR #5811, opened 2026-07-22
feat(http): [Data Collection 13] Apply cookie collection policy

adinauer:
## PR Stack (Data Collection)

- #5759
- #5760
- #5761
- #5799
- #5800
- #5801
- #5802
- #5803
- #5804
- #5805
- #5806
- #5807
- #5810
- #5811
- #5812
- #5831
- #5832
- #5834
- #5937
- #5983
- #6036
- #6038
- #6064
- #6065
- #6075
- #6076
- #6122

---

## :scroll: Description

Apply the cookie Data Collection policy to automatically captured request and response cookies.

The policy is used by Spring MVC/WebFlux, OkHttp, Ktor, and Apollo 3/4 failed-request events. Explicit Data Collection supports off, deny-list, and allow-list behavior. Built-in sensitive cookie names and Spring integration-provided session cookie names are always filtered, including when allow-listed. `Set-Cookie` values are parsed separately so attributes such as `Path`, `HttpOnly`, and `SameSite` are preserved.

When Data Collection is absent, integrations preserve their existing `sendDefaultPii` behavior, including Spring's legacy security-cookie filtering and raw HTTP-client cookie values.

## :bulb: Motivation and Context

Cookie collection previously relied on `sendDefaultPii` as an on/off gate, with filtering only in Spring integrations. This wires the existing `dataCollection.cookies` option across cookie capture paths and provides consistent explicit-mode filtering without changing the legacy bridge for applications that have not configured Data Collection.

Refs #5666

## :green_heart: How did you test it?

- `./gradlew spotlessApply apiDump`
- `./gradlew :sentry:apiCheck`
- Core cookie filtering tests
- OkHttp and Ktor failed-request tests
- Spring, Spring Jakarta, and Spring 7 request tests
- Apollo 3 and Apollo 4 test suites
- `git diff --check`

## :pencil: Checklist

- [x] I added GH Issue ID _&_ Linear ID
- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [ ] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.
- [x] Public API changes reviewed by another Mobile SDK team member or implemented according to the [develop docs](https://develop.sentry.dev/) spec.

## :crystal_ball: Next steps

[...]

PR #5833, opened 2026-07-24
fix(android): [Data Collection 17] Apply user info policy to distinct ID

adinauer:
## PR Stack (Data Collection)

- #5759
- #5760
- #5761
- #5799
- #5800
- #5801
- #5802
- #5803
- #5804
- #5805
- #5806
- #5807
- #5810
- #5811
- #5812
- #5831
- #5832
- #5833
- #5834
- #5937
- #5983
- #5985
- #6036
- #6038

---

## :scroll: Description

Apply the resolved user information policy when generating Android's default installation-backed `distinctId`.

Move default `distinctId` generation out of `AndroidOptionsInitializer` and run it after the programmatic `SentryAndroid.init` callback. Generate the installation ID only when no distinct ID is configured and `dataCollection.userInfo` allows automatic identity collection.

Existing runtime gates in Android event, exit-info, device-context, and hybrid scope enrichment remain unchanged.

Android logs and metrics use `distinctId` as an automatic `user.id` fallback when no scope user exists. The previous unconditional default generation left that SDK-generated identity available even when applications explicitly configured `dataCollection.userInfo=false`.

Evaluating the policy after the programmatic callback prevents automatic `distinctId` generation for opted-out applications while preserving custom IDs and the legacy absent-Data-Collection behavior.

Refs #5666

- `./gradlew :sentry-android-core:testDebugUnitTest --tests='io.sentry.android.core.AndroidOptionsInitializerTest' --tests='io.sentry.android.core.ApplicationExitInfoEventProcessorTest' --tests='io.sentry.android.core.DeviceInfoUtilTest'`
- `./gradlew :sentry-android-core:apiCheck :sentry-android-core:spotlessJavaCheck :sentry-android-core:spotlessKotlinCheck`
- `git diff --check`

## :pencil: Checklist

## :crystal_ball: Next steps

Add external and Android manifest Data Collection configuration before finalizing the stack.

#skip-changelog

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
@@ -149,6 +149,14 @@ public final class SentryAndroid {
                       "Error in the 'OptionsConfiguration.configure' callback.",
                       t);
             }
+            if (options.getDistinctId() == null
+                && options.getDataCollectionResolver().isUserInfoWithLegacyAlways()) {
+              try {
+                options.setDistinctId(Installation.id(context));
+              } catch (RuntimeException e) {
+                options.getLogger().log(SentryLevel.ERROR, "Could not generate distinct Id.", e);
+              }
+            }
 
             // if SentryPerformanceProvider was disabled or removed,
             // we set the app start / sdk init time here instead
--- a/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
@@ -149,14 +149,6 @@ public final class SentryAndroid {
                       "Error in the 'OptionsConfiguration.configure' callback.",
                       t);
             }
-            if (options.getDistinctId() == null
-                && options.getDataCollectionResolver().isUserInfoWithLegacyAlways()) {
-              try {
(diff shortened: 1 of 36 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
```diff
@@ -149,6 +149,14 @@ public static void init(
                       "Error in the 'OptionsConfiguration.configure' callback.",
                       t);
             }
+            if (options.getDistinctId() == null
```
adinauer: Moving this here no longer allows customers to override it in the options configuration callback passed to init since it used to run before but now is set afterwards. It can still be removed via `beforeSend`. Is that OK?

lbloder: Good question, as this is a change in behaviour for existing users. I.e. if they update and have previously set the id to null in order not to track the id, it would be turned back on after updating the sdk. Also, `beforeSend` doesn't cover logs or metrics, correct?

lbloder (approved):

sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroid.java
adinauer: This is a leftover, reverting since we decided not to guard `Installation.id()`

PR #5810, opened 2026-07-22
feat(http): [Data Collection 12] Apply query parameter policy

adinauer:
## PR Stack (Data Collection)

- #5759
- #5760
- #5761
- #5799
- #5800
- #5801
- #5802
- #5803
- #5804
- #5805
- #5806
- #5807
- #5810
- #5811
- #5812
- #5831
- #5832
- #5834
- #5937
- #5983
- #6036
- #6038
- #6064
- #6065
- #6075
- #6076
- #6122

---

## :scroll: Description

Apply the query-parameter Data Collection policy to Sentry-owned automatic HTTP request data, spans, breadcrumbs, and failed-request events.

The policy is used by Servlet, Spring MVC/WebFlux, OkHttp, Ktor, Apollo 2/3/4, OpenFeign, RestTemplate, and WebClient. Explicit Data Collection supports off, deny-list, and allow-list behavior. Built-in sensitive parameter names are always filtered, including when allow-listed. Encoded parameter names are decoded for matching while the original query encoding is retained.

When Data Collection is absent, integrations continue to attach raw query values as before. Completed OpenTelemetry attributes also retain their existing values because the exporter cannot distinguish manually supplied attributes from values added by auto-instrumentation.

## :bulb: Motivation and Context

Query strings collected by Sentry-owned automatic instrumentation need a dedicated privacy control. This wires `dataCollection.urlQueryParams` into those Java HTTP integration paths while preserving the legacy bridge for applications that have not configured Data Collection.

Refs #5666

## :green_heart: How did you test it?

- `./gradlew spotlessApply apiDump`
- `./gradlew :sentry:apiCheck`
- Core resolver, URL, and query-filter tests
- Servlet query tests
- OkHttp query integration test
- `./gradlew :sentry-opentelemetry:sentry-opentelemetry-core:test --tests '*OpenTelemetryAttributesExtractorTest'`
- Full Ktor, Apollo 2/3/4, OpenFeign, Spring, Spring Jakarta, and Spring 7 test suites
- `git diff --check`

## :pencil: Checklist

- [x] I added GH Issue ID _&_ Linear ID
- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [ ] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.
[...]

PR #5807, opened 2026-07-21
feat(http): [Data Collection 11] Apply response header policy

adinauer:
## PR Stack (Data Collection)

- #5759
- #5760
- #5761
- #5799
- #5800
- #5801
- #5802
- #5803
- #5804
- #5805
- #5806
- #5807
- #5810
- #5811
- #5812
- #5831
- #5832
- #5834
- #5937
- #5983
- #6036
- #6038
- #6064
- #6065
- #6075
- #6076
- #6122

---

## :scroll: Description

Apply the response-header Data Collection policy to OkHttp, Ktor, and Apollo 3/4 failed-request events.

Explicit Data Collection supports off, deny-list, and allow-list behavior with canonical sensitive-key filtering. When Data Collection is absent, `sendDefaultPii=false` continues to omit response headers and `sendDefaultPii=true` continues to collect headers while omitting legacy-sensitive names.

## :bulb: Motivation and Context

These are the current global response-header consumers. Central filtering provides granular response-header control while preserving historical behavior for applications that have not opted into Data Collection. Session Replay network headers remain governed by Replay-specific integration controls.

Refs #5666

## :green_heart: How did you test it?

- `./gradlew spotlessApply apiDump`
- Full OkHttp, Ktor, Apollo 3, and Apollo 4 tests
- `git diff --check`

## :pencil: Checklist

- [x] I added GH Issue ID _&_ Linear ID
- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [ ] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

## :crystal_ball: Next steps

Migrate cookie and query-parameter collection. Document that Session Replay remains governed by Replay-specific privacy controls.

#skip-changelog

> ⚠️ **Merge this PR using a merge commit** (not squash). Only the collection branch is squash-merged into main.

PR #5806, opened 2026-07-21
feat(http): [Data Collection 10] Apply request header policy

adinauer:
## PR Stack (Data Collection)

- #5759
- #5760
- #5761
- #5799
- #5800
- #5801
- #5802
- #5803
- #5804
- #5805
- #5806
- #5807
- #5810
- #5811
- #5812
- #5831
- #5832
- #5834
- #5937
- #5983
- #6036
- #6038
- #6064
- #6065
- #6075
- #6076
- #6122

---

## :scroll: Description

Apply the request-header Data Collection policy across Servlet, Spring MVC and WebFlux, OkHttp, Ktor, and Apollo 3/4 automatic instrumentation.

Explicit Data Collection supports off, deny-list, and allow-list behavior with canonical sensitive-key filtering. When Data Collection is absent, each integration retains its previous `sendDefaultPii` and legacy sensitive-header behavior.

Completed OpenTelemetry attributes retain their existing behavior because the exporter cannot distinguish manually supplied attributes from values added by auto-instrumentation.

## :bulb: Motivation and Context

Request header collection historically differs by integration. A shared filter provides specification behavior for opted-in applications without changing existing applications that continue to use legacy configuration.

Refs #5666

## :green_heart: How did you test it?

- `./gradlew spotlessApply apiDump`
- Full relevant core, Servlet, Spring, OkHttp, Ktor, Apollo 3, and Apollo 4 tests
- `./gradlew :sentry-opentelemetry:sentry-opentelemetry-core:test --tests '*OpenTelemetryAttributesExtractorTest'`
- `./gradlew apiCheck`

## :pencil: Checklist

- [x] I added GH Issue ID _&_ Linear ID
- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [ ] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

## :crystal_ball: Next steps

Apply the response-header policy and migrate cookies separately.

#skip-changelog

> ⚠️ **Merge this PR using a merge commit** (not squash). Only the collection branch is squash-merged into main.

PR #5805, opened 2026-07-21
feat(graphql): [Data Collection 9] Apply outgoing response body policy

adinauer:
## PR Stack (Data Collection)

- #5759
- #5760
- #5761
- #5799
- #5800
- #5801
- #5802
- #5803
- #5804
- #5805
- #5806
- #5807
- #5810
- #5811
- #5812
- #5831
- #5832
- #5834
- #5937
- #5983
- #6036
- #6038
- #6064
- #6065
- #6075
- #6076
- #6122

---

## :scroll: Description

Apply the outgoing response body policy to GraphQL Java execution results attached by `ExceptionReporter`.

Explicit Data Collection controls the serialized GraphQL response independently from the request document. When Data Collection is absent, the existing `sendDefaultPii` and `maxRequestBodySize` gate remains unchanged.

## :bulb: Motivation and Context

A server-side GraphQL execution result is outgoing response body content and can contain customer data. It requires a directional policy distinct from the incoming GraphQL document and variables.

Refs #5666

## :green_heart: How did you test it?

- `./gradlew spotlessApply apiDump`
- `DataCollectionResolverTest`
- `ExceptionReporterTest`
- `./gradlew apiCheck`

## :pencil: Checklist

- [x] I added GH Issue ID _&_ Linear ID
- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [ ] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

## :crystal_ball: Next steps

Apply request and response header collection behavior across HTTP integrations.

#skip-changelog

> ⚠️ **Merge this PR using a merge commit** (not squash). Only the collection branch is squash-merged into main.

PR #5834, opened 2026-07-24
fix(spring): [Data Collection 18] Bind key-value policies

adinauer:
## PR Stack (Data Collection)

- #5759
- #5760
- #5761
- #5799
- #5800
- #5801
- #5802
- #5803
- #5804
- #5805
- #5806
- #5807
- #5810
- #5811
- #5812
- #5831
- #5832
- #5834
- #5937
- #5983
- #6036
- #6038
- #6064
- #6065
- #6075
- #6076
- #6122

---

## :scroll: Description

Make `KeyValueCollectionBehavior` JavaBean-bindable so Spring Boot can configure Data Collection policies for cookies, URL query parameters, and request/response headers.

Add a public no-argument constructor with the default deny-list behavior and setters for `mode` and `terms`. The existing `off`, `denyList`, and `allowList` factories remain available, and terms continue to be defensively copied into an immutable list.

Spring Boot could bind scalar Data Collection fields but silently ignored nested key-value policies because `KeyValueCollectionBehavior` only exposed private construction and immutable fields. A scalar property could therefore make the namespace explicit while the requested cookie, query, or header restriction was not applied.

Refs #5666

- `./gradlew :sentry:test --tests io.sentry.KeyValueCollectionBehaviorTest :sentry-spring-boot:test --tests 'io.sentry.spring.boot.SentryAutoConfigurationTest.data collection key value properties are applied to SentryOptions' :sentry-spring-boot-jakarta:test --tests 'io.sentry.spring.boot.jakarta.SentryAutoConfigurationTest.data collection key value properties are applied to SentryOptions' :sentry-spring-boot-4:test --tests 'io.sentry.spring.boot4.SentryAutoConfigurationTest.data collection key value properties are applied to SentryOptions'`
- `git diff --check`

## :pencil: Checklist

## :crystal_ball: Next steps

Add the remaining external and Android manifest Data Collection configuration paths.

#skip-changelog


lbloder (approved):
LGTM 👍

sentry/src/main/java/io/sentry/KeyValueCollectionBehavior.java
```diff
@@ -53,10 +56,18 @@ private KeyValueCollectionBehavior(final @NotNull Mode mode, final @NotNull List
     return mode;
   }
 
+  public void setMode(final @NotNull Mode mode) {
```
runningcode, line 59: Up until this PR, `KeyValueCollectionBehavior` was immutable. This is important because `DataCollectionResolver` holds two static final `KeyValueCollectionBehavior`s.
This means that users can modify the static singleton which is not a good idea.
I think we should rather keep this class immutable to prevent issues.

adinauer: Ah good catch!

Unfortunately this isn't easily possible for Spring Boot property binding. I will simply inline the static fields to call the factory methods directly.

runningcode (approved):
LGTM!

PR #5837, opened 2026-07-24
fix(replay): Post checkCanRecord to main thread to prevent deadlock

romtsn:
## Summary
- `onScreenshotRecorded` can run on the replay executor thread (PixelCopy masked-capture and emit paths). `checkCanRecord()` → `pauseInternal()` acquires `lifecycleLock` — if another thread holds that lock and submits to the same single-threaded executor, we deadlock.
- Post `checkCanRecord()` to the main looper so it never runs on the executor thread.

This was a pre-existing issue introduced with SurfaceView capturing (masked-capture path calls `onScreenshotRecorded` from executor at `PixelCopyStrategy.kt:209`), and is also relevant for PR #5808 which adds the `emitLastScreenshot` executor path.

## Test plan
- [x] Existing `onScreenshotRecorded pauses replay when rate-limited for sessions` test updated to idle looper
- [x] `./gradlew :sentry-android-replay:testReleaseUnitTest --tests="*ReplayIntegrationTest*"` passes

### DIFF
--- a/sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
+++ b/sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
@@ -391,21 +391,23 @@ public class ReplayIntegration(
       recorder?.close()
       recorder = null
       rootViewsSpy.close()
-      if (lazyReplayExecutor.isInitialized()) {
-        if (options.threadChecker.isMainThread) {
-          replayExecutor.gracefulShutdown()
-        } else {
-          replayExecutor.shutdown()
-        }
-      }
-      if (lazyPersistingExecutor.isInitialized()) {
-        if (options.threadChecker.isMainThread) {
-          persistingExecutor.gracefulShutdown()
-        } else {
-          persistingExecutor.shutdown()
-        }
-      }
       lifecycle.currentState = CLOSED
+    }
+    // ponytail: shutdown outside lock — awaiting termination while holding lifecycleLock deadlocks
+    // if any executor task tries to acquire the same lock
+    if (lazyReplayExecutor.isInitialized()) {
+      if (options.threadChecker.isMainThread) {
+        replayExecutor.gracefulShutdown()
+      } else {
+        replayExecutor.shutdown()
+      }
+    }
+    if (lazyPersistingExecutor.isInitialized()) {
+      if (options.threadChecker.isMainThread) {
+        persistingExecutor.gracefulShutdown()
+      } else {
+        persistingExecutor.shutdown()
+      }
     }
   }
 
--- a/sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
+++ b/sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
@@ -393,7 +393,7 @@ public class ReplayIntegration(
       rootViewsSpy.close()
       lifecycle.currentState = CLOSED
     }
-    // ponytail: shutdown outside lock — awaiting termination while holding lifecycleLock deadlocks
+    // shutdown outside lock — awaiting termination while holding lifecycleLock deadlocks
     // if any executor task tries to acquire the same lock
     if (lazyReplayExecutor.isInitialized()) {
       if (options.threadChecker.isMainThread) {
--- a/sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
+++ b/sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
@@ -131,7 +131,7 @@ public class ReplayIntegration(
   private var replayCaptureStrategyProvider: ((isFullSession: Boolean) -> CaptureStrategy)? = null
   private var mainLooperHandler: MainLooperHandler = MainLooperHandler()
   private var gestureRecorderProvider: (() -> GestureRecorder)? = null
-  private val lifecycleLock = AutoClosableReentrantLock()
+  internal val lifecycleLock = AutoClosableReentrantLock()
   private val lifecycle = ReplayLifecycle()
 
   override fun register(scopes: IScopes, options: SentryOptions) {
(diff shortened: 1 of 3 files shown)


sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
```diff
@@ -352,7 +353,15 @@ public class ReplayIntegration(
...
-    checkCanRecord()
+    // When called from the replay executor (PixelCopy masked-capture and emit paths),
+    // checkCanRecord -> pauseInternal acquires lifecycleLock. If another thread holds
+    // that lock while submitting to the same executor, we deadlock. Post to main in
+    // that case; on main thread call directly to keep the existing synchronous behavior.
+    if (Looper.myLooper() == Looper.getMainLooper()) {
+      checkCanRecord()
+    } else {
+      mainLooperHandler.post { checkCanRecord() }
+    }
```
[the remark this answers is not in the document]
romtsn: technically correct, but was the case before this PR -- and adding one more frame when a rate-limit is hit is not at all harmful, so we accept it here

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
```diff
@@ -352,7 +353,7 @@ public class ReplayIntegration(
       }
       addFrame(bitmap, frameTimeStamp, screen)
     }
-    checkCanRecord()
+    postOnMainThread { checkCanRecord() }
```
0xadam-brown, line 356: m: Looks like we have (at least) two choices for how we fix our deadlock:

1. never allow code that may acquire `lifecycleLock` to run on the replay executor thread (ie, this PR's current approach), or
2. don't allow code holding lifecycleLock to try to shut down the replay executor.

I like (2) because it's easier to confine to a single place (the `close()` method), both now and over time; it's reasonable to have a policy that prohibits trying to shut down executors (or do other potentially blocking things) while holding locks (because of deadlocks like this one); and tracking down exactly which code is actually executing onScreenshotRecorded() is...tricky.

That said, maybe (2) is a pain to implement? If so, I'm okay with (1), but my vote would be to rename `postOnMainThread()` -> `dontBlockExecutorShutdown()` (or something like that). The rename would put us on notice that we're being clever, and it'd also avoid the suggestion that use of the main thread is itself significant.

******

**_Background explaining why I think releasing lifecycleLock before shutting down the replay executor will prevent our deadlock:_**

AFAICT, our deadlock emerges when:

1. Thread A holds `lifecycleLock`.
2. Replay executor on thread B runs onScreenshotRecorded(), which calls checkCanRecord().
3. checkCanRecord() tries to acquire lifecycleLock via pauseInternal() but can't because thread A holds it.
4. Thread A finishes its work and – before releasing lifecycleLock – calls replayExecutor.shutdown().
5. But shutdown() only completes if all tasks managed by the replayExecutor have terminated. 
6. The onScreenshotRecorded() task hasn't completed because its waiting for the lifecycleLock. --> and our deadlock is born.

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
```diff
@@ -444,6 +445,17 @@ public class ReplayIntegration(
     captureStrategy?.onTouchEvent(event)
   }
 
+  // Runs [block] on the main thread. If already there, executes inline; otherwise posts via
+  // the main looper handler. Prevents deadlocks when lifecycle-lock-acquiring code (e.g.
+  // checkCanRecord -> pauseInternal) is called from the replay executor thread.
```
0xadam-brown, line 452: l: If we keep approach (1), worth explicitly mentioning that the invariant we need to maintain is never to allow the replay executor to try to acquire `lifecycleLock`, as it could prevent the executor from being shut down.

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
romtsn: thanks, that's a good point, however I don't think `replayExecutor.shutdown()` alone is the problem. There's also this scenario (as we'd attempted to fix earlier: https://github.com/getsentry/sentry-java/pull/4788):
  1. Thread A holds `lifecycleLock` (in pause(), stop(), start(), whatever)
  2. Executor thread runs onScreenshotRecorded → checkCanRecord() → pauseInternal() → tries to acquire lifecycleLock → blocks
  3. Thread A (still holding the lock) submits something to the same single-threaded executor → blocks waiting for a slot → deadlock

So I think to satisfy both, we actually need both option 1 and 2. I can add the 2nd one 👍

0xadam-brown: Ah, makes sense – good deal 👍 

**Side note:** Nothing that should delay us, but it'd be great to think about how we might move toward a world where task implementation and task execution policy (eg, thread application, locking, etc.) are cleanly separated so that implementations can ignore their execution context. Otherwise, reasoning about correctness gets... tricky 😅 . (We could also look into thread confinement policies that remove lots of locking issues from the get-go.)

romtsn: yeah, my plan was to confine replay lifecycle methods to the main thread as part of https://github.com/getsentry/sentry/issues/78018. Filed a sub-issue for that 👍  https://github.com/getsentry/sentry-java/issues/5847

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
romtsn: i think i'll skip this one since we're gonna rework this in #5847 soon anyway (plan is to start with this next week)

0xadam-brown (approved):
Thanks for the update 💯 

Up to you whether you think it worthwhile to add tests to `ReplayIntegrationTest` verifying that we shutdown outside the lifecycleLock, and that we don't deadlock when invoking onScreenshotRecorded() from the replay executor path while lifecycleLock is held. Otherwise lgtm.

PR #5842, opened 2026-07-27
fix(replay): Don't let a wedged video encoder freeze the app

### LINKED ISSUE #3556: ApplicationNotResponding sun.misc.Unsafe in park
### Platform

Flutter Mobile Android

### Obfuscation

Disabled

### Debug Info

Disabled

### Doctor

```
[✓] Flutter (Channel stable, 3.35.6, on openSUSE Tumbleweed 6.19.5-2-default, locale en_DK.UTF-8) [31ms]
    • Flutter version 3.35.6 on channel stable at /home/d/fvm/versions/3.35.6
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 9f455d2486 (5 months ago), 2025-10-08 14:55:31 -0500
    • Engine revision d2913632a4
    • Dart version 3.9.2
    • DevTools version 2.48.0
    • Feature flags: enable-web, enable-linux-desktop, enable-macos-desktop, enable-windows-desktop, enable-android, enable-ios, no-cli-animations, enable-lldb-debugging

[!] Android toolchain - develop for Android devices (Android SDK version 35.0.0) [1,883ms]
    • Android SDK at /home/d/Android/Sdk
    • Emulator version 36.4.9.0 (build_id 14788078) (CL:N/A)
    • Platform android-36, build-tools 35.0.0
    • ANDROID_HOME = /home/d/Android/Sdk
    • ANDROID_SDK_ROOT = /home/d/Android/Sdk
    • Java binary at: /home/d/.local/share/JetBrains/Toolbox/apps/android-studio-2/jbr/bin/java
      This JDK is specified in your Flutter configuration.
      To change the current JDK, run: `flutter config --jdk-dir="path/to/jdk"`.
    • Java version OpenJDK Runtime Environment (build 21.0.8+-14196175-b1038.72)
    ! Some Android licenses not accepted. To resolve this, run: flutter doctor --android-licenses

[✗] Chrome - develop for the web (Cannot find Chrome executable at google-chrome) [5ms]
    ! Cannot find Chrome. Try setting CHROME_EXECUTABLE to a Chrome executable.

[✓] Linux toolchain - develop for Linux desktop [216ms]
    • clang version 21.1.8
    • cmake version 4.2.3
    • ninja version 1.13.2
    • pkg-config version 2.5.1
    • OpenGL core renderer: AMD Radeon 610M (radeonsi, raphael_mendocino, ACO, DRM 3.64, 6.19.5-2-default)
    • OpenGL core version: 4.6 (Core Profile) Mesa 26.0.1
    • OpenGL core shading language version: 4.60
    • OpenGL ES renderer: AMD Radeon 610M (radeonsi, raphael_mendocino, ACO, DRM 3.64, 6.19.5-2-default)
    • OpenGL ES version: OpenGL ES 3.2 Mesa 26.0.1
    • OpenGL ES shading language version: OpenGL ES GLSL ES 3.20
    • GL_EXT_framebuffer_blit: yes
    • GL_EXT_texture_format_BGRA8888: yes

[✓] Android Studio (version 2025.2.3) [4ms]
    • Android Studio at /home/d/.local/share/JetBrains/Toolbox/apps/android-studio-2
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
    • android-studio-dir = /home/d/.local/share/JetBrains/Toolbox/apps/android-studio-2
    • Java version OpenJDK Runtime Environment (build 21.0.8+-14196175-b1038.72)

[✓] Connected device (2 available) [47ms]
    • sdk gphone64 x86 64 (mobile) • emulator-5554 • android-x64 • Android 16 (API 36) (emulator)
    • Linux (desktop)              • linux         • linux-x64   • openSUSE Tumbleweed 6.19.5-2-default

[✓] Network resources [470ms]
    • All expected network resources are available.

! Doctor found issues in 2 categories.
```

### Version

9.12.0

### Steps to Reproduce

1. take app to background
2. bring it up to foreground after some time

I see this error in the sentry itself. It has already been [presumably fixed in version 7](https://github.com/getsentry/sentry-dart/issues/1573), yet I am using version 9 and it's still happening. I have got 10 events in last 30 days, which is quite a lot, considering my small user base of 14 people.

### Expected Result

Works

### Actual Result

```
io.sentry.android.core.ApplicationNotResponding: Background ANR
    at jdk.internal.misc.Unsafe.park
    at java.util.concurrent.locks.LockSupport.park(LockSupport.java:221)
    at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(AbstractQueuedSynchronizer.java:754)
    at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(AbstractQueuedSynchronizer.java:990)
    at java.util.concurrent.locks.ReentrantLock$Sync.lock(ReentrantLock.java:153)
    at java.util.concurrent.locks.ReentrantLock.lock(ReentrantLock.java:322)
    at io.sentry.util.AutoClosableReentrantLock.acquire(r8-map-id-4bd4561c6bc95b6e9eaa50ad6a5155dc278353ee29bbb24ceb2796fa31b59f1e:1)
    at io.sentry.android.replay.ReplayIntegration.start(r8-map-id-4bd4561c6bc95b6e9eaa50ad6a5155dc278353ee29bbb24ceb2796fa31b59f1e:3)
    at io.sentry.android.core.LifecycleWatcher.startSession(r8-map-id-4bd4561c6bc95b6e9eaa50ad6a5155dc278353ee29bbb24ceb2796fa31b59f1e:59)
    at io.sentry.android.core.LifecycleWatcher.onForeground(r8-map-id-4bd4561c6bc95b6e9eaa50ad6a5155dc278353ee29bbb24ceb2796fa31b59f1e:1)
    at io.sentry.android.core.AppState$LifecycleObserver.onStart(r8-map-id-4bd4561c6bc95b6e9eaa50ad6a5155dc278353ee29bbb24ceb2796fa31b59f1e:25)
    at androidx.lifecycle.e.b(r8-map-id-4bd4561c6bc95b6e9eaa50ad6a5155dc278353ee29bbb24ceb2796fa31b59f1e:75)
    at androidx.lifecycle.s.a(r8-map-id-4bd4561c6bc95b6e9eaa50ad6a5155dc278353ee29bbb24ceb2796fa31b59f1e:25)
    at androidx.lifecycle.t.f(r8-map-id-4bd4561c6bc95b6e9eaa50ad6a5155dc278353ee29bbb24ceb2796fa31b59f1e:361)
    at androidx.lifecycle.t.e(r8-map-id-4bd4561c6bc95b6e9eaa50ad6a5155dc278353ee29bbb24ceb2796fa31b59f1e:91)
    at androidx.lifecycle.E$a.onActivityPostStarted(r8-map-id-4bd4561c6bc95b6e9eaa50ad6a5155dc278353ee29bbb24ceb2796fa31b59f1e:24)
    at android.app.Activity.dispatchActivityPostStarted(Activity.java:1714)
    at android.app.Activity.performStart(Activity.java:9474)
    at android.app.ActivityThread.handleStartActivity(ActivityThread.java:4728)
    at android.app.servertransaction.TransactionExecutor.performLifecycleSequence(TransactionExecutor.java:217)
    at android.app.servertransaction.TransactionExecutor.cycleToPath(TransactionExecutor.java:197)
    at android.app.servertransaction.TransactionExecutor.executeLifecycleItem(TransactionExecutor.java:169)
    at android.app.servertransaction.TransactionExecutor.executeTransactionItems(TransactionExecutor.java:104)
    at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:83)
    at android.app.ActivityThread$H.handleMessage(ActivityThread.java:3085)
    at android.os.Handler.dispatchMessage(Handler.java:114)
    at android.os.Looper.loopOnce(Looper.java:267)
    at android.os.Looper.loop(Looper.java:360)
    at android.app.ActivityThread.main(ActivityThread.java:10054)
    at java.lang.reflect.Method.invoke
    at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:616)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1115)

Thread: main
    at jdk.internal.misc.Unsafe.park
    at java.util.concurrent.locks.LockSupport.park(LockSupport.java:221)
    at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(AbstractQueuedSynchronizer.java:754)
    at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(AbstractQueuedSynchronizer.java:990)
    at java.util.concurrent.locks.ReentrantLock$Sync.lock(ReentrantLock.java:153)
    at java.util.concurrent.locks.ReentrantLock.lock(ReentrantLock.java:322)
    at io.sentry.util.AutoClosableReentrantLock.acquire(r8-map-id-4bd4561c6bc95b6e9eaa50ad6a5155dc278353ee29bbb24ceb2796fa31b59f1e:1)
    at io.sentry.android.replay.ReplayIntegration.start(r8-map-id-4bd4561c6bc95b6e9eaa50ad6a5155dc278353ee29bbb24ceb2796fa31b59f1e:3)
    at io.sentry.android.core.LifecycleWatcher.startSession(r8-map-id-4bd4561c6bc95b6e9eaa50ad6a5155dc278353ee29bbb24ceb2796fa31b59f1e:59)
    at io.sentry.android.core.LifecycleWatcher.onForeground(r8-map-id-4bd4561c6bc95b6e9eaa50ad6a5155dc278353ee29bbb24ceb2796fa31b59f1e:1)
    at io.sentry.android.core.AppState$LifecycleObserver.onStart(r8-map-id-4bd4561c6bc95b6e9eaa50ad6a5155dc278353ee29bbb24ceb2796fa31b59f1e:25)
    at androidx.lifecycle.e.b(r8-map-id-4bd4561c6bc95b6e9eaa50ad6a5155dc278353ee29bbb24ceb2796fa31b59f1e:75)
    at androidx.lifecycle.s.a(r8-map-id-4bd4561c6bc95b6e9eaa50ad6a5155dc278353ee29bbb24ceb2796fa31b59f1e:25)
    at androidx.lifecycle.t.f(r8-map-id-4bd4561c6bc95b6e9eaa50ad6a5155dc278353ee29bbb24ceb2796fa31b59f1e:361)
    at androidx.lifecycle.t.e(r8-map-id-4bd4561c6bc95b6e9eaa50ad6a5155dc278353ee29bbb24ceb2796fa31b59f1e:91)
    at androidx.lifecycle.E$a.onActivityPostStarted(r8-map-id-4bd4561c6bc95b6e9eaa50ad6a5155dc278353ee29bbb24ceb2796fa31b59f1e:24)
    at android.app.Activity.dispatchActivityPostStarted(Activity.java:1714)
    at android.app.Activity.performStart(Activity.java:9474)
    at android.app.ActivityThread.handleStartActivity(ActivityThread.java:4728)
    at android.app.servertransaction.TransactionExecutor.performLifecycleSequence(TransactionExecutor.java:217)
    at android.app.servertransaction.TransactionExecutor.cycleToPath(TransactionExecutor.java:197)
    at android.app.servertransaction.TransactionExecutor.executeLifecycleItem(TransactionExecutor.java:169)
    at android.app.servertransaction.TransactionExecutor.executeTransactionItems(TransactionExecutor.java:104)
    at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:83)
    at android.app.ActivityThread$H.handleMessage(ActivityThread.java:3085)
    at android.os.Handler.dispatchMessage(Handler.java:114)
    at android.os.Looper.loopOnce(Looper.java:267)
    at android.os.Looper.loop(Looper.java:360)
    at android.app.ActivityThread.main(ActivityThread.java:10054)
    at java.lang.reflect.Method.invoke
    at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:616)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1115)
```

### Are you willing to submit a PR?

No

### LINKED ISSUE #5870: Permanent app freeze on foreground: main thread blocks in ReplayIntegration.start while replay executor hangs in MediaCodec.dequeueOutputBuffer holding the replay lock (HyperOS 3 / Android 16)
### Integration

sentry-android (via sentry-flutter)

### Java Version

N/A (Flutter app)

### Version

sentry-android 8.49.0 (bundled by sentry_flutter 9.25.0)

### Steps to Reproduce

Device: Xiaomi 14 (houji), HyperOS 3 / Android 16, build `OS3.0.303.0.WNCCNXM` (latest stable).

Replay config:

```dart
options.replay.onErrorSampleRate = 1.0;  // buffer mode for all sessions
options.replay.sessionSampleRate = 0.1;
```

1. Use the app normally, then send it to background (screen off or app switch).
2. HyperOS 3 aggressively freezes the cached process (cgroup freezer). This appears to leave the in-flight `MediaCodec` session in a broken state.
3. Return to the app.

### Expected Result

App resumes normally. A hung replay video encoder must never block the main thread.

### Actual Result

The app is **permanently frozen**: no touch, no back gesture, window never gains focus. The OS records an ANR with subject *"Input dispatching timed out (Application does not have a focused window)"*. Only force-stop recovers.

It is a classic lock-ordering hazard rather than a device quirk:

**Thread 1 —** `SentryReplayIntegration-0` is stuck forever inside `MediaCodec.dequeueOutputBuffer` (native `ALooper::awaitResponse` never returns after the process was frozen mid-encode), while **holding the replay lock**:

```
"SentryReplayIntegration-0" daemon prio=5 tid=4 Native
  native: #03 pc 0001f2c8  libstagefright_foundation.so (android::ALooper::awaitResponse+284)
  native: #04 pc 0002432c  libstagefright_foundation.so (android::AMessage::postAndAwaitResponse+452)
  native: #05 pc 001d6390  libstagefright.so (android::MediaCodec::dequeueOutputBuffer+436)
  at android.media.MediaCodec.native_dequeueOutputBuffer(Native method)
  at android.media.MediaCodec.dequeueOutputBuffer(MediaCodec.java:4308)
  at io.sentry.android.replay.video.SimpleVideoEncoder.drainCodec(SimpleVideoEncoder.kt:218)
  at io.sentry.android.replay.video.SimpleVideoEncoder.release(SimpleVideoEncoder.kt:288)
  at io.sentry.android.replay.ReplayCache.createVideoOf(ReplayCache.kt:212)
  at io.sentry.android.replay.capture.CaptureStrategy$Companion.createSegment(CaptureStrategy.kt:92)
  at io.sentry.android.replay.capture.BaseCaptureStrategy.createSegmentInternal(BaseCaptureStrategy.kt:142)
  at io.sentry.android.replay.capture.BufferCaptureStrategy.createCurrentSegment$lambda$5(BufferCaptureStrategy.kt:241)
  at io.sentry.android.replay.util.ReplayExecutorService.submit$lambda$0(ReplayExecutorService.kt:26)
```

**Thread 2 —** `main` then blocks forever acquiring that same lock during activity start, killing the whole app:

```
"main" prio=5 tid=1 Waiting
  at jdk.internal.misc.Unsafe.park(Native method)
  at java.util.concurrent.locks.ReentrantLock.lock(ReentrantLock.java:322)
  at io.sentry.util.AutoClosableReentrantLock.acquire(AutoClosableReentrantLock.java:37)
  at io.sentry.android.replay.ReplayIntegration.start(ReplayIntegration.kt:171)
  at io.sentry.android.core.LifecycleWatcher.startSession(LifecycleWatcher.java:87)
  at io.sentry.android.core.LifecycleWatcher.onForeground(LifecycleWatcher.java:62)
  at io.sentry.android.core.AppState$LifecycleObserver.onStart(AppState.java:202)
  at androidx.lifecycle.ProcessLifecycleOwner.activityStarted$lifecycle_process_release(ProcessLifecycleOwner.kt:97)
  at android.app.Activity.performStart(Activity.java:9522)
```

A second full dump taken on another day shows the identical pair of stacks, so this is stable, not a one-off.

### Impact

Our analytics defined a "frozen resume" signature (app opened from background → zero further events → cold start within 120 s, i.e. user force-killed a dead app). Over 30 days:

* Xiaomi devices on Android 16 (HyperOS 3): **6.3 %** of resumes (1785 sampled devices)
* Xiaomi on Android ≤ 15: 0.3–0.6 %
* Every other brand on Android 16: 0.3–0.8 %

The rate climbed week over week as the HyperOS 3 OTA rolled out, independent of our app releases. Battery-optimization whitelisting on the device does not prevent it. We had to disable Session Replay entirely to fix our users, confirmed by an affected user: freeze gone with a replay-disabled build.

### Suggestions

* `ReplayIntegration.start()` (and `stop`/`pause`/`resume`) run on the main thread via `LifecycleWatcher`; acquiring a lock there that can be held across a **blocking native** `MediaCodec` **call** turns any codec hang into a whole-app deadlock. Either move the work off the main thread, use `tryLock` with a short timeout, or never hold the replay lock across encoder drain/release.
* `SimpleVideoEncoder.drainCodec` passes an infinite/long timeout to `dequeueOutputBuffer` in the `endOfStream`/release path; a bounded timeout with bail-out would make a dead codec merely drop a segment instead of hanging the executor while holding the lock.

Possibly related: getsentry/sentry-java#5529 (encoder kept running while rate-limited).

romtsn:
#### 📜 Description

Fixes an ANR where Android apps freeze on foreground/background transitions until the system kills the process.

A reporter pulled the full deobfuscated 131-thread dump from Play Console (the Sentry ANR event only carries `main`, which is why the lock holder was invisible), giving us the exact chain:

```
main ──waits──▶ lifecycleLock ──held by──▶ timer thread
                                              │ waits
                                              ▼
                                          encoderLock ──held by──▶ SentryReplayIntegration-0
                                                                       │ parked forever
                                                                       ▼
                                                          MediaCodec.dequeueOutputBuffer
```

Two cooperating defects, so two bounds — both are needed:

1. **`SimpleVideoEncoder.drainCodec()`** had a `while (true)` with no exit on the `endOfStream = true` path. Some hardware encoders never emit `BUFFER_FLAG_END_OF_STREAM` after `signalEndOfInputStream()`, so the loop spun forever while holding `encoderLock`. It now gives up after 10 consecutive no-progress iterations (~1s at the existing 100ms poll) and drops the remaining frames.
2. **`ReplayCache.close()`** used an unbounded `encoderLock.acquire()`. It runs inline on the caller's thread from `BaseCaptureStrategy.stop()`, which `ReplayIntegration.stop()` invokes while holding `lifecycleLock` — that nesting is what promotes a stuck encoder into a frozen main thread. It now waits at most 2s and skips the release otherwise.

The loop bound alone isn't sufficient: the native frames show `MediaCodec::dequeueOutputBuffer` → `AMessage::postAndAwaitResponse` → `ALooper::awaitResponse` with `MediaCodec_looper` idle. `TIMEOUT_USEC` is handed *to* the codec looper, which is supposed to reply once it elapses — a dead looper never replies, and `awaitResponse` has no deadline of its own. So a single call can never return, and only bounding the *lock wait* keeps the lifecycle path free.

On the timeout path the (already-dead) codec is not released, leaking a native handle. That's the deliberate trade: leaking a handle in a process whose encoder is wedged beats freezing the app. `isClosed.set(true)` still runs on every path, since `persistSegmentValues` gates on it.

The other four `encoderLock.acquire()` sites in `ReplayCache` are unchanged — they all run on the replay worker, the thread that would be stuck, so a deadline there buys nothing.

Adds `AutoClosableReentrantLock.tryAcquire(timeout, unit)` (`@ApiStatus.Internal`, additive) for the second bound. Returns the token on success and `null` on timeout, so callers must branch explicitly rather than silently no-op via `?.use {}`.

Net effect: a wedged codec degrades to "replay stops working for this process" instead of "the app freezes."

Fixes https://github.com/getsentry/sentry-dart/issues/3556
Fixes https://github.com/getsentry/sentry-java/issues/5870

#### 💚 How did you test it?

`ReplayShadowMediaCodec` gained two hooks, since it ignores `timeoutUs` and always eventually produces EOS: `neverSignalEos` and `blockOnDequeue`. New tests cover both bounds in `ReplayCacheTest`, plus 4 for `tryAcquire` in `AutoClosableReentrantLockTest`.

Both `ReplayCacheTest` cases were mutation-checked — reverting each fix in turn makes the matching test fail with its assertion message rather than hang. Full `:sentry-android-replay:testReleaseUnitTest` passes; `apiDump` adds exactly the one `tryAcquire` line.

No device repro is available (Redmi Note 14 Pro 5G / Pro+ 5G, Android 16, MediaTek), so the shadow-based tests are the verification of record.

#### 📝 Checklist

- [ ] No new PII added or SDK only sends newly added PII if `sendDefaultPii` is enabled.

### DIFF
--- a/sentry-android-replay/build.gradle.kts
+++ b/sentry-android-replay/build.gradle.kts
@@ -83,6 +83,7 @@ dependencies {
   testImplementation(libs.androidx.test.ext.junit)
   testImplementation(libs.androidx.test.runner)
   testImplementation(libs.awaitility.kotlin)
+  testImplementation(libs.google.truth)
   testImplementation(libs.mockito.kotlin)
   testImplementation(libs.mockito.inline)
   testImplementation(libs.androidx.compose.ui)
--- a/sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
+++ b/sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
@@ -300,9 +300,10 @@ public class ReplayCache(private val options: SentryOptions, private val replayI
       }
     } catch (e: InterruptedException) {
       Thread.currentThread().interrupt()
+    } finally {
+      // has to happen on all paths, callers rely on it to stop persisting segment values
+      isClosed.set(true)
     }
-    // has to happen on all paths, callers rely on it to stop persisting segment values
-    isClosed.set(true)
   }
 
   // TODO: it's awful, choose a better serialization format
--- a/sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
+++ b/sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
@@ -22,6 +22,7 @@ import java.io.File
 import java.io.StringReader
 import java.util.Date
 import java.util.LinkedList
+import java.util.concurrent.TimeUnit.MILLISECONDS
 import java.util.concurrent.atomic.AtomicBoolean
 
 /**
@@ -280,10 +281,27 @@ public class ReplayCache(private val options: SentryOptions, private val replayI
   }
 
   override fun close() {
-    encoderLock.acquire().use {
-      encoder?.release()
-      encoder = null
+    // close() is called inline from the lifecycle path (ReplayIntegration.stop/close), which holds
+    // its own lock, so blocking here can freeze the main thread. If the encoder is wedged in a
+    // native MediaCodec call we'd never get the lock, so we give up instead: the already-dead codec
+    // is not released (leaking a native handle), which beats an ANR.
+    try {
+      val token = encoderLock.tryAcquire(ENCODER_RELEASE_TIMEOUT_MS, MILLISECONDS)
+      if (token == null) {
+        options.logger.log(
+          WARNING,
+          "Timed out waiting for the video encoder, skipping its release to not block the caller",
+        )
+      } else {
+        token.use {
+          encoder?.release()
+          encoder = null
+        }
+      }
+    } catch (e: InterruptedException) {
+      Thread.currentThread().interrupt()
     }
+    // has to happen on all paths, callers rely on it to stop persisting segment values
     isClosed.set(true)
   }
 
@@ -314,6 +332,13 @@ public class ReplayCache(private val options: SentryOptions, private val replayI
   }
 
   internal companion object {
+    /**
+     * How long [close] waits for the video encoder to become available. Below Android's ~5s ANR
+     * budget, and above the encoder's own bail-out (see MAX_EOS_STALL_ITERATIONS), so an encoder
+     * that's merely slow is still awaited rather than abandoned.
+     */
+    private const val ENCODER_RELEASE_TIMEOUT_MS = 2000L
+
     internal const val ONGOING_SEGMENT = ".ongoing_segment"
 
     internal const val SEGMENT_KEY_HEIGHT = "config.height"
--- a/sentry-android-replay/src/main/java/io/sentry/android/replay/video/SimpleVideoEncoder.kt
+++ b/sentry-android-replay/src/main/java/io/sentry/android/replay/video/SimpleVideoEncoder.kt
@@ -37,6 +37,7 @@ import android.media.MediaFormat
 import android.os.Build
 import android.view.Surface
 import io.sentry.SentryLevel.DEBUG
+import io.sentry.SentryLevel.WARNING
 import io.sentry.SentryOptions
 import io.sentry.android.replay.util.SystemProperties
 import java.io.File
@@ -301,7 +302,7 @@ internal class SimpleVideoEncoder(
 
       if (stalledIterations >= MAX_EOS_STALL_ITERATIONS) {
         options.logger.log(
-          DEBUG,
+          WARNING,
           "[Encoder]: encoder made no progress for $stalledIterations iterations, dropping the remaining frames",
         )
         break // out of while
--- a/sentry-android-replay/src/main/java/io/sentry/android/replay/video/SimpleVideoEncoder.kt
+++ b/sentry-android-replay/src/main/java/io/sentry/android/replay/video/SimpleVideoEncoder.kt
@@ -45,6 +45,16 @@ import kotlin.LazyThreadSafetyMode.NONE
 
 private const val TIMEOUT_USEC = 100_000L
 
+/**
+ * How many consecutive [MediaCodec.dequeueOutputBuffer] calls may come back without producing
+ * anything before we give up on the encoder. At [TIMEOUT_USEC] per call that's ~1s.
+ *
+ * Some hardware encoders never emit [MediaCodec.BUFFER_FLAG_END_OF_STREAM] after
+ * [MediaCodec.signalEndOfInputStream], which used to spin the drain loop forever while holding the
+ * encoder lock, wedging the whole replay pipeline (and with it the app's lifecycle callbacks).
+ */
+private const val MAX_EOS_STALL_ITERATIONS = 10
+
 @SuppressLint("UseRequiresApi")
 @TargetApi(26)
 internal class SimpleVideoEncoder(
@@ -214,19 +224,26 @@ internal class SimpleVideoEncoder(
       mediaCodec.signalEndOfInputStream()
     }
     var encoderOutputBuffers: Array<ByteBuffer?>? = mediaCodec.outputBuffers
+    // counts consecutive iterations that made no progress, so a codec that never signals EOS can't
+    // spin us forever, see MAX_EOS_STALL_ITERATIONS
+    var stalledIterations = 0
     while (true) {
       val encoderStatus: Int = mediaCodec.dequeueOutputBuffer(bufferInfo, TIMEOUT_USEC)
       if (encoderStatus == MediaCodec.INFO_TRY_AGAIN_LATER) {
         // no output available yet
         if (!endOfStream) {
           break // out of while
-        } else if (options.sessionReplay.isDebug) {
+        }
+        stalledIterations++
+        if (options.sessionReplay.isDebug) {
           options.logger.log(DEBUG, "[Encoder]: no output available, spinning to await EOS")
         }
       } else if (encoderStatus == MediaCodec.INFO_OUTPUT_BUFFERS_CHANGED) {
+        stalledIterations = 0
         // not expected for an encoder
         encoderOutputBuffers = mediaCodec.outputBuffers
       } else if (encoderStatus == MediaCodec.INFO_OUTPUT_FORMAT_CHANGED) {
+        stalledIterations = 0
         // should happen before receiving buffers, and should only happen once
         if (frameMuxer.isStarted()) {
           throw RuntimeException("format changed twice")
@@ -245,8 +262,10 @@ internal class SimpleVideoEncoder(
             "[Encoder]: unexpected result from encoder.dequeueOutputBuffer: $encoderStatus",
           )
         }
-        // let's ignore it
+        // let's ignore it, but still count it as no progress so we can't loop on it forever
+        stalledIterations++
       } else {
+        stalledIterations = 0
         val encodedData =
           encoderOutputBuffers?.get(encoderStatus)
             ?: throw RuntimeException("encoderOutputBuffer $encoderStatus was null")
@@ -278,6 +297,14 @@ internal class SimpleVideoEncoder(
           }
           break // out of while
         }
+      }
+
+      if (stalledIterations >= MAX_EOS_STALL_ITERATIONS) {
+        options.logger.log(
+          DEBUG,
+          "[Encoder]: encoder made no progress for $stalledIterations iterations, dropping the remaining frames",
+        )
+        break // out of while
       }
     }
   }
--- a/sentry-android-replay/src/test/java/io/sentry/android/replay/ReplayCacheTest.kt
+++ b/sentry-android-replay/src/test/java/io/sentry/android/replay/ReplayCacheTest.kt
@@ -66,6 +66,7 @@ class ReplayCacheTest {
     ReplayShadowMediaCodec.neverSignalEos = false
     ReplayShadowMediaCodec.blockOnDequeue = null
     ReplayShadowMediaCodec.blockedOnDequeue = CountDownLatch(1)
+    ReplayShadowMediaCodec.released = false
     ShadowBitmapFactory.setAllowInvalidImageData(true)
   }
 
@@ -720,9 +721,29 @@ class ReplayCacheTest {
       // giving up on the lock still counts as closed, otherwise we'd keep persisting segments
       replayCache.persistSegmentValues(SEGMENT_KEY_ID, "0")
       assertThat(File(replayCache.replayCacheDir, ONGOING_SEGMENT).exists()).isFalse()
+
+      assertWithMessage("encoder should not be released when the lock times out")
+        .that(ReplayShadowMediaCodec.released)
+        .isFalse()
     } finally {
       wedge.countDown()
       encoder.join(SECONDS.toMillis(10))
     }
+  }
+
+  @Test
+  fun `close releases the encoder when the lock is available`() {
+    ReplayShadowMediaCodec.neverSignalEos = true
+    val replayCache = fixture.getSut(tmpDir)
+
+    val bitmap = Bitmap.createBitmap(1, 1, ARGB_8888)
+    replayCache.addFrame(bitmap, 1)
+
+    // createVideoOf bails out via the stall bound, but still releases the encoder
+    replayCache.createVideoOf(1000L, 0L, 0, 100, 200, 1, 20_000)
+
+    assertWithMessage("encoder should be released even when EOS was never signalled")
+      .that(ReplayShadowMediaCodec.released)
+      .isTrue()
   }
 }
--- a/sentry-android-replay/src/test/java/io/sentry/android/replay/ReplayCacheTest.kt
+++ b/sentry-android-replay/src/test/java/io/sentry/android/replay/ReplayCacheTest.kt
@@ -732,14 +732,14 @@ class ReplayCacheTest {
   }
 
   @Test
-  fun `close releases the encoder when the lock is available`() {
+  fun `createVideoOf releases the encoder even when EOS is never signalled`() {
     ReplayShadowMediaCodec.neverSignalEos = true
     val replayCache = fixture.getSut(tmpDir)
 
     val bitmap = Bitmap.createBitmap(1, 1, ARGB_8888)
     replayCache.addFrame(bitmap, 1)
 
-    // createVideoOf bails out via the stall bound, but still releases the encoder
+    // the stall bound breaks the drain loop, but release() must still be called
     replayCache.createVideoOf(1000L, 0L, 0, 100, 200, 1, 20_000)
 
     assertWithMessage("encoder should be released even when EOS was never signalled")
--- a/sentry-android-replay/src/test/java/io/sentry/android/replay/ReplayCacheTest.kt
+++ b/sentry-android-replay/src/test/java/io/sentry/android/replay/ReplayCacheTest.kt
@@ -4,6 +4,8 @@ import android.graphics.Bitmap
 import android.graphics.Bitmap.CompressFormat.JPEG
 import android.graphics.Bitmap.Config.ARGB_8888
 import androidx.test.ext.junit.runners.AndroidJUnit4
+import com.google.common.truth.Truth.assertThat
+import com.google.common.truth.Truth.assertWithMessage
 import io.sentry.DateUtils
 import io.sentry.SentryOptions
 import io.sentry.SentryReplayEvent.ReplayType
@@ -24,7 +26,9 @@ import io.sentry.rrweb.RRWebInteractionEvent.InteractionType.TouchEnd
 import io.sentry.rrweb.RRWebInteractionEvent.InteractionType.TouchStart
 import java.io.File
 import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit.SECONDS
 import java.util.concurrent.atomic.AtomicReference
+import kotlin.concurrent.thread
 import kotlin.test.BeforeTest
 import kotlin.test.Test
 import kotlin.test.assertEquals
@@ -59,6 +63,9 @@ class ReplayCacheTest {
   fun `set up`() {
     ReplayShadowMediaCodec.framesToEncode = 5
     ReplayShadowMediaCodec.throwOnStart = false
+    ReplayShadowMediaCodec.neverSignalEos = false
+    ReplayShadowMediaCodec.blockOnDequeue = null
+    ReplayShadowMediaCodec.blockedOnDequeue = CountDownLatch(1)
     ShadowBitmapFactory.setAllowInvalidImageData(true)
   }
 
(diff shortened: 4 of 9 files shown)


sentry-android-replay/src/main/java/io/sentry/android/replay/video/SimpleVideoEncoder.kt
```diff
@@ -45,6 +45,16 @@ import kotlin.LazyThreadSafetyMode.NONE
...
 
+/**
+ * How many consecutive [MediaCodec.dequeueOutputBuffer] calls may come back without producing
+ * anything before we give up on the encoder. At [TIMEOUT_USEC] per call that's ~1s.
+ *
+ * Some hardware encoders never emit [MediaCodec.BUFFER_FLAG_END_OF_STREAM] after
+ * [MediaCodec.signalEndOfInputStream], which used to spin the drain loop forever while holding the
+ * encoder lock, wedging the whole replay pipeline (and with it the app's lifecycle callbacks).
+ */
+private const val MAX_EOS_STALL_ITERATIONS = 10
```
runningcode, line 57: how did we determine `10` here? Just curious. How fast do these iterations happen?

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
```diff
@@ -280,10 +281,27 @@ public class ReplayCache(private val options: SentryOptions, private val replayI
...
+          "Timed out waiting for the video encoder, skipping its release to not block the caller",
+        )
+      } else {
+        token.use {
+          encoder?.release()
+          encoder = null
+        }
+      }
+    } catch (e: InterruptedException) {
+      Thread.currentThread().interrupt()
```
runningcode, line 302: do we need to surround this with `try/catch` as well?

sentry-android-replay/build.gradle.kts
```diff
@@ -83,6 +83,7 @@ dependencies {
   testImplementation(libs.androidx.test.ext.junit)
   testImplementation(libs.androidx.test.runner)
   testImplementation(libs.awaitility.kotlin)
+  testImplementation(libs.google.truth)
```
runningcode, line 86: nice!

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
```diff
@@ -280,10 +281,27 @@ public class ReplayCache(private val options: SentryOptions, private val replayI
...
-      encoder = null
+    // close() is called inline from the lifecycle path (ReplayIntegration.stop/close), which holds
+    // its own lock, so blocking here can freeze the main thread. If the encoder is wedged in a
+    // native MediaCodec call we'd never get the lock, so we give up instead: the already-dead codec
+    // is not released (leaking a native handle), which beats an ANR.
+    try {
+      val token = encoderLock.tryAcquire(ENCODER_RELEASE_TIMEOUT_MS, MILLISECONDS)
+      if (token == null) {
+        options.logger.log(
+          WARNING,
```
runningcode, line 292: ok so just to check my understanding, we try to get the lock only to make sure nobody else is holding it before calling `release` but if we can't acquire the lock, what state are we in? is this safe to skip the `release`? won't we cause a memory leak?

runningcode (commented):
looks good, i left a question for a discussion!

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
```diff
@@ -280,10 +281,27 @@ public class ReplayCache(private val options: SentryOptions, private val replayI
   }
 
   override fun close() {
-    encoderLock.acquire().use {
-      encoder?.release()
-      encoder = null
+    // close() is called inline from the lifecycle path (ReplayIntegration.stop/close), which holds
```
0xadam-brown, line 284: m: Thoughts about lifting this decision to the call site? That'd let us reduce how much we peg ReplayCache's implementation to its current caller. 

Eg, we could:

1. create a new ReplayCache.tryClose() method and have ReplayIntegration call it instead; or
2. add a new ShutdownMode param to close() and have ReplayIntegration choose BEST_EFFORT. 


*******************

**Long-term side note:** If we had a general pattern where, say, the Integration owned the execution context / threading decisions, we could leave everything beneath it thread _**un**_safe on the assumption that our Integrations would serialize execution via thread confinement by default. That'd make the bulk of our integration code much easier to implement and reason about. 

I haven't looked in detail, but seems like that sort of pattern should be possible, given the very task-oriented nature of our processing + the fact that the host app needs very little from the SDK in real-time.

sentry-android-replay/src/test/java/io/sentry/android/replay/ReplayCacheTest.kt
```diff
@@ -654,4 +661,68 @@ class ReplayCacheTest {
...
+        .isTrue()
+
+      // giving up on the lock still counts as closed, otherwise we'd keep persisting segments
+      replayCache.persistSegmentValues(SEGMENT_KEY_ID, "0")
+      assertThat(File(replayCache.replayCacheDir, ONGOING_SEGMENT).exists()).isFalse()
+    } finally {
+      wedge.countDown()
+      encoder.join(SECONDS.toMillis(10))
+    }
+  }
```
0xadam-brown, line 732: l: Consider adding tests showing that the encoder is or isn't invoked depending on whether lock acquisition times out in ReplayCache.close().

0xadam-brown (approved):
A few quick comments for your consideration (no blockers); otherwise lgtm

runningcode (approved):

sentry-android-replay/src/main/java/io/sentry/android/replay/video/SimpleVideoEncoder.kt
romtsn: honestly just gut feeling (10 x 100ms = 1s wait at most). Unfortunately, I don't have a device at hand to reproduce it, but I will order one to actually check it.

On my two devices that I have (HMD Pulse and Pixel 2XL) each iteration takes exactly around 100ms (the full `TIMEOUT_USEC`), so we're well under a second here.

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
romtsn: this should never throw, but I now set `isClosed` flag to `true` in a `finally` block so we have all the paths covered 👍

sentry-android-replay/src/test/java/io/sentry/android/replay/ReplayCacheTest.kt
romtsn: added in https://github.com/getsentry/sentry-java/pull/5842/commits/6d7152c26fad544acb4130bd724d97344ed78110

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
romtsn: yup, it could cause a memory leak, but that's the trade-off - if the encoder is wedged in the native layer, calling `release` would also hang in the native layer anyway. When the process dies the ref should get reclaimed still, so I guess we pick a lesser evil here.

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
romtsn: There's only one caller (`BaseCaptureStrategy.stop()`), and `ReplayCache.close()` should never block indefinitely regardless of who calls it (otherwise, that's just a broken close). Not sure if adding tryClose() or a ShutdownMode for a single call site would justify it? Besides, `close()` is actually inherited from `Closeable`, so not sure if we should diverge from the known contract.

That said, the longer-term plan (your side-note about integration owning threading) is exactly where we're headed - I've updated #5847. Once lifecycle methods are confined to main thread, `cache.close()` runs on the replay executor (same thread as the encoder), so `encoderLock` goes away entirely and this whole question disappears. I'm already in the process of implementing that, so I'd rather not poke this PR any further, if you don't mind :)

PR #5851, opened 2026-07-29
perf: Drop pending timer tasks on shutdown to unblock close (JAVA-653)

runningcode:
This is a follow up on this comment: https://github.com/getsentry/sentry-java/pull/5814/changes#r3667006204

The timer executor is shut down in `Scopes.close()` via `shutdown()` followed by `awaitTermination(shutdownTimeoutMillis)`. `ScheduledThreadPoolExecutor` keeps queued delayed tasks across `shutdown()` by default, so `awaitTermination` blocks for the **full shutdown timeout (2s by default)** whenever a long timeout is still pending at close time:

- an unfinished transaction's idle/deadline timer (`SentryTracer`)
- the 30s `LifecycleWatcher` end-session task, only cancelled on foregrounding
- a rate limit lifted notification (`RateLimiter`, up to `retry_after`)

Setting `executeExistingDelayedTasksAfterShutdownPolicy(false)` drops those tasks at `shutdown()` instead. This is behaviour preserving: today they are discarded by the subsequent `shutdownNow()` once the timeout expires anyway, so the only thing lost is the wait.

The policy only affects `shutdown()`, so the SDK restart path that intentionally leaves the timer executor running (`Scopes.close(isRestarting = true)`) is unaffected. Scoped to the timer executor's constructor — the general-purpose executor keeps the default.

Measured with a standalone repro using the same executor configuration, one pending 60s task and a 2s timeout:

| | `terminated` | blocked |
|---|---|---|
| before | `false` | 2000ms |
| after | `true` | 0ms |

### DIFF
--- a/sentry/src/main/java/io/sentry/SentryExecutorService.java
+++ b/sentry/src/main/java/io/sentry/SentryExecutorService.java
@@ -55,6 +55,10 @@ public final class SentryExecutorService implements ISentryExecutorService {
     executorService.setRemoveOnCancelPolicy(removeOnCancelPolicy);
     executorService.setKeepAliveTime(keepAliveTime, keepAliveTimeUnit);
     executorService.allowCoreThreadTimeOut(true);
+    // by default shutdown() keeps queued delayed tasks, so awaitTermination blocks for the full
+    // shutdown timeout whenever a long timeout is still pending. Those tasks are discarded by the
+    // subsequent shutdownNow() anyway, so dropping them upfront only saves the wait.
+    executorService.setExecuteExistingDelayedTasksAfterShutdownPolicy(false);
   }
 
   public SentryExecutorService() {
--- a/sentry/src/test/java/io/sentry/SentryExecutorServiceTest.kt
+++ b/sentry/src/test/java/io/sentry/SentryExecutorServiceTest.kt
@@ -8,7 +8,6 @@ import java.util.concurrent.LinkedBlockingQueue
 import java.util.concurrent.ScheduledThreadPoolExecutor
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicBoolean
-import kotlin.system.measureTimeMillis
 import kotlin.test.Test
 import kotlin.test.assertEquals
 import kotlin.test.assertFailsWith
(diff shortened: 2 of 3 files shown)


sentry/src/test/java/io/sentry/SentryExecutorServiceTest.kt
```diff
@@ -110,6 +111,27 @@ class SentryExecutorServiceTest {
     sentryExecutor.close(15000)
   }
 
+  @Test
+  fun `SentryExecutorService discards pending delayed tasks on shutdown when requested`() {
```
runningcode: Not sure we really need these tests.

sentry/src/main/java/io/sentry/SentryExecutorService.java
```diff
@@ -56,6 +56,10 @@ public SentryExecutorService(final @Nullable SentryOptions options) {
     executorService.setRemoveOnCancelPolicy(removeOnCancelPolicy);
     executorService.setKeepAliveTime(keepAliveTime, keepAliveTimeUnit);
     executorService.allowCoreThreadTimeOut(true);
+    // by default shutdown() keeps queued delayed tasks, so awaitTermination blocks for the full
+    // shutdown timeout whenever a long timeout is still pending. Those tasks are discarded by the
+    // subsequent shutdownNow() anyway, so dropping them upfront only saves the wait.
```
0xadam-brown, line 60: 👍

sentry/src/test/java/io/sentry/SentryExecutorServiceTest.kt
0xadam-brown: I think the inline comment is enough to prevent regressions, so my vote would be to remove...

0xadam-brown (approved):
Nice!

Does it make sense to apply the same policy to other / all of our Sentry executor services? (Doesn't have to stop us from merging this PR)

romtsn (approved):

sentry/src/test/java/io/sentry/SentryExecutorServiceTest.kt
runningcode: removed!

PR #5852, opened 2026-07-29
fix(anr): Use Proguard ID from origination ANR process with ANR profie chunks

0xadam-brown:
## :scroll: Description

Prior to this PR, we'd always (rightly) bind originating process Proguard IDs to ANR events, but we'd (wrongly) bind current Proguard ID to ANR profile chunks. Usually the originating Proguard ID == the current Proguard ID, and so the discrepancy didn't matter. But the two differ in the situation when a user updates the host app after the ANR occurs but before it's reported. When that happens, deobfuscation for ANR profile chunks in the Sentry UI can break.

This PR ensures both the ANR event and the ANR profile chunk receive the same originating Proguard ID by: 

1. having the entity that determines the originating Proguard ID (ApplicationExitInfoEventProcessor) pass that ID to the profile chunk pipeline; and
2. updating our DebugMeta.buildDebugMeta(ProfileChunk, ...) method so that the profile chunk pipeline honors (rather than clobbers) the Proguard ID the event processor bakes into the DebugMeta owned by the incoming profile chunk.

_(Sorry for the wall of text ahead, but it's worth a quick read! Yours truly wrote it to make sure my LLM's approach was sane, so it should be easy to follow. It lays out the problem in detail and our possible options – including closing this PR and living with the edge case.)_

### Background

We report ANRs from app processs launched _after_ each ANR occurs. Usually the same app installation is used with both processes and so the Proguard ID used to deobfuscate stack traces is the same for each. But not always! A user could have installed a new version of the app in between, so we need to make sure we grab the Proguard ID associated with the original process. 

Nicely, logic to do that already exists in ApplicationExitInfoEventProcessor ([link](https://github.com/getsentry/sentry-java/blob/a30048b0fc30280201fdba44a7ab4a91ad98c80a/sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java#L584)). Not so nicely, the Proguard ID produced by that logic isn't used by the profile chunk we associate with the ANR event – or at least it wasn't prior to this PR. (The profile chunk would always use the Proguard ID for the current process.)

The upshot is that app owners would see properly deobfuscated stack traces for the ANR event but possibly broken / still-obfuscated stack traces for the profile chunk in the rare case that someone installs a new version of the host app before the ANR can be reported out.

### So what's going on?

ANRs are composites of two separate entities: an ANR event proper and an ANR profile chunk. That distinction goes all the way down, as the Sentry UI displays ANR events and profile chunks in different dashboards. Each entity owns a DebugMeta instance, which is the data structure used to associate a Proguard ID with the entity it belongs to. 

In the SDK, the ApplicationExitInfoEventProcessor owns the pipeline for constructing both the ANR event and its profile chunk. It goes like this:

0. User launches the app after an ANR was recorded.
1. ApplicationExitInfoEventProcessor recreates the ANR event proper + creates DebugMeta for the event with the Proguard ID from the originating process.
2. ApplicationExitInfoEventProcessor recreates the profile chunk associated with the ANR and then passes it on to SentryClient.captureProfileChunk(). 
3. SentryClient creates DebugMeta for the chunk with the Proguard ID from the current process (which is usually fine but is wrong in our edge case). SentryClient then sends the profile chunk to Relay.
5. ApplicationExitInfoEventProcessor binds the profile chunk ID to the ANR event and then sends the ANR event to Relay.

Step (2) receives the DebugMeta instance with the correct Proguard ID from ApplicationExitInfoEventProcessor. But our profile chunk processing ignores it and instead selects the Proguard ID from the current process. That’s because ApplicationExitInfoEventProcessor delegates to the existing profile chunk pipeline used for non-ANR profiles, and non-ANR profiles are always concerned with the current app process. (Note that the profile chunk pipeline is shared by the Java. Ideally that’d mean it’s Android-agnostic, but that isn’t quite the case atm, eg, it knows about Proguard IDs.)

### What should we do?

Possible approaches:

**[A]** Introduce the concept of “originating process" into our profile chunk pipeline.

**[B]** Update the profile chunk API to accept a pre-selected Proguard ID (whether directly as a string param, via a Context/Hint-esque parameter, or by lookup).

**[C]** Include the originating Proguard ID inside the chunk already passed to our profile chunk pipeline by ApplicationExitInfoEventProcessor + update DebugMeta.builder(incomingAnrDebugMeta, …) so that it honors that ID rather than clobbering it.

**[D]** Do nothing / close this PR and live with the edge case of potentially (but not inevitably) broken or misleading ANR profile chunk stack traces.

### The path I chose

This PR implements [C]. 

Approach [A] sounded best in the abstract, but it’d be a big lift as a number of the inputs ApplicationExitInfoEventProcessor relies on to decide which Proguard ID to use are either Android-specific or aren’t currently available to our profile chunk pipeline. Option [B] was a more explicit version of [C] but at the cost of modifying the public API in ways that are, at best, debatable. 

Option [D] is legit, esp if you look at this PR and think the fix is more of a maintenance burden than it’s worth. (Chime in if you do!)

Addresses: [JAVA-549](https://linear.app/getsentry/issue/JAVA-549/deobfuscate-anr-profile-stack-frames)

Unit tests.

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/test/java/io/sentry/android/core/ApplicationExitInfoEventProcessorTest.kt
+++ b/sentry-android-core/src/test/java/io/sentry/android/core/ApplicationExitInfoEventProcessorTest.kt
@@ -11,6 +11,7 @@ import io.sentry.Hint
 import io.sentry.IScopes
 import io.sentry.IpAddressUtils
 import io.sentry.NoOpLogger
+import io.sentry.ProfileChunk
 import io.sentry.Sentry
 import io.sentry.SentryBaseEvent
 import io.sentry.SentryEvent
@@ -75,7 +76,9 @@ import org.junit.rules.TemporaryFolder
 import org.junit.runner.RunWith
 import org.mockito.Mockito.mockStatic
 import org.mockito.kotlin.any
+import org.mockito.kotlin.argumentCaptor
 import org.mockito.kotlin.mock
+import org.mockito.kotlin.verify
 import org.mockito.kotlin.whenever
 import org.robolectric.annotation.Config
 import org.robolectric.shadow.api.Shadow
@@ -1012,6 +1015,60 @@ class ApplicationExitInfoEventProcessorTest {
 
       assertNotNull(processed?.contexts?.profile)
       assertNotNull(processed.contexts.profile?.profilerId)
+    }
+  }
+
+  @Test
+  fun `uses persisted proguard uuid for ANR profile chunk after app update`() {
+    fixture.options.anrProfilingSampleRate = 1.0
+    fixture.options.proguardUuid = "current-uuid"
+    val processor =
+      fixture.getSut(
+        tmpDir,
+        populateScopeCache = false,
+        populateOptionsCache = false,
+        isSendDefaultPii = false,
+      )
+    fixture.persistOptions(PROGUARD_UUID_FILENAME, "previous-uuid")
+    setLastUpdateTime(2_000)
+
+    val hint =
+      HintUtils.createWithTypeCheckHint(
+        AbnormalExitHint(mechanism = "anr_foreground", timestamp = 1_000)
+      )
+
+    AnrProfileManager(
+        fixture.options,
+        AnrProfileRotationHelper.getFileForRecording(File(fixture.options.cacheDirPath!!)),
+      )
+      .apply {
+        add(
+          AnrStackTrace(
+            1_000,
+            arrayOf(
+              StackTraceElement("com.example.MyApp", "blocked", "MyApp.java", 42),
+              StackTraceElement("android.os.Handler", "dispatchMessage", "Handler.java", 5678),
+            ),
+          )
+        )
+        close()
+      }
+    AnrProfileRotationHelper.rotate()
+
+    val scopes = mock<IScopes>()
+    whenever(scopes.captureProfileChunk(any())).thenReturn(SentryId())
+
+    mockStatic(Sentry::class.java).use { mockedSentry ->
+      mockedSentry.`when`<Any> { Sentry.getCurrentScopes() }.thenReturn(scopes)
+
+      processor.process(SentryEvent(), hint)
+
+      val chunkCaptor = argumentCaptor<ProfileChunk>()
+      verify(scopes).captureProfileChunk(chunkCaptor.capture())
+      val images = chunkCaptor.firstValue.debugMeta!!.images!!
+      assertEquals(1, images.size)
+      assertEquals(DebugImage.PROGUARD, images[0].type)
+      assertEquals("previous-uuid", images[0].uuid)
     }
   }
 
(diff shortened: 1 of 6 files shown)


sentry/src/main/java/io/sentry/protocol/DebugMeta.java
```diff
@@ -57,6 +57,19 @@ public void setSdkInfo(final @Nullable SdkInfo sdkInfo) {
   @ApiStatus.Internal
   public static @Nullable DebugMeta buildDebugMeta(
       final @Nullable DebugMeta eventDebugMeta, final @NotNull SentryOptions options) {
+    final @NotNull List<DebugImage> optionDebugImages = createDebugImagesFromOptions(options);
+
+    if (eventDebugMeta == null && optionDebugImages.isEmpty()) {
+      return null;
+    }
+
+    DebugMeta debugMeta = eventDebugMeta != null ? eventDebugMeta : new DebugMeta();
```
markushi: `nit`

```suggestion
    final @NotNull DebugMeta debugMeta = eventDebugMeta != null ? eventDebugMeta : new DebugMeta();
```

sentry/src/main/java/io/sentry/protocol/DebugMeta.java
```diff
@@ -73,21 +86,56 @@ public void setSdkInfo(final @Nullable SdkInfo sdkInfo) {
...
+        return false;
       }
+    }
+    return true;
+  }
 
-      return debugMeta;
+  private static boolean isMatchingDebugImage(
+      final @NotNull DebugImage image, final @NotNull DebugImage candidate) {
+    // There can only be one ProGuard mapping per payload, so an existing ProGuard image takes
```
markushi, line 126: 💯

sentry-android-core/src/test/java/io/sentry/android/core/ApplicationExitInfoEventProcessorTest.kt
```diff
@@ -1015,6 +1018,60 @@ class ApplicationExitInfoEventProcessorTest {
...
+      )
+    fixture.persistOptions(PROGUARD_UUID_FILENAME, "previous-uuid")
+    setLastUpdateTime(2_000)
+
+    val hint =
+      HintUtils.createWithTypeCheckHint(
+        AbnormalExitHint(mechanism = "anr_foreground", timestamp = 1_000)
+      )
+
+    AnrProfileManager(
```
markushi, line 1040: I feel bad that we have no easier way of mocking an ANR profile, but I guess there's no better way of doing this right now 🙈

markushi (approved):
LGTM!

sentry/src/main/java/io/sentry/protocol/DebugMeta.java
0xadam-brown: Ah, good catch. Remembering my Java skills slowly...

PR #5866, opened 2026-07-29
perf(android): Move start-reason binder off main thread (JAVA-616)

### LINKED ISSUE #5702: Defer the API 35+ getHistoricalProcessStartReasons binder call off the main thread
## Problem

`AppStartMetrics.registerLifecycleCallbacks` performs an `ActivityManager.getHistoricalProcessStartReasons(1)` binder call synchronously on the main thread on API 35+ (`sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java:467-475`). Under auto-init this runs at ContentProvider time via `SentryPerformanceProvider.onAppLaunched`. getsentry/sentry-java#5275 already restricted it to API 35+, but on those devices it still blocks the main thread on a system_server round-trip during the coldest part of startup.

## Proposal

The result (`appStartType` / `cachedStartInfo`) is only consumed later when app-start spans are assembled — investigate moving the lookup to a background thread/executor with the consumer reading a Future or volatile, or deferring until first frame.

## Risks

* `registerLifecycleCallbacks` runs very early (ContentProvider), before the Sentry executor exists — may need a lazy or Handler-based defer rather than `options.getExecutorService()`.

runningcode:
## 📜 Description

On API 35+, `AppStartMetrics.registerLifecycleCallbacks` performed an `ActivityManager.getHistoricalProcessStartReasons(1)` binder call **synchronously on the main thread**. Under auto-init this runs at `ContentProvider` time (`SentryPerformanceProvider.onAppLaunched`), blocking the main thread on a `system_server` round-trip during the coldest part of app start.

This moves the lookup onto a short-lived background daemon thread. The Sentry executor doesn't exist this early (ContentProvider time), so a plain thread is used. `cachedStartInfo` is now `volatile` and consumed best-effort:

* Cold/warm classification moves into `onActivityCreated`: it prefers the resolved `ApplicationStartInfo` and falls back to the existing pre-API-35 heuristic if the lookup hasn't resolved yet.
* The headless-start path consults the resolved start info before defaulting to `COLD`.
* `getAppStartReason()` returns `null` if the lookup hasn't resolved by the time it is read.

No public API change (`apiDump` clean).

## 💡 Motivation and Context

getsentry/sentry-java#5702. Fixes getsentry/sentry-java#5702.

The binder call stalled the main thread at the most latency-sensitive moment of startup. Measured on a Pixel 10 (API 36) across cold starts, the synchronous call cost **\~0.5–2.2 ms of main-thread time** (median \~0.7 ms, mean \~0.9 ms). Deferring it reclaims that time from the main thread.

## 💚 How did you test it?

* `AppStartMetricsTestApi35` unit tests updated to inject a synchronous executor and drive `onActivityCreated`; all pass.
* On-device (Pixel 10, API 36, cold starts via force-stop → launch): **before** ≈ 0.5–2.2 ms on the main thread (median \~0.7 ms); **after**, the same work runs on the background thread and `cachedStartInfo` was ready before the first activity in every observed cold start — even with the sample's artificial start-padding removed — so cold/warm and `app.start.reason` are preserved in practice.

## :pencil: Checklist

## 🔮 Next steps

* Consider a follow-up to reuse a shared background executor once one is available that early, instead of a one-off thread.

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
@@ -1,5 +1,6 @@
 package io.sentry.android.core.performance;
 
+import android.annotation.SuppressLint;
 import android.app.Activity;
 import android.app.ActivityManager;
 import android.app.Application;
@@ -35,6 +36,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -70,6 +72,13 @@ public class AppStartMetrics extends ActivityLifecycleCallbacksAdapter {
 
   private @NotNull AppStartType appStartType = AppStartType.UNKNOWN;
   private @Nullable volatile Boolean appLaunchedInForeground;
+  // True/false once the deferred ApplicationStartInfo start reason says whether the OS spawned this
+  // process for a user interaction; null while unresolved or when the reason is inconclusive.
+  // Written from the lookup thread, read from the main thread.
+  private @Nullable volatile Boolean startReasonForeground;
+  // Set once an Activity or the headless check has observed the actual launch. Those signals beat
+  // the start reason, so it stops being consulted from then on.
+  private volatile boolean appLaunchedInForegroundObserved;
   private volatile long firstIdle = -1;
 
   private final @NotNull TimeSpan appStartSpan;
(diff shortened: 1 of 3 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
```diff
@@ -99,7 +101,17 @@ public enum AppStartType {
   private @Nullable String appStartSentryTraceHeader;
   private @Nullable String appStartBaggageHeader;
   private @Nullable SentryDate appStartEndTime;
-  private @Nullable ApplicationStartInfo cachedStartInfo;
+  // Volatile: written from the background lookup thread, read from the main thread. Null until the
+  // deferred getHistoricalProcessStartReasons lookup resolves.
+  private volatile @Nullable ApplicationStartInfo cachedStartInfo;
+  // Runs the getHistoricalProcessStartReasons binder call off the main thread. The Sentry executor
+  // does not exist this early (ContentProvider time), so a plain daemon thread is used by default.
+  private @NotNull Executor startInfoExecutor =
```
markushi, line 109: We could have the executor as a local variable and immediately call shut down after enqueuing [like here](https://github.com/getsentry/sentry-java/blob/83277d1441dc432f7bd542b19fcb0d3725eb97a2/sentry-android-core/src/main/java/io/sentry/android/core/DefaultAndroidEventProcessor.java#L56-L70), to avoid any dangling executors.

sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
```diff
@@ -99,7 +101,17 @@ public enum AppStartType {
   private @Nullable String appStartSentryTraceHeader;
   private @Nullable String appStartBaggageHeader;
   private @Nullable SentryDate appStartEndTime;
-  private @Nullable ApplicationStartInfo cachedStartInfo;
+  // Volatile: written from the background lookup thread, read from the main thread. Null until the
+  // deferred getHistoricalProcessStartReasons lookup resolves.
+  private volatile @Nullable ApplicationStartInfo cachedStartInfo;
```
markushi, line 106: How about turning this into a `LazyEvaluator<ApplicationStartInfo> cachedStartInfo`? Then the executor can simple call get(), to trigger evaluation off the main thread.

runningcode:
@markushi thanks for the feedback on this, with the changes to main that created merge conflicts and the fact that this can cause a timing issue on startup I will close this.

Issue #5709, opened 2026-07-03
Lazy-allocate the breadcrumb queue in Scope


runningcode:
## Problem

Each `Scope` preallocates its breadcrumb ring buffer at `maxBreadcrumbs` capacity (default 100) in the constructor (`createBreadcrumbsList` in `sentry/src/main/java/io/sentry/Scope.java`; the vendored `CircularFifoQueue` allocates its backing array eagerly). Init alone creates three scopes — global (`Sentry.java:68`), root + isolation (`Sentry.java:339-340`) — and scope forking copies pay the same, so the arrays multiply across the process lifetime even when breadcrumbs are disabled (`maxBreadcrumbs = 0` is handled, but the default path always allocates).

## Proposal

Allocate the queue on first `addBreadcrumb`. Check the same pattern for other per-scope members (`FeatureFlagBuffer`, `attachments`, `eventProcessors`) — several could be empty-singleton until first write.

## Constraints

* `Scope.clone`/fork paths and the synchronized queue wrapper need to handle the not-yet-allocated state without racing; keep the fast path allocation-free.

Allocation/GC-pressure win; verify with an ART allocation trace (below cold-start macrobenchmark noise individually).

Issue #5708, opened 2026-07-03
Avoid constructing a full SentryOptions for the static globalScope


runningcode:
## Problem

`Sentry`'s static initializer creates `globalScope = new Scope(SentryOptions.empty())` (`sentry/src/main/java/io/sentry/Sentry.java:68`) at class-load time — a complete `SentryOptions` (sub-options objects, ~10 concurrent collections, plus the two default integrations added in the ctor) and a full `Scope` (preallocated breadcrumb ring, ~14 collections, locks). All of it is superseded moments later when `Sentry.init` calls `globalScope.replaceOptions(options)` (`Sentry.java:337`).

Under auto-init this allocation happens on the main thread at ContentProvider time, before `Application.onCreate`.

## Proposal

Use a shared minimal/immutable empty-options singleton, or defer `globalScope` construction until first access (lazy holder). Pure allocation win at class-load; combine with the lazy sub-options investigation.

## Constraints

* `SentryOptions.empty()` semantics: some code paths rely on a mutable options instance pre-init (e.g. `globalScope` observers?); audit `Scope` usage before `replaceOptions`.

Issue #5707, opened 2026-07-03
Lazy-allocate SentryOptions sub-options objects


runningcode:
## Problem

Every `SentryOptions` eagerly constructs `SentryReplayOptions`, `SentryFeedbackOptions`, `ExperimentalOptions`, logs/metrics sub-options, `ClientReportRecorder`, etc. (field initializers + ctor in `sentry/src/main/java/io/sentry/SentryOptions.java`). This is paid even by throwaway instances:

* the static `globalScope` uses `SentryOptions.empty()` (`Sentry.java:68`),
* `SentryPerformanceProvider` builds a full options object just to deserialize the app-start profiling config (`sentry-android-core/.../SentryPerformanceProvider.java:122`).

## Proposal

Investigate lazy holders for sub-options (allocate on first getter access). Manifest parsing touches several of them during init, so the win materializes mainly for secondary options instances and disabled features — measure the allocation delta with an ART allocation trace before committing.

## Constraints

* Getters are called from multiple threads post-init; lazy init must be thread-safe without adding per-call lock overhead (e.g. double-checked volatile, or the existing `LazyEvaluator` — but note getsentry/sentry-java#5642/getsentry/sentry-java#5644 reduced per-instance lock churn; don't reintroduce it).

PR #5867, opened 2026-07-29
perf(android): Parse app start profiling config without JsonSerializer

runningcode:
## 📜 Description

`JsonSerializer` is a massive object with 60 serializers but we only need one of them in `SentryPerformanceProvider`. We can't re-use the one held in `SentryOptions` because the `SentryPerformanceProvider` content provider has a higher priority than the `SentryInitProvider`.  <br>So instead we create a serializer with only the serializer we need for this specific purpose.

Measured on a Pixel 10 (API 37) with androidx Microbenchmark, release build:

|  | allocations | time (median) |
| -- | -- | -- |
| via `JsonSerializer` (before) | 221 | 7548 ns |
| direct `Deserializer` (after) | **33** | **3687 ns** |

`allocationCount` is deterministic, so the allocation figures are exact. The timings are steady-state (JIT-warmed, clocks unlocked), so they understate the real cold-start saving — warmup amortizes away the class loading and verification of all those deserializer classes, which the real one-shot path pays in full. No claim is made about measurable end-to-end startup improvement; \~3.9 µs is small against a \~1 s cold start. The point is main-thread work removed from before `Application.onCreate`.

## 💡 Motivation and Context

Perf improvement

Related: getsentry/sentry-java#5708 and getsentry/sentry-java#5709 were previously closed for the same "small allocation win, not worth the trade off" reason.

GH Issue: getsentry/sentry-java#5707  <br>Linear: getsentry/sentry-java#5707

## 💚 How did you test it?

Unit tests

## :pencil: Checklist

## 🔮 Next steps

`createAndStartContinuousProfiler` also calls `SentryOptions.empty()`, for `TracesSampler`. That one genuinely needs a mutable options object (`setProfileSessionSampleRate`) and only runs when a sampled continuous-profiling config exists, so it is deliberately left alone.

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/SentryPerformanceProvider.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/SentryPerformanceProvider.java
@@ -13,7 +13,7 @@ import io.sentry.IContinuousProfiler;
 import io.sentry.ILogger;
 import io.sentry.ISentryLifecycleToken;
 import io.sentry.ITransactionProfiler;
-import io.sentry.JsonSerializer;
+import io.sentry.JsonObjectReader;
 import io.sentry.SentryAppStartProfilingOptions;
 import io.sentry.SentryExecutorService;
 import io.sentry.SentryLevel;
@@ -117,8 +117,7 @@ public final class SentryPerformanceProvider extends EmptySecureContentProvider 
     try (final @NotNull Reader reader =
             new BufferedReader(new InputStreamReader(new FileInputStream(configFile)))) {
       final @Nullable SentryAppStartProfilingOptions profilingOptions =
-          new JsonSerializer(SentryOptions.empty())
-              .deserialize(reader, SentryAppStartProfilingOptions.class);
+          deserializeProfilingConfig(reader);
 
       if (profilingOptions == null) {
         logger.log(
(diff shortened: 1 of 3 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/SentryPerformanceProvider.java
```diff
@@ -166,6 +165,29 @@ private void launchAppStartProfiler(final @NotNull AppStartMetrics appStartMetri
...
   }
 
+  /**
+   * Parses the app start profiling config with only the deserializer it needs. Going through {@link
+   * io.sentry.JsonSerializer} would allocate a full {@link SentryOptions} plus every registered
+   * deserializer on the main thread before {@code Application.onCreate}, to use exactly one of
+   * them.
+   *
+   * <p>Returns null on malformed input, matching what {@code JsonSerializer.deserialize} did, so
+   * callers keep reporting it as a deserialization failure rather than a read error. The vendored
```
0xadam-brown: nit if you want it: Personally I'd either inline "The vendored..." above the catch clause, or drop it altogether. (The latter makes sense to me as we're all aware of the virtues of not catching `Error`s.)

0xadam-brown (approved):
Ftw!

romtsn (approved):
nice!

sentry-android-core/src/main/java/io/sentry/android/core/SentryPerformanceProvider.java
runningcode: thank you. simplified

PR #5872, opened 2026-07-30
fix(android): Prevent Relay from rejecting single-sample ANR profiles 

0xadam-brown:
## :scroll: Description

Relay currently rejects ANR profiles with a single sample, even though they're still useful to customers. This PR updates our StackTraceConverter to add a second, synthetic sample whenever an ANR profile would otherwise only contain one.

The synthetic sample duplicates the original, save for a minor offset to its timestamp. (Lets us avoid misleading the user about the ANRs cause or duration.)

Customers occasionally see ANR events with no attached profile despite having sufficient quota and ANR profiling enabled. That's because Relay drops any profile chunk with fewer than two samples, and an ANR chunk can have exactly one sample if the main thread is caught on the very first poll after the 1-second suspicion threshold. 

Relay's policy was developed with continuous profiles in mind, where a one-point series has no meaningful time axis. But for ANR profiles a single sample is highly actionable, as it shows exactly where the main thread was stuck when the ANR began.

Resolves: [JAVA-550](https://linear.app/getsentry/issue/JAVA-550/prevent-single-sample-anr-profile-chunks-from-being-silently-dropped)

## Options considered

**[A] Make it more likely that the SDK satisfies Relay's requirements** -> In particular, narrow the polling interval used with ANRs. Would only reduce the risk of single-sample ANRs but not eliminate it + requires more resources from host app.

**[B] Guarantee the SDK satisfies Relay's requirements** -> The approach this PR takes.

**[C] Relax Relay's requirements** ->  Too big of a lift to implement and would require substantial input by the Relay folks.

Unit tests + integration test + had my LLM check E2E against Relay and entire backend pipeline.

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/anr/StackTraceConverter.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/anr/StackTraceConverter.java
@@ -174,11 +174,8 @@ public final class StackTraceConverter {
    */
   @NotNull
   private static SentrySample createSyntheticSample(@NotNull SentrySample originalSample) {
-    final @NotNull SentrySample syntheticSample = new SentrySample();
+    final @NotNull SentrySample syntheticSample = new SentrySample(originalSample);
     syntheticSample.setTimestamp(originalSample.getTimestamp() + SYNTHETIC_SAMPLE_OFFSET_SECONDS);
-    syntheticSample.setStackId(originalSample.getStackId());
-    syntheticSample.setThreadId(originalSample.getThreadId());
-    syntheticSample.setUnknown(originalSample.getUnknown());
     return syntheticSample;
   }
 }
--- a/sentry-android-core/src/main/java/io/sentry/android/core/anr/StackTraceConverter.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/anr/StackTraceConverter.java
@@ -34,6 +34,14 @@ public final class StackTraceConverter {
   private static final String MAIN_THREAD_NAME = "main";
 
   /**
+   * Timestamp offset used with synthetic ANR profile samples. (Currently 33 ms.)
+   *
+   * <p>Places the synthetic sample halfway to the next ANR polling tick.
+   */
+  private static final double SYNTHETIC_SAMPLE_OFFSET_SECONDS =
+      (AnrProfilingIntegration.POLLING_INTERVAL_MS / 2.0d) / 1000.0d;
+
+  /**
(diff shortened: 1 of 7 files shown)


runningcode:
This seems to me like a server side problem, can we fix this in Relay or would that break other SDKs?

0xadam-brown:
> This seems to me like a server side problem, can we fix this in Relay or would that break other SDKs?

Good callout, and agreed that'd be ideal. As mentioned in the PR description, it'd be a big lift implementation-wise and it'd require a lot of coordination with the Relay folks. We could certainly do it at some point, but this gets us 90% of the way there at super low cost 👍

sentry-android-core/src/main/java/io/sentry/android/core/anr/StackTraceConverter.java
```diff
@@ -33,6 +33,14 @@ public final class StackTraceConverter {
   private static final String MAIN_THREAD_ID = "0";
   private static final String MAIN_THREAD_NAME = "main";
 
+  /**
+   * Timestamp offset used with synthetic ANR profile samples. (Currently 33 ms.)
+   *
+   * <p>Places the synthetic sample halfway to the next ANR polling tick.
```
markushi, line 39: 💡 smart!

sentry-android-core/src/main/java/io/sentry/android/core/anr/StackTraceConverter.java
```diff
@@ -147,4 +164,21 @@ private static SentryStackFrame createSentryStackFrame(@NotNull StackTraceElemen
...
+
+  /**
+   * Creates a {@link SentrySample} identical to {@code originalSample}, save that its timestamp is
+   * advanced by {@link #SYNTHETIC_SAMPLE_OFFSET_SECONDS}.
+   *
+   * <p>Lets us produce a plausible synthetic sample without misleading the user about the ANR's
+   * actual duration or cause.
+   */
+  @NotNull
+  private static SentrySample createSyntheticSample(@NotNull SentrySample originalSample) {
```
markushi, line 176: `nit`: How about doing a copy ctor + a `setTimestamp()` call here? This way the responsibility of setting all attributes stays within `SentrySample`.

markushi (approved):
Awesome, LGTM!

sentry-android-core/src/main/java/io/sentry/android/core/anr/StackTraceConverter.java
0xadam-brown: Good call – updated!

PR #5874, opened 2026-07-30
fix(core): Report no-op executor tasks as cancelled

runningcode:
## :scroll: Description

`NoOpSentryExecutorService` returned `new FutureTask<>(() -> null)` from `submit`/`schedule` — a task that is never run and never cancelled. Callers had no way to distinguish it from a genuinely queued task: `isCancelled()` and `isDone()` both stay `false` forever, and `get()` blocks until its timeout.

It now returns an already-cancelled `Future`, matching what `SentryExecutorService` already does when its work queue is full. `CancelledFuture` moves out of `SentryExecutorService` into a shared package-private class, and `ISentryExecutorService` documents the requirement so future implementations honour it.

No public API change — `CancelledFuture` is package-private and no existing signature changed shape, so `.api` files are untouched.

Found [by cursor here](https://github.com/getsentry/sentry-java/pull/5791#discussion_r3681990250). That PR's `PersistingScopeObserver` uses `future.isCancelled()` to detect a flush task the executor dropped, and releases its `hasPendingFlush` latch when it sees one. Against a no-op executor the latch would be set with no task to ever clear it, silently stopping scope persistence for the rest of the process.

That is not a reachable bug today because the observer is only installed after `options.activate()` has swapped in a real executor so this more to prevent regressions in the future. But `setExecutorService` is public, and the contract the caller relies on was never written down anywhere. Fixing it here, off `main`, keeps #5791 focused and makes the guarantee available to any other caller that wants to check whether its task was accepted.

Tests pass

## :pencil: Checklist

## :crystal_ball: Next steps

#5791 builds on this guarantee for its scope-write batching.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 - Avoid a CPU busy-loop when recording discarded log or metric envelopes under rate limiting ([#5835](https://github.com/getsentry/sentry-java/pull/5835))
   - `ClientReportRecorder` now reads the item count from the envelope item header instead of deserializing the payload, which under sustained rate limiting could pin CPU cores while repeatedly throwing exceptions
 - Report tasks handed to a no-op `ISentryExecutorService` as cancelled ([#5874](https://github.com/getsentry/sentry-java/pull/5874))
-  - `NoOpSentryExecutorService` returned a `Future` that was never run and never cancelled, so callers could not tell a dropped task from a queued one and `get()` would block until its timeout
+  - `NoOpSentryExecutorService` previously returned a `Future` that was never run and never cancelled, so callers could not tell a dropped task from a queued one and `get()` would block until its timeout
 
 ### Performance
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 - Avoid a CPU busy-loop when recording discarded log or metric envelopes under rate limiting ([#5835](https://github.com/getsentry/sentry-java/pull/5835))
   - `ClientReportRecorder` now reads the item count from the envelope item header instead of deserializing the payload, which under sustained rate limiting could pin CPU cores while repeatedly throwing exceptions
+- Report tasks handed to a no-op `ISentryExecutorService` as cancelled ([#5874](https://github.com/getsentry/sentry-java/pull/5874))
+  - `NoOpSentryExecutorService` returned a `Future` that was never run and never cancelled, so callers could not tell a dropped task from a queued one and `get()` would block until its timeout
 
 ### Performance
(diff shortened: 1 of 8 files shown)


markushi (approved):
LGTM!

sentry/src/main/java/io/sentry/ISentryExecutorService.java
```diff
@@ -6,15 +6,23 @@
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 
-/** Sentry Executor Service that sends cached events and envelopes on App. start. */
+/**
+ * Sentry Executor Service that sends cached events and envelopes on App. start.
```
0xadam-brown:
```suggestion
 * Sentry Executor Service that sends cached events and envelopes on App start.
```

sentry/src/main/java/io/sentry/ISentryExecutorService.java
```diff
@@ -6,15 +6,23 @@
...
+ *
+ * <p>Implementations that drop a task instead of running it — a no-op service, a full work queue —
+ * must report that through the returned {@link Future}: either throw {@link
+ * RejectedExecutionException} or return a Future that is already cancelled. Callers rely on this to
```
0xadam-brown:
```suggestion
 * RejectedExecutionException} or return a {@link CancelledFuture}. Callers rely on this to
```

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 - Avoid a CPU busy-loop when recording discarded log or metric envelopes under rate limiting ([#5835](https://github.com/getsentry/sentry-java/pull/5835))
   - `ClientReportRecorder` now reads the item count from the envelope item header instead of deserializing the payload, which under sustained rate limiting could pin CPU cores while repeatedly throwing exceptions
+- Report tasks handed to a no-op `ISentryExecutorService` as cancelled ([#5874](https://github.com/getsentry/sentry-java/pull/5874))
+  - `NoOpSentryExecutorService` returned a `Future` that was never run and never cancelled, so callers could not tell a dropped task from a queued one and `get()` would block until its timeout
```
0xadam-brown:
```suggestion
  - `NoOpSentryExecutorService` previously returned a `Future` that was never run and never cancelled, so callers could not tell a dropped task from a queued one and `get()` would block until its timeout
```

0xadam-brown (approved):
Nice! A few optional nits; otherwise lgtm pending fix of [the SentryBot issue](https://github.com/getsentry/sentry-java/pull/5874#discussion_r3682405988).

PR #5875, opened 2026-07-30
perf(core): Use build-time class availability

romtsn:
## :scroll: Description

Allow `LoadClass.isClassAvailable` to use class-availability results supplied by build tooling before falling back to `Class.forName`. The lookup map is nullable and private, so applications without compatible tooling keep the existing reflection behavior.

Android SDK initialization probes optional integrations reflectively. Build tooling already knows whether their owning modules are present, so it can avoid those runtime checks.

Refs [JAVA-654](https://linear.app/getsentry/issue/JAVA-654/move-reflection-checks-to-compile-time)

- `./gradlew :sentry:test --tests '*LoadClassTest*'`
- Pixel 2 XL microbenchmark: map construction and lookups were about 160× faster than the equivalent absent-heavy 11-class reflection batch.
- In the absent-heavy startup macrobenchmark, SDK initialization was 1.08% faster and full startup was 0.88% faster, though variance exceeded the measured delta. The representative 10-present/1-absent case showed no measurable improvement.

## :pencil: Checklist

## :crystal_ball: Next steps

[getsentry/sentry-android-gradle-plugin#1375](https://github.com/getsentry/sentry-android-gradle-plugin/pull/1375) injects the build-time results.

#skip-changelog

### DIFF
--- a/sentry/src/main/java/io/sentry/util/LoadClass.java
+++ b/sentry/src/main/java/io/sentry/util/LoadClass.java
@@ -13,7 +13,7 @@ import org.jetbrains.annotations.Nullable;
 public class LoadClass {
 
   // Populated by the Sentry Android Gradle plugin for class names it can resolve at build time.
-  private static @Nullable Map<String, Boolean> classAvailability;
+  static @Nullable Map<String, Boolean> classAvailability;
 
   /**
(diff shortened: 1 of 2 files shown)


runningcode (approved):
Seems simple enough!

PR #5880, opened 2026-07-30
fix(app-start): Fix detection of foreground app start

### LINKED ISSUE #4693: Long ui.load transactions exceeding deadlineTimeout for Android Activities
### Integration

sentry-android

### Build System

Gradle

### AGP Version

5.9.0

### Proguard

Enabled

### Version

8.19.1

### Steps to Reproduce

Customer is seeing long ui.load traces that **exceed** the default `deadlineTimeout` defined (30s) 

### Expected Result

Transaction duration should not exceed the defined `deadlineTimeout`

### Actual Result

In certain cases, we're seeing transaction duration exceed `deadlineTimeout`

Some common observations:
- All suspicious long transactions that are ui.load have a `Activity.onStart` but not a `Activity.onCreate`
- None of the suspicious long transactions have a `app.start.cold` span or a `app.start.warm` span

Considering the above, the issue might be that the app goes to the BG before the deadline (and before any other spans are opened) and then it resumes from BG and closes out the open span like an hour later. This could happen if the process starts, but the Activity never gets through onCreate/onStart.

markushi:
## Description

Pre Android 15, `AppStartMetrics.appLaunchedInForeground` uses a `LazyEvaluator<Boolean>` that only
evaluated `ContextUtils.isForegroundImportance()` on the first read. Since the read can happen at a later point in time, e.g. during `Activity.onCreate()` the process importance might have changed by now, and we incorrectly detect foreground / background starts - causing long app starts.

This PR changes the behavior, and we now immediately evaluate the foreground flag. This will make foreground/background app starts pre Android 15 more correct, at the cost of an extra binder call via `ContextUtils.isForegroundImportance()`.

Fixes https://github.com/getsentry/sentry-java/issues/4693 (pre Android 15)

Unit Tests.

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
@@ -534,7 +534,7 @@ public class AppStartMetrics extends ActivityLifecycleCallbacksAdapter {
         }
       }
     }
-    // Fallback, if not matching ApplicationStartInfo is available
+    // Fallback, if no matching ApplicationStartInfo is available
     if (appLaunchedInForeground == null) {
       appLaunchedInForeground = ContextUtils.isForegroundImportance();
     }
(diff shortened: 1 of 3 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
```diff
@@ -543,6 +534,10 @@ public void registerLifecycleCallbacks(final @NotNull Application application) {
         }
       }
     }
+    // Fallback, if not matching ApplicationStartInfo is available
```
0xadam-brown:
```suggestion
    // Fallback, if no matching ApplicationStartInfo is available
```

0xadam-brown (approved):
Nice!

Approving pending addition of the CHANGELOG entry 👍

PR #5883, opened 2026-08-03
fix(android): Reduce SQLite cursor SDK crash false-positives

0xadam-brown:
## :scroll: Description

PR helps us avoid incorrectly attributing host app SQLite crashes to the Sentry SDK in situations where a pass-through cursor method shows up in the crash stack trace. 

In particular, it replaces Kotlin interface delegation in SentryCrossProcessCursor in favor of Android's CursorWrapper. That keeps lazy-query span instrumentation on getCount, onMove, and fillWindow, but avoids generating Sentry-owned methods like getString for pass-through cursor calls.

Reduces false-positive crash rates associated with the Sentry Android SDK. Something we saw a lot of in 8.44.1 and 8.43.3, as mentioned [here](https://sentry.slack.com/archives/G01PW6HRQNN/p1785748167877479?thread_ts=1785741441.590429&cid=G01PW6HRQNN).

## :pencil: Checklist

## :crystal_ball: Next steps

 #skip-changelog

### DIFF
--- a/sentry-android-sqlite/src/main/java/io/sentry/android/sqlite/SentryCrossProcessCursor.kt
+++ b/sentry-android-sqlite/src/main/java/io/sentry/android/sqlite/SentryCrossProcessCursor.kt
@@ -2,20 +2,20 @@ package io.sentry.android.sqlite
 
 import android.database.CrossProcessCursor
 import android.database.CursorWindow
+import android.database.CursorWrapper
 
 /*
  * SQLiteCursor executes the query lazily, when one of getCount() and onMove() is called.
  * Also, by docs, fillWindow() can be used to fill the cursor with data.
  * So we wrap these methods to create a span.
- * SQLiteCursor is never used directly in the code, but only the Cursor interface.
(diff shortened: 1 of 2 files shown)


runningcode (approved):
I didn't know about `CursorWrapper` Nice find! Are there other place we should apply this to?

0xadam-brown:
> I didn't know about `CursorWrapper` Nice find! Are there other place we should apply this to?

This is the only spot. The other SQLite wrappers don't have a framework Wrapper equivalent, so we're out of luck as far as this trick is concerned.

The bigger false-positive volume is on instrumented OpenHelper/Database paths, which need SDKCD ignore rules rather than another wrapper change. I've created a branch for that ([link](https://github.com/getsentry/sentry/tree/fix/android-sqlite-sdkcd-false-positives)), but I'm reluctant to do anything about it because we need to rely on a heuristic that risks suppressing real crashes triggered by us (eg, a Sentry wrapper reuses a closed statement or cursor).

PR #5884, opened 2026-08-03
fix(core): Restore directory processor interrupt flag

0xadam-brown:
## :scroll: Description

Restores thread interrupt status when the inter-envelope processing delay is interrupted so callers can observe the status after directory processing returns.

Fixes [JAVA-666](https://linear.app/getsentry/issue/JAVA-666/directoryprocessor-interruptedexception-caught-but-interrupt-flag)

Via a new regression test

## :pencil: Checklist

### DIFF
--- a/sentry/src/main/java/io/sentry/DirectoryProcessor.java
+++ b/sentry/src/main/java/io/sentry/DirectoryProcessor.java
@@ -93,10 +93,14 @@ abstract class DirectoryProcessor {
         // InterruptedException will be handled by the outer try-catch
         Thread.sleep(ENVELOPE_PROCESSING_DELAY);
       }
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      logger.log(
+          SentryLevel.INFO,
+          e,
+          "Thread interrupted during processing '%s'",
+          directory.getAbsolutePath());
     } catch (Throwable e) {
-      if (e instanceof InterruptedException) {
-        Thread.currentThread().interrupt();
-      }
       logger.log(SentryLevel.ERROR, e, "Failed processing '%s'", directory.getAbsolutePath());
     }
   }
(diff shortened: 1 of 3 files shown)


sentry/src/main/java/io/sentry/DirectoryProcessor.java
```diff
@@ -94,6 +94,9 @@ public void processDirectory(final @NotNull File directory) {
         Thread.sleep(ENVELOPE_PROCESSING_DELAY);
       }
     } catch (Throwable e) {
+      if (e instanceof InterruptedException) {
+        Thread.currentThread().interrupt();
+      }
       logger.log(SentryLevel.ERROR, e, "Failed processing '%s'", directory.getAbsolutePath());
     }
```
runningcode, line 105: Not sure if it is more idomatic to catch the exception separately instead of doing an `instanceOf` check? Then we can log it at the INFO level since we expect it to happen.

```suggestion
    } catch (InterruptedException e) {
      Thread.currentThread().interrupt();
  logger.log(SentryLevel.INFO, e, "Thread interrupted during processing '%s'", directory.getAbsolutePath());
    } catch (Throwable e) {
      logger.log(SentryLevel.ERROR, e, "Failed processing '%s'",   directory.getAbsolutePath());
    }
```
```

sentry/src/main/java/io/sentry/DirectoryProcessor.java
runningcode, line 93: ah, seems like the catch `Throwable` wasn't doing this right!

runningcode (approved):
Nice find and fix! I wonder if there are other places with the same issue because we catch Throwable too broadly!

sentry/src/main/java/io/sentry/DirectoryProcessor.java
0xadam-brown: I considered catching them separately but declined because it meant duplicating the logging, which seemed like a wash at best. But I didn't think to question the existing log level (you're right - info is better, as this is a normal shutdown path). 

That^^ makes updating worthwhile 👍

PR #5886, opened 2026-08-04
fix(android): Defer frame metrics reflection during init

0xadam-brown:
## :scroll: Description

Moves Choreographer private field lookup out of the frame metrics collector constructor so that `Sentry.init()` doesn't synchronously perform framework reflection. Helps us reduce the likelihood of another common class of init ANRs (see [here](https://sentry.sentry.io/issues/6138715212/?project=4506812075540480&referrer=seer.agent.in-chat-link)).

We've seen 583k ANR events involving the SentryFrameMetricsCollector across 257 affected users since November 2024 ([link](https://sentry.sentry.io/issues/6138715212/?referrer=seer.agent.in-chat-link)), making it the highest-volume SDK-caused ANR we track. 

Those numbers got lots better thanks to @runningcode's fix in [#5641](https://github.com/getsentry/sentry-java/pull/5641): events peaked at ~12k per 3-day window in June 2026, then dropped sharply to (a fairly consistent) ~4k/3d. But the reflection this PR addresses continues to contribute to those remaining ANRs across all SDK versions, including the newest ones.

### Any behavior changes?

Behavior change from the user's perspective should be nonexistant in most cases, and minor in the worst case. 

The choreographer and choreographerLastFrameTimeField properties are still initialized by a main-thread Handler post made during collector construction, before later startCollection() calls post frame-listener registration work to the same main looper. Since those main looper tasks run in order, the Choreographer fallback should be populated before any collected frame or pending-frame interpolation normally needs it. If it's not ready yet, the failure mode is a missed/less precise first pending-frame calculation rather than a crash.

Unit tests + manual smoke tests on Android sample app.

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
@@ -56,8 +56,8 @@ public final class SentryFrameMetricsCollector implements Application.ActivityLi
   private final WindowFrameMetricsManager windowFrameMetricsManager;
 
   private @Nullable Window.OnFrameMetricsAvailableListener frameMetricsAvailableListener;
-  private @Nullable Choreographer choreographer;
-  private @Nullable Field choreographerLastFrameTimeField;
+  private volatile @Nullable Choreographer choreographer;
+  private volatile @Nullable Field choreographerLastFrameTimeField;
   private long lastFrameStartNanos = 0;
   private long lastFrameEndNanos = 0;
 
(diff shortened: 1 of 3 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
```diff
@@ -126,27 +126,24 @@ public SentryFrameMetricsCollector(
     // Most considerations regarding timestamps of frames are inspired from JankStats library:
```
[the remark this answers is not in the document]
0xadam-brown: Untouched by this PR. Leaving as-is.

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
```diff
@@ -56,8 +56,8 @@ public final class SentryFrameMetricsCollector implements Application.ActivityLi
   private final WindowFrameMetricsManager windowFrameMetricsManager;
 
   private @Nullable Window.OnFrameMetricsAvailableListener frameMetricsAvailableListener;
-  private @Nullable Choreographer choreographer;
-  private @Nullable Field choreographerLastFrameTimeField;
+  private volatile @Nullable Choreographer choreographer;
```
runningcode, line 59: If we only ever access these from the main thread why do we need to add `volatile` here?

runningcode (approved):
Looks good with one nit! Thanks for taking this further!

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java
0xadam-brown: Good question! Because after this PR we now assign those fields on the main thread but read from another thread (via `getLastKnownFrameStartTimeNanos()`, which is reached from the frame metrics callback registered with our background handler). (Before they were assigned and read from the same.)

I've updated the inline comment to be more precise about the fact that we're initializing the choreographer out of the init path, rather than using the previous "accessed on the main thread" language 👍

PR #5888, opened 2026-08-04
fix(android): Prevent duplicated breadcrumbs on tombstone-merged events

lucas-zimerman:
## :scroll: Description
Tombstone-merged native crash events already carry their own breadcrumb history captured at crash time. `ApplicationExitInfoEventProcessor.setBreadcrumbs()` was unconditionally appending disk-persisted scope breadcrumbs on top, duplicating overlapping entries (`navigation`, `device.event`, `ui.click`, etc.).

Root cause of getsentry/sentry-dart#3856 — `TombstoneMerged` events on Android showed duplicated breadcrumbs compared to the regular `signalhandler` event for the same crash.

`ApplicationExitInfoEventProcessor` backfills persisted scope data onto `Backfillable` events (ANR and Tombstone). Every other field it backfills only fills in what's missing, but `setBreadcrumbs` blindly `addAll`'d the persisted list even when the event already had its own breadcrumbs. ANR events never hit this branch (they always start breadcrumb-less), but Tombstone-merged events reuse the native SDK's own captured event, which already has breadcrumbs — so the persisted, overlapping list got appended on top.

Fix mirrors the existing guard in `SentryClient.applyScope()` for `Cached` native events: skip the backfill when the event already carries breadcrumbs.

Linear: JAVA-627, DART-429

- Added a regression test simulating a tombstone-merged event (pre-existing breadcrumbs + overlapping persisted breadcrumbs) asserting no duplication.
- Updated an existing test that had locked in the buggy merge-and-duplicate behavior.
- `./gradlew :sentry-android-core:testReleaseUnitTest --tests "*ApplicationExitInfoEventProcessorTest*"` — all 50 tests pass.
- `./gradlew spotlessApply apiDump` — clean, no public API surface changed (private method).

## :pencil: Checklist

## :crystal_ball: Next steps
Open against `getsentry/sentry-java` once ready.


lucas-zimerman:
The fix seems simple, and it kinda is, both capture event and add breadcrumbs are being asynchronously sent to the NDK layer, so at the end, they end up duplicated when checked.


Here is a sample event before the patch: https://sentry-sdks.sentry.io/issues/7652870323/events/8c52a48aa88c4060366c04557a092f4d/?project=5428562&referrer=discover-events-table
And one after: https://sentry-sdks.sentry.io/issues/7652870323/?project=5428562&query=is%3Aunresolved&referrer=issue-stream 


Also since both observers on Native are async, I thought we would lose some breadcrumb, but by doing some testing on a 10ms loop, it seems like there were no missing breadcrumbs.
Example event with the breadcrumb loop before the changes: https://sentry-sdks.sentry.io/issues/7652870323/events/068f90dcbf0c4d6e6e24f160327e4919/?project=5428562&referrer=discover-events-table

Happy to do a sanity check on merging the breadcrumbs if necessary.

sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
```diff
@@ -356,9 +356,9 @@ private void setBreadcrumbs(final @NotNull SentryBaseEvent event) {
     }
     if (event.getBreadcrumbs() == null) {
```
romtsn: `m`: wondering if we should now introduce a check for `event.getBreadcrumbs().isEmpty()` and still prefer the persisted breadcrumbs if the event's ones are empty? Or would it sacrifice correctness and we'd append the persisted breadcrumbs which are irrelevant?

romtsn: if we do that here, probably good to do the same in `SentryClient.applyScope`

sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
romtsn, line 359: `l`: theoretically we could skip reading from disk (=unnecessary I/O + they would be discarded later anyway) if we know the event already has breadcrumbs and return early?

romtsn (approved):
couple of comments, but none of them are blocking, still needs a changelog entry. 

I think we're good with this approach, merging + deduplicating would likely be fragile (due to timestamps losing precision during ser/deser, etc.), but we could keep it in mind if someone ever complains about missing crumbs

sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
lucas-zimerman: Good point, added on SentryClient here 3b6a1d2

sentry-android-core/src/main/java/io/sentry/android/core/ApplicationExitInfoEventProcessor.java
lucas-zimerman: That's a good point indeed! Avoided some I/O here 3b6a1d2

CHANGELOG.md
```diff
@@ -22,6 +26,7 @@
 
 ### Fixes
 
+- Prevent duplicated breadcrumbs on tombstone-merged native crash events ([#5888](https://github.com/getsentry/sentry-java/pull/5888))
```
lucas-zimerman: Duplicated from the merge conflict
```suggestion
```

lucas-zimerman:
@romtsn  could you do a final check here?

romtsn (approved):
looks good!

PR #5895, opened 2026-08-05
fix(core): Reset contexts when a scope is cleared

### LINKED ISSUE #5894: scope.clear() does not remove contexts
I'm adding scopes API to the Godot Engine SDK (hybrid sdk) and our tests caught an issue with the `clear` method not removing contexts from a scope object. It clears the other fields fine, just not the contexts. I'm now using a [workaround](https://github.com/getsentry/sentry-godot/blob/497f94e02023358a81391eea529193456e0b12cd/android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt#L943-L948).

limbonaut:
## :scroll: Description

`Scope.clear()` is documented as resetting the scope to its default state, but it never touched `contexts`, so anything set through `setContexts` survived a clear. Every other field the public API writes to was already being reset. This adds a `clear()` to `Contexts` and calls it from `Scope.clear()`; `CombinedContextsView` overrides it to delegate to the default scope, matching how it already handles `put` and `remove`. Other SDKs, such as JavaScript, Python, Cocoa and native all clear contexts as part of their scope clear. Surfaced while implementing per-capture scopes for the Godot SDK, where identical game code dropped the context on desktop but kept it on Android.

[Dev docs](https://develop.sentry.dev/sdk/foundations/state-management/scopes/#scope-data-methods-1) describe `clear` as

> Reset scope to defaults, keeping event processors.

So this change brings implementation closer to spec.

* Resolves https://github.com/getsentry/sentry-java/issues/5894

## :pencil: Checklist


sentry/src/main/java/io/sentry/Scope.java
```diff
@@ -575,6 +575,7 @@ public void clear() {
     tags.clear();
     attributes.clear();
     extra.clear();
+    contexts.clear();
     eventProcessors.clear();
```
limbonaut, line 579: Also worth mentioning another thing from [Scopes](https://develop.sentry.dev/sdk/foundations/state-management/scopes/) spec: "keeping event processors". Currently, they are cleared. I'm not sure if it's an intentional deviation.

markushi:
Dang, not sure what went wrong here, but this is a dup of (the already) merged https://github.com/getsentry/sentry-java/pull/5902

On the upside the diff within the implementation itself is empty 😅

limbonaut:
Gotcha :laughing:  Closing as superseded by #5902

PR #5902, opened 2026-08-06
fix(core): Clear contexts on Scope.clear()

### LINKED ISSUE #5894: scope.clear() does not remove contexts

runningcode:
## :scroll: Description

`Scope.clear()` reset every field on the scope except `contexts`, so anything set via `setContexts(...)` survived a `clear()`. The root cause is that `Contexts` had no `clear()` method (unlike the `tags`, `extra`, and `fingerprint` collections that `clear()` already resets), so contexts were simply left untouched.

This PR:
- Adds `Contexts.clear()`.
- Overrides `clear()` in `CombinedContextsView` to delegate to the default context, consistent with how `put`/`remove` already delegate (otherwise `clear()` on the view would no-op against its always-empty inherited storage).
- Clears contexts in `Scope.clear()` and notifies scope observers with the now-empty contexts, so persisted scopes (e.g. Android's scope-on-disk used for native crashes) stay in sync — matching what `setContexts`/`clearBreadcrumbs` already do.

Reported while adding the scopes API to the Godot Engine (hybrid) SDK: `clear()` cleared everything except contexts, forcing a [manual workaround](https://github.com/getsentry/sentry-godot/blob/497f94e02023358a81391eea529193456e0b12cd/android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt#L943-L948) that removes each context key by hand.

Fixes GH-5894
Fixes JAVA-683

Added unit tests:
- `ContextsTest` — `clear()` empties the contexts.
- `CombinedContextsViewTest` — `clear()` clears only the default context.
- `ScopeTest` — extended the reset test to cover contexts, plus a new test verifying `clear()` notifies scope observers with empty contexts.

## :pencil: Checklist

## :crystal_ball: Next steps

Hybrid SDKs (e.g. Godot) can drop their manual context-clearing workaround once this ships.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixes
 
+- Clear contexts when calling `Scope.clear()` ([#5902](https://github.com/getsentry/sentry-java/pull/5902))
 - Preserve custom `Throwable` identities when R8 optimizes Android apps ([#5881](https://github.com/getsentry/sentry-java/pull/5881))
 
 ## 8.52.0
--- a/sentry/api/sentry.api
+++ b/sentry/api/sentry.api
@@ -227,6 +227,7 @@ public final class io/sentry/CheckInStatus : java/lang/Enum {
 
 public final class io/sentry/CombinedContextsView : io/sentry/protocol/Contexts {
 	public fun <init> (Lio/sentry/protocol/Contexts;Lio/sentry/protocol/Contexts;Lio/sentry/protocol/Contexts;Lio/sentry/ScopeType;)V
+	public fun clear ()V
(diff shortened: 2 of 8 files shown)


markushi (approved):
LGTM!

PR #5907, opened 2026-08-07
feat(core): Add ExceptionUtils.handleFatal to rethrow non-recoverable throwables

### LINKED ISSUE #5865: Create a rethrow function to properly handle non-recoverable Throwables
Related to getsentry/sentry-java#5513, we want to create a function/lambda so that we can catch `Throwable` and safely rethrow errors that we shouldn't catch.  <br>  <br>The types of exceptions and errors that should be rethrown should be OOMs, ProcessDeath, Stackoverflow, InterruptedException and other types of errors (research this).  <br>  <br>Example

```kotlin
fun rethrowFatal(t: Throwable) {
  if (t is X) {
     throw t
  } else if (t is Y) {
     throw t
  }
  // just ignore or log?
}
```

so it can be used like:

```kotlin
try {
  doDangerousThing()
} catch(t: Throwable) {
  rethrowFatal(t)
}
```

  <br>**Ideally we shouldn't use this function often, just when we really don't want to crash.**

markushi:
## Summary

Adds `ExceptionUtils.handleFatal(Throwable)`, a small utility for use inside broad `catch (Throwable t)` blocks. It rethrows `VirtualMachineError` (e.g. `OutOfMemoryError`, `StackOverflowError`) and `ThreadDeath` as-is, and restores the thread's interrupt flag for `InterruptedException` instead of swallowing it. All other throwables are left untouched for the caller to handle as before.

Closes getsentry/sentry-java#5865

#skip-changelog

### DIFF
--- a/sentry/src/main/java/io/sentry/util/ExceptionUtils.java
+++ b/sentry/src/main/java/io/sentry/util/ExceptionUtils.java
@@ -29,4 +29,24 @@ public final class ExceptionUtils {
       final @NotNull Throwable throwable) {
     return ignoredExceptionsForType.contains(throwable.getClass());
   }
+
+  /**
+   * Handles non-recoverable {@link Throwable}s that should never be swallowed. Rethrows {@link
+   * VirtualMachineError} (e.g. OutOfMemoryError/StackOverflowError) and {@link ThreadDeath} as-is.
+   * For {@link InterruptedException}, restores the thread's interrupted status instead of
+   * rethrowing, since it is a checked exception. All other throwables are left untouched for the
+   * caller to handle/log/ignore as before.
+   *
+   * @param throwable - the throwable to check
+   */
+  public static void handleFatal(final @NotNull Throwable throwable) {
+    // VirtualMachineError covers OutOfMemoryError, StackOverflowError, InternalError, and
+    // UnknownError
+    if (throwable instanceof VirtualMachineError || throwable instanceof ThreadDeath) {
+      throw (Error) throwable;
+    }
+    if (throwable instanceof InterruptedException) {
+      Thread.currentThread().interrupt();
+    }
+  }
 }
--- a/sentry/src/main/java/io/sentry/util/ExceptionUtils.java
+++ b/sentry/src/main/java/io/sentry/util/ExceptionUtils.java
@@ -32,17 +32,27 @@ public final class ExceptionUtils {
 
   /**
    * Rethrows non-recoverable {@link Throwable}s that should never be swallowed: {@link
-   * VirtualMachineError} (e.g. OutOfMemoryError/StackOverflowError) and {@link ThreadDeath} are
-   * rethrown as-is. For {@link InterruptedException}, the thread's interrupted status is restored
-   * instead of rethrowing, since it is a checked exception. All other throwables are left untouched
-   * for the caller to handle/log/ignore as before.
+   * VirtualMachineError} (e.g. OutOfMemoryError/StackOverflowError), {@link ThreadDeath} and {@link
+   * LinkageError} are rethrown as-is. For {@link InterruptedException}, the thread's interrupted
+   * status is restored instead of rethrowing, since it is a checked exception. All other throwables
+   * are left untouched for the caller to handle/log/ignore as before.
+   *
+   * <p>Note on {@link LinkageError}: its subclasses (e.g. {@link NoClassDefFoundError}, {@link
+   * NoSuchMethodError}, {@link AbstractMethodError}, {@link UnsatisfiedLinkError}) are also the
+   * expected runtime signal for a missing or version-mismatched optional dependency, which is a
+   * normal condition for integrations built against {@code compileOnly} dependencies. Call sites
+   * that probe for such a dependency must catch the relevant {@link LinkageError} subclass locally
+   * <em>before</em> delegating to this method — see {@code SentrySQLiteDriver.hasConnectionPool}
+   * and {@link LoadClass} for examples.
    *
    * @param throwable - the throwable to check
    */
   public static void rethrowIfFatal(final @NotNull Throwable throwable) {
     // VirtualMachineError covers OutOfMemoryError, StackOverflowError, InternalError, and
     // UnknownError
-    if (throwable instanceof VirtualMachineError || throwable instanceof ThreadDeath) {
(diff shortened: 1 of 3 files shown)


sentry/src/main/java/io/sentry/util/ExceptionUtils.java
```diff
@@ -29,4 +29,24 @@ public static boolean isIgnored(
...
+   *
+   * @param throwable - the throwable to check
+   */
+  public static void handleFatal(final @NotNull Throwable throwable) {
+    // VirtualMachineError covers OutOfMemoryError, StackOverflowError, InternalError, and
+    // UnknownError
+    if (throwable instanceof VirtualMachineError || throwable instanceof ThreadDeath) {
+      throw (Error) throwable;
+    }
+    if (throwable instanceof InterruptedException) {
```
markushi, line 58: I was a bit torn apart if this utility method should take care of interrupts as well, and decided to do it, as it will keep all call sites more tight. Along with that change the method is now a generic `handleFatal` instead of a `rethrowFatal`. That doesn't tell a lot about the method, but I guess it will be a common pattern everywhere. - Happy for any opinions on this!

sentry/src/main/java/io/sentry/util/ExceptionUtils.java
```diff
@@ -29,4 +29,24 @@ public static boolean isIgnored(
...
+  /**
+   * Handles non-recoverable {@link Throwable}s that should never be swallowed. Rethrows {@link
+   * VirtualMachineError} (e.g. OutOfMemoryError/StackOverflowError) and {@link ThreadDeath} as-is.
+   * For {@link InterruptedException}, restores the thread's interrupted status instead of
+   * rethrowing, since it is a checked exception. All other throwables are left untouched for the
+   * caller to handle/log/ignore as before.
+   *
+   * @param throwable - the throwable to check
+   */
+  public static void handleFatal(final @NotNull Throwable throwable) {
```
0xadam-brown: l: Thoughts about `rethrowIfFatal()`?

To me, that'd make its purpose more obvious at the call site (and I'm fine with its cheating a bit w/r/t restoring the interrupted flag).

sentry/src/main/java/io/sentry/util/ExceptionUtils.java
0xadam-brown: Seems reasonable to me, and definitely nice to have a common facility for restoring the interrupt flag.  

The only potential concern I can think of are methods that declare `throws InterruptedException` and need to use `handleFatal()`, as sometimes clients might expect either an InterruptedException or an interrupted flag to be set, but not both. But those methods could simply re-throw the `InterruptedException` before calling `handleFatal()`.

The alternative would be to have separate methods similar to what Project Reactor does: 

- ExceptionUtils.rethrowIfJvmFatal()
- ExceptionUtils.restoreInterruptIfNeeded()
- ExceptionUtils.rethrowIfCancellation() (for coroutine CancellationException)

But I prefer the current all-in-one approach more 👍

sentry/src/main/java/io/sentry/util/ExceptionUtils.java
```diff
@@ -29,4 +29,24 @@ public static boolean isIgnored(
...
+    // VirtualMachineError covers OutOfMemoryError, StackOverflowError, InternalError, and
+    // UnknownError
+    if (throwable instanceof VirtualMachineError || throwable instanceof ThreadDeath) {
```
0xadam-brown: (Agree that we should omit re-throwing LinkageError b/c of the `compileOnly` issue illustrated [here](https://github.com/getsentry/sentry-java/blob/1664d46c82b86ce7e7e0e264d2d1f878d9da8d26/sentry-android-sqlite/src/main/java/io/sentry/sqlite/SentrySQLiteDriver.kt#L44).)

0xadam-brown (approved):
Nice – and thanks for this! 

One renaming comment for your consideration, but no blockers 🥇

sentry/src/main/java/io/sentry/util/ExceptionUtils.java
markushi: Yeah sounds good to me too, let's cheat a little 😅

sentry/src/main/java/io/sentry/util/ExceptionUtils.java
markushi: That's some good input! We could improve the ergonomics slightly by introducing method chaining. E.g.

```
ExceptionUtils.restoreInterrupt().rethrowIfJvmFatal()
```

Let me task the clanker to have a look at this and the actual impact on our codebase

sentry/src/main/java/io/sentry/util/ExceptionUtils.java
runningcode: That's a valid point but I think we should re-throw `LinkageError` and only catch it where we expect it to be thrown as you did there.

sentry/src/main/java/io/sentry/util/ExceptionUtils.java
runningcode: Agreed with `rethrowIfFatal`!

sentry/src/main/java/io/sentry/util/ExceptionUtils.java
runningcode: I think it depends on the codebase and if we have use cases where we would need to do one and not the other. Can you ask the clanker to give you a report if we have such cases?

runningcode (commented):
Thanks for doing this! It looks good so far.

sentry/src/main/java/io/sentry/util/ExceptionUtils.java
markushi: Alright, I checked - we do not seem to have any call sites where we want to distinguish them. Which kinda makes sense as the interrupt flag should be always be set if necessary.

There are a few cases were we just catch an `InterruptedException` right now, e.g.
https://github.com/getsentry/sentry-java/blob/9054d65b80c7a724faf4fefeaf104bba1f6e3f92/sentry-apache-http-client-5/src/main/java/io/sentry/transport/apache/ApacheHttpClientTransport.java#L193-L203

but I'd leave those call sites as-is, they don't need any broad exception catching.

PR #5909, opened 2026-08-07
ci(macrobenchmark): Run the startup benchmark on Sauce Labs and recover its results

runningcode:
## :scroll: Description

Runs the `sentry-uitest-android-macrobenchmark` cold-start benchmark on a Sauce Labs real device and gets the numbers back into CI. See the results [here](https://github.com/getsentry/sentry-java/actions/runs/31159874391).

In a future PR, we will either decide to post this as PR comments OR run this as a pre-release task against the last relase. But for now I'm just getting the mechanism working.

Q: **Why is this so complicated?**
A: Sauce does not provide any APIs to pull files off the devices unlike firebase test lab or other tools. [Here's a link to Sauce docs in case you find something I didn't.](https://docs.saucelabs.com/mobile-apps/automated-testing/espresso-xcuitest/espresso/) So to work around this we echo the `benchmarkData.json` into logcat in chunks and sauce does give us back a `device.log`. `scripts/parse-macrobenchmark-log.py` uses that to reassemble the `benchmarkData.json` and parse it.

The things my clanker tried are documented in the README of the module so others can check/verify.

Note: sauce labs devices run with unlocked CPU clocks so the benchmarks will be unstable!!

Get a reliable Sdk init benchmark!

Green run on a Pixel 9 Pro XL (api 35) — [31159874391](https://github.com/getsentry/sentry-java/actions/runs/31159874391) — recovered all three metrics over 12 iterations:

| Metric | min | median | max | CoV |
|---|--:|--:|--:|--:|
| SentryAndroid.initSumMs | 15.7 | 21.3 | 24.3 | 13.4% |
| SentryAndroid.initCount | 1.0 | 1.0 | 1.0 | 0.0% |
| timeToInitialDisplayMs | 443.4 | 538.6 | 612.3 | 11.9% |

## :pencil: Checklist

## :crystal_ball: Next steps

Before this leaves draft:

- **Remove the temporary `push:` trigger** on this branch from the workflow — it only exists because `workflow_dispatch` cannot target a workflow absent from the default branch. It is marked `Remove before merge` in the file.
- **Confirm one green run on the current head**, since the device was reverted after the Pixel 4 attempts.
- Consider `if-no-files-found: error` on the upload step. A failed Sauce run currently uploads nothing while the step reports success.

Follow-ups, out of scope here:

#skip-changelog


.github/workflows/integration-tests-macrobenchmark.yml
```diff
@@ -0,0 +1,71 @@
+name: 'Integration Tests - Macrobenchmark'
+# Runs the sentry-uitest-android-macrobenchmark cold-start benchmark on a Sauce Labs real
+# device and recovers timeToInitialDisplay from the device log.
+#
+on:
+  workflow_dispatch:
+  # Temporary scaffolding: workflow_dispatch cannot target a workflow that does not exist on
+  # the default branch yet, so trigger on pushes to this branch while we validate the flow.
+  # Remove before merge.
+  push:
```
runningcode: We want to remove this before merging and just use workflow_dispatch until we decide how we want to use this.

.github/workflows/integration-tests-macrobenchmark.yml
```diff
@@ -0,0 +1,71 @@
...
+        run: |
+          # Without pipefail the step passes on `tee`'s exit code, so a failed recovery
+          # would report success while silently producing no results.
+          set -o pipefail
+          python3 scripts/parse-macrobenchmark-log.py ./artifacts \
+            --json-out ./artifacts/benchmarkData.json | tee -a "$GITHUB_STEP_SUMMARY"
+
+      - name: Upload Sauce artifacts
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+        if: always()
```
romtsn: `l`: should we gate this (and the one above) with `!cancelled()`?

romtsn (approved):
nice!

.github/workflows/integration-tests-macrobenchmark.yml
runningcode: good idea. done!

PR #5916, opened 2026-08-10
feat(core): [Unhandled Sessions 1] Add Unhandled session state and pending marker

buenaflor:
## PR Stack (Unhandled Sessions)

- #5915 (collection)
- #5916
- #5917
- #5918

---

## :scroll: Description

Adds the `unhandled` session status to the session model.

- `Session.State.Unhandled` — the protocol status for an unhandled error that did **not** terminate the process.
- A `pendingUnhandled` marker on `Session` (`pending_unhandled` when serialized) that records "this session saw an unhandled error" without ending the session or changing its status while it is still alive.
- `end()` finalizes a session carrying the marker as `Unhandled` instead of `Exited`. `Crashed` and `Abnormal` keep taking precedence, and `update(Crashed, ...)` clears the marker so a real crash always wins.
- `clone()` and the (de)serializer carry the marker; it is omitted from JSON when unset.

Nothing sets the marker yet — this PR is inert on its own. It is the model layer for the rest of the stack.

## :bulb: Motivation and Context

Hybrid runtimes such as Flutter report `handled=false` exceptions that do not kill the process. Today those go through the terminating capture path, which marks the session `crashed` and starts a replacement session even though the app keeps running, incorrectly lowering crash-free session rates.

The session protocol has had a dedicated status for exactly this case [since 1.6.0](https://develop.sentry.dev/sdk/telemetry/sessions/): `unhandled` — "an unhandled error occurred but the process did not terminate". Relay accepts it (`SessionStatus::Unhandled` → `"unhandled"`).

## :green_heart: How did you test it?

New `SessionTest` covering the marker's atomic update, terminal-state precedence (`Crashed`/`Abnormal` win), `clone()`, and serialization round-trips. `PreviousSessionFinalizerTest` covers recovery of a previous session carrying the marker, including a native crash escalating it to `crashed`. `SessionSerializationTest` covers the JSON round-trip.

## :pencil: Checklist

- [ ] I added GH Issue ID _&_ Linear ID
- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [x] I updated the docs if needed.
- [x] I updated the wizard if needed.
- [x] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.

## :crystal_ball: Next steps

Persisting the marker across process death and exposing the capture API, in the following PRs in this stack.

#skip-changelog

[...]

PR #5917, opened 2026-08-10
fix(core): [Unhandled Sessions 2] Make session cache writes session-id aware

buenaflor:

- #5915 (collection)
- #5916
- #5917
- #5918

---

## :scroll: Description

Makes the session file cache aware of *which* session it is acting on.

Two paths in `EnvelopeCache.storeInternal` previously ignored session identity:

- **`SessionEnd`** deleted `session.json` unconditionally. A delayed `End(A)` arriving after a newer session B was persisted would delete B's file.
- **`SessionStart`** always rotated `session.json` into `previous_session.json` and overwrote it. A delayed `Start(A)` arriving after A's own state had already advanced on disk would overwrite the newer snapshot.

Both now compare session ids (and, for `SessionEnd`, start times) before deleting or rotating. A stale end only preserves the current file when the current session is genuinely newer — a different id **and** a later start time. Everything else keeps the existing behaviour, including malformed envelopes and unreadable session files, which still delete as before.

Also adds `EnvelopeCache.persistCurrentSession(Session)` so a caller can flush the active session to disk outside the envelope-store path, and extracts `readSessionFromEnvelope` / `readSessionFromDisk` helpers.

Needed by the next PR in the stack, which mutates the live session in place (marking it pending-unhandled) and persists it so the marker survives process death. Without session-id awareness, an unrelated in-flight `SessionEnd` or `SessionStart` could clobber that snapshot.

In practice this only changes observable behaviour for sessions carrying the pending-unhandled marker from PR 1 — the preserve branch requires `isPendingUnhandled()` — so existing SDK flows are unaffected.

`EnvelopeCacheTest` grew coverage for delayed same-id `SessionStart` preserving a newer pending or higher-error snapshot, different-id starts still rotating, matching and mismatching `SessionEnd`, null session ids on both paths, malformed envelopes, and unreadable session files. Also verified the pre-existing `AbnormalExit` and `NativeCrashExit` recovery paths still win over a pending marker. 36 tests, all passing.

## :pencil: Checklist

## :crystal_ball: Next steps

The capture API that sets the marker, in the next PR in this stack.

#skip-changelog

### DIFF
--- a/sentry/api/sentry.api
+++ b/sentry/api/sentry.api
@@ -4862,6 +4862,7 @@ public class io/sentry/cache/EnvelopeCache : io/sentry/cache/IEnvelopeCache {
 	public static fun getPreviousSessionFile (Ljava/lang/String;)Ljava/io/File;
 	public fun iterator ()Ljava/util/Iterator;
 	public fun movePreviousSession (Ljava/io/File;Ljava/io/File;)V
+	public fun persistCurrentSession (Lio/sentry/Session;)V
 	public fun store (Lio/sentry/SentryEnvelope;Lio/sentry/Hint;)V
 	public fun storeEnvelope (Lio/sentry/SentryEnvelope;Lio/sentry/Hint;)Z
 	public fun waitPreviousSessionFlush ()Z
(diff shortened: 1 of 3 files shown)


buenaflor:
Not actually merged into `main`. This PR was auto-closed when the collection branch `feat/unhandled-sessions` was accidentally fast-forwarded to the tip of the stack, which made GitHub consider every stack branch merged into its base.

The collection branch has been restored to its empty commit and the stack rebuilt unchanged. This PR continues as #5920.

PR #5918, opened 2026-08-10
feat(android): [Unhandled Sessions 3] Add internal non-terminating envelope capture

buenaflor:

- #5915 (collection)
- #5916
- #5917
- #5918

---

## :scroll: Description

Adds `InternalSentrySdk.captureEnvelopeNonTerminating(byte[])` for hybrid runtimes where an unhandled exception does **not** terminate the process.

Compared to `captureEnvelope(byte[], boolean)`, the new path does not treat `handled=false` as a crash. Instead it:

- marks the current session pending-unhandled and increments its error count,
- keeps the session `Ok` with the same session id on the scope,
- attaches no session item to the envelope and starts no new session,
- persists the session so the marker survives process death.

The session is finalized later by normal lifecycle (`endSession`, background, or previous-session recovery) as `unhandled`, unless a native crash escalates it to `crashed`. `Abnormal` and `Crashed` remain authoritative, and `captureEnvelope(byte[], boolean)` is unchanged for the SDKs already using it.

Also in this PR:

- `Scope.IWithSession` is widened to public so `InternalSentrySdk` can mutate the session under the scope lock. This is its only consumer, hence it lives here rather than in PR 1.
- A shared private `readEnvelope` helper for both capture paths, and `catch (Throwable)` narrowed to `catch (Exception)` so `OutOfMemoryError` and friends propagate instead of being swallowed. Note `IEnvelopeReader.read` declares only `IOException` but also throws unchecked `IllegalArgumentException` for malformed payloads, which is why the helper still catches `Exception`.

Flutter currently forwards `handled=false` events through the terminating hybrid capture path. That marks the session `crashed` and may start a replacement session even though the Flutter process keeps running, incorrectly lowering crash-free session rates.

This follows the session protocol's [`unhandled` status](https://develop.sentry.dev/sdk/telemetry/sessions/) for exceptions that are unhandled by the application but do not terminate the process.

New `InternalSentrySdkTest` coverage for: the session staying `Ok` with the same id and the marker persisted to disk; `endSession` afterwards finalizing as `unhandled`; and a subsequent hard crash through the existing terminating API finalizing the old session as `crashed` (clearing the marker) and starting a fresh `Ok` session. Existing `captureEnvelope` tests confirm the terminating path is unchanged, including malformed-payload handling.

## :pencil: Checklist

## :crystal_ball: Next steps

Update the Flutter Android bridge to use `captureEnvelopeNonTerminating` for non-terminating unhandled events.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@
   - [changelog](https://github.com/getsentry/sentry-native/blob/master/CHANGELOG.md#0162)
   - [diff](https://github.com/getsentry/sentry-native/compare/0.16.1...0.16.2)
 
+### Internal
+
+- Add `InternalSentrySdk.captureEnvelopeNonTerminating` for hybrid SDKs (e.g. Flutter) so unhandled exceptions that don't terminate the process no longer end the session as `crashed` ([#5918](https://github.com/getsentry/sentry-java/pull/5918))
+
 ## 8.52.0
 
 ### Fixes
--- a/sentry-android-core/api/sentry-android-core.api
+++ b/sentry-android-core/api/sentry-android-core.api
@@ -315,6 +315,7 @@ public abstract interface class io/sentry/android/core/IDebugImagesLoader {
 public final class io/sentry/android/core/InternalSentrySdk {
 	public fun <init> ()V
 	public static fun captureEnvelope ([BZ)Lio/sentry/protocol/SentryId;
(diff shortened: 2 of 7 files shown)


buenaflor:
Not actually merged into `main`. This PR was auto-closed when the collection branch `feat/unhandled-sessions` was accidentally fast-forwarded to the tip of the stack, which made GitHub consider every stack branch merged into its base.

The collection branch has been restored to its empty commit and the stack rebuilt unchanged. This PR continues as #5921.

PR #5919, opened 2026-08-10
feat(core): [Unhandled Sessions 1] Add Unhandled session state and non-terminating error flag

buenaflor:

- #5919
- #5920
- #5921
- #5990

---

## :scroll: Description

Adds the session protocol's `unhandled` status to the `Session` model.

- `Session.State.Unhandled` — an unhandled error the process survived.
- A `hasNonTerminatingUnhandledError` flag (`non_terminating_unhandled_error` in JSON), set by `recordNonTerminatingUnhandledError()`, which counts the error without ending the session or changing its status while it is alive.
- `end()` finalizes a flagged session as `Unhandled` rather than `Exited`. `Crashed` and `Abnormal` still take precedence, and `update(Crashed, ...)` clears the flag.
- `clone()` and the (de)serializer carry the flag; it is omitted from JSON when unset.

Nothing sets the flag yet, so this PR is inert on its own.

The flag is named for what separates this case from a native crash, which is also unhandled but terminates the process and ends the session as `crashed`.

Hybrid runtimes such as Flutter report `handled=false` exceptions that do not kill the process. Those currently go through the terminating capture path, which marks the session `crashed` and starts a replacement even though the app keeps running, incorrectly lowering crash-free session rates.

The protocol has had [`unhandled`](https://develop.sentry.dev/sdk/telemetry/sessions/) for exactly this case since 1.6.0, and Relay accepts it.

`SessionSerializationTest` gets a `session_unhandled.json` fixture covering the full unhandled document, round-tripped through the same serialize/deserialize pair as `session.json`. Omission when the flag is unset needs no test of its own: the existing `session.json` has no `non_terminating_unhandled_error` key and those tests compare exact strings, so a leaked key fails them.

`PreviousSessionFinalizerTest` covers recovering a flagged previous session, including a native crash escalating it to `crashed`. `EnvelopeCacheTest` covers the same precedence through the `AbnormalExit` and `NativeCrashExit` recovery paths.

## :pencil: Checklist

## :crystal_ball: Next steps

Persistence across process death in #5920, the capture API in #5921.


sentry/src/test/java/io/sentry/protocol/SessionSerializationTest.kt
```diff
@@ -53,6 +54,26 @@ class SessionSerializationTest {
     assertEquals(expectedJson, actualJson)
   }
 
+  @Test
+  fun `serialize and deserialize round-trips Unhandled status and non-terminating flag`() {
```
runningcode: i wonder if we can just have a single json / json test with all the different valid options instead of one test per json option

runningcode (approved):
Nice! Thanks for breaking this up

sentry/src/main/java/io/sentry/Session.java
```diff
@@ -188,6 +192,42 @@ public int errorCount() {
...
+  /**
+   * Whether the session experienced an unhandled error that did <em>not</em> terminate the process,
+   * e.g. an unhandled Flutter exception, and so finalizes as {@link State#Unhandled} rather than
+   * {@link State#Exited}. A native crash is also unhandled, but it kills the process and ends the
+   * session as {@link State#Crashed} instead.
+   *
+   * <p>Never sent as a status while the session is alive; it is only persisted with the session.
+   */
+  @ApiStatus.Internal
+  public boolean hasNonTerminatingUnhandledError() {
```
runningcode, line 209: do we need to grab the `sessionLock` in order to read this?

otherwise if this could be set to true before `errorCount` is incremented. would that be a valid state?

sentry/src/main/java/io/sentry/Session.java
```diff
@@ -21,7 +21,8 @@ public enum State {
     Ok,
     Exited,
     Crashed,
-    Abnormal
+    Abnormal,
+    Unhandled
```
runningcode, line 30: should we javadoc this? i know the other ones don't but this feels like it is worht documenting given the surprising behavior

sentry/src/main/java/io/sentry/Session.java
```diff
@@ -188,6 +192,42 @@ public int errorCount() {
...
+   *
+   * <p>Never sent as a status while the session is alive; it is only persisted with the session.
+   */
+  @ApiStatus.Internal
+  public boolean hasNonTerminatingUnhandledError() {
+    return hasNonTerminatingUnhandledError;
+  }
+
+  /**
+   * Records that an active session experienced an unhandled error which did not terminate the
```
runningcode, line 214: should we document that this is only meant to be called from hybrid sdks and never from java/kotlin/android?

sentry/src/main/java/io/sentry/Session.java
buenaflor: none of the other getters take the lock either, only the write paths do

flag=true with a stale error count isn't a real state we care about, both get set in the same locked block in record…(), and end() reads the flag under the lock too

sentry/src/main/java/io/sentry/Session.java
```diff
@@ -66,6 +72,9 @@ public enum State {
   /** the Abnormal mechanism, e.g. what was the reason for session to become abnormal (ANR) */
   private @Nullable String abnormalMechanism;
 
+  /** Whether an unhandled error occurred that did not terminate the process */
+  private boolean hasNonTerminatingUnhandledError;
```
markushi, line 76: this should be volatile, but it's not done anywhere else in this class, so good to keep it as-is.

sentry/src/main/java/io/sentry/Session.java
```diff
@@ -188,6 +197,45 @@ public int errorCount() {
...
+  /**
+   * Records that an active session experienced an unhandled error which did not terminate the
+   * process, counting the error and advancing the session's sequence without ending it. On {@link
+   * #end()} the session is finalized as {@link State#Unhandled} unless a terminal status such as
+   * {@link State#Crashed} or {@link State#Abnormal} took over first.
+   *
+   * <p>Hybrid SDKs whose unhandled errors do not kill the process. Native Java/Android capture
+   * should not call this.
+   *
+   * @return whether the session was updated, i.e. false if it had already reached a terminal state
```
markushi, line 222: nice!

markushi (approved):

PR #5921, opened 2026-08-10
feat(android): [Unhandled Sessions 3] Add internal non-terminating envelope capture

buenaflor:

- #5915 (collection)
- #5919
- #5920
- #5921
- #5990

---

## :scroll: Description

Adds `InternalSentrySdk.captureEnvelopeNonTerminating(byte[])` for hybrid runtimes where an unhandled exception does **not** terminate the process.

Unlike `captureEnvelope(byte[], boolean)`, it does not treat `handled=false` as a crash. Instead it:

- flags the current session and increments its error count,
- keeps the session `Ok` with the same session id on the scope,
- attaches no session item and starts no new session,
- persists the session so the flag survives process death.

The session is finalized later by normal lifecycle (`endSession`, background, or previous-session recovery) as `unhandled`, unless a terminal status takes over first, such as `crashed` for a native crash or `abnormal` for an ANR. `captureEnvelope(byte[], boolean)` is unchanged.

Also in this PR:

- `Scope.IWithSession` becomes public so `InternalSentrySdk` can mutate the session under the scope lock; it is the only consumer. It carries `@ApiStatus.Internal` like `IWithTransaction` and `IWithPropagationContext`, so it is not supported API despite being public.
- A shared `eventStateOf` returning `EnvelopeEventState`, replacing two loops that computed the same booleans by different routes, plus a shared `readEnvelope`.
- The code added here catches `Exception` rather than `Throwable`, so `OutOfMemoryError` and friends propagate instead of being swallowed. The existing `captureEnvelope` keeps its `catch (Throwable)`.

Flutter forwards `handled=false` events through the terminating hybrid capture path. That marks the session `crashed` and may start a replacement session even though the Flutter process keeps running, incorrectly lowering crash-free session rates.

New `InternalSentrySdkTest` coverage: the session staying `Ok` with the same id and the flag persisted to disk; `endSession` afterwards finalizing as `unhandled`; and a later hard crash finalizing the old session as `crashed` and starting a fresh `Ok` session. Existing `captureEnvelope` tests confirm that path is unchanged.

## :pencil: Checklist

## :crystal_ball: Next steps

Update the Flutter Android bridge to use `captureEnvelopeNonTerminating` for non-terminating unhandled events.

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
@@ -203,8 +203,8 @@ public final class InternalSentrySdk {
       final SentryEnvelope repackagedEnvelope =
           new SentryEnvelope(envelope.getHeader(), envelopeItems);
       return scopes.captureEnvelope(repackagedEnvelope);
-    } catch (Exception e) {
-      options.getLogger().log(SentryLevel.ERROR, "Failed to capture envelope", e);
+    } catch (Throwable t) {
+      options.getLogger().log(SentryLevel.ERROR, "Failed to capture envelope", t);
     }
     return null;
   }
--- a/sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
@@ -41,7 +41,6 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
-import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.atomic.AtomicReference;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
@@ -175,39 +174,45 @@ public final class InternalSentrySdk {
       return null;
     }
 
+    final @NotNull ISerializer serializer = options.getSerializer();
+    final @NotNull EnvelopeEventState eventState;
     try {
-      final @NotNull ISerializer serializer = options.getSerializer();
-      final @NotNull EnvelopeEventState eventState = eventStateOf(envelope, serializer);
-
-      final @NotNull List<SentryEnvelopeItem> envelopeItems = new ArrayList<>();
-      for (SentryEnvelopeItem item : envelope.getItems()) {
-        envelopeItems.add(item);
-      }
-
-      // update session and add it to envelope if necessary
-      final @Nullable Session.State status =
-          eventState == EnvelopeEventState.UNHANDLED ? Session.State.Crashed : null;
-      final @Nullable Session session =
-          updateSession(scopes, options, status, eventState != EnvelopeEventState.NONE);
-      if (session != null) {
-        final SentryEnvelopeItem sessionItem = SentryEnvelopeItem.fromSession(serializer, session);
-        envelopeItems.add(sessionItem);
-        deleteCurrentSessionFile(
-            options,
-            // should be sync if going to crash or already not a main thread
-            !maybeStartNewSession || !scopes.getOptions().getThreadChecker().isMainThread());
-        if (maybeStartNewSession) {
-          scopes.startSession();
-        }
-      }
-
-      final SentryEnvelope repackagedEnvelope =
-          new SentryEnvelope(envelope.getHeader(), envelopeItems);
-      return scopes.captureEnvelope(repackagedEnvelope);
-    } catch (Throwable t) {
-      options.getLogger().log(SentryLevel.ERROR, "Failed to capture envelope", t);
+      eventState = eventStateOf(envelope, serializer);
+    } catch (Exception e) {
+      // getEvent reads through a Callable, whose call() declares Exception
+      options.getLogger().log(SentryLevel.ERROR, "Failed to inspect envelope events", e);
+      return null;
     }
-    return null;
+
+    final @NotNull List<SentryEnvelopeItem> envelopeItems = new ArrayList<>();
+    for (SentryEnvelopeItem item : envelope.getItems()) {
+      envelopeItems.add(item);
+    }
+
+    // update session and add it to envelope if necessary
+    final @Nullable Session.State status =
+        eventState == EnvelopeEventState.UNHANDLED ? Session.State.Crashed : null;
+    final @Nullable Session session =
+        updateSession(scopes, options, status, eventState != EnvelopeEventState.NONE);
+    if (session != null) {
+      try {
+        envelopeItems.add(SentryEnvelopeItem.fromSession(serializer, session));
+      } catch (IOException e) {
+        options.getLogger().log(SentryLevel.ERROR, "Failed to add session to envelope", e);
+        return null;
+      }
+      deleteCurrentSessionFile(
+          options,
+          // should be sync if going to crash or already not a main thread
+          !maybeStartNewSession || !scopes.getOptions().getThreadChecker().isMainThread());
+      if (maybeStartNewSession) {
+        scopes.startSession();
+      }
+    }
+
+    final SentryEnvelope repackagedEnvelope =
+        new SentryEnvelope(envelope.getHeader(), envelopeItems);
+    return scopes.captureEnvelope(repackagedEnvelope);
   }
 
   /**
@@ -262,7 +267,11 @@ public final class InternalSentrySdk {
   }
 
   /**
-   * Flags and persists the current session for a non-terminating hybrid error.
+   * Flags the current session for a non-terminating hybrid error and persists it before returning,
+   * so the marker survives an immediate process death.
+   *
+   * <p>The write stays inside {@code withSession}: the scope's session lock is what keeps the
+   * session from being ended or replaced mid-write.
    *
    * @param unhandled {@code true} if the error was unhandled ({@code mechanism.handled=false})
    */
@@ -287,30 +296,10 @@ public final class InternalSentrySdk {
                       unhandled
                           ? session.recordNonTerminatingUnhandledError()
                           : session.update(null, null, true, null);
-                  if (recorded) {
-                    schedulePersistSession(options, session.clone());
+                  if (recorded && options.getEnvelopeDiskCache() instanceof EnvelopeCache) {
+                    ((EnvelopeCache) options.getEnvelopeDiskCache()).persistCurrentSession(session);
                   }
                 }));
-  }
-
-  /**
-   * This function is mostly called from Flutter where capturing envelope is run in a background
-   * thread.
-   *
-   * <p>Nonetheless we should run this in the same executor service as the deleteCurrentSessionFile
-   * function for consistency.
-   */
-  private static void schedulePersistSession(
-      final @NotNull SentryOptions options, final @NotNull Session session) {
-    if (!(options.getEnvelopeDiskCache() instanceof EnvelopeCache)) {
-      return;
-    }
-    final @NotNull EnvelopeCache cache = (EnvelopeCache) options.getEnvelopeDiskCache();
-    try {
-      options.getExecutorService().submit(() -> cache.persistCurrentSession(session));
-    } catch (RejectedExecutionException e) {
-      options.getLogger().log(WARNING, "Submission of session persisting rejected.", e);
-    }
   }
 
   /** What the events inside an envelope amount to, from the session's point of view. */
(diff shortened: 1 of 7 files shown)


sentry-android-core/src/test/java/io/sentry/android/core/InternalSentrySdkTest.kt
```diff
@@ -452,6 +469,110 @@ class InternalSentrySdkTest {
...
+    val scopeSession = AtomicReference<Session>()
+    Sentry.configureScope { scope -> scopeSession.set(scope.session) }
+    assertThat(scopeSession.get().status).isEqualTo(Session.State.Ok)
+    assertThat(scopeSession.get().hasNonTerminatingUnhandledError()).isTrue()
+    assertThat(scopeSession.get().sessionId).isEqualTo(originalSid.get())
+
+    // and it is persisted so the flag survives process death
+    val sessionFile = EnvelopeCache.getCurrentSessionFile(fixture.options.cacheDirPath!!)
+    val persistedSession =
+      fixture.options.serializer.deserialize(sessionFile.reader(), Session::class.java)!!
```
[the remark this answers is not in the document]
buenaflor: False positive: `JsonSerializer.deserialize` wraps the reader in a try-with-resources `JsonObjectReader`, which closes the underlying reader.

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
```diff
@@ -163,35 +169,25 @@ public static SentryId captureEnvelope(
...
-          }
-        }
       }
 
       // update session and add it to envelope if necessary
-      final @Nullable Session session = updateSession(scopes, options, status, crashedOrErrored);
+      final @Nullable Session.State status =
+          eventState == EnvelopeEventState.UNHANDLED ? Session.State.Crashed : null;
+      final @Nullable Session session =
+          updateSession(scopes, options, status, eventState != EnvelopeEventState.NONE);
```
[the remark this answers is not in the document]
buenaflor: Right, the description was overstating it. `captureEnvelope` keeps its `catch (Throwable)`; only the code added here catches `Exception`. Description updated.

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
```diff
@@ -213,6 +209,122 @@ public static SentryId captureEnvelope(
...
+  /**
+   * Reads an envelope from the given bytes. Besides the declared {@link IOException}, {@link
+   * io.sentry.IEnvelopeReader#read(InputStream)} also rejects malformed payloads with an unchecked
+   * {@link IllegalArgumentException}, hence the broader catch.
+   */
+  private static @Nullable SentryEnvelope readEnvelope(
+      final @NotNull SentryOptions options, final @NotNull byte[] envelopeData) {
+    try (final InputStream envelopeInputStream = new ByteArrayInputStream(envelopeData)) {
+      return options.getEnvelopeReader().read(envelopeInputStream);
+    } catch (Exception e) {
```
runningcode: We should only catch the exceptions expect to throw here. Looks like we expect `IOException` and `IllegalArgumentException` ?

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
runningcode, line 206: since we arent doing disk I/O inside the try/catch, why do we need this? which parts can throw `Exception`s ?

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
```diff
@@ -213,6 +209,122 @@ public static SentryId captureEnvelope(
...
+                      options
+                          .getLogger()
+                          .log(INFO, "Session is null on captureEnvelopeNonTerminating");
+                    }
+                  });
+            });
+      }
+
+      return scopes.captureEnvelope(envelope);
+    } catch (Exception e) {
```
runningcode, line 254: We should try to capture the narrowest exception possible as written in the AGENTS.md file now. Might need to rebase to get this update.

runningcode (commented):
This looks good aside from the exception handling!

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
buenaflor: ah yes, thx, will update it

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
buenaflor: in this case we are since Exception is the narrowest that compiles here, eventStateOf reads events via SentryEnvelopeItem.getEvent → getData() → dataFactory.call(), and dataFactory is a java.util.concurrent.Callable<byte[]>, whose call() declares throws Exception

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
buenaflor: there is disk I/O (`deleteCurrentSessionFile`)

however due to eventStateOf which declares throws Exception we can't narrow it down 

should we use `rethrowIfFatal`? it's referenced in the AGENTS.md

runningcode: IMO, we should only capture `IOException` if that is the only exception that can happen here. otherwise we will silenty drop errors that we don't expect and we won't know if this feature works.

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
runningcode: ah dang. i feel like we should use another primitive that doesn't throw such a broad exception becuase of this reason. usually callables are passed to executors. oh well. this isn't related to your PR.

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
```diff
@@ -213,6 +209,132 @@ public static SentryId captureEnvelope(
...
+    }
+    if (unhandled) {
+      return EnvelopeEventState.UNHANDLED;
+    }
+    return errored ? EnvelopeEventState.ERRORED : EnvelopeEventState.NONE;
+  }
+
+  /**
+   * Reads an envelope from the given bytes. {@link io.sentry.IEnvelopeReader#read(InputStream)}
+   * declares {@link IOException} and additionally rejects malformed payloads with an unchecked
```
runningcode: what does this mean? seems like we are catching `IllegalArgumentException` here.

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
```diff
@@ -213,6 +209,132 @@ public static SentryId captureEnvelope(
...
+        scope -> {
+          scope.withSession(
+              session -> {
+                if (session != null) {
+                  final boolean updated =
+                      crashed
+                          ? session.recordNonTerminatingUnhandledError()
+                          : session.update(null, null, true, null);
+                  if (updated && options.getEnvelopeDiskCache() instanceof EnvelopeCache) {
+                    ((EnvelopeCache) options.getEnvelopeDiskCache()).persistCurrentSession(session);
```
runningcode, line 298: From the previous PR, this grabs a lock and writes to disk. We need to be absolutely sure that this is never called from the main thread. Can you double check that? My clanker seems to think this method is called from the main thread.

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
```diff
@@ -213,6 +209,132 @@ public static SentryId captureEnvelope(
...
+   * outside it lets a concurrent caller's older snapshot land last and drop the unhandled marker.
+   *
+   * @param crashed {@code true} if the error was unhandled ({@code mechanism.handled=false})
+   */
+  private static void updateSessionNonTerminating(final boolean crashed) {
+    final @NotNull IScopes scopes = ScopesAdapter.getInstance();
+    final @NotNull SentryOptions options = scopes.getOptions();
+    scopes.configureScope(
+        scope -> {
+          scope.withSession(
```
runningcode: There are tw locks over a `Session` in this codebase and this leads to a situation where it doesn't seem like we're grabbing the right lock when we need to.

`withSession` acquires `Scope.sessionLock`, but `Session.update()` / `end()` take only `Session.sessionLock` internally. Two different locks. Serializing the live session inside `withSession` is therefore safe only if every writer holds `Scope.sessionLock`, and `updateSession` (line 422) doesn't: it goes through bare configureScope + scope.getSession().

`Session.serialize` also reads its fields without `Session.sessionLock`, so a concurrent crash update lands mid-write. The persisted file can come out as "status": "ok" alongside a finished session's duration and the unhandled marker — a state that never existed at any instant, which Session.Deserializer accepts and previous_session.json then replays.

Cheapest fix I see is routing updateSession through withSession too; `AutoClosableReentrantLock` is reentrant, so the nested clearSession() won't self-deadlock. Let me know if there's a reason it can't move.

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
```diff
@@ -213,6 +209,132 @@ public static SentryId captureEnvelope(
...
+              session -> {
+                if (session != null) {
+                  final boolean updated =
+                      crashed
+                          ? session.recordNonTerminatingUnhandledError()
+                          : session.update(null, null, true, null);
```
runningcode, line 296: `Session.update(null, null, true, null)` returns true purely because `addErrorsCount` incremented the counter, it never checks terminal state, unlike `recordNonTerminatingUnhandledError()` which returns false when `status != Ok`. So updated gates the persist on "we incremented something", and we can write an already-finalized session into `session.json`.

That file means "not delivered yet", so PreviousSessionFinalizer re-sends whatever it finds there on the next launch, same session id twice, and `crashedLastRun` re-armed if the status was Crashed. Reachable via the ANR path: `SentryClient.updateSessionData` ends a session as Abnormal without calling `scope.clearSession()`, so scope.session is still the finalized one when the next handled error arrives here.

Suggest the guard SentryClient already applies at its own call site:
```
if (session.isTerminated()) {
  options.getLogger().log(INFO, "Session already terminated, not recording the error.");
  return;
}
```

runningcode (commented):
Hey, looks like there are some concurrency issues with this change. Let me know if you have questions and we can go through them together!

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
buenaflor: removed the javadocs here

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
buenaflor: switched over to running sync and documenting that this method should be called off the main thread

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
```diff
@@ -213,6 +210,151 @@ public static SentryId captureEnvelope(
...
+      final @NotNull SentryOptions options, final @NotNull Session session) {
+    if (!(options.getEnvelopeDiskCache() instanceof EnvelopeCache)) {
+      return;
+    }
+    final @NotNull EnvelopeCache cache = (EnvelopeCache) options.getEnvelopeDiskCache();
+    try {
+      options.getExecutorService().submit(() -> cache.persistCurrentSession(session));
+    } catch (RejectedExecutionException e) {
+      options.getLogger().log(WARNING, "Submission of session persisting rejected.", e);
+    }
```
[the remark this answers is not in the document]
buenaflor: @runningcode afaict this is real. I think we should just run the persist as sync.

This is only called by Flutter where we run it on a background thread so maybe we should just document it clearly here

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
buenaflor: I updated the codeblocks to narrow it down

sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java
runningcode: don't forget to add this doc to this method! might be worth adding why like "this method writes to disk so it shouldn't be called from the main thread"

runningcode (approved):
looks good! thanks for taking the time to go through this together to get the best solution!

PR #5923, opened 2026-08-10
feat(android): Forward logs.enabled to sentry-native

### LINKED ISSUE #5911: Support logging via native APIs on Android
### Problem Statement

We would like to use the native `sentry_log_*()` / `sentry_log()` APIs in our Android app.

While we can enable logs via `SentryAndroidOptions` (`logs.enabled` / `io.sentry.logs.enabled` manifest key), when the native SDK is initialized through the NDK integration this flag is not passed on:

- `SentryNdk.init()` builds the native options from `NdkOptions`, which only carries DSN, debug, outbox path, release, environment, dist, max breadcrumbs, and SDK name (plus handler strategy, traces sample rate, and app-hang settings) — see [SentryNdk.java](https://github.com/getsentry/sentry-java/blob/main/sentry-android-ndk/src/main/java/io/sentry/android/ndk/SentryNdk.java)
- `NdkOptions` has no logs-related field — see [NdkOptions.java](https://github.com/getsentry/sentry-native/blob/master/ndk/lib/src/main/java/io/sentry/ndk/NdkOptions.java)
- sentry-native options are immutable after `sentry_init()`, which the NDK integration performs — so the app cannot set `enable_logs` itself afterward.

As a result, `enable_logs` is always false in the native SDK on Android, and every call to `sentry_log_*()` / `sentry_log()` from native code returns `SENTRY_LOG_RETURN_DISABLED` and is silently dropped — even when the user has explicitly enabled logs in their Android SDK configuration.

### Solution Brainstorm

- Forward `options.getLogs().isEnabled()` in `SentryNdk.init()` in sentry-java.
- Add an `enableLogs` field to `NdkOptions` and apply it in the NDK init path via `sentry_options_set_enable_logs()`.
- Possibly also forward `logs_with_attributes` and consider routing `beforeSendLog` semantics, though the `enable_logs` flag alone would unblock the feature.

Native logs are sent through the outbox/envelope path, so ingestion should work with the existing transport once the option is enabled.

romtsn:
## Summary

* Forwards `SentryOptions.Logs.isEnabled()` to `NdkOptions.setEnableLogs()` during NDK init
* Unblocks `sentry_log_*()` / `sentry_log()` calls from native code on Android

**Depends on** [https://github.com/getsentry/sentry-native/pull/1971](<https://github.com/getsentry/sentry-native/pull/1971>) — needs a `sentry-native-ndk` release with the new `NdkOptions.enableLogs` field before this can merge.

Fixes getsentry/sentry-java#5911

## Test plan

- [ ] Bump `sentry-native-ndk` version once the dependency PR is released
- [ ] Verify native logs reach Sentry when `logs.enabled = true`
- [ ] Verify native logs are dropped when `logs.enabled = false` (default)

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)

### DIFF
--- a/sentry-android-ndk/src/main/java/io/sentry/android/ndk/SentryNdk.java
+++ b/sentry-android-ndk/src/main/java/io/sentry/android/ndk/SentryNdk.java
@@ -76,6 +76,8 @@ public final class SentryNdk {
         ndkOptions.setEnableAppHangTracking(options.isEnableNdkAppHangTracking());
         ndkOptions.setAppHangTimeoutMillis(options.getNdkAppHangTimeoutIntervalMillis());
 
+        ndkOptions.setEnableLogs(options.getLogs().isEnabled());
+
         //noinspection UnstableApiUsage
         io.sentry.ndk.SentryNdk.init(ndkOptions);
 


romtsn:
We're going to drop the `enableLogs` flag altogether, so there's no need in this PR anymore

PR #5926, opened 2026-08-10
perf(android): Read cpu time via Process.getElapsedCpuTime

### LINKED ISSUE #5925: Optimize AndroidCpuCollector
`AndroidCpuCollector` reads and parses `/proc/self/stat` on every sample to obtain the process cpu time. `collect()` runs every 100ms for the whole duration of every transaction, and each sample allocates the file reader buffers, the intermediate strings and a regex split of all 52 fields — measured at 33623 bytes per call on a Pixel 3 — to extract four numbers.

Reduce the cost of the collector, and of constructing it (the regex `Pattern` is compiled on the `SentryAndroid.init` path, whether or not performance collection ever starts).

runningcode:
Closes getsentry/sentry-java#5925  <br>Linear: getsentry/sentry-java#5925

## 📜 Description

TL;DR We can get the elapsed CPU time by calling `android.os.Process.getElapsedCpuTime()` instead of parsing it from a file. This is more performant and uses less memory and fixes one small bug. I hope it will reduce per device variation/issues. Let's see!

Ok on to the clanker generated details:  <br>`AndroidCpuCollector` obtained the process cpu time by reading `/proc/self/stat` and parsing `utime`/`stime`/`cutime`/`cstime` out of it. It now calls `android.os.Process.getElapsedCpuTime()`, which AOSP implements as a `@CriticalNative` wrapper around `clock_gettime(CLOCK_PROCESS_CPUTIME_ID)`.

That removes, per sample:

* the `FileReader`/`BufferedReader` buffers (an 8 kB `StreamDecoder` byte buffer plus a 16 kB `char[]`), the `StringBuilder`, the line `String` and the `toString()` copy
* a regex split of all 52 fields of `/proc/self/stat` — a `Matcher`, an `ArrayList` and \~52 `String`s, of which 4 were used
* five syscalls (`readText` stats the file three times via `exists`/`isFile`/`canRead` before the open)

It also removes the `Pattern.compile()` from the constructor, which ran on the `SentryAndroid.init` path whether or not performance collection ever started, along with the `_SC_CLK_TCK` sysconf and the IO/parse error handling that `isEnabled = false` existed for.

The new source excludes the cpu time of reaped child processes, which an app process does not have, and has millisecond rather than clock-tick (10 ms) resolution.

Second, unrelated fix in the same file: `setup()` seeded `lastCpuNanos` but not `lastRealtimeNanos`, so the first sample of every transaction divided the cpu delta by the time since device boot and reported \~0%.

No public API change (`apiDump` produces no diff).

## 💡 Motivation and Context

`collect()` runs every 100 ms for the whole duration of every transaction. On a Pixel 3 each call allocated 33623 bytes, i.e. \~336 kB/s of garbage while a transaction is in flight, to read four numbers out of a \~300 byte file.

| per `collect()` | before | after |
| -- | -- | -- |
| bytes allocated | 33623 B | 16 B |
| wall time | 478 µs | 2.95 µs |

The remaining 16 bytes are the `Double` boxing in `setCpuUsagePercentage`, which this PR does not touch.

## 💚 How did you test it?

Existing unit tests pass. They can't cover the value itself — with `isReturnDefaultValues = true` the android.jar stubs return 0 for both clocks, so `when collect cpu is collected` really asserts `NaN != 0.0`, before and after this change.

So verification was done on a **Pixel 3 (blueline, Android 12 / API 31, 8 cores)** by dexing the class files out of the release build output and driving them with `app_process`.

**1.** `getElapsedCpuTime()` **agrees with** `/proc/self/stat`**:**

```
_SC_CLK_TCK = 100, _SC_NPROCESSORS_CONF = 8

idle 1000ms            wall=1002ms  getElapsedCpuTime=   1ms  /proc/self/stat=   0ms  diff= 1ms
busy 1 thread 1000ms   wall=1002ms  getElapsedCpuTime=1005ms  /proc/self/stat=1000ms  diff= 5ms
busy 4 threads 1000ms  wall=1003ms  getElapsedCpuTime=3966ms  /proc/self/stat=3970ms  diff=-4ms
busy 8 threads 1000ms  wall=1005ms  getElapsedCpuTime=7701ms  /proc/self/stat=7710ms  diff=-9ms
```

The unit is milliseconds, it counts all threads, and it tracks the `/proc` value to within one clock tick (10 ms) — the residual is the tick quantization on the `/proc` side.

**2. The collector reports the same percentages as before**, sampled at the 100 ms interval `DefaultCompositePerformanceCollector` uses:

```
idle                                   1.1%    1.4%    0.0%    0.4%    0.5%
1 of 8 cores busy (expect ~12.5%)      4.7%   12.8%   12.7%   12.7%   12.3%
4 of 8 cores busy (expect ~50%)       16.5%   49.7%   49.9%   49.5%   50.1%
8 of 8 cores busy (expect ~100%)      33.4%   95.3%   98.5%   96.0%   98.1%
```

(The low first value in each row is the load threads ramping up inside that 100 ms window.) The pre-change collector run through the same harness produces the same numbers.

**3. The first sample after** `setup()`**, taken under full load:**

```
before:  0.0%
after:  92.9%
```

**4. Allocation and timing**, via `Debug.getRuntimeStat("art.gc.bytes-allocated")` over 20k `collect()` calls, two runs each — the table above. Note `app_process` runs imageless (JIT-cold), so the absolute µs figures are inflated; the byte counts are exact.

## :pencil: Checklist

## 🔮 Next steps

`AndroidMemoryCollector` runs on the same 100 ms timer and is worth the same look.

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidCpuCollector.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidCpuCollector.java
@@ -11,13 +11,11 @@ import io.sentry.util.Objects;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 
-// The approach to get the cpu usage info was taken from
-// https://eng.lyft.com/monitoring-cpu-performance-of-lyfts-android-applications-4e36fafffe12
-// The process cpu time itself comes from Process.getElapsedCpuTime(), a @CriticalNative wrapper
-// around clock_gettime(CLOCK_PROCESS_CPUTIME_ID), rather than from parsing /proc/self/stat: reading
-// and parsing that file allocated on every sample, and collect() runs 10 times per second for the
-// whole duration of a transaction. It does not include the cpu time of reaped child processes,
-// which an app process doesn't have.
+// The process cpu time comes from Process.getElapsedCpuTime(), a @CriticalNative wrapper around
+// clock_gettime(CLOCK_PROCESS_CPUTIME_ID), rather than from parsing /proc/self/stat: reading and
+// parsing that file allocated on every sample, and collect() runs 10 times per second for the whole
+// duration of a transaction. It does not include the cpu time of reaped child processes, which an
+// app process doesn't have.
 @ApiStatus.Internal
 public final class AndroidCpuCollector implements IPerformanceSnapshotCollector {
 
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidCpuCollector.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidCpuCollector.java
@@ -11,11 +11,12 @@ import io.sentry.util.Objects;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 
-// Process.getElapsedCpuTime() is a @CriticalNative wrapper around
-// clock_gettime(CLOCK_PROCESS_CPUTIME_ID), i.e. the cpu time of all threads of this process. It
-// replaces reading utime/stime out of /proc/self/stat, which allocated ~25 kB per sample (file
-// reader buffers plus a regex split of all 52 fields) while collect() runs 10 times per second for
-// the whole duration of a transaction. It does not include the cpu time of reaped child processes,
+// The approach to get the cpu usage info was taken from
+// https://eng.lyft.com/monitoring-cpu-performance-of-lyfts-android-applications-4e36fafffe12
+// The process cpu time itself comes from Process.getElapsedCpuTime(), a @CriticalNative wrapper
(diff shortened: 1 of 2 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/AndroidCpuCollector.java
```diff
@@ -1,56 +1,44 @@
...
-import io.sentry.SentryLevel;
-import io.sentry.util.FileUtils;
 import io.sentry.util.Objects;
-import java.io.File;
-import java.io.IOException;
-import java.util.regex.Pattern;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 
 // The approach to get the cpu usage info was taken from
```
romtsn: i guess we could drop this part and dismiss the warden's review

romtsn (approved):
dope 🚀 as mentioned, worth taking a look at the MemoryCollector to see if we can also improve it too

PR #5930, opened 2026-08-11
docs(agents): Remove duplicated agent rule files

runningcode:
## :scroll: Description

Removes two redundant agent rule files and consolidates the rule index into `AGENTS.md`.

- **`.cursor/rules/coding.mdc`** — reduced to a pointer at `AGENTS.md`. Its "Contributing Guidelines", "Coding rules", "Useful Resources", and "Key Commands" sections were near-verbatim duplicates of sections already in `AGENTS.md`, including the Google Truth paragraph in full. The file stays (with `alwaysApply: true`) so Cursor still has an entry point.
- **`.cursor/rules/overview_dev.mdc`** — deleted. It described each rule twice (a bullet list per rule, then a keyword→rule table), while every rule file already carries the same information in its own frontmatter `description`, which is what Cursor uses for auto-selection. It also instructed agents to use `fetch_rules`, a Cursor-only tool unavailable to other agents.
- **`AGENTS.md`** — the "STOP — Required Reading" block is replaced by a single table mapping each rule to when it applies.
- **`CLAUDE.md`** — updated to match, since it referenced the removed section.

Agents now read one file to get oriented instead of three.

The required-reading chain made every session load `AGENTS.md` → `coding.mdc` → `overview_dev.mdc` before doing anything, and two of those three reads added no information. Duplicated guidance across files also drifts: the same rules were maintained in several places at once.

I did what Alex suggested, asked claude code `which otel modes exist for java sdk?` and then `which context files did you use so far?`, it printed 
```
Two, both read in full at the start of this turn:

1. AGENTS.md — loaded per the instruction in CLAUDE.md (the project CLAUDE.md is injected into context automatically; I read AGENTS.md with the Read tool). It's what pointed me at the opentelemetry rule.
2. .cursor/rules/opentelemetry.mdc — the domain rule for sentry-opentelemetry-*, agent vs agentless, span processing, and OTLP.
```

## :pencil: Checklist

#skip-changelog


CLAUDE.md
```diff
@@ -1,6 +1,5 @@
 # CLAUDE.md
 
-## STOP — Required Reading (Do This First)
-
-Before doing ANYTHING else (including answering questions), you MUST use the Read tool to load [AGENTS.md](AGENTS.md) and follow ALL of its instructions, including reading the required `.cursor/rules/*.mdc` files it references.
-Do NOT skip this step. Do NOT proceed without reading these files first.
+Read [AGENTS.md](AGENTS.md) and follow its instructions. It is the single source of truth
```
runningcode: The language is now less strong here. If we feel strongly about needing the strong language, I can add it back.

.cursor/rules/coding.mdc
```diff
@@ -5,49 +5,10 @@ description: Cursor Coding Rules
...
-6. Once you're done, format the code and regenerate the .api files using the following command `./gradlew spotlessApply apiDump`
-7. As a last step, git stage the relevant files and propose (but not execute) a single git commit command (e.g. `git commit -m "<git commit message>"`)
-
-
-## Useful Resources
-
-- Main SDK documentation: https://develop.sentry.dev/sdk/overview/
-- Internal contributing guide: https://docs.sentry.io/internal/contributing/
-- Git commit messages conventions: https://develop.sentry.dev/engineering-practices/commit-messages/
+Before doing ANYTHING else (including answering questions), you MUST read
```
adinauer: Is this just for Cursor? Shouldn't that be auto loading `AGENTS.md` anyways and render this irrelevant? In that case we could drop the file completely.

AGENTS.md
```diff
@@ -2,15 +2,30 @@
...
-Then identify and read any topically relevant `.cursor/rules/*.mdc` files for the area you're working on (e.g., `opentelemetry.mdc` for OTel work, `metrics.mdc` for metrics work). Use the Glob tool on `.cursor/rules/*.mdc` to discover available rule files.
-
-Do NOT skip this step. Do NOT proceed without reading these files first.
+## Domain-Specific Rules
+
+This file covers the whole repository. Before working on a specific area, read the matching
+rule file in `.cursor/rules/`:
+
+| Rule | Read it when working on |
+|---|---|
```
adinauer, line 11: Should we also port over the more detailed instructions to ensure it's able to find the right rules files?

runningcode: It was able to find the opentelemetry file with these rules, do you have a specific example I could test to make sure?

.cursor/rules/coding.mdc
runningcode: I didn't know that. I'll remove this then!

markushi (approved):

PR #5933, opened 2026-08-11
perf(core): Remove per-sample allocations from performance collection

runningcode:
While a transaction or profile chunk is running, `DefaultCompositePerformanceCollector` collects performance data every 100ms for up to 30 seconds (~300 samples per transaction). Two allocation sources on that path were doing avoidable work.

### Boxed measurements

`PerformanceCollectionData` stored its measurements as `@Nullable Double`/`@Nullable Long`, using `null` for "no collector reported this". Heap sizes are far outside the `Long` cache, so every sample allocated a box per measurement — 2 per sample on Android, ~600 per 30s transaction.

The fields are now primitives with an explicit presence flag, and consumers call `hasUsedHeapMemory()` etc. instead of null-checking.

I first tried sentinel values (`-1` for memory, matching the pre-8.0 `MemoryCollectionData`, and `NaN` for cpu) to avoid the extra fields. `AndroidCpuCollectorTest` caught that `AndroidCpuCollector` can legitimately produce `NaN` — it computes `cpuNanosDiff / realTimeNanosDiff` and observes a zero-length interval on the first collection after `setup()`. Since no in-band value is safe, the presence flags are explicit.

### Redundant clock reads

`CompositeData.addDataAndCheckTimeout` called `options.getDateProvider().now().nanoTimestamp()` once *per in-flight transaction* to check the 30s timeout, allocating a `SentryNanotimeDate` (plus `System.currentTimeMillis()` + `System.nanoTime()`) each time — even though the enclosing timer task had just read the same clock for the sample it was distributing. That reading is now passed in, so a collection round reads the clock once rather than once per transaction.

### Notes

No behavior change: which measurements get reported, and their values, are unchanged.

`PerformanceCollectionData` is `@ApiStatus.Internal`, so the signature changes in `sentry.api` are not part of the public surface.

`sentry-android-core` did not have Truth wired up; added `testImplementation(libs.google.truth)` per `AGENTS.md` for the tests touched here.

This came out of a review of the memory collector. Two things I found but deliberately left alone:

- `AndroidMemoryCollector` computes native memory as `Debug.getNativeHeapSize() - Debug.getNativeHeapFreeSize()`, which is **two** `mallinfo()` calls where `getNativeHeapAllocatedSize()` would be one. On Scudo, `mallinfo()` reaches `GlobalStats::get()`, which takes a global mutex and walks every thread's stats cache — so each sample takes that allocator-wide lock twice, 10x/second, while profiling. Worth fixing, but `usmblks - fordblks` and `uordblks` are not the same number, so it would change reported `memory_native_footprint` values. That deserves its own PR and a measurement.
- The double `Runtime.getRuntime()` call in `AndroidMemoryCollector` is a static field read that folds away; not worth touching.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Fixes
 
+- Remove per-sample allocations from performance data collection ([#5933](https://github.com/getsentry/sentry-java/pull/5933))
+  - `PerformanceCollectionData` now stores its measurements as primitives instead of boxed types, and the 30s transaction timeout check reads the clock once per collection round instead of once per in-flight transaction
 - Clear contexts when calling `Scope.clear()` ([#5902](https://github.com/getsentry/sentry-java/pull/5902))
 - Preserve custom `Throwable` identities when R8 optimizes Android apps ([#5881](https://github.com/getsentry/sentry-java/pull/5881))
 
--- a/sentry-android-core/build.gradle.kts
+++ b/sentry-android-core/build.gradle.kts
@@ -115,6 +115,7 @@ dependencies {
   testImplementation(libs.androidx.test.ext.junit)
   testImplementation(libs.androidx.test.runner)
   testImplementation(libs.awaitility.kotlin)
+  testImplementation(libs.google.truth)
   testImplementation(libs.mockito.kotlin)
   testImplementation(libs.mockito.inline)
   testImplementation(projects.sentryTestSupport)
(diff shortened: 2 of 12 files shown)


runningcode:
Split into two independent PRs so each concern can be reviewed on its own:

- #5934 — read the clock once per performance collection round
- #5935 — store performance measurements as primitives

Both are branched off current `main` and touch disjoint files, so they can merge in either order.

PR #5935, opened 2026-08-11
perf(core): Store performance measurements as primitives

runningcode:
## :scroll: Description

While a transaction or profile chunk is running, `DefaultCompositePerformanceCollector` collects performance data every 100ms for up to 30 seconds (~300 samples per transaction). Every sample allocated a box per measurement — 2 per sample on Android, ~600 per 30s transaction.

The fields are now primitives with an explicit presence flag, and consumers call `hasUsedHeapMemory()` etc. instead of null-checking.

Reduce boxing and improve performance.

Unit tests pass

## :pencil: Checklist


romtsn (approved):
clever!

PR #5940, opened 2026-08-12
feat(logback): [Logs and Metrics Enable Flags 1] Add Logs opt-in

adinauer:
## PR Stack (Logs and Metrics Enable Flags)

- #5939
- #5940
- #5941
- #5942
- #5943
- #5945
- #5946
- #5947
- #5949
- #5950
- #5951
- #5952
- #5953
- #5954
- #5955
- #5956
- #5957
- #5960
- #5961
- #5964
- #5966
- #5970

---

## :scroll: Description

Adds a `logsEnabled` option to the Logback `SentryAppender`. The option defaults to `false` and can be configured through Java or Logback XML.

The appender now requires both this local opt-in and the existing aggregate core Logs flag before forwarding Logback records as Sentry Logs. Event and breadcrumb capture remain unchanged.

Logging integrations need explicit local opt-ins before the aggregate core Logs flag can be removed later in this stack. This prevents applications from unexpectedly forwarding framework logs when core Logs capture becomes available without a global enable flag.

- `./gradlew :sentry-logback:test`
- Added tests for the default-disabled behavior, explicit opt-in, and independent event/breadcrumb capture

## :pencil: Checklist

## :crystal_ball: Next steps

Continue adding integration-local Logs opt-ins before removing the aggregate core Logs flag.

### DIFF
--- a/sentry-logback/api/sentry-logback.api
+++ b/sentry-logback/api/sentry-logback.api
@@ -14,6 +14,8 @@ public class io/sentry/logback/SentryAppender : ch/qos/logback/core/Unsynchroniz
 	public fun getMinimumBreadcrumbLevel ()Lch/qos/logback/classic/Level;
 	public fun getMinimumEventLevel ()Lch/qos/logback/classic/Level;
 	public fun getMinimumLevel ()Lch/qos/logback/classic/Level;
+	public fun isEnableLogs ()Z
+	public fun setEnableLogs (Z)V
 	public fun setEncoder (Lch/qos/logback/core/encoder/Encoder;)V
 	public fun setMinimumBreadcrumbLevel (Lch/qos/logback/classic/Level;)V
 	public fun setMinimumEventLevel (Lch/qos/logback/classic/Level;)V
--- a/sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
+++ b/sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
@@ -52,6 +52,7 @@ public class SentryAppender extends UnsynchronizedAppenderBase<ILoggingEvent> {
   private @NotNull Level minimumBreadcrumbLevel = Level.INFO;
   private @NotNull Level minimumEventLevel = Level.ERROR;
   private @NotNull Level minimumLevel = Level.INFO;
+  private boolean enableLogs = false;
   private @Nullable Encoder<ILoggingEvent> encoder;
 
   static {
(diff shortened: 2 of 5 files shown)


sentry-logback/api/sentry-logback.api
```diff
@@ -14,6 +14,8 @@ public class io/sentry/logback/SentryAppender : ch/qos/logback/core/Unsynchroniz
 	public fun getMinimumBreadcrumbLevel ()Lch/qos/logback/classic/Level;
 	public fun getMinimumEventLevel ()Lch/qos/logback/classic/Level;
 	public fun getMinimumLevel ()Lch/qos/logback/classic/Level;
+	public fun isEnableLogs ()Z
```
runningcode, line 17: would it be more idiomatic to make this `logsEnabled()`?

sentry-logback/api/sentry-logback.api
```diff
@@ -14,6 +14,8 @@ public class io/sentry/logback/SentryAppender : ch/qos/logback/core/Unsynchroniz
...
+	public fun setEnableLogs (Z)V
```
runningcode, line 18: same suggestion as above
```suggestion
	public fun setLogsEnabled (Z)V
```

sentry-logback/src/main/java/io/sentry/logback/SentryAppender.java
```diff
@@ -52,6 +52,7 @@ public class SentryAppender extends UnsynchronizedAppenderBase<ILoggingEvent> {
   private @NotNull Level minimumBreadcrumbLevel = Level.INFO;
   private @NotNull Level minimumEventLevel = Level.ERROR;
   private @NotNull Level minimumLevel = Level.INFO;
+  private boolean enableLogs = false;
```
runningcode, line 55:
```suggestion
  private boolean logsEnabled = false;
```

runningcode (approved):
I put a global suggestion on the naming, probably worth a discussion with the team but otherwise going to approve

sentry-logback/api/sentry-logback.api
adinauer: Afaik other SDKs are calling the option `enableLogs` too, so I wanted to align.

runningcode: good point, this is the classic tension of consistency vs feeling native to the platform. Given our new goal hierarchy, I would think it should be the one that feels native.

runningcode: Discussed in slack, we should make this feel native.

runningcode (changes requested):
After a discussion, I think we should make this feel native!

sentry-logback/api/sentry-logback.api
lbloder: Adding my 2 cents here: I do believe the getter should still follow JavaBean conventions and be `isLogsEnabled()`.

My reasoning here:
- Getters through the whole SDK use this convention
- Frameworks like Spring (Boot) may rely on this convention as well

Issue #5518, opened 2026-06-09
Deadlock between ReplayIntegration.close() and connectivity callbacks blocks main thread (ANR)


buenaflor:
### Integration

sentry-android

### Build System

Gradle

### AGP Version

8.6.0

### Proguard

Disabled

### Other Error Monitoring Solution

No

### Version

8.41.3

### Steps to Reproduce

1. `SentryAndroid.init` with Session Replay enabled (e.g. `sessionSampleRate = 1.0`) — ReplayIntegration registers as a connection status observer and `AndroidConnectionStatusProvider`
  registers a `ConnectivityManager` network callback.
2. Call `Sentry.close()` from the main thread shortly after init (or whenever a connectivity notification is in flight — Android delivers the current network state immediately after
  callback registration, so closing soon after init hits this reliably).
  3. Race: loop init→close with replay enabled on a loaded emulator; reproduces within ~30 cycles.

### Expected Result

Sentry.close() completes and the SDK shuts down cleanly.

### Actual Result

`Sentry.close()` never returns — AB-BA deadlock between `ReplayIntegration.lifecycleLock` and `AndroidConnectionStatusProvider.lock`. When `close()` runs on the main thread this is a permanent
  main-thread block (ANR).

  - **Closing thread**: `ReplayIntegration.close()` holds lifecycleLock (`ReplayIntegration.kt:365`) → `removeConnectionStatusObserver()` waits on the provider lock
  (`AndroidConnectionStatusProvider.java:441`).
  - **ConnectivityThread**: `updateCacheAndNotifyObservers` holds the provider lock while notifying observers (`AndroidConnectionStatusProvider.java:274`) →
  `ReplayIntegration.onConnectionStatusChanged` → `resumeInternal()` waits on lifecycleLock (`ReplayIntegration.kt:220`).

```
  "main" Waiting
    at io.sentry.android.core.internal.util.AndroidConnectionStatusProvider.removeConnectionStatusObserver(AndroidConnectionStatusProvider.java:441)
    at io.sentry.android.replay.ReplayIntegration.close(ReplayIntegration.kt:370)
    at io.sentry.Scopes.close(Scopes.java:439)
    at io.sentry.Sentry.close(Sentry.java:727)

  "ConnectivityThread" Waiting
    at io.sentry.android.replay.ReplayIntegration.resumeInternal(ReplayIntegration.kt:220)
[...]

PR #5965, opened 2026-08-13
ref(android): Confine replay lifecycle to main thread

romtsn:
## :scroll: Description

Serialize Session Replay lifecycle changes through the Android main looper and keep the active lifecycle, capture strategy, replay ID, and generation in one atomic state value. Every lifecycle command is queued, including calls already made from main, which preserves ordering across callers and defers replay startup beyond the SDK initialization critical path.

This removes the replay lifecycle and encoder locks that could block Android lifecycle callbacks and cause deadlocks or ANRs. Encoder work and cache cleanup remain ordered on the replay executor. Close performs main-thread teardown before executor shutdown; if a background close times out, the already-queued teardown still shuts the executors down after cleanup has been scheduled.

Event correlation remains synchronous where required. The internal `ReplayController.captureReplay()` contract now returns the active replay ID and sets it on the scope before deferred conversion. Terminating capture reaches the active strategy immediately so a main-thread crash cannot block it behind the looper. Deferred capture callbacks validate the replay generation, ID, and strategy before changing state, preventing an old replay from mutating a restarted one.

This is a breaking internal API change for hybrid SDKs that implement or bridge `ReplayController`. Returning the ID binds the triggering event or feedback to the exact sampled replay snapshot; falling back to the scope preserves an existing association when no new capture is accepted.

Prepare Session Replay for public start and stop APIs without allowing calls from arbitrary threads to race, deadlock, or block Android lifecycle callbacks.

Always queuing startup also addresses the Session Replay initialization delay tracked in JAVA-656.

Refs JAVA-665
Refs JAVA-656

- Ran all Session Replay unit tests: 241 passed and 1 skipped.
- Ran the full `SentryAndroidTest` suite.
- Added focused coverage for lifecycle ordering, concurrent capture and stop, stale callbacks, crash termination, background-close timeouts, executor cleanup ordering, and deferred foreground initialization.

## :pencil: Checklist

## :crystal_ball: Next steps

- Update hybrid SDK bridges for the `ReplayController.captureReplay()` return type.
- Add the public Session Replay start and stop APIs in JAVA-325.
- Document manual replay lifecycle control in JAVA-691.


sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
```diff
@@ -152,28 +150,26 @@ public class ReplayCache(private val options: SentryOptions, private val replayI
...
+        )
+        .apply {
+          // the constructor already opened the MediaMuxer, so release it if start() fails,
+          // otherwise the encoder is never assigned and its resources leak (CloseGuard warning)
+          try {
+            start()
+          } catch (t: Throwable) {
+            release()
+            throw t
           }
```
runningcode, line 171: do you know why we catch and rethrow the throwable here instead of a `finally` ?
```suggestion
          } finally {
            release()
          }
```

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
```diff
@@ -380,63 +384,88 @@ public class ReplayIntegration(
...
+    }
 
-      options.connectionStatusProvider.removeConnectionStatusObserver(this)
-      scopes?.rateLimiter?.removeRateLimitObserver(this)
-      stop()
-      recorder?.close()
-      recorder = null
-      rootViewsSpy.close()
-      lifecycle.currentState = CLOSED
+    val isMainThread = Looper.myLooper() == Looper.getMainLooper()
```
runningcode: how come we don't use options.threadChecker.isMainThread here ?

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
```diff
@@ -169,82 +169,84 @@ public class ReplayIntegration(
...
-      }
-
-      if (!lifecycle.isAllowed(STARTED)) {
-        options.logger.log(
-          DEBUG,
-          "Session replay is already being recorded, not starting a new one",
-        )
-        return
-      }
+    postOnMainThread { startInternal() }
```
runningcode: `postOnMainThread` is a bit of a misnomer since it doesn't post if we're already on the main thread.
Two issues:
1. when called during startup, it doesn't post it for later meaning it doesn't improve the startup time.
2. calls can get out of order. not sure if this was an explicit goal, if we always post to the main thread, then the events are queue in order, if we sometimes post and sometimes execute inline then we dno't have any ordering guarantee if these are called from different threads.

runningcode (commented):
left two thoughts!

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
```diff
@@ -165,110 +165,170 @@ public class ReplayIntegration(
...
+    strategy.start()
+    val replayId: SentryId? = strategy.currentReplayId
+    state.set(
+      ReplayState(
+        generation = current.generation + 1,
+        lifecycleState = STARTED,
+        replayId = replayId ?: SentryId.EMPTY_ID,
+        captureStrategy = strategy,
+      )
+    )
```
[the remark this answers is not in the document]
romtsn: This ordering is intentional. We publish the ReplayState only after strategy.start() so other threads cannot observe a strategy before its cache and replay ID are initialized. WindowRecorder.start() only flips isRecording; actual screenshot capture starts later from onConfigurationChanged() through the delayed capturer, and main-thread callbacks cannot interleave with startInternal(). A background capture during initialization correctly treats replay as not ready yet, so there is no frame-loss window here.

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
romtsn: good point, addressed in https://github.com/getsentry/sentry-java/pull/5965/commits/b72ae831b65221bbfbc6b40d200a5deb630a3cbb

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
romtsn: that's intentional - we don't want to release() on the happy path because we're gonna use the encoder right after and release after encoding a segment

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
romtsn: yeah that's a great observation, but you were too soon to review 😅 I reworked this in https://github.com/getsentry/sentry-java/pull/5965/commits/c9a7891a5e6a4ab422777cdd8d466b1b80e7669d and now we always enqueue on the main thread + use snapshot-style state management to make it truly lock-free

romtsn: sorry about marking it as ready, was mainly to get bots' review

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayCache.kt
```diff
@@ -209,20 +205,15 @@ public class ReplayCache(private val options: SentryOptions, private val replayI
...
 
-    var videoDuration: Long
-    encoderLock.acquire().use {
-      encoder?.release()
-      videoDuration = encoder?.duration ?: 0
-      encoder = null
-    }
+    encoder?.release()
+    val videoDuration = encoder?.duration ?: 0
+    encoder = null
```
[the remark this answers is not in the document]
romtsn: Duration is retained in memory, and `release()` drains remaining frames first. Reading beforehand could be less accurate.

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
```diff
@@ -165,110 +164,175 @@ public class ReplayIntegration(
...
+        "Replay wasn't sampled by onErrorSampleRate, not capturing for event",
+      )
+      return SentryId.EMPTY_ID
+    }
+
+    // Set it synchronously so the event that triggered the flush picks it up before conversion.
+    scopes?.configureScope { it.replayId = current.replayId }
+    if (isTerminating == true) {
+      // A main-thread crash blocks the looper while flushing, so mark termination synchronously.
+      current.captureStrategy?.captureReplay(true) {}
```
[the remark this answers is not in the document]
romtsn: that's by design. `true` comes from an unhandled/crashing event; the strategy is intentionally terminal and recovery happens next launch.

runningcode (approved):
Nice! Thanks for addressing the PR comments!

GoldenMars:
Hi @romtsn @buenaflor - when will sentry_flutter bundle sentry-android ≥ 8.54.0 with this fix?
Any rough ETA for the Flutter SDK release that includes it?
We're on sentry_flutter 9.13.0 (sentry-android 8.32.0); the #5518 replay deadlock is causing a significant ANR spike in production for us, so getting this fix into Flutter is very important.

buenaflor:
@GoldenMars I'll merge `8.54.0` today into Flutter then have a release ready by tomorrow

PR #5969, opened 2026-08-14
perf(android): Use generated manifest metadata

romtsn:
## :scroll: Description
Use manifest metadata injected by the Sentry Android Gradle plugin when available. The existing PackageManager Bundle remains the fallback, and all existing runtime option precedence remains unchanged.

Reading ApplicationInfo metadata and converting every manifest option during startup showed up prominently in customer traces. The paired plugin change in getsentry/sentry-android-gradle-plugin#1405 resolves typed metadata from the final merged manifest and avoids that runtime work, reducing median SDK initialization time by 6.5% in the all-options benchmark.

Ran the ManifestMetadataReader release unit tests, including typed generated values, absent keys, auto-init, PackageManager fallback, and no Context access on the generated path. Also ran spotlessApply and apiDump.

## :pencil: Checklist
- [x] No new PII added or SDK only sends newly added PII if sendDefaultPII is enabled.

## :crystal_ball: Next steps

Release with getsentry/sentry-android-gradle-plugin#1405 to enable the generated path.


sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
```diff
@@ -18,13 +18,18 @@
...
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 /** Class responsible for reading values from manifest and setting them to the options */
 final class ManifestMetadataReader {
 
+  // Populated by the Sentry Android Gradle plugin with manifest metadata resolved at build time.
+  // When set, it is authoritative: the manifest is not read at runtime.
+  static @Nullable Map<String, Object> buildTimeMetadata;
```
runningcode: I wonder if we should separate the data to a separate class that just holds the data. That would make refactorings of `ManifestMetadataReader` easier and separate the "reading" from the "data" holder concern.

romtsn: that's a great idea! However, I think the bytecode changes we're making from the gradle plugin side are very minimal (just a static field assignment), so we wouldn't gain too much by this (and I don't expect us to refactor this class in a way that it affects this static field)

sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
```diff
@@ -828,11 +856,21 @@ private static boolean readBool(
...
+      if (bundleValue == -1) {
+        bundleValue = ((Integer) ((Bundle) metadata).getInt(key, -1)).doubleValue();
+      }
+      value = bundleValue;
+    } else {
+      final Object raw = ((Map<?, ?>) metadata).get(key);
+      value =
+          raw instanceof Float
+              ? ((Float) raw).doubleValue()
+              : raw instanceof Integer ? ((Integer) raw).doubleValue() : -1;
```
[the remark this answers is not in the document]
romtsn: we don't inject Doubles from the plugin, only Floats and Integers

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Performance
+
+- Use manifest metadata resolved at build time, reducing Android SDK initialization time by 6.5% ([#5969](https://github.com/getsentry/sentry-java/pull/5969))
```
runningcode: i think we shouldn't quote specific percentages in the release notes since it only came from your test device and isn't wildly applicable

runningcode: i think we shouldn't quote specific percentages in the release notes since it only came from your test device and isn't wildly applicable

sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
```diff
@@ -828,11 +856,21 @@ private static boolean readBool(
...
-      final @NotNull Bundle metadata, final @NotNull ILogger logger, final @NotNull String key) {
-    // manifest meta-data only reads float
-    double value = ((Float) metadata.getFloat(key, -1)).doubleValue();
-    if (value == -1) {
-      value = ((Integer) metadata.getInt(key, -1)).doubleValue();
+      final @NotNull Object metadata, final @NotNull ILogger logger, final @NotNull String key) {
+    final double value;
+    if (metadata instanceof Bundle) {
+      // manifest meta-data only reads float
+      double bundleValue = ((Float) ((Bundle) metadata).getFloat(key, -1)).doubleValue();
```
runningcode: it looks so ugly. i wonder if there's a way to simplify this. like what if the bytecode generator created a bundle, would that make this simpler?

runningcode (approved):
LGTM from the sentry-java side. i just had one thought but not sure how feasible it is.

sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
runningcode: it looks so ugly. i wonder if there's a way to simplify this. like what if the bytecode generator created a bundle, would that make this simpler?

CHANGELOG.md
runningcode: i think we shouldn't quote specific percentages in the release notes since it only came from your test device and isn't wildly applicable

runningcode (approved):
LGTM from the sentry-java side. i just had one thought but not sure how feasible it is.

sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java
```diff
@@ -841,12 +879,18 @@ private static double readDouble(
...
       final long defaultValue) {
-    // manifest meta-data only reads int if the value is not big enough
-    final long value = metadata.getInt(key, (int) defaultValue);
+    final long value;
+    if (metadata instanceof Bundle) {
+      // manifest meta-data only reads int if the value is not big enough
+      value = ((Bundle) metadata).getInt(key, (int) defaultValue);
+    } else {
+      final Object raw = ((Map<?, ?>) metadata).get(key);
+      value = raw instanceof Integer ? (Integer) raw : defaultValue;
```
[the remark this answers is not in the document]
romtsn: we don't inject longs from the gradle plugin either

PR #5978, opened 2026-08-19
feat(replay): Add manual replay control API

romtsn:
Depends on #5965.

## :scroll: Description

Add `Sentry.replay()` controls for starting full-session or buffered recording, stopping, pausing, resuming, and flushing.

Lifecycle-driven background pauses remain separate from explicit user pauses. Foregrounding resumes only lifecycle-paused recording; an explicit pause remains active across automatic replay restarts in the same process until `resume()` is called. Starting an already-running replay is a no-op; flushing a stopped replay starts a full session, matching the JavaScript SDK. Explicit starts bypass replay sample rates, which continue to control automatic startup.

Give Android applications explicit control over Session Replay for opt-in flows and sensitive screens while preserving automatic lifecycle behavior.

Refs JAVA-325

- Ran all `sentry-android-core` unit tests: 1,601 tests, 0 failures, 1 skipped.
- Ran all `sentry-android-replay` unit tests successfully.
- Added focused coverage for the public API, manual pause behavior, buffering, flushing, and lifecycle transitions.

## :pencil: Checklist

## :crystal_ball: Next steps

- Merge the Android documentation in getsentry/sentry-docs#19077.


sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
```diff
@@ -334,6 +346,17 @@ public class ReplayIntegration(
 
   override fun getReplayId(): SentryId = state.get().replayId
 
+  override fun flush() {
```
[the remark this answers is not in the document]
romtsn: that is by design - when errorSampleRate is 0 we shouldn't force-upload manually started replay, the customers should call `flush()` (this is also aligned with JS)

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
```diff
@@ -167,10 +159,38 @@ public class ReplayIntegration(
...
   }
 
-  private fun startInternal() {
+  override fun startBuffering() {
+    enqueueOnMainThread { startInternal(isFullSession = false) }
+  }
+
+  override fun onAppForegrounded(startNewSession: Boolean) {
+    enqueueOnMainThread {
+      if (!isEnabled.get()) {
```
runningcode, line 172: would it make more sense to check this condition before enqueing?

runningcode, line 172: would it make more sense to check this condition before enqueing?

sentry/src/main/java/io/sentry/IReplayApi.java
```diff
@@ -1,7 +1,39 @@
 package io.sentry;
 
+/**
+ * Controls Session Replay. Methods may be called from any thread and return before the requested
+ * operation completes.
+ */
 public interface IReplayApi {
 
+  /** Starts a new replay session. Does nothing if a replay is already being recorded. */
```
runningcode, line 9: just for my understanding why is `start` and `startBuffering` two different calls. Why do we want to start a session separately from Buffering?

sentry/src/main/java/io/sentry/IReplayApi.java
```diff
@@ -1,7 +1,39 @@
...
+  void start();
+
+  /**
+   * Starts replay buffering. The rolling buffer is sent when {@link #flush()} is called or an error
+   * is captured. After the buffer is sent, recording continues in session mode unless the process
+   * is terminating.
+   */
+  void startBuffering();
+
+  /** Stops the current replay. A subsequent {@link #start()} begins a new replay session. */
```
runningcode: does this mean the session is stopped or the buffering is stopped or both?

I ask because we have two different start methods but only a single `stop/pause` which makes it not clear which one the stop stops

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
```diff
@@ -334,6 +346,17 @@ public class ReplayIntegration(
 
   override fun getReplayId(): SentryId = state.get().replayId
 
+  override fun flush() {
+    enqueueOnMainThread {
+      val current = state.get()
+      if (!current.isRecording) {
+        startInternal(isFullSession = true)
```
runningcode, line 368: why does `flush` call `startInternal` if we're not recording?

runningcode (commented):
I did a first pass quick review, ping me for a second round.

sentry/src/main/java/io/sentry/IReplayApi.java
romtsn: we give customers an option to choose which mode they want to start recording in. This could've been a method argument, but we're mimicking the [JS API](https://docs.sentry.io/platforms/javascript/guides/react/session-replay/understanding-sessions/#manually-starting-replay) here to stay aligned.

romtsn: session mode would continuously send replay segments while in buffer mode customers would have to explicitly call `flush()` or raise an error to have anything sent

sentry/src/main/java/io/sentry/IReplayApi.java
romtsn: both, the `stop` logic does not differ between the two (also updated javadoc)

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
romtsn: Also following the [JS API](https://docs.sentry.io/platforms/javascript/guides/react/session-replay/understanding-sessions/#manually-flushing-recording-data) :) I agree it can be a real footgun though - there's one issue open in sentry-javascript re. that https://github.com/getsentry/sentry-javascript/issues/12664. I think I'd still not introduce `recordingMode` as they suggest in the issue above for now, until someone actually hits this (we document the behaviour). Or we could also diverge from JS here and not start a new recording in `flush` at all.

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
```diff
@@ -167,10 +159,38 @@ public class ReplayIntegration(
...
+        if (!isFullSession && !options.sessionReplay.isSessionReplayForErrorsEnabled) {
+          options.logger.log(
+            INFO,
+            "Session replay is not started, full session was not sampled and onErrorSampleRate is not specified",
+          )
+        } else {
+          startInternal(isFullSession)
+        }
+      }
+      resumeInternal()
```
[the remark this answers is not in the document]
romtsn: STARTED already means actively recording: startInternal() calls recorder.start() and strategy.start(). RESUMED is specifically for PAUSED → RESUMED, so forbidding STARTED → RESUMED is intentional.

  The trailing resumeInternal() handles an existing paused replay and correctly no-ops for a newly started one. Existing recorder lifecycle tests confirm this.

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
romtsn: apparently it can break a subtle edge-case https://github.com/getsentry/sentry-java/pull/5978#discussion_r3842227572 so i reverted it to be safe

sentry/src/main/java/io/sentry/IReplayApi.java
runningcode: Can we document that `flush()` sends it up to Sentry? I think it isn't so clear (it could just be flushing to disk).

runningcode (commented):
looks good. I still think the start and startBuffering methods are confusing. Not sure if better javadocs would be less confusing. I understand that it is that way to align with js.

runningcode (approved):
lgtm!

sentry/src/main/java/io/sentry/IReplayApi.java
romtsn: documented!

sentry-android-replay/src/main/java/io/sentry/android/replay/capture/SessionCaptureStrategy.kt
```diff
@@ -91,6 +92,17 @@ internal class SessionCaptureStrategy(
...
+  override fun flush(onSegmentSent: (Date) -> Unit) {
+    createCurrentSegment("flush") { segment ->
+      if (segment is ReplaySegment.Created) {
+        segment.capture(scopes)
+        currentSegment++
+        segmentTimestamp = segment.replay.timestamp
+        isFlushed = true
+      }
+    }
+  }
```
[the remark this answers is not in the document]
romtsn: This one is a false positive. Session flush and stop work are FIFO on the same replay executor. If stop wins on the main thread, its segment teardown and cache cleanup are queued behind the earlier flush; the old cache is then closed/deleted, so this state cannot survive cleanup. Moving these mutations into the main-thread callback would instead let later replay-executor frame tasks observe the old segment ID/timestamp and produce duplicate segments. The timeline therefore stays serialized on the replay executor.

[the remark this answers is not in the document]
romtsn: Good catch — fixed in de61eebb6e and refined in ff3b7aeea2. The mutable segment start and ID are read when the task runs on the replay executor, after any earlier natural boundary. The request cutoff time and replay ID remain invocation snapshots, so executor backlog does not extend the segment. Regression tests cover both queued boundaries and delayed execution.

sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt
```diff
@@ -167,10 +160,39 @@ public class ReplayIntegration(
...
+    enqueueOnMainThread { startInternal(isFullSession = false) }
+  }
+
+  override fun onAppForegrounded(startNewSession: Boolean) {
+    enqueueOnMainThread {
+      if (!isEnabled.get()) {
+        return@enqueueOnMainThread
+      }
+      if (startNewSession) {
+        stopInternal()
```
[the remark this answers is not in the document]
romtsn: You're right — changed in 59505b3647. A manual pause is now preserved when automatic app-session rollover replaces the replay in the same process. The replacement strategy is paused before the main looper can capture a frame and remains paused until resume() is called. The API docs, changelog, Android docs, and regression test now state this behavior.

sentry-android-replay/src/main/java/io/sentry/android/replay/capture/SessionCaptureStrategy.kt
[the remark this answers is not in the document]
romtsn: `ReplayIntegration` holds the same strategy object, not copies of its timeline fields. The callback is only needed when buffer mode converts to a differentstrategy.

sentry-android-replay/src/main/java/io/sentry/android/replay/capture/SessionCaptureStrategy.kt
```diff
@@ -179,15 +191,16 @@ internal class SessionCaptureStrategy(
...
       ReplayRunnable("$TAG.$taskName") {
+        // Keep the request time so executor delays do not extend the segment, but read the mutable
+        // timeline here so a queued natural boundary cannot make it stale.
+        val currentSegmentTimestamp = segmentTimestamp ?: return@ReplayRunnable
         val segment =
           createSegmentInternal(
-            duration,
+            requestedAt - currentSegmentTimestamp.time,
             currentSegmentTimestamp,
             replayId,
```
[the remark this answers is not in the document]
romtsn: Fixed in 69a9fb0f5e. Natural segment boundaries and the replay deadline now use the frame timestamp captured before enqueueing, so executor backlog no longer counts as recorded time or advances the timeline ahead of queued request-time operations. The regression test delays executor execution past both thresholds and verifies neither is triggered.

sentry-android-replay/src/main/java/io/sentry/android/replay/capture/SessionCaptureStrategy.kt
```diff
@@ -179,17 +191,17 @@ internal class SessionCaptureStrategy(
...
+        // cannot make this snapshot stale.
+        val now = dateProvider.currentTimeMillis
+        val currentSegmentTimestamp = segmentTimestamp ?: return@ReplayRunnable
         val segment =
           createSegmentInternal(
-            duration,
+            now - currentSegmentTimestamp.time,
             currentSegmentTimestamp,
-            replayId,
+            currentReplayId,
```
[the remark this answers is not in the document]
romtsn: Fixed in 69a9fb0f5e. resume() now captures the resume time on the caller thread but applies the timeline update on the replay executor, behind any queued pause, flush, or frame work. A regression test verifies that a queued pause still captures the pre-resume segment.

sentry-android-replay/src/main/java/io/sentry/android/replay/capture/SessionCaptureStrategy.kt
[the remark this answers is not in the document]
romtsn: This is intentional. ReplayIntegration holds the same SessionCaptureStrategy instance, so it observes currentSegment, segmentTimestamp, and isFlushed directly. The callback is needed for buffer-to-session conversion because that swaps strategy instances; invoking it here would apply the same session update twice. Both natural boundaries and flush mutate this strategy on the replay executor, which preserves their ordering.

PR #5979, opened 2026-08-21
fix(android): Handle default-package ANR frame modules (JAVA-696)

0xadam-brown:
## :scroll: Description

Updates `ThreadDumpParser` so ANR frames for classes in Java's default package are emitted with the correct module name.

Prior to this PR, when the parser saw a Java or JNI frame whose class has no package, it would reconstruct the frame's module as though a package existed. That turned a class like `MainActivity` into `null.MainActivity`. 

This PR fixes that by teaching `ThreadDumpParser` to build the module name conditionally:

- if a package name is present, emit `package.class`
- if no package name is present, emit just `class`

The same logic is applied consistently to both Java frames and JNI frames, since both parsing paths share the same optional-package shape.

Absent this PR, we'd see:

- an invalid synthetic module name for default-package classes;
- distored downstream in-app classification, since that logic depends on the module name; and
- internally inconsistent parsed ANR frames, where the function, file, and line point at one class but the module name points at another, non-existent one.

resolves: JAVA-696

TDD-style unit tests.

## :pencil: Checklist


romtsn (approved):
nice!

PR #5989, opened 2026-08-25
fix(core): Make feature flag buffer merging thread-safe

runningcode:
## :scroll: Description

When we merge the feature flags in merge(), we iterate through the lists without copying them or obtaining a lock.
The add() method obtains a lock, and does the add by first removing an existing flag (if exists) then adding it again and then doing another remove if the list is too long.

That means that if we call merge and then another thread calls add, there's no guarantee that anything will work or that it won't throw an exception.

The fix here is to use immutable lists and swap them under the lock. This makes reading the list produce an immutable snapshot.

Other improvements because of that change and also in this PR:
- `clone()` is now free — it shares the list rather than copying it. Since the class doc calls scope cloning the optimized path, this is the one that matters.
- `add()` drops from three full array copies to one.
- `CopyOnWriteArrayList` disappears from the class; `clear()` becomes an empty-list swap.
- `maxSize` becomes `final`, which it always should have been — as written, a thread obtaining a buffer through a data race could legally observe `maxSize == 0`.
- `FeatureFlagEntry.nanos` becomes a primitive `long` instead of a boxed `@NotNull Long`, and its stale `@SuppressWarnings("UnusedVariable")` is gone.
- Merge tie-breaks now favour the most specific scope (CURRENT > ISOLATION > GLOBAL) instead of the reverse. Previously strict `>` meant whichever buffer was tested first won equal timestamps.

`getFeatureFlags()` is left annotated `@Nullable` even though it never returns null, to stay consistent with the `IFeatureFlagBuffer` signature.

* resolves: JAVA-704

Added a test that merges in a loop while another thread adds flags. Against the old implementation it fails with `ArrayIndexOutOfBoundsException: Index 100 out of bounds for length 100`; against the new one it passes. Also added a test pinning `clone()` independence, since clone changed from copying the list to sharing it. Full `:sentry:test` suite passes and `apiDump` produces no changes.

## :pencil: Checklist

## :crystal_ball: Next steps

JAVA-704 also notes an optional follow-up: replacing `System.nanoTime()` with a static `AtomicLong` sequence, which would give a total order across buffers with no ties at all and no dependence on clock granularity. Left out here to keep the change small.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixes
 
 - Prevent duplicated breadcrumbs on tombstone-merged native crash events ([#5888](https://github.com/getsentry/sentry-java/pull/5888))
+- Prevent events from being dropped when feature flags are added while an event is being captured ([#5989](https://github.com/getsentry/sentry-java/pull/5989))
 - Prevent a class of Session Replay deadlocks by confining lifecycle state changes to Android's main thread ([#5965](https://github.com/getsentry/sentry-java/pull/5965))
 
 ### Performance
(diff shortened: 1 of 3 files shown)


markushi (approved):
Nice, LGTM!

adinauer (approved):
LGTM!
Wrote down benchmark results and reasoning in https://github.com/getsentry/sentry-java/pull/5994

PR #5993, opened 2026-08-25
feat(profiling): drop profiler context from transactions when profiler gets rate-limited

markushi:
## :scroll: Description

Transactions are tagged with the continuous profiler's `profiler_id` as soon as they start, but the OS only tells us whether a Perfetto profile actually exists later.

This PR adds a callback path so anything already tagged with a dead id can drop it before being sent.

Rate limiting is the common case on API 35+, and it is reported roughly 1 ms after the request. Without this, a rate-limited session produces transactions that link to profiles the backend never receives, which surfaces in the UI as dangling profile references.

Unfortunately we do not have any backend logic in place to remove dangling references.

Added unit tests.

## :pencil: Checklist

## :crystal_ball: Next steps

Known gaps left deliberately, happy to fold any of them in:

- Cancellation is scoped to the profiler id, not chunk_id. Since the id survives restarts, one failed chunk also invalidates transactions covered by an earlier chunk that was sent. This only affects unsent transactions/spans where the profile would not cover it's execution anyway.
- A cancellation landing between the tracer reading the profiler id and registering itself is not observed. The window is a few instructions and both other interleavings are handled by the chunk teardown.
- `finish()` unregisters unconditionally, and tracers that are never finished stay registered until `close(true)`.


sentry/src/main/java/io/sentry/Scopes.java
```diff
@@ -992,9 +992,6 @@ && getOptions().getProfileLifecycle() == ProfileLifecycle.TRACE
       transaction =
           spanFactory.createTransaction(
               transactionContext, this, transactionOptions, compositePerformanceCollector);
-      //          new SentryTracer(
```
runningcode, line 995: is this intentional?

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -89,7 +94,8 @@ public SentryTracer(
     final @NotNull SentryId continuousProfilerId = getProfilerId();
 
     if (!continuousProfilerId.equals(SentryId.EMPTY_ID) && Boolean.TRUE.equals(isSampled())) {
-      this.contexts.setProfile(new ProfileContext(continuousProfilerId));
+      contexts.setProfile(new ProfileContext(continuousProfilerId));
+      scopes.getOptions().getContinuousProfiler().registerProfilingCanceledCallback(this);
```
runningcode, line 98: `registerProfilingCanceledCallback` requires obtaining a lock which means this could stall the constructor. The constructor seems to be called from the main thread so this seems to have the potential to cause ANRs.

sentry/src/test/java/io/sentry/profiling/ProfilingServiceLoaderTest.kt
```diff
@@ -63,6 +64,14 @@ class ContinuousProfilerStub() : IContinuousProfiler {
     TODO("Not yet implemented")
   }
 
+  override fun registerProfilingCanceledCallback(callback: IProfilingCanceledCallback) {
+    TODO("Not yet implemented")
```
runningcode, line 68: I would use a different exception for a stub. this implies it will be implemented in the future.

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
```diff
@@ -162,6 +167,60 @@ public void stopProfiler(final @NotNull ProfileLifecycle profileLifecycle) {
...
+      if (isRunning) {
+        stopInternal(false);
+      }
+      // Iterated from a copy, as a listener may unregister itself while being notified.
+      callbacks = new ArrayList<>(profilingCanceledCallbacks);
+    }
+    for (final @NotNull IProfilingCanceledCallback callback : callbacks) {
+      try {
+        callback.onProfilingCanceled(canceledProfilerId);
+      } catch (Throwable t) {
```
runningcode, line 215: My main concern here is how would we know if this new feature is working or not working correctly? How would we get alerted in the future if a change causes this to break?

I'm not sure I understand the comment about this being an OS binder thread. My understanding is that an uncaught exception always ends the process. The `onProfilingCanceled` is our own callback so we should understand and control the code there.

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
```diff
@@ -162,6 +167,60 @@ public void stopProfiler(final @NotNull ProfileLifecycle profileLifecycle) {
...
+
+    final @NotNull List<IProfilingCanceledCallback> callbacks;
+    try (final @NotNull ISentryLifecycleToken ignored = lock.acquire()) {
+      // The OS can report the failure a few ms into a chunk that would otherwise run for another
+      // minute. Without tearing it down here, transactions started in the meantime would read the
+      // id that was just invalidated, and register too late to ever be told about it.
+      if (isRunning) {
+        stopInternal(false);
+      }
+      // Iterated from a copy, as a listener may unregister itself while being notified.
```
runningcode, line 209: how can they unregister if they require a lock to unregister?

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -261,6 +267,8 @@ public void finish(
                   }
                 });
           });
+      scopes.getOptions().getContinuousProfiler().unregisterProfilingCanceledCallback(this);
```
runningcode, line 270: This might not be called if the condition on line 211 isn't true `root.isFinished()
        && (!transactionOptions.isWaitForChildren() || hasAllChildrenFinished()`. would that cause a leak?

runningcode (commented):
gave it a first pass! looking good!

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
```diff
@@ -162,6 +167,60 @@ public void stopProfiler(final @NotNull ProfileLifecycle profileLifecycle) {
...
+
+  @Override
+  public void unregisterProfilingCanceledCallback(@NotNull IProfilingCanceledCallback callback) {
+    try (final @NotNull ISentryLifecycleToken ignored = lock.acquire()) {
+      this.profilingCanceledCallbacks.remove(callback);
+    }
+  }
+
+  /**
+   * Invoked once it is known that no profile chunk will be produced for the given profiler id, so
```
0xadam-brown, line 185: l: "no profile chunk" -> "no additional profile chunks"

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
```diff
@@ -162,6 +167,60 @@ public void stopProfiler(final @NotNull ProfileLifecycle profileLifecycle) {
...
+   * that anything already tagged with it can drop the reference before being sent.
+   *
+   * <p>The id outlives a single chunk, so this may fire more than once for the same id, and it also
```
0xadam-brown, line 188: l: Fwiw, seems like most of this comment would be better suited for `SentryTracer.onProfilingCanceled()` (but up to you). 

Also might be worth noting that transactions are bound to profiling sessions via profiler_id, and a single profiling session can produce 0 to N profile chunks, as that's what leads to the "Losing a valid profile link..." tradeoff you (helpfully!) mention.

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
```diff
@@ -162,6 +167,60 @@ public void stopProfiler(final @NotNull ProfileLifecycle profileLifecycle) {
...
+    logger.log(
+        SentryLevel.DEBUG,
+        "No profile chunk will be produced for profiler id %s, dropping it.",
+        canceledProfilerId);
+
+    final @NotNull List<IProfilingCanceledCallback> callbacks;
+    try (final @NotNull ISentryLifecycleToken ignored = lock.acquire()) {
+      // The OS can report the failure a few ms into a chunk that would otherwise run for another
+      // minute. Without tearing it down here, transactions started in the meantime would read the
+      // id that was just invalidated, and register too late to ever be told about it.
```
0xadam-brown, line 205: l: My vote would be to implement (and comment) PerfettoContinuousProfiler as if it has a policy that simply cancels the current profiling session whenever we receive a rate-limited response from the ProfilingManager. We could then include discussions about how that policy interacts with transactions higher up our stack, eg, in SentryTracer. 

That'd^^ let us be more explicit about the contract in the profiler, while keeping comments about how that contract interacts with transactions in the abstraction that naturally has to care about both.

(Happy to defer to you, of course...)

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoProfiler.java
```diff
@@ -104,6 +109,41 @@ public boolean start(final long durationMs) {
     return true;
   }
 
+  /**
+   * Sets the callback invoked once it is known that this session will not produce a trace file,
```
0xadam-brown, line 113: l: Consider: "this session" -> "the current request"

(although that's^^ assuming our docs use the term "session" the same way my comments have been, viz., a sequence of 0 to N requests made by PerfettoContinuousProfiler that are associated with the same profiler_id. Maybe we have a separate concept of a ProfilingManager session that exists for a single .pftrace request?? If so, nvm.)

sentry/src/main/java/io/sentry/IProfilingCanceledCallback.java
```diff
@@ -0,0 +1,18 @@
...
+
+import io.sentry.protocol.SentryId;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Notified when the profiler learns that no profile will ever exist for a profiler id, so that
+ * anything already tagged with that id can drop the reference before being sent.
+ *
+ * <p>Implementations are invoked on whichever thread learns about the failure: an OS binder thread,
```
0xadam-brown, line 11: 🙏

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -22,7 +22,7 @@
 import org.jetbrains.annotations.TestOnly;
 
 @ApiStatus.Internal
-public final class SentryTracer implements ITransaction {
+public final class SentryTracer implements ITransaction, IProfilingCanceledCallback {
```
0xadam-brown, line 25: m: Looks like we have two options:

1. update SentryTracer in real time as the current profile session is updated (eg., via a callback); or
2. have SentryTracer repair its provider_id and the provider_id bound to any of its children when someone calls SentryTracer.finish().

Off the bat, I'm inclined toward (2) because it lets avoid having to copy profiler state here (eg, `profilingCanceled`) and it makes for a more change-resistant IContinuousProfiler API (ie, a single IContinuousProfiler.getSessionData()-esque method that whose return type can be built out over time, rather than callbacks that we'll likely need to add to and document how they relate to each other). SentryTracer could then grab that session data, see whether it's status is CANCELED (etc.), and take appropriate action.

Thoughts? (And correct me if I'm being too naive!)

sentry/src/main/java/io/sentry/SentryTracer.java
```diff
@@ -1021,6 +1031,31 @@ public void addFeatureFlag(final @Nullable String flag, final @Nullable Boolean
     this.root.addFeatureFlag(flag, result);
   }
 
+  @Override
+  public void onProfilingCanceled(final @NotNull SentryId profilerId) {
```
0xadam-brown, line 1035: m: Thoughts about also passing info describing whether any ProfileChunks have been successfully produced by the current session + their start and end times? (in raw form or – probably preferably – in the form of a `SessionData`-type object that lets us query against it)

That'd^^ let us avoid declaring an entire profiling session invalid in cases where one or more ProfileChunks were successfully emitted while the current transaction was alive. And it'd have the nice property of allowing us to solve the more general problem of what we should do for transactions bound to profile sessions that don't produce any profiling data. (Solving the more general problem would mean taking action like we do here both when a session ends exceptionally – eg, rate-limiting – and normally – eg, someone manually calls stop() or no active transaction exists.)

Doesn't have to block this PR, but worth considering...


### Background

<summary> Click below for helpful background on my comment (and to double-check my reasoning!)
<details>


**[I]** ***Continuous profiling happens in sessions, which consist of 0 to N requests for Perfetto .pftrace files from the Android OS***

Our PerfettoContinuousProfiler implements a profiling session model, where a session:

- is bound to a single `profiler_id`; 
- consists of one or more sequential calls to Android's ProfilingManager, where each call either generates a .pftrace + ProfileChunk pair, or no-ops if we've been rate-limited by the OS.

Each .pftrace + ProfileChunk covers a 60 second interval.

The profiling session can end in one of two ways: 

1. naturally (ie, when someone manually calls stop() or no active transaction is present, depending on the profiling mode used); or
2. when we're rate-limited by the Android OS.

So long as the session hasn't ended, PerfettoContinuousProfiler automatically calls Android's ProfilingManager again after each .pftrace + ProfileChunk pair is processed.

**[II]** ***Creating a transaction eagerly binds it to the currently active profiling session via the latter's profiler_id***

...and if no profiling session is active, creating a transaction will start one. Each transaction is bound to the current profiling session before we know whether that session will actually produce profiling data (ie, .pftrace + ProfileChunk pairs). 

Sentry's backend makes sure that only ProfileChunks i) associated with a transaction's bound profiler_id and ii) generated during a transaction's lifetime are displayed in the Sentry UI for that transaction (i.e., the backend filters out ProfileChunks generated by the profiling session that didn't occur during the transaction's lifetime). 

**[III]** ***So what do we do if a transaction is bound to a profiling session that never produces any profiling data?***

That, of course, is the question this PR addresses. But as @markushi points out in the PR description, it only addresses a reduced version, viz., what do we do if the Android OS rate-limits us partway through a profiling session? 

Three options: 

1. deem the entire profiling session invalid, including any previously shipped ProfileChunks (the approach taken by this PR);
2. deem any prior work done during the session valid, but invalidate the session going forward (what I propose above); or
3. do nothing (ie, keep the status quo and the bad Sentry UI experience it produces).
</details>
</summary>

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
 - Prevent duplicated breadcrumbs on tombstone-merged native crash events ([#5888](https://github.com/getsentry/sentry-java/pull/5888))
 - Prevent a class of Session Replay deadlocks by confining lifecycle state changes to Android's main thread ([#5965](https://github.com/getsentry/sentry-java/pull/5965))
 - Symbolicate tombstone native frames for libraries loaded directly from APKs ([#5992](https://github.com/getsentry/sentry-java/pull/5992))
+- Drop the `profiler_id` from in-flight transactions when Android's `ProfilingManager` rejects profiling requests ([#5993](https://github.com/getsentry/sentry-java/pull/5993))
+  - Previously a rate-limited or failed Perfetto profiling request still left a `profiler_id` on transactions, pointing to non-existent profiles
```
0xadam-brown, line 12: m: Should we warn folks that we've added non-default methods to IContinuousProfiler, given that it can be vended by an SPI implemented in a different JAR ([link](https://github.com/getsentry/sentry-java/blob/27e07a50655360d837728b67bc2804a808883a4b/sentry/src/main/java/io/sentry/profiling/JavaContinuousProfilerProvider.java#L16))? (I get that the interface is `@Internal`, but it's a bit unusual given the runtime discovery.)

0xadam-brown (commented):
Looking great, and thanks for this 💯 

I had two bigger picture comments for your consideration ([here](https://github.com/getsentry/sentry-java/pull/5993#discussion_r3862355513) and [here](https://github.com/getsentry/sentry-java/pull/5993#discussion_r3862364284)). Up to you whether to incorporate those ideas now, later, or never. 

I also included a few doc comments – but only because I bothered to deep dive into the implementation of our profiling and figured capturing some of that knowledge could be helpful. They're all optional, of course.

Note that I haven't carefully reviewed issues around thread-safety or call sequence in case we end up changing the approach. Happy to take a second pass as needed 👍

markushi:
Closing in favor of #6015, as it seems a much more lightweight change and achieves the same goal. I'll try to transfer the learnings/feedback from here to the other PR, so not all is wasted.

PR #5994, opened 2026-08-25
fix(core): Snapshot feature flags before merging

### LINKED ISSUE #5988: FeatureFlagBuffer.merged() can throw and drop events under concurrency
`sentry/src/main/java/io/sentry/featureflags/FeatureFlagBuffer.java`

## Bug

`merged()` claims to snapshot the incoming buffers:

```java
// Capture references to avoid inconsistencies from concurrent modifications
final @Nullable CopyOnWriteArrayList<FeatureFlagEntry> globalFlags = globalBuffer.flags;
```

That comment isn't true. `flags` is assigned only in the three constructors and never reassigned, so capturing the reference yields the *same live list*, not a snapshot. `CopyOnWriteArrayList` makes each individual `size()`/`get(i)` atomic, but gives nothing across the ~`maxSize` calls the merge loop makes.

`add()` mutates in three separate steps — `remove(i)`, `add(...)`, `remove(0)` — so the list transiently shrinks. A merging thread that reads `size() == n` and then calls `get(n-1)` while a writer sits between its `remove` and its `add` gets an `ArrayIndexOutOfBoundsException`. Indexing a shifting list can also duplicate or skip entries.

This is reachable: the global-scope buffer is shared across all threads, and `merged()` runs on the error-capture path (`SentryClient#applyScope` → `CombinedScopeView#getFeatureFlagBuffer`). `Scopes.captureEventInternal` catches `Throwable`, so it won't crash the host app — but **the event is silently dropped**, including events from the uncaught-exception handler.

`clone()` and `getFeatureFlags()` are fine: the `CopyOnWriteArrayList(Collection)` constructor and the COW iterator both take a real atomic snapshot. Only `merged()` indexes.

## Related cleanup in the same class

* `volatile` on `flags` is dead weight — the field is never reassigned. It costs a fence per read and, worse, signals a snapshot-swap design that isn't there (likely how the misleading comment above arose).
* `maxSize` is neither `final` nor `volatile` despite only being set in constructors. As written, a thread obtaining a `FeatureFlagBuffer` through a data race can legally observe `maxSize == 0`. `final` gives the freeze guarantee for free — same for `flags`.
* `CopyOnWriteArrayList` + `lock` is redundant: every mutation already holds the lock, so COW's write-side atomicity is never needed. We pay for it three times per `add()` — with the default `maxFeatureFlags = 100`, that's three 100-element array copies per flag evaluation.
* `FeatureFlagEntry.nanos` is a boxed `@NotNull Long` — an extra allocation per entry, unboxed twice per merge-loop comparison. Should be a primitive `long`.
* `@SuppressWarnings("UnusedVariable")` on `nanos` is stale; `merged()` reads it.
* `getFeatureFlags()` is annotated `@Nullable` but can never return null.
* Merge tie-break favors GLOBAL > ISOLATION > CURRENT (strict `>`, so whichever is tested first wins equal timestamps). The more specific scope should win; changing isolation's and current's comparison to `>=` fixes it.

## Proposed fix

Make the snapshot real: an immutable list behind a volatile field, swapped under the existing lock — the design the current code already gestures at.

```java
private volatile @NotNull List<FeatureFlagEntry> flags = Collections.emptyList();
private final int maxSize;

@Override
public void add(final @Nullable String flag, final @Nullable Boolean result) {
  if (flag == null || result == null) {
    return;
  }
  try (final @NotNull ISentryLifecycleToken ignored = lock.acquire()) {
    final @NotNull List<FeatureFlagEntry> updated = new ArrayList<>(flags.size() + 1);
    for (final @NotNull FeatureFlagEntry entry : flags) {
      if (!entry.flag.equals(flag)) {
        updated.add(entry);
      }
    }
    updated.add(new FeatureFlagEntry(flag, result, System.nanoTime()));
    if (updated.size() > maxSize) {
      updated.remove(0);
    }
    flags = Collections.unmodifiableList(updated);
  }
}
```

What it buys:

* `merged()` becomes correct — one volatile read per buffer yields a list that can never change, so the index arithmetic is trivially safe and the existing comment becomes accurate.
* `clone()` gets cheaper than today: `new FeatureFlagBuffer(other.maxSize, other.flags)` shares the immutable list with zero copying. Since the class doc calls clone the optimized path, this is the one that matters.
* `add()` drops from three array copies to one.
* `clear()` becomes `flags = Collections.emptyList()`; `CopyOnWriteArrayList` disappears from the class entirely.

Invariant to preserve: correctness now rests on every *write* holding the lock (they are read-modify-write), while reads need no lock at all.

Add a test that hammers `add()` from one thread while another calls `merged()`.

## Notes

* Left `AutoClosableReentrantLock` alone — it is the repo idiom.
* Deliberately not swapping the list for a `LinkedHashMap`: that would make `add()` O(1) but force `clone()` to copy, trading away the property this class exists to optimize.
* Optional: replace `System.nanoTime()` with a static `AtomicLong` sequence to get a total order across buffers with no ties and no dependence on clock granularity.

adinauer:
## :scroll: Description

Snapshot each scope's `FeatureFlagBuffer` before merging feature flags into an event. This keeps the existing mutable `CopyOnWriteArrayList` write path while ensuring that the merge traverses structurally stable lists.

Fixes https://github.com/getsentry/sentry-java/issues/5988

`FeatureFlagBuffer` keeps a mutable `CopyOnWriteArrayList` wrapper. Writers update that wrapper through separate remove, add, and eviction operations, while `merged()` previously captured only the wrapper reference and then used separate `size()` and indexed `get()` calls.

A `CopyOnWriteArrayList` makes each individual operation thread-safe, but retaining its wrapper does not retain one backing-array snapshot. A concurrent writer can therefore produce this interleaving:

1. A writer adds a flag and the list reaches 101 entries.
2. The merging thread reads a size of 101.
3. The writer evicts the oldest flag and the list returns to 100 entries.
4. The merging thread calls `get(100)` and throws `ArrayIndexOutOfBoundsException` during event capture.

The same structural race is possible when another thread clears a buffer. Constructing an independent COW wrapper for each input buffer gives the merge a stable size and stable indexed reads without acquiring the global, isolation, and current buffer locks together. This avoids lock-ordering risk and preserves the benchmark-favored write path.

Each per-buffer snapshot is taken under that buffer's writer lock, so it represents a completed add or clear operation. Global, isolation, and current snapshots are captured sequentially rather than as one atomic cross-buffer transaction, preserving best-effort cross-scope semantics.

Refs JAVA-704

- Added concurrent regression tests that pre-fill the buffer, verify writer progress during add/evict churn, and confirm merges and direct reads never observe partial updates.
- Confirmed the regression test fails on `main` with `ArrayIndexOutOfBoundsException` and passes with this change.
- Ran `./gradlew ':sentry:test' --tests='*FeatureFlagBufferTest*' --info` (26 tests, 0 failures).
- Ran `./gradlew spotlessApply apiDump` successfully.

## :pencil: Checklist

## :crystal_ball: Next steps

None.

### DIFF
--- a/sentry/src/main/java/io/sentry/featureflags/FeatureFlagBuffer.java
+++ b/sentry/src/main/java/io/sentry/featureflags/FeatureFlagBuffer.java
@@ -135,7 +135,9 @@ public final class FeatureFlagBuffer implements IFeatureFlagBuffer {
       final @Nullable FeatureFlagBuffer isolationBuffer,
       final @Nullable FeatureFlagBuffer currentBuffer) {
 
-    // Capture snapshots to avoid inconsistencies from concurrent modifications
+    // Capture structurally stable snapshots before indexed traversal. Passing a
+    // CopyOnWriteArrayList directly allows its collection constructor to reuse the immutable
+    // backing array instead of copying the elements on runtimes that support this optimization.
     final @Nullable CopyOnWriteArrayList<FeatureFlagEntry> globalFlags =
         globalBuffer == null ? null : new CopyOnWriteArrayList<>(globalBuffer.flags);
     final @Nullable CopyOnWriteArrayList<FeatureFlagEntry> isolationFlags =
--- a/sentry/src/main/java/io/sentry/featureflags/FeatureFlagBuffer.java
+++ b/sentry/src/main/java/io/sentry/featureflags/FeatureFlagBuffer.java
@@ -44,7 +44,14 @@ public final class FeatureFlagBuffer implements IFeatureFlagBuffer {
 
   private FeatureFlagBuffer(@NotNull FeatureFlagBuffer other) {
     this.maxSize = other.maxSize;
-    this.flags = new CopyOnWriteArrayList<>(other.flags);
+    this.flags = other.snapshot();
+  }
+
+  @ApiStatus.Internal
+  private @NotNull CopyOnWriteArrayList<FeatureFlagEntry> snapshot() {
+    try (final @NotNull ISentryLifecycleToken ignored = lock.acquire()) {
+      return new CopyOnWriteArrayList<>(flags);
+    }
   }
 
   @Override
@@ -139,11 +146,11 @@ public final class FeatureFlagBuffer implements IFeatureFlagBuffer {
     // CopyOnWriteArrayList directly allows its collection constructor to reuse the immutable
     // backing array instead of copying the elements on runtimes that support this optimization.
     final @Nullable CopyOnWriteArrayList<FeatureFlagEntry> globalFlags =
-        globalBuffer == null ? null : new CopyOnWriteArrayList<>(globalBuffer.flags);
+        globalBuffer == null ? null : globalBuffer.snapshot();
     final @Nullable CopyOnWriteArrayList<FeatureFlagEntry> isolationFlags =
-        isolationBuffer == null ? null : new CopyOnWriteArrayList<>(isolationBuffer.flags);
+        isolationBuffer == null ? null : isolationBuffer.snapshot();
     final @Nullable CopyOnWriteArrayList<FeatureFlagEntry> currentFlags =
-        currentBuffer == null ? null : new CopyOnWriteArrayList<>(currentBuffer.flags);
+        currentBuffer == null ? null : currentBuffer.snapshot();
 
     final int globalSize = globalFlags == null ? 0 : globalFlags.size();
     final int isolationSize = isolationFlags == null ? 0 : isolationFlags.size();
--- a/sentry/src/main/java/io/sentry/featureflags/FeatureFlagBuffer.java
+++ b/sentry/src/main/java/io/sentry/featureflags/FeatureFlagBuffer.java
@@ -8,6 +8,7 @@ import io.sentry.protocol.FeatureFlags;
 import io.sentry.util.AutoClosableReentrantLock;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Iterator;
 import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
 import org.jetbrains.annotations.ApiStatus;
@@ -85,9 +86,16 @@ public final class FeatureFlagBuffer implements IFeatureFlagBuffer {
 
   @Override
   public @Nullable FeatureFlags getFeatureFlags() {
-    List<FeatureFlag> featureFlags = new ArrayList<>();
-    for (FeatureFlagEntry entry : flags) {
-      featureFlags.add(entry.toFeatureFlag());
+    final int size;
+    final @NotNull Iterator<FeatureFlagEntry> snapshot;
+    try (final @NotNull ISentryLifecycleToken ignored = lock.acquire()) {
+      size = flags.size();
+      snapshot = flags.iterator();
+    }
+
+    final @NotNull List<FeatureFlag> featureFlags = new ArrayList<>(size);
+    while (snapshot.hasNext()) {
+      featureFlags.add(snapshot.next().toFeatureFlag());
     }
     return new FeatureFlags(featureFlags);
   }
--- a/sentry/src/main/java/io/sentry/featureflags/FeatureFlagBuffer.java
+++ b/sentry/src/main/java/io/sentry/featureflags/FeatureFlagBuffer.java
@@ -135,13 +135,13 @@ public final class FeatureFlagBuffer implements IFeatureFlagBuffer {
       final @Nullable FeatureFlagBuffer isolationBuffer,
       final @Nullable FeatureFlagBuffer currentBuffer) {
 
-    // Capture references to avoid inconsistencies from concurrent modifications
+    // Capture snapshots to avoid inconsistencies from concurrent modifications
     final @Nullable CopyOnWriteArrayList<FeatureFlagEntry> globalFlags =
-        globalBuffer == null ? null : globalBuffer.flags;
+        globalBuffer == null ? null : new CopyOnWriteArrayList<>(globalBuffer.flags);
     final @Nullable CopyOnWriteArrayList<FeatureFlagEntry> isolationFlags =
-        isolationBuffer == null ? null : isolationBuffer.flags;
+        isolationBuffer == null ? null : new CopyOnWriteArrayList<>(isolationBuffer.flags);
     final @Nullable CopyOnWriteArrayList<FeatureFlagEntry> currentFlags =
-        currentBuffer == null ? null : currentBuffer.flags;
+        currentBuffer == null ? null : new CopyOnWriteArrayList<>(currentBuffer.flags);
 
     final int globalSize = globalFlags == null ? 0 : globalFlags.size();
     final int isolationSize = isolationFlags == null ? 0 : isolationFlags.size();
--- a/sentry/src/main/java/io/sentry/featureflags/FeatureFlagBuffer.java
+++ b/sentry/src/main/java/io/sentry/featureflags/FeatureFlagBuffer.java
@@ -27,7 +27,7 @@ import org.jetbrains.annotations.Nullable;
 @ApiStatus.Internal
 public final class FeatureFlagBuffer implements IFeatureFlagBuffer {
 
-  private volatile @NotNull CopyOnWriteArrayList<FeatureFlagEntry> flags;
+  private final @NotNull CopyOnWriteArrayList<FeatureFlagEntry> flags;
   private final @NotNull AutoClosableReentrantLock lock = new AutoClosableReentrantLock();
   private int maxSize;
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixes
 
+- Prevent crashes when feature flags are merged during concurrent scope updates ([#5994](https://github.com/getsentry/sentry-java/pull/5994))
 - Prevent duplicated breadcrumbs on tombstone-merged native crash events ([#5888](https://github.com/getsentry/sentry-java/pull/5888))
 - Prevent a class of Session Replay deadlocks by confining lifecycle state changes to Android's main thread ([#5965](https://github.com/getsentry/sentry-java/pull/5965))
 
--- a/sentry/src/test/java/io/sentry/featureflags/FeatureFlagBufferTest.kt
+++ b/sentry/src/test/java/io/sentry/featureflags/FeatureFlagBufferTest.kt
@@ -397,4 +397,47 @@ class FeatureFlagBufferTest {
 
     assertThat(writerFailure.get()).isNull()
   }
+
+  @Test
+  fun `merging does not observe a partially updated buffer`() {
+    val options = SentryOptions().also { it.maxFeatureFlags = 100 }
+    val globalBuffer = FeatureFlagBuffer.create(options)
+    val isolationBuffer = FeatureFlagBuffer.create(options)
+    val currentBuffer = FeatureFlagBuffer.create(options)
+    val expectedFlags = (0 until options.maxFeatureFlags).map { "flag$it" }
+
+    expectedFlags.forEach { globalBuffer.add(it, true) }
+
+    val stop = AtomicBoolean(false)
+    val writerFailure = AtomicReference<Throwable?>(null)
+    val writerStarted = CountDownLatch(1)
+    val writer = Thread {
+      try {
+        var i = 0
+        while (!stop.get()) {
+          globalBuffer.add(expectedFlags[i++ % expectedFlags.size], true)
+          writerStarted.countDown()
+        }
+      } catch (e: Exception) {
+        writerFailure.set(e)
+      }
+    }
+
+    writer.start()
+    try {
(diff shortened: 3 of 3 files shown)


adinauer:
## Feature flag buffer strategy benchmarks

I benchmarked the three implementations discussed here:

- **A — Current `main` (buggy):** mutable `CopyOnWriteArrayList`, with `merged()` traversing the live wrapper.
- **B — This PR:** same mutable write path, but `merged()` creates stable COW wrapper snapshots for the three scope buffers.
- **C — PR #5989:** immutable lists published through a volatile field; writers copy and publish a completed list.

The benchmark reproduces the historical `add()` workloads from #4812 and also measures merging three populated scopes and cloning.

### Add workloads

Each operation creates a buffer and performs the complete workload.

| Workload | A: Current | B: This PR | C: PR #5989 |
|---|---:|---:|---:|
| Unique | 10.09 µs / 30,848 B | 10.35 µs / 30,848 B | 19.14 µs / 41,137 B |
| Duplicates | 2.17 µs / 7,312 B | 2.16 µs / 7,312 B | 2.43 µs / 16,016 B |
| Full + duplicates | 25.72 µs / 117,249 B | 30.00 µs / 117,249 B | 51.26 µs / 97,137 B |
| Realistic | 6.90 µs / 27,168 B | 6.95 µs / 27,168 B | 13.97 µs / 32,256 B |

A and B have identical `add()` implementations, so their timing differences are benchmark variance. This PR preserves the existing write-path performance.

Compared with current `main`, PR #5989 was approximately:

- 2× slower for unique, full-duplicate, and realistic workloads;
- 12% slower for duplicate-only updates;
- 19% more allocation for the realistic workload.

### Three-scope merge

Each scope contains the same flag names, modeling substantial deduplication between global, isolation, and current scopes.

| Entries per scope | A: Current | B: This PR | C: PR #5989 |
|---:|---:|---:|---:|
| 0 | 2.42 ns / ~0 B | 2.90 ns / 48 B | 1.34 ns / ~0 B |
| 10 | 262.96 ns / 1,304 B | 273.63 ns / 1,424 B | 248.96 ns / 1,232 B |
| 100 | 2,143.05 ns / 7,024 B | 2,258.90 ns / 7,144 B | 1,945.93 ns / 6,592 B |

At 100 entries per scope:

- This PR costs **5.4% more merge time and 120 additional bytes** versus current `main`.
- On the benchmark's OpenJDK 17 runtime, the fixed 120-byte cost is three COW wrapper allocations; their backing arrays are shared rather than copied.
- PR #5989 is **9.2% faster** than current `main` for merging and allocates **432 fewer bytes**.

### Clone at 100 entries

| Strategy | Time | Allocation |
|---|---:|---:|
| A: Current | 6.56 ns | 112 B |
| B: This PR | 6.82 ns | 112 B |
| C: PR #5989 | 4.16 ns | 72 B |

### Trade-off

This PR keeps the hot feature-flag evaluation/write path unchanged and pays a small cost when capturing an event. It fixes the structural `size()`/indexed-`get()` crash, but a lock-free snapshot can still observe an intermediate step of the existing compound `add()` operation.

PR #5989 provides stronger completed-write snapshot semantics and cheaper merge/clone operations, but moves substantially more allocation and CPU cost onto feature-flag additions. Based on these measurements, this PR favors preserving write performance while applying the smallest fix for the demonstrated crash.

### Methodology and caveats

- JMH on OpenJDK 17, with 3 warmup iterations, 5 measurement iterations, and 2 forks.
- GC profiler used for normalized allocation measurements.
- Single-threaded microbenchmark; lock contention is not measured.
- The harness mirrors the implementations rather than invoking SDK classes directly.
- Android's `CopyOnWriteArrayList` implementation may have different constructor costs; sharing the backing array is an implementation optimization, not a public complexity guarantee.

sentry/src/main/java/io/sentry/featureflags/FeatureFlagBuffer.java
```diff
@@ -135,13 +135,13 @@ public void clear() {
       final @Nullable FeatureFlagBuffer isolationBuffer,
       final @Nullable FeatureFlagBuffer currentBuffer) {
 
-    // Capture references to avoid inconsistencies from concurrent modifications
+    // Capture snapshots to avoid inconsistencies from concurrent modifications
     final @Nullable CopyOnWriteArrayList<FeatureFlagEntry> globalFlags =
-        globalBuffer == null ? null : globalBuffer.flags;
+        globalBuffer == null ? null : new CopyOnWriteArrayList<>(globalBuffer.flags);
```
runningcode: why does this need to be a `CopyOnWriteArrayList` ? could it be `Collections.unmodifiableList` or just an `ArrayList` ?

runningcode (approved):
Can we still remove the volatile on the `flags` ?

sentry/src/main/java/io/sentry/featureflags/FeatureFlagBuffer.java
adinauer: We make use of the optimization in modern JDK / Android:

```
    public CopyOnWriteArrayList(Collection<? extends E> c) {
        Object[] es;
        if (c.getClass() == CopyOnWriteArrayList.class)
            es = ((CopyOnWriteArrayList<?>)c).getArray();
        else {
            es = c.toArray();
            if (c.getClass() != java.util.ArrayList.class)
                es = Arrays.copyOf(es, es.length, Object[].class);
        }
        setArray(es);
    }
```

This avoids copying the underlying array and instead reuses it until the next modification.

runningcode: Ah gotcha. I think we should add a comment about that for future readers of the codebase since this is somewhat unexpected here since we never modify the list.

sentry/src/main/java/io/sentry/featureflags/FeatureFlagBuffer.java
```diff
@@ -135,13 +135,15 @@ public void clear() {
       final @Nullable FeatureFlagBuffer isolationBuffer,
       final @Nullable FeatureFlagBuffer currentBuffer) {
 
-    // Capture references to avoid inconsistencies from concurrent modifications
+    // Capture structurally stable snapshots before indexed traversal. Passing a
+    // CopyOnWriteArrayList directly allows its collection constructor to reuse the immutable
+    // backing array instead of copying the elements on runtimes that support this optimization.
     final @Nullable CopyOnWriteArrayList<FeatureFlagEntry> globalFlags =
-        globalBuffer == null ? null : globalBuffer.flags;
+        globalBuffer == null ? null : new CopyOnWriteArrayList<>(globalBuffer.flags);
```
runningcode: After thinking about this overnight, how can we guarantee that the `CopyOnWriteArrayList` is created from an atomic copy? For example, if the copy is created in the middle of an add operation, we might see a missing feature flag or `maxSize + 1` feature flags.

runningcode (changes requested):
Hey after thinking about this overnight I don't think this is safe as it is. Going to request changes until we discuss.

sentry/src/main/java/io/sentry/featureflags/FeatureFlagBuffer.java
adinauer: Good catch! Will fix.

adinauer:
New benchmarks with latest changes:

```
 ### Add workloads

 ┌───────────────────┬──────────────────────┬──────────────────────┬─────────────────────┐
 │ Workload          │ main                 │ #5994                │ #5989               │
 ├───────────────────┼──────────────────────┼──────────────────────┼─────────────────────┤
 │ Unique            │ 10.64 µs / 32,464 B  │ 9.65 µs / 32,464 B   │ 19.00 µs / 41,153 B │
 ├───────────────────┼──────────────────────┼──────────────────────┼─────────────────────┤
 │ Duplicates        │ 2.39 µs / 8,928 B    │ 2.29 µs / 8,928 B    │ 2.47 µs / 16,032 B  │
 ├───────────────────┼──────────────────────┼──────────────────────┼─────────────────────┤
 │ Full + duplicates │ 27.67 µs / 120,465 B │ 28.15 µs / 120,465 B │ 53.69 µs / 97,154 B │
 ├───────────────────┼──────────────────────┼──────────────────────┼─────────────────────┤
 │ Realistic         │ 7.15 µs / 28,624 B   │ 6.99 µs / 28,624 B   │ 14.02 µs / 32,272 B │
 └───────────────────┴──────────────────────┴──────────────────────┴─────────────────────┘

 #5994 preserves main write performance. #5989 remains approximately 2× slower for realistic writes.

 ### Three-scope merge

 ┌───────────────┬───────────────────────┬───────────────────────┬───────────────────────┐
 │ Entries/scope │ main                  │ #5994                 │ #5989                 │
 ├───────────────┼───────────────────────┼───────────────────────┼───────────────────────┤
 │ 0             │ 2.44 ns / ~0 B        │ 8.93 ns / 48 B        │ 1.63 ns / ~0 B        │
 ├───────────────┼───────────────────────┼───────────────────────┼───────────────────────┤
 │ 10            │ 244.34 ns / 1,248 B   │ 246.43 ns / 1,296 B   │ 245.06 ns / 1,200 B   │
 ├───────────────┼───────────────────────┼───────────────────────┼───────────────────────┤
 │ 100           │ 2,182.49 ns / 6,968 B │ 2,135.00 ns / 7,016 B │ 2,075.99 ns / 6,560 B │
 └───────────────┴───────────────────────┴───────────────────────┴───────────────────────┘

 The three uncontended lock acquisitions in #5994 have little measurable effect for populated buffers. Empty merges are notably slower, though still only ~9 ns.

 ### Clone and direct read at 100 entries

 ┌───────────┬─────────────────────┬─────────────────────┬─────────────────────┐
 │ Operation │ main                │ #5994               │ #5989               │
 ├───────────┼─────────────────────┼─────────────────────┼─────────────────────┤
 │ Clone     │ 5.08 ns / 80 B      │ 6.99 ns / 80 B      │ 2.84 ns / 40 B      │
 ├───────────┼─────────────────────┼─────────────────────┼─────────────────────┤
 │ Read      │ 244.69 ns / 3,800 B │ 200.17 ns / 2,840 B │ 460.57 ns / 2,872 B │
 └───────────┴─────────────────────┴─────────────────────┴─────────────────────┘

 #5989 wins clearly on cloning. #5994’s pre-sized result list makes direct reads faster than main; #5989’s current unmodifiable-list iterator is slower in this benchmark.
```

And

```
If scope forks greatly outnumber feature-flag additions, #5989 becomes faster overall for the buffer.

 Using the realistic-add results:

 - #5994 saves about 78 ns per feature-flag addition.
 - #5989 saves about 4.15 ns per buffer clone.

 CPU break-even is therefore approximately:

 ```text
   78 / 4.15 ≈ 19 clones per feature-flag addition
 ```

 So:

 - Fewer than ~19 clones per addition: #5994 likely wins.
 - More than ~19 clones per addition: #5989 likely wins.
 - No feature flags added but scopes frequently forked: #5989 clearly wins.

 For allocation, the crossover is much earlier:

 - #5989 saves 40 B per clone.
 - It adds roughly 41 B per realistic addition.

 Thus, at roughly one clone per addition, #5989 already becomes competitive on allocations.

 This can matter with OpenTelemetry, where scope forking may happen for every span while feature flags are added relatively infrequently. However, the measured clone difference is only the FeatureFlagBuffer portion—about 4 ns and 40 B. A complete scope clone performs considerably more work, so an end-to-end scope-fork benchmark would be needed to determine how visible that saving is in the full application flow.
```

Also

```
maxFeatureFlags = 100 does not preallocate 100 slots. Both start empty, but #5994 allocates an empty CopyOnWriteArrayList, while #5989 shares the Collections.emptyList() singleton.

 I added targeted no-add benchmarks:

 ┌──────────────────────┬──────────────────┬────────────────┬────────────────┐
 │ Operation            │ #5994            │ #5989          │ #5994 overhead │
 ├──────────────────────┼──────────────────┼────────────────┼────────────────┤
 │ Create empty buffer  │ 6.87 ns / 96 B   │ 3.13 ns / 40 B │ 3.74 ns / 56 B │
 ├──────────────────────┼──────────────────┼────────────────┼────────────────┤
 │ Clone an empty chain │ 10.70 ns / 128 B │ 3.04 ns / 40 B │ 7.66 ns / 88 B │
 └──────────────────────┴──────────────────┴────────────────┴────────────────┘

 The clone-chain benchmark repeatedly clones the previous result. This matters because #5994’s snapshot acquires the source buffer’s lazily initialized lock. With no prior add(), that can initialize a ReentrantLock during cloning. #5989’s clone neither touches a lock nor creates a list—it only shares the immutable empty-list reference.

 For repeated sibling clones from an already-locked source, the earlier result was:

 ```text
   #5994: 6.69 ns / 80 B
   #5989: 2.87 ns / 40 B
 ```

 So with feature flags never used, #5989 is consistently cheaper, particularly in allocations. For a full fork cloning two feature-flag buffers, the nested-chain model saves approximately 15 ns and 176 B per fork.
```

adinauer:
Closing this PR in favor of #5989 .
While this PRs version is faster during `add()`, it's slower when cloning/forking `Scope` and not everyone uses feature flags. Given the default of `maxFeatureFlags=100`, #5989 is faster for anyone not using feature flags.

When capturing an event #5989 is also faster (merge performance).

PR #6003, opened 2026-08-26
fix(okhttp): keep the wrapped EventListener per Call

### LINKED ISSUE #5967: SentryOkHttpEventListener breaks EventListener.Factory contract
### Integration

sentry-okhttp

### Java Version

8 but it's unrelated

No

### Other Error Monitoring Solution Name

_No response_

### Version

8.21.1 but it's unrelated

### Steps to Reproduce

This might be more related to the Sentry Android Gradle plugin than the Java project per se, but if we want to use SentryOkHttpEventListener (which I presume sends some useful Breadcrumbs? at least) I think the fix will need to be here (or for the Gradle plugin to use EventListener.plus in OkHttp 5.13, I think). 

It looks like SentryOkHttpEventListener isn't designed for concurrency and that it breaks the EventListener.Factory contract.

If you use a custom EventListener.Factory with the Sentry Android Gradle plugin, the custom Factory gets wrapped by the SentryOkHttpEventListener.

However, the SentryOkHttpEventListener doesn't respect the EventListener.Factory contract, which expects that the EventListener instance is used for the lifetime of the Call. Instead, `SentryOkHttpEventListener` will pass the _latest_ EventListener instance to every concurrent Call, because it's updating the 1 instance it uses after every `SentryOkHttpEventListener.callStart`.

The steps to repro were simply:

1. use SentryOkHttpEventListener around a custom EventListener.Factory (only 1 eventListener or 1 eventListenerFactory can be set on an OkHttpClient at the same time)
2. in the custom EventListener.Factory, capture the original Call
3. launch multiple overlapping requests

```
    private class MismatchDetectingListener(
        private val ownCall: Call,
        private val mismatches: MutableList<String>,
    ) : EventListener() {
        private fun verify(method: String, cbCall: Call) {
            if (cbCall !== ownCall) {
                val msg = "MISMATCH $method: built for ${ownCall.request().url} invoked with ${cbCall.request().url} " +
                    "thread=${Thread.currentThread().name}"
                mismatches += msg
                Log.e(TAG, msg)
            }
        }

        override fun callStart(call: Call) = verify("callStart", call)
        override fun dnsStart(call: Call, domainName: String) = verify("dnsStart", call)
    }
.... 
    val perCallFactory = EventListener.Factory { call -> MismatchDetectingListener(call, mismatches) }
    val client = OkHttpClient.Builder()
      .eventListener(SentryOkHttpEventListener(originalEventListenerFactory = perCallFactory))
      .build()
....

    val response1 = async { client.newCall(request1).execute() }
    val response2 = async { client.newCall(request2).execute() }
    val response3 = async { client.newCall(request3).execute() }
    val response4= async { client.newCall(request4).execute() }

```

### Expected Result

Each EventListener should only ever receive calls from the Call it was created alongside in EventListener.Factory.create

### Actual Result

The latest EventListener that was created receives every callback

markushi:
## :scroll: Description

`SentryOkHttpEventListener` held the wrapped `EventListener` in a single mutable field that
`callStart` overwrote for each call. It is now kept in a per-`Call` map, the same pattern the class
already uses for `eventMap`. No public API change.

OkHttp uses one listener instance for all calls, thus concurrent calls were all delegated to the
listener made for the call that started last. This breaks the `EventListener.Factory` contract and
loses the terminal `callEnd`/`callFailed` of every overlapping call.

* resolves: #5967
* resolves: JAVA-695

Added unit tests.

## :pencil: Checklist

### DIFF
--- a/sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpEventListener.kt
+++ b/sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpEventListener.kt
@@ -45,7 +45,7 @@ public open class SentryOkHttpEventListener(
   private val scopes: IScopes = ScopesAdapter.getInstance(),
   private val originalEventListenerCreator: ((call: Call) -> EventListener)? = null,
 ) : EventListener() {
-  private var originalEventListener: EventListener? = null
+  private val originalEventListenerMap: MutableMap<Call, EventListener> = ConcurrentHashMap()
 
   public companion object {
     internal const val PROXY_SELECT_EVENT = "http.client.proxy_select_ms"
@@ -85,18 +85,23 @@ public open class SentryOkHttpEventListener(
   ) : this(scopes, originalEventListenerCreator = { originalEventListenerFactory.create(it) })
 
   override fun callStart(call: Call) {
-    originalEventListener = originalEventListenerCreator?.invoke(call)
+    // The EventListener.Factory contract binds a listener to a single call, so the wrapped
+    // listener is kept per call instead of in a field shared by all concurrent calls
+    val originalEventListener = originalEventListenerCreator?.invoke(call)
+    if (originalEventListener != null) {
+      originalEventListenerMap[call] = originalEventListener
+    }
     originalEventListener?.callStart(call)
     // If the wrapped EventListener is ours, we can just delegate the calls,
     // without creating other events that would create duplicates
-    if (canCreateEventSpan()) {
+    if (canCreateEventSpan(originalEventListener)) {
       eventMap[call] = SentryOkHttpEvent(scopes, call.request())
     }
   }
 
   override fun proxySelectStart(call: Call, url: HttpUrl) {
-    originalEventListener?.proxySelectStart(call, url)
-    if (!canCreateEventSpan()) {
+    originalEventListenerMap[call]?.proxySelectStart(call, url)
+    if (!canCreateEventSpan(call)) {
       return
     }
     val okHttpEvent: SentryOkHttpEvent = eventMap[call] ?: return
@@ -104,8 +109,8 @@ public open class SentryOkHttpEventListener(
   }
 
   override fun proxySelectEnd(call: Call, url: HttpUrl, proxies: List<Proxy>) {
-    originalEventListener?.proxySelectEnd(call, url, proxies)
-    if (!canCreateEventSpan()) {
+    originalEventListenerMap[call]?.proxySelectEnd(call, url, proxies)
+    if (!canCreateEventSpan(call)) {
       return
     }
     val okHttpEvent: SentryOkHttpEvent = eventMap[call] ?: return
@@ -117,8 +122,8 @@ public open class SentryOkHttpEventListener(
   }
 
   override fun dnsStart(call: Call, domainName: String) {
-    originalEventListener?.dnsStart(call, domainName)
-    if (!canCreateEventSpan()) {
+    originalEventListenerMap[call]?.dnsStart(call, domainName)
+    if (!canCreateEventSpan(call)) {
       return
     }
     val okHttpEvent: SentryOkHttpEvent = eventMap[call] ?: return
@@ -126,8 +131,8 @@ public open class SentryOkHttpEventListener(
   }
 
   override fun dnsEnd(call: Call, domainName: String, inetAddressList: List<InetAddress>) {
-    originalEventListener?.dnsEnd(call, domainName, inetAddressList)
-    if (!canCreateEventSpan()) {
+    originalEventListenerMap[call]?.dnsEnd(call, domainName, inetAddressList)
+    if (!canCreateEventSpan(call)) {
       return
     }
     val okHttpEvent: SentryOkHttpEvent = eventMap[call] ?: return
@@ -140,8 +145,8 @@ public open class SentryOkHttpEventListener(
   }
 
   override fun connectStart(call: Call, inetSocketAddress: InetSocketAddress, proxy: Proxy) {
-    originalEventListener?.connectStart(call, inetSocketAddress, proxy)
-    if (!canCreateEventSpan()) {
+    originalEventListenerMap[call]?.connectStart(call, inetSocketAddress, proxy)
+    if (!canCreateEventSpan(call)) {
       return
     }
     val okHttpEvent: SentryOkHttpEvent = eventMap[call] ?: return
@@ -149,8 +154,8 @@ public open class SentryOkHttpEventListener(
   }
 
   override fun secureConnectStart(call: Call) {
-    originalEventListener?.secureConnectStart(call)
-    if (!canCreateEventSpan()) {
+    originalEventListenerMap[call]?.secureConnectStart(call)
+    if (!canCreateEventSpan(call)) {
       return
     }
     val okHttpEvent: SentryOkHttpEvent = eventMap[call] ?: return
@@ -158,8 +163,8 @@ public open class SentryOkHttpEventListener(
   }
 
   override fun secureConnectEnd(call: Call, handshake: Handshake?) {
-    originalEventListener?.secureConnectEnd(call, handshake)
-    if (!canCreateEventSpan()) {
+    originalEventListenerMap[call]?.secureConnectEnd(call, handshake)
+    if (!canCreateEventSpan(call)) {
       return
     }
     val okHttpEvent: SentryOkHttpEvent = eventMap[call] ?: return
@@ -172,8 +177,8 @@ public open class SentryOkHttpEventListener(
     proxy: Proxy,
     protocol: Protocol?,
   ) {
-    originalEventListener?.connectEnd(call, inetSocketAddress, proxy, protocol)
-    if (!canCreateEventSpan()) {
+    originalEventListenerMap[call]?.connectEnd(call, inetSocketAddress, proxy, protocol)
+    if (!canCreateEventSpan(call)) {
       return
     }
     val okHttpEvent: SentryOkHttpEvent = eventMap[call] ?: return
@@ -188,8 +193,8 @@ public open class SentryOkHttpEventListener(
     protocol: Protocol?,
     ioe: IOException,
   ) {
-    originalEventListener?.connectFailed(call, inetSocketAddress, proxy, protocol, ioe)
-    if (!canCreateEventSpan()) {
+    originalEventListenerMap[call]?.connectFailed(call, inetSocketAddress, proxy, protocol, ioe)
+    if (!canCreateEventSpan(call)) {
       return
     }
     val okHttpEvent: SentryOkHttpEvent = eventMap[call] ?: return
@@ -202,8 +207,8 @@ public open class SentryOkHttpEventListener(
   }
 
   override fun connectionAcquired(call: Call, connection: Connection) {
-    originalEventListener?.connectionAcquired(call, connection)
-    if (!canCreateEventSpan()) {
+    originalEventListenerMap[call]?.connectionAcquired(call, connection)
+    if (!canCreateEventSpan(call)) {
       return
     }
     val okHttpEvent: SentryOkHttpEvent = eventMap[call] ?: return
@@ -211,8 +216,8 @@ public open class SentryOkHttpEventListener(
   }
 
   override fun connectionReleased(call: Call, connection: Connection) {
-    originalEventListener?.connectionReleased(call, connection)
-    if (!canCreateEventSpan()) {
+    originalEventListenerMap[call]?.connectionReleased(call, connection)
+    if (!canCreateEventSpan(call)) {
       return
     }
     val okHttpEvent: SentryOkHttpEvent = eventMap[call] ?: return
@@ -220,8 +225,8 @@ public open class SentryOkHttpEventListener(
   }
 
   override fun requestHeadersStart(call: Call) {
-    originalEventListener?.requestHeadersStart(call)
-    if (!canCreateEventSpan()) {
+    originalEventListenerMap[call]?.requestHeadersStart(call)
+    if (!canCreateEventSpan(call)) {
       return
     }
     val okHttpEvent: SentryOkHttpEvent = eventMap[call] ?: return
@@ -229,8 +234,8 @@ public open class SentryOkHttpEventListener(
   }
 
   override fun requestHeadersEnd(call: Call, request: Request) {
-    originalEventListener?.requestHeadersEnd(call, request)
-    if (!canCreateEventSpan()) {
+    originalEventListenerMap[call]?.requestHeadersEnd(call, request)
+    if (!canCreateEventSpan(call)) {
       return
     }
     val okHttpEvent: SentryOkHttpEvent = eventMap[call] ?: return
@@ -238,8 +243,8 @@ public open class SentryOkHttpEventListener(
   }
 
   override fun requestBodyStart(call: Call) {
-    originalEventListener?.requestBodyStart(call)
-    if (!canCreateEventSpan()) {
+    originalEventListenerMap[call]?.requestBodyStart(call)
+    if (!canCreateEventSpan(call)) {
       return
     }
     val okHttpEvent: SentryOkHttpEvent = eventMap[call] ?: return
@@ -247,8 +252,8 @@ public open class SentryOkHttpEventListener(
   }
 
   override fun requestBodyEnd(call: Call, byteCount: Long) {
-    originalEventListener?.requestBodyEnd(call, byteCount)
-    if (!canCreateEventSpan()) {
+    originalEventListenerMap[call]?.requestBodyEnd(call, byteCount)
+    if (!canCreateEventSpan(call)) {
       return
     }
     val okHttpEvent: SentryOkHttpEvent = eventMap[call] ?: return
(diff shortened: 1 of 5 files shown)


sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpEventListener.kt
```diff
@@ -370,26 +397,26 @@ public open class SentryOkHttpEventListener(
   }
 
   override fun canceled(call: Call) {
-    originalEventListener?.canceled(call)
+    originalEventListenerMap[call]?.canceled(call)
```
0xadam-brown: h: We only delegate `canceled()` if `originalEventListenerMap` contains the call, but [OkHttp docs](https://github.com/lysine-dev/okhttp/blob/b830f03c8f161883e0213c22c170c3d1f5f5eaa6/okhttp/src/commonJvmAndroid/kotlin/okhttp3/EventListener.kt#L432) say it's legal to call `canceled()` before `callStart()` and after `callEnd()`.

Looks like we're regressing in cases where `canceled()` follows `callEnd()` for the single-call case that used to delegate through the stale field.

0xadam-brown (commented):
Thanks for this 💯 !

One comment worth addressing; otherwise looking good.

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpEventListener.kt
markushi: Confirmed, and fixed in b50807017c.

`EventListener.canceled()` is documented as "It may be invoked at any point in a call's life, including before [callStart] and after [callEnd]", and `RealCall.cancel()` calls `eventListener.canceled(this)` directly, independent of the callStart/callEnd window. So the map lookup can miss on both edges.

`canceled()` now falls back to `originalEventListenerCreator?.invoke(call)` when the map has no entry. For the single-listener constructors the creator returns the same instance, so the pre-callStart and post-callEnd cases behave exactly as before this PR. The created listener is deliberately *not* put into `originalEventListenerMap`: a call that is canceled before it starts never gets a `callEnd()`/`callFailed()`, so nothing would ever remove the entry.

Covered by four new cases in `SentryOkHttpEventListenerDelegationTest`: `cancel before callStart is delegated`, `cancel after callEnd is delegated`, `cancel outside of a call does not bind a listener to the call`, and `a single wrapped listener receives cancels outside of the call window`.

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpEventListener.kt
```diff
@@ -330,8 +355,9 @@ public open class SentryOkHttpEventListener(
   }
 
   override fun responseFailed(call: Call, ioe: IOException) {
+    val originalEventListener = originalEventListenerMap[call]
     originalEventListener?.responseFailed(call, ioe)
-    if (!canCreateEventSpan()) {
+    if (!canCreateEventSpan(originalEventListener)) {
       return
     }
     val okHttpEvent: SentryOkHttpEvent = eventMap[call] ?: return
```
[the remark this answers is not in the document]
markushi: No code change — the premise is wrong. `canceled()` is not a terminal event in OkHttp.

From `okhttp3.EventListener.canceled` (5.3.2): "Note that cancellation is best-effort and that a call may proceed normally after it has been canceled" and "It may be invoked at any point in a call's life, including before [callStart] and after [callEnd]." `RealCall.cancel()` only sets the `canceled` flag and cancels the in-flight exchange; the call still terminates through `callEnd()` or `callFailed()`, both of which already `remove(call)` from `originalEventListenerMap`.

Removing the entry in `canceled()` would therefore not fix a leak, it would drop delegation of every remaining callback of a canceled-but-still-running call. The new test `cancel during a call is delegated to the listener of that call` asserts that `callStart` → `canceled` → `callFailed` all reach the same wrapped listener.

The one real gap on this line was the opposite one (raised in the other thread): a `canceled()` with no map entry was silently dropped. That is fixed in b50807017c without inserting into the map, so there is still no unbounded growth.

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpEventListener.kt
```diff
@@ -370,26 +397,32 @@ public open class SentryOkHttpEventListener(
   }
 
   override fun canceled(call: Call) {
-    originalEventListener?.canceled(call)
+    // Unlike the other callbacks, canceled() can occur before callStart() and after
+    // callEnd()/callFailed(), so there is not always a listener bound to the call. Create one on
+    // the fly in that case, but do not put it in the map: nothing would remove it again, because
+    // a call that is canceled before it starts never gets a callEnd() or callFailed().
+    val originalEventListener =
```
0xadam-brown: We're close! (and thanks for the great updates)

We still need to preserve the contract of EventListener.Factory that ensures only one EventListener instance is produced per Call lifecycle. Ie, the listener returned by the factory for a given call needs to be the listener that captures i) all of that call's lifecycle and ii) no other call's lifecycle. 

We've fixed (ii), but we're still violating (i) in the case of cancelation because we're creating an extra listener for early and late cancel() invocations.

**Possible solution**

Thoughts about using a weak per-call map for the wrapped listener instead? Something like a WeakHashMap<Call, EventListener> guarded by synchronized, with a getOrCreateOriginalEventListener(call) helper used by both callStart and canceled(). 

That'd^^ let us avoid removing entries on callEnd / callFailed, and completed calls would be gc'd as soon as the Call instance is unreachable.

sentry-okhttp/src/test/java/io/sentry/okhttp/SentryOkHttpEventListenerDelegationTest.kt
```diff
@@ -0,0 +1,193 @@
...
+    sut.canceled(call)
+    sut.callFailed(call, IOException())
+
+    assertThat(fixture.listeners.single().received.map { it.first })
+      .containsExactly("callStart", "canceled", "callFailed")
+      .inOrder()
+  }
+
+  @Test
+  fun `cancel before callStart is delegated`() {
```
0xadam-brown: Thanks for the new tests 💯 

Bonus points if our cancellation tests can assert the stronger factory-contract invariant 👍 

(Right now `cancel before callStart is delegated` and `cancel after callEnd is delegated` prove that *some* listener receives `canceled()`, rather than that a single listener receives all lifecycle callbacks.)

0xadam-brown (commented):
Excellent! One tweak more to satisfy the EventListener.Factory contract, and I think we'll be there 🥇

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpEventListener.kt
markushi: Fixed in 728129d9f5. You were right that creating an extra listener was the wrong trade — and a `WeakHashMap` turned out not to be usable here, for a reason our own code demonstrates.

**Why not weak keys.** `WeakHashMap` holds values by ordinary strong references, and its javadoc warns that a value which strongly refers to its own key prevents the key from being discarded. That is exactly our shape: a factory is *handed* the `Call`, so listeners that keep it are the normal case (our own `RecordingListener(val ownCall: Call)` does it). Worse, the sibling `eventMap` already has this cycle inside the SDK: `SentryOkHttpEvent` holds a `Response`, `Response.exchange` is an `Exchange`, and `Exchange.call` is the `RealCall`. So a weak map would stop collecting as soon as a response is recorded. It is also unsynchronized, and `getTable()` calls `expungeStaleEntries()`, so even reads would need the monitor on a path OkHttp explicitly allows to run concurrently.

**What we do instead.** OkHttp stores the listener on the call itself (`RealCall:73`, `client.eventListenerFactory.create(this)`), so its lifetime is the `Call` object's lifetime. Ours is the `callStart()..callEnd()` window. `canceled()` is the one event that escapes that window, so it is the one that needed handling — and `Call.isExecuted()` tells the two edges apart without any storage of our own:

- **not executed** — the cancel precedes `callStart()`. `computeIfAbsent` creates *and binds* the listener, and `callStart()` then reuses it, so one listener sees the whole lifecycle. It is self-cleaning: `getResponseWithInterceptorChain` does `if (canceled) throw IOException("Canceled")`, so a pre-canceled call that is later executed still runs `callStart` → `callFailed` and the entry is removed.
- **executed, entry present** — in flight, delegated to the bound listener.
- **executed, entry absent** — the terminal event already passed. Ignored. `Call.cancel()` is documented as "Requests that are already complete cannot be canceled", so there is nothing to report, and fabricating a second listener would both break the contract and leak the entry.

A stored "terminal" flag would have worked too, but it needs a per-`Call` marker that no later event can ever remove — the same unbounded growth, just with a smaller value behind a `Call` key. `isExecuted()` is that flag, maintained by OkHttp, for free.

Two supporting changes: `getOrCreateEventListener` uses `ConcurrentHashMap.computeIfAbsent` rather than a get-then-put, so a cancel racing `callStart()` cannot make the factory produce two listeners for one call; and the constructors that wrap a **single** `EventListener` now keep it in a field. That instance is shared across calls by definition — it is precisely what OkHttp's own `EventListener.asFactory()` does — so it exists independently of the window and receives every cancel, which restores the pre-PR behaviour you flagged with no leak and no contract question.

Residual gap, stated plainly: a call that is canceled before it starts and then *never* executed keeps its map entry, because no terminal event ever arrives. That is far narrower than the late-cancel leak it replaces, and bounding it would need the weak keys that do not work here.

sentry-okhttp/src/test/java/io/sentry/okhttp/SentryOkHttpEventListenerDelegationTest.kt
markushi: Done in 728129d9f5. The cancellation tests now assert the single-listener invariant rather than "some listener got it".

The blocker was that these tests drive the listener by hand with `client.newCall(...)`, so `Call.isExecuted()` is always `false` and the post-terminal branch was unreachable. Added a `Fixture.mockCall(path, isExecuted)` helper so each test states the call state it is exercising.

- `cancel before callStart binds the listener that callStart then reuses` — asserts `fixture.listeners` has size **1** and that the one listener receives `canceled, callStart, dnsStart, callEnd` in order.
- `cancel after the terminal event is ignored` — size **1**, receiving exactly `callStart, callEnd`; no second listener is created and no stray `canceled` is delivered.
- `cancel after a failed call is ignored` — same, via `callFailed`.
- `cancel during a call is delegated to the listener of that call` — `callStart, canceled, callFailed` all on one listener.
- `a single wrapped listener receives cancels outside of the call window` — covers the fixed-instance constructors, which keep delegating cancels at any time.

The `hasSize(1)` assertions are the ones carrying the factory contract: they fail if we ever invoke the factory more than once for a call.

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpEventListener.kt
markushi: Correction to my previous reply — 7b7f12e4ab simplifies this further and removes the residual leak I disclosed there.

The `Call.isExecuted()` branch is gone. It bound a listener for a cancel that precedes `callStart()`, and that is exactly the case that can never be cleaned up: `RealCall.cancel()` fires `canceled()` unconditionally, but `callStart` is only reached from `execute()`/`enqueue()`, so `newCall()` → `cancel()` → never executed produces **no terminal event at all**. Any entry added there would retain the `Call` forever. (I checked the other paths — `AsyncCall.run()`, `failRejected()`, and cancel-while-queued all reach `callDone()` — so never-executed calls are the only such case.)

`canceled()` is now simply:

```kotlin
val originalEventListener = originalEventListenerMap[call] ?: fixedOriginalEventListener
originalEventListener?.canceled(call)
```

Out-of-window cancels are not delegated to factory-created listeners, and no listener is fabricated to receive them. The rationale is written into a comment on the method itself rather than left in this thread.

Little is actually lost: if the call is never executed there is no lifecycle to observe, and if it is executed after being canceled, OkHttp fails it with `IOException("Canceled")`, so the listener still learns about the cancellation through `callFailed()`.

This gives your invariant without qualification — for any `Call`, the factory is invoked at most once, and that single listener sees the whole `callStart()`..terminal window and nothing from any other call. `getOrCreateEventListener` keeps `computeIfAbsent` so concurrent callbacks cannot produce two listeners for one call. Listeners passed as a single instance still receive every cancel, since they are shared by all calls by definition — the same as `EventListener.asFactory()`.

Tests updated accordingly: `cancel before callStart is not delegated and creates no listener`, `a call canceled before callStart still gets a single listener when it starts`, `cancel after the terminal event is ignored`, `cancel after a failed call is ignored`. The mocked `Call` is no longer needed, so they run against real `Call` instances again.

sentry-okhttp/src/test/java/io/sentry/okhttp/SentryOkHttpEventListenerDelegationTest.kt
markushi: Follow-up: the cancellation tests changed again in 7b7f12e4ab, along with the behaviour (see the other thread). The invariant they assert is unchanged and now stronger, because there is no case left where a second listener could appear.

- `cancel before callStart is not delegated and creates no listener` — `fixture.listeners` is empty.
- `a call canceled before callStart still gets a single listener when it starts` — size **1**, receiving `callStart, callFailed`.
- `cancel after the terminal event is ignored` — size **1**, receiving `callStart, callEnd`.
- `cancel after a failed call is ignored` — size **1**, receiving `callStart, callFailed`.
- `cancel during a call is delegated to the listener of that call` — `callStart, canceled, callFailed` on one listener.
- `a single wrapped listener receives cancels outside of the call window` — the fixed-instance constructors keep delegating cancels at any time.

The mocked `Call` from my previous update is gone; these run against real `Call` instances again. Ten tests in the class, 94 in the module.

victorlai-stripe:
Lurking - hi @0xadam-brown! Crazy seeing you around, hope you're doing well!

0xadam-brown:
> Lurking - hi @0xadam-brown! Crazy seeing you around, hope you're doing well!

@victorlai-stripe – so good to see your name show up here! Definitely doing well (and enjoying Sentry). Hope Stripe's been an excellent change-up on your end too 💯 !

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpEventListener.kt
markushi: One correction to my last reply: `getOrCreateEventListener` no longer uses `computeIfAbsent`. AnimalSniffer rejected it against `gummy-bears-api-21` — both `ConcurrentHashMap.computeIfAbsent` and `java.util.function.Function` need Android API 24, so it broke the Java 17 build job.

Fixed in 0600e6c1ad by dropping the helper and putting the listener into the map directly in `callStart()`.

Nothing is weakened by this. `computeIfAbsent` was guarding against a `canceled()` racing `callStart()` on the same `Call`, and that race disappeared when `canceled()` stopped creating listeners. `callStart()` is now the only writer, and `EventListener.callStart` is documented as "invoked only once for a single [Call]", so exactly one listener is created per `Call` without needing an atomic map method.

The factory-contract invariant is unchanged, and `animalsnifferMain`, the 10 delegation tests and all 94 `sentry-okhttp` tests pass locally.

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpEventListener.kt
```diff
@@ -65,7 +70,9 @@ public open class SentryOkHttpEventListener(
 
   public constructor(
```
0xadam-brown, line 98: m: Given that we're deviating from the EventListener contract by ignoring cancel() calls outside callStart() and callEnd(), could be sensible to add KDocs to the constructors to put folks on notice. 

Eg, maybe something like:

```kotlin
/**
 * Wraps a single [okhttp3.EventListener] instance and shares it across [Call]s. 
 * 
 * All callbacks are forwarded to the listener, including those delivered outside
 * the normal `callStart()`..`callEnd()` / `callFailed()` window.
 */                                                                                                                                                                                                                                                                                                                                              
public constructor(originalEventListener: EventListener)    
```

```kotlin
/**                                                                                                                                                                                                                                                                                                                                              
  * Wraps an [okhttp3.EventListener.Factory] and creates a separate
  * [okhttp3.EventListener] per [Call]. 
  * 
  * Callbacks that OkHttp may deliver before `callStart()` or after `callEnd()` / 
  * `callFailed()`, such as [okhttp3.EventListener.canceled], are not guaranteed to
  * be forwarded to the listener.
  */                                                                                                                                                                                                                                                                                                                                              
public constructor(originalEventListenerFactory: Factory)      
```

sentry-okhttp/src/test/java/io/sentry/okhttp/SentryOkHttpEventListenerDelegationTest.kt
```diff
@@ -0,0 +1,212 @@
...
+    sut.canceled(call)
+    sut.callFailed(call, IOException())
+
+    assertThat(fixture.listeners.single().received.map { it.first })
+      .containsExactly("callStart", "canceled", "callFailed")
+      .inOrder()
+  }
+
+  @Test
+  fun `cancel before callStart is not delegated and creates no listener`() {
```
0xadam-brown, line 152: l: Might be worth a quick KDoc noting that we're deliberately deviating from the EventListener contract here + why we're doing so. (Same for relevant tests below.)

0xadam-brown (commented):
A few quick comments, but no blockers.

0xadam-brown (approved):

sentry-okhttp/src/main/java/io/sentry/okhttp/SentryOkHttpEventListener.kt
markushi: I've added it to the class doc 👍

PR #6005, opened 2026-08-26
feat(android): Set app start vitals on standalone app start children

buenaflor:
## :scroll: Description
Set `app.vitals.start.type` on the standalone `app.start` root, and copy `app.vitals.start.screen` + `app.vitals.start.type` onto all of its children — including user spans under `app.start.extended`. Headless starts omit screen. The `ui.load`-attached app start path is unchanged.

Mobile vitals app start breakdown with `app.start` children with the root using these attributes. They were only on the root, so child spans did not group. Matches Flutter (getsentry/sentry-dart#3988).

Example of the sample app app start breakdown (previously empty without this PR):
image

Related to https://github.com/getsentry/sentry/pull/122577

Unit tests in `PerformanceAndroidEventProcessorTest` (foreground, user descendants under `app.start.extended`, headless, `ui.load` non-propagation). Verified on `sentry-samples-android` (debug) that the envelope includes both attributes on the root, `app.start.extended`, and user children.

## :pencil: Checklist

## :crystal_ball: Next steps

Made with [Cursor](https://cursor.com)


markushi (approved):
LGTM!

sentry-android-core/src/main/java/io/sentry/android/core/PerformanceAndroidEventProcessor.java
```diff
@@ -186,6 +191,29 @@ public SentryEvent process(@NotNull SentryEvent event, @NotNull Hint hint) {
...
+      return;
+    }
+
+    traceContext.setData(APP_START_TYPE_DATA, appStartType);
+    final @Nullable Object screen = traceContext.getData().get(APP_START_SCREEN_DATA);
+
+    for (final @NotNull SentrySpan span : transaction.getSpans()) {
+      @Nullable Map<String, Object> data = span.getData();
+      if (data == null) {
+        data = new ConcurrentHashMap<>();
```
runningcode, line 207: if we're iterating through all the transaction's spans, is it possible that another thread can also be modifying these spans? specifically the data? If so we have a race condition here.

buenaflor: whoops, sry I just merged it, let me double check

buenaflor: as far as I can tell it's fine

SentryTransaction.spans is a private ArrayList built once and never leaves the capture pipeline, so nothing mutates it while we iterate. The data map is shared with the live Span, but it's a ConcurrentHashMap and we only put into it

the same behaviour happens below in `setContributingFlags`

PR #6006, opened 2026-08-26
fix(android): Read the app start type once per event processor pass

runningcode:
## :scroll: Description

`PerformanceAndroidEventProcessor.process()` read `AppStartMetrics.getAppStartType()` three separate times while building a single transaction:

1. to pick the measurement key (`app_start_cold` / `app_start_warm`)
2. in `attachAppStartSpans` to decide whether to attach the cold-start breakdown children
3. for `contexts.app.start_type` — which #6005 now also stamps onto the trace context and every child span as `app.vitals.start.type`

This snapshots the type once at the top of `process()` and passes it down, and makes `AppStartMetrics.appStartType` `volatile`.

`appStartType` is written on the main thread (activity lifecycle callbacks, the `ApplicationStartInfo` block) and read by the event processor, which runs on whichever thread finished the transaction — `SentryTracer.finish()` calls `captureTransaction()` synchronously, so for idle and deadline timeouts that is the Sentry timer thread, not the main thread. The field was neither `volatile` nor guarded, so the read was a plain data race.

Beyond the JMM issue, the repeated reads can genuinely disagree. `onActivityDestroyed` sets `appStartType = WARM` when the last activity goes away, so backgrounding the app while a transaction finishes on the timer thread can produce a transaction whose measurement says `app_start_cold` while `contexts.app.start_type` and every span's `app.vitals.start.type` say `warm`. Since #6005 that value is the grouping key for the whole mobile vitals app start breakdown, so a mismatch splits the breakdown rather than just mislabelling one field.

Two tests in `PerformanceAndroidEventProcessorTest` asserting that the measurement key, `contexts.app.start_type`, the trace context and every span agree, for cold and warm standalone app starts. Full `:sentry-android-core:testReleaseUnitTest` passes.

Note that the race itself is not deterministically unit-testable — there is no seam to flip the type between the reads inside a single `process()` call — so the tests lock in the invariant rather than reproducing the interleaving.

## :pencil: Checklist

## :crystal_ball: Next steps

Follow-up to #6005. See also the companion PR removing the `AppStartExtension` lock-ordering deadlock.


sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java
```diff
@@ -68,7 +68,9 @@ public enum AppStartType {
   public static final @NotNull AutoClosableReentrantLock staticLock =
       new AutoClosableReentrantLock();
 
-  private @NotNull AppStartType appStartType = AppStartType.UNKNOWN;
+  // Written on the main thread (activity lifecycle, ApplicationStartInfo) and read on whichever
```
markushi, line 71: 💯

markushi (approved):
Nice, consistency for the win!

PR #6009, opened 2026-08-26
docs: Add a concurrency section to AGENTS.md

runningcode:
## :scroll: Description

Adds a **Concurrency** section to `AGENTS.md`, next to the existing exception-handling guidance. The core rule: never hold a lock across a call that can re-enter the SDK — read what you need under the lock, release it, then make the call.

Also covers the second-lock escape hatch for compound operations that were only atomic because the lock spanned them, plus two supporting rules (`volatile` for cross-thread fields, and reading mutable shared state once per operation).

Requested by @buenaflor while reviewing #6007.

Two production ANRs came from the same mistake. In #6007, `AppStartExtension` held its lock while calling `finish()` on the app start transaction; finishing captures synchronously and runs `PerformanceAndroidEventProcessor`, which takes its own lock and then calls back into `AppStartExtension` — opposite lock order, main thread blocked, ANR. JAVA-553 was the same shape in `ReplayIntegration`. The `volatile` and single-read rules come from #6006, where `AppStartMetrics.appStartType` was read three times in one pass and could disagree with itself.

The exception-handling section already establishes the format for this kind of repository-wide rule, so this follows it: the rule, why it fails, a wrong/right code pair, and the scope.

Documentation only. `./gradlew spotlessApply` is clean.

## :pencil: Checklist

## :crystal_ball: Next steps

Independent of #6007 — this documents the rule, that PR fixes the instance.

#skip-changelog

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -155,6 +155,54 @@ Apply that pattern only where a broad catch is genuinely unavoidable — an entr
 arbitrary user code or third-party callbacks. Everywhere else, name the exception types. Say in the
 PR description why the broad catch is necessary.
 
+### Concurrency
+
+**Never hold a lock across a call that can re-enter the SDK.** Read what you need under the lock,
+release it, then make the call.
+
+Finishing a span or transaction, capturing an event, and invoking a user-supplied callback all run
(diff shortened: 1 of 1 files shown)


markushi (approved):
Nice!

PR #6015, opened 2026-08-27
feat(profiling): Drop profiler ids from spans no profile covers

markushi:
## :scroll: Description

Transactions and spans are tagged with the continuous profiler's `profiler_id` as soon as they start, but the OS only tells us later whether a Perfetto profile really exists. This PR lets anything tagged with an id that leads nowhere drop the reference before it is sent.

`IContinuousProfiler` gets one new method:

```java
@NotNull ProfileRecordingState getProfileRecordingState(
    SentryId profilerId, SentryDate startTime, SentryDate endTime);
```

`SentryTracer.finish()` asks it once for the root span window and once per child span window, and removes the `ProfileContext` plus the `profiler_id` span data only on `NOT_RECORDED`. `RECORDED` and `UNKNOWN` change nothing, so an outcome we do not know never costs a valid link. Only a window whose chunks all failed loses the id: a window no chunk covers stays `UNKNOWN`, as it may reach further back than the history, or a back-dated span may read as outside the chunk it ran in.

`PerfettoContinuousProfiler` answers from a bounded history of chunk records, guarded by a lock of its own so that a finishing transaction never waits for a chunk start, a chunk stop or an OS callback. The answer for a window follows one policy:

| Scenario | Answer |
|---|---|
| No chunk ever ran | `UNKNOWN` |
| A chunk covering part of the window was recorded | `RECORDED` |
| No covering chunk was recorded, but one has no outcome yet | `UNKNOWN` |
| Every covering chunk failed | `NOT_RECORDED` |
| No chunk covers the window | `UNKNOWN` |

The outcome of a chunk has a single owner on each side. `PerfettoProfiler` only ever reports a failure — an OS error code, a missing or empty trace file, or the result timeout — as soon as it knows, which is what makes the common rate-limit case cheap. `PerfettoContinuousProfiler` alone declares a chunk recorded, from the trace file it actually received, since the OS can report success and still leave nothing usable behind. `NOT_RECORDED` is final, so a result the OS delivers late cannot revive a chunk that was already given up on.

`NoOpContinuousProfiler`, `AndroidContinuousProfiler` and `JavaContinuousProfiler` answer `UNKNOWN`, which leaves every non-Perfetto path exactly as it was.

This is an alternative to #5993, which solved the same problem with callbacks from the profiler into the tracer. Pulling the state at send time removes the listener registration, the unregistration on finish, and the tracers that stay registered because they never finish.

Rate limiting is the common case on API 35+, and the OS reports it roughly 1 ms after the request. Without this, a rate-limited session produces transactions that link to profiles the backend never receives, which shows up in the UI as dangling profile references. There is no backend logic that removes such references.

Unit tests, on three levels:

- `ChunkRecordTest` — the state machine (`NOT_RECORDED` is final), whether a chunk has ended, and the window arithmetic, including the exact boundaries and a chunk that is still running.
- `PerfettoProfilerTest` — the record outcome for an OS error, a null result path, a missing trace file, the result timeout, and a result that arrives after the timeout, plus the two cases where a successful result leaves the record untouched.
- `PerfettoContinuousProfilerTest` — the answer for a running chunk, a chunk that is still being collected, a collected chunk, a failed chunk, a failed and an unknown chunk under one window, a window after the last chunk, a window that starts before the profiler did, per-chunk judgement across a failed and a recorded chunk, eviction, and close.
- `SentryTracerTest` — dropped, kept and unknown outcomes, per-span judgement, a span that never finished, and a `profiler_id` the SDK did not write.

## :pencil: Checklist

## :crystal_ball: Next steps

Known limits, left out deliberately:

- `PerfettoProfiler.start(long)` was public and is now package-private, since it takes the internal chunk record. The class is `@ApiStatus.Internal` and `@RequiresApi(35)`, but the method was on the published surface in 8.51.0–8.53.0.
- Only `SentryTracer` drops the id. OpenTelemetry spans take another path and are not covered. This has no effect today, as only the Android Perfetto profiler ever answers anything but `UNKNOWN`.
- Chunks that Sentry itself drops after a valid trace file was produced — a client-side rate limit, or cache eviction while offline — still read as recorded. The scope here is the OS reporting that no profile exists.
- The chunk history holds 10 chunks, so about 10 minutes. A transaction older than that reads as a transaction no chunk covers, and keeps its id.

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
@@ -23,6 +23,7 @@ import io.sentry.SentryNanotimeDate;
 import io.sentry.SentryOptions;
 import io.sentry.TracesSampler;
 import io.sentry.android.core.internal.util.SentryFrameMetricsCollector;
+import io.sentry.profiling.ProfileRecordingState;
 import io.sentry.protocol.SentryId;
 import io.sentry.transport.RateLimiter;
 import io.sentry.util.AutoClosableReentrantLock;
@@ -356,6 +357,18 @@ public class AndroidContinuousProfiler
   @Override
   public @NotNull SentryId getChunkId() {
     return chunkId;
+  }
+
+  /**
+   * This profiler does not track the outcome of its profiling requests, so the answer is always
+   * unknown.
+   */
+  @Override
+  public @NotNull ProfileRecordingState getProfileRecordingState(
+      final @NotNull SentryId profilerId,
+      final @NotNull SentryDate startTime,
+      final @NotNull SentryDate endTime) {
+    return ProfileRecordingState.UNKNOWN;
   }
 
   private void sendChunks(final @NotNull IScopes scopes, final @NotNull SentryOptions options) {
--- a/sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
@@ -82,8 +82,8 @@ public class PerfettoContinuousProfiler
    * so a handful of chunks (a minute each) is plenty.
    *
    * <p>The history is therefore assumed to be long enough for every window a span can ask about. A
-   * window whose chunks all fell out of it needs no special treatment, and is judged by the chunks
-   * that are left.
+   * window whose chunks all fell out of it needs no special treatment: it reads as a window no
+   * chunk covers, and keeps its profiler id.
    */
   @VisibleForTesting static final int MAX_CHUNK_HISTORY_SIZE = 10;
 
@@ -244,10 +244,7 @@ public class PerfettoContinuousProfiler
       final @NotNull SentryDate startTime,
       final @NotNull SentryDate endTime) {
     try (final @NotNull ISentryLifecycleToken ignored = chunkHistoryLock.acquire()) {
-      if (chunkHistory.isEmpty()) {
-        return ProfileRecordingState.UNKNOWN;
-      }
-
+      boolean hasFailedChunk = false;
       boolean hasUnknownChunk = false;
 
       for (final @NotNull ChunkRecord chunk : chunkHistory) {
@@ -260,6 +257,8 @@ public class PerfettoContinuousProfiler
         }
         if (state == ProfileRecordingState.UNKNOWN) {
           hasUnknownChunk = true;
+        } else {
+          hasFailedChunk = true;
         }
       }
 
@@ -268,8 +267,14 @@ public class PerfettoContinuousProfiler
         return ProfileRecordingState.UNKNOWN;
       }
 
-      // Every chunk overlapping the window failed, or no chunk ran during the window at all
-      return ProfileRecordingState.NOT_RECORDED;
+      // Every chunk covering the window failed
+      if (hasFailedChunk) {
+        return ProfileRecordingState.NOT_RECORDED;
+      }
+
+      // No chunk covers the window. It may have fallen out of the history, or a back-dated span may
+      // read as outside a chunk it really ran in, so the window gets the benefit of the doubt
+      return ProfileRecordingState.UNKNOWN;
     }
   }
 
--- a/sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
@@ -295,6 +295,12 @@ public class PerfettoContinuousProfiler
     }
   }
 
+  private void removeChunkRecord(final @NotNull ChunkRecord chunk) {
+    try (final @NotNull ISentryLifecycleToken ignored = chunkHistoryLock.acquire()) {
+      chunkHistory.remove(chunk);
+    }
+  }
+
   /**
    * Ends the running chunk, which is always the newest one, as a chunk only starts once the one
    * before it ended. Returns null if there is none, or if it already ended.
@@ -366,14 +372,19 @@ public class PerfettoContinuousProfiler
       profilerId = new SentryId();
     }
 
-    final @Nullable ChunkRecord chunkRecord =
-        perfettoProfiler.start(startProfileChunkTimestamp, profilerId, MAX_CHUNK_DURATION_MILLIS);
-    if (chunkRecord == null) {
+    // The chunk is known before profiling starts, so that a transaction finishing right after the
+    // request cannot miss the chunk it ran in and drop its profiler id
+    final @NotNull ChunkRecord chunkRecord =
+        new ChunkRecord(profilerId, startProfileChunkTimestamp);
+    addChunkRecord(chunkRecord);
+
+    if (!perfettoProfiler.start(chunkRecord, MAX_CHUNK_DURATION_MILLIS)) {
+      // No chunk ran, so the record is dropped rather than kept as a failure: it would take a slot
+      // of the history away from the chunks that did run
+      removeChunkRecord(chunkRecord);
       profilerId = SentryId.EMPTY_ID;
       chunkId = SentryId.EMPTY_ID;
-      logger.log(
-          SentryLevel.ERROR,
-          "Failed to start Perfetto profiling. PerfettoProfiler.start() returned no chunk.");
+      logger.log(SentryLevel.ERROR, "Failed to start Perfetto profiling.");
       return;
     }
 
@@ -382,7 +393,6 @@ public class PerfettoContinuousProfiler
       chunkId = new SentryId();
     }
 
-    addChunkRecord(chunkRecord);
     chunkMeasurements.start(performanceCollector, chunkId.toString());
 
     try {
--- a/sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
@@ -23,9 +23,11 @@ import io.sentry.SentryLevel;
 import io.sentry.SentryNanotimeDate;
 import io.sentry.SentryOptions;
 import io.sentry.TracesSampler;
+import io.sentry.android.core.internal.profiling.ChunkRecord;
 import io.sentry.android.core.internal.util.SentryFrameMetricsCollector;
 import io.sentry.profilemeasurements.ProfileMeasurement;
 import io.sentry.profilemeasurements.ProfileMeasurementValue;
+import io.sentry.profiling.ProfileRecordingState;
 import io.sentry.protocol.SentryId;
 import io.sentry.transport.RateLimiter;
 import io.sentry.util.AutoClosableReentrantLock;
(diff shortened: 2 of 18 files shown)


markushi:
@sentry review

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
```diff
@@ -96,6 +108,12 @@ public class PerfettoContinuousProfiler
 
   private final AutoClosableReentrantLock lock = new AutoClosableReentrantLock();
 
+  private final @NotNull ArrayDeque<ChunkRecord> chunkHistory =
+      new ArrayDeque<>(MAX_CHUNK_HISTORY_SIZE);
+  private @Nullable ChunkRecord currentChunk = null;
```
0xadam-brown: l: Can we just use `chunkHistory.peekLast()` instead?

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
```diff
@@ -213,6 +233,65 @@ public boolean isRunning() {
...
+        return ProfileRecordingState.UNKNOWN;
+      }
+
+      for (final @NotNull ChunkRecord chunk : chunkHistory) {
+        // A chunk that is still running, or that is still being collected, is assumed to be
+        // recorded in the end
+        if (chunk.getProfilerId().equals(profilerId)
+            && chunk.overlaps(startTime, endTime)
+            && chunk.getRecordingState() != ProfileRecordingState.NOT_RECORDED) {
+          return ProfileRecordingState.RECORDED;
```
0xadam-brown, line 256: m: Thoughts about updating our approach to match the following policy?

| Scenario | Returned value |                                                                                                                                                                                                                                                                                                                      
|---|---|                                                                                                                                                                                                                                                                                                                                               
| No chunks have ever run | `UNKNOWN` | 
| No recorded chunk overlaps window but one or more overlapping chunk remain undecided | `UNKNOWN` | 
| At least one overlapping chunk was recorded | `RECORDED` | 
| Overlapping chunks exist, but all failed | `NOT_RECORDED` |
| History exists, but no chunk for that profiler/window overlaps | `NOT_RECORDED` | 

We could implement it like this:

```java
 @Override                                                                                                                                                                                                                                                                                                                                               
     public @NotNull ProfileRecordingState getProfileRecordingState(                                                                                                                                                                                                                                                                                         
         final @NotNull SentryId profilerId,                                                                                                                                                                                                                                                                                                                 
         final @NotNull SentryDate startTime,                                                                                                                                                                                                                                                                                                                
         final @NotNull SentryDate endTime) {                                                                                                                                                                                                                                                                                                                
       try (final @NotNull ISentryLifecycleToken ignored = chunkHistoryLock.acquire()) {                                                                                                                                                                                                                                                                     
         if (chunkHistory.isEmpty()) {                                                                                                                                                                                                                                                                                                                       
           return ProfileRecordingState.UNKNOWN;                                                                                                                                                                                                                                                                                                             
         }                                                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                             
         boolean hasUnknownOverlappingChunk = false;                                                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                                                             
         for (final @NotNull ChunkRecord chunk : chunkHistory) {                                                                                                                                                                                                                                                                                             
           if (!chunk.getProfilerId().equals(profilerId) || !chunk.overlaps(startTime, endTime)) {                                                                                                                                                                                                                                                           
             continue;                                                                                                                                                                                                                                                                                                                                       
           }                                                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                             
           final @NotNull ProfileRecordingState state = chunk.getRecordingState();                                                                                                                                                                                                                                                                           
           if (state == ProfileRecordingState.RECORDED) {                                                                                                                                                                                                                                                                                                    
             return ProfileRecordingState.RECORDED;                                                                                                                                                                                                                                                                                                          
           }                                                                                                                                                                                                                                                                                                                                                 
           if (state == ProfileRecordingState.UNKNOWN) {                                                                                                                                                                                                                                                                                                     
             hasUnknownOverlappingChunk = true;                                                                                                                                                                                                                                                                                                              
           }                                                                                                                                                                                                                                                                                                                                                 
         }                  

         if (hasUnknownOverlappingChunk) {                                                                                                                                                                                                                                                                                                                   
           return ProfileRecordingState.UNKNOWN;                                                                                                                                                                                                                                                                                                             
         }                                                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                             
         return ProfileRecordingState.NOT_RECORDED;                                                                                                                                                                                                                                                                                                          
       }                                                                                                                                                                                                                                                                                                                                                     
     }                       
```

That'd let us avoid returning `RECORDED` in situations where we don't actually know yet. (Doesn't change current behavior b/c we only act on `NOT_RECORDED`, but it could matter going forward.)

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
```diff
@@ -265,23 +344,27 @@ private void startInternal() {
...
     }
-    if (!perfettoProfiler.start(MAX_CHUNK_DURATION_MILLIS)) {
+    if (SentryId.EMPTY_ID.equals(profilerId)) {
+      profilerId = new SentryId();
+    }
+
+    final @Nullable ChunkRecord chunkRecord =
+        perfettoProfiler.start(startProfileChunkTimestamp, profilerId, MAX_CHUNK_DURATION_MILLIS);
+    if (chunkRecord == null) {
+      profilerId = SentryId.EMPTY_ID;
```
0xadam-brown, line 390: m: What's the thinking behind updating the profilerId? 

Fwiw, I would've figured we'd keep the profilerId intact and let the ProfileRecordingState inform the outside world about the success vs failure of recording – but perhaps I'm missing something?

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoProfiler.java
```diff
@@ -152,6 +178,15 @@ private void onProfilingResult(final @NotNull ProfilingResult result) {
         result.getErrorCode(),
         result.getResultFilePath());
 
+    final @Nullable ChunkRecord record = chunkRecord;
+    if (record != null) {
+      final int errorCode = result.getErrorCode();
+      record.setRecordingState(
+          errorCode == ProfilingResult.ERROR_NONE
```
0xadam-brown: m: Should this be:

```java
if (record != null && result.getErrorCode() != ProfilingResult.ERROR_NONE) {                                                                                                                                                                                                                                                                          
         record.setRecordingState(ProfileRecordingState.NOT_RECORDED);                                                                                                                                                                                                                                                                                       
       }   
``` 

Worried that we're claiming the profile has been recorded because we haven't seen an error.

0xadam-brown (commented):
Nice! Much easier to follow, IMO. A few initial comments. Haven't checked for any threading issues. But the APIs are looking clean 💯

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
markushi: There's a public `getProfilerId()` method, which could be read by anyone (for Android it's just `SentryTracer` right now), so clearing it avoids any stale reads.

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
markushi: yeah, good one! 👀

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoProfiler.java
markushi: ah yeah, that's some dangling leftovers when moving from a Boolean to the enum, thanks for calling out!

sentry-android-core/src/main/java/io/sentry/android/core/internal/profiling/ChunkRecord.java
```diff
@@ -0,0 +1,72 @@
...
+
+  public @NotNull SentryId getProfilerId() {
+    return profilerId;
+  }
+
+  public @NotNull SentryDate getStartTimestamp() {
+    return startTimestamp;
+  }
+
+  public boolean overlaps(final @NotNull SentryDate startTime, final @NotNull SentryDate endTime) {
```
0xadam-brown, line 70: Worth a quick Javadoc reminding folks that this logic should be kept in sync with the rules we use on the backend to decide which chunks to display for a given transaction – at least until our hoped-for tombstone protocol shows up 😄.

sentry-android-core/src/main/java/io/sentry/android/core/internal/profiling/ChunkRecord.java
```diff
@@ -0,0 +1,72 @@
...
+    if (endTime.isBefore(startTimestamp)) {
+      return false;
+    }
+    // A chunk that is still running has no end yet, and covers everything from its start on
```
0xadam-brown:
```suggestion
    // A chunk that is still running has no end yet, and covers everything from its start onward
```

0xadam-brown: (Fwiw, at first I thought the sentence had been truncated.)

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
```diff
@@ -188,6 +205,8 @@ public void close(final boolean isTerminating) {
       if (isTerminating) {
         stopInternal(false);
         isClosed.set(true);
+        // sendChunk drops everything once isClosed is set, so the chunk that just ended is lost
+        markLastChunkNotRecordedIfUnknown();
```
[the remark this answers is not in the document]
0xadam-brown: Looks like this one's worth fixing. 

PerfettoContinuousProfiler.close(true) has two jobs: stop the current profiling chunk and mark the profiler as closed so no more chunks are sent. But those two decisions are ordered in a way that lets one callback observe the old state and another step observe the new state, meaning a chunk can be marked as recorded in history even though shutdown guarantees it will never be sent.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
```diff
@@ -358,6 +359,18 @@ public void close(final boolean isTerminating) {
...
+  /**
+   * This profiler does not track the outcome of its profiling requests, so the answer is always
+   * unknown.
+   */
+  @Override
+  public @NotNull ProfileRecordingState getProfileRecordingState(
+      final @NotNull SentryId profilerId,
+      final @NotNull SentryDate startTime,
+      final @NotNull SentryDate endTime) {
+    return ProfileRecordingState.UNKNOWN;
```
0xadam-brown, line 371: Thoughts about replacing with a default method on `IConinuousProfiler` that returns `UNKNOWN` (same throughout)?

0xadam-brown (approved):
A few minor comments + worth addressing [the SentryBot issue](https://github.com/getsentry/sentry-java/pull/6015/changes#r3879037552); otherwise lgtm

sentry-android-core/src/main/java/io/sentry/android/core/PerfettoContinuousProfiler.java
```diff
@@ -265,23 +362,27 @@ private void startInternal() {
     if (perfettoProfiler == null) {
       return;
     }
-    if (!perfettoProfiler.start(MAX_CHUNK_DURATION_MILLIS)) {
+    if (SentryId.EMPTY_ID.equals(profilerId)) {
+      profilerId = new SentryId();
+    }
+
+    final @Nullable ChunkRecord chunkRecord =
```
runningcode: there's a gap here between when we call `perfettoProfiler.start` and when we call `addChunkRecord` which means that if a transaction finishes between those two events, it won't check the newest chunkRecord since it hasn't been added yet.
my clanker suggests the best fix is to write the chunkRecord first and then remove it if it fails.

sentry-android-core/src/main/java/io/sentry/android/core/internal/profiling/ChunkRecord.java
```diff
@@ -0,0 +1,78 @@
...
+   * <p>Keep this in sync with the rules the backend uses to decide which chunks it shows for a
+   * transaction, for as long as there is no protocol for a chunk that produced nothing.
+   */
+  public boolean overlaps(final @NotNull SentryDate startTime, final @NotNull SentryDate endTime) {
+    if (endTime.isBefore(startTimestamp)) {
+      return false;
+    }
+    // A chunk that is still running has no end yet, and covers everything from its start onward
+    final @Nullable SentryDate endTimestamp = this.endTimestamp;
+    return endTimestamp == null || !startTime.isAfter(endTimestamp);
```
runningcode, line 76: App start spans are back-dated with a `SentryLongDate`, so `isBefore`/`isAfter` fall back to the wall clock here and we know Android's clock steps backwards a few seconds after boot (from the tests we did a few weeks ago). A skewed miss reads as `NOT_RECORDED` and strips `profiler_id` off the app start transaction. I would instead suggest that "no covering chunk" return `UNKNOWN` instead? This way the `profilder_id` isn't stripped off.

I really need to get my clock changes...

runningcode (approved):
Thanks for taking on this complicated fix! Looks good! I left a two comments, one with a locking issue and the other with a clock issue. Let me know if you'd like to discuss either.

It's also valid to say they aren't blockers or to do them in a follow up so I'll approve this for now.

sentry-android-core/src/main/java/io/sentry/android/core/AndroidContinuousProfiler.java
markushi: Sticking to the existing pattern here instead: every other `IContinuousProfiler` method is declared on the interface and implemented per profiler, so a default would be the odd one out.

PR #6028, opened 2026-08-31
feat(time): Add a monotonic clock abstraction (JAVA-571)

runningcode:

- this PR
- #6045
- #6029
- #6030
- #6032
- #6041
- #6043

---

## :scroll: Description

Adds a clock abstraction to `io.sentry.time`. This just adds the new APIs. They are never called. 

Here's a list of what is added (indents for class hierarchy).

```
MonotonicClock            long tickNanos()  — a nanosecond counter with an arbitrary origin,
                                              including time spent in deep sleep
JavaMonotonicClock        System.nanoTime()                    singleton
AndroidMonotonicClock     SystemClock.elapsedRealtimeNanos()   singleton
Deadline, Stopwatch       take a MonotonicClock
```

We also add `TestMonotonicClock` in `sentry-test-support` that advances by an amount *and a unit* to make testing easier.

To help review the new APIs with concrete use cases, of these new APIs that aren't called in this PR, I created draft PRs. Please don't review these yet, since I haven't fully reviewed it myself but here they are so you can see the new APIs:
- Usages of `Deadline`: https://github.com/getsentry/sentry-java/pull/6029 and https://github.com/getsentry/sentry-java/pull/6030/
- Usages of `Stopwatch`: https://github.com/getsentry/sentry-java/pull/6032

Everything is `@ApiStatus.Internal`, so nothing here is a public contract.

`ICurrentDateProvider.getCurrentTimeMillis()` has two implementations that mean different things: `CurrentDateProvider` returns wall time, `AndroidCurrentDateProvider` returns `SystemClock.uptimeMillis()`, which is monotonic and pauses in deep sleep. Every consumer hand-picks the one matching whatever it compares against; a wrong pairing compiles silently, and the tests inject fakes so nothing catches it.

Having one clock with a stated guarantee is the fix. `MonotonicClock` says in its javadoc that a tick never goes backwards, that its origin is arbitrary and may be negative, that only differences are meaningful, that it must never be serialized, and that it counts deep sleep. A site that needs a wall-clock instant reaches for `SentryDateProvider` instead, which is a different type returning a different thing.

`Deadline` and `Stopwatch` exist so callers never do arithmetic on raw ticks. Three decisions worth reviewing:

- `Deadline.passed()` gives "not populated yet" a representation outside the numeric range — `0` is a real and very recent instant on any boot-relative clock.
- `hasPassed()` subtracts rather than compares, so it holds for a negative or wrapping origin.
- `remaining()` rounds up, so a caller scheduling work for it never wakes to find the deadline still standing.

* resolves: JAVA-571 (partially — this is the API, not the migration)

Unit tests!

## :pencil: Checklist

## :crystal_ball: Next steps

This is the first of six pre-v9 PRs; the rest convert call sites whose current semantics are preserved:

1. **This PR** — the clock types and the options seam
2. Android connection-status cache → the clock added here (fixes JAVA-717, where `0` meant "unset" on a boot-relative clock)
3. `RateLimiter` → `MonotonicClock` (JAVA-574)
4. All four check-in duration paths → `MonotonicClock` (JAVA-576)
5. ANR thresholds → `MonotonicClock` (JAVA-579)
6. Deprecate `ICurrentDateProvider`, `DateUtils.getCurrentDateTime()` and `AndroidDateUtils` (JAVA-571)

Everything that changes a **serialized** measurement — span and transaction durations, session durations, app-start spans, replay and profiler timings — is deliberately held for v9.


sentry/src/main/java/io/sentry/time/MonotonicClock.java
```diff
@@ -0,0 +1,20 @@
+package io.sentry.time;
+
+import org.jetbrains.annotations.ApiStatus;
+
+/**
+ * A monotonically increasing nanosecond counter.
+ *
+ * <p>This type deliberately promises very little: a tick is a number that does not go backwards,
```
markushi, line 9: in theory the number could overflow, but practically that's not happening (e.g. on Android it would take 292 years of uptime)

sentry/src/main/java/io/sentry/time/UptimeClock.java
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * <p>This is the clock for measuring how long the CPU was actually available — most importantly ANR
+ * detection, where counting suspended time would report a responsive main thread as blocked.
+ *
+ * <p>On Android this is {@code CLOCK_MONOTONIC}, the same clock behind {@code
+ * SystemClock.uptimeMillis()}. On the JVM there is no comparable suspend state, so uptime and
+ * elapsed real time coincide.
+ */
+@ApiStatus.Internal
+public interface UptimeClock extends Ticker {}
```
markushi: To be honest I would remove this one for now.

sentry/src/main/java/io/sentry/time/Deadline.java
```diff
@@ -0,0 +1,79 @@
...
+
+  /**
+   * Whether this deadline falls after {@code other}.
+   *
+   * @throws IllegalArgumentException if the two were created from different clocks, whose origins
+   *     are unrelated and whose ticks are therefore not comparable.
+   */
+  public boolean isAfter(final @NotNull Deadline other) {
+    if (clock != other.clock) {
+      throw new IllegalArgumentException(
```
markushi, line 71: 💯

sentry/src/main/java/io/sentry/time/Ticker.java
```diff
@@ -0,0 +1,20 @@
...
+ * measured from an origin that is arbitrary and may be negative. Only <em>differences</em> between
+ * two ticks from the same instance are meaningful, and a tick must never be persisted, serialized,
+ * or compared against a value from another clock.
+ *
+ * <p>Do not implement or depend on {@code Ticker} directly. It exists so that {@link Deadline} and
+ * {@link Stopwatch} can be written once; callers declare {@link UptimeClock} or {@link
+ * ElapsedRealtimeClock}, whose names state which guarantee they provide.
+ */
+@ApiStatus.Internal
+public interface Ticker {
```
markushi: What's the motivation behind using both terms "Ticker" and "Clock"?

markushi: (I haven't looked at PRs on top of this one)
I guess we also need some mechanism to translate the ticks to unix time, right?

sentry/src/main/java/io/sentry/time/Stopwatch.java
```diff
@@ -0,0 +1,35 @@
...
+
+  private Stopwatch(final @NotNull Ticker clock) {
+    this.clock = clock;
+    this.startNanos = clock.tickNanos();
+  }
+
+  public static @NotNull Stopwatch started(final @NotNull Ticker clock) {
+    return new Stopwatch(clock);
+  }
+
```
markushi, line 27: maybe it could be useful to have a `reset()` method too, which sets startNanos to the current time

sentry-android-core/src/main/java/io/sentry/android/core/internal/time/AndroidMonotonicClock.java
```diff
@@ -0,0 +1,29 @@
...
+
+  public static @NotNull ElapsedRealtimeClock getInstance() {
+    return instance;
+  }
+
+  private AndroidElapsedRealtimeClock() {}
+
+  @Override
+  public long tickNanos() {
+    return SystemClock.elapsedRealtimeNanos();
```
markushi, line 27: I double checked: both `uptimeMillis()` and [`elapsedRealtimeNanos()`](https://android.googlesource.com/platform/frameworks/base/%2B/HEAD/core/java/android/os/SystemClock.java#280) are marked as `@ CriticalNative`, indicating they both should be fast ™️

markushi (commented):
Left some early feedback (which maybe are already resolved on follow up PRs)

sentry/src/main/java/io/sentry/time/Stopwatch.java
runningcode: the way it is designed you need to create a new stopwatch. this way it can only measure a single interval.
do you know of a use case for the reset method where creating a new stopwatch wouldn't make sense?

sentry/src/main/java/io/sentry/time/Ticker.java
runningcode: It is a very important point. Maybe this wasn't clear in the PR description but anything that extends `Ticker` should not be used for wall clocks or translating to wall clocks. So for translating to wall clocks I'm working on these APIs: https://github.com/getsentry/sentry-java/pull/6045 but that's not ready yet. Main issue is interop with existing APIs like SentryDateProvider and Spans etc.

runningcode: It comes from Guava. https://guava.dev/releases/27.1-jre/api/docs/com/google/common/base/Ticker.html
It represents a nanosecond counter with an arbitrary origin.
The `Clock` means that we have some sort of guarantee attached to the counter. For example if it includes time spent in deep sleep.
To be fair, most of the APIs are copied from Guava here.

sentry/src/main/java/io/sentry/time/UptimeClock.java
runningcode: Why is that? We will need it for ANRs.

runningcode: I've removed it. We now have a single type of Monotonic clock

sentry-android-core/src/main/java/io/sentry/android/core/internal/time/AndroidMonotonicClock.java
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * <p>That is {@code CLOCK_BOOTTIME}, so it keeps counting while the device is suspended — unlike
+ * {@link System#nanoTime()}, which the core module falls back to and which stops in deep sleep.
+ */
+@ApiStatus.Internal
+public final class AndroidMonotonicClock implements MonotonicClock {
+
+  private static final AndroidMonotonicClock instance = new AndroidMonotonicClock();
+
+  public static @NotNull MonotonicClock getInstance() {
```
0xadam-brown: Probably worth adding a test verifying that repeat calls return the same instance, as that's critical to correct operation, and you never know what an errant clanker might do.... 

(Same with JavaMonotonicClock.)

sentry/src/main/java/io/sentry/time/Deadline.java
```diff
@@ -0,0 +1,79 @@
...
+
+  /** A deadline {@code amount} of {@code unit} from now. */
+  public static @NotNull Deadline after(
+      final @NotNull MonotonicClock clock, final long amount, final @NotNull TimeUnit unit) {
+    return new Deadline(clock, clock.tickNanos() + unit.toNanos(amount));
+  }
+
+  /**
+   * A deadline that has already passed. Use for state that has not been populated yet, so that
+   * "never set" needs no numeric sentinel and cannot be mistaken for fresh — {@code 0} is a real
```
0xadam-brown: m: So is this basically an "empty" value for Deadline? (I'm having a little trouble reading between the lines of the clanker's prose, which seems to reflect an implementation that's no longer with us.) 

If so, should we just use a `Deadline.UNSET` singleton instead? I'm worried that `Deadline.passed()` makes `hasPassed()` ambiguous for everyone who uses it. (I can only be certain `hasPassed()` is meaningful if my deadline wasn't created via `passed()`, and that means tracing through code to find where deadlines are initialized.)

sentry/src/main/java/io/sentry/time/Deadline.java
```diff
@@ -0,0 +1,79 @@
...
+
+  private Deadline(final @NotNull MonotonicClock clock, final long deadlineNanos) {
+    this.clock = clock;
+    this.deadlineNanos = deadlineNanos;
+  }
+
+  /** A deadline {@code amount} of {@code unit} from now. */
+  public static @NotNull Deadline after(
+      final @NotNull MonotonicClock clock, final long amount, final @NotNull TimeUnit unit) {
+    return new Deadline(clock, clock.tickNanos() + unit.toNanos(amount));
```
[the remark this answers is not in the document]
0xadam-brown: Agree that this is worth protecting against.

sentry/src/main/java/io/sentry/time/Deadline.java
```diff
@@ -0,0 +1,79 @@
...
+  private final long deadlineNanos;
+
+  private Deadline(final @NotNull MonotonicClock clock, final long deadlineNanos) {
+    this.clock = clock;
+    this.deadlineNanos = deadlineNanos;
+  }
+
+  /** A deadline {@code amount} of {@code unit} from now. */
+  public static @NotNull Deadline after(
+      final @NotNull MonotonicClock clock, final long amount, final @NotNull TimeUnit unit) {
```
0xadam-brown: Do we want to prevent folks from passing negative amounts?

sentry/src/main/java/io/sentry/time/Stopwatch.java
```diff
@@ -0,0 +1,35 @@
...
+  private final @NotNull MonotonicClock clock;
+  private final long startNanos;
+
+  private Stopwatch(final @NotNull MonotonicClock clock) {
+    this.clock = clock;
+    this.startNanos = clock.tickNanos();
+  }
+
+  public static @NotNull Stopwatch started(final @NotNull MonotonicClock clock) {
```
0xadam-brown: `started()` -> `start()`? 

(The latter is what I personally expect based on Guava, etc. Up to you, though...)

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -3059,6 +3061,18 @@ public void setDateProvider(final @NotNull SentryDateProvider dateProvider) {
     this.dateProvider.setValue(dateProvider);
   }
 
+  /**
+   * Returns the clock used to measure elapsed time, such as rate-limit windows, cache expiry and
+   * ANR thresholds.
+   *
+   * <p>Android overrides this with a {@code SystemClock.elapsedRealtimeNanos()}-backed clock, which
```
0xadam-brown: l: Consider referencing SentryAndroidOptions explicitly, as that's where folks will need to go to get the elapsed realtime version.

sentry/src/main/java/io/sentry/time/MonotonicClock.java
```diff
@@ -0,0 +1,22 @@
...
+ * two ticks from the same instance are meaningful, and a tick must never be persisted, serialized,
+ * or compared against a value from another clock.
+ *
+ * <p>On Android this is {@code CLOCK_BOOTTIME}, via {@code SystemClock.elapsedRealtimeNanos()}, so
+ * an interval measured across a suspend reports the real time that passed rather than only the time
+ * the CPU was awake. On the JVM there is no comparable suspend state, so {@link System#nanoTime()}
+ * is equivalent.
+ */
+@ApiStatus.Internal
+public interface MonotonicClock {
```
0xadam-brown: m: Guava uses `Ticker` for this (as I know you know). Thoughts about `MonotonicTicker` to save the concept of "clocks" for the wall-clock / epochal variety? 

No super strong opinions, but worth a quick think b/c this API will be with us for a while. 

And personally, I do find this: 

```
Deadline.after(MonotonicTicker, 2, MINUTES)
```

more obvious than this:

```
Deadline.after(MonotonicClock, 2, MINUTES)
```

sentry/src/main/java/io/sentry/time/Deadline.java
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * A point in the future, measured on a {@link MonotonicClock}.
+ *
+ * <p>Exists so that callers never do arithmetic on raw ticks. A tick carries no unit and no epoch,
+ * so spelling out {@code now - then < ttl} at every call site is where unit mix-ups, sentinels that
+ * happen to mean "boot", and wrap-unsafe {@code <} comparisons come from. Each of those is decided
+ * once, here.
+ */
+@ApiStatus.Internal
+public final class Deadline {
```
0xadam-brown, line 16: m: Part of me loves the name Deadline, but part of me also strongly associates it with wall clocks and things that can be serialized. And this is definitely not an epochal class. 

Two possibilities if you feel similarly:

**[1] Rename `Deadline` -> `Cutoff` (or something that doesn't suggest calendars or dates) and keep the API as-is.**

Then we could have: 

```
Cutoff.after(MonotonicTicker, 2, MINUTES)
```

**[2] Rework the API as an `Interval` (or `TimeWindow`, or whatever we want to call it) instead.**

Our API could become: 

- Interval.lasting(monotonicTicker, 2, MINUTES)
- interval.hasEnded()
- interval.remaining(SECONDS)
- interval.endsAfter(otherInterval)

We could also fold Stopwatch into it if we wanted (but I'm less sure about that):

- interval.elapsed(SECONDS)

Personally, I like [1], with the option to go with [2] at some point if we find it makes more sense. But open to anything.

Thoughts?

sentry/src/test/java/io/sentry/time/DeadlineTest.kt
```diff
@@ -0,0 +1,93 @@
...
+
+    clock.advance(2, MINUTES)
+
+    assertTrue(deadline.hasPassed())
+  }
+
+  @Test
+  fun `a passed deadline is never fresh, even at tick zero`() {
+    // Regression guard: elapsedRealtimeNanos() starts at 0 on boot, so a numeric sentinel of 0
+    // would read as fresh for a whole TTL after every boot.
```
0xadam-brown: l: Doesn't `hasPassed()` return true if `clock.tickNanos() == `deadlineNanos`? So if 0 were the sentinel and we got 0 from `clock.tickNanos()` on boot, we'd return true from `hasPassed()`. (Not that the clanker doesn't have a point worth making, but seems like it may be confused on how to make it – or else I am??)

sentry/src/test/java/io/sentry/time/DeadlineTest.kt
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  @Test
+  fun `isAfter compares two deadlines`() {
+    val clock = TestMonotonicClock()
+    val shorter = Deadline.after(clock, 1, SECONDS)
+    val longer = Deadline.after(clock, 5, SECONDS)
+
+    assertTrue(longer.isAfter(shorter))
+    assertFalse(shorter.isAfter(longer))
```
0xadam-brown, line 82: We should also verify that `isAfter()` returns false when the two deadlines are equal.

sentry/src/test/java/io/sentry/time/DeadlineTest.kt
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  @Test
+  fun `isAfter rejects deadlines from different clocks`() {
+    val deadline = Deadline.after(TestMonotonicClock(), 1, SECONDS)
+    val fromAnotherClock = Deadline.after(TestMonotonicClock(), 5, SECONDS)
```
0xadam-brown: Even better if we compare deadlines that have the same duration (eg, make this 1 rather than 5).

sentry/src/test/java/io/sentry/time/DeadlineTest.kt
```diff
@@ -0,0 +1,93 @@
...
+
+    assertFailsWith<IllegalArgumentException> { deadline.isAfter(fromAnotherClock) }
+  }
+
+  @Test
+  fun `comparisons hold when the tick origin is negative`() {
```
0xadam-brown, line 95: Probably worth testing all three tick origin possibilities: negative, 0, positive.

0xadam-brown (commented):
Great to see this, and excellent all around 💯 . A few comments for your consideration – the biggest of which is around the naming + approach for Deadline, and the naming of MonotonicClock. 

Happy to discuss in real-time if that's easier.

sentry/src/main/java/io/sentry/time/MonotonicClock.java
0xadam-brown: ...looking at the next PR in this sequence, I like `MonotonicTicker` even more. Eg:

```
AnchoredClock(EpochClock, MonotonicTicker, ...)
```

sentry/src/main/java/io/sentry/time/Deadline.java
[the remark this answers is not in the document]
runningcode: It would take 292 years of uptime on Android as Markus wrote here: https://github.com/getsentry/sentry-java/pull/6028#discussion_r3913637906 so I don't think it is worth protecting against. Do you see another scenario where it is worth protecting against?

sentry-android-core/src/main/java/io/sentry/android/core/internal/time/AndroidMonotonicClock.java
runningcode: I've written a lot of singletons in my career but I've never written a test that they return the same instance. I'm not sure that the existence of AI changes that. 😄 

This is a standard java singleton and it isn't critical to correct operation as there is no state. What part of it being a singleton is critical to the operation?

sentry/src/main/java/io/sentry/time/Deadline.java
runningcode: Yeah, that's exactly the point. 😄 The deadline is always in the past. The use case is for the `AndroidConnectionStatusProvider` which had a bug where it would read an empty cache on startup as "fresh": https://github.com/getsentry/sentry-java/pull/6029/changes#diff-aadb2b267f52db6cae59f8995d3a5c08b63578296e8835f1ed689a94f76f80d0R94

To me `UNSET` is ambigous because it isn't clear if a deadline was never set and therefore it would be unclear if `hasPassed` is `true` or `false`. When we use `Deadline.passed()` it becomes very clear that `hasPassed = true`.

sentry/src/main/java/io/sentry/time/Deadline.java
runningcode: Thanks for the ideas.
I thought about it a bit and to me a Deadline isn't related to a wall clock/epoch time. For example, you can have a deadline in 2 hours or a deadline on 9 Sep at 13:00. A cutoff isn't related to time, you can have a frequency cut off, a height cut off, etc. And you can also say the cut off for submitting the assignments is in 2 hours or the cutoff for submitting the assigment is 9 Sep at 13:00. Either way, IMO Deadline is the right word choice here.

To add to that, if you look at the codebase, we already have a few places that measure deadlines in terms of milliseconds like `TransactionOptions.DEFAULT_DEADLINE_TIMEOUT_AUTO_TRANSACTION` and `SentryTracer.scheduleDeadlineTimeout()`
 
For 2, an interval doesn't make sense here. For example in the connection cache, we just want to know whether or not the connection cache is expired. On startup, we need to assume it is past the cache's deadline.

sentry/src/main/java/io/sentry/time/Deadline.java
runningcode: i can add it

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -3059,6 +3061,18 @@ public void setDateProvider(final @NotNull SentryDateProvider dateProvider) {
...
+   * this module cannot reference. On the JVM there is no suspend state to account for.
+   */
+  @ApiStatus.Internal
+  public @NotNull MonotonicClock getMonotonicClock() {
```
0xadam-brown: m: Looking at the next PR in this sequence, do we even want to expose MonotonicClock via SentryOptions? Seems like we should direct folks to AnchoredClock instead.

sentry/src/test/java/io/sentry/time/DeadlineTest.kt
runningcode: I'm not sure I follow your point. It is correct if the deadline were `0` and the clock is currently `0`, we should return `hasPassed() = true`.
I can make the comment more clear if that's what you are asking?

sentry/src/test/java/io/sentry/time/DeadlineTest.kt
runningcode: Well the positive is tested on line 14,24, and 66.
And the 0 case is tested on line 34. did you mean something else?

sentry/src/test/java/io/sentry/time/DeadlineTest.kt
runningcode: Isn't that the same as your suggestion for line 72? Maybe i misunderstood.

sentry/src/main/java/io/sentry/time/Deadline.java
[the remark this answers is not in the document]
0xadam-brown: Ha! Well, shows me. Nvm then 😄

sentry-android-core/src/main/java/io/sentry/android/core/internal/time/AndroidMonotonicClock.java
0xadam-brown: Nope - my bad. We always call through to the system clock, so it doesn't matter for correctness, and so obviously not worth protecting against regressions.

sentry/src/main/java/io/sentry/time/MonotonicClock.java
runningcode: Makes sense. I can make this change.

sentry/src/main/java/io/sentry/time/Deadline.java
0xadam-brown: I suspect it's the Javadoc that's throwing me off. You have a concrete use case where being able to set a Deadline that immediately returns true for hasPassed() makes sense. That's reasonable. 

My vote would be just to drop the "Use for state that has..." part of the Javadoc, as it makes it sound like this is a general purpose sentinel value that any user of the Deadline API may need to worry if they're provided a deadline that they didn't create.

sentry/src/test/java/io/sentry/time/DeadlineTest.kt
0xadam-brown: > It is correct if the deadline were 0 and the clock is currently 0, we should return hasPassed() = true.

AFAICT, the comment says the opposite – which is why I was confused. Disregard if I'm misreading that. (I'm interpreting "fresh" as "has not passed", and maybe that's wrong.)

sentry/src/test/java/io/sentry/time/DeadlineTest.kt
0xadam-brown: Close. Here I meant: 

```kotlin
 val deadline = Deadline.after(TestMonotonicClock(), 1, SECONDS)
 val fromAnotherClock = Deadline.after(TestMonotonicClock(), 1, SECONDS)
```

On line 72 I meant:

```kotlin
assertTrue(longer.isAfter(shorter))
assertFalse(timeA.isAfter(timeB)) // where timeA and timeB have the same duration.
assertFalse(shorter.isAfter(longer))
```

The first is about conditions Deadline uses to determine whether the clocks are different (and having everything be the same save for the instance is a stronger test).

sentry/src/test/java/io/sentry/time/DeadlineTest.kt
0xadam-brown: Nw if we've covered the bases already – just read the test name and commented based on that.

sentry/src/main/java/io/sentry/SentryOptions.java
runningcode: Which PR do you mean? There's the PR to add the `AnchoredClock` but we still need a `MonotonicClock` to construct a `Deadline`and a `Stopwatch`.
Up for debate is whether we want to change `Deadline` to not require passing in a `MonotonicClock` via the static constructor which would then allow this. But then we'd need to introduce a third class like `DeadlineFactory` which is stateless. Happy to debate this.

sentry/src/main/java/io/sentry/time/Stopwatch.java
runningcode: What do you mean? I would expect the former if we are comparing this to `Guava` which is where the naming came from. This is the same pattern as Guava's `createStarted()`. https://guava.dev/releases/23.0/api/docs/com/google/common/base/Stopwatch.html#createStarted--

sentry/src/main/java/io/sentry/time/Deadline.java
runningcode: Good point. I'll clean up that javadoc. It was written based on that narrow use case and thanks for pointing out how confusing it can be to read!

sentry/src/main/java/io/sentry/SentryOptions.java
0xadam-brown: Just asking about the level of abstraction we want to advertise / make easily available. Happy to defer to you, as you know where all of these PRs are ultimately headed. Just wanted to make sure we were thinking through what we were exposing via SentryOptions in particular, as protecting folks from reaching for low-level APIs that they could possibly misuse would be a win.

sentry/src/main/java/io/sentry/time/Stopwatch.java
0xadam-brown: This is just me reading portions of PRs faster than I should, and needing to get my Java chops back. 

Seems I'm used to relying on constructors, companion objects, or `create`-type langauge as signals that I'm looking at a factory method. But this is static and returns a `Stopwatch`, and you're definitely right: it matches Guava's `*Started()`. I recant!

sentry/src/main/java/io/sentry/time/Deadline.java
0xadam-brown: Deadline is okay by me 👍

sentry/src/main/java/io/sentry/SentryOptions.java
runningcode: Thanks, we need some way to get it. Either way someone could call `getInstance()` on any of the clock instances. This abstraction at least gives you the correct one depending on which platform you are on.

sentry/src/test/java/io/sentry/time/DeadlineTest.kt
runningcode: done

sentry/src/test/java/io/sentry/time/DeadlineTest.kt
runningcode: comment updated

sentry/src/main/java/io/sentry/time/Deadline.java
runningcode: updated

sentry/src/main/java/io/sentry/time/MonotonicClock.java
runningcode: done

sentry/src/test/java/io/sentry/time/DeadlineTest.kt
runningcode: done

0xadam-brown (approved):

PR #6029, opened 2026-08-31
fix(android): Treat an unpopulated connection cache as stale (JAVA-717)

runningcode:

- #6028
- #6045
- this PR
- #6030
- #6032
- #6041
- #6043

---

## :scroll: Description

`AndroidConnectionStatusProvider` used `0` to mean "cache never populated", and compared it against `SystemClock.uptimeMillis()`:

```java
private volatile long lastCacheUpdateTime = 0;
private static final long CACHE_TTL_MS = 2 * 60 * 1000L;

private boolean isCacheValid() {
  return (timeProvider.getCurrentTimeMillis() - lastCacheUpdateTime) < CACHE_TTL_MS;
}
```

`uptimeMillis()` is 0 at boot, so `0` does not mean "unset" — it means "populated at boot". For the first two minutes of every boot, an empty cache reads as fresh.

The cache now holds a `Deadline`, so "never populated" is expired by construction and has no numeric value to get wrong.

Stacked on #6028, which adds the clock types this uses.

* resolves: JAVA-717

A regression test asserts that a provider constructed at tick 0 populates the cache before reading it. It fails if the deadline is constructed fresh instead of passed, which is the shape of the original bug.

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/internal/util/AndroidConnectionStatusProvider.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/internal/util/AndroidConnectionStatusProvider.java
@@ -19,10 +19,12 @@ import io.sentry.SentryOptions;
 import io.sentry.android.core.AppState;
 import io.sentry.android.core.BuildInfoProvider;
 import io.sentry.android.core.ContextUtils;
-import io.sentry.transport.ICurrentDateProvider;
+import io.sentry.time.Deadline;
+import io.sentry.time.MonotonicTicker;
 import io.sentry.util.AutoClosableReentrantLock;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
@@ -41,7 +43,7 @@ public final class AndroidConnectionStatusProvider
   private final @NotNull Context context;
   private final @NotNull SentryOptions options;
   private final @NotNull BuildInfoProvider buildInfoProvider;
-  private final @NotNull ICurrentDateProvider timeProvider;
+  private final @NotNull MonotonicTicker ticker;
   private final @NotNull List<IConnectionStatusObserver> connectionStatusObservers;
   private final @Nullable Handler handler;
   private final @NotNull AutoClosableReentrantLock lock = new AutoClosableReentrantLock();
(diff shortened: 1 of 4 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/internal/util/AndroidConnectionStatusProvider.java
```diff
@@ -231,7 +234,7 @@ private void clearCacheAndNotifyObservers() {
               try (final @NotNull ISentryLifecycleToken ignored = lock.acquire()) {
                 cachedNetworkCapabilities = null;
                 currentNetwork = null;
-                lastCacheUpdateTime = timeProvider.getCurrentTimeMillis();
+                cacheFreshUntil = Deadline.after(ticker, CACHE_TTL_MINUTES, TimeUnit.MINUTES);
```
0xadam-brown, line 237: Seeing this in action, I wonder if there isn't a better name than Deadline.after()? ("after" makes me think that perhaps we're offsetting the start of the deadline until the time we pass to after())

Thoughts about `until()`? I prefer it b/c my mind defaults to thinking that whatever I'm returned is "valid"; a deadline that's passed isn't; and `until()` points to the valid time segment. 

(I also considered `at()`, which reads nicely here, but (I think) mostly because my mind is sneaking in the idea of a clock, and not reading this as a pure duration.)

Happy to defer to you...

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/AndroidConnectionStatusProvider.java
```diff
@@ -400,13 +403,13 @@ private void updateCache(@Nullable NetworkCapabilities networkCapabilities) {
...
         cachedNetworkCapabilities = null;
-        lastCacheUpdateTime = timeProvider.getCurrentTimeMillis();
+        cacheFreshUntil = Deadline.after(ticker, CACHE_TTL_MINUTES, TimeUnit.MINUTES);
       }
     }
   }
 
   private boolean isCacheValid() {
-    return (timeProvider.getCurrentTimeMillis() - lastCacheUpdateTime) < CACHE_TTL_MS;
+    return !cacheFreshUntil.hasPassed();
```
0xadam-brown, line 412: l: We could add a hasNotPassed() method to Deadline, as well. I bet it'll get a lot of use.

0xadam-brown (approved):
A couple quick comments if you want them; no blockers.

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/AndroidConnectionStatusProvider.java
runningcode: Its a good point that there are a lot of callers that invert the condition. I don't like methods with negative words in the name. When you invert the condition it becomes harder to reason about.

sentry-android-core/src/main/java/io/sentry/android/core/internal/util/AndroidConnectionStatusProvider.java
runningcode: I originally had it as `in` but that's a reserved keyword in kotlin so it was an awkward API once we get to the kotlin code. I don't see a big enough difference between `until` and `after`. My reasoning with `after` is that the deadline expires after the specified time.

We can change it later, but I'd rather start getting some PRs merged. Rebasing is getting tedious with all the stacked PRs.

PR #6030, opened 2026-08-31
ref(transport): Measure rate-limit backoff on a monotonic clock (JAVA-574)

runningcode:
## :scroll: Description

I wrote all these lines by hand in case you care to read them. :)

### Wall clocks are bad

So this PR started as a way to fix the rate limiter using a wall clock. That's clearly bad™️ because a wall clock can drift which means if the clock jumps backward or forward suddenly we could send a ton of requests when the ratelimiter expires or extend the rate limiter for far longer.

I think the code is a LOT easier to read and speaks for itself.

### Static constructors are good
With that change, we need to change the object's constructor because we don't pass in a `ICurrentDateProvider` anymore. So the old constructors are kept (and deprecated) only to maintain backward compatibility with any possible existing customers. We should remove the constructors in v9. We also use a static constructor because Josh Bloch is my man. (See Effective Java item 1 "Consider static factory methods instead of constructors.")

### SentryOptions is a god object
Since we are changing the constructor, I decided it would be a good idea to NOT pass in the `SentryOptions` god object and instead hide it behind an interface that only gives us what we need from that object. There's no functionality change but it makes it easier to test and confines the scope our `SentryOptions` usage to make future refactors easier.

#### Another route instead of the `RateLimiterConfig`
Another option would be to only pass in the objects from `SentryOptions` that we need to the RateLimiter's constructor. I really wanted to do this. The issue is that all the fields are publicly mutable so we can't pass them in to the constructor because they can change. 😢 Mutable state is very fun /s

So TL;DR on what's changed:
1. Use monotonic clock to determine rate limit
2. Use static constructor
3. RateLimiterConfig interface instead of SentryOptions.

A wall clock is bad for RateLimiting since it can jump backwards.

* resolves: JAVA-574

Tests pass.

## :pencil: Checklist

### DIFF
--- a/sentry/src/main/java/io/sentry/SentryOptions.java
+++ b/sentry/src/main/java/io/sentry/SentryOptions.java
@@ -29,6 +29,7 @@ import io.sentry.transport.ITransport;
 import io.sentry.transport.ITransportGate;
 import io.sentry.transport.NoOpEnvelopeCache;
 import io.sentry.transport.NoOpTransportGate;
+import io.sentry.transport.RateLimiterConfig;
 import io.sentry.util.AutoClosableReentrantLock;
 import io.sentry.util.LazyEvaluator;
 import io.sentry.util.LoadClass;
@@ -58,7 +59,7 @@ import org.jetbrains.annotations.TestOnly;
 
 /** Sentry SDK options */
 @Open
-public class SentryOptions {
+public class SentryOptions implements RateLimiterConfig {
 
   @ApiStatus.Internal public static final @NotNull String DEFAULT_PROPAGATION_TARGETS = ".*";
 
@@ -847,6 +848,7 @@ public class SentryOptions {
    *
    * @return the logger
    */
+  @Override
   public @NotNull ILogger getLogger() {
     return logger;
   }
@@ -1610,6 +1612,7 @@ public class SentryOptions {
    * @return the timer executor service
    */
   @ApiStatus.Internal
+  @Override
   @NotNull
   public ISentryExecutorService getTimerExecutorService() {
     return timerExecutorService;
@@ -2601,6 +2604,7 @@ public class SentryOptions {
    * @return a client report recorder or NoOp
    */
   @ApiStatus.Internal
+  @Override
   public @NotNull IClientReportRecorder getClientReportRecorder() {
     return clientReportRecorder;
   }
--- a/sentry/src/main/java/io/sentry/time/Deadline.java
+++ b/sentry/src/main/java/io/sentry/time/Deadline.java
@@ -24,15 +24,18 @@ public final class Deadline {
   }
 
   /**
-   * A deadline {@code amount} of {@code unit} from now.
+   * A deadline {@code amount} of {@code unit} from now, or one that has already {@link #passed} if
+   * {@code amount} is negative.
    *
-   * @throws IllegalArgumentException if {@code amount} is negative. A deadline that starts out in
-   *     the past is a sign error at the call site; {@link #passed} says it deliberately.
+   * <p>Negative amounts are tolerated because callers pass durations parsed from server headers,
+   * where a bogus value must degrade to "no wait" rather than throw out of response handling.
+   * Clamping rather than adding a negative offset also keeps the tick arithmetic away from
+   * wrapping.
    */
   public static @NotNull Deadline after(
       final @NotNull MonotonicTicker ticker, final long amount, final @NotNull TimeUnit unit) {
     if (amount < 0) {
-      throw new IllegalArgumentException("Deadline amount must not be negative, but was " + amount);
+      return passed(ticker);
     }
     return new Deadline(ticker, ticker.tickNanos() + unit.toNanos(amount));
   }
--- a/sentry/src/main/java/io/sentry/transport/RateLimiter.java
+++ b/sentry/src/main/java/io/sentry/transport/RateLimiter.java
@@ -14,6 +14,8 @@ import io.sentry.clientreport.DiscardReason;
 import io.sentry.hints.DiskFlushNotification;
 import io.sentry.hints.Retryable;
 import io.sentry.hints.SubmissionResult;
+import io.sentry.time.Deadline;
+import io.sentry.time.MonotonicTicker;
 import io.sentry.util.AutoClosableReentrantLock;
 import io.sentry.util.HintUtils;
 import io.sentry.util.StringUtils;
@@ -22,7 +24,6 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
-import java.util.Date;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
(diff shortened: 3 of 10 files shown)


sentry/src/main/java/io/sentry/transport/RateLimiter.java
```diff
@@ -354,7 +357,7 @@ private long parseRetryAfterOrDefault(final @Nullable String retryAfterHeader) {
         // let's use the default then
       }
     }
-    return retryAfterMillis;
+    return Deadline.after(ticker, retryAfterMillis, TimeUnit.MILLISECONDS);
```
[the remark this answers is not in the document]
runningcode: Ah dang. we should just remove that exception then.

sentry/src/main/java/io/sentry/transport/RateLimiter.java
```diff
@@ -354,7 +357,7 @@ private long parseRetryAfterOrDefault(final @Nullable String retryAfterHeader) {
...
   }
 
   private void notifyRateLimitObservers() {
```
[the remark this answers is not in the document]
runningcode: yeah, i will remove that exception

sentry/src/main/java/io/sentry/time/Deadline.java
```diff
@@ -24,15 +24,18 @@ private Deadline(final @NotNull MonotonicTicker ticker, final long deadlineNanos
   }
 
   /**
-   * A deadline {@code amount} of {@code unit} from now.
+   * A deadline {@code amount} of {@code unit} from now, or one that has already {@link #passed} if
+   * {@code amount} is negative.
    *
-   * @throws IllegalArgumentException if {@code amount} is negative. A deadline that starts out in
-   *     the past is a sign error at the call site; {@link #passed} says it deliberately.
+   * <p>Negative amounts are tolerated because callers pass durations parsed from server headers,
```
0xadam-brown, line 30: :swish:

sentry/src/main/java/io/sentry/time/Deadline.java
```diff
@@ -24,15 +24,18 @@ private Deadline(final @NotNull MonotonicTicker ticker, final long deadlineNanos
...
+   * where a bogus value must degrade to "no wait" rather than throw out of response handling.
+   * Clamping rather than adding a negative offset also keeps the tick arithmetic away from
+   * wrapping.
    */
   public static @NotNull Deadline after(
       final @NotNull MonotonicTicker ticker, final long amount, final @NotNull TimeUnit unit) {
     if (amount < 0) {
-      throw new IllegalArgumentException("Deadline amount must not be negative, but was " + amount);
+      return passed(ticker);
```
0xadam-brown, line 38: Good moves here.

sentry/src/main/java/io/sentry/transport/RateLimiter.java
```diff
@@ -22,40 +24,63 @@
...
+    this.config = config;
   }
 
+  public static @NotNull RateLimiter create(
+      final @NotNull MonotonicTicker ticker, final @NotNull RateLimiterConfig config) {
+    return new RateLimiter(ticker, config);
+  }
+
+  /**
+   * @deprecated use the create static constructor instead
```
0xadam-brown: nit: Could add a link to make this foolproof if you want (same below)

sentry/src/main/java/io/sentry/transport/RateLimiter.java
```diff
@@ -22,40 +24,63 @@
...
+   */
+  @Deprecated
```
0xadam-brown: ⚠️  Can we do this given that MonotonicClock is still internal? Some folks might fail compilation on warnings...

(Same below)

sentry/src/main/java/io/sentry/transport/RateLimiter.java
```diff
@@ -107,16 +132,11 @@ public RateLimiter(final @NotNull SentryOptions options) {
...
-    final Date currentDate = new Date(currentDateProvider.getCurrentTimeMillis());
-
     // check all categories
-    final Date dateAllCategories = sentryRetryAfterLimit.get(DataCategory.All);
-    if (dateAllCategories != null) {
-      if (!currentDate.after(dateAllCategories)) {
-        return true;
-      }
+    final @Nullable Deadline allCategories = sentryRetryAfterLimit.get(DataCategory.All);
+    if (allCategories != null && !allCategories.hasPassed()) {
```
0xadam-brown, line 147: (future aside: still thinking `hasNotPassed()` would be nice at some point 😄... )

sentry/src/main/java/io/sentry/transport/RateLimiterConfig.java
```diff
@@ -0,0 +1,29 @@
+package io.sentry.transport;
+
+import io.sentry.ILogger;
+import io.sentry.ISentryExecutorService;
+import io.sentry.clientreport.IClientReportRecorder;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * The configuration {@link RateLimiter} reads. Declared next to its consumer rather than alongside
```
0xadam-brown, line 10: l: Consider moving a generalized version of the "Declared next..." comment into SentryOptions where we declare its implementation of these interfaces, since that'll be a pattern we want to continue (we could then remove it here).

sentry/src/main/java/io/sentry/transport/RateLimiterConfig.java
```diff
@@ -0,0 +1,29 @@
...
+ * the implementation, so that the collaborators a rate limiter actually touches are three lines to
+ * read instead of three hundred, and a test can supply them without building a {@link
+ * io.sentry.SentryOptions}.
+ *
+ * <p>Implementations are expected to delegate to live configuration rather than snapshot it, so
+ * that a logger or executor replaced after {@code Sentry.init} is still picked up.
+ */
+@ApiStatus.Internal
```
0xadam-brown, line 18: 👍

sentry/src/main/java/io/sentry/transport/RateLimiterConfig.java
```diff
@@ -0,0 +1,29 @@
...
+public interface RateLimiterConfig {
+
+  @NotNull
+  ILogger getLogger();
```
0xadam-brown, line 22: future aside: maybe we'll want to hoist some of these want-them-everywhere deps into their own interface at some point and let RateLimiterConfig, etc. implement that.

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -58,7 +59,7 @@
 
 /** Sentry SDK options */
 @Open
-public class SentryOptions {
+public class SentryOptions implements RateLimiterConfig {
```
0xadam-brown, line 62: The many prices paid for service locators... sigh. (But I like this move given that we have one!)

0xadam-brown (commented):
So good! Just [one potential blocker](https://github.com/getsentry/sentry-java/pull/6030#discussion_r3970777089); otherwise looks fantastic 💯 

> A wall clock is bad for RateLimiting since it can jump backwards.

:poetry:

> We also use a static constructor because Josh Bloch is my man. (See Effective Java item 1 "Consider static factory methods instead of constructors.")

🐐 🏅

0xadam-brown (approved):
Thanks for the update 💯 🕐 🚀

PR #6031, opened 2026-08-31
test(android): Fix flaky AnrV2 scope persistence assertion

runningcode:
## :scroll: Description

The final assertions of `SentryAndroidTest."AnrV2 events get enriched with previously persisted scope and options data, the new data gets persisted after that"` read `transaction.json` and `release.json` straight off disk. Both are written from the Sentry executor, and nothing in the test joins it. Poll for the persisted values instead of reading them once.

The test is flaky on CI. https://scans.gradle.com/s/x5fecocehp2i4/tests/task/:sentry-android-core:testReleaseUnitTest/details/io.sentry.android.core.SentryAndroidTest/AnrV2%20events%20get%20enriched%20with%20previously%20persisted%20scope%20and%20options%20data%2C%20the%20new%20data%20gets%20persisted%20after%20that?focused-exception-line=0-0&page=eyJvdXRwdXQiOnsiMCI6Mn19&top-execution=1

Ran `:sentry-android-core:testReleaseUnitTest --tests io.sentry.android.core.SentryAndroidTest` — passes, with no change to the class's wall-clock time.

To confirm the polling actually engages rather than the assertion just passing on the first read, I temporarily changed the expected value to one never written. The test then polled until the timeout and reported `was:<[Test]Activity>`, showing both that the write does land and that the retry loop is active.

## :pencil: Checklist

## :crystal_ball: Next steps

None.

#skip-changelog

### DIFF
--- a/sentry-android-core/src/test/java/io/sentry/android/core/SentryAndroidTest.kt
+++ b/sentry-android-core/src/test/java/io/sentry/android/core/SentryAndroidTest.kt
@@ -464,7 +464,7 @@ class SentryAndroidTest {
     // the reset assertions run on the executor, so nothing reports their AssertionError unless we
     // observe the Future: without this the test would pass no matter what they found
     try {
-      resetAssertions.get(30, TimeUnit.SECONDS)
+      resetAssertions.get(2, TimeUnit.SECONDS)
     } catch (e: ExecutionException) {
       throw requireNotNull(e.cause)
     }
--- a/sentry-android-core/src/test/java/io/sentry/android/core/SentryAndroidTest.kt
+++ b/sentry-android-core/src/test/java/io/sentry/android/core/SentryAndroidTest.kt
@@ -464,17 +464,23 @@ class SentryAndroidTest {
     // Execute all posted tasks
     Shadows.shadowOf(Looper.getMainLooper()).idle()
 
-    // assert that persisted values have changed
-    assertEquals(
-      "TestActivity",
-      options
-        .findPersistingScopeObserver()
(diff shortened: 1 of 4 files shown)


sentry-android-core/src/test/java/io/sentry/android/core/SentryAndroidTest.kt
```diff
@@ -464,17 +464,24 @@ class SentryAndroidTest {
...
-        ?.read(options, TRANSACTION_FILENAME, String::class.java),
-    )
-    assertEquals(
-      "io.sentry.sample@1.1.0+220",
-      PersistingOptionsObserver.read(options, RELEASE_FILENAME, String::class.java),
-    )
+    // Both values are persisted from the Sentry executor, and nothing above joins it: the
+    // beforeSend callback we awaited runs on an earlier task in the same queue, so it flipping
+    // says nothing about the writes. Poll instead of reading once and racing them.
+    await
```
0xadam-brown, line 474: If we wanted something more deterministic than waiting, we could pass a no-op marker to options.executorService, wait for its Future to return, and only then do the final disk assertions. 

(Assumes we keep both persistence paths on the same executor going forward, of course, but that's not unreasonable.)

Up to you...

0xadam-brown (approved):
A possible alt approach if you want it; looks good regardless 👍

sentry-android-core/src/test/java/io/sentry/android/core/SentryAndroidTest.kt
runningcode: That's a great point. I didn't add submitting a no-op marker but I added joining on the executor!

PR #6041, opened 2026-09-02
ref(android): Measure ANR thresholds on the monotonic clock (JAVA-579)

runningcode:
## :scroll: Description

Moves the ANR integrations to use the `MonotonicTicker` abstraction. This also makes them more testable since we aren't hard coding a direct access to `SystemClock`.

A note: We previously used `SystemClock.uptimeMillis` which does not include time spent in sleep. The implementation of `MonotonicTicker` on Android uses `SystemClock.elapsedRealtimeNanos()`. Since we aren't using these to measure background ANRs, nothing changes.

Also `ANRWatchdog` was previously using a `ICurrentDateProvider`. Which has two different implementations, a wall clock and a monotonic clock. It happens to work by chance on Android because on Android it should always return a monotonic clock but this is quite brittle.

Fixing clocks and making the code more sane.
* resolves: JAVA-579

`./gradlew :sentry-android-core:testReleaseUnitTest :sentry-android-core:apiCheck` — green.

## :pencil: Checklist

### DIFF
--- a/sentry-android-core/src/main/java/io/sentry/android/core/ANRWatchDog.java
+++ b/sentry-android-core/src/main/java/io/sentry/android/core/ANRWatchDog.java
@@ -30,11 +30,13 @@ import static android.app.ActivityManager.ProcessErrorStateInfo.NOT_RESPONDING;
 import android.app.ActivityManager;
 import android.content.Context;
 import android.os.Debug;
-import android.os.SystemClock;
 import io.sentry.ILogger;
 import io.sentry.SentryLevel;
-import io.sentry.transport.ICurrentDateProvider;
+import io.sentry.android.core.internal.time.AndroidMonotonicTicker;
+import io.sentry.time.Deadline;
+import io.sentry.time.MonotonicTicker;
 import java.util.List;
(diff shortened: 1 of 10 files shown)


sentry-android-core/src/main/java/io/sentry/android/core/ANRWatchDog.java
```diff
@@ -30,60 +30,60 @@
...
 /** A watchdog timer thread that detects when the UI thread has frozen. */
 @SuppressWarnings("UnusedReturnValue")
 final class ANRWatchDog extends Thread {
 
+  private static final long DEFAULT_POLLING_INTERVAL_MS = 500;
+
   private final boolean reportInDebug;
   private final ANRListener anrListener;
   private final MainLooperHandler uiHandler;
-  private final ICurrentDateProvider timeProvider;
```
runningcode, line 49: This returns EITHER a monotonic clock or a wall clock depending on the platform. This is a bad abstraction.
On android it gives you a monotonic clock which is what we want for this class, but the fact that it has `Date` in the name but gives you a monotonic clock is quite confusing.

PR #6032, opened 2026-08-31
ref(checkin): Measure check-in durations with Stopwatch (JAVA-576)

runningcode:
## :scroll: Description

Switch CheckIns to use Stopwatch to measure durations. This makes the code easier to reason about and makes the code more testable. This has no logical changes since this is only used in Java and the monotonic clock uses the same `System.nanoTime()` under the hood.

## :bulb: Motivation and Context

This is a pure refactor to make the code easier to reason about and easier to test.

* resolves: JAVA-576

## :green_heart: How did you test it?

Existing tests pass and added new ones.

## :pencil: Checklist

- [x] I added GH Issue ID _&_ Linear ID
- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [ ] Review from the native team if needed.
- [x] No breaking change or entry added to the changelog.
- [x] No breaking change for hybrid SDKs or communicated to hybrid SDKs.
- [x] Public API changes reviewed by another Mobile SDK team member or implemented according to the [develop docs](https://develop.sentry.dev/) spec.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6045, opened 2026-09-02
feat(time): Add Timestamp, EpochClock and AnchoredClock (JAVA-572)

runningcode:

- #6028
- this PR
- #6029
- #6030
- #6032
- #6041
- #6043

---

## :scroll: Description

Adds the wall-clock half of the time API — `Timestamp`, `EpochClock` and `AnchoredClock` — on top of
the `MonotonicClock`/`Stopwatch` primitives from #6028. **Nothing calls any of it yet**, like #6028.

```java
Timestamp      // an epoch instant, plus the anchor that projected it (or null if read directly).
               // No arithmetic between instants; equality is by instant.
EpochClock     // now() stamps a moment that leaves the process. Cannot report a duration.
AnchoredClock  // one epoch reading pinned to one tick. now()/at(tick) project, tickOf() inverts
               // exactly, driftNanos() reports how far the projection trails the wall clock.
```

`SentryOptions.getEpochClock()` is the injection point.

### Why anchoring, rather than more careful types

The bug class is not "wall vs monotonic". It is **pairwise arithmetic between two independently
produced instants**. Better types narrow that; they do not close it, because a mixed pair is still
reachable and still has to answer.

So the fix is to stop producing the instants independently. A group that will be compared against
each other — the spans of a transaction, the samples of a profile chunk, the segments of a replay —
reads the epoch **once** and projects the rest through the monotonic clock:

```
epoch(t) = anchorEpoch + (tick(t) − anchorTick)
```

The projection is affine with slope 1, so subtracting two instants from one anchor *is* subtracting
two ticks. A duration is then monotonic by construction rather than by convention, and a clock step
cannot make a child span start before its parent. It also buys resolution the wall clock does not
have: Android's epoch is millisecond-granular, so a directly read instant is truncated while a
projected one carries nanoseconds. OpenTelemetry's SDK does the same thing, per local root span, for
the same two reasons.

`tickOf()` **refusing** an instant it did not project is what makes this safer rather than merely
tidier, and it is why `Timestamp` references its anchor at all: mixing domains raises an
`IllegalArgumentException` instead of returning a plausible-looking wrong number. An instant read
straight from the wall clock, or stated by something outside the process, has no anchor and can only
be serialized.

* resolves: JAVA-572 (partially — this is the additive, behaviour-free half)

### The epoch clock does not go through SentryDateProvider

`SystemEpochClock` reads the wall clock directly, picking precision the way `SentryAutoDateProvider`
does: `Instant.now()` on JVM 9+, `System.currentTimeMillis()` otherwise. Android is always the
latter — `Instant` is millisecond-granular there whether or not the build desugars it (#2451). The
values are byte-identical to what the provider returns on every platform; what changes is that
reading one no longer allocates a `SentryDate`, and on Android no longer takes a `System.nanoTime()`
reading that an `EpochClock` never looks at.

`setDateProvider` therefore does not reach the epoch clock. Faking time means overriding
`getEpochClock()` on `SentryOptions`, the way `SentryAndroidOptions` already overrides
`getMonotonicClock()`. There is no `setEpochClock` because nothing consumes it yet.

`./gradlew :sentry:test` — 3551 tests, 0 failures. `spotlessApply apiDump` clean; the `.api` diff
against the merge base is **additions only**, with no `<init>` leaks.

14 new tests on `AnchoredClock`, including the ones that were impossible to write before:

- step the epoch clock backwards and forwards *after* taking the anchor, and assert projected instants
  and the differences between them do not move
- `tickOf` inverts a projection exactly, and throws for a bare instant and for another anchor's instant
- a millisecond anchor still projects nanoseconds
- `driftNanos()` is zero while the wall clock keeps pace, and reports the signed size of a step

## :pencil: Checklist

## :crystal_ball: Next steps

`SentryTracer` creates one `AnchoredClock` per transaction, before its root `Span`, and every `Span`
below projects from that anchor. `Span` then holds two `Timestamp`s instead of two `SentryDate`s,
serialization reads `end().epochNanos()` instead of `laterDateNanosTimestampByDiff`, and the two
sentinel hacks call `anchor.tickOf(...)`. That retires `laterDateNanosTimestampByDiff`,
`SentryNanotimeDate`'s `diff`/`compareTo`/`nanotimeDiff` overrides,
`SentryAutoDateProvider`/`SentryInstantDate`, and two `SentryDate` allocations per span endpoint.

Two things that PR has to settle, flagged here so they get argued before code exists:

- **Clock base.** `Choreographer` hands us frame timestamps in the `System.nanoTime()` timebase, while
  `AndroidMonotonicClock` is `SystemClock.elapsedRealtimeNanos()`; the two differ by accumulated
  suspend. The plan is to keep a single `MonotonicClock` and put that last hop inside
  `SpanFrameMetricsCollector`, which already has an `onSpanStarted` hook to capture both readings and
  can detect sleep by comparing the deltas — skipping attribution honestly rather than returning a
  plausible-but-wrong projection.
- **v9 gating.** Eager projection moves every *child* span's `start_timestamp` by sub-millisecond
  amounts (root spans are unaffected — the anchor is read at root start). Under the "serialized values
  are frozen until the major" rule that puts the flip behind v9, even though it is an improvement.

Separately, JAVA-572's original subject — the tracer idle/deadline timeout — is stale on the timer half
(`SentryTracer` already uses `getTimerExecutorService()`), but "clamp the finish timestamp when the
deadline fires late" is still real and is the actual fix for the multi-hour `ui.load` artifact. It
wants its own ticket. `QueuedThreadPoolExecutor`'s wall-clock backoff is internal control flow, so it
can be fixed before the major, like #6030.

> :warning: **Merge this PR using a merge commit** (not squash), so the rest of the stack keeps a clean history.

### DIFF
--- a/sentry/src/main/java/io/sentry/SentryOptions.java
+++ b/sentry/src/main/java/io/sentry/SentryOptions.java
@@ -21,8 +21,10 @@ import io.sentry.metrics.DefaultMetricsBatchProcessorFactory;
 import io.sentry.metrics.IMetricsBatchProcessorFactory;
 import io.sentry.protocol.SdkVersion;
 import io.sentry.protocol.SentryTransaction;
+import io.sentry.time.EpochClock;
 import io.sentry.time.JavaMonotonicTicker;
 import io.sentry.time.MonotonicTicker;
+import io.sentry.time.SystemEpochClock;
 import io.sentry.transport.ITransport;
 import io.sentry.transport.ITransportGate;
 import io.sentry.transport.NoOpEnvelopeCache;
@@ -3059,6 +3061,19 @@ public class SentryOptions {
   @ApiStatus.Internal
   public void setDateProvider(final @NotNull SentryDateProvider dateProvider) {
     this.dateProvider.setValue(dateProvider);
+  }
+
+  /**
+   * Returns the wall clock, for stamping an instant that will be serialized.
+   *
+   * <p>Reports the same epoch as {@link #getDateProvider()}, but a {@link io.sentry.time.Timestamp}
+   * carries no {@link System#nanoTime()} tick of its own the way a {@link SentryNanotimeDate} does.
+   * Instants that will be subtracted from each other come from an {@link
+   * io.sentry.time.AnchoredClock} built on this and {@link #getMonotonicTicker()}.
+   */
+  @ApiStatus.Internal
+  public @NotNull EpochClock getEpochClock() {
+    return SystemEpochClock.getInstance();
   }
 
   /**
--- a/sentry/src/main/java/io/sentry/time/AnchoredClock.java
+++ b/sentry/src/main/java/io/sentry/time/AnchoredClock.java
@@ -1,0 +1,57 @@
+package io.sentry.time;
+
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * One wall-clock reading plus a monotonic ticker, producing timestamps that are safe to subtract
+ * from each other.
+ *
+ * <p>Wall clocks jump. Read the wall clock twice and the device may have changed it in between, so
+ * the gap between the two readings is not the time that actually passed — it can come out too
+ * short, too long, or negative. That is how a child span ends up starting before its parent.
+ *
+ * <p>So for a group of timestamps that will be compared with each other — the spans of a
+ * transaction, the samples of a profile chunk, the frames of a replay segment — read the wall clock
+ * only once (the anchor) and build the rest by adding the time a {@link MonotonicTicker} has
+ * measured since, which only ever moves forward. Gaps between those timestamps are then real
+ * elapsed time. Spans need exactly that: the protocol sends a start and an end timestamp and no
+ * duration, so the server subtracts them.
+ */
+@ApiStatus.Internal
+public final class AnchoredClock {
+
+  private final @NotNull MonotonicTicker ticker;
+  private final long epochNanos;
+  private final long anchorTick;
+
+  private AnchoredClock(
+      final @NotNull MonotonicTicker ticker, final long epochNanos, final long anchorTick) {
+    this.ticker = ticker;
+    this.epochNanos = epochNanos;
+    this.anchorTick = anchorTick;
+  }
+
+  /** Takes the anchor now: one wall-clock reading and one tick, back to back. */
+  public static @NotNull AnchoredClock create(
+      final @NotNull EpochClock epoch, final @NotNull MonotonicTicker ticker) {
+    return new AnchoredClock(ticker, epoch.now().epochNanos(), ticker.tickNanos());
+  }
+
+  /**
+   * The anchor itself: the one timestamp here that was read from the wall clock rather than built
+   * from a tick. Reads no clock and never changes.
+   */
+  public @NotNull Timestamp startTime() {
+    return Timestamp.ofEpochNanos(epochNanos);
+  }
+
+  /** Now: {@link #startTime()} plus the time the ticker has measured since the anchor. */
+  public @NotNull Timestamp now() {
+    return at(ticker.tickNanos());
+  }
+
+  private @NotNull Timestamp at(final long tickNanos) {
+    return Timestamp.ofEpochNanos(epochNanos + (tickNanos - anchorTick));
+  }
+}
--- a/sentry/src/main/java/io/sentry/time/EpochClock.java
+++ b/sentry/src/main/java/io/sentry/time/EpochClock.java
@@ -1,0 +1,20 @@
+package io.sentry.time;
+
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * The source of wall-clock {@link Timestamp}s (aka, instants).
+ *
+ * <p>Stamps a moment that will leave this process, an event, a breadcrumb, a session, and nothing
+ * else. This class should <b>not</b> be used to compute durations. For Durations, use {@link
+ * Stopwatch}, and a group of instants that will be subtracted from each other belongs to an {@link
+ * AnchoredClock}, which reads this once and projects the rest.
+ */
+@ApiStatus.Internal
+public interface EpochClock {
+
+  /** The current instant. Serialize it; do not subtract it from another one. */
+  @NotNull
+  Timestamp now();
+}
--- a/sentry/src/main/java/io/sentry/time/InstantEpochNanos.java
+++ b/sentry/src/main/java/io/sentry/time/InstantEpochNanos.java
@@ -1,0 +1,25 @@
+package io.sentry.time;
+
+import io.sentry.DateUtils;
+import java.time.Instant;
+import org.jetbrains.annotations.ApiStatus;
+
+/**
+ * Reads the epoch from {@link Instant}.
+ *
+ * <p>A class of its own so the reference to {@code java.time} is loaded only where {@link
+ * SystemEpochClock} decided to use it. Android's minSdk is below the API 26 that introduced {@code
+ * Instant}.
+ */
+@ApiStatus.Internal
+@SuppressWarnings("NewApi")
+final class InstantEpochNanos {
+
+  private InstantEpochNanos() {}
+
+  static long read() {
+    final Instant now = Instant.now();
+    // No long overflow until year 2262
+    return DateUtils.secondsToNanos(now.getEpochSecond()) + now.getNano();
+  }
+}
--- a/sentry/src/main/java/io/sentry/time/Timestamp.java
+++ b/sentry/src/main/java/io/sentry/time/Timestamp.java
@@ -1,0 +1,57 @@
+package io.sentry.time;
+
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * An instant on the wall clock, as nanoseconds since the Unix epoch.
+ *
+ * <p>Unlike a {@link MonotonicTicker} tick, a timestamp means something outside this process: it
+ * can be serialized, stored, and compared against a value from another machine.
+ *
+ * <p>It deliberately offers no arithmetic between instants. Subtracting two independent wall-clock
+ * readings gives a duration the device's clock can lengthen, shorten or make negative. Durations
+ * come from a {@link Stopwatch}, or from two instants an {@link AnchoredClock} projected from the
+ * same tick.
+ *
+ * <p>Nanoseconds since the epoch overflow a long in the year 2262.
+ */
+@ApiStatus.Internal
+public final class Timestamp {
+
+  private final long epochNanos;
+
+  private Timestamp(final long epochNanos) {
+    this.epochNanos = epochNanos;
+  }
+
+  public static @NotNull Timestamp ofEpochNanos(final long epochNanos) {
+    return new Timestamp(epochNanos);
+  }
+
+  public long epochNanos() {
+    return epochNanos;
+  }
+
+  @Override
+  public boolean equals(final @Nullable Object other) {
+    if (this == other) {
+      return true;
+    }
+    if (!(other instanceof Timestamp)) {
+      return false;
+    }
+    return epochNanos == ((Timestamp) other).epochNanos;
+  }
+
+  @Override
+  public int hashCode() {
+    return (int) (epochNanos ^ (epochNanos >>> 32));
+  }
+
+  @Override
+  public @NotNull String toString() {
+    return "Timestamp{epochNanos=" + epochNanos + '}';
+  }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 ### Internal
 
 - Add an internal `MonotonicTicker` abstraction with `Deadline` and `Stopwatch` primitives ([#6028](https://github.com/getsentry/sentry-java/pull/6028))
+- Add internal `Timestamp`, `EpochClock` and `AnchoredClock`, so related instants project from one wall-clock reading instead of each reading the clock ([#6045](https://github.com/getsentry/sentry-java/pull/6045))
 
 ## 8.55.0
 
--- a/sentry/api/sentry.api
+++ b/sentry/api/sentry.api
@@ -3707,6 +3707,7 @@ public class io/sentry/SentryOptions {
 	public fun getEnvelopeDiskCache ()Lio/sentry/cache/IEnvelopeCache;
 	public fun getEnvelopeReader ()Lio/sentry/IEnvelopeReader;
 	public fun getEnvironment ()Ljava/lang/String;
+	public fun getEpochClock ()Lio/sentry/time/EpochClock;
 	public fun getEventProcessors ()Ljava/util/List;
 	public fun getExecutorService ()Lio/sentry/ISentryExecutorService;
 	public fun getExperimental ()Lio/sentry/ExperimentalOptions;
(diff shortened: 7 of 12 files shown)


sentry/src/main/java/io/sentry/time/AnchoredClock.java
```diff
@@ -0,0 +1,101 @@
+package io.sentry.time;
+
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * One wall-clock reading pinned to one monotonic tick, from which related instants are projected.
```
0xadam-brown: If you're up for improving on the clanker, maybe something like: 

```java
 * An {@link EpochClock} and a {@link MonotonicClock} that together let us generate multiple {@link
 * Timestamp}s (aka, instants) that share a common epochal anchor.
 *
 * <p>Lets us:
 *
 * <ol>
 *   <li>safely create accurate interval timings (e.g., for spans) by avoiding clock drift, NTP
 *       resets, or other sources of skew that intervals are subject to when they're computed from
 *       multiple wall clock readings; and
 *   <li>produce intervals with nanosecond precision.
 * </ol>
 *
 * <p>(Item (2) especially helpful on Android, as nano-precise system time is only available for API
 * 33+.)
 *
 * <p>Because {@code AnchoredClock} uses a single epochal time source, the timestamps it generates
 * are accurate relative to one another, but are always subject to any initial inaccuracy of the
 * epochal anchor. The {@link #driftNanos} method lets you estimate that inaccuracy by comparing
 * against more current wall clock readings.
```

Up to you of course...

(If your curious, you can see the comment [here](https://github.com/getsentry/sentry-java/blob/8400a7f84a9777af044b359bbafde11d410c6893/sentry-android-sqlite/src/main/java/io/sentry/sqlite/DriverSpans.kt#L82) for one of the unexpected – and impactful – consequences of not having (2).)

sentry/src/main/java/io/sentry/time/AnchoredClock.java
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  /** Takes the anchor now: one epoch reading, one tick, as close together as a call allows. */
+  public static @NotNull AnchoredClock create(
+      final @NotNull EpochClock epoch, final @NotNull MonotonicClock clock) {
+    return new AnchoredClock(epoch, clock, epoch.now().epochNanos(), clock.tickNanos());
+  }
+
+  /** The anchor itself — the one instant here that was read rather than projected. */
+  public @NotNull Timestamp start() {
```
0xadam-brown: l: My vote would be for `started()` here, since we're returning a timestamp from the past rather than staring a timer.

sentry/src/main/java/io/sentry/time/AnchoredClock.java
```diff
@@ -0,0 +1,101 @@
...
+  public long tickOf(final @NotNull Timestamp timestamp) {
+    if (timestamp.anchor() != this) {
+      throw new IllegalArgumentException(
+          "Timestamp was not projected by this AnchoredClock: " + timestamp);
+    }
+    return anchorTick + (timestamp.epochNanos() - epochNanos);
+  }
+
+  /**
+   * How far this anchor's projection has fallen behind or ahead of the wall clock, in nanoseconds.
```
0xadam-brown: ...or really, the relative distance between the original anchor wall clock time and an updated reading of that wall clock. (To my ear, the current wording makes it sound like any returned value besides 0 means the anchor has become less accurate relative to "true time", when it could have been perfectly accurate to begin with and has since skewed.)

...which raises the more fundamental question: **How exactly do we expect folks to use this method?** Do we plan on taking a bunch of these measures and using a computed average (or whatever) to offset what we actually report to Relay?

sentry/src/main/java/io/sentry/time/AnchoredClock.java
```diff
@@ -0,0 +1,101 @@
...
+   *
+   * <p>Zero means the wall clock advanced by exactly the time this clock measured. Anything else is
+   * a clock step, or — where {@link MonotonicClock} and the wall clock disagree about suspend —
```
0xadam-brown: l: Not quite sure I understand what "a clock step" means, and I'm skeptical that we can claim "anything else" is caused by one of these two things. (Clankers gonna clank, apparently.)

sentry/src/main/java/io/sentry/time/AnchoredClock.java
```diff
@@ -0,0 +1,101 @@
...
+   * device sleep. Reads the epoch and the tick in the same order as {@link #create}, so the gap
+   * between the two reads biases the result the same way it biased the anchor.
+   */
```
0xadam-brown: l: Worth mentioning that this can return negative values.

sentry/src/main/java/io/sentry/time/EpochClock.java
```diff
@@ -0,0 +1,20 @@
+package io.sentry.time;
+
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * The source of wall-clock time.
```
0xadam-brown: Maybe:

```java
 * A source of wall-clock {@link Timestamp}s (aka, instants).
 *
 * <p>This class should <b>not</b> be used to compute durations. Instead, see {@link AnchoredClock}
 * if you need shareable timestamps for an interval's start and end point, or {@link Deadline} if
 * you need a monotonic endpoint. 
```

Again, up to you – but could be worth avoiding the clanker for these Javadocs as clarity is esp valuable for such central APIs.

sentry/src/main/java/io/sentry/time/InstantEpochNanos.java
```diff
@@ -0,0 +1,25 @@
...
+ * SystemEpochClock} decided to use it. Android's minSdk is below the API 26 that introduced {@code
+ * Instant}.
+ */
+@ApiStatus.Internal
+@SuppressWarnings("NewApi")
+final class InstantEpochNanos {
+
+  private InstantEpochNanos() {}
+
+  static long read() {
```
0xadam-brown, line 20: l: Thoughts about an API like this instead:

```
InstantUtils.toEpochNanos(Instant)
```

...fwiw, that^^ seems a bit more intuitive to me than `InstantEpochNanos.read()`, and it also lets callers supply their own Instant made at call time or at some other time.

Up to you of course.

sentry/src/main/java/io/sentry/time/Timestamp.java
```diff
@@ -0,0 +1,79 @@
...
+ * an {@link AnchoredClock} produced references that clock, which lets {@link AnchoredClock#tickOf}
+ * recover the tick it came from and reject instants it did not produce.
+ *
+ * <p>Nanoseconds since the epoch overflow a long in the year 2262.
+ */
+@ApiStatus.Internal
+public final class Timestamp {
+
+  private final long epochNanos;
+  private final @Nullable AnchoredClock anchor;
```
0xadam-brown: m: This is surprising to me, as I'd expect the dependency to be solely from AnchoredClock -> Timestamp, rather than circular. It's also an instance that can't be gc'd when it otherwise might be. 

Thoughts about living without this field? Or do we really need AnchoredClock.tickOf() (its sole caller at present via Timestamp.anchored())? 

Not sure about what you had in mind for tickOf() and anchored(), but maybe we could live with a general purpose tickOf() that returns a long based on its AnchoredClock without caring where it came from / trusting the caller?

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -527,6 +529,9 @@ public class SentryOptions {
   private final @NotNull LazyEvaluator<SentryDateProvider> dateProvider =
       new LazyEvaluator<>(() -> new SentryAutoDateProvider());
 
+  private final @NotNull LazyEvaluator<EpochClock> epochClock =
```
0xadam-brown: l: Curious as to why we evaluate lazily here but not for the monotonic clock exposed via SentryOptions?

sentry/src/main/java/io/sentry/SentryOptions.java
```diff
@@ -3061,6 +3066,19 @@ public void setDateProvider(final @NotNull SentryDateProvider dateProvider) {
...
+  /**
+   * Returns the wall clock, for stamping an instant that will be serialized.
+   *
+   * <p>Reports the same epoch as {@link #getDateProvider()}, but a {@link io.sentry.time.Timestamp}
+   * carries no {@link System#nanoTime()} tick of its own the way a {@link SentryNanotimeDate} does.
+   * Instants that will be subtracted from each other come from an {@link
+   * io.sentry.time.AnchoredClock} built on this and {@link #getMonotonicClock()}.
+   */
+  @ApiStatus.Internal
+  public @NotNull EpochClock getEpochClock() {
```
0xadam-brown, line 3075: m: Similarly to [here](https://github.com/getsentry/sentry-java/pull/6028/changes#r3957771463), do we want to expose this via SentryOptions, or just AnchoredClock instead?

0xadam-brown (commented):
Nice! 🥇 

Some comments for your consideration. (Apologies in advance for the doc suggestions, but hope they'll be helpful in a cut-and-paste way – seems worthwhile one-up the clanker because these APIs are so important.) Looking great though 💯

sentry/src/main/java/io/sentry/SentryOptions.java
runningcode: that's slop, thanks for pointing this out.

sentry/src/main/java/io/sentry/SentryOptions.java
runningcode: I understand that there's no PR right now that uses this so maybe it is confusing.

There are some uses cases where we want an epoch to serialize from `EpochClock.now()` for example in `Breadcrumb` and `SentryEvent`. Do you mean to not expose this and then all the callers grab a singleton directly from whichever implementation of `EpochClock` they need?

sentry/src/main/java/io/sentry/time/Timestamp.java
runningcode: This is a good point. 😄 There are two design decisions that impact this here. The second one we discussed on the call but I'll repeat it here.

Choice one: **Ability to verify we are comparing against the same anchor**
Without this, we don't have any way to check that two timestamps are being compared against the same anchor. Otherwise the timestamps are not relative to each other and the comparison is invalid.

Choice two: **Single anchor or multiple anchors**
An alternative design would have been where we have a single Anchor for the entire process. In that world, we don't need a way to check that we are comparing two Timestamps against the same anchor. We made the choice here to have multiple anchors. The idea was to make it easier to mesh with our current span design.

If it helps understand how the `anchor` method is used, check the usage here: https://github.com/getsentry/sentry-java/pull/6055/changes#diff-8fb51c60f3cd9eba198cd0702cf5b06a1b4e0ff676dccaa371840a44a5a5bd8eR334 

I didn't follow your point about GC, can you explain it to me?

If we do reverse the dependency, what data structure would we use to keep track of all the `Timestamp`s in clock and make it thread safe?

sentry/src/main/java/io/sentry/time/InstantEpochNanos.java
runningcode: The issue with callers supplying their own `Instant` is that it won't compile since we still have min SDK < 26. I also personally dislike classes named with `Utils` after working at Square ;)

sentry/src/main/java/io/sentry/time/EpochClock.java
runningcode: I think the wording on "will leave this process" is important. i'll add your point about not computing durations but otherwise i thought the existing comment was good.

sentry/src/main/java/io/sentry/time/AnchoredClock.java
runningcode: good point!

sentry/src/main/java/io/sentry/time/AnchoredClock.java
runningcode: This is a good point. I had a branch where I was using it to attach metadata to spans (we could also calculate sleep time by subtracting the difference of the two monotonic clocks) but both blow up the scope significantly. I will remove this!

sentry/src/main/java/io/sentry/time/AnchoredClock.java
runningcode: I'm following Guava's convention here where static methods have the `ed` ending and instance methods don't.
I do see your point that this isn't the best name though as it doesn't distinguish between this and `now()`.  I've updated to `origin()` to make it more clear what it does.

sentry/src/main/java/io/sentry/SentryOptions.java
0xadam-brown: > There are some uses cases where we want an epoch to serialize from EpochClock.now() for example in Breadcrumb and SentryEvent.

That sounds a good enough reason to expose this to me. (Just wanted to make sure we had considered whether hiding some of these abstractions was possible / desired. No concerns about exposing them if we have specific use cases supporting it.)

sentry/src/main/java/io/sentry/time/Timestamp.java
0xadam-brown: > I didn't follow your point about GC, can you explain it to me?

B/c each Timestamp has a reference to its AnchoredClock, no AnchoredClock can be gc'd until all the Timestamps it created are gone too. If a Timstamp escapes somewhere, the corresponding AnchoredClock has to stick around with it. 

> If we do reverse the dependency, what data structure would we use to keep track of all the Timestamps in clock and make it thread safe?

We could use an ID or token of some sort if we wanted to avoid passing the whole AnchoredClock to Timestamp. Even a marker object would do. Happy to leave it to your discretion – I was mainly surprised to see us passing AnchoredClock, but nothing that has to block us.

> Thoughts about living without this field? Or do we really need AnchoredClock.tickOf() 

Quoting my original question again just so it doesn't get lost in the back-and-forth: do we know that we need tickOf() at all? Put another way, do we want to support inverse mappings from Timestamps back to ticks / longs? 

I'm fine with that if we have a good use case or two (and thanks for providing the link in your last comment!). I just don't know (yet) how easily that example could be worked around vs how much we need the inverse mappings, as I haven't dug into the application PRs much. Feel free to disregard if we really do need it, but worth a second thought in case we can do without it. (Do java.time, kotlin.time, or Guava support Instant -> tick mappings? They'll be lots more insightful than I am, of course...)

0xadam-brown (approved):
Thanks! Responded to your major questions [here](https://github.com/getsentry/sentry-java/pull/6045/changes#r3958913765), but nothing that has to block us 👍

sentry/src/main/java/io/sentry/time/Timestamp.java
runningcode: You raise a good point about keeping the AnchoredClock longer than it should but it isn't a heavy object so it shouldn't cause an issue. An alternative design would be a singleton AnchoredClock that is alive for the whole process. We should be creating a ton and keeping them around so I don't see it becoming an issue.


Yeah the `tickOf` is used in that link above https://github.com/getsentry/sentry-java/pull/6055/changes#diff-8fb51c60f3cd9eba198cd0702cf5b06a1b4e0ff676dccaa371840a44a5a5bd8eR334

It is to avoid the hack that was previously there to compare to the sentinel along with the `diff does ¯\_(ツ)_/¯` in the comment.

```
date.diff(EMPTY_NANO_TIME);
```

Using `tickOf` makes it a lot more clear. It is essentially needed since Spans are anchored to wall clocks and we need the ticks to compare to the base anchor.

I'll remove it for now, and i can add it back in the 6045 PR that uses it. Thanks for the discussion.

sentry/src/main/java/io/sentry/time/AnchoredClock.java
runningcode: Thanks for the rewrite suggestion. I will incorporate some of your ideas and simplify this!

PR #6049, opened 2026-09-03
fix(compose): Prevent dangling spans in SentryTraced + tag composition outcome

0xadam-brown:
## :scroll: Description

PR repairs a few defects in the implementation of `SentryTraced` while aiming to maintain parity with the previous approach. In particular it:

1. Prevents dangling composition spans in cases where the Compose runtime abandons the composition. (We now delay creating the span until a SideEffect executes.)

2. Prevents dangling render spans in cases where drawContent() throws and the host app recovers. (Similarly, we delay creating the span until drawContent() completes successfully.)

Dangling spans remain attached to their parent transaction and can delay transaction capture if `waitForChildren` is enabled. They're also confusing in the Sentry dashboard because they look like poor performing compositions, when in fact they will have simply been abandoned in the ordinary course by the Compose runtime.

## Spans before vs after

I had my clanker exercise SentryTraced composables in our sample app to test span durations before this change vs after. Expand below to see the results (tl;dr they're comparable, which is what we want).

<details>

| Flow | Baseline | Latest rerun |
| --- | --- | --- |
| Stock MainActivity -> Integrations -> Compose | 11 `ui.compose` + 11 `ui.render` child spans | 21 `ui.compose` + 21 `ui.render` child spans |
| Extra `Integrations` spans in stock flow | No | Yes |
| `composition.result` tag | No | No |
| Child `ui.render` origin | `manual` | `manual` |
| ComposeActivity-first child span set | 7 `ui.compose` + 7 `ui.render` | 7 `ui.compose` + 7 `ui.render` |

_Note: We're likely seeing more spans after this PR (21 vs 11) because of a race that's always been present between when the root transaction becomes active vs when we start the composition span. We now start later, which means the transaction has more time to set itself up, and therefore fewer spans are lost. I plan to tackle this problem fully with the introduction of `LocalSentrySpan` in a subsequent PR._

| Span description | Baseline (ms) | Latest rerun (ms) | Delta (ms) |
| --- | ---: | ---: | ---: |
| Jetpack Compose Initial Composition | 106.549 | 113.869 | +7.320 |
| navhost | 10.408 | 16.849 | +6.441 |
| buttons_page | 43.769 | 28.652 | -15.117 |
| button_nav_github | 24.341 | 10.637 | -13.704 |
| button_nav_github_args | 1.702 | 0.755 | -0.947 |
| button_crash | 2.749 | 0.643 | -2.106 |
| button_dialog (first) | 2.671 | 0.770 | -1.901 |
| button_dialog (second) | 8.746 | 14.314 | +5.568 |

| Span description | Baseline (ms) | Latest rerun (ms) | Delta (ms) |
| --- | ---: | ---: | ---: |
| Jetpack Compose Initial Render | 7.311 | 10.561 | +3.250 |
| navhost | 7.311 | 10.561 | +3.250 |
| buttons_page | 7.006 | 10.377 | +3.371 |
| button_nav_github | 3.287 | 2.813 | -0.474 |
| button_nav_github_args | 0.535 | 1.163 | +0.628 |
| button_crash | 0.245 | 0.761 | +0.516 |
| button_dialog (first) | 0.251 | 0.937 | +0.686 |
| button_dialog (second) | 1.902 | 2.842 | +0.940 |
</details>

Unit tests and manual tests.

## :pencil: Checklist

## :crystal_ball: Next steps

Two more `SentryTraced` related PRs:

1. Behavior changes: Only report a composition span for the first successful composition + include origin info with the render span like we do with the composition span.
3. Fix the currently broken CompositionLocals that lead to the possibility of a stale parent transaction via a new `LocalSentrySpan` CompositionLocal.

### DIFF
--- a/sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
+++ b/sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
@@ -4,6 +4,7 @@ import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.BoxScope
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.Immutable
+import androidx.compose.runtime.SideEffect
 import androidx.compose.runtime.compositionLocalOf
 import androidx.compose.runtime.remember
 import androidx.compose.ui.ExperimentalComposeUiApi
@@ -11,6 +12,7 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.drawWithContent
 import io.sentry.ISpan
 import io.sentry.Sentry
+import io.sentry.SentryDate
 import io.sentry.SpanOptions
 import io.sentry.compose.SentryModifier.sentryTag
 
@@ -21,14 +23,6 @@ private const val OP_PARENT_RENDER = "ui.compose.rendering"
 private const val OP_RENDER = "ui.render"
 
 private const val OP_TRACE_ORIGIN = "auto.ui.jetpack_compose"
-
-@Immutable private class ImmutableHolder<T>(var item: T)
-
-private fun getRootSpan(): ISpan? {
-  var rootSpan: ISpan? = null
-  Sentry.configureScope { rootSpan = it.transaction }
-  return rootSpan
-}
 
 private val localSentryCompositionParentSpan = compositionLocalOf {
   ImmutableHolder(
@@ -62,6 +56,15 @@ private val localSentryRenderingParentSpan = compositionLocalOf {
   )
 }
 
+@Immutable internal class ImmutableHolder<T>(var item: T)
+
+/**
+ * Creates spans for tracking the time required to compose the wrapped [content], and a span for its
+ * initial draw.
+ *
+ * Spans are approximate and include work performed by any composables [content] invokes. Abandoned
+ * recompositions are ignored.
+ */
 @ExperimentalComposeUiApi
 @Composable
 public fun SentryTraced(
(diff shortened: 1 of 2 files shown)


sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
```diff
@@ -22,14 +24,63 @@ private const val OP_RENDER = "ui.render"
...
+) : RememberObserver {
+
+  override fun onRemembered() {
+    recordCompositionSpan(resultTag = TAG_VALUE_COMPOSITION_SUCCESSFUL)
+  }
+
+  override fun onForgotten() = Unit
+
+  override fun onAbandoned() {
+    recordCompositionSpan(resultTag = TAG_VALUE_COMPOSITION_ABANDONED)
```
0xadam-brown: I'm actually going to post a follow-up PR proposing that we only record a span for the first successful composition. 

That^^ seemed to be the original intent of [#2507](https://github.com/getsentry/sentry-java/pull/2507); it matches the current description we use with the composition spans' parent ("Jetpack Compose Initial Composition"); and I strongly suspect it's what developers care about (and that abandoned recompositions in particular aren't of interest, as they're an ordinary course artifact of Compose runtime).

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
```diff
@@ -70,32 +129,35 @@ public fun SentryTraced(
...
+  val alreadyRendered = remember { ImmutableHolder(false) }
   val baseModifier = if (enableUserInteractionTracing) modifier.sentryTag(tag) else modifier
 
+  val parentCompositionSpan = localSentryCompositionParentSpan.current.item
+  val parentRenderingSpan = localSentryRenderingParentSpan.current.item
+
+  val compositionStartTimestamp = Sentry.getCurrentScopes().options.dateProvider.now()
+  remember(parentCompositionSpan, tag, compositionStartTimestamp) {
+    CompositionSpanRecorder(parentCompositionSpan, tag, compositionStartTimestamp)
+  }
```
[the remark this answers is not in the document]
0xadam-brown: This is existing behavior. I'll be addressing it in a follow-on to this PR.

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
```diff
@@ -70,32 +73,72 @@ public fun SentryTraced(
...
-          } else {
-            null
-          }
-        drawContent()
-        firstRendered.item = true
-        renderSpan?.finish()
+        if (alreadyRendered.item) {
+          drawContent()
+        } else {
+          val renderStart = dateProvider.now()
```
runningcode: this should be measured using a monotonic clock.

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
```diff
@@ -70,32 +73,72 @@ public fun SentryTraced(
...
-  val firstRendered = remember { ImmutableHolder(false) }
-
+  val alreadyRendered = remember { ImmutableHolder(false) }
   val baseModifier = if (enableUserInteractionTracing) modifier.sentryTag(tag) else modifier
 
+  val parentCompositionSpan = localSentryCompositionParentSpan.current.item
+  val parentRenderingSpan = localSentryRenderingParentSpan.current.item
+  val dateProvider = Sentry.getCurrentScopes().options.dateProvider
+
+  val compositionStart = dateProvider.now()
```
runningcode: it seems to me that we measure compositionstart and compositionEnd even if `parentCompositionSpan` is null which means we end up creating the SentryNanotDatetime object and then throwing it away.
Like i said, I'm not a compose expert but i think we should try to avoid unecessary work here in the rendering path.

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
```diff
@@ -70,32 +73,72 @@ public fun SentryTraced(
...
+        }
       },
     propagateMinConstraints = true,
   ) {
     content()
   }
-  compositionSpan?.finish()
+  val compositionEnd = dateProvider.now()
+
+  if (parentCompositionSpan != null) {
```
runningcode: nit: recordCompositionSpan checks if `parentSpan` is null and does nothing already. i guess this just avoids creating the `SideEffect` so probably still makes sense.

0xadam-brown:
Sorry for the churn. Need to incorporate a few additional changes. Back shortly...

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
```diff
@@ -62,6 +56,15 @@ private val localSentryRenderingParentSpan = compositionLocalOf {
   )
 }
 
+@Immutable internal class ImmutableHolder<T>(var item: T)
```
runningcode, line 59: `ImmutableHolder` holds a `var` is that intentional? i see you just moved this so it isn't code you added but seems weird to me unless it some compose thing.

runningcode (approved):
LGTM! I'm not an expert in compose so feel free to ignore these comments

0xadam-brown:
Ready for review again. No changes from before – the other work was substantial enough that I'll create a separate PR for it.

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
markushi: If I remember correctly, this was the only trick back then to avoid any recomposition caused by `SentryTraced`. As we only do measurements, and those are tightly scoped to our use case / not exposed, they should not trigger unwanted recompositions.

E.g. the bool `alreadyRendered`, should not trigger a recomposition when set to `true`.

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
0xadam-brown: Good catch – updated in 5b2d3729b9b450aa10c9c0c9c95839450d4b3eb8 to avoid allocating dates unless we have a parent span.

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
0xadam-brown: > i guess this just avoids creating the SideEffect 

Exactly.

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
0xadam-brown: Good question. The mutability is intentional. We need to ask Compose to `remember` the value of our already-rendered flag across compositions, while also being able to update that flag from within the composition. That means we need a container type with a variable field. 

The "immutable" language comes from the fact that the class is marked with `@Immutable`, which is used by Compose to make optimization decisions about what needs to be updated when state changes and what can be left alone. For obvious reasons, that annotation is a bit misleading. I suspect it was originally added because we wanted to tell the runtime that changes to Sentry state shouldn't result in the user's code being re-composed (which is true, we don't want that to happen).

...that said, given that our type isn't tracked composable State, I'm not sure it's doing anything. I can look into removing it in one of the later PRs in this sequence.

markushi (approved):

PR #6051, opened 2026-09-03
chore(compose): Update SentryTraced to emit a span only for initial composition

0xadam-brown:
## :scroll: Description

PR updates SentryTraced to:

1. emit a span _only_ for initial composition, rather than for every composition of the wrapped composable; and
2. add an origin to `ui.render` spans, just like we already do for composition spans.

I've also cleaned up use of ImmutableHolder and renamed it to MutableRef in response to @runningcode's reasonable thoughts [here](https://github.com/getsentry/sentry-java/pull/6049#discussion_r3922636110).

These are behavior changes to follow the fixes in [#6049](https://github.com/getsentry/sentry-java/pull/6049).

⚠️ Item (1):

- matches the current description we apply to the composition spans' parent ("Jetpack Compose **_Initial_** Composition"); 
- helps prevent the composition parent from getting overcrowded (see the diagram in the SentryTraced KDoc for why that's the case);
- reflects (AFAICT) the original intent of [#2507](https://github.com/getsentry/sentry-java/pull/2507); and
- is what I'd expect developers to actually care about (chime in if you think differently!).

I had my clanker exercise SentryTraced composables in our sample app and verify the output.

## :pencil: Checklist

## :crystal_ball: Next steps

Fix the currently broken CompositionLocals that lead to the possibility of a stale parent transaction.

### DIFF
--- a/sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
+++ b/sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
@@ -3,7 +3,6 @@ package io.sentry.compose
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.BoxScope
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.Immutable
 import androidx.compose.runtime.SideEffect
 import androidx.compose.runtime.compositionLocalOf
 import androidx.compose.runtime.remember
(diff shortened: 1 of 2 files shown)


sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
```diff
@@ -25,45 +24,67 @@ private const val OP_RENDER = "ui.render"
...
+      SpanOptions().apply {
+        isTrimStart = true
+        isTrimEnd = true
+        isIdle = true
+      },
+    )
+    ?.apply { spanContext.origin = OP_TRACE_ORIGIN }
 }
 
-@Immutable internal class ImmutableHolder<T>(var item: T)
```
0xadam-brown, line 59: The `@Immutable` annotation wasn't doing anything here as ImmutableHolder isn't snapshot state and we never pass it as an argument to a composable function.

0xadam-brown: cc @markushi in case I'm overlooking something^^

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
```diff
@@ -73,9 +94,19 @@ public fun SentryTraced(
...
   content: @Composable BoxScope.() -> Unit,
 ) {
-  val alreadyRendered = remember { ImmutableHolder(false) }
   val baseModifier = if (enableUserInteractionTracing) modifier.sentryTag(tag) else modifier
 
+  val parentCompositionSpan = localSentryCompositionParentSpan.current
+  val parentRenderingSpan = localSentryRenderingParentSpan.current
+
+  val alreadyComposed = remember(parentCompositionSpan) { MutableRef(false) }
+  val alreadyRendered = remember(parentRenderingSpan) { MutableRef(false) }
```
0xadam-brown, line 103: Keying on the parent span doesn't actually change anything yet (though the optics are right) because our implementation of the CompositionLocals above caches the transaction seen by the first SentryTraced to enter the composition in the app's process and never lets it go. That means both that:

1. the parents never change (making our keying on the parent spans here an optics-only update for now), and 
2. once the root transaction finishes, all spans generated by all SentryTraced composables no-op for the remainder of the app process's life. 

I'll be fixing (2) in the next PR in this sequence, which will make the parent span keys meaningful.

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
```diff
@@ -73,9 +94,19 @@ public fun SentryTraced(
...
+  val dateProvider = Sentry.getCurrentScopes().options.dateProvider
+
+  // Only record spans if we have a parent for them.
+  val compositionStart =
+    if (!alreadyComposed.value) parentCompositionSpan?.let { dateProvider.now() } else null
+
   val parentCompositionSpan = localSentryCompositionParentSpan.current.item
   val parentRenderingSpan = localSentryRenderingParentSpan.current.item
   val dateProvider = Sentry.getCurrentScopes().options.dateProvider
```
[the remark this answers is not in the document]
0xadam-brown: Fixed in b9706dc47c5291471c77c2db3417484b95780499

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
runningcode: just for my understanding, is snapshot state something that causes a re-render?

runningcode (approved):
I'll give it a 👍 with the caveat that I'm not a compose expert.

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
0xadam-brown: That's right. Zach actually had a superb write up of it that I continue to hear referenced: [link](https://blog.zachklipp.com/introduction-to-the-compose-snapshot-system/)

Compose has a robust observable state system (usu referred to as "compose State", "compose Mutable State", or the "snapshot state system"; they're each technically different, but they're part of the same set of concepts). It's based on MVCC principles, hence talk of "snapshots".

The basic idea is that composable functions are subdivided into "recompose scopes" by the machinery the compose compiler plugin generates, and each of those scopes subscribes to the observables for the state it's aware of. If a scope observes a state change, it invalidates itself. When the Choreographer next ticks, the Recomposer recomposes all invalidated scopes (ie, invokes all of their composables, perhaps more than once, and ultimately writes any updates to the slot table).

PR #6057, opened 2026-09-05
fix(compose): Stop SentryTraced from reusing stale parent spans

0xadam-brown:
## :scroll: Description

PR fixes the stale parent problem that causes spans to be dropped from all SentryTraced instances for the entire app process after the initial active transaction finishes. 

Prior to this PR, SentryTraced used process-wide composition locals to bind the transaction active whenever the first SentryTraced for an app process entered the composition, and to reuse it for all SentryTraced instances thereafter. That meant all SentryTraced spans for the entire app would be dropped for the lifetime of the app process once the initial transaction finished.

This PR sets things right by having each SentryTraced composable request the current active transaction and update the generation of spans accordingly. 

Because we're no longer relying on a single transaction + parent span pair, SentryTraced now needs to manage the creation of possibly multiple parent span pairs, as the owning transaction updates. That logic lives in the new ParentSpans class.

(See above.)

## Measuring performance

Performance of SentryTraced is essentially unchanged, if not slightly better with the introduction of this PR. Expand below to see the details.

<details>

I had my clanker run 8 time trials, comparing `main` against this PR.

### Key Timing Results

| Suite | Metric | Main | PR | Notes |
|---|---|---:|---:|---|
| Standard 8-run | Stock flow `ui.compose` child count | 21 | 28 | PR still attaches the extra `ComposeActivity` child set in the stock multi-screen flow |
| Reduced-noise 8-run | Stock flow `ui.compose` child count | 21 | 28 | Same semantic difference remains under reduced-noise conditions |
| Standard 8-run | `ComposeActivity` `buttons_page` `ui.compose` avg ms | 41.406 | 34.949 | PR lower |
| Standard 8-run | `ComposeActivity` `button_nav_github` `ui.compose` avg ms | 20.250 | 14.905 | PR lower |
| Standard 8-run | `ComposeActivity` `button_dialog#2` `ui.compose` avg ms | 11.465 | 11.358 | Essentially flat |
| Reduced-noise 8-run | `ComposeActivity` `buttons_page` `ui.compose` avg ms | 29.628 | 25.104 | PR lower |
| Reduced-noise 8-run | `ComposeActivity` `button_nav_github` `ui.compose` avg ms | 14.878 | 12.046 | PR lower |
| Reduced-noise 8-run | `ComposeActivity` `button_dialog#2` `ui.compose` avg ms | 7.694 | 7.369 | Essentially flat |
| Reduced-noise 8-run | `ComposeActivity` `buttons_page` `ui.render` avg ms | 4.617 | 4.644 | Flat |
| Reduced-noise 8-run | `ComposeActivity` `navhost` `ui.render` avg ms | 4.753 | 4.872 | Flat |

### Executive Summary

| Metric | Main | PR | Takeaway |
|---|---:|---:|---|
| Stock flow `ui.compose` child count | 21 | 28 | PR still records the extra `ComposeActivity` child spans in the stock flow |
| Reduced-noise `buttons_page` `ui.compose` avg ms | 29.628 | 25.104 | No slowdown here; PR is slightly lower |
| Reduced-noise `button_nav_github` `ui.compose` avg ms | 14.878 | 12.046 | No slowdown here; PR is slightly lower |
</details>

New unit tests + I had my clanker run manual tests for both correctness and performance against the Compose and Nav2 sample apps.

## :pencil: Checklist


sentry-compose/build.gradle.kts
```diff
@@ -62,11 +62,14 @@ kotlin {
     }
     getByName("androidUnitTest") {
       dependencies {
+        implementation(libs.androidx.compose.foundation)
+        implementation(libs.androidx.compose.foundation.layout)
```
0xadam-brown, line 66: Fyi, new deps are test-only. 

`foundation` and `foundation-layout` are needed to keep the Robolectric Compose test runtime from picking up `navigation-compose`'s obsolete transitive foundation artifacts, which led to a `BoxKt.maybeCachedBoxMeasurePolicy(...)` linkage failure (given the `Box` used by `SentryTraced`).

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
```diff
@@ -4,74 +4,47 @@ import androidx.compose.foundation.layout.Box
...
 
-private const val OP_PARENT_RENDER = "ui.compose.rendering"
-private const val OP_RENDER = "ui.render"
+private const val DESCRIPTION_RENDER_PARENT = "Jetpack Compose Initial Render"
+private const val OP_RENDER_PARENT = "ui.compose.rendering"
+private const val OP_RENDER_CHILD = "ui.render"
 
 private const val OP_TRACE_ORIGIN = "auto.ui.jetpack_compose"
 
-private val localSentryCompositionParentSpan = compositionLocalOf {
```
0xadam-brown, line 26: These composition locals were the cause of the stale parent bug. 

The lambda here is executed when `localSentryCompositionParentSpan.current` is first read (ie, by the first SentryTraced in the app process to be invoked). CompositionLocal then caches that default value and returns it every time `.current` is called. Once the original transaction returned by getRootSpan() expired, all remaining spans generated by SentryTraced through the lifetime of the app process were dropped.

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
```diff
@@ -130,44 +102,137 @@ public fun SentryTraced(
...
+/**
+ * Weak holder of [OP_COMPOSITION_PARENT] and [OP_RENDER_PARENT] spans generated by [SentryTraced]
+ * composables.
+ *
+ * Instances are cached in a weak map keyed by owner span, so multiple [SentryTraced] composables
+ * with the same owner share one composition parent and one render parent.
+ *
+ * **Not threadsafe:** Access must be confined to Compose UI-thread callbacks.
+ */
+private class ParentSpans {
```
0xadam-brown: This class manages the creation of new parent spans as the owner span updates. It does so for all SentryTraced instances throughout the app process. (Process-wide logic and state lives in the companion object; per-SentryTraced state is implemented by the instance properties.) 

It has to make sure it tracks which parent spans have been created for which owner span, so it knows when new parent spans need to be created. But it doesn't own the lifecycles of either.

sentry-compose/src/androidUnitTest/kotlin/io/sentry/compose/SentryTracedTest.kt
```diff
@@ -0,0 +1,438 @@
...
+import org.robolectric.annotation.GraphicsMode
+
+@OptIn(ExperimentalComposeUiApi::class)
+@RunWith(AndroidJUnit4::class)
+@Config(sdk = [30])
+@GraphicsMode(GraphicsMode.Mode.NATIVE)
+class SentryTracedTest {
+
+  // workaround for robolectric tests with composeRule taken
+  // from https://github.com/robolectric/robolectric/pull/4736#issuecomment-1831034882
```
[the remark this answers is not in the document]
0xadam-brown: This is a false positive. My use of "taken from" triggered the alarm.

This is actually just a reference to a GitHub issue discussion where someone mentioned the workaround applied here. In particular:

- The code is not Robolectric source copied into Sentry.                                                                                                                                                  
- It's a small test setup pattern using Robolectric’s public test API: Shadows.shadowOf(...).addActivityIfNotPresent(...).                                                                               
- The cited PR/comment is explanatory context for why this setup is needed, not the source of a vendored implementation.                                                                                  
- THIRD_PARTY_NOTICES.md is meant for adapted third-party code, not every external issue/PR that explains a workaround. 

I've updated the language in 68f3f69c43b770ab4b169cf53a7f9b855485da86 to keep Warden happy.

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
```diff
@@ -130,44 +102,137 @@ public fun SentryTraced(
...
+      ownerSpan.startChild(
+        operation,
+        description,
+        SpanOptions().apply {
+          setStartTimestamp(startTimestamp)
+          isTrimStart = true
+          isTrimEnd = true
+          isIdle = true
+        },
+      )
```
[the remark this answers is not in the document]
0xadam-brown: Fixed in c0b7806b9337e578165f15b372cd0623f94ac4df

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
runningcode: Nice find and explanation!

sentry-compose/build.gradle.kts
runningcode: do you have a build-scan with this issue? I would be curious to examine the dependency tree when this happened to figure out where it came from.
could this happen to our customers?

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
```diff
@@ -130,44 +103,141 @@ public fun SentryTraced(
...
+ * (Process-wide logic and state lives in the companion object; per-SentryTraced state is
+ * implemented by the instance properties.)
+ *
+ * Under the hood this class tracks which parent spans have been created for which owner span, so it
+ * knows when new parent spans need to be created. But it doesn't own the lifecycle of either and
+ * holds only weak references.
+ *
+ * **Not threadsafe:** Access must be confined to Compose UI-thread callbacks.
+ */
+private class ParentSpans {
```
runningcode: I got confused between `Owner` and `Parent`. If I understand correctly the `Owner` is still the `Parent` of the `Parent` spans which is why it gets confusing. Can we call these buckets or something else? This way we have the `CompositionBucket` and the `RenderBucket` ?

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
```diff
@@ -130,44 +103,141 @@ public fun SentryTraced(
...
+
+  private var compositionParentSpan: WeakReference<ISpan>? = null
```
runningcode: it seems these are `WeakReferences` in order to avoid a circular reference back to the key.
example:
`ownerSpanToParentSpans` has a `ParentSpans` value which has a `compositionParentSpan` which holds a `SentryTracer` which in this case is the "owner span" which is also the map's key.

the code works as intended. there's no bug here. i'm just saying we should comment why it is a `WeakReference` here otherwise it isn't immediately obvious why these are `WeakReferences `and someone might try to change them to be strong references which means it'll never be cleaned up

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
```diff
@@ -4,74 +4,48 @@ import androidx.compose.foundation.layout.Box
...
+ * All spans produced are rooted under an owner span defined by the environment `SentryTraced` runs
+ * in. `SentryTraced` composables with the same owner share two common parent spans
+ * (`ui.compose.composition` and `ui.compose.rendering`). Each `SentryTraced` in the group emits at
+ * most one `ui.compose` span to the composition parent and one `ui.render` span to the render
+ * parent.
+ *
+ * The end result looks something like this:
  * ```
- * Root span
+ * Owner span
```
runningcode, line 48: I don't have a strong opinion on this change, but others might. I wonder if `Root span` is a Sentry-wide lingo that we should keep.

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
```diff
@@ -130,44 +103,141 @@ public fun SentryTraced(
...
   }
 }
 
-private fun getRootSpan(): ISpan? {
-  var rootSpan: ISpan? = null
-  Sentry.configureScope { rootSpan = it.transaction }
-  return rootSpan
-}
-
 private fun recordCompositionSpan(
```
runningcode, line 120:
```suggestion
private fun recordCompositionSpanForOwner(
```

just a thought since the param is named `ownerSpan` but you often don't check method parameter names.

runningcode (commented):
Looks good! I'll note that this is not a deep dive since I am not a compose expert. I added some nits and happy to discuss those if needed.

For my understanding, what are the units the benchmark you performed? and was it a macrobenchmark?

runningcode (approved):

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
0xadam-brown: I ended up renaming 'parents' -> 'buckets', but I kept 'owner'. 

I tried out 'root' but it too strongly suggested not just the span where you happen to decide to hang the bucket spans and their children, but the highest entry point in your entire span graph. (Admittedly, the latter is what we currently give you with auto-instrumentation, but it's not necessarily the experience of power users, and the proposed introduction of `LocalSentrySpan` would make it even less true, including possibly for auto-instrumentation.)

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
0xadam-brown: Ah, good thinking. I'm worried mentioning the owner in the method name will be confusing in its own way, since we record the composition span immediately under the bucket, not the owner. I've kept the names as-is but added KDocs to each method clarifying what's going on (and giving future clankers enough context).

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
0xadam-brown: Always appreciate the fresh eyes – updated to use 'bucket' rather than 'parent' 👍

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
0xadam-brown: Thanks for the callout – added a comment!

sentry-compose/build.gradle.kts
0xadam-brown: Good questions. No impact on customers b/c:

1. these are test deps, and our Compose deps in release builds are compileOnly;
2. Robolectric Compose uses significantly newer versions of Compose than we currently pin in libs.versions.toml;
3. the new test deps are brought in so that Robolectric Compose has the version of Compose's Box it assumes is available to it (as Robolectric makes use of those newer APIs).

The real solution here would be to consider bumping our Compose deps in libs.versions.toml. That's too much of a lift for this PR, but I'll take a look at it this week if I have a few spare cycles. (My clanker tells me we'd need to upgrade androidx.navigation:navigation-compose to at least 2.8.0 to make Robolectric happy. We're currently on 2.4.2, which was released in 2022.)

**Build scans**

**[a]** This PR without the new test deps: https://scans.gradle.com/s/23t7c6ow25r2e
**[b]** This PR with the new test deps: https://scans.gradle.com/s/iv75ao6ullh5a

Running the sentry-compose unit tests with [a] produces a NoSuchMethod error with every test: 

```
java.lang.AssertionError: Expected an exception of class java.lang.IllegalStateException to be thrown, but was java.lang.NoSuchMethodError: 'androidx.compose.ui.layout.MeasurePolicy androidx.compose.foundation.layout.BoxKt.maybeCachedBoxMeasurePolicy(androidx.compose.ui.Alignment, boolean)'
```

sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
runningcode: sounds great!

sentry-compose/build.gradle.kts
runningcode: Ready for a Gradle deep dive? Here we go....

As you say, we have a very old `androidx.navigation:navigation-compose:2.4.2` which brings `foundation-layout:1.0.1`. [Scan link here](https://scans.gradle.com/s/23t7c6ow25r2e/dependencies?focusedDependency=WzIsNzUsMzE5MyxbMiw3NSxbMTcwMSwzMTkzXV1d&toggled=W1syXSxbMiw3NV0sWzIsNzUsWzEzNTddXSxbMiw3NSxbMTM1NywxMzYzXV0sWzIsNzUsWzE3MDFdXSxbMiw3NSxbMTcwMSwzMTkzXV1d)
When you add the new dependencies, that bumps the `foundation-layout` to `1.7.0` which includes that method. [Scan link here](https://scans.gradle.com/s/iv75ao6ullh5a/dependencies?expandAll&focusedDependency=WzIsNzUsMzMzMyxbMiw3NSxbMTcwMywzMzMzXV1d&toggled=W1swXSxbMV0sWzJdLFsyXSxbMiw3NV0sWzIsNzUsWzMyODNdXSxbMiw3NSxbMTcwM11dXQ)

Surprisingly, we only specify version `1.6.3` of `foundation-layout` in our [libs.versions.toml](https://github.com/getsentry/sentry-java/blob/main/gradle/libs.versions.toml#L7) but you can see in the build scan that there is a conflict between `1.7.0, 1.6.3, 1.0.1 and 1.6.0` and we ended up resolving `1.7.0` which is luckly because my clanker tells me that `maybeCachedBoxMeasurePolicy` was introduced in `1.7.0`.

Scrolling around on the build scan [you can see here](https://scans.gradle.com/s/iv75ao6ullh5a/dependencies?expandAll&focusedDependency=WzIsNzUsMzI4MixbMiw3NSxbMzI4Ml1dXQ&toggled=W1swXSxbMV0sWzJdLFsyXSxbMiw3NV0sWzIsNzUsWzMyODNdXSxbMiw3NSxbMTcwM11dLFsyLDc1LFszMjgyXV0sWzIsNzUsWzMyODIsMzI4NV1dXQ) `Constraint: prevents a regression in Overscroll` which [comes from here](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/ui/ui/build.gradle;l=199?q=prevents%20regression%20in%20overscroll&sq=&ss=androidx) in the AndroidX source code.

That version constraint is ONLY applied to `foundation` and not `foundation-layout` and it comes from the `ui-android:1.9.0` artifact. So that essentially is saying if you are using `compose-ui` and you include `foundation`, then it must be at least `foundation 1.7.0`. Then `foundation` and `foundation-layout` are in the same atomic group so they are bumped together.


Ok so that was interesting. Here's the TL;DR:
* We are adding dependencies here so that the added dependencies use a version constraint to bump the version. We should instead bump the versions directly.
* We should bump compose from 1.6.3 since we never actually resolve that version. 
* We should definitely bump that 4 year old navigation-compose dependency to make it less fragile.
* Our code requires at least `androidx.compose.foundation:foundation-layout:1.7.0` so we should declare this constraint. Otherwise consumers will see the same error you saw in your test.

0xadam-brown: Excellent, and thanks for the deep dive! That tl;dr will come in handy. I'll see if I can't bump in a follow-on (and clean up the test deps here) 🙌

PR #6058, opened 2026-09-05
fix(compose): Ensure SentryTraced spans respect ignore-by-origin requests

0xadam-brown:
## :scroll: Description

Fixes a bug where we set span origins for SentryTraced after their creation via the span context, resulting in our ignore-span-origins logic not being able to see them and spans being produced when they should have been ignored / suppressed.

This issue has been with us for a while, but was discovered (along with a number of others) in connection with my Nav3 work. 

The general rule is that we need to set span origins during creation and not afterwards via span context or else checks list `ignoreSpanOrigins` will be ignored.

## What went wrong?

SentryTracer checks `ignoredSpanOrigins` during span creation, before the returned span can be mutated:

```java
if (SpanUtils.isIgnored(scopes.getOptions().getIgnoredSpanOrigins(), spanOptions.getOrigin())) {
  return NoOpSpan.getInstance();
}
```

That check reads SpanOptions.origin, not the SpanContext.origin we were setting after creation.

So if a host app did this...

```kotlin
options.setIgnoredSpanOrigins(listOf("auto.ui.jetpack_compose"))
```

...it would no-op and the Compose spans were created nonetheless.

The issue affected both levels of the SentryTraced span hierarchy (ie, both parent spans and child spans).

Unit regression test + I had my clanker verify the fix works and is correct via our Android sample apps.

## :pencil: Checklist

### DIFF
--- a/sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
+++ b/sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
@@ -125,10 +125,15 @@ private fun recordCompositionSpan(
 ) {
   val bucketSpan = BucketSpans.getOrCreateCompositionSpan(ownerSpan, startTimestamp) ?: return
 
-  bucketSpan.startChild(OP_COMPOSITION_SPAN, tag, startTimestamp).apply {
-    spanContext.origin = OP_TRACE_ORIGIN
-    finish(null, endTimestamp)
-  }
+  bucketSpan
+    .startChild(
+      OP_COMPOSITION_SPAN,
+      tag,
+      startTimestamp,
+      Instrumenter.SENTRY,
+      SpanOptions().apply { origin = OP_TRACE_ORIGIN },
+    )
+    .run { finish(null, endTimestamp) }
 }
 
 /**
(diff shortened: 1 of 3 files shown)


sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryComposeTracing.kt
```diff
@@ -117,10 +117,15 @@ private fun recordCompositionSpan(
...
-  parentSpan.startChild(OP_COMPOSITION_CHILD, tag, startTimestamp).apply {
-    spanContext.origin = OP_TRACE_ORIGIN
-    finish(null, endTimestamp)
-  }
+  parentSpan
+    .startChild(
+      OP_COMPOSITION_CHILD,
+      tag,
+      startTimestamp,
+      Instrumenter.SENTRY,
```
0xadam-brown, line 133: Note: We have to specify the instrumenter explicitly b/c the only `startChild()` method that permits setting both the start timestamp and SentryOptions requires it. (This doesn't change behavior, as SentryTraced has always started children via Instrumenter.SENTRY under the hood.)

runningcode (approved):
This is quite a subtle bug! :O  Is there a way of changing the API to make it impossible to do it the wrong way? Or throwing an exception when doing this the wrong way?

0xadam-brown:
Rebased on an updated target + fixed minor merge conflicts; otherwise unchanged.

0xadam-brown:
> This is quite a subtle bug! :O Is there a way of changing the API to make it impossible to do it the wrong way? Or throwing an exception when doing this the wrong way?

That's indeed the question. I also ran into a similar issue w/r/t setting the start timestamp for spans in another SentryTraced PRs. I think a deep dive into our ISpan API and the constraints driving its behavior could be worthwhile at some point...

PR #6059, opened 2026-09-05
chore(compose): Rename file for SentryTraced

0xadam-brown:
## :scroll: Description

Renamed `SentryComposeTracing.kt` -> `SentryTraced.kt` to match our usual conventions + make it easier to find SentryTraced in the IDE. 

Rename keeps binary compatibility with previous file name for Java clients.

Having file names match public composable APIs is especially helpful b/c composables are functions, not classes, so most IDEs don't have a type-based global search for them (but do let you search files by name).

## :pencil: Checklist

## :crystal_ball: Next steps

#skip-changelog

### DIFF
--- a/sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryTraced.kt
+++ b/sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryTraced.kt
@@ -1,3 +1,5 @@
+@file:JvmName("SentryComposeTracingKt")
+
 package io.sentry.compose
 
 import androidx.compose.foundation.layout.Box


sentry-compose/src/androidMain/kotlin/io/sentry/compose/SentryTraced.kt
```diff
@@ -1,3 +1,5 @@
+@file:JvmName("SentryComposeTracingKt")
```
0xadam-brown, line 1: Note: SDK Size Analysis is failing because this rename preserves the old top-level JVM class name via @file:JvmName("SentryComposeTracingKt"), and the size-test app’s R8 config strips Kotlin runtime deps so the shrinker can’t resolve kotlin.jvm.JvmName. 

This is a size-test classpath quirk, not a production/runtime breakage for Java callers.

runningcode (approved):

0xadam-brown:
Rebased on an updated target; otherwise unchanged.

0xadam-brown:
Closing this PR, as my clanker tells me the issue mentioned [here](https://github.com/getsentry/sentry-java/pull/6059/changes#r3942803689) will continue indefinitely if this PR is merged, ie, we'll cause the SDK Size Path job to fail perpetually. (Fwiw, I also notice that no files in sentry-java use `@file:JvmName`.) We can try again if/when we update the R8 rules for the SDK Size Path job.

PR #6060, opened 2026-09-05
chore(android): Add product list to Nav2 sample

0xadam-brown:
## :scroll: Description

Adds a new hide/show product list button to the Nav2Activity. For the Compose tab, doing so lets us observe the span-producing of multiple SentryTraced composables inside the same nav transaction. 

The list was also added to the Fragments tab to maintain UI parity.

Better manual test (and LLM) coverage for SentryTraced.

By hand.

## :pencil: Checklist

## :crystal_ball: Next steps

#skip-changelog

### DIFF
--- a/sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/navigation/Nav2RouteFragment.kt
+++ b/sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/navigation/Nav2RouteFragment.kt
@@ -68,6 +68,7 @@ class Nav2RouteFragment : Fragment() {
                 activity.navigateTo(Nav2Destination.ProductDetail("7", "product-list"))
               },
             ),
+          trailingContent = { addProductListItemsToggle() },
         )
 
       Nav2RouteNames.PRODUCT_DETAIL -> productDetailLayout(activity)
@@ -131,7 +132,6 @@ class Nav2RouteFragment : Fragment() {
             activity.navigateTo(Nav2Destination.Checkout(productId))
           },
         ),
-      trailingContent = { addProductDetailItemsToggle(productId) },
     )
   }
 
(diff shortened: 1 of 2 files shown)


sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/navigation/Nav2RouteFragment.kt
```diff
@@ -182,9 +188,121 @@ class Nav2RouteFragment : Fragment() {
...
+    } finally {
+      span?.finish()
+    }
+  }
+
+  private fun productListItemRow(itemNumber: Int): View =
+    LinearLayout(requireContext()).apply {
+      orientation = LinearLayout.HORIZONTAL
+      setPadding(12.dp)
+      setBackgroundColor(color(android.R.color.white))
```
runningcode, line 289: just for my understanding, in compose is there some way to get the themed background?
i imagine that hard coding to white wouldn't work for light/dark themes. i don't think we need to support that but Im just curious for my knowledge.

runningcode (approved):

sentry-samples/sentry-samples-android/src/main/java/io/sentry/samples/android/navigation/Nav2RouteFragment.kt
0xadam-brown: > in compose is there some way to get the themed background?

There is indeed. Normally that code looks something like this:

```kotlin
MyTheme {
  MyComposableContent()
}
```

...where `MyTheme` is a composable function that sets one or more CompositionLocals containing all of the adjustments applied by your theme, eg, primary and secondary colors, font, font sizes, etc. 

(`MyComposableContent()` is the composable that acts as the root of your composable tree, ie, all the screens managed at this root. Widget composables in those screens would read the values off of the relevant CompositionLocals and theme themselves with them.)

runningcode: Thank you! So adding the `MyTheme` wrapper will override the background color set as `android.R.color.white`?

PR #6090, opened 2026-09-10
fix(replay): Measure recording deadlines on a monotonic clock (JAVA-575)

### LINKED ISSUE #5578: Session Replay drives interval/window math from the wall clock
Audit finding **B3 — actual bug, MEDIUM** (manifests when the clock steps mid-recording).

Epoch is required for RRWeb payload timestamps (that part is correct), but the same wall values also drive windows/durations:

* Segment durations and the 1h max-session deadline are `now - startEpoch` diffs — `sentry-android-replay/src/main/java/io/sentry/android/replay/capture/SessionCaptureStrategy.kt:107` (segment) and `:127` (session duration).
* `BufferCaptureStrategy` trim-to-last-30s and `ReplayCache.createVideoOf` iterate epoch-millis windows; frame files are named by epoch millis.
* `ReplayGestureConverter.kt`: `timeOffset = now - touchMoveBaseline`.

A backward step mid-recording → frames "newer than now": trim can wipe valid frames, segment windows miss/duplicate frames, gesture offsets go negative. A forward step → premature 1h cutoff. NTP/carrier/user steps on phones are realistic.

Source: [JAVA-557](<https://linear.app/getsentry/issue/JAVA-557/clock-usage-audit-wall-vs-monotonic-time-issues-across-the-sdk>) §B3.

runningcode:
## :scroll: Description

SessionReplay has a 1 hour session duration cap and a 50ms `ACTION_MOVE` debounce. These were both derived from wall clocks and not monotonic clocks. This PR fixes both of them to use a monotonic clock.

Here's a nice little table my agent came up with the explain the changes:
| Site | Before | After |
|---|---|---|
| 1h `sessionDuration` cap | `frameTimestamp - replayStartTimestamp >= sessionDuration` | `Deadline.after(ticker, sessionDuration, MILLISECONDS)` |
| 50 ms `ACTION_MOVE` debounce | `lastCapturedMoveEvent + 50 > now` | `Deadline.after(ticker, 50, MILLISECONDS)` |

### What this PR deliberately does not change

Everything else JAVA-575 lists stays on the wall clock since these values are serialized:

- segment timestamps and durations (`SessionCaptureStrategy`, `BufferCaptureStrategy`)
- frame filenames and the `ReplayCache.createVideoOf` / `rotate` windows
- gesture `timeOffset`s

Since those are customer facing / serialized values, we'll fix them in v9. 

A backward wall-clock step mid-recording (NTP correction, carrier time, user change — all realistic on phones) makes frames look "newer than now"; a forward step makes a recording look older than it is. For the two sites converted here that means:

- **Forward step past the cap** → a healthy recording is stopped on the very next frame with `Session replay deadline exceeded (1h)`.
- **Backward step** → the recording never reaches its cap, and every move event is suppressed until the wall clock catches back up.

* resolves: #5578
* resolves: JAVA-575 (partially — the serialized half is v9)

Existing tests pass and new tests added:

- `onScreenshotRecorded does not stop replay when the wall clock jumps past the deadline`
- `onScreenshotRecorded stops replay when the wall clock steps backwards past the deadline`
- `wall clock stepping backwards does not suppress move events`
- `wall clock stepping forwards does not lift the debounce early`

## :pencil: Checklist

## :crystal_ball: Next steps

Fix the serialized values in v9 as mentioned above.


sentry-android-replay/src/main/java/io/sentry/android/replay/capture/SessionCaptureStrategy.kt
```diff
@@ -111,6 +114,7 @@ internal class SessionCaptureStrategy(
     // reflecting the exact time of when it was captured
     val currentConfig = recorderConfig
     val frameTimestamp = dateProvider.currentTimeMillis
+    val deadlineExceeded = replayDeadline?.hasPassed() == true
```
runningcode, line 117: we check here because this preserves the existing behavior since we were previously comparing against `frameTimestamp`

Issue #5576, opened 2026-06-22
Session-end timer runs on java.util.Timer


runningcode:
Audit finding **B2 — actual bug, MEDIUM**.

`LifecycleWatcher` schedules session end with `java.util.Timer` (`new Timer(true)` / `scheduleEndSession`, `sentry-android-core/src/main/java/io/sentry/android/core/LifecycleWatcher.java:106-122`), sharing B1's mechanics:

* Device sleeps within the 30s background window → session ends only at wake; `Session.end()` stamps wake time → inflated session durations in release health. Replay `stop()` and `ContinuousProfiler.close(false)` also run hours late.
* The foreground check `lastUpdatedSession + sessionIntervalMillis <= now` is a wall-clock interval → a clock step causes spurious or missed session rotation.

Source: [JAVA-557](<https://linear.app/getsentry/issue/JAVA-557/clock-usage-audit-wall-vs-monotonic-time-issues-across-the-sdk>) §B2.

PR #6096, opened 2026-09-11
fix(android): Decide session rotation on a monotonic clock (JAVA-573)

runningcode:
## :scroll: Description

Background context:
The SDK decides it is a new session if you leave the app for more than 30 seconds. So if you come back under 30 seconds you use the same session. More than 30 seconds you get a new session.

The problem is, yup, you guessed it, we're using a wall clock time to determine these 30 seconds. The wall clock could jump backwards or forwards within these 30 seconds and then we make an incorrect decision on whether or not create a new session.

This swaps the logic to use a monotonic clock and a `Deadline`.

The code is also simpler now. Before there were two places that would check to end the session. One was to check `lastUpdatedSession` against the `currentTimeMillis` which is the WallClock. The other was the `.schedule(endSession, sessionIntervalMillis)` which is a monotonic clock. They could both have different values.

**Not fixed here, deliberately:** the other half of JAVA-573. A session that slept through its background window is still stamped when the SDK *noticed* rather than when the window is due, so its duration is inflated by the suspend. That value is serialized which is why we aren't fixing it here. #6091 is a template for fixing it. Its `Expiry` is the shape to reuse, and probably wants lifting out of `SentryTracer` first.

* resolves: JAVA-573
* relates to: #5576

Part of the clock semantics hardening tracked in JAVA-557 §B2. Note that §B2's premise is stale on one point: `java.util.Timer` was already replaced by the shared timer executor in #5819, so this PR is about which clock the window is *measured* on, not about the timer type.

Its hard to reproduce deep sleep and clock pinning but the logic is tested with some new unit tests.

## :pencil: Checklist

No `.api` diff — `LifecycleWatcher` is package-private.

## :crystal_ball: Next steps

Stamping the session end at the instant the window is due, so a suspended session stops reporting the sleep as app usage.


sentry-android-core/src/main/java/io/sentry/android/core/LifecycleWatcher.java
```diff
@@ -61,40 +60,48 @@ public void onForeground() {
...
+   *
+   * <p>Used when no background window is pending, which means the session was started by SDK init
+   * rather than by leaving and returning to the app. Nothing captured a tick back then, and the
+   * only record of when the session started is {@link Session#getStarted()} — a wall-clock instant,
+   * because it is sent to Sentry. So this check stays on the wall clock, clock steps included.
+   *
+   * <p>TODO [MAJOR]: let a session remember the tick it started on, so this can use a {@link
+   * Deadline} too. That tick must not be serialized.
+   */
+  private boolean isSessionOnScopeStale() {
```
markushi, line 86: nice refactor, this will keep future timing changes more isolated 👍

sentry-android-core/src/main/java/io/sentry/android/core/LifecycleWatcher.java
```diff
@@ -61,40 +60,48 @@ public void onForeground() {
...
+          }
+        });
+    return stale.get();
   }
 
   // App went to background and triggered this callback after 700ms
   // as no new screen was shown
   @Override
   public void onBackground() {
-    final long currentTimeMillis = currentDateProvider.getCurrentTimeMillis();
```
markushi, line 95: ~doesn't this change the behavior of the code?~
~Previously `lastUpdatedSession` was set when starting a new session, or when the app goes to background.~
~So when startSession was called, it would only start a new session when `lastUpdatedSession + sessionIntervalMillis` was beyond now.~

All good, I see this is now handled via `sessionEnd`.

markushi (approved):
Nice!

Issue #5582, opened 2026-06-22
Remaining low-priority clock issues and clock-step regression coverage


runningcode:
Tracking issue for the remaining low-priority audit findings, plus the project-wide regression sweep. Split any item into its own issue if it grows.

**Regression coverage:** unit tests already inject `ICurrentDateProvider` fakes everywhere. This project should add cases simulating backward/forward clock steps and deep-sleep delays across the affected paths (tracer, session, replay, check-in) so the fixes are covered and future regressions are caught.

**Remaining low-priority findings (audit §C4–C7, all LOW):**

* **C4 — App-start anchor projection.** `TimeSpan` (uptime + wall anchor) captures the wall anchor once; `setStartedAt()` back-projects assuming no step since process start → NTP sync shortly after boot shifts app-start span timestamps relative to later-anchored spans.
* **C5 —** `SpanFrameMetricsCollector.toNanoTime()`**.** Re-anchors wall-based `SentryLongDate`s into the `nanoTime` base using the current offset; wrong by any step since the date was created, and across deep sleep → frames attributed to the wrong span windows.
* **C6 — Wall-clock TTLs/cleanup.** `HostnameCache` 5h TTL; `Sentry.classCreationTimestamp` vs `File.lastModified()` for profiling-traces cleanup; `CacheStrategy` envelope rotation ordered by `lastModified()`; `DefaultCompositePerformanceCollector` 30s auto-stop via wall diff + sampling on `java.util.Timer`.
* **C7 — Informational.** `Breadcrumb.compareTo` orders by captured `System.nanoTime()` (restored-from-disk breadcrumbs get fresh nanos at parse → cross-restart ordering is parse-order); cross-type `SentryDate` arithmetic (`SentryNanotimeDate` vs `SentryLongDate`) silently degrades to ms-precision wall math.

Source: [JAVA-557](<https://linear.app/getsentry/issue/JAVA-557/clock-usage-audit-wall-vs-monotonic-time-issues-across-the-sdk>) §C4–C7 + "Verification idea".

PR #6097, opened 2026-09-11
fix(core): Order breadcrumbs by their own timestamp (JAVA-579)

runningcode:
## :scroll: Description

`Breadcrumb.compareTo` ordered breadcrumbs purely by a `System.nanoTime()` reading taken in the constructor. That reading is meaningful only for breadcrumbs created in the same process run, and two paths violate that:

- **Deserialized breadcrumbs.** `Breadcrumb.fromMap` and `Breadcrumb.Deserializer` both go through the public `Breadcrumb(Date)` constructor, so a breadcrumb read back from disk or handed over by a hybrid SDK takes its tick at parse time. A breadcrumb recorded yesterday sorted as if it had just happened, and the order `CombinedScopeView` merges the three scopes into became parse order rather than recorded order.
- **Clones.** The copy constructor also took a fresh tick, so `Scope`'s breadcrumb cloning moved every copied breadcrumb to the end of the order.

This orders by the recorded `timestamp` instead, keeping the creation tick as the tie-breaker it was added for in #3355. Cloning also now carries the original's `creationTick`.

Separately, there was a second `Comparator` in `SentryClient`. This is now removed so that the Breadcrumb's compare method is the single source of truth.

Audit finding §C7 from the clock-usage audit: a `System.nanoTime()` value was driving an ordering decision across a process boundary, where it carries no meaning.

* resolves: JAVA-579 (partially — §C7 breadcrumb ordering only)
* part of: #5582

Four unit tests in `BreadcrumbTest`, covering live breadcrumbs sharing a timestamp, both deserialization paths (`fromMap` and `Breadcrumb.Deserializer`), and clones.

## :pencil: Checklist

Note for hybrid SDKs: breadcrumbs passed in via `Breadcrumb.fromMap` now sort by the timestamp they carry rather than by the moment they were handed over. That is the intended fix, but it does change the order hybrid breadcrumbs land in relative to native ones. Where a hybrid and a native breadcrumb share a millisecond, the hybrid one sorts after the native one, since it takes its tie-breaker at the moment it is handed over rather than when it was recorded.


sentry/src/main/java/io/sentry/Breadcrumb.java
```diff
@@ -59,21 +66,33 @@ public final class Breadcrumb implements JsonUnknown, JsonSerializable, Comparab
...
     this.timestamp = timestamp;
     this.timestampMs = null;
   }
 
+  /**
+   * A breadcrumb rebuilt from a serialized one — read back from disk, or handed over by a hybrid
+   * SDK. It carries the timestamp it was serialized with and no creation tick.
+   */
+  static @NotNull Breadcrumb deserialized(final @NotNull Date timestamp) {
+    return new Breadcrumb(timestamp, null);
```
adinauer, line 85: `h` I think we should fill in the `creationTick` here too. I'm worried about losing order on deserialization. The bug should still be fixed by preferring `timestamp` for the comparison and just using `creationTick` as tie breaker.

adinauer (commented):
We should always fill in `creationTick` as otherwise we risk a different bug.

sentry/src/test/java/io/sentry/BreadcrumbTest.kt
```diff
@@ -365,6 +366,46 @@ class BreadcrumbTest {
...
+    val sorted = listOf(third, first, second).sorted().map { it.message }
+
+    assertThat(sorted).containsExactly("first", "second", "third").inOrder()
+  }
+
+  @Test
+  fun `a deserialized breadcrumb is ordered by its own timestamp, not by when it was parsed`() {
+    val live = Breadcrumb(Date(1_600_000_000_000)).apply { message = "live" }
+    val restored =
+      Breadcrumb.fromMap(
```
0xadam-brown, line 385: l: Consider testing `Breadcrumb.deserialize()` as well, since we also call `.deserialized()` there.

0xadam-brown (approved):
A couple quick comments for your consideration, but no blockers save for responding to @adinauer's concern 👍

sentry/src/main/java/io/sentry/Breadcrumb.java
```diff
@@ -831,8 +850,21 @@ public void setUnknown(@Nullable Map<String, Object> unknown) {
...
+    // Creation ticks break the tie in the order they were actually recorded.
+    if (creationTick == null) {
+      // Deserialized, so from an earlier process run than anything holding a tick.
+      return o.creationTick == null ? 0 : -1;
+    }
+    if (o.creationTick == null) {
+      return 1;
+    }
+    return creationTick.compareTo(o.creationTick);
   }
```
0xadam-brown, line 868: m: Looks like we're doing the following: 

- if timestamps differ, sort by timestamps;
- if tied, put deserialized timestamps before live;
- if both live, use tick.

But the PR description (assuming I'm interpreting it correctly) mentions:

- if timestamps differ, sort by timestamps;
- if tied and both live, use tick;
- otherwise treat as equal and let stable sort preserve existing order.

The latter seems a bit preferable to me, but no super strong opinions / happy to defer to you. Just want to make sure we're choosing deliberately 👍

sentry/src/main/java/io/sentry/Breadcrumb.java
runningcode: This is a good point. Right now this will still work since all the sorting is stable. (They keep the same order that we passed in to it). I'll make this change though since it significantly reduces the churn in this PR.

sentry/src/test/java/io/sentry/BreadcrumbTest.kt
runningcode: do you mean the new static constructor? that's a good point, but I removed it to follow Alex's suggestion.

sentry/src/main/java/io/sentry/Breadcrumb.java
runningcode: Very good catch. I updated the PR to `always` create the creationTick as before so the diff is much smaller now.

sentry/src/main/java/io/sentry/Breadcrumb.java
```diff
@@ -832,6 +834,12 @@ public void setUnknown(@Nullable Map<String, Object> unknown) {
...
   @SuppressWarnings("JavaUtilDate")
   public int compareTo(@NotNull Breadcrumb o) {
+    final int byTimestamp = getTimestamp().compareTo(o.getTimestamp());
+    if (byTimestamp != 0) {
+      return byTimestamp;
+    }
+    // Timestamps are millisecond-granular, so breadcrumbs recorded in the same millisecond tie.
+    // nanos is only meaningful within a process run, which is all a tie-breaker has to cover.
     return nanos.compareTo(o.nanos);
   }
```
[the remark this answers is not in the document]
runningcode: this is so true. why did someone build a customer comparator there!!?

sentry/src/main/java/io/sentry/SentryClient.java
```diff
@@ -1851,13 +1849,4 @@ private boolean sample() {
     }
     return true;
   }
-
-  private static final class SortBreadcrumbsByDate implements Comparator<Breadcrumb> {
-
```
runningcode, line 1856: This is such a footgun. I'm glad Seer found this.

adinauer (approved):
LGTM, thanks for updating.
Good catch on the `SortBreadcrumbsByDate`!

